PR #85, opened 2015-06-29
Minor reformatting to installation instructions

mbookman:
Update to README.md and installing.md to highlight supported platforms and installation prerequisites.

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -1,7 +1,5 @@
 
-# Prerequisites
-
-## Operating system
+# Operating systems
 
 gcsfuse has been tested successfully with the following operating systems:
 
@@ -9,7 +7,6 @@ gcsfuse has been tested successfully with the following operating systems:
 *   OS X (minimum version 10.10.2)
 
 It may or may not work correctly with other operating systems and older versions.
-
 ___
 ### OS X fast track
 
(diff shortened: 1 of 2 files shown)


docs/installing.md
```diff
@@ -1,4 +1,17 @@
...
+
+gcsfuse has been tested successfully with the following operating systems:
+
+*   Linux (minimum kernel version 3.10)
+*   OS X (minimum version 10.10.2)
+
+It may or may not work correctly with other operating systems and older versions.
+
+___
+### OS X fast track
```
jacobsa: Thanks for doing this, but it doesn't make sense to nest the "fast track" instructions under "Prerequisites", since they take care of the prerequisites for you. If the problem is that the operating systems aren't up front, I suggest making "Operating Systems" a top-level section, then the fast track section, then prerequisites.

mbookman:
Updated. Let me know what you think.

docs/installing.md
```diff
@@ -1,4 +1,14 @@
...
+
+# Operating systems
+
+gcsfuse has been tested successfully with the following operating systems:
+
+*   Linux (minimum kernel version 3.10)
+*   OS X (minimum version 10.10.2)
+
+It may or may not work correctly with other operating systems and older versions.
+___
```
jacobsa: Remove the line here? It looks odd, especially if the thing below is supposed to be a sub-section.

mbookman: Removed both horizontal lines around the OS X fast track section.

jacobsa:
Thanks!

PR #91, opened 2015-07-09
Fixes #89 - Make help text prettier

### LINKED ISSUE #89: gcsfuse help is pretty ugly
### How gcsfuse shows help

```
jjg@jjg1:~/projects$ gcsfuse
Usage: gcsfuse [flags] bucket_name mount_point

Flags:
  -dir-mode=493: Permissions bits for directories. Default is 0755.
  -file-mode=420: Permissions bits for files. Default is 0644.
  -fuse.debug=false: Write FUSE debugging messages to stderr.
  -fuse.trace_by_pid=false: Enable a hacky mode that uses reqtrace to group all ops from each individual PID. Not a good idea to use in production; races, bugs, and resource leaks likely lurk.
  -fuseutil.random_delays=false: If set, randomly delay each op received, to help expose concurrency issues.
  -gcs-chunk-size=16777216: If set to a non-zero value N, split up GCS objects into multiple chunks of size at most N when reading, and do not read or cache unnecessary chunks.
  -gcs.debug=false: Write FUSE debugging messages to stderr.
  -gid=-1: If non-negative, the GID that owns all inodes. The default is the GID of the gcsfuse process.
  -httputil.debug=false: Dump information about HTTP requests.
  -implicit-dirs=false: Implicitly define directories based on their content. See docs/semantics.md.
  -limit-bytes-per-sec=-1: If positive, a limit on the GCS -> gcsfuse bandwidth for reading objects, measured over a 30-second window.
  -limit-ops-per-sec=5: If positive, a limit on the rate at which we send requests to GCS, measured over a 30-second window.
  -o=map[]: Additional system-specific mount options. Be careful!
  -reqtrace.enable=false: Collect and print traces.
  -stat-cache-ttl=1m0s: How long to cache StatObject results from GCS.
  -syncutil.check_invariants=false: Crash when registered invariants are violated.
  -temp-dir="": The temporary directory in which to store local copies of GCS objects. If empty, the system default (probably /tmp) will be used.
  -temp-dir-bytes=2147483648: A desired limit on the number of bytes used in --temp-dir. May be exceeded for dirty files that have not been flushed or closed.
  -type-cache-ttl=1m0s: How long to cache name -> file/dir type mappings in directory inodes.
  -uid=-1: If non-negative, the UID that owns all inodes. The default is the UID of the gcsfuse process.
```

### How a typical binary shows help

```
jjg@jjg1:~/projects$ cat --help
Usage: cat [OPTION]... [FILE]...
Concatenate FILE(s), or standard input, to standard output.

  -A, --show-all           equivalent to -vET
  -b, --number-nonblank    number nonempty output lines, overrides -n
  -e                       equivalent to -vE
  -E, --show-ends          display $ at end of each line
  -n, --number             number all output lines
  -s, --squeeze-blank      suppress repeated empty output lines
  -t                       equivalent to -vT
  -T, --show-tabs          display TAB characters as ^I
  -u                       (ignored)
  -v, --show-nonprinting   use ^ and M- notation, except for LFD and TAB
      --help     display this help and exit
      --version  output version information and exit

With no FILE, or when FILE is -, read standard input.

Examples:
  cat f - g  Output f's contents, then standard input, then g's contents.
  cat        Copy standard input to standard output.

Report cat bugs to bug-coreutils@gnu.org
GNU coreutils home page: <http://www.gnu.org/software/coreutils/>
General help using GNU software: <http://www.gnu.org/gethelp/>
For complete documentation, run: info coreutils 'cat invocation'
```

### Key issues to fix

- [ ] No short summary of what the binary does (`cat` has "Concatenate FILE(s), or standard input, to standard output.")
- [x] No proper spacing
- [x] No short and long names (`cat` has `-A, --show-all`)
- [x] Long names use underscores instead of hyphens (`bucket_name` instead of `bucket-name` or just `bucket`)
- [ ] Dots in parameters (`fuse.debug`) ...?
- [x] Defaults list as `=` instead of `Default: ...`
- [ ] Lower-case placeholders (`[flags]` instead of `[FLAGS]`)
- [ ] No examples (ie, `cat f - g  Output f's contents, then standard input, then g's contents.`)
- [ ] No end summary (where to report bugs, links to pages, etc)

jgeewax:
New output:

```
(master)jjg@jjg1:~/projects/gcsfuse$ go build . && ./gcsfuse --help
NAME:
   gcsfuse - Mount a GCS bucket locally

USAGE:
   gcsfuse [global options] bucket mountpoint
   
VERSION:
   0.1.0
   
GLOBAL OPTIONS:
   --dir-mode, -d               Permissions bits for directories. (default: 0755)
   --file-mode, -f              Permission bits for files (default: 0644)
   --gcs-chunk-size             Max chunk size for loading GCS objects. (default: 16777216)
   --gid                        GID owner of all inodes.
   --implicit-dirs              Implicitly define directories based on content. Seedocs/semantics.md
   --limit-bytes-per-sec        Bandwidth limit for reading data, measured over a 30-second window. (use -1 for no limit) (default: -1)
   --limit-ops-per-sec          Operations per second limit, measured over a 30-second window (use -1 for no limit) (default: 5)
   --mount-options, -o          Additional system-specific mount options. Be careful!
   --stat-cache-ttl             How long to cache StatObject results from GCS. (default: "1m0s")
   --type-cache-ttl             How long to cache name -> file/dir mappings in directory inodes. (default: "1m0s")
   --temp-dir, -t               Temporary directory for local GCS object copies. (default: system default, likely /tmp)
   --temp-dir-bytes             Size limit of the temporary directory. (default: 2147483648)
   --uid                        UID owner of all inodes.
   --help, -h                   print this help text
   --version, -v                print the version
```

I'm using a fork of https://github.com/codegangsta/cli that has a couple extra features. Once we get those merged into the main stream, we can switch this to his version.

Old output:

```
(master)jjg@jjg1:~/projects/gcsfuse$ gcsfuse --help
Usage: gcsfuse [flags] bucket_name mount_point

Flags:
  -dir-mode=493: Permissions bits for directories. Default is 0755.
  -file-mode=420: Permissions bits for files. Default is 0644.
  -fuse.debug=false: Write FUSE debugging messages to stderr.
  -fuse.trace_by_pid=false: Enable a hacky mode that uses reqtrace to group all ops from each individual PID. Not a good idea to use in production; races, bugs, and resource leaks likely lurk.
  -fuseutil.random_delays=false: If set, randomly delay each op received, to help expose concurrency issues.
  -gcs-chunk-size=16777216: If set to a non-zero value N, split up GCS objects into multiple chunks of size at most N when reading, and do not read or cache unnecessary chunks.
  -gcs.debug=false: Write FUSE debugging messages to stderr.
  -gid=-1: If non-negative, the GID that owns all inodes. The default is the GID of the gcsfuse process.
  -httputil.debug=false: Dump information about HTTP requests.
  -implicit-dirs=false: Implicitly define directories based on their content. See docs/semantics.md.
  -limit-bytes-per-sec=-1: If positive, a limit on the GCS -> gcsfuse bandwidth for reading objects, measured over a 30-second window.
  -limit-ops-per-sec=5: If positive, a limit on the rate at which we send requests to GCS, measured over a 30-second window.
  -o=map[]: Additional system-specific mount options. Be careful!
  -reqtrace.enable=false: Collect and print traces.
  -stat-cache-ttl=1m0s: How long to cache StatObject results from GCS.
  -syncutil.check_invariants=false: Crash when registered invariants are violated.
  -temp-dir="": The temporary directory in which to store local copies of GCS objects. If empty, the system default (probably /tmp) will be used.
  -temp-dir-bytes=2147483648: A desired limit on the number of bytes used in --temp-dir. May be exceeded for dirty files that have not been flushed or closed.
  -type-cache-ttl=1m0s: How long to cache name -> file/dir type mappings in directory inodes.
  -uid=-1: If non-negative, the UID that owns all inodes. The default is the UID of the gcsfuse process.
```

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -15,11 +15,100 @@
 package main
 
 import (
-	"flag"
 	"time"
 
 	mountpkg "github.com/googlecloudplatform/gcsfuse/mount"
+	"github.com/jgeewax/cli"
 )
+
+func getApp() (app *cli.App) {
+	app = cli.NewApp()
+	app.Name = "gcsfuse"
+	app.Usage = "Mount a GCS bucket locally"
+	app.ArgumentUsage = "bucket mountpoint"
+	app.HideHelp = true
+	app.Version = "0.1.0"
+	app.Flags = []cli.Flag{
+		cli.IntFlag{
+			Name: "dir-mode, d",
+			Usage: "Permissions bits for directories. (default: 0755)",
+			HideDefault: true,
+		},
+		cli.IntFlag{
+			Name: "file-mode, f",
+			Value: 0644,
+			Usage: "Permission bits for files (default: 0644)",
+			HideDefault: true,
+		},
+		cli.IntFlag{
+			Name: "gcs-chunk-size",
+			Value: 1<<24,
+			Usage: "Max chunk size for loading GCS objects.",
+		},
+		cli.IntFlag{
+			Name: "gid",
+			Value: -1,
+			HideDefault: true,
+			Usage: "GID owner of all inodes.",
+		},
+		cli.BoolFlag{
+			Name: "implicit-dirs",
+			Usage: "Implicitly define directories based on content. See" +
+			"docs/semantics.md",
+		},
+		cli.Float64Flag{
+			Name: "limit-bytes-per-sec",
+			Value: -1,
+			Usage: "Bandwidth limit for reading data, measured over a 30-second " +
+			"window. (use -1 for no limit)",
+		},
+		cli.Float64Flag{
+			Name: "limit-ops-per-sec",
+			Value: 5.0,
+			Usage: "Operations per second limit, measured over a 30-second window " +
+			"(use -1 for no limit)",
+		},
+		cli.StringFlag{
+			Name: "mount-options, o",
+			HideDefault: true,
+			Usage: "Additional system-specific mount options. Be careful!",
+		},
+		cli.DurationFlag{
+			Name: "stat-cache-ttl",
+			Value: time.Minute,
+			Usage: "How long to cache StatObject results from GCS.",
+		},
+		cli.DurationFlag{
+			Name: "type-cache-ttl",
+			Value: time.Minute,
+			Usage: "How long to cache name -> file/dir mappings in directory " +
+			"inodes.",
+		},
+		cli.StringFlag{
+			Name: "temp-dir, t",
+			Value: "",
+			HideDefault: true,
+			Usage: "Temporary directory for local GCS object copies. " +
+			"(default: system default, likely /tmp)",
+		},
+		cli.IntFlag{
+			Name: "temp-dir-bytes",
+			Value: 1<<31,
+			Usage: "Size limit of the temporary directory.",
+		},
+		cli.IntFlag{
+			Name: "uid",
+			Value: -1,
+			HideDefault: true,
+			Usage: "UID owner of all inodes.",
+		},
+		cli.BoolFlag{
+			Name: "help, h",
+			Usage: "print this help text",
+		},
+	}
+	return app
+}
 
 type flagStorage struct {
 	MountOptions                       map[string]string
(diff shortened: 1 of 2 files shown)


jgeewax:
I think some of these are missing -- not sure where they were coming from as they weren't in flags.go... Everything in flagStorage is populated, so not sure where I'd put things like fuse.debug, etc.

jacobsa:
The dotted names come from other packages; for example `fuse.debug` comes from [here](https://github.com/jacobsa/fuse/blob/069fff34a5ce9adba098672b60e84ffdb3056522/debug.go#L26). In general I think it's bad form to stick flags in reusable libraries, but I've used it exclusively for debugging options, which I think is sane.

Do you have any opinion on what to do about these? I quite like this change otherwise; this is the biggest sticking point for me. (The only other one: I'm not sure some of the options are worthy of short names; in particular `-f` and `-d` might be better reserved for some other use in the future.)

jgeewax:
> The dotted names come from other packages

Any chance you can point me at where exactly those get intercepted and passed along? I had some trouble finding that... (Or maybe I just spaced and missed it...)

I'm fine with keeping them - just need to know where exactly they're coming from...

> I'm not sure some of the options are worthy of short names; in particular -f and -d might be better reserved for some other use in the future.

That's a great point -- I'll chop those out.

jacobsa:
> Any chance you can point me at where exactly those get intercepted and passed along? I had some trouble finding that... (Or maybe I just spaced and missed it...)

Sorry, I'm not sure I understand the question. Are you asking where they are defined? Here is a list that I think is exhaustive (please confirm):

* [`fuse.debug`](https://github.com/jacobsa/fuse/blob/069fff34a5ce9adba098672b60e84ffdb3056522/debug.go#L26-L29)
* [`fuse.trace_by_pid`](https://github.com/jacobsa/fuse/blob/069fff34a5ce9adba098672b60e84ffdb3056522/connection.go#L34-L39)
* [`fuseutil.random_delays`](https://github.com/jacobsa/fuse/blob/069fff34a5ce9adba098672b60e84ffdb3056522/fuseutil/file_system.go#L28-L30)
* [`gcs.debug`](https://github.com/jacobsa/gcloud/blob/8289aa9f1290a6dead04fa77e9c29b7f0fe74cb7/gcs/debug_bucket.go#L30-L33)
* [`httputil.debug`](https://github.com/jacobsa/gcloud/blob/8289aa9f1290a6dead04fa77e9c29b7f0fe74cb7/httputil/debugging_round_tripper.go#L26-L29)
* [`reqtrace.enable`](https://github.com/jacobsa/reqtrace/blob/e4daabf6acca9c74b7efc8d786251336313c829e/reqtrace.go#L26)
* [`syncutil.check_invariants`](https://github.com/jacobsa/syncutil/blob/47ae5379a1cd4265867a8748ed01355df33e91a7/invariant_mutex.go#L22)

They are not passed along; in each case they're only used by the package (and usually file) in which they're defined.

jgeewax:
Sorry let me clarify. The process I see from our code in main.go is:

1. Parse flags from the command-line
1. Put flags into a flagStorage struct
1. Pass that struct along to the `mount()` method (https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/main.go#L165)

So... If I add these other flags to the list, what exactly do I do with them? There is no place for them in the flagStorage struct, and I don't see where we pass them along...?

jgeewax:
@jacobsa : Hey I don't think I'll have time to rebase and fix all of this in the next day -- but Jay says he needs it ASAP for some announcement stuff. Can you steal this PR from me ?

jacobsa:
The debug flags aren't propagated via the struct. They're defined in one package and used in only that package, and are magically set by the call to `flag.Parse` in the `main` function.

I'm still not sure what can/should be done about them in the context of this larger change of how flags work, so I don't know if I can steal the whole PR. But I'll make the help text look prettier using the existing package in time for Jay.

jacobsa:
Actually, that's probably harder. Will make this work; details to follow in #89.

Minecraftly:
Interesting, what if I want to mount this for Minecraft server world with lots of .dat files, which flags should I consider?

P/S: Currently I get lots of "(LookUpInode(parent=2, name="players")) error: no such file or directory" and "(*bazilfuse.GetxattrRequest(inode=6)) error: function not implemented" errors, not sure where they are from.

jgeewax:
Hi @Minecraftly 

I'm not sure familiar with the technical pieces of running a Minecraft server, but I'd suspect that it's acting similarly to how a database might act. If that is the case, GCS Fuse might not be the best way to do this. There's actually a the note on the [GCS Fuse docs page](https://cloud.google.com/storage/docs/gcs-fuse?hl=en#notes):

> Small random reads are slow due to latency to first byte (don't run a database over GCS Fuse!)

You might want to explore using PD-SSD to get great performance and high data durability, and I suspect that these would be at least comparable in cost.

I also just created a `gcsfuse` tag on StackOverflow -- feel free to ask questions like these over there too :)

PR #181, opened 2016-08-05
Update README.md

benchong408:
Clarify "status".

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@ Storage][gcs].
 
 # Current status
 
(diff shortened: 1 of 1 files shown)


jacobsa:
I'm not sure it makes sense to address such rumors on the front page of the project -- that seems to be to be unnecessarily (and fishily) defensive. Maybe it's better to retitle that issue and post a definitive update there? And also to update the Google-side documentation that gave people this idea.

benchong408:
Sounds good.

PR #182, opened 2016-08-13
Support for key_file fstab option

vaygr:
This adds basic support for `key_file` fstab option, which can be used for mounting with specific credentials (like for cross-project buckets and VMs).

It's quite useful, because `mount.gcsfuse` doesn't respect `GOOGLE_APPLICATION_CREDENTIALS` env var, but only `gcsfuse` does. But even if it did, it would be quite hacky to make it work under systems like Puppet, for example.

The option simplifies this by a lot:

```
my-bucket /mount/point gcsfuse rw,noauto,user,key_file=/path/to/key.json
```

it re-uses already existing `--key-file` argument for `gcsfuse`.

### DIFF
--- a/docs/mounting.md
+++ b/docs/mounting.md
@@ -22,6 +22,10 @@ Console:
 
     GOOGLE_APPLICATION_CREDENTIALS=/path/to/key.json gcsfuse [...]
 
+On OS X and Linux you can also use fstab option in `/etc/fstab` system file:
+
+    my-bucket /mount/point gcsfuse rw,noauto,user,key_file=/path/to/key.json
+
 [gce]: https://cloud.google.com/compute/
 [gce-service-accounts]: https://cloud.google.com/compute/docs/authentication
 [gcloud tool]: https://cloud.google.com/sdk/gcloud/
@@ -97,6 +101,7 @@ with dashes instead of underscores:
 *   `implicit_dirs`
 *   `dir_mode`
 *   `file_mode`
+*   `key_file`
 
 On both OS X and Linux, you can also add entries to your `/etc/fstab` file like
 the following:
--- a/docs/mounting.md
+++ b/docs/mounting.md
@@ -22,7 +22,7 @@ Console:
 
     GOOGLE_APPLICATION_CREDENTIALS=/path/to/key.json gcsfuse [...]
 
-On OS X and Linux you can also use fstab option in `/etc/fstab` system file:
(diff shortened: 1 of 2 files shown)


vaygr:
I signed it!

tools/mount_gcsfuse/main.go
```diff
@@ -87,6 +87,14 @@ func makeGcsfuseArgs(
 			)
 
 		// Special case: support mount-like formatting for gcsfuse string flags.
+		case "key_file":
+			args = append(
+				args,
+				"--"+strings.Replace(name, "_", "-", -1),
```
jacobsa: Please run gofmt.

docs/mounting.md
```diff
@@ -22,6 +22,10 @@ Console:
 
     GOOGLE_APPLICATION_CREDENTIALS=/path/to/key.json gcsfuse [...]
 
+On OS X and Linux you can also use fstab option in `/etc/fstab` system file:
```
jacobsa: Could you touch up the wording here a bit (and drop the references to OS X and Linux, since these are the only systems we support)? Probably:

When mounting with an fstab entry, you can use the `key_file` option. For example:

jacobsa:
Thanks for this! I've left a few comments.

tools/mount_gcsfuse/main.go
vaygr: I did. Double-checked that now.

docs/mounting.md
vaygr: Well, good point. My comment was based on the existing last note of this file though:

```
On both OS X and Linux, you can also add entries to your /etc/fstab file like the following:
```

tools/mount_gcsfuse/main.go
jacobsa: My apologies, I had misjudged what gofmt would do here. Thanks!

docs/mounting.md
jacobsa: Ah you're right. In that case, could you actually move this information to the fstab section of this file? So it says something like

> If you need to specify custom credentials in your fstab, use the `key_file=/path/to/file` option.

jacobsa:
Actually this looks good as-is; will accept and touch up if I feel it needs it later. Thanks!

docs/mounting.md
vaygr: Sure thing.

PR #194, opened 2016-10-03
Patch for gcsfuse systemd integration

jbaublitz:
This pull request is meant to enable gcsfuse mounting from systemd. After digging a little bit, it seems systemd supplies the ```-n``` (```--no-mtab```) option to whatever mount executable helper is invoked based on the type of filesystem specified as the mount in the ```.mount``` file. See [this link](https://www.freedesktop.org/software/systemd/man/systemd.mount.html#) for more information on ```.mount``` files but the basic idea is to expose filesystem mounts as a service that is callable from systemd and allows service dependency resolution based on what filesystems have been mounted. While including the ```-n``` flag might make sense in the systemd context given the developments around turning ```/etc/mtab``` into a symlink to ```/proc/self/mounts``` which is read only on my system, gcsfuse sees this as a bad flag so it cannot be integrated into a systemd mount file. There does not seem to be a flag to disable the ```-n``` flag when systemd invokes the mount helper so the simplest solution seems to be to ignore the ```-n``` flag from the gcsfuse mount helper. I'm open to discussion around this but given that ```-n``` is a standard flag in the ```mount``` command, I think it might be beneficial not just for this use case but for others as well to accept and ignore it for compatibility reasons similar to this one. I'm going to add an integration test so this is not yet ready to be merged but have submitted this patch so that we can at least start a discussion in case anyone working on this project is opposed or has other suggestions.

### DIFF
--- a/tools/integration_tests/mount_helper_test.go
+++ b/tools/integration_tests/mount_helper_test.go
@@ -134,6 +134,17 @@ func (t *MountHelperTest) BadUsage() {
 	}
 }
 
+func (t *MountHelperTest) IgnoreMtabFlag() {
+	var err error
+
+	// Mount.
+	args := []string{canned.FakeBucketName, t.dir, "-n"}
+
+	err = t.mount(args)
+	AssertEq(nil, err)
+	unmount(t.dir)
+}
+
 func (t *MountHelperTest) SuccessfulMount() {
 	var err error
 	var fi os.FileInfo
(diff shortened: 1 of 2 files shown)


jbaublitz:
I signed it!

jacobsa:
I'm happy with this in principle, thanks. Please add a bit more context to the comment in the code and an integration test.

jbaublitz:
@jacobsa I have provided an integration test and verified that it fails without my code to ignore ```-n```.

jacobsa:
Merged in 9009346f86b8e733eee945fdd5f3131f73b7c7fb. Thanks!

PR #231, opened 2017-06-06
Added another CentOS setup instruction

Nerdoholic:
The step "yum update" was missing.
Without it the system won't be able to find the package gcsfuse, because it has yet to initialize the repo.


Nerdoholic:
I signed it!

jacobsa:
Thanks!

PR #253, opened 2017-11-28
Rename googlecloudplatform to GoogleCloudPlatform

zchee:
import path respect case-sensitive word. can't import 'internal' packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,20 +34,20 @@ install: true
 #
 # Normally this is papered over by the fact that the default `install` step for
 # go does `go get -v ./...` from within  that directory. That winds up fetching
-# another copy into `$GOPATH/src/github.com/googlecloudplatform/gcsfuse`, so
+# another copy into `$GOPATH/src/github.com/GoogleCloudPlatform/gcsfuse`, so
 # the imports resolve correctly. (But this is bad! We get the wrong versions if
 # commits are pushed quickly.)
 #
 # However we disabled the install step above, so this doesn't happen and
 # instead we get errors like the following:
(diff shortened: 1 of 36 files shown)


jacobsa:
Sorry, can you please give more detail about what you're trying to do here and how it's not working? I haven't had trouble building gcsfuse on Linux or macOS before because of this.

nmousavi:
I guess this is related to this err msg

```
go get github.com/GoogleCloudPlatform/gcsfuse
package github.com/GoogleCloudPlatform/gcsfuse
        imports github.com/googlecloudplatform/gcsfuse/internal/canned: use of internal package not allowed
package github.com/GoogleCloudPlatform/gcsfuse
        imports github.com/googlecloudplatform/gcsfuse/internal/fs: use of internal package not allowed
package github.com/GoogleCloudPlatform/gcsfuse
        imports github.com/googlecloudplatform/gcsfuse/internal/gcsx: use of internal package not allowed
package github.com/GoogleCloudPlatform/gcsfuse
        imports github.com/googlecloudplatform/gcsfuse/internal/mount: use of internal package not allowed
package github.com/GoogleCloudPlatform/gcsfuse
        imports github.com/googlecloudplatform/gcsfuse/internal/perms: use of internal package not allowed
```

If you run `go get ...`. My workaround for this was to clone origin (doesn't work for my fork), and build it there (`go install .`)

jacobsa:
The problem is that you should run `go get github.com/GoogleCloudPlatform/gcsfuse`. (Or use the gcsfuse release binaries.)

Issue #262, opened 2018-02-01
gcsfuse becomes very slow when accessed by multiple threads/processes


nmousavi:
Experimenting with using gcsfuse for DeepVariant (https://github.com/google/deepvariant), I've noticed a big performance degradation when more than one instance is reading the data via gcsfuse (read only). 

Observations:

- Input data is ~70GB.
- A single instance processes 1/2048 of the data in 15min when input data is downloaded from gcs.
- A single instance processes 1/2048 of the data in 12min when input data is fused using gcsfuse.
- 8 instances (running in parallel in a GCE VM) process 8/2048 of the data in 16min when input data is downloaded from gcs.
- 8 instances (running in parallel in a GCE VM) process 8/2048 of the data in ~30hours when input data is fused using gcsfuse.

Access pattern is interleaved, i.e., 

process 1 reads bytes 1-1000, 2048*1000 - 2048*1000 + 1000, so on
process 2 reads bytes 2000-3000, 2048*1000 + 2000 - 2048* 1000 + 3000, son 
.....

Any idea why the performance is so poor in multiprocess mode?  Please let me know if debug info needed. Thanks!

jacobsa:
To confirm, are you running all of these processes on a single VM? Have you considered running them on separate VMs (this would fully isolate them), or with separate gcsfuse mount points? I expect you'd see the performance problem entirely go away with either of these.

I imagine the bottleneck here is either a lock in the kernel (you could probably see this with `gcsfuse --foreground --debug_fuse`) or maybe a lock in gcsfuse (we might be able to figure this out by looking at `gcsfuse --foreground --debug_fuse --debug_gcs`).

nmousavi:
Yes, we used a single VM. Running them on separate VMs is not an scalable option for us, but running separate gcsfuse is an idea we will try (and update you). Is there any limit on the number of separate mount points we can/should create?

I am still interested in knowing why this happens in multi-process scenario, so collected the logs here:

https://paste.googleplex.com/6560735928778752
https://paste.googleplex.com/6478377296855040

Related question: why is `limit-ops-per-sec` set to 5? I have changed it to no limit (-1), but haven't seen any improvement .

jacobsa:
There's no limit on the number of mount points. `--limit-ops-per-sec` is set to a reasonable value by default just to avoid surprise huge bills from GCS.

[...]

PR #263, opened 2018-02-05
Include timestamp in the debug logs.

nmousavi:
Include timestamp in the debug logs
The need for this was brought up in issue #262

Tested:
  no


jacobsa:
Thanks!

PR #300, opened 2018-09-28
uid and gid not work without allow_other

bastiendonjon:
tested on Ubuntu 16


bjornleffler:
I fixed this last week. Thanks for the suggestion.

PR #310, opened 2018-11-19
Allow passing debugging options through `mount_gcsfuse`

riccardomurri:
Allow passing debugging options like `--debug_gcs` through `mount_gcsfuse` as mount options (e.g., `mount -t gcsfuse -o debug_gcs ...`).

This may be useful when debugging `gcsfuse` mounts from non-interactive sessions like systemd, autofs, or simply `/etc/fstab`.

### DIFF
--- a/tools/mount_gcsfuse/main.go
+++ b/tools/mount_gcsfuse/main.go
@@ -94,6 +94,13 @@ func makeGcsfuseArgs(
 				value,
 			)
 
+			// Special case: support mount-like formatting for gcsfuse debug flags.
+		case "debug_fuse", "debug_gcs", "debug_http", "debug_invariants":
+			args = append(
+				args,
+				"--"+strings,
+			)
+
 		// Pass through everything else.
 		default:
 			var formatted string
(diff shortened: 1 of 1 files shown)


bjornleffler:
Does this work? Line 101 looks wrong.

riccardomurri:
> Line 101 looks wrong.

What does look wrong in it?  It's basically the same code as in lines 86 and 93,except there's no need to replace `_` with `-` because the FUSE command-line actually uses underscores in the option names.

bjornleffler:
Does this code compile? Have you tried that it works?

strings is a package name, not a variable.

riccardomurri:
> strings is a package name, not a variable.

Oh, I see.  No, I haven't tried it -- I'm no Go programmer so I don't even have a build environment for it.  I thought that if there were some blatant errors, they would immediately catch the eye... should that be `'--'+name` instead of `'--'+strings` ?

bjornleffler:
Yes, please change line 101 to
'--'+name

I tested that this works today.

riccardomurri:
Committed the fix.  Many thanks for your patience in reviewing and pointing out the error!

PR #316, opened 2019-02-20
Update semantics.md now that GCS bucket listing is strongly consistent

bramp:
For at-least the past two years, GCS bucket listing has been strongly consistent [1, 2]. Thus update the semantics in-light of this fact.

[1] https://cloud.google.com/blog/products/gcp/how-google-cloud-storage-offers-strongly-consistent-object-listing-thanks-to-spanner
[2] https://cloud.google.com/storage/docs/consistency

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -184,14 +184,6 @@ more thorough discussion):
 *   The feature requires an additional request to GCS for each name lookup,
     which may have costs in terms of request budget and latency.
 
-*   GCS object listings are only eventually consistent, so directories that
-    recently implicitly sprang into existence due to the creation of a child
-    object may not show up for several minutes (or, in rare extreme cases,
-    hours or days). Similarly, recently deleted objects may continue for a time
-    to implicitly define directories that eventually wink out of existence.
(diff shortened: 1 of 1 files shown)


bramp:
@jacobsa any chance you could review this? Thanks! Trying to remove a source of confusion our GCS customers have.

jacobsa:
@bjornleffler owns the project now; leaving it to him.

bjornleffler:
I'm fine with this change. Given that GCS is now strongly consistent for most operations, we should probably call that out, explain what it means for GCS Fuse and add a reference to the documentation.

bramp:
Thanks @jacobsa and @bjornleffler.

@bjornleffler I did review the rest of the document and the semantics seem correct.  I'm happy to send another PR with text explicitly calling out that it is strongly consistent. I didn't do that originally as I felt it is the default with file systems that listing would be strongly consistent. I guess in this case it would help advertise the fact it has changed.

bjornleffler:
Please do. S3 is not strongly consistent, and also offers a similar client to GCSFuse. It would be nice to have the documentation call out how GCSFuse is benefitting from the strong consistency of GCS.

PR #327, opened 2019-05-10
Consume mtimes set by gsutil cp / rsync

RJacksonm1:
Hi folks!

I'd recently migrated a ton of data from bare disks to Google Cloud Storage, using `gsutil rsync` for initial the bulk copy, and then mounting via gcsfuse to maintain a filesystem interface to those files.

Once running on gcsfuse, the mtimes recorded against the files was the time of the bulk copy of data, rather than the original files' mtimes.

Turns out that gsutil does [preserves the original `mtime`](https://github.com/GoogleCloudPlatform/gsutil/blob/0d9d0175b2b10430471c7b744646e56210f991d3/gslib/commands/rsync.py#L1448-L1450) when copying or syncing files, under the metadata key [`goog-reserved-file-mtime`](https://github.com/GoogleCloudPlatform/gsutil/blob/0d9d0175b2b10430471c7b744646e56210f991d3/gslib/utils/posix_util.py#L50).

Gcsfuse resolves mtimes by consuming a `gcsfuse_mtime` attribute in the object's metadata, if present, otherwise it falls back to the object's own updated time.

To address the problem for my users, I quickly threw together [a tool](https://github.com/RJacksonm1/gcs-consolidate-mtimes/blob/master/cmd/gcs_consolidate_mtimes/main.go) to go through each object in my bucket and copy the time from `goog-reserved-file-mtime` to `gcsfuse_mtime`.

But I am curious if there'd be any appetite for gcsfuse to handle `goog-reserved-file-mtime` itself?

If so I've had a crack at building that with this PR, although I'm only a novice with Go so feedback is definitely welcome!

It's worth noting that [`gcsfuse_mtime` will be always be set when a file is saved/updated via gcsfuse](https://github.com/GoogleCloudPlatform/gcsfuse/blob/eebad4c2f5a72142409dea471289093f6fc75fea/internal/gcsx/syncer.go#L94), so it will always be the canonical value – hence the ordering of interpretation within this PR; `gcsfuse_mtime` always takes priority over `goog-reserved-file-mtime`.

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -17,6 +17,7 @@ package inode
 import (
 	"fmt"
 	"io"
+	"strconv"
 	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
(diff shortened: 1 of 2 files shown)


RJacksonm1:
CLA signed 😉

bjornleffler:
I like the general idea of this.

bjornleffler:
Looks good. Thanks again for contributing this.

PR #333, opened 2019-06-10
Add another good option for the slow --implicit-dirs parameter

rayfoss:
This is the option I went with... it's worked really well.

It would be great if such a script was part of gcsfuse. Example: `gcsfuse fix-dirs my-bucket`

You might ask, what about the poor soul who has a bucket named `fix-dirs`? ... well, that's me, and [I give you full permission to ignore me.](https://storage.googleapis.com/fix-dirs/permission.txt)

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -202,7 +202,7 @@ You could automate this with a webhook to reduce the delay between creating file
 and them apearing in gcsfuse. [issue-7-bash][Example bash script]
 
 [issue-7-bash]: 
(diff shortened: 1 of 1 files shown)


bjornleffler:
Oups, I totally missed this one. Sorry about that.

Rather than updating the documentation, we should probably turn this into a script.

rayfoss:
agreed, it should probably be a script that's included and intended to be executed from cron with parameters

PR #381, opened 2020-03-13
Remove unnecessary "yum update" step from CentOS install

mfielding:
Unlike apt-get, yum doesn't require an update run before installing a package from a new repo.


bjornleffler:
This is fixed. Thanks for highlighting this issue.

PR #383, opened 2020-04-03
Add options for successful auto-mounting

jmgreger:
On a GCE centos VM, auto-mounting failed as it tries to mount before the network is ready. Adding systemd.requires=network-online.target solves this issue.


jmgreger:
@googlebot I signed it!

bjornleffler:
Thanks for finding this. Do you know if it also works for Debian?

jmgreger:
I'm not sure, I haven't had a chance to test on Debian.  I'll give it a shot and let you know what I find

PR #389, opened 2020-04-14
http_proxy support for mount_gcsfuse

### LINKED ISSUE #388: Input/output error behind proxy whithout --foreground
### Current Setup:

CentOS Linux release 8.1.1911 (Core)
gcsfuse version 0.28.1 (Go version go1.9.7)

### Error Message:

ls: reading directory '/srv/demo-fs-mount/': Input/output error

### How to reproduce

Everything is done under the root user, no sudo/su

**Mount the bucket without the --foreground option**
```
[root@v0303 ~]# gcsfuse --key-file /root/key.json demo-bucket /srv/demo-fs-mount
Using mount point: /srv/demo-fs-mount
Opening GCS connection...
Opening bucket...
Mounting file system...
File system has been successfully mounted.
```
**Try to access the folder**
```
[root@v0303 ~]# ls /srv/demo-fs-mount/
ls: reading directory '/srv/demo-fs-mount/': Input/output error
```

**Unmount**
```
[root@v0303 ~]# umount /srv/demo-fs-mount
```

**Mount the bucket with the --foreground option**
```
[root@v0303 ~]# gcsfuse --foreground --key-file /root/key.json demo-bucket /srv/demo-fs-mount &
[1] 3521
Using mount point: /srv/demo-fs-mount
Opening GCS connection...

WARNING: gcsfuse invoked as root. This will cause all files to be owned by
root. If this is not what you intended, invoke gcsfuse as the user that will
be interacting with the file system.

Opening bucket...
Mounting file system...
File system has been successfully mounted.
```
**Try to access the folder**
```
[root@v0303 ~]# ls /srv/demo-fs-mount/
total 1
2 drwxr-xr-x. 1 root root  0 Apr 12 13:20 base
3 drwxr-xr-x. 1 root root  0 Apr 12 13:20 master
4 drwxr-xr-x. 1 root root  0 Apr 12 13:20 ssh-keys
```

So currently everything seems to work as long as the gcsfuse is run with the --foreground option, what only makes sense for debugging I guess.
As soon as I try to use a fstab entry or do the mount without the --foreground option the folder is not accessible anymore.
Also I'm not sure how to debug this as all debug features are relying on --foreground set and the issue does not occur with that option set.

peter-fe:
Added http_proxy to the mount.gcsfuse helper as it was done previously for the cli part (v.0.29.0), Closes #388

### DIFF
--- a/tools/mount_gcsfuse/main.go
+++ b/tools/mount_gcsfuse/main.go
@@ -232,6 +232,16 @@ func run(args []string) (err error) {
 	// Run gcsfuse.
 	cmd := exec.Command(gcsfusePath, gcsfuseArgs...)
 	cmd.Env = append(cmd.Env, fmt.Sprintf("PATH=%s", path.Dir(fusermountPath)))
+
+        // Pass through the https_proxy/http_proxy environment variable,
+	// in case the host requires a proxy server to reach the GCS endpoint.
+	// http_proxy has precedence over http_proxy, in case both are set
+        if p, ok := os.LookupEnv("https_proxy"); ok {
+                cmd.Env = append(cmd.Env, fmt.Sprintf("https_proxy=%s", p))
+        } else if p, ok := os.LookupEnv("http_proxy"); ok {
+                cmd.Env = append(cmd.Env, fmt.Sprintf("http_proxy=%s", p))
+        }
+
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
--- a/main.go
+++ b/main.go
@@ -319,10 +319,21 @@ func runCLIApp(c *cli.Context) (err error) {
 		if p, ok := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS"); ok {
 			env = append(env, fmt.Sprintf("GOOGLE_APPLICATION_CREDENTIALS=%s", p))
 		}
-		// Pass through the http_proxy environment variable, in case the host
-		// requires a HTTP proxy server to reach the GCS endpoint.
-		if p, ok := os.LookupEnv("http_proxy"); ok {
+		// Pass through the https_proxy/http_proxy environment variable,
+		// in case the host requires a proxy server to reach the GCS endpoint.
+		// http_proxy has precedence over http_proxy, in case both are set
+		if p, ok := os.LookupEnv("https_proxy"); ok {
+			env = append(env, fmt.Sprintf("https_proxy=%s", p))
+			fmt.Fprintf(
+				os.Stdout,
+				"Added environment https_proxy: %s\n",
+				p)
+		} else if p, ok := os.LookupEnv("http_proxy"); ok {
 			env = append(env, fmt.Sprintf("http_proxy=%s", p))
+			fmt.Fprintf(
+				os.Stdout,
+				"Added environment http_proxy: %s\n",
+				p)
 		}
 
 		// Run.


peter-fe:
@googlebot I signed it!

peter-fe:
@bjornleffler, can you do a review?
Would be great, thx a lot!

bjornleffler:
Thanks for contributing this.

I would change:
- Stderr to Stdout. It's a notification, not an error.
- Remove empty line 240.

Questions:
- Have you tried this. Does it work as expected?
- GCS should use mostly https. How about we prefer https_proxy with a fallback to http_proxy?

Thanks,
Bjorn

peter-fe:
@bjornleffler 

Good point with the https preference, I added the fallback to gcsfuse as well as there was only http_proxy respected. (actually the part I've stolen the code snippet from).

Adjusted the output to stdout and moved the output to the main gcsfuse code (otherwhise you'll see it twice as gcsfuse and mount.gcsfuse are get called in series) and removed the output from mount.gcsfuse.

It's not 100% perfect as actually two binaries get called, but should be sufficient to indicate that the https_proxy/http_proxy was used as long as the proxy parts stays in sync in both places.

Sadly it seems that the build pipeline is broken since some time, some of the checks fail due internal errors (internal/gcsx).

tools/mount_gcsfuse/main.go
```diff
@@ -232,6 +232,16 @@ func run(args []string) (err error) {
 	// Run gcsfuse.
 	cmd := exec.Command(gcsfusePath, gcsfuseArgs...)
 	cmd.Env = append(cmd.Env, fmt.Sprintf("PATH=%s", path.Dir(fusermountPath)))
+
+        // Pass through the https_proxy/http_proxy environment variable,
```
kumowoon1025, line 236: What happens if https_proxy is specified in both the env var and e.g. an entry in fstab?

peter-fe: Do you mean something like this in fstab?

`my-bucket /my-mount-point gcsfuse rw,dir_mode=770,file_mode=660,https_proxy=http://my-proxy:8080`

kumowoon1025: That's what I meant, but I guess the same thing happens if it's defined in the shell environment and you also specify it as an option with the CLI tool, seems to take environment variable value.

peter-fe: Actually, at least on my linux, when appended to the fstab it gets as -o https_proxy=... to fusermount and fusermount fails with an unkown parameter exception.

peter-fe:
@bjornleffler,

Can you give it another shot and maybe get the pipeline running again?

Br

bjornleffler:
Looks good. Thanks for the contribution.

I'll build a new release soon.

peter-fe:
@bjornleffler

Great, thank you!

peter-fe:
@bjornleffler

Would the new release also include the .rpm  again? That would be great, as it was missing in that last release.

Thx for keeping this project alive!

Br

peter-fe:
@bjornleffler 

Hey Bjorn, have you found some time to make a new release including this changes?
I'm mainly interested in the rpm & dedicated tag for the change.

Br
Peter

bjornleffler:
Oups, I totally dropped the ball on this.

We're preparing some major changes to GCS Fuse, so I'm going to cut a new version tomorrow. Let me make sure that I get the signed rpms out quickly this time.

bjornleffler:
Please have a look and see if release 0.30.0 works for you.

I have pushed the new release to the Google rpm repository.

peter-fe:
Thank you very much Bjorn!
Install and testing went well, perfekt!

PR #397, opened 2020-06-05
Update installing.md

dcloud9:
`yum update` will "upgrade" all packages as opposed to just refreshing the package indexes. Use optional `-y` to suppress answer `yes`.


bjornleffler:
Thanks for highlighting this. It seems like the update step isn't really necessary with yum, so I removed it.

bjornleffler:
I updated the installation instructions today.

PR #403, opened 2020-06-20
Refactor the gcs bucket used in file system

lezh:
The refactoring is to make it easier to add the feature of mounting all accessible buckets.

Before this change, the file system is not ready to support multiple buckets, because assumptions of "just one bucket" is prevalent throughout the system. This refactor, however, removes the assumption by

(1) the file system no longer owns a copy of the bucket,
(2) the buckets managed by a new object, BucketManager, allowing dynamically setting up more buckets, 
(3) each inode will inherit its bucket from the parent inode, allowing different buckets for different indoes.

### DIFF
--- a/internal/fs/bucket_manager.go
+++ b/internal/fs/bucket_manager.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package fs
 
 import (
 	"fmt"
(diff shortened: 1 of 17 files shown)


internal/fs/bucket_manager.go
```diff
@@ -37,26 +38,54 @@ type BucketConfig struct {
...
+	//
+	// 1. Write out a temporary object containing the appended contents whose
+	//    name begins with TmpObjectPrefix.
+	//
+	// 2. Compose the original object and the temporary object on top of the
+	//    original object.
+	//
+	// 3. Delete the temporary object.
+	//
+	// Note that if the process fails or is interrupted the temporary object will
```
bjornleffler: Let's automate this.

sshrdp:
@bjornleffler gentle ping on this :)

bjornleffler:
I merged these changes into a new branch called "mountall" today.

PR #412, opened 2020-07-10
Cache implicit dirs to reduce ls latency

lezh:
Implicit dirs are the gcs directories that don't have an object to back up.

The dir inodes used to memoize the subdirs, but not distinguishing the implicit subdirs. When it looked up subdir children, it had to issue an gcs ListObject api call to tell if the child was implicit or not, which was very expensive. This PR allows the dir inodes to cache the dirs as implicit dirs, to avoid such expensive api calls.

As a result, the latency of `ls a/b/c/d/` of 1024 files could be reduced from 10 min to 1.5 seconds, due to 4096 api calls avoided.

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -645,6 +645,7 @@ func (d *dirInode) LookUpChild(
 	now := d.cacheClock.Now()
 	cacheSaysFile := d.cache.IsFile(now, name)
 	cacheSaysDir := d.cache.IsDir(now, name)
+	cacheSaysImplicitDir := d.cache.IsImplicitDir(now, name)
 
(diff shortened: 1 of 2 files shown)


bjornleffler:
I like it, but it needs to a command line option and configurable TTL. Leave it disabled by default.

lezh:
@bjornleffler 

> a command line option and configurable TTL

Yes, it's a part of the current "type cache". Thus, `--type-cache-ttl` sets its ttl.

> Leave it disabled by default.

It's disabled by default, since "implicit dir" is not disabled by default. The user needs to pass `--implicit-dirs' to enable the implicit directory, and this cache.

bjornleffler:
All good then.

PR #416, opened 2020-07-14
Remove "yum update" step from CentOS/RH install

mfielding:
Under debian, an "apt-get update" step is required when adding repositories.  This step is not required in CentOS/Red Hat, and actually has a different effect: updating all packages on the system, which may not be what the user intended.  Hre we remove the step entirely, siply adding the repo and then running yum install.


bjornleffler:
Thanks for highlighting this issue. I updated the documentation today.

PR #422, opened 2020-07-18
Make root directory containing buckets readable

lezh:
When the GCS bucket name is omitted, the root directory contains
subdirectories that represent buckets. Currently, reading the root
directory will end up in an error. This commit changes the behavior by
providing a list of bucket names for such operation.

The implementation uses the [official GCS client library](
https://godoc.org/cloud.google.com/go/storage). So, for it to be built, 
you may need to call `go get` to install the dependencies at GOPATH.

For this API to work, 
- It must know the project id, which can be passed  in as 
  `--billing-project=...`.
- The user or the service account must have the permission 
  `storage.buckets.list`.

TEST

The test is done by listing the all buckets on a GCE instance.

### DIFF
--- a/internal/fs/fs_test.go
+++ b/internal/fs/fs_test.go
@@ -288,6 +288,14 @@ type fakeBucketManager struct {
 
 func (bm *fakeBucketManager) ShutDown() {}
 
+func (bm *fakeBucketManager) ListBuckets(
+	ctx context.Context) (names []string, err error) {
+	for name, _ := range bm.buckets {
+		names = append(names, name)
+	}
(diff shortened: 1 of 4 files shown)


bjornleffler:
Why billing project, as opposed to just project? Is that because of the official golang SDK?

lezh:
@bjornleffler "billing project" is a term already used by gcsfuse. If we change the flag name, we potentially break existing use cases.

bjornleffler:
Fair enough. I'll look into this before we build a public release of this code.

PR #423, opened 2020-07-20
Tune TCP connections for HTTP/1.1

lezh:
The default protocol used by Go's http transport is HTTP/2. However,
I found HTTP/1.1 with multiple TCP connections can perform better
compared to HTTP/2 with TCP multiplexing.

This commit adds new flags `--disable-http2` and `--max-conns-per-host`.
It disables HTTP/2 and sets MaxConnsPerHost option for HTTP/1.1, which
can potentially improve read throughput by 2x.

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,9 +21,11 @@
 package main
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
 	"os/signal"
 	"path"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -209,6 +211,20 @@ func getConn(flags *flagStorage) (c gcs.Conn, err error) {
 		MaxBackoffSleep: flags.MaxRetrySleep,
 	}
 
+	// The default HTTP transport uses HTTP/2 with TCP multiplexing, which
+	// does not create new TCP connections even when the idle connections
+	// run out. To specify multiple connections per host, HTTP/2 is disabled
+	// on purpose.
+	if flags.MaxConnsPerHost > 0 {
+		cfg.Transport = &http.Transport{
+			MaxConnsPerHost: 100,
```
bjornleffler, line 220: Copy the value?

bjornleffler:
Overall this looks good, but we need some recommendations on sizing. A maximum of 10 is probably a good place to start?

lezh:
@bjornleffler Sounds good. I'll use two flags to specify http version and connection pool size respectively.

lezh:
Updated to incorporate the comments.

PR #431, opened 2020-08-09
Add prometheus metrics for monitoring

lezh:
This PR adds 4 prometheus metrics:
- number of GCS requests by type,
- number of bytes read,
- number of readers created,
- number of readers closed.

They can be helpful for debugging performance issues as well as tracking the progress of some client application relying on gcsfuse.

TEST

Acquire the stats from localhost:8086 while running gcsfuse:
```
# HELP gcsfuse_bytes_read Number of bytes read from GCS.
# TYPE gcsfuse_bytes_read counter
gcsfuse_bytes_read{bucket="tf-ent-dev",object="datasets/imagenet/combined/train-00001-of-01024"} 1.33300224e+08
# HELP gcsfuse_gcs_requests Number of GCS requests.
# TYPE gcsfuse_gcs_requests counter
gcsfuse_gcs_requests{bucket="tf-ent-dev",method="ListObjects"} 10
gcsfuse_gcs_requests{bucket="tf-ent-dev",method="NewReader"} 1
gcsfuse_gcs_requests{bucket="tf-ent-dev",method="StatObject"} 15
# HELP gcsfuse_object_readers_closed Number of object readers already closed.
# TYPE gcsfuse_object_readers_closed counter
gcsfuse_object_readers_closed 1
# HELP gcsfuse_object_readers_created Number of object readers created.
# TYPE gcsfuse_object_readers_created counter
gcsfuse_object_readers_created 1
```

### DIFF
--- a/internal/gcsx/bucket_manager.go
+++ b/internal/gcsx/bucket_manager.go
@@ -212,6 +212,9 @@ func (bm *bucketManager) SetUpBucket(
 	// Enable content type awareness
 	b = NewContentTypeBucket(b)
 
+	// Enable monitoring
+	b = NewMonitoringBucket(b)
+
 	// Enable Syncer
 	if bm.config.TmpObjectPrefix == "" {
 		err = errors.New("You must set TmpObjectPrefix.")
--- a/flags.go
+++ b/flags.go
@@ -200,6 +200,17 @@ func newApp() (app *cli.App) {
 			},
 
 			/////////////////////////
+			// Monitoring
+			/////////////////////////
+
+			cli.IntFlag{
+				Name:  "monitoring-port",
+				Value: 0,
+				Usage: "The port used to export prometheus metrics for monitoring. " +
+					"The default value 0 indicates no monitoring metrics.",
+			},
+
+			/////////////////////////
 			// Debugging
 			/////////////////////////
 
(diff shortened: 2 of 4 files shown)


internal/gcsx/bucket_manager.go
```diff
@@ -212,6 +212,9 @@ func (bm *bucketManager) SetUpBucket(
 	// Enable content type awareness
 	b = NewContentTypeBucket(b)
 
+	// Enable monitoring
+	b = NewMonitoringBucket(b)
```
bjornleffler, line 216: Only if monitoring is enabled?

lezh: The current switch only controls whether the metrics are exported. The metrics will always be counted internally. The cost of keeping them is negligible. 

If we decide to use one switch to turn off all the metrics, although this might be easy to implement now since all the metrics exist in one place (monitoring_bucket.go), it can be quite complicated in the future when we have metrics scattered around different files. That means the value of the switch has to be passed around, and "if-else" clauses executed everywhere to decide whether the metrics should be counted, which can be quite ugly.

What do you think? Is it necessary to have all the internal metrics as optional?

bjornleffler: Good points. Let me think about it.

PR #432, opened 2020-08-19
Monitor the fuse requests

lezh:
Internally, gcsfuse has a "FileSystem" struct to handle fuse requests. This PR modifies each method of the FileSystem with these monitoring metrics:

- the number of each requests,
- the number of errors,
- the latency of ReadFile and OpenFile requests.

These metrics will be very helpful for debugging performance issues.

### DIFF
--- a/internal/fs/monitoring_fs.go
+++ b/internal/fs/monitoring_fs.go
@@ -16,6 +16,7 @@ package fs
 
 import (
 	"context"
+	"time"
 
 	"github.com/jacobsa/fuse/fuseops"
 	"github.com/jacobsa/fuse/fuseutil"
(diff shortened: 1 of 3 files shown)


internal/fs/monitoring_fs.go
```diff
@@ -0,0 +1,342 @@
...
+			Name:    "gcsfuse_fs_open_file_latency",
+			Help:    "The latency of executing an OpenFile request in ms.",
+			Buckets: prometheus.ExponentialBuckets(0.01, 10, 8),
+		},
+	)
+	latencyReadFile = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "gcsfuse_fs_read_file_latency",
+			Help:    "The latency of executing a ReadFile request in ms.",
+			Buckets: prometheus.ExponentialBuckets(0.01, 10, 8),
```
bjornleffler, line 56: Please add a comment to explain what buckets are generated. I usually use:
// [0.1, 1.5, 30] generates 30 buckets between 0.1 ms and 12 seconds.

internal/gcsx/monitoring_bucket.go
```diff
@@ -55,6 +56,20 @@ var (
 			Help: "Number of object readers already closed.",
 		},
 	)
+	latencyNewReader = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "gcsfuse_object_new_reader_latency",
+			Help:    "The latency of creating a GCS object reader in ms.",
+			Buckets: prometheus.ExponentialBuckets(0.01, 10, 8),
```
bjornleffler, line 63: Same here.

lezh:
@bjornleffler 
Updated ExponentialBuckets with the suggestion

PR #433, opened 2020-08-20
Measure the latency of LookUpInode

lezh:
LookUpInode was the most common operation. It could easily become the bottleneck. So, monitoring its latency is necessary.

### DIFF
--- a/internal/fs/monitoring_fs.go
+++ b/internal/fs/monitoring_fs.go
@@ -42,6 +42,15 @@ var (
 			"method",
 		},
 	)
+	latencyLookUpInode = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name: "gcsfuse_fs_look_up_inode_latency",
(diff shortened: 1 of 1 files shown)


internal/fs/monitoring_fs.go
```diff
@@ -42,6 +42,15 @@ var (
 			"method",
 		},
 	)
+	latencyLookUpInode = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name: "gcsfuse_fs_look_up_inode_latency",
+			Help: "The latency of executing an LookUpInode request in ms.",
```
bjornleffler, line 48: Let's make it a bit more generic. This way, we can reuse the downstream aggregation templates.

Something like:
latency = prometheus.NewHistogram(
		prometheus.HistogramOpts{
			Name: "gcsfuse_latency",

Add a label for the type. Maybe operation="lookup_inode"?

lezh:
@bjornleffler Updated with the feedback

PR #486, opened 2021-01-25
Update installing.md

leozitor:
using `brew cask install osxfuse` throws this erro:
Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
So changed to `brew install osxfuse` works


lezh:
Similar PR #488

leozitor:
> Similar PR #488

Sorry, my bad for not checking before.

PR #493, opened 2021-02-13
fix brew install command

antonpolishko:


antonpolishko:
@googlebot I signed it!

PR #507, opened 2021-03-24
Set symlink atime and ctime to mtime

gaul:
This mimics the file logic and avoids nonsense values.  Found via
pjdfstests.


gaul:
You can verify this via:

```
prove -rv /path/to/pjdfstest/tests/utimensat/05.t
```

PR #512, opened 2021-04-06
Disable repo_gpgcheck in yum install instructions

ericdand:
repo_gpgcheck is buggy and nobody else uses it. We're following Red Hat et al. and disusing it too.


lezh:
Duplicate of #510

PR #538, opened 2021-06-11
Allow configuring non-Google endpoint

gaul:
This allows use of other implementations like fake-gcs-server.  Setting a
non-Google endpoint disables OAuth.

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"strconv"
 	"time"
@@ -120,6 +121,12 @@ func newApp() (app *cli.App) {
 			/////////////////////////
 			// GCS
 			/////////////////////////
+
+			cli.StringFlag{
+				Name:  "endpoint",
+				Value: "https://www.googleapis.com:443",
+				Usage: "The endpoint to connect to.",
+			},
 
 			cli.StringFlag{
 				Name:  "billing-project",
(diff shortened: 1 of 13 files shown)


gaul:
Go 1.14 build fails due to timeout but the others succeed?

flags.go
```diff
@@ -285,6 +292,7 @@ type flagStorage struct {
 	RenameDirLimit int64
 
 	// GCS
+	Url                                *url.URL
```
lezh, line 295: The fields in this struct match the names of the flags. Should we change `Url` to `Endpoint` instead, to match the new flag `--endpoint`?

internal/gcsx/connection.go
```diff
@@ -36,9 +36,17 @@ func NewConnection(cfg *gcs.ConnConfig) (c *Connection, err error) {
 		return
 	}
 
+	var auth option.ClientOption
+	if cfg.TokenSource == nil {
```
lezh, line 40: When the endpoint is overridden to a non-google url, the TokenSource would be 
```
oauth2.StaticTokenSource(&oauth2.Token{})
```
according to `getConn`. Why do we need to handle `cfg.TokenSource == nil` which never happens?

lezh (changes requested):

flags.go
gaul: Done.

internal/gcsx/connection.go
gaul: Removed.  I tested this against GCS and fake-gcs-server to make sure that both still work.

gaul:
@lezh Any other comments?

PR #588, opened 2022-01-18
fix:Using double quote to prevent globbing and word splitting.

bluPhy:
Using double quote to prevent globbing and word splitting.


lezh:
This PR changes the files in the vendored packages. The changes would be overwritten when the vendored packages are updated in the future. It's better to contribute to those packages and then update Gcsfuse.

PR #592, opened 2022-01-27
Fix integration tests and create fuzzy read/write test

thepenguinco:
Fix failing integration tests and add read/write test for files containing unicode and rune characters.


tools/integration_tests/mounting/gcsfuse_test.go
```diff
@@ -157,7 +157,10 @@ func (t *GcsfuseTest) NonEmptyMountPoint() {
 	args := []string{canned.FakeBucketName, t.dir}
 
 	err = t.runGcsfuse(args)
-	ExpectThat(err, Error(HasSubstr("is not empty")))
+	ExpectThat(err, Error(HasSubstr("exit status 1")))
+	// TODO ezl the below string is not returned by cmd.CombinedOutput, we should make changes
```
thepenguinco, line 161: I suspect this might be from FUSE library.

PR #595, opened 2022-02-03
[DO NOT MERGE]: gcsfuse read only content cache

thepenguinco:
Implemented content cache under --local-file-cache flag to persist files on disk in tmp directory (otherwise in specified folder using --temp-dir <yourdirectoryhere>)

Currently the tests spill files onto disk without cleaning up which is not intentional.

File names consist of an autogenerated string of numbers by ioutil.TempFile (wrapper for os.CreateTemp).

We will implement a function to prefetch the cache on gcsfuse startup as well as implement a metadata store (JSON file[s]).


lezh:
There are multiple changes in the PR. Can you split it? Some of them could be merged directly.

Issue #590, opened 2022-01-24
Improve CI test coverage


gaul:
gcsfuse could reuse s3fs tests, pjdfstest, or xfstests in CI by using [fake-gcs-server](https://github.com/fsouza/fake-gcs-server).  https://github.com/gaul/s3fs-fuse/tree/gcsfuse shows some partial work.

avidullu:
https://github.com/GoogleCloudPlatform/gcsfuse/pull/777 added the fake-gcs-server to our repo and we are already building tests around the same.

PR #597, opened 2022-02-06
Do not allow renaming into a non-empty directory

gaul:
This allows pjdfstest rename/20.t to pass.  References
GoogleCloudPlatform/gcsfuse#590.

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1501,14 +1501,20 @@ func (fs *fileSystem) renameDir(
 	}
 	pendingInodes = append(pendingInodes, oldDir)
 
-	// Fetch all the descendants of the directory recuirsively
+	// Fetch all the descendants of the old directory recursively
 	descendants, err := oldDir.ReadDescendants(ctx, int(fs.renameDirLimit+1))
 	if len(descendants) > int(fs.renameDirLimit) {
 		return fmt.Errorf("too many objects to be renamed: %w", syscall.EMFILE)
 	}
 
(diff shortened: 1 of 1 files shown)


internal/fs/fs.go
```diff
@@ -1501,14 +1501,20 @@ func (fs *fileSystem) renameDir(
...
+	// Fetch all the descendants of the old directory recursively
 	descendants, err := oldDir.ReadDescendants(ctx, int(fs.renameDirLimit+1))
 	if len(descendants) > int(fs.renameDirLimit) {
 		return fmt.Errorf("too many objects to be renamed: %w", syscall.EMFILE)
 	}
 
-	// Create the backing object of the new directory.
 	newParent.Lock()
+	// Fetch a single descendant of the new parent
+	if newParentDescendents, err := newParent.ReadDescendants(ctx, 1); err == nil && len(newParentDescendents) > 0 {
```
lezh, line 1512: Here, the `newParent` refers to the parent directory of the target. For instance, in

```
mv /gcs/foo/bar/ /gcs/foo/baz/
```

both the oldParent and newParent are the directory `/gcs/foo/`. Therefore, we cannot abort the renaming if the newParent isn't empty.

gaul: Sorry about this!  CI also should have caught this -- I opened #606 to prevent further regressions.

lezh: @gaul No worries. Thanks for your contribution!

PR #599, opened 2022-02-09
Add cache and metadata support for on disk cache

thepenguinco:
We add metadata and file support in memory for the cache.  We add the in memory map in contentcache.go and create supporting structs and modifications to support metadata in gcsx.TempFile.

This change should not impact or break any unit tests: run go test ./... in the gcsfuse directory and all tests should continue to pass.

### DIFF
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -22,18 +22,32 @@ import (
 	"github.com/jacobsa/timeutil"
 )
 
+// CacheObjectKey contains
+type CacheObjectKey struct {
+	BucketName string
+	ObjectName string
+}
+
 // ContentCache is a directory on local disk to store the object content.
 type ContentCache struct {
-	tempDir    string
-	mtimeClock timeutil.Clock
+	tempDir        string
+	localFileCache bool
+	fileMap        map[CacheObjectKey]gcsx.TempFile
+	mtimeClock     timeutil.Clock
 }
 
 // New creates a ContentCache.
 func New(tempDir string, mtimeClock timeutil.Clock) *ContentCache {
 	return &ContentCache{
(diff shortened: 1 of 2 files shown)


internal/gcsx/temp_file.go
```diff
@@ -71,6 +74,14 @@ type StatResult struct {
 	Mtime *time.Time
 }
 
+// Metadata store struct
+type Metadata struct {
```
lezh, line 78: Metadata is a broad and ambiguous term, especially in the package "gcsx". It can be the metadata of the object, the temp file, or something else. In this case, we are creating this struct for ObjectMetadata used by TempFile. Therefore, we can fix this in any of the three ways:

(1) Use "TempFileObjectMetadata"
(2) Move the temp_file.go into a new package "tempfile", and keep the name "Metadata" - "tempfile.ObjectMetadata" also clear
(3) Keep it private by changing the name to "objectMetadata".

internal/contentcache/contentcache.go
```diff
@@ -22,20 +22,34 @@ import (
...
 	return &ContentCache{
 		tempDir:    tempDir,
+		fileMap:    make(map[CacheObjectKey]gcsx.TempFile),
 		mtimeClock: mtimeClock,
 	}
 }
 
+// Populates the ContentCache with files on disk
+func (c *ContentCache) PopulateCache() {
+
```
lezh, line 50: Remove this empty function or leave a TODO comment in the function body.

lezh (changes requested):
The changes in contentcache.go and temp_file.go are independent of each other. Thus, they should be ideally put in separate PRs.

thepenguinco:
I will split up changes more in subsequent commits.

PR #603, opened 2022-02-10
Allow creation of temp cache files and metadata files

thepenguinco:
Add way to persist temp cache files and metadata files on disk.

One concern I have is whether we want to store a pointer to the file on disk or if storing the struct in memory is enough.  The metadata on disk is only used to allow gcsfuse to hotload the cache on startup and isn't used any other time.

Testing:

Run go test ./....  and unit tests should pass. There should be no impacting changes on unit tests.

### DIFF
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -51,8 +51,7 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
 
-// NewTempFile returns a handle for a temporary file on the disk. The caller
-// must call Destroy on the TempFile before releasing it.
+// NewCacheFile returns a handle for a cache file on the disk.
 func (c *ContentCache) NewCacheFile(rc io.ReadCloser, metadata *gcsx.TempFileObjectMetadata) (gcsx.TempFile, error) {
 	return gcsx.NewCacheFile(rc, metadata, c.tempDir, c.mtimeClock)
 }
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -50,3 +50,9 @@ func New(tempDir string, mtimeClock timeutil.Clock) *ContentCache {
 func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
+
+// NewTempFile returns a handle for a temporary file on the disk. The caller
+// must call Destroy on the TempFile before releasing it.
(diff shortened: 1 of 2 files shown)


internal/gcsx/temp_file.go
```diff
@@ -108,6 +113,38 @@ func NewTempFile(
 	return
 }
 
+// NewCacheFile creates a temp file whose initial contents are given by the
+// supplied reader. dir is a directory on whose file system the file will live,
+// or the system default temporary location if empty.
+func NewCacheFile(
```
lezh, line 119: The package "gcsx" means "extended gcs". It contains the structs for communicating with the GCS API.

This NewCacheFile function is specifically used by the cache. How a cache file is written to the disk doesn't matter to the GCS communication. So, it should live in the contentcache package instead.

internal/contentcache/contentcache.go
```diff
@@ -50,3 +50,8 @@ func New(tempDir string, mtimeClock timeutil.Clock) *ContentCache {
 func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
+
+// NewCacheFile returns a handle for a cache file on the disk.
+func (c *ContentCache) NewCacheFile(rc io.ReadCloser, metadata *gcsx.TempFileObjectMetadata) (gcsx.TempFile, error) {
```
lezh, line 55: The name is a little confusing. Initially, I thought this was the file storing the TempFileObjectMetadata. 

Can we clarify the file content in the comment above, like "a cache file on the disk storing the object content"?

lezh (changes requested):

internal/gcsx/temp_file.go
thepenguinco: @lezh  I thought we wanted GCS API to interact with cached files that instead of the TempFile.  This allows users to modify files and "writethrough" the changes to the cache before pushing the changes to GCS.

lezh: We should clarify the responsibilities of each package. Depending on your design, the scope of the `fs` and `contentcache` packages could be either of these cases:

1. A look-aside cache. The `fs` package should be able to use the `gcsx` package to download the content from GCS API, and save it via the `contentcache` package.
2. A write-through cache. The `fs` package only talks to the `contentcache` package, which then uses the `gcsx` package to communicate with the APIs.

In either case, the responsibility of `gcsx` should be restricted to communications with the GCS APIs.

thepenguinco: I agree.  I will add a task to refactor this in a later PR but I think this change can go in for now.

internal/contentcache/contentcache.go
lezh: Please address the it by modifying the code comment

PR #604, opened 2022-02-10
gcsfuse cache method implementation

thepenguinco:
Implement addReplace, remove and get methods to interact with the cache.

Run go test ./...

There should be no failures.

TODO: implement some basic unit tests later to verify correctness of the cache datastructure.

### DIFF
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -51,6 +51,35 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
 
+// Function to add or update existing cache file
+func (c *ContentCache) AddOrReplace(cacheObjectKey *CacheObjectKey, generation int64, rc io.ReadCloser) (gcsx.TempFile, error) {
+	if _, exists := c.fileMap[*cacheObjectKey]; exists {
+		c.fileMap[*cacheObjectKey].Destroy()
+	}
+	metadata := &gcsx.TempFileObjectMetadata{
+		BucketName: cacheObjectKey.BucketName,
+		ObjectName: cacheObjectKey.ObjectName,
+		Generation: generation,
+	}
+	file, err := c.NewCacheFile(rc, metadata)
+	c.fileMap[*cacheObjectKey] = file
+	return file, err
+}
+
+// Retrieve temp file from the cache
+func (c *ContentCache) Get(cacheObjectKey *CacheObjectKey) (gcsx.TempFile, bool) {
+	file, exists := c.fileMap[*cacheObjectKey]
+	return file, exists
+}
+
+// Function to remove and destroy cache file and metadata on disk
+func (c *ContentCache) Remove(cacheObjectKey *CacheObjectKey) {
+	if _, exists := c.fileMap[*cacheObjectKey]; exists {
+		c.fileMap[*cacheObjectKey].Destroy()
+	}
+	delete(c.fileMap, *cacheObjectKey)
+}
+
 // NewCacheFile creates a cache file on the disk storing the object content
 // TODO ezl we should refactor reading/writing cache files and metadata to a different package
 func (c *ContentCache) NewCacheFile(rc io.ReadCloser, metadata *gcsx.TempFileObjectMetadata) (gcsx.TempFile, error) {
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -17,6 +17,7 @@
 package contentcache
 
 import (
+	"fmt"
 	"io"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
@@ -52,7 +53,7 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
 
-// Add or replace creates a new cache file or updates an existing cache file
+// AddOrReplace creates a new cache file or updates an existing cache file
 func (c *ContentCache) AddOrReplace(cacheObjectKey *CacheObjectKey, generation int64, rc io.ReadCloser) (gcsx.TempFile, error) {
 	if cacheObject, exists := c.fileMap[*cacheObjectKey]; exists {
 		cacheObject.Destroy()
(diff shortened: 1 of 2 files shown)


internal/contentcache/contentcache.go
```diff
@@ -50,3 +50,38 @@ func New(tempDir string, mtimeClock timeutil.Clock) *ContentCache {
...
+		BucketName: cacheObjectKey.BucketName,
+		ObjectName: cacheObjectKey.ObjectName,
+		Generation: generation,
+	}
+	file, err := c.NewCacheFile(rc, metadata)
+	c.fileMap[*cacheObjectKey] = file
+	return file, err
+}
+
+// Retrieve temp file from the cache
```
lezh, line 69: Let's follow [the golang guidance](https://go.dev/doc/effective_go#commentary) that recommends the function comment to start with the function name. 

>  If every [doc](https://go.dev/cmd/go/#hdr-Show_documentation_for_package_or_symbol) comment begins with the name of the item it describes, you can use the doc subcommand of the [go](https://go.dev/cmd/go/) tool and run the output through grep.

So, we can change it to
```go
// Get retrieves a temp file from the cache given the GCS object name and its generation.
```

internal/contentcache/contentcache.go
```diff
@@ -50,3 +50,38 @@ func New(tempDir string, mtimeClock timeutil.Clock) *ContentCache {
 func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
+
+// Function to add or update existing cache file
```
lezh, line 54: The cache is going to be used concurrently. Any idea how to ensure thread safety? If the caller is required to ensure it's used in a thread safe way, we should note that in the comment.

internal/contentcache/contentcache.go
```diff
@@ -50,3 +50,38 @@ func New(tempDir string, mtimeClock timeutil.Clock) *ContentCache {
...
+func (c *ContentCache) AddOrReplace(cacheObjectKey *CacheObjectKey, generation int64, rc io.ReadCloser) (gcsx.TempFile, error) {
+	if _, exists := c.fileMap[*cacheObjectKey]; exists {
+		c.fileMap[*cacheObjectKey].Destroy()
+	}
+	metadata := &gcsx.TempFileObjectMetadata{
+		BucketName: cacheObjectKey.BucketName,
+		ObjectName: cacheObjectKey.ObjectName,
+		Generation: generation,
+	}
+	file, err := c.NewCacheFile(rc, metadata)
```
lezh, line 64: If there is an error, should we proceed with the file?

internal/contentcache/contentcache.go
```diff
@@ -50,3 +50,38 @@ func New(tempDir string, mtimeClock timeutil.Clock) *ContentCache {
 func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
+
+// Function to add or update existing cache file
+func (c *ContentCache) AddOrReplace(cacheObjectKey *CacheObjectKey, generation int64, rc io.ReadCloser) (gcsx.TempFile, error) {
+	if _, exists := c.fileMap[*cacheObjectKey]; exists {
```
lezh, line 56: If we already extracted a value from the map, why do we do it again in the next line?

lezh (changes requested):

internal/contentcache/contentcache.go
thepenguinco: Will add a comment, this in memory map is not concurrent safe.

internal/contentcache/contentcache.go
thepenguinco: We do not include it in our map and return err to the caller.

internal/contentcache/contentcache.go
```diff
@@ -51,6 +52,38 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
 	return gcsx.NewTempFile(rc, c.tempDir, c.mtimeClock)
 }
 
+// Add or replace creates a new cache file or updates an existing cache file
```
lezh, line 55:
```
// AddOrReplace creates a new cache file or updates an existing cache file
```

internal/contentcache/contentcache.go
```diff
@@ -51,6 +52,38 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
...
+		cacheObject.Destroy()
+	}
+	metadata := &gcsx.TempFileObjectMetadata{
+		BucketName: cacheObjectKey.BucketName,
+		ObjectName: cacheObjectKey.ObjectName,
+		Generation: generation,
+	}
+	file, err := c.NewCacheFile(rc, metadata)
+	if err != nil {
+		return nil, err
```
lezh, line 67: Use `fmt.Errorf` to explain where the error comes from

internal/contentcache/contentcache.go
```diff
@@ -51,6 +52,38 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
...
+// Get retrieves a file from the cache given the GCS object name and bucket name
+func (c *ContentCache) Get(cacheObjectKey *CacheObjectKey) (gcsx.TempFile, bool) {
+	file, exists := c.fileMap[*cacheObjectKey]
+	return file, exists
+}
+
+// Remove removes and destroys the specfied cache file and metadata on disk
+func (c *ContentCache) Remove(cacheObjectKey *CacheObjectKey) {
+	if _, exists := c.fileMap[*cacheObjectKey]; exists {
+		c.fileMap[*cacheObjectKey].Destroy()
```
lezh, line 82: This is mentioned in a previous comment still unresolved. Unnecessary map lookup.

lezh (changes requested):

internal/contentcache/contentcache.go
```diff
@@ -51,6 +53,38 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
...
+	file, exists := c.fileMap[*cacheObjectKey]
+	return file, exists
+}
+
+// Remove removes and destroys the specfied cache file and metadata on disk
+func (c *ContentCache) Remove(cacheObjectKey *CacheObjectKey) {
+	if cacheObject, exists := c.fileMap[*cacheObjectKey]; exists {
+		cacheObject.Destroy()
+	}
+	delete(c.fileMap, *cacheObjectKey)
```
lezh, line 85: Why do we delete the entry even if the key does not exist?

lezh (changes requested):

internal/contentcache/contentcache.go
thepenguinco: Delete should be a no-op if the key doesn't exist in the map, but I can move it inside.

lezh: Please move it inside for better readability and less overhead.

PR #605, opened 2022-02-10
Create pre-populate cache plumbing

thepenguinco:
Create populate cache function and mark TODO.

Run go test ./... and no unit tests should fail.

TODO: add testing once the function implementation is complete.

### DIFF
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -16,7 +16,11 @@
 package contentcache
 
 import (
+	"fmt"
 	"io"
+	"io/ioutil"
+	"log"
+	"regexp"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
 	"github.com/jacobsa/timeutil"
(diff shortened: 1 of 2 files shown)


internal/contentcache/contentcache.go
```diff
@@ -36,6 +40,34 @@ type ContentCache struct {
...
 }
 
+// Populates the cache with existing persisted files when gcsfuse starts
+func (c *ContentCache) PopulateCache() {
+	if c.tempDir == "" {
+		c.tempDir = "/tmp"
+	}
+	files, err := ioutil.ReadDir(c.tempDir)
+	if err != nil {
+		log.Fatal(err)
```
lezh, line 50: This can happen. For instance, if the disk is network attached, it's possible to lose the connection to the dir. To crash the gcsfuse process seems too aggressive to handle such a condition.

internal/contentcache/contentcache.go
```diff
@@ -36,6 +40,34 @@ type ContentCache struct {
 	mtimeClock     timeutil.Clock
 }
 
+// Populates the cache with existing persisted files when gcsfuse starts
+func (c *ContentCache) PopulateCache() {
```
lezh, line 44: PopulateCache sounds like adding stuff to an empty cache. Would "RecoverCache" or "SetupCache" be a better name?

internal/fs/fs.go
```diff
@@ -121,13 +121,16 @@ func NewFileSystem(
 
 	mtimeClock := timeutil.RealClock()
 
+	contentCache := contentcache.New(cfg.TempDir, mtimeClock)
+	contentCache.PopulateCache()
```
lezh, line 125: What if the gcsfuse doesn't want a cache? Would this scan the `/tmp` unnecessarily?

internal/contentcache/contentcache.go
```diff
@@ -36,6 +40,34 @@ type ContentCache struct {
...
+	}
+	files, err := ioutil.ReadDir(c.tempDir)
+	if err != nil {
+		log.Fatal(err)
+	}
+	for _, file := range files {
+		// validate not a directory and matches gcsfuse pattern
+		if !file.IsDir() && matchPattern(file.Name()) {
+			// TODO ezl: load the files from disk to the in memory map
+			fmt.Printf(file.Name())
```
lezh, line 56: This is for debug purpose, which should not be merged into the master. We don't want this message to show up in a production environment distracting the user.

lezh (changes requested):

internal/contentcache/contentcache.go
thepenguinco: RecoverCache makes sense.

internal/fs/fs.go
thepenguinco: I will change this so "RecoverCache()" is only called when the user specifies the content cache flag.

internal/contentcache/contentcache.go
thepenguinco: I agree.  A better solution is to disable the content cache and print a log to the console if the cache directory cannot be read.

lezh (changes requested):

PR #606, opened 2022-02-11
Run tests during CI

gaul:
Tests previously ran in Travis but were accidentally removed in
2dee82339ae07fc8f2d8b2a7b592c560ab2ce2b4.  This would have prevented a
regression introduced in #597.  References #590.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,5 +26,6 @@ jobs:
     - name: Build
       run: |
         go build ./...
+        go test -v ./...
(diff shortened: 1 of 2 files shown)


gaul:
When running tests locally, `fs.test` hangs as a zombie.

gaul:
These tests pass locally but fail in CI.  But at least they are now running?

thepenguinco:
They seem to be hanging.   I suspect it has to do with how the tests are using the tmp directory to read/write files (see local_modifications_test.go for example).

PR #608, opened 2022-02-15
gcsfuse use content cache when enabled

thepenguinco:
PR is larger since many changes are required to allow gcsfuse to serve data from the content cache.


internal/fs/inode/file.go
```diff
@@ -99,13 +100,14 @@ func NewFileInode(
...
-		name:         name,
-		attrs:        attrs,
-		contentCache: contentCache,
-		src:          *o,
+		bucket:         bucket,
+		mtimeClock:     mtimeClock,
+		id:             id,
+		name:           name,
+		attrs:          attrs,
+		localFileCache: localFileCache,
```
lezh, line 108: Can we use `contentCache != nil` to represent if `localFileCache` is enabled?

internal/fs/inode/file.go
```diff
@@ -178,38 +180,74 @@ func (f *FileInode) clobbered(ctx context.Context) (b bool, err error) {
 	return
 }
 
-// Ensure that f.content != nil
+// Helper method to open a reader
+// (should only be called by methods that have locks)
```
lezh, line 184: What locks? Is it `f.mu`? Then, it is already mentioned in line 186. This comment is redundant.

internal/fs/inode/file.go
```diff
@@ -178,38 +180,74 @@ func (f *FileInode) clobbered(ctx context.Context) (b bool, err error) {
 	return
 }
 
-// Ensure that f.content != nil
+// Helper method to open a reader
```
lezh, line 183: This does not provide any information in addition to the function name `openReader`. You can move the line 188 here which is more informative about the reader.

internal/fs/inode/file.go
```diff
@@ -178,38 +180,74 @@ func (f *FileInode) clobbered(ctx context.Context) (b bool, err error) {
...
+				return
+			}
+		}
 
-	// Update state.
-	f.content = tf
+		// open reader
+		rc, err := f.openReader(ctx)
+		if err != nil {
+			return err
```
lezh, line 220: All the
```
return err
```
must be replaced with 
```
return fmt.Errorf("where the error comes from: %w", err)
```

Otherwise it's almost impossible to debug

internal/fs/inode/file.go
```diff
@@ -178,38 +180,74 @@ func (f *FileInode) clobbered(ctx context.Context) (b bool, err error) {
...
-		err = fmt.Errorf("NewTempFile: %w", err)
-		return
-	}
+// Ensure that content exists and is not stale
+//
+// LOCKS_REQUIRED(f.mu)
+func (f *FileInode) ensureContent(ctx context.Context) (err error) {
+	if f.localFileCache {
+		// Fetch content from the cache
+		// we validate generation numbers here again
```
lezh, line 207: Be consistent with the comments. All the comments start with a capitalized letter.

internal/fs/inode/file.go
```diff
@@ -178,38 +180,74 @@ func (f *FileInode) clobbered(ctx context.Context) (b bool, err error) {
...
+		if file, exists := f.contentCache.Get(cacheObjectKey); exists {
+			if file.ValidateGeneration(f.src.Generation) {
+				f.content = file
+				return
+			}
+		}
 
-	// Update state.
-	f.content = tf
+		// open reader
```
lezh, line 217: Useless comment

lezh (changes requested):

lezh (changes requested):
This could be done in a few PRs with the proposed changes in the comments:

PR list
1. Decouple Syncer from FileInode: The latter should delete the temp file once syncing is done.
2. Change `ContentCache` to an interface with `EnsureContent` and `DestroyContent` functions. The `contentcache.New` function should take a flag indicating if the read path is cached.
3. Create a `writeOnlyContentCache` and tests
4. Migrate the current file inode to use `writeOnlyContentCache`
5. Create a `readWriteContentCache` and tests
6. Enable the file inodes to use `readWriteContentCache`

thepenguinco:
> This could be done in a few PRs with the proposed changes in the comments:
> 
> PR list
> 
> 1. Decouple Syncer from FileInode: The latter should delete the temp file once syncing is done.
> 2. Change `ContentCache` to an interface with `EnsureContent` and `DestroyContent` functions. The `contentcache.New` function should take a flag indicating if the read path is cached.
> 3. Create a `writeOnlyContentCache` and tests
> 4. Migrate the current file inode to use `writeOnlyContentCache`
> 5. Create a `readWriteContentCache` and tests
> 6. Enable the file inodes to use `readWriteContentCache`

I will make these changes in future PRs, current PR should be ready for review.

internal/fs/fs.go
```diff
@@ -121,12 +121,15 @@ func NewFileSystem(
...
+	// TODO ezl Consider setting contentcache to nil when the local file cache is disabled
+	// This means we need to refactor and separate the local file cache logic away from
+	// temp files being used when users want to write/modify files in the filesystem
 	contentCache := contentcache.New(cfg.TempDir, mtimeClock)
 
 	if cfg.LocalFileCache {
 		err := contentCache.RecoverCache()
 		if err != nil {
-			fmt.Printf("Encountered error retrieving files from cache directory, disabling local file cache: %v", err)
+			fmt.Printf("encountered error retrieving files from cache directory, disabling local file cache: %v", err)
```
lezh, line 132: Logs (`fmt.Printf`) should start with capitalized letter whereas error strings (`fmt.Errorf`) should not.

internal/gcsx/temp_file.go
```diff
@@ -227,8 +226,14 @@ func (tf *tempFile) CheckInvariants() {
 
 func (tf *tempFile) Destroy() {
 	tf.state = fileDestroyed
-	// Throw away the file.
+	// Throw away the file (for anonymous files).
 	tf.f.Close()
+
+	// TODO ezl: move this logic out once we refactor cache disk file handling outside of gcsx.Tempfile
+	// Throw away the cache file and metadata
+	os.Remove(fmt.Sprintf("%s.json", tf.f.Name()))
```
lezh, line 234: When `local-file-cache` is true, we should not evict any cache entries (not for milestone 1/2). 

When `local-file-cache` is false, we should not have `json` files.

So, I'm confused when this would be needed.

internal/gcsx/temp_file.go
```diff
@@ -227,8 +226,14 @@ func (tf *tempFile) CheckInvariants() {
...
+	os.Remove(tf.f.Name())
```
lezh, line 235: Same as the comment above. Why do we need to delete this file?

internal/gcsx/syncer.go
```diff
@@ -154,7 +154,7 @@ type syncer struct {
 func (os *syncer) SyncObject(
 	ctx context.Context,
 	srcObject *gcs.Object,
-	content TempFile) (o *gcs.Object, err error) {
+	content TempFile, localFileCache bool) (o *gcs.Object, err error) {
```
lezh, line 157: Can we make the changes of syncer.go out of this PR? It isn't relevant as syncer is only used for the write path.

For the read path, the cache files won't be deleted.
For the write path, the cache files have to be deleted anyway.

So, I'm not sure why this has to be included in this PR.

internal/fs/inode/file.go
```diff
@@ -100,24 +101,25 @@ func NewFileInode(
...
 	f.lc.Init(id)
 
 	// Set up invariant checking.
 	f.mu = syncutil.NewInvariantMutex(f.checkInvariants)
 
-	if localFileCache {
-		// The gcs object is cached as local temp file
-		f.ensureContent(context.Background())
-	}
+	// if localFileCache {
```
lezh, line 119: Remove the code that has been commented out.

internal/fs/inode/file.go
```diff
@@ -46,10 +46,11 @@ type FileInode struct {
...
 
-	id           fuseops.InodeID
-	name         Name
-	attrs        fuseops.InodeAttributes
-	contentCache *contentcache.ContentCache
+	id             fuseops.InodeID
+	name           Name
+	attrs          fuseops.InodeAttributes
+	contentCache   *contentcache.ContentCache
+	localFileCache bool
```
lezh, line 53: Leave a TODO here to remove this bool flag by refactoring the content cache using two different implementations.

lezh (changes requested):

internal/gcsx/temp_file.go
thepenguinco: This is to maintain cache integrity.  The cache design already allows for dynamic writes and cache eviction when files become stale (out of sync with gcs).

internal/gcsx/syncer.go
thepenguinco: I have rolled the changes from #614 here so that we don't modify the syncer interface.  However, in order to preserve original functionality, I have added a check to destroy the temp file if local file cache is disabled.

internal/fs/inode/file.go
thepenguinco: This is mistakenly commented out.  Thanks for catching.

internal/gcsx/temp_file.go
thepenguinco: Same as above.  We clear these files whenever destroy is called.  These should be moved once we decouple cache files from gcsx.TempFile.

internal/gcsx/syncer.go
```diff
@@ -219,8 +219,10 @@ func (os *syncer) SyncObject(
 		return
 	}
 
-	// Destroy the temp file.
-	content.Destroy()
+	// Destroy the temp file after syncing if the local file cache is disabled
+	if !localFileCache {
+		content.Destroy()
```
lezh, line 224: This is no longer needed once it is rebased onto https://github.com/GoogleCloudPlatform/gcsfuse/pull/616

lezh (changes requested):

internal/fs/inode/file.go
```diff
@@ -279,11 +315,12 @@ func (f *FileInode) DecrementLookupCount(n uint64) (destroy bool) {
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) Destroy() (err error) {
 	f.destroyed = true
-
-	if f.content != nil {
+	if f.localFileCache {
+		cacheObjectKey := &contentcache.CacheObjectKey{BucketName: f.bucket.Name(), ObjectName: f.name.objectName}
+		f.contentCache.Remove(cacheObjectKey)
```
lezh, line 320: Why not calling this?
```
f.content.Destroy()
```

lezh (changes requested):

internal/fs/inode/file.go
thepenguinco: f.contentCache.Remove() calls Destroy on the tempfile retrieved from the map.

internal/fs/inode/file.go
```diff
@@ -503,7 +540,7 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 	}
 
 	// If we wrote out a new object, we need to update our state.
-	if newObj != nil {
+	if newObj != nil && !f.localFileCache {
```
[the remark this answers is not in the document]
thepenguinco: We don't delete content because we support dynamic writing in the cache.  The file on disk doesn't need to be deleted as we apply writes to that file in cache whenever we make edits to the file in the VFS.  The cache AddOrReplace method handles the case when the file on disk is stale (mismatching generation numbers) and we pull a fresh file from GCS. 

There is however an edge case that I haven't tested for, the metadata file locally is only updated when we discover a stale file on disk during lookup and we pull the file, so that means if we write to our file (updating our cache file), it is possible we have to do an unnecessary pull of the file next time gcsfuse starts up again as we still have the old generation number in our metadata file.  Not sure if this is worth handling right away.

PR #611, opened 2022-02-16
gcsfuse Recover Cache

thepenguinco:
Recover the cache from disk on gcsfuse startup.

Depends on #608.

### DIFF
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -17,12 +17,20 @@
 package contentcache
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 	"github.com/jacobsa/timeutil"
 )
 
@@ -34,9 +42,45 @@ type CacheObjectKey struct {
 
 // ContentCache is a directory on local disk to store the object content
 type ContentCache struct {
+	debug      *log.Logger
 	tempDir    string
 	fileMap    map[CacheObjectKey]gcsx.TempFile
 	mtimeClock timeutil.Clock
+}
+
+// RecoverFileFromCache recovers a file from the cache via metadata
+func (c *ContentCache) RecoverFileFromCache(metadataFile fs.FileInfo) {
+	// validate not a directory and matches gcsfuse pattern
+	if metadataFile.IsDir() {
+		return
+	}
+	if !matchPattern(metadataFile.Name()) {
+		return
+	}
+	var metadata gcsx.TempFileObjectMetadata
+	metadataAbsolutePath := path.Join(c.tempDir, metadataFile.Name())
+	contents, err := ioutil.ReadFile(metadataAbsolutePath)
+	// TODO ezl should we exec some sort of cleanup in the cache if there are errors
+	if err != nil {
+		c.debug.Printf("Skip metadata file %v due to read error: %s", metadataFile.Name(), err)
+		return
+	}
+	err = json.Unmarshal(contents, &metadata)
+	if err != nil {
+		c.debug.Printf("Skip metadata file %v due to file corruption: %s", metadataFile.Name(), err)
+		return
+	}
+	cacheObjectKey := &CacheObjectKey{BucketName: metadata.BucketName, ObjectName: metadata.ObjectName}
+	// TODO ezl we should probably store the cached file name inside the cache file metadata instead of using the json file name
+	fileName := strings.TrimSuffix(metadataAbsolutePath, filepath.Ext(metadataFile.Name()))
+	// TODO ezl linux fs limits single process to open max of 1024 file descriptors
+	// so this is not scalable
+	file, err := os.Open(fileName)
+	if err != nil {
+		c.debug.Printf("Skip cache file %v due to error: %s", fileName, err)
+		return
+	}
+	c.AddOrReplace(cacheObjectKey, metadata.Generation, file)
 }
 
 // RecoverCache recovers the cache with existing persisted files when gcsfuse starts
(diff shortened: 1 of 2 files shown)


internal/contentcache/contentcache.go
```diff
@@ -44,15 +48,39 @@ func (c *ContentCache) RecoverCache() error {
 	if c.tempDir == "" {
 		c.tempDir = "/tmp"
 	}
+	fmt.Printf("Recovering cache:\n")
```
lezh, line 51: Let's use `logger.Infof` instead of `fmt.Printf` (in all places) because the latter does not write to log files.

internal/contentcache/contentcache.go
```diff
@@ -44,15 +48,39 @@ func (c *ContentCache) RecoverCache() error {
...
-		if !file.IsDir() && matchPattern(file.Name()) {
-			// TODO ezl: load the files from disk to the in memory map
+		if !metadataFile.IsDir() && matchPattern(metadataFile.Name()) {
+			var metadata gcsx.TempFileObjectMetadata
+			// TODO ezl inefficient string operation
+			metadataAbsolutePath := c.tempDir + "/" + metadataFile.Name()
+			contents, err := ioutil.ReadFile(metadataAbsolutePath)
+			// TODO ezl should we exec some sort of cleanup in the cache if there are errors
+			if err != nil {
+				fmt.Printf("skip metadata file %v due to read error: %s", metadataFile.Name(), err)
```
lezh, line 66: This is a DEBUG message. Use `logger.NewDebug` to create a logger at DEBUG level and use that to write the debug logs. Logs should start with capitalized letters.

internal/contentcache/contentcache.go
```diff
@@ -44,15 +48,39 @@ func (c *ContentCache) RecoverCache() error {
...
+				fmt.Printf("skipping...")
```
lezh, line 67: This line is useless. We should reduce the noise of any logs (in production they cost money).

internal/contentcache/contentcache.go
```diff
@@ -44,15 +48,39 @@ func (c *ContentCache) RecoverCache() error {
...
 	if err != nil {
 		// if we fail to read the specified directory, log and return error
 		return fmt.Errorf("recover cache: %w", err)
 	}
-	for _, file := range files {
+	for _, metadataFile := range files {
 		// validate not a directory and matches gcsfuse pattern
-		if !file.IsDir() && matchPattern(file.Name()) {
-			// TODO ezl: load the files from disk to the in memory map
+		if !metadataFile.IsDir() && matchPattern(metadataFile.Name()) {
```
lezh, line 59: This entire chunk should be moved to a helper function.

internal/contentcache/contentcache.go
```diff
@@ -61,7 +89,7 @@ func (c *ContentCache) RecoverCache() error {
 // Helper function that matches the format of a gcsfuse file
 func matchPattern(fileName string) bool {
 	// TODO ezl: replace with constant defined in gcsx.TempFile
-	match, err := regexp.MatchString(fmt.Sprintf("gcsfuse[0-9]+[.]json"), fileName)
+	match, err := regexp.MatchString(fmt.Sprintf("gcsfusecache[0-9]+[.]json"), fileName)
```
lezh, line 92: I see the file name formats used by both contentcache.go and temp_file.go. They should be controlled in one place.

lezh (changes requested):

internal/contentcache/contentcache.go
thepenguinco: I will temporarily move this to gcsx, but in reality we really should be using the new package.

thepenguinco: Next PR will be a refactor.

internal/contentcache/contentcache.go
```diff
@@ -39,29 +45,59 @@ type ContentCache struct {
 	mtimeClock timeutil.Clock
 }
 
+// RecoverFileFromCache recovers a file from the cache via metadata
+func (c *ContentCache) RecoverFileFromCache(metadataFile fs.FileInfo) {
+	debug := logger.NewDebug("recover file from cache: ")
```
thepenguinco, line 50: I don't think we need to keep creating a new debug logger for each file.

internal/contentcache/contentcache.go
```diff
@@ -39,29 +45,59 @@ type ContentCache struct {
...
+	// validate not a directory and matches gcsfuse pattern
+	if !metadataFile.IsDir() && matchPattern(metadataFile.Name()) {
```
lezh, line 52: nit: This function is only executed when these 2 conditions are true. A better way to express this is:
```go
if maybeMetadataFile.IsDir() {
  return
}
if !isMetadataFileName(maybeMetadataFile.Name()) {
  return
}
// recover the metadata file ...
```

internal/contentcache/contentcache.go
```diff
@@ -39,29 +45,59 @@ type ContentCache struct {
...
+		var metadata gcsx.TempFileObjectMetadata
+		// TODO ezl handle '/' char by stripping it and appending ourselves
+		// so that the user can specify folder name with or without '/'
+		metadataAbsolutePath := fmt.Sprintf("%v/%v", c.tempDir, metadataFile.Name())
```
lezh, line 56: You can use [path.Join](https://pkg.go.dev/path#Join)

internal/contentcache/contentcache.go
```diff
@@ -39,29 +45,59 @@ type ContentCache struct {
...
+		}
+		err = json.Unmarshal(contents, &metadata)
+		if err != nil {
+			debug.Printf("Skip metadata file %v due to file corruption: %s", metadataFile.Name(), err)
+			return
+		}
+		cacheObjectKey := &CacheObjectKey{BucketName: metadata.BucketName, ObjectName: metadata.ObjectName}
+		// TODO ezl we should probably store the cached file name inside the cache file metadata instead of using the json file name
+		fileName := strings.TrimSuffix(metadataAbsolutePath, filepath.Ext(metadataFile.Name()))
+		file, err := os.Open(fileName)
```
lezh, line 71: Some OS has a limit for open file to be 1024. So, this is not scalable. A better way to handle it is to open on demand.

lezh (changes requested):

internal/contentcache/contentcache.go
thepenguinco: I left a TODO, will address in refactor.

PR #618, opened 2022-02-26
Create prefetch tool to cache gcs objects on disk

thepenguinco:
Creates a catch prefetch tool to download files into local disk cache from specified bucket.  We use a typical single-producer multi-consumer concurrency pattern.

Usage:
`prefetch_cache_gcsfuse cache_dir bucket_name [prefix]`

To test, run the binary with the specified command line arguments.

Verify cache_dir is populated with files and metadata files.

Next try running gcsfuse specifying --local-file-cache flag with the value of the same cache_dir.  This will allow gcsfuse to use the cached files on disk.

The constant NUM_WORKERS specifies how many consumer workers to download files are created.

A higher NUM_WORKERS value creates more goroutines which can speed up downloads (up to a limit).

In the future, NUM_WORKERS can be a configurable command line flag exposed to the user.

### DIFF
--- a/tools/prefetch_cache_gcsfuse/main.go
+++ b/tools/prefetch_cache_gcsfuse/main.go
@@ -1,0 +1,72 @@
+// Copyright 2022 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Prefetches files into local disk cache from specified bucket
+//
+// Usage:
+//
+//     prefetch_cache_gcsfuse cache_dir bucket_name [prefix]
+//
+// This will prefetch the cache files from the specified bucket
+// with an optional file prefix to filter the GCS objects
+// and download them into the specified cache directory
+//
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+)
+
+func run(args []string) (err error) {
+	// Extract arguments.
+	if len(args) < 2 || len(args) > 3 {
+		err = fmt.Errorf("Usage: %s cache_dir bucket_name [prefix]", os.Args[0])
+		return
+	}
+
+	cacheDir := args[0]
+	bucketName := args[1]
+	var prefix string
+
+	if len(args) > 2 {
+		prefix = args[2]
+	}
+
+	log.Printf("Using settings:")
+	log.Printf("  cacheDir:  %s", cacheDir)
+	log.Printf("  bucketName:  %s", bucketName)
+	log.Printf("  prefix:  %s", prefix)
+
+	// Try to preload persistent cache on disk using specified parameters
+	err = prefetchCache(cacheDir, bucketName, prefix)
+	if err != nil {
+		err = fmt.Errorf("prefetch_cache_gcsfuse: %w", err)
+	}
+
+	return
+}
+
+func main() {
+	log.SetFlags(log.Lmicroseconds)
+	flag.Parse()
+
+	err := run(flag.Args())
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
--- a/tools/prefetch_cache_gcsfuse/prefetch.go
+++ b/tools/prefetch_cache_gcsfuse/prefetch.go
@@ -1,0 +1,147 @@
+// Copyright 2022 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"cloud.google.com/go/storage"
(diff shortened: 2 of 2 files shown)


tools/prefetch_cache_gcsfuse/main.go
```diff
@@ -0,0 +1,77 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Prefetches files into local disk cache from specified bucket
+//
+// Usage:
+//
+//     prefetch_cache_gcsfuse cache_dir bucket_name
+//
+// This will prefetch the cache file
```
thepenguinco, line 21: Will update this comment

tools/prefetch_cache_gcsfuse/main.go
```diff
@@ -0,0 +1,77 @@
...
+	cacheDir := args[0]
+	bucketName := args[1]
+	var prefix string
+	var delimiter string
+
+	if len(args) > 2 {
+		prefix = args[2]
+	}
+
+	// TODO ezl delimiter is not quite supported yet
```
lezh, line 48: We don't need to specify the delimiter in the args.

The only use case for the preloader is:
```
> preload_gcsfuse /tmp/cache/path/ gs://bucket-name/foo/
```

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if err != nil {
+		err = fmt.Errorf("ioutil.TempFile: %w", err)
+		return
+	}
+
+	rc, err := client.Bucket(object.Bucket).Object(object.Name).NewReader(ctx)
+
+	if err != nil {
+		return fmt.Errorf("Object(%q).NewReader: %v", object.Name, err)
```
lezh, line 58: Do we need to call `f.Close()` before this return?

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if _, err := io.Copy(f, rc); err != nil {
+		return fmt.Errorf("io.Copy: %v", err)
+	}
+
+	if err = f.Close(); err != nil {
+		return fmt.Errorf("f.Close: %v", err)
+	}
+
+	file, err := json.MarshalIndent(*tempFileObjectMetadata, "", " ")
```
lezh, line 70: The `file` is a variable that stores metadata, which has nothing to do with file. Prefer variable names like `metadata` or `file_content`

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,142 @@
...
+
+func prefetchCache(cacheDir, bucketName, prefix, delimiter string) (err error) {
+	start := time.Now()
+
+	var wg sync.WaitGroup
+
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %v", err)
```
lezh, line 86: Use `%w` instead of `%v`.

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,142 @@
...
+				log.Printf("prefetchCache: %v", err)
+			}
+		}()
+	}
+
+	// Wait for all goroutines downloading files to finish before returning
+	wg.Wait()
+
+	elapsed := time.Since(start)
+	log.Printf("Prefetch cache took %s", elapsed)
```
lezh, line 139: It would be nice to print out stats about how many objects are attempted vs. downloaded.

lezh (changes requested):

tools/prefetch_cache_gcsfuse/main.go
thepenguinco: Got it.  I've updated the usage as indicated in the comment:

`prefetch_cache_gcsfuse cache_dir bucket_name [prefix]`

tools/prefetch_cache_gcsfuse/prefetch.go
thepenguinco: I added a defer instead.

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// Should we set a higher timeout or let this be configurable by the user
+	ctx, cancel := context.WithTimeout(ctx, time.Minute*5)
+	defer cancel()
+
+	it := client.Bucket(bucketName).Objects(ctx, &storage.Query{
+		Prefix: prefix,
+	})
+
+	// Concurrently download files from specified gcs bucket with prefix and delimiter
```
lezh, line 98: This can potentially throttle the API or the disk by creating too many concurrent connections.

A better way of handling such a task is:
1. Create two channels: `pendingTasks` and `finishedTasks`
2. Create a pool of N workers by running `go func() {...}` for each worker as a goroutine. Then, each worker is going to take one `pendingTask`, do it, and put one in `finishedTask`.
3. The main goroutine calls `it.Next()` to get the objects, push them into the `pendingTasks`.
4. The main goroutine keeps getting the `finishedTasks` until all the tasks are finished.

When N = 1, this implementation becomes sequential.

lezh (changes requested):

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		err = fmt.Errorf("ioutil.TempFile: %w", err)
+		return
+	}
+	defer f.Close()
+
+	rc, err := client.Bucket(object.Bucket).Object(object.Name).NewReader(ctx)
+
+	if err != nil {
+		return fmt.Errorf("Object(%q).NewReader: %v", object.Name, err)
```
lezh, line 64: Use `%w` for errors, in ALL `fmt.Errorf` calls.

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,147 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
lezh, line 1: Use 2022.

lezh (changes requested):
Looks good overall. A couple of nits

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -0,0 +1,147 @@
...
+
+	rc, err := client.Bucket(object.Bucket).Object(object.Name).NewReader(ctx)
+
+	if err != nil {
+		return fmt.Errorf("Object(%q).NewReader: %w", object.Name, err)
+	}
+	defer rc.Close()
+
+	if _, err := io.Copy(f, rc); err != nil {
+		return fmt.Errorf("io.Copy: %v", err)
```
lezh, line 69: One more thing here

lezh (changes requested):

PR #624, opened 2022-03-09
Refactor content cache to use wrapper structs

thepenguinco:
Refactor content cache to support dynamic cache feature implementation.

To test, verify the read-only cache still works and that there is no change in behaviour.

### DIFF
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -25,14 +25,14 @@ import (
 	"log"
 	"os"
 	"path"
-	"path/filepath"
 	"regexp"
-	"strings"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 	"github.com/jacobsa/timeutil"
 )
+
+const CACHE_FILE_PREFIX = "gcsfusecache"
 
 // CacheObjectKey uniquely identifies GCS objects by bucket name and object name
 type CacheObjectKey struct {
@@ -44,8 +44,50 @@ type CacheObjectKey struct {
 type ContentCache struct {
 	debug      *log.Logger
 	tempDir    string
-	fileMap    map[CacheObjectKey]gcsx.TempFile
+	fileMap    map[CacheObjectKey]*CacheObject
 	mtimeClock timeutil.Clock
+}
+
+// Metadata store struct
+type CacheFileObjectMetadata struct {
+	CacheFileNameOnDisk string
+	BucketName          string
+	ObjectName          string
+	Generation          int64
+	MetaGeneration      int64
+}
+
+type CacheObject struct {
+	MetadataFileName        string
+	CacheFileObjectMetadata *CacheFileObjectMetadata
+	CacheFile               gcsx.TempFile
+}
+
+func (c *CacheObject) ValidateGeneration(generation int64) bool {
+	if c.CacheFileObjectMetadata == nil {
+		return false
+	}
+	return c.CacheFileObjectMetadata.Generation == generation
+}
+
+func (c *ContentCache) WriteMetadataCheckpointFile(cacheFileName string, cacheFileObjectMetadata *CacheFileObjectMetadata) (metadataFileName string, err error) {
+	var file []byte
+	file, err = json.MarshalIndent(cacheFileObjectMetadata, "", " ")
+	if err != nil {
+		err = fmt.Errorf("json.MarshalIndent failed for object metadata: %w", err)
+		return
+	}
+	metadataFileName = fmt.Sprintf("%s.json", cacheFileName)
+	err = ioutil.WriteFile(metadataFileName, file, 0644)
+	if err != nil {
+		err = fmt.Errorf("WriteFile for JSON metadata: %w", err)
+		return
+	}
+	return
+}
+
+func (c *CacheObject) Destroy() {
+	// TODO ezl clean up
 }
 
 // RecoverFileFromCache recovers a file from the cache via metadata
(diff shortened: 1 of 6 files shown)


internal/contentcache/contentcache.go
```diff
@@ -127,27 +165,31 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
...
+	metadata := &CacheFileObjectMetadata{
+		CacheFileNameOnDisk: file.Name(),
+		BucketName:          cacheObjectKey.BucketName,
+		ObjectName:          cacheObjectKey.ObjectName,
+		Generation:          generation,
 	}
-	file, err := c.NewCacheFile(rc, metadata)
+	var metadataFileName string
+	metadataFileName, err = c.WriteMetadataCheckpointFile(file.Name(), metadata)
+	cacheObject := &CacheObject{MetadataFileName: metadataFileName, CacheFileObjectMetadata: metadata, CacheFile: file}
```
lezh, line 181: 1. Please use multiple lines to specify the fields
2. This line should be after the error handling

internal/contentcache/contentcache.go
```diff
@@ -127,27 +165,31 @@ func (c *ContentCache) NewTempFile(rc io.ReadCloser) (gcsx.TempFile, error) {
...
-func (c *ContentCache) AddOrReplace(cacheObjectKey *CacheObjectKey, generation int64, rc io.ReadCloser) (gcsx.TempFile, error) {
+func (c *ContentCache) AddOrReplace(cacheObjectKey *CacheObjectKey, generation int64, rc io.ReadCloser) (*CacheObject, error) {
 	if cacheObject, exists := c.fileMap[*cacheObjectKey]; exists {
 		cacheObject.Destroy()
 	}
-	metadata := &gcsx.TempFileObjectMetadata{
-		BucketName: cacheObjectKey.BucketName,
-		ObjectName: cacheObjectKey.ObjectName,
-		Generation: generation,
+	file, err := c.NewCacheFile(rc)
```
lezh, line 172: This error is not handled

tools/prefetch_cache_gcsfuse/prefetch.go
```diff
@@ -36,21 +36,19 @@ const NUM_WORKERS = 10
...
-	// Create a temporary cache file on disk
-	// TODO ezl concerns about partial download and cache file corruption
 	// We may want a way to verify the files are fully downloaded
 	// and either resuming the download or discarding and redownloading the file
 	// We may also want to do cleanup if files are created on disk but aren't populated in time
-	f, err := ioutil.TempFile(cacheDir, gcsx.CACHE_FILE_PREFIX)
+
+	f, err := ioutil.TempFile(cacheDir, contentcache.CACHE_FILE_PREFIX)
+
+	metadata := &contentcache.CacheFileObjectMetadata{
```
lezh, line 45: This shouldn't be executed if `err != nil`

internal/contentcache/contentcache.go
```diff
@@ -44,10 +44,51 @@ type CacheObjectKey struct {
 type ContentCache struct {
 	debug      *log.Logger
 	tempDir    string
-	fileMap    map[CacheObjectKey]gcsx.TempFile
+	fileMap    map[CacheObjectKey]*CacheObject
 	mtimeClock timeutil.Clock
 }
 
+// Metadata store struct
+type CacheFileObjectMetadata struct {
```
lezh, line 52: This needs a unit test

internal/contentcache/contentcache.go
```diff
@@ -44,10 +44,51 @@ type CacheObjectKey struct {
...
+func (c *ContentCache) WriteMetadataCheckpointFile(cacheFileName string, cacheFileObjectMetadata *CacheFileObjectMetadata) (metadataFileName string, err error) {
+	var file []byte
+	file, err = json.MarshalIndent(cacheFileObjectMetadata, "", " ")
+	if err != nil {
+		return
+	}
+	metadataFileName = fmt.Sprintf("%s.json", cacheFileName)
+	err = ioutil.WriteFile(metadataFileName, file, 0644)
+	if err != nil {
+		err = fmt.Errorf("%w", err)
```
lezh, line 82: This is useless because it keeps the error string unchanged

internal/contentcache/contentcache.go
```diff
@@ -44,10 +44,51 @@ type CacheObjectKey struct {
...
+		return false
+	}
+	return c.CacheFileObjectMetadata.Generation == generation
+}
+
+func (c *ContentCache) WriteMetadataCheckpointFile(cacheFileName string, cacheFileObjectMetadata *CacheFileObjectMetadata) (metadataFileName string, err error) {
+	var file []byte
+	file, err = json.MarshalIndent(cacheFileObjectMetadata, "", " ")
+	if err != nil {
+		return
```
lezh, line 77: Add an error prefix to distinguish the origin of the error

lezh (changes requested):

internal/contentcache/contentcache.go
thepenguinco: I will make a separate PR to create some basic unit tests for content_cache.go

lezh: The unit test for X should be put in the same PR where X is introduced.

A recommended fix for this PR is to create another PR2 to introduce CacheFileObjectMetadata and its test, and rebase this PR onto the PR2.

lezh (changes requested):

internal/contentcache/contentcache_test.go
```diff
@@ -0,0 +1,65 @@
...
+func TestValidateGeneration(t *testing.T) {
+	objectMetadata := contentcache.CacheFileObjectMetadata{
+		CacheFileNameOnDisk: "foobar",
+		BucketName:          "foo",
+		ObjectName:          "baz",
+		Generation:          testGeneration,
+		MetaGeneration:      1,
+	}
+	cacheObject := contentcache.CacheObject{CacheFileObjectMetadata: &objectMetadata}
+	cacheObject.ValidateGeneration(testGeneration)
```
lezh, line 41: The result is not checked

internal/contentcache/contentcache_test.go
```diff
@@ -0,0 +1,65 @@
...
+		MetaGeneration:      1,
+	}
+	metadataFileName, err := contentCache.WriteMetadataCheckpointFile(objectMetadata.ObjectName, &objectMetadata)
+	AssertEq(err, nil)
+	newObjectMetadata := contentcache.CacheFileObjectMetadata{}
+	contents, err := ioutil.ReadFile(metadataFileName)
+	AssertEq(err, nil)
+	err = json.Unmarshal(contents, &newObjectMetadata)
+	AssertEq(err, nil)
+	ExpectEq(true, objectMetadata == newObjectMetadata)
```
lezh, line 63: It should be `ExpectEq(objectMetadata, newObjectMetadata)`. When it fails, the predicate will print out both values for debugging, which can't be done in the current expression.

lezh (changes requested):

PR #636, opened 2022-03-14
gcsfuse content cache thread safety

thepenguinco:
The changes move ensureContent from inode creation to the file read path so files are only cached during reads and not during file inode creation (which happens during directory listings etc.) to improve performance.

In order to support these changes, the content cache now needs to be thread safe as there are no synchronization primitives protecting against concurrent cache access.  (previously we had filesystem and inode locks as cache access occurred at the inode creation step, but such locks do not guard against concurrent cache access during reads)

### DIFF
--- a/internal/contentcache/contentcache.go
+++ b/internal/contentcache/contentcache.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"sync"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/internal/logger"
@@ -41,7 +42,9 @@ type CacheObjectKey struct {
 }
 
 // ContentCache is a directory on local disk to store the object content
+// ContentCache is thread-safe
 type ContentCache struct {
+	mu         sync.Mutex
 	debug      *log.Logger
 	tempDir    string
 	fileMap    map[CacheObjectKey]*CacheObject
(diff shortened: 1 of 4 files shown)


internal/fs/handle/file.go
```diff
@@ -134,6 +134,9 @@ func (fh *FileHandle) checkInvariants() {
 // LOCKS_REQUIRED(fh)
 // LOCKS_REQUIRED(fh.inode)
 func (fh *FileHandle) tryEnsureReader() (err error) {
+	// If content cache enabled, CacheEnsureContent forces the file handler to fall through to the inode
+	// and fh.inode.SourceGenerationIsAuthoritative() will return false
+	fh.inode.CacheEnsureContent(context.Background())
```
lezh, line 139: The context should be passed from the caller of `tryEnsureReader`.

lezh (changes requested):

internal/contentcache/contentcache.go
```diff
@@ -235,6 +247,12 @@ func (c *ContentCache) NewCacheFile(rc io.ReadCloser, f *os.File) (gcsx.TempFile
 	return gcsx.NewCacheFile(rc, f, c.tempDir, c.mtimeClock)
 }
 
+// RecoverCacheFile returns a tempfile wrapper around a prepopulated cache file from disk
```
lezh, line 250: This is not thread-safe?

internal/contentcache/contentcache.go
```diff
@@ -235,6 +247,12 @@ func (c *ContentCache) NewCacheFile(rc io.ReadCloser, f *os.File) (gcsx.TempFile
...
 func (c *ContentCache) RecoverCacheFile(f *os.File) (gcsx.TempFile, error) {
 	return gcsx.RecoverCacheFile(f, c.tempDir, c.mtimeClock)
 }
+
+// Size returns the size of the in memory map of cache files
+func (c *ContentCache) Size() int {
```
lezh, line 256: This is not thread-safe?

lezh (changes requested):

lezh (approved):

PR #651, opened 2022-04-11
Setting MaxResults to 5000 and projection to noAcl in ListObjects call to fetch larger data set

vadlakondaswetha:
Right now we don't set the maxResults param in listObjects api. By default 1000 results are returned. ListObjects api supports fetching upto 5000 results. Hence utilising the maxResults param to fetch larger data set which will improve the directory listing.
https://github.com/GoogleCloudPlatform/gcsfuse/issues/652

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -31,6 +31,11 @@ import (
 	"golang.org/x/net/context"
 )
 
+// ListObjects call supports fetching upto 5000 results via maxResults param in
+// one call. By default 1000 results are returned if value is not set.
+// Defining a constant to set maxResults param.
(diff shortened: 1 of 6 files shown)


internal/fs/inode/dir.go
```diff
@@ -31,6 +31,11 @@ import (
 	"golang.org/x/net/context"
 )
 
+// ListObjects call supports fetching upto 5000 results via maxResults param in
+// one call. By default 1000 results are returned if value is not set.
```
avidullu, line 35: Can you also add that the 5000 is supportable only when the projection is not full

vadlakondaswetha: done.

avidullu (approved):

PR #658, opened 2022-04-22
Changes to fetch statObject from gcs before syncing the file

vadlakondaswetha:
Object metadata is fetched in different flows like lookupInode, ListObject etc. Based on the flow either complete metadata or subset is fetched from gcs and stored in cache.  So cache will not have all the object metadata. Hence before syncing the file which recreates the file in gcs - we need to fetch the all the object properties and set those properties to the new file.

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -517,8 +517,24 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 		return
 	}
 
+	// When listObjects call is made, we fetch data with projection set as noAcl
+	// which means acls and owner properties are not returned. So the f.src object
+	// here will not have acl information even though there are acls present on
+	// the gcsObject.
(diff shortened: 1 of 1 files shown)


internal/fs/inode/file.go
```diff
@@ -517,8 +517,24 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 		return
 	}
 
+	// When listObjects call is made, we fetch data with projection set as noAcl
+	// which means acls and owner properties are not returned. So the f.src object
+	// here will not have acl information even though there are acls present on
+	// the gcsObject.
+	// Hence, we are making an explicit gcs stat call to fetch the latest
```
avidullu, line 524: Can update this comment to call out that Stat call sets the full projection.

vadlakondaswetha: done

avidullu (approved):

PR #681, opened 2022-06-07
Made changes in read.go to integrate Storage Client Library

thesixpathguy:
The readflows will use Go Storage Client Library to interact with GCS backend instead of using the Jacobsa library's HTTP request.


vendor/github.com/jacobsa/gcloud/gcs/read.go
```diff
@@ -41,6 +42,16 @@ func (b *bucket) NewReader(
 	// In Google-internal bug 19718068, it was clarified that the intent is that
 	// each of the bucket and object names are encoded into a single path
 	// segment, as defined by RFC 3986.
+	
+
+	// Switching to Go Storage Client Library 
+	if true{
```
avidullu, line 48: Could you fork this repo and make changes in that so that we can ensure this repo does not break while we test these changes?

vendor/github.com/jacobsa/gcloud/gcs/read.go
```diff
@@ -138,6 +149,32 @@ func (b *bucket) NewReader(
...
+        req *ReadObjectRequest,bucketName string) (rc io.ReadCloser, err error){
+        // initialising the bucket
+        client, err := storage.NewClient(ctx)
+        if err != nil {
+                err = fmt.Errorf("Error in creating the client")
+                return
+        }
+        fmt.Println("New Storage Client Created for: ", req.Name)
+        bkt := client.Bucket(bucketName) //check if the bucket name is valid or not
+        obj:=bkt.Object(req.Name) //check if the object name is valid or not
```
avidullu, line 165: space before and after := .. please run gofmt or follow the style guide of this file/repo

vendor/github.com/jacobsa/gcloud/gcs/read.go
thesixpathguy: Hey Avi! Actually I have made the PR on storagelibrary branch (created by Swetha) which is different than master branch. So I think it should not affect the main code in the master branch.

vendor/github.com/jacobsa/gcloud/gcs/read.go
thesixpathguy: yeah, will do that

thesixpathguy:
Also Today I will write the unit tests for the code and merge them into the same PR.

vendor/github.com/jacobsa/gcloud/gcs/read.go
avidullu: My bad. Missed that.

vendor/github.com/jacobsa/gcloud/gcs/read.go
```diff
@@ -138,6 +153,38 @@ func (b *bucket) NewReader(
...
 }
 
+
+// Custom function made to create a new reader using Storage Client Library.
+func NewReaderSCL(
+        ctx context.Context,
+        req *ReadObjectRequest, bucketName string) (rc io.ReadCloser, err error){
+	// Create a client if there is not one already created.
+        if sclClient == nil {
+		fmt.Println("Client Created")
```
avidullu, line 163: Shouldn't this message be printed after the client is actually created?

vendor/github.com/jacobsa/gcloud/gcs/read.go
```diff
@@ -138,6 +153,38 @@ func (b *bucket) NewReader(
...
+		sclClient, err = storage.NewClient(ctx)
+		if err != nil {
+			err = fmt.Errorf("Error in creating the client: %v", err)
+			return
+		}
+	}
+
+	// Initialising the starting offset and the length to be read by the reader.
+        start := int64((*req.Range).Start)
```
avidullu, line 172: Is it possible that the Range is not being set?

thesixpathguy: I think the range will always be valid.
[https://github.com/GoogleCloudPlatform/gcsfuse/blob/eaf4a4d58ee31765a9fb239b2f672e29f70aa441/internal/gcsx/random_reader.go#L332](url)
Whenever the native NewReader method is being called (as above), the range is always passed, consisting of start and end parameters. Also the start and end variables are in turn formed from offset and length variables which are passed by the OS to the ReadFile method in fs.go. So I think the arguments passed by the OS should be correct, but anyways many checks are performed on offset, length, start and end variables before passing it to the NewReader method. So I think the range will always be valid.
Also I wrote edge cases of various ranges in my unit tests, and they were all passing.

avidullu (approved):

vendor/github.com/jacobsa/gcloud/gcs/read.go
```diff
@@ -138,6 +153,38 @@ func (b *bucket) NewReader(
...
 
+
+// Custom function made to create a new reader using Storage Client Library.
+func NewReaderSCL(
+        ctx context.Context,
+        req *ReadObjectRequest, bucketName string) (rc io.ReadCloser, err error){
+	// Create a client if there is not one already created.
+        if sclClient == nil {
+		fmt.Println("Client Created")
+		sclClient, err = storage.NewClient(ctx)
```
avidullu, line 164: .Close() at some point on this?

thesixpathguy: But if we will close, then we will need to create the client again and in the documentation it was mentioned that we should reuse the client as much as possible.

PR #685, opened 2022-06-15
Set up Go Storage Client as a parameter of bucket struct in jacobsa library.

thesixpathguy:
Moved Go Storage Client declaration from NewReaderSCL method in read.go to newBucket method in bucket.go.
The earlier declaration in NewReaderSCL method was not thread protected i.e. multiple clients can be formed from different threads whereas we want a single client for all threads.
Moved to newBucket method because it is only called once after starting gcsfuse irrespective of different threads. Also if passed as a argument to bucket, the same client can be used by read, write, get methods without creating different clients for different methods.

### DIFF
--- a/vendor/github.com/jacobsa/gcloud/gcs/bucket.go
+++ b/vendor/github.com/jacobsa/gcloud/gcs/bucket.go
@@ -126,7 +126,7 @@ type Bucket interface {
 
 type bucket struct {
 	client         *http.Client
-	sclClient      *storage.Client //Go Storage Library Client.
+	storageClient  *storage.Client //Go Storage Library Client.
 	url            *url.URL
 	userAgent      string
 	name           string
@@ -359,18 +359,18 @@ func newBucket(
 	name string,
 	billingProject string) Bucket {
 
-	// Creating client through GSCL for the sclClient parameter of bucket.
-	var sclClient *storage.Client = nil
-	sclClient, err := storage.NewClient(ctx)
+	// Creating client through Go Storage Client Library for the storageClient parameter of bucket.
+	var storageClient *storage.Client = nil
+	storageClient, err := storage.NewClient(ctx)
 	if err != nil {
-		err = fmt.Errorf("Error in creating the client through GSCL: %v", err)
+		err = fmt.Errorf("Error in creating the client through Go Storage Library: %v", err)
 	} else {
-		fmt.Println("GSCL Client Created")
+		fmt.Println("Go Storage Client Created")
 	}
 
 	return &bucket{
 		client:         client,
-		sclClient:      sclClient,
+		storageClient:  storageClient,
 		url:            url,
 		userAgent:      userAgent,
 		name:           name,
--- a/vendor/github.com/jacobsa/gcloud/gcs/bucket.go
+++ b/vendor/github.com/jacobsa/gcloud/gcs/bucket.go
@@ -364,8 +364,6 @@ func newBucket(
 	storageClient, err := storage.NewClient(ctx)
 	if err != nil {
 		err = fmt.Errorf("Error in creating the client through Go Storage Library: %v", err)
-	} else {
-		fmt.Println("Go Storage Client Created")
 	}
 
 	return &bucket{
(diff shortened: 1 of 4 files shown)


vendor/github.com/jacobsa/gcloud/gcs/bucket.go
```diff
@@ -350,13 +352,25 @@ func (b *bucket) DeleteObject(
...
+
+	// Creating client through GSCL for the sclClient parameter of bucket.
+	var sclClient *storage.Client = nil
+	sclClient, err := storage.NewClient(ctx)
+	if err != nil {
+		err = fmt.Errorf("Error in creating the client through GSCL: %v", err)
+	} else {
+		fmt.Println("GSCL Client Created")
+	}
+
```
vadlakondaswetha, line 370: we should use loggers instead of printing messages to console. Also in this case you can avoid printing a message. Its obvious that if error is not printed the client is created.

vendor/github.com/jacobsa/gcloud/gcs/bucket.go
```diff
@@ -350,13 +352,25 @@ func (b *bucket) DeleteObject(
...
 	url *url.URL,
 	userAgent string,
 	name string,
 	billingProject string) Bucket {
+
+	// Creating client through GSCL for the sclClient parameter of bucket.
+	var sclClient *storage.Client = nil
+	sclClient, err := storage.NewClient(ctx)
+	if err != nil {
+		err = fmt.Errorf("Error in creating the client through GSCL: %v", err)
```
vadlakondaswetha, line 366: what exactly are you doing here. formatting the error message but neither we are returning the error nor printing it, what does it mean?

vendor/github.com/jacobsa/gcloud/gcs/read.go
```diff
@@ -49,12 +45,12 @@ func (b *bucket) NewReader(
...
 	if true {
-                rc, err = NewReaderSCL(ctx, req, b.name)
-                if err != nil {
-                        err = fmt.Errorf("Error in creating client through NewReaderSCL")
-                }
-                return
-        }
+		rc, err = NewReaderSCL(ctx, req, b.name, b.sclClient)
+		if err != nil {
+			err = fmt.Errorf("Error in creating client through NewReaderSCL")
```
vadlakondaswetha, line 50: you can get because of different scenarios from newReaderSCL. You cannot simply say error in creating client for all types of errors. IMO there is no need to format every error we receive, we can return as is

vendor/github.com/jacobsa/gcloud/gcs/bucket.go
thesixpathguy: okay. I'll remove this.

vendor/github.com/jacobsa/gcloud/gcs/bucket.go
thesixpathguy: Yeah I'll change it to print the error.

vendor/github.com/jacobsa/gcloud/gcs/read.go
thesixpathguy: sure, will remove it.

vendor/github.com/jacobsa/gcloud/gcs/read.go
```diff
@@ -153,36 +149,31 @@ func (b *bucket) NewReader(
...
-		if err != nil {
-			err = fmt.Errorf("Error in creating the client: %v", err)
-			return
-		}
-		fmt.Println("Client Created")
+	ctx context.Context,
+	req *ReadObjectRequest, bucketName string, sclClient *storage.Client) (rc io.ReadCloser, err error) {
+	// If client is "nil", it means that there was some problem in initializing client in newBucket function of bucket.go file.
+	if sclClient == nil {
+		err = fmt.Errorf("Error in creating client through GSCL: %v", err)
```
vadlakondaswetha, line 158: same here also. err is the return variable here and by the time the code reaches here its not even initialized, why are you using err in fmt.Errorf()

thesixpathguy: Since err is a named return parameter so I think we don't have to initialize it beforehand. It will already be initialized once the function is called.

vadlakondaswetha (approved):

PR #687, opened 2022-06-17
build_gcsfuse: apply buildArgs to all targets

gromgit:
I can't think of a build flag that could only be applied to `gcsfuse`, and I currently need to apply `-buildvcs=false` to both binaries.


talsag-dev (approved):

ashmeenkaur (commented):
Please fix formatting (lint checks)

PR #690, opened 2022-06-22
Adding code to extract fio metrics

adata111:


vadlakondaswetha (approved):
This code is reviewed on google3

PR #695, opened 2022-06-23
Correcting job time calculation in fio metrics

adata111:


perfmetrics/scripts/fio/fio_metrics.py
```diff
@@ -63,13 +63,15 @@
...
-def _convert_value(value, conversion_dict):
+def _convert_value(value, conversion_dict, default_unit=''):
   """Converts data strings to a particular unit based on conversion_dict.
 
   Args:
-    value: String, contains data value+unit
+    value: String, contains data value[+unit]
     conversion_dict: Dictionary containing units and their respective
       multiplication factor
+    default_unit: String, specifies the default unit, used if no unit is present
```
vadlakondaswetha, line 73: can you elaborate the comment on why the unit is not present in value, is it because fio is giving units for some or not for others?

vadlakondaswetha (approved):

perfmetrics/scripts/fio/fio_metrics.py
adata111: Updated the comment

PR #702, opened 2022-07-04
Bash Script to setup machine before running ML model automation Script.

DrashtiKoladiya:


DrashtiKoladiya:
Already reviewd in google3, no changes after that.

PR #720, opened 2022-07-13
Making final changes to fio_metrics

adata111:
fio_metrics.py
- handling startdelay from start of job file execution issue
- passing worksheet name to _add_to_google_sheet function

fio_metrics_test.py
- Updating tests for the startdelay change
- Updating tests for gsheet.py changes (deleting entries in sheet before writing)

fetch_metrics.py
- passing worksheet name to fio and vm metrics extraction methods

Also added new job files and removed unused job files


perfmetrics/scripts/fio/fio_metrics_test.py
```diff
@@ -18,8 +18,7 @@
 MULTIPLE_JOBS_JOB_OPTIONS_FILE = 'multiple_jobs_job_options.json'
 
 SPREADSHEET_ID = '1kvHv1OBCzr9GnFxRu9RTJC7jjQjc9M4rAiDnhyak2Sg'
```
avidullu, line 20: As mentioned in a separate cl, should we make this a flag which can be overridden? If its hairy then having as is is also fine by me.

avidullu (commented):
Mostly nits.

perfmetrics/scripts/fio/fio_metrics_test.py
avidullu: Got the response.

PR #727, opened 2022-07-27
Go Client Flag added. Added parameters to improve performance.

thesixpathguy:


main.go
```diff
@@ -93,13 +93,14 @@ func getConn(flags *flagStorage) (c *gcsx.Connection, err error) {
...
-		MaxConnsPerHost:     flags.MaxConnsPerHost,
-		MaxIdleConnsPerHost: flags.MaxIdleConnsPerHost,
+		Url:                        flags.Endpoint,
+		TokenSource:                tokenSrc,
+		UserAgent:                  fmt.Sprintf("gcsfuse/%s %s", getVersion(), flags.AppName),
+		MaxBackoffSleep:            flags.MaxRetrySleep,
+		DisableHTTP2:               flags.DisableHTTP2,
+		MaxConnsPerHost:            flags.MaxConnsPerHost,
+		MaxIdleConnsPerHost:        flags.MaxIdleConnsPerHost,
+		EnableStorageClientLibrary: flags.EnableStorageClientLibrary,
```
avidullu, line 103: Is it possible for you to forwat the code using golangci-lint  ?

vendor/cloud.google.com/go/storage/bucket.go
```diff
@@ -1977,7 +1977,7 @@ func (it *ObjectIterator) fetch(pageSize int, pageToken string) (string, error)
 	if projection == ProjectionDefault {
 		projection = ProjectionFull
 	}
-	req.Projection(projection.String())
+	req.Projection("noAcl")
```
avidullu, line 1980: For this I think we should debug as to why we are not getting the right value from the projection.string when we set it to noAcl in gcsfuse code.

vendor/cloud.google.com/go/storage/bucket.go
```diff
@@ -1991,9 +1991,7 @@ func (it *ObjectIterator) fetch(pageSize int, pageToken string) (string, error)
 	if it.bucket.userProject != "" {
 		req.UserProject(it.bucket.userProject)
 	}
-	if pageSize > 0 {
-		req.MaxResults(int64(pageSize))
-	}
+	req.MaxResults(5000)
```
avidullu, line 1994: Please add a comment here that this is done because the default is 1000 and we need to find a way to make this configurable

vendor/github.com/jacobsa/gcloud/gcs/create_object.go
```diff
@@ -265,6 +265,7 @@ func CreateObjectSCL(
 	// Creating a NewWriter with requested attributes, using Go Storage Client.
 	// Chuck size for resumable upload is deafult i.e. 16MB.
 	wc := obj.NewWriter(ctx)
+	wc.ChunkSize = 0 // This will enable one shot upload and thus increase performance.
```
avidullu, line 268: What is the difference you get with this?

thesixpathguy:
Checked with FIO test and the performance of both clients is as intended.

main.go
thesixpathguy: Hey I used golangci-lint but it's not showing any suggestion to format this code piece.

vendor/cloud.google.com/go/storage/bucket.go
thesixpathguy: done.

vendor/cloud.google.com/go/storage/bucket.go
thesixpathguy: done.

vendor/cloud.google.com/go/storage/bucket.go
thesixpathguy: Checked by performing manual listing operation and it's working correctly now.

vendor/cloud.google.com/go/storage/bucket.go
avidullu: Where is the comment?

vendor/cloud.google.com/go/storage/bucket.go
avidullu: Please delete the lines 1976-1979 (which use the projection variable) and add a comment that this is done to ask for higher number of results.

vendor/cloud.google.com/go/storage/bucket.go
thesixpathguy: No the files will not get updated here.
please check: [link](https://github.com/GoogleCloudPlatform/gcsfuse/pull/727/files)

PR #731, opened 2022-07-27
Add no proxy

rpjeff:
Adds pass through of no_proxy environment variable to compliment existing pass through of https_proxy and http_proxy environment variables.


main.go
```diff
@@ -281,6 +281,14 @@ func runCLIApp(c *cli.Context) (err error) {
 				"Added environment http_proxy: %s\n",
 				p)
 		}
+		// Pass through the no_proxy enviroment variable.
```
avidullu, line 284: Similar in spirit to the https_proxy above, please call out in the comments as to when and why one would need this.

PR #742, opened 2022-08-04
Made changes to export metrics for gcs read calls along with type

vadlakondaswetha:
1. Added two new metrics - gcs/read_count to capture number of read calls along with type, gcs/downloaded_bytes_count to capture amount of data downloading from gcs.
2. gcsfuse_read_bytes is same as gcs/read_bytes_count. Hence removed gcsfuse_read_bytes

### DIFF
--- a/internal/gcsx/random_reader.go
+++ b/internal/gcsx/random_reader.go
@@ -61,19 +61,19 @@ const random = "Random"
 // Initialize the metrics.
 func init() {
 	if err := view.Register(
-		      &view.View{
-									Name:        "gcsfuse_read_bytes",
-									Measure:     readBytes,
-									Description: "The number of bytes read from GCS",
-									Aggregation: view.Sum(),
-					},
-					&view.View{
-									Name:        "gcs_read_count",
-									Measure:     gcsReadCount,
-									Description: "Specifies the number of gcs reads made along with type- Sequential/Random",
-									Aggregation: view.Sum(),
-									TagKeys:     []tag.Key{tags.ReadType},
-					},
+		&view.View{
+			Name:        "gcsfuse_read_bytes",
+			Measure:     readBytes,
+			Description: "The number of bytes read from GCS",
+			Aggregation: view.Sum(),
+		},
+		&view.View{
+			Name:        "gcs_read_count",
+			Measure:     gcsReadCount,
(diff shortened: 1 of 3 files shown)


internal/gcsx/random_reader.go
```diff
@@ -47,15 +54,27 @@ const maxReadSize = 8 * MB
...
 // Initialize the metrics.
 func init() {
-	v := &view.View{
-		Name:        "gcsfuse_read_bytes",
-		Measure:     readBytes,
-		Description: "The number of bytes read from GCS",
-		Aggregation: view.Sum(),
-	}
-	if err := view.Register(v); err != nil {
+	if err := view.Register(
```
avidullu, line 63: nit: formatting looks off.

internal/gcsx/random_reader.go
```diff
@@ -350,5 +370,25 @@ func (rr *randomReader) startRead(
 	rr.start = start
 	rr.limit = end
 
+	captureMetrics(ctx, rr.seeks)
```
avidullu, line 373: why pass the seeks here? Can we make a readType variable in this function, populate it based on the conditions in this function and then pass it to the captureMetrics?

avidullu, line 373: I think we can also record the size (end - start) of the read? Or is that been done elsewhere?

internal/gcsx/random_reader.go
vadlakondaswetha: It looks proper in the intellij: https://screenshot.googleplex.com/FAhYcbCsieQ9t5z.png

internal/gcsx/random_reader.go
vadlakondaswetha: We have gcsfuse_read_bytes metrics which captures the read bytes. Not based on data fetched from gcs but data read by user.  do you want to capture the amount of data fetched from gcs?

avidullu: right, that's the intention.

internal/gcsx/random_reader.go
avidullu: I see it as the attached image, hence the comment.

codeformat
t

avidullu (approved):

PR #746, opened 2022-08-05
Fixing Hang Issue in KeyFile Integration Test

raj-prince:
**Issue:** By default when we pass wrong key-file it retries several time and returns the response after ~200 secs. We have two test-cases in that particular test (KeyFile), so it returns the response after ~400 secs, which feels like the hang in the integration test.

**Fix:**    By setting the --max-retry-sleep to 0, we only try to connect to the server once which also removes the concept of wait time and we are able to complete the test in 4-5 secs.

### DIFF
--- a/tools/integration_tests/mounting/gcsfuse_test.go
+++ b/tools/integration_tests/mounting/gcsfuse_test.go
@@ -185,7 +185,7 @@ func (t *GcsfuseTest) KeyFile() {
 	const nonexistent = "/tmp/foobarbazdoesntexist"
 
 	// Specify a non-existent key file in two different ways.
-	// We pass --max-retry-sleep 0, just to limit number of retry to 1 with no wait time
+	// We pass --max-retry-sleep 0, just to limit the number of retry to 0 with no wait time
 	testCases := []struct {
 		extraArgs []string
 		env       []string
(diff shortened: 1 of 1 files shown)


tools/integration_tests/mounting/gcsfuse_test.go
```diff
@@ -21,7 +21,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-
```
avidullu, line 24: Is this formatting done by some IDE?

tools/integration_tests/mounting/gcsfuse_test.go
```diff
@@ -194,12 +191,13 @@ func (t *GcsfuseTest) KeyFile() {
 	}{
 		// Via flag
 		0: {
-			extraArgs: []string{fmt.Sprintf("--key-file=%s", nonexistent)},
+			extraArgs: []string{fmt.Sprintf("--key-file=%s", nonexistent), "--max-retry-sleep=0"},
```
avidullu, line 194: Add a comment as to why this max-retry-sleep flag is needed.

tools/integration_tests/mounting/gcsfuse_test.go
raj-prince: yes.

tools/integration_tests/mounting/gcsfuse_test.go
raj-prince: done.

avidullu: I don't see the comment. Do you need to "commit something" ?

raj-prince: Ohh, I updated the comment in description part, now it's done.

avidullu (approved):

PR #749, opened 2022-08-06
ReadWrite integration tests fixes

vadlakondaswetha:
1. Using fusermount instead of umount since umount requires sudo permissions.
2. Clearing kernel page cache after write to ensure reads are made from gcsfuse
3. Deleting files from mount directory to ensure gcsbucket is not overloaded.

### DIFF
--- a/tools/integration_tests/readwrite/readwrite_test.go
+++ b/tools/integration_tests/readwrite/readwrite_test.go
@@ -96,6 +96,13 @@ func unMount() error {
 }
 
 func clearKernelCache() error {
+	if _, err := os.Stat("/proc/sys/vm/drop_caches"); err != nil {
+		log.Printf("Kernel cache file not found: %v", err)
+		// No need to stop the test execution if cache file is not found. Further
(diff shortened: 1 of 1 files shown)


tools/integration_tests/readwrite/readwrite_test.go
```diff
@@ -83,14 +83,23 @@ func mountGcsfuse() error {
...
 	if _, err := cmd.CombinedOutput(); err != nil {
-		return fmt.Errorf("umount error: %w", err)
+		return fmt.Errorf("fusermount error: %w", err)
+	}
+	return nil
+}
+
+func clearKernelCache() error {
+	// sudo permission is required to clear kernel page cache.
+	cmd := exec.Command("sudo", "sh", "-c", "echo 3 > /proc/sys/vm/drop_caches")
```
avidullu, line 100: This may not be present on all the platforms. Should we do a check if this exists and only write if it does?

avidullu (approved):

avidullu (approved):

PR #751, opened 2022-08-08
Flag changes to have ErrorLogger as optional

avidullu:
PR adds the --log-fuse-errors as a default true flag which can be optionally set to false in case a user wants to ignore the error logs from fuse.  Attached a screenshot which confirms that the fuse: messages disappear only when the flag is explicitly set to false.

image

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -255,16 +255,16 @@ func newApp() (app *cli.App) {
 				Usage: "The format of the log file: 'text' or 'json'.",
 			},
 
+			/////////////////////////
+			// Debugging
+			/////////////////////////
+
 			cli.BoolTFlag{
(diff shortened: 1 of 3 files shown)


flags.go
```diff
@@ -255,6 +255,13 @@ func newApp() (app *cli.App) {
 				Usage: "The format of the log file: 'text' or 'json'.",
 			},
 
+			cli.BoolTFlag{
+				Name: "log-fuse-errors",
```
vadlakondaswetha, line 259: should we rename this to debug_fuse_errors to be consistent with debug_fuse name?

avidullu: Done

vadlakondaswetha (approved):

PR #752, opened 2022-08-08
Updated README.md for Go Storage Client

thesixpathguy:
Added info about Go Storage Client.


README.md
```diff
@@ -191,3 +191,53 @@ Versioning][semver]. Note that the current major version is `0`, which means
 that we reserve the right to make backwards-incompatible changes.
 
 [semver]: http://semver.org/
+
+# Switching to Go Storage Client
```
avidullu, line 195: I think you should add a couple of lines here (maybe a sub heading also) to describe what is different about this branch i.e. the fact that this branch has Go Storage Client integrated and can be activated via flags. After that you can describe how to use GCSFuse with Go Storage Client and then can describe about performance

avidullu (approved):

PR #773, opened 2022-08-26
Adding API to Create Bucket Handle

raj-prince:
>> Added api to create bucket handle.
>> Added a new dependency to the project to create fake go storage client.


internal/storage/storage_handle.go
```diff
@@ -77,6 +81,19 @@ func NewStorageHandle(ctx context.Context,
...
+	sh = &StorageClient{client: sc}
+	return
+}
+
+func (sh *StorageClient) BucketHandle(bucketName string) (bh *bucketHandle,
+	err error) {
+	storageBucketHandle := sh.client.Bucket(bucketName)
+	attrs, err := storageBucketHandle.Attrs(context.Background())
+
+	if err != nil || attrs == nil {
```
vadlakondaswetha, line 93: what does attrs == nil mean? should we set something to error otherwise, callers will not know what happened

internal/storage/storage_handle.go
```diff
@@ -13,7 +13,11 @@ import (
 	"google.golang.org/api/option"
 )
 
-type storageClient struct {
+type StorageHandle interface {
+	BucketHandle(bucketName string) (bh *bucketHandle, err error)
+}
+
+type StorageClient struct {
```
vadlakondaswetha, line 20: why is this made public?

internal/storage/storage_handle_test.go
```diff
@@ -5,9 +5,63 @@ import (
 	"testing"
 	"time"
 
+	"github.com/fsouza/fake-gcs-server/fakestorage"
 	"golang.org/x/oauth2"
 )
 
+const correctBucketName string = "will-be-present-in-fake-server"
```
vadlakondaswetha, line 12: instead of correct/wrong it would be good to rename to valid/invalid

internal/storage/storage_handle_test.go
```diff
@@ -5,9 +5,63 @@ import (
...
+	if err != nil {
+		t.Fatalf("Server creation failed")
+	}
+	defer server.Stop()
+
+	fakeClient := server.Client()
+	fakeStorageClient := &StorageClient{client: fakeClient}
+	bucketHandle, err := fakeStorageClient.BucketHandle(correctBucketName)
+
+	if err != nil {
```
vadlakondaswetha, line 41: given that we are going to use 3rd party libraries for unit tests, go/gcs-clients-coding-guidelines on testing is not valid. Please use ooglematchers/ooglemock like in other test files to do asserting.

Sorry for back and forth.

vadlakondaswetha (changes requested):

internal/storage/storage_handle.go
raj-prince: Got a warning that return type of a Exported (public) method should also be public. I can revert if it doesn't make sense.

vadlakondaswetha: you can return an interface type here instead of structType

vadlakondaswetha: Like this: https://github.com/jacobsa/gcloud/blob/8d1a66f33ec963197b578d3bd491df1f6deea91b/gcs/bucket.go#L352

internal/storage/storage_handle.go
raj-prince: We will only check for non-nil err, since attrs == nil signifies some error occurred. Err will be set automatically.

internal/storage/storage_handle_test.go
raj-prince: But here we are using feature specific 3rd party libraries, not like ooglematchers/ooglemock which was specially for testing (feature agnostic).

vadlakondaswetha: Right. the guideline going forward is " We are okay to use 3rd party libraries for unit testing". And since we already use ooglematchers/ooglemock in gcsfuse, lets use it here also.

internal/storage/storage_handle.go
vadlakondaswetha: are you sure that gcs will set the error incase attrs==nil? Then why do you have a check on attrs==nil?

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/googlecloudplatform/gcsfuse
 go 1.18
 
 require (
-	cloud.google.com/go/storage v1.22.1
```
vadlakondaswetha, line 6: please do go.mod changes in seperate cls.

internal/storage/storage_handle_test.go
raj-prince: Sure.

go.mod
raj-prince: Cool. Then this CL will contain only the changes to integrate the dependency. Will create a new cl for the feature task and unit test.

internal/storage/storage_handle.go
raj-prince: I tried for invalid bucket, it returns attrs as nil and error as "bucket doesn't exist".

raj-prince:
Closing this PR. Created a new PR for this change: https://github.com/GoogleCloudPlatform/gcsfuse/pull/790

PR #776, opened 2022-08-29
Test for write at start of the file

abhinav2603:
Integration test for testing gcsfuse if write made at the start of file is handled correctly

### DIFF
--- a/tools/integration_tests/implicitdir/write_test.go
+++ b/tools/integration_tests/implicitdir/write_test.go
@@ -30,7 +30,6 @@ func TestWriteAtEndOfFile(t *testing.T) {
 	if _, err = f.WriteString("line 3\n"); err != nil {
 		t.Errorf("AppendString: %v", err)
 	}
-
 	f.Close()
 
 	compareFileContents(t, fileName, "line 1\nline 2\nline 3\n")
@@ -46,7 +45,6 @@ func TestWriteAtStartOfFile(t *testing.T) {
 	if _, err = f.WriteAt([]byte("line 4\n"), 0); err != nil {
 		t.Errorf("WriteString-Start: %v", err)
 	}
-
 	f.Close()
 
 	compareFileContents(t, fileName, "line 4\nline 2\n")
--- a/tools/integration_tests/implicitdir/write_test.go
+++ b/tools/integration_tests/implicitdir/write_test.go
@@ -37,7 +37,7 @@ func TestWriteAtEndOfFile(t *testing.T) {
 }
 
 func TestWriteAtStartOfFile(t *testing.T){
-	createTempFile()
+	fileName := createTempFile()
 	f, err := os.OpenFile(fileName, os.O_WRONLY, 0600)
 	if err != nil {
 	    t.Errorf("Open file for write at start: %v", err)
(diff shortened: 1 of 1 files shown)


tools/integration_tests/implicitdir/implicitdir_test.go
```diff
@@ -32,10 +32,13 @@ import (
...
-	binFile string
-	logFile string
-	mntDir  string
+	testDir  string
+	binFile  string
+	logFile  string
+	mntDir   string
+	tmpDir   string
+	fileName string
+	err      error
```
vadlakondaswetha, line 41: fileName and err shouldn't be global variables. Make the fileName as return parameter for the function call. 
For the err, i can see it not be need to sent back to the caller. So you can make it a local variable inside the method.

vadlakondaswetha (commented):
Update the cl description to mention that you added two test - write at start of file and writeAtEndOfFile

tools/integration_tests/implicitdir/write_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Provides integration tests for write flows with implicit_dir flag set.
+package implicitdir_test
+
+import (
+	"os"
+	"testing"
+)
+
+func TestAppendToEndOfFile(t *testing.T) {
```
vadlakondaswetha, line 23: please make the test method names consistent. Either call it append or write in both the cases.

tools/integration_tests/implicitdir/implicitdir_test.go
```diff
@@ -111,6 +114,40 @@ func clearKernelCache() error {
...
+}
+
+func logAndExit(s string) {
+	log.Printf(s)
+	os.Exit(1)
+}
+
+func createTempFile() {
+	// A temporary file is created and some lines are added
+	// to it for testing purposes
```
vadlakondaswetha, line 143: nit: Comment is like a statement. Please start with capital letter and end with full stop.

abhinav2603: Done

tools/integration_tests/implicitdir/write_test.go
abhinav2603: Done

tools/integration_tests/implicitdir/implicitdir_test.go
abhinav2603: Done

tools/integration_tests/implicitdir/write_test.go
```diff
@@ -35,3 +35,19 @@ func TestWriteAtEndOfFile(t *testing.T) {
...
+	fileName := createTempFile()
+	f, err := os.OpenFile(fileName, os.O_WRONLY, 0600)
+	if err != nil {
+		t.Errorf("Open file for write at start: %v", err)
+	}
+
+	if _, err = f.WriteAt([]byte("line 4\n"), 0); err != nil {
+		t.Errorf("WriteString-Start: %v", err)
+	}
+
```
vadlakondaswetha, line 49: in test cases, we should have the formatting like this
test initialisation
<new line>
test invocation
<new line>
assert statements.

f.close is part of test invocation only. so remove new line before f.close.

vadlakondaswetha (approved):

PR #790, opened 2022-09-07
Adding an API to Create a Bucket Handel from StorageHandle

raj-prince:


internal/storage/storage_handle.go
```diff
@@ -77,6 +94,18 @@ func NewStorageHandle(ctx context.Context,
...
+	client := &storageClient{client: sc}
+	return client, err
+}
+
+func (sh *storageClient) BucketHandle(bucketName string) (bh *bucketHandle,
+	err error) {
+	storageBucketHandle := sh.client.Bucket(bucketName)
+	_, err = storageBucketHandle.Attrs(context.Background())
+	if err != nil {
+		return
```
vadlakondaswetha, line 106: do you want the format the error to mention that specified bucket is not found

internal/storage/storage_handle.go
```diff
@@ -77,6 +94,18 @@ func NewStorageHandle(ctx context.Context,
 		}),
 		storage.WithPolicy(storage.RetryAlways))
 
-	sh = &storageClient{client: sc}
+	client := &storageClient{client: sc}
+	return client, err
```
vadlakondaswetha, line 98: simple return should be sufficient. no need to mention client, err

internal/storage/storage_handle.go
```diff
@@ -77,6 +94,18 @@ func NewStorageHandle(ctx context.Context,
...
+}
+
+func (sh *storageClient) BucketHandle(bucketName string) (bh *bucketHandle,
+	err error) {
```
vadlakondaswetha, line 102: dont break the function definitions inbetween. Put the entire definition on one line. https://github.com/golang/go/wiki/CodeReviewComments#line-length

internal/storage/storage_handle_test.go
```diff
@@ -20,36 +61,54 @@ func getDefaultStorageClientConfig() (clientConfig storageClientConfig) {
 	}
 }
 
-func invokeAndVerifyStorageHandle(t *testing.T, sc storageClientConfig) {
-	handleCreated, err := NewStorageHandle(context.Background(), sc)
+func (t *StorageHandleTest) TestBucketHandleWhenBucketExists() {
+	server, err := createFakeServer()
```
vadlakondaswetha, line 65: should we create the server once for all tests and add a client variable to StorageHandleTest struct.
Any specific reason we need one server for every test?

internal/storage/storage_handle_test.go
```diff
@@ -20,36 +61,54 @@ func getDefaultStorageClientConfig() (clientConfig storageClientConfig) {
...
+	fakeClient := server.Client()
+	fakeStorageClient := &storageClient{client: fakeClient}
+
+	bucketHandle, err := fakeStorageClient.BucketHandle(invalidBucketName)
+
+	AssertNe(nil, err)
+	AssertEq(nil, bucketHandle)
+}
+
+func (t *StorageHandleTest) invokeAndVerifyStorageHandle(sc storageClientConfig) {
```
vadlakondaswetha, line 90: put the private methods before all tests or after all tests.

vadlakondaswetha (changes requested):

internal/storage/storage_handle.go
raj-prince: Removed the client variable.

internal/storage/storage_handle.go
raj-prince: Right now we get error message as "storage: bucket doesn't exist".

internal/storage/storage_handle_test.go
```diff
@@ -20,36 +39,79 @@ func getDefaultStorageClientConfig() (clientConfig storageClientConfig) {
...
-	}
-	if handleCreated == nil {
-		t.Fatalf("Storage handle is null")
-	}
-	if nil == handleCreated.client {
-		t.Fatalf("Storage client handle is null")
-	}
+func (t *StorageHandleTest) TestBucketHandleWhenBucketExists() {
+	fakeClient := t.fakeStorageServer.Client()
+	fakeStorageClient := &storageClient{client: fakeClient}
```
vadlakondaswetha, line 81: nit: dont call it fakeStorageClient, rename it to storageClient.
It gives the impression that we are creating a fake instance and invoking test method on it. but we are creating a storageClient instance which has fakeClient.

vadlakondaswetha (approved):

vadlakondaswetha (approved):

PR #792, opened 2022-09-07
Updating version jacobsa gcloud

Tulsishah:
Updating version of jacobsa gcloud


Tulsishah:
As it was my forked project, seems like it does not fetch original changes. I created new branch and created PR again which does not have any conflict now.

PR #794, opened 2022-09-07
Updating version of Jacobsa gcloud

Tulsishah:
Updating version of jacobsa gcloud


tools/integration_tests/mounting/test.txt
```diff
@@ -0,0 +1,24 @@
+{"name":"root","levelname":"INFO","severity":"INFO","message":"Start gcsfuse/fake_version (Go version go1.19-pre4 cl/455575533 +12f49fe0ed) for app \"\" using mount point: /tmp/gcsfuse_test1743820428\n","timestampSeconds":1662540221,"timestampNanos":695732107}
```
avidullu, line 1: I think this file should be deleted.
@raj-prince I think these files are created as a side effect of running go test all in the gcsfuse repo. Can we make sure it gets deleted after the tests are completed? Or make this in a temporary directory?

vendor/github.com/fsouza/fake-gcs-server/internal/backend/metadata_file.go
```diff
@@ -1,30 +0,0 @@
-// Copyright 2022 Francisco Souza. All rights reserved.
```
avidullu, line 1: These files should not be deleted i,ho
@raj-prince just added these today.

avidullu (commented):
I am not sure why we're having so many files in this PR. This PR should only have files in the vendor/github.com/jacobsa/gcloud directory which you updated in the other PR.

PR #803, opened 2022-09-14
Fixing Integration Test Related to Relative Log/Key File Path

raj-prince:
>> Issue: whenever we run the test related to relative log/key file path. One new test.txt file gets created in the working directory.

>> Fix: We were passing different working directory while creating the absolute path for given relative path. Now, I made changes to get the current working directory from os.Getwd()

Question: Why the behavior is different for desktop machine and GCP machine?
>> For desktop, finally mounting happens with **fusermount3** package binary and for GCP vm, it is done using **fusermount**.   Both are different package binary, if we want the same behavior, we need to install **fusermount3** on the GCP_vm. So, fusermount3 looks updated version and it mounts the non-empty directory too, which is not the case for fusermout.

### DIFF
--- a/tools/integration_tests/mounting/gcsfuse_test.go
+++ b/tools/integration_tests/mounting/gcsfuse_test.go
@@ -22,6 +22,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+
 	//"runtime"
 	"syscall"
 	"testing"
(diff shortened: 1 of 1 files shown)


avidullu:
Would this PR also clean up the test.txt file which is created during the test execution?

raj-prince:
> Would this PR also clean up the test.txt file which is created during the test execution?

> Yes.

avidullu (approved):

vadlakondaswetha (approved):

PR #809, opened 2022-09-22
Support to Create Bucket Object Using Go Client Library

raj-prince:


internal/storage/bucket_handle.go
```diff
@@ -96,3 +96,39 @@ func (b *bucketHandle) StatObject(ctx context.Context, req *gcs.StatObjectReques
...
+			obj = obj.If(storage.Conditions{GenerationMatch: *req.GenerationPrecondition, MetagenerationMatch: *req.MetaGenerationPrecondition})
+		} else {
+			obj = obj.If(storage.Conditions{GenerationMatch: *req.GenerationPrecondition})
+		}
+	}
+
+	// Creating a NewWriter with requested attributes, using Go Storage Client.
+	// Chuck size for resumable upload is default i.e. 16MB.
+	wc := obj.NewWriter(ctx)
+	wc.ChunkSize = 0 // This enables one shot upload.
```
raj-prince, line 114: This change disables the chunking mechanism, which improves the performance and the memory usage during the upload. But it prevents the writer in retrying in case of a transient error.

@vadlakondaswetha Please let me know, what should we do here?
There are three options:-
(a) We can make it configurable - need to pass an argument for this.
(b) We can with the current change - no retry but improves memory and performance.
(c) The default chunk size - 16 MB.

internal/storage/bucket_handle.go
```diff
@@ -96,3 +96,40 @@ func (b *bucketHandle) StatObject(ctx context.Context, req *gcs.StatObjectReques
 
 	return
 }
+
+func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectRequest) (o *gcs.Object, err error) {
+	obj := bh.bucket.Object(req.Name)
+
+	if req.GenerationPrecondition != nil && *req.GenerationPrecondition != 0 {
```
vadlakondaswetha, line 103: why not two conditions seperately like it is done for other methods. It avoid nesting of code and improves readability.

internal/storage/bucket_handle.go
vadlakondaswetha: Configurable shouldn't be required for now.
how it improves the memory usage in case of single shot upload?
I would suggest create a bug to load test the write flow to check what should be the chunk size

internal/storage/bucket_handle.go
```diff
@@ -96,3 +96,40 @@ func (b *bucketHandle) StatObject(ctx context.Context, req *gcs.StatObjectReques
...
+
+	// Copy the contents to the writer.
+	if _, err = io.Copy(wc, req.Contents); err != nil {
+		err = fmt.Errorf("error in io.Copy: %w", err)
+		return
+	}
+
+	// We can't use defer to close the writer, because we need to close the
+	// writer successfully before calling Attrs() method of writer.
+	if err = wc.Close(); err != nil {
```
vadlakondaswetha, line 125: just curious, when exactly does write to gcs happens, when do you wc.close or io.copy?

internal/storage/bucket_handle.go
```diff
@@ -96,3 +96,40 @@ func (b *bucketHandle) StatObject(ctx context.Context, req *gcs.StatObjectReques
...
+		err = fmt.Errorf("error in closing writer: %v", err)
+		return
+	}
+
+	attrs := wc.Attrs() // Retrieving the attributes of the created object.
+
```
vadlakondaswetha, line 131: nit: remove new line here.

internal/storage/bucket_handle.go
```diff
@@ -96,3 +96,40 @@ func (b *bucketHandle) StatObject(ctx context.Context, req *gcs.StatObjectReques
 
 	return
 }
+
+func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectRequest) (o *gcs.Object, err error) {
+	obj := bh.bucket.Object(req.Name)
+
+	if req.GenerationPrecondition != nil && *req.GenerationPrecondition != 0 {
+		if req.MetaGenerationPrecondition != nil && *req.MetaGenerationPrecondition != 0 {
```
vadlakondaswetha, line 104: can you add a comment what does generation and metageneration preconditions do in case of createObject

vadlakondaswetha (changes requested):

internal/storage/bucket_handle.go
raj-prince: When chunk size is zero, it doesn't keep buffer in order to retry the failed request.

raj-prince: If ChunkSize is set to zero, chunking will be disabled and the object will be uploaded in a single request without the use of a buffer.

raj-prince: Sure. I'll create a bug to tune the value of chunk-size according to the performance trade-off.

internal/storage/bucket_handle.go
raj-prince: As I can see in other methods present in this file, request object contains one Generation (not GenerationPrecondition) and one MetagenerationPrecondition. This is the first method, in which request contains both precondition. Also, we can assign one storage.Conditions to obj (using obj.If). If we will write separate if statement for both condition, there might be chance that the later condition assignment will override the previous assignment.

internal/storage/bucket_handle.go
raj-prince: io.copy writes the data to gcs (internally it open writer method of httpStorageClient, which creates pipe to write it to the gcs), and close method completes the write operation and flush any remaining buffered data in writer-pipe.

internal/storage/bucket_handle.go
vadlakondaswetha: 1. what is the difference between generation and generation precondition?
2. lets not have might be answers. Please add two conditions seperately and check what is happening. I believe you can check the query in debug mode to see what is generated.

internal/storage/bucket_handle.go
raj-prince: Cool.

internal/storage/bucket_handle.go
raj-prince: Generation is used to select the object on which we want to perform the operation.
While the preconditions is used to make the operation conditional. like, IfGenerationMatch, IfGenerationNotMatch, IfMetaGenerationMatch, IfMetaGenerationNotMatch.

For details: https://cloud.google.com/storage/docs/json_api/v1/objects/patch

internal/storage/bucket_handle_test.go
```diff
@@ -185,3 +186,33 @@ func (t *BucketHandleTest) TestStatObjectMethodWithMissingObject() {
...
+	obj, err := t.bucketHandle.CreateObject(context.Background(),
+		&gcs.CreateObjectRequest{
+			Name:                   "test_object",
+			Contents:               strings.NewReader(content),
+			CRC32C:                 &crc32,
+			GenerationPrecondition: &generation,
+		})
+
+	AssertEq(nil, obj)
+	AssertNe(nil, err)
```
vadlakondaswetha, line 217: please assert the type of error

vadlakondaswetha (approved):

internal/storage/bucket_handle.go
raj-prince: Create a small taks for this: https://b.corp.google.com/issues/248985083

PR #820, opened 2022-09-28
Adding experimental flag and test case

Tulsishah:


flags.go
```diff
@@ -357,6 +367,9 @@ type flagStorage struct {
 	DebugHTTP       bool
 	DebugInvariants bool
 	DebugMutex      bool
+
+	// client
+	ExperimentalEnableStorageClientLibrary bool
```
vadlakondaswetha, line 372: this should not be called experimental. The flag which is exposed to the customer will be experimental, internally we will call it EnableStorageClientLibrary. Once the flag is productionized, only flag will be updated, no code changes should be done

Eg: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/flags.go#L213

vadlakondaswetha (changes requested):

flags.go
Tulsishah: fixed

vadlakondaswetha (approved):

PR #822, opened 2022-09-28
Updating Storagehandle file for making StorageClientConfig public

Tulsishah:


internal/storage/storage_handle.go
```diff
@@ -35,11 +35,11 @@ type storageClient struct {
 	client *storage.Client
 }
 
-type storageClientConfig struct {
+type StorageClientConfig struct {
 	disableHTTP2        bool
 	maxConnsPerHost     int
 	maxIdleConnsPerHost int
-	tokenSrc            oauth2.TokenSource
```
vadlakondaswetha, line 42: why is this made public?

Tulsishah: because while calling NewStorageHandle we need to set particular token in storageclientConfig, as before it was private was not able to set it up.

vadlakondaswetha: others fields also need to be set right from outside. Why are they not public

Tulsishah: because we need to set on only tokensrc only, that's why I made chnages in that variable only

vadlakondaswetha: please update the comment and resolve it.

Tulsishah: made all fields public. 
fixed

vadlakondaswetha (approved):

PR #830, opened 2022-10-11
Created Storage Handle Instance

Tulsishah:


main.go
```diff
@@ -138,6 +139,19 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
 ////////////////////////////////////////////////////////////////////////
 
 // Mount the file system according to arguments in the supplied context.
+func CreateStorageHandle(flags *flagStorage) (storageHandle storage.StorageHandle, err error) {
+	var tokenSrc oauth2.TokenSource
+
+	tokenSrc, err = auth.GetTokenSource(context.Background(), flags.KeyFile, flags.TokenUrl, true)
+	var storageClientConfigToken = storage.StorageClientConfig{TokenSrc: tokenSrc}
```
vadlakondaswetha, line 146: when will you set other fields in storageClientConfig?

vadlakondaswetha, line 146: variables should be named after their original types. rename this  to storageClientConfig

main.go
```diff
@@ -138,6 +139,19 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
...
+	if err != nil {
```
vadlakondaswetha, line 147: you should check for err != nil before creating storageClientConfigToken instance

main.go
```diff
@@ -138,6 +139,19 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
...
+		err = fmt.Errorf("get token source: %w", err)
+		return
+	}
```
vadlakondaswetha, line 150: please add new lines as and when appropriate. Here we have a err check due to which we might exit from the method execution, so its good to add a new line here to distinguish the logic

main_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"testing"
+
+	. "github.com/jacobsa/ogletest"
+)
+
+const TestBucketName string = "gcsfuse-default-bucket"
+
+func TestMains(t *testing.T) { RunTests(t) }
```
vadlakondaswetha, line 11: why is it plural? rename it to TestMain

main_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+type MainTest struct {
+}
+
+func init() { RegisterTestSuite(&MainTest{}) }
+
+func (t *MainTest) TestCreateStorageHandleEnableStorageClientLibraryIsTrue() {
+	storageHandle, err := CreateStorageHandle(&flagStorage{
+		EnableStorageClientLibrary: true,
+	})
```
vadlakondaswetha, line 25: nit: new line after test method invocation.

vadlakondaswetha (changes requested):

main_test.go
Tulsishah: It is sawing error "Wrong test signature"

main.go
Tulsishah: fixed

main.go
Tulsishah: fixed

main_test.go
Tulsishah: fixed

main.go
Tulsishah: fixed

Tulsishah: fixed

main.go
```diff
@@ -138,6 +139,27 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
...
+func createStorageHandle(flags *flagStorage) (storageHandle storage.StorageHandle, err error) {
+	var tokenSrc oauth2.TokenSource
+
+	tokenSrc, err = auth.GetTokenSource(context.Background(), flags.KeyFile, flags.TokenUrl, true)
+	if err != nil {
+		err = fmt.Errorf("get token source: %w", err)
+		return
+	}
+
+	var storageClientConfig = storage.StorageClientConfig{DisableHTTP2: true,
```
vadlakondaswetha, line 151: this is wrong.  If you hardcode the values here then what is the purpose of defining them as flags.
We define a parameter as flag, so that user has the feasibility to set whatever value they want. That flag values should be passed to the storageHandle

main.go
```diff
@@ -138,6 +139,28 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
...
 // Mount the file system according to arguments in the supplied context.
+func createStorageHandle(flags *flagStorage) (storageHandle storage.StorageHandle, err error) {
+	var tokenSrc oauth2.TokenSource
+
+	tokenSrc, err = auth.GetTokenSource(context.Background(), flags.KeyFile, flags.TokenUrl, true)
+	if err != nil {
+		err = fmt.Errorf("get token source: %w", err)
+		return
+	}
+	var storageClientConfig = storage.StorageClientConfig{
```
vadlakondaswetha, line 150: instead of var storageClientConfig - can you simply do storageClientConfig :=

main.go
```diff
@@ -138,6 +139,28 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
 ////////////////////////////////////////////////////////////////////////
 
 // Mount the file system according to arguments in the supplied context.
+func createStorageHandle(flags *flagStorage) (storageHandle storage.StorageHandle, err error) {
+	var tokenSrc oauth2.TokenSource
+
+	tokenSrc, err = auth.GetTokenSource(context.Background(), flags.KeyFile, flags.TokenUrl, true)
```
vadlakondaswetha, line 145: here also can you do tokenSrc, err := instead of declaring tokenSrc as a variable on top.

vadlakondaswetha (approved):

main.go
Tulsishah: fixed

main.go
Tulsishah: fixed

main.go
Tulsishah: fixed

PR #831, opened 2022-10-13
Create ListObjects method

Tulsishah:


internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +256,18 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
 	AssertEq(nil, obj)
 	AssertTrue(strings.Contains(err.Error(), "Error 412: Precondition failed"))
 }
+
+func (t *BucketHandleTest) TestListObjectMethod() {
+	obj, err := t.bucketHandle.ListObjects(context.Background(),
+		&gcs.ListObjectsRequest{
+			Prefix:                   "gcsfuse/default.txt",
```
vadlakondaswetha, line 263: can you add more test cases to cover the following scenarios:
1. testing prefix parameter - > with data present and not present
2. testing delimiter paramters -> for different type of results
3. including trailing delimiter param
4. projection value

internal/storage/bucket_handle.go
```diff
@@ -179,3 +180,37 @@ func (b *bucketHandle) CopyObject(ctx context.Context, req *gcs.CopyObjectReques
 	o = storageutil.ObjectAttrsToBucketObject(objAttrs)
 	return
 }
+
+func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequest) (listing *gcs.Listing, err error) {
+	// Converting *ListObjectsRequest to type *storage.Query as expected by the Go Storage Client.
+	query := &storage.Query{
+		Delimiter:                req.Delimiter,
+		Prefix:                   req.Prefix,
+		Projection:               storage.Projection(req.ProjectionVal),
```
vadlakondaswetha, line 189: why is projectionVal conversion not handled here
https://github.com/GoogleCloudPlatform/gcsfuse/blob/5651aa43ab0f75f56bb5c3353df60d9c950c75a9/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L285

vadlakondaswetha (changes requested):

internal/storage/bucket_handle.go

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle.go
```diff
@@ -179,3 +180,56 @@ func (b *bucketHandle) CopyObject(ctx context.Context, req *gcs.CopyObjectReques
...
+	switch int(req.ProjectionVal) {
+	// Projection Value 0 in jacobsa/gcloud maps to Projection Value 1 in Go Client API, that is for "full".
+	case 0:
+		convertedProjection = storage.Projection(1)
+	// Projection Value 1 in jacobsa/gcloud maps to Projection Value 2 in Go Client API, that is for "noAcl".
+	case 1:
+		convertedProjection = storage.Projection(2)
+	// Default Projection value in jacobsa/gcloud library is 0 that maps to 1 in Go Client API interface, and that is for "full".
+	default:
+		convertedProjection = storage.Projection(1)
```
vadlakondaswetha, line 196: please extract this into a seperate method and add unit tests for it

internal/storage/bucket_handle.go
```diff
@@ -179,3 +180,56 @@ func (b *bucketHandle) CopyObject(ctx context.Context, req *gcs.CopyObjectReques
...
+		}
+
+		// Converting attrs to *Object type.
+		if req.IncludeTrailingDelimiter {
+			currObject := storageutil.ObjectAttrsToBucketObject(attrs)
+			list.Objects = append(list.Objects, currObject)
+		}
+
+		currCollapsedRuns := attrs.Name
+		list.CollapsedRuns = append(list.CollapsedRuns, currCollapsedRuns)
```
vadlakondaswetha, line 230: what is this collapsed runs and why do you need it

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +256,51 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+		&gcs.ListObjectsRequest{
+			Prefix:                   "gcsfuse/",
+			Delimiter:                "/",
+			IncludeTrailingDelimiter: true,
+			ContinuationToken:        "ContinuationToken",
+			MaxResults:               7,
+			ProjectionVal:            0,
+		})
+
+	AssertNe(nil, obj.Objects)
```
vadlakondaswetha, line 271: can you check the object contents. Checking for nil is not sufficient here, we need to check if correct object is returned.

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +256,51 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+		&gcs.ListObjectsRequest{
+			Prefix:                   "InvalidPrefix",
+			Delimiter:                "/",
+			IncludeTrailingDelimiter: true,
+			ContinuationToken:        "ContinuationToken",
+			MaxResults:               7,
+			ProjectionVal:            0,
+		})
+
+	AssertEq(nil, obj.Objects)
```
vadlakondaswetha, line 287: same here also.

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +256,51 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+			Delimiter:                "/",
+			IncludeTrailingDelimiter: true,
+			ContinuationToken:        "ContinuationToken",
+			MaxResults:               7,
+			ProjectionVal:            0,
+		})
+
+	AssertNe(nil, obj.Objects)
+	AssertNe(nil, obj.CollapsedRuns)
+	AssertEq(nil, err)
```
vadlakondaswetha, line 273: you should always check for error first. 
If error is not present, then we check for data. No point in checking for error after data check

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +256,51 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+}
+
+func (t *BucketHandleTest) TestListObjectMethodWithInValidPrefix() {
```
vadlakondaswetha, line 276: please rename this test method. Its not invalid prefix its just that object with given prefix doesn't exist/

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +256,51 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+		&gcs.ListObjectsRequest{
+			Prefix:                   "gcsfuse/",
+			Delimiter:                "/",
+			IncludeTrailingDelimiter: false,
+			ContinuationToken:        "ContinuationToken",
+			MaxResults:               7,
+			ProjectionVal:            0,
+		})
+
+	AssertEq(nil, obj.Objects)
```
vadlakondaswetha, line 303: same here. check for object.

Also what is the difference between this test and first check. can you differentiate with assert statements.

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +256,51 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+	AssertEq(nil, obj.CollapsedRuns)
+	AssertEq(nil, err)
+}
+
+func (t *BucketHandleTest) TestListObjectMethodWithIncludeTrailingDelimiter() {
+	obj, err := t.bucketHandle.ListObjects(context.Background(),
+		&gcs.ListObjectsRequest{
+			Prefix:                   "gcsfuse/",
+			Delimiter:                "/",
+			IncludeTrailingDelimiter: false,
```
vadlakondaswetha, line 297: also what about tests with includetrailing delimiter, max result and continuation token?

vadlakondaswetha (changes requested):

internal/storage/bucket_handle.go
Tulsishah: //  1. Prefix: "foo/", Delimiter: "/", IncludeTrailingDelimiter: true
	//     -> "foo/bar/" exists in both listing.CollapsedRuns and
	//     listing.Objects.
	//  2. Prefix: "foo/", Delimiter: "/", IncludeTrailingDelimiter: false
	//     -> "foo/bar/" exists in only listing.CollapsedRuns but not
	//     listing.Objects.

for IncludeTrailingDelimiter this is conditions, before we were not setting CollapsedRuns, that's why I made this condition

internal/storage/bucket_handle.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: Object is nil. there is data exist to check property

internal/storage/bucket_handle_test.go
Tulsishah: Object is nil. there is No data exist to check property

internal/storage/bucket_handle_test.go
Tulsishah: fakeserverobject is only checking Prefix,Delimiter.
https://github.com/GoogleCloudPlatform/gcsfuse/blob/9b96e9dc586959661a1ad54622bbb7419e7e3e68/vendor/github.com/fsouza/fake-gcs-server/fakestorage/object.go#L287

internal/storage/bucket_handle.go
```diff
@@ -179,3 +180,59 @@ func (b *bucketHandle) CopyObject(ctx context.Context, req *gcs.CopyObjectReques
...
 	return
 }
+
+func getProjectionValue(req gcs.Projection) storage.Projection {
+	// Explicitly converting Projection Value because the ProjectionVal interface of jacobsa/gcloud and Go Client API are not coupled correctly.
+	var convertedProjection storage.Projection // Stores the Projection Value according to the Go Client API Interface.
+	switch int(req) {
+	// Projection Value 0 in jacobsa/gcloud maps to Projection Value 1 in Go Client API, that is for "full".
+	case 0:
+		convertedProjection = storage.Projection(1)
```
vadlakondaswetha, line 190: instead of integers please use enum names. Like ProjectionNoAcl, ProjectionFull

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +257,90 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
 	AssertEq(nil, obj)
 	AssertTrue(strings.Contains(err.Error(), "Error 412: Precondition failed"))
 }
+
+func (t *BucketHandleTest) TestGetProjectionValueEqualZero() {
```
vadlakondaswetha, line 261: same here also, rename the test cases using projection names instead of integters.
Like TestGetProjectValueWhenGcloudProjectionIsNoAcl

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +257,90 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+	AssertEq(nil, obj.Objects)
+	AssertEq(nil, obj.CollapsedRuns)
+}
+
+func (t *BucketHandleTest) TestListObjectMethodWithIncludeTrailingDelimiter() {
+	obj, err := t.bucketHandle.ListObjects(context.Background(),
+		&gcs.ListObjectsRequest{
+			Prefix:                   "gcsfuse/",
+			Delimiter:                "/",
+			IncludeTrailingDelimiter: true,
```
vadlakondaswetha, line 317: add testcase for includetrailingdelimiter false scenario. I see include trailingdelimiter is handled: https://github.com/GoogleCloudPlatform/gcsfuse/blob/9b96e9dc586959661a1ad54622bbb7419e7e3e68/vendor/github.com/fsouza/fake-gcs-server/fakestorage/object.go#L316

vadlakondaswetha (changes requested):

internal/storage/bucket_handle.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle.go
```diff
@@ -179,3 +180,60 @@ func (b *bucketHandle) CopyObject(ctx context.Context, req *gcs.CopyObjectReques
...
+		if err == iterator.Done {
+			err = nil
+			break
+		}
+		if err != nil {
+			err = fmt.Errorf("Error in iterating through objects: %v", err)
+			return
+		}
+
+		// If Prefix of an attribute is set, only prefix will be populated through CollapsedRun.
```
vadlakondaswetha, line 226: Rephrase this comment.
Prefix attribute will be set for the objects returned as part of Prefix[] array in list response.

Also add a link to Prefix definition which says this.

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +257,94 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+			Prefix:                   "gcsfuse/",
+			Delimiter:                "/",
+			IncludeTrailingDelimiter: false,
+			ContinuationToken:        "ContinuationToken",
+			MaxResults:               7,
+			ProjectionVal:            0,
+		})
+
+	AssertEq(nil, err)
+	AssertEq(TestObjectRootFolderName, obj.Objects[0].Name)
```
vadlakondaswetha, line 326: also check that no other objects should be present in the response by adding an assert on the size. same for all tests.

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +257,94 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
 	AssertEq(nil, obj)
 	AssertTrue(strings.Contains(err.Error(), "Error 412: Precondition failed"))
 }
+
+func (t *BucketHandleTest) TestGetProjectValueWhenGcloudProjectionIsNoAcl() {
+	proj := getProjectionValue(gcs.Projection(storage.ProjectionNoACL))
```
vadlakondaswetha, line 262: I didnt follow. what are you trying to do here. you need to  pass gcs.Projection type to getProjectionValue method. 

Why are you creating a gcs.Projection using storage.Projection?

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +257,94 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+
+	AssertEq(storage.Projection(1), proj)
```
vadlakondaswetha, line 264: please compare enum values so its easier to understand

AssertEq(storage.ProjectionNoAcl, proj)

internal/storage/bucket_handle.go

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +257,102 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+			IncludeTrailingDelimiter: true,
+			ContinuationToken:        "ContinuationToken",
+			MaxResults:               7,
+			ProjectionVal:            0,
+		})
+
+	AssertEq(nil, err)
+	AssertEq(nil, obj.Objects)
+	AssertEq(nil, obj.CollapsedRuns)
+	AssertEq(0, len(obj.Objects))
```
vadlakondaswetha, line 314: if obj.objects and obj.collapsedruns is nil, why to check for length.

internal/storage/bucket_handle_test.go
```diff
@@ -256,3 +257,102 @@ func (t *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObjectNotExi
...
+			MaxResults:               7,
+			ProjectionVal:            0,
+		})
+
+	AssertEq(nil, err)
+	AssertEq(TestObjectRootFolderName, obj.Objects[0].Name)
+	AssertEq(TestObjectName, obj.Objects[1].Name)
+	AssertEq(TestObjectSubRootFolderName, obj.CollapsedRuns[0])
+	AssertEq(2, len(obj.Objects))
+	AssertEq(1, len(obj.CollapsedRuns))
```
vadlakondaswetha, line 334: you should always check for length first because verifying the list contents. make change in all test methods.

vadlakondaswetha (approved):

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle_test.go
Tulsishah: fixed

PR #840, opened 2022-11-03
Handling continuation token and Max results param from listobjects request and setting continuation token in response

Tulsishah:


internal/storage/bucket_handle.go
```diff
@@ -210,11 +210,16 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
...
 	pi := itr.PageInfo()
 	pi.MaxSize = req.MaxResults
+	pi.Token = req.ContinuationToken
 	var list gcs.Listing
 
 	// Iterating through all the objects in the bucket and one by one adding them to the list.
 	for {
 		var attrs *storage.ObjectAttrs
+		if len(list.Objects) == req.MaxResults {
+			err = nil
```
vadlakondaswetha, line 220: 1. add a comment why this if condition is required
2. err is nil by default. Do we still need to set err=nil?

internal/storage/bucket_handle_test.go
```diff
@@ -353,3 +353,21 @@ func (t *BucketHandleTest) TestListObjectMethodWithEmptyDelimiter() {
 	AssertEq(TestObjectGeneration, obj.Objects[0].Generation)
 	AssertEq(nil, obj.CollapsedRuns)
 }
+
+func (t *BucketHandleTest) TestListObjectMethodForMaxResult() {
```
vadlakondaswetha, line 357: how many objects did you add in the fakeserver?

vadlakondaswetha (changes requested):

internal/storage/bucket_handle_test.go
Tulsishah: four

internal/storage/bucket_handle.go

internal/storage/bucket_handle_test.go
vadlakondaswetha: add a comment here that we have 4 objects in fakeserver

internal/storage/bucket_handle.go
```diff
@@ -210,14 +210,18 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 	itr := b.bucket.Objects(ctx, query) // Returning iterator to the list of objects.
 	pi := itr.PageInfo()
 	pi.MaxSize = req.MaxResults
+	pi.Token = req.ContinuationToken
 	var list gcs.Listing
 
 	// Iterating through all the objects in the bucket and one by one adding them to the list.
 	for {
 		var attrs *storage.ObjectAttrs
+		// Fetching number of objects equals to requested MaxResults
```
vadlakondaswetha, line 219: Please update the comment like this. itr.next returns all the objects present in the bucket. Hence adding a check to break after required number of objects are returned.

vadlakondaswetha (changes requested):
Please update the pr description.

internal/storage/bucket_handle_test.go
Tulsishah: fixed

internal/storage/bucket_handle.go

vadlakondaswetha (approved):

PR #841, opened 2022-11-08
Initialising start and end for range equal to nil condition

Tulsishah:


internal/storage/bucket_handle.go
```diff
@@ -41,8 +41,12 @@ func (bh *bucketHandle) NewReader(
 	ctx context.Context,
 	req *gcs.ReadObjectRequest) (rc io.ReadCloser, err error) {
 	// Initialising the starting offset and the length to be read by the reader.
-	start := int64((*req.Range).Start)
-	end := int64((*req.Range).Limit)
+	start := int64(0)
+	end := int64(-1)
```
vadlakondaswetha, line 45: setting end to -1 and computing length from it is confusing. can you directly set the length here and update it in the if block below like this:

length: -1
if(rangle !=nil) {
end = int64(req.Range.limit)
length = end-start
}

internal/storage/bucket_handle.go
```diff
@@ -41,8 +41,12 @@ func (bh *bucketHandle) NewReader(
 	ctx context.Context,
 	req *gcs.ReadObjectRequest) (rc io.ReadCloser, err error) {
 	// Initialising the starting offset and the length to be read by the reader.
-	start := int64((*req.Range).Start)
-	end := int64((*req.Range).Limit)
+	start := int64(0)
```
vadlakondaswetha, line 44: do we need to make it int64 explicitly?

internal/storage/bucket_handle.go
```diff
@@ -41,8 +41,12 @@ func (bh *bucketHandle) NewReader(
...
+	end := int64(-1)
+	if req.Range != nil {
```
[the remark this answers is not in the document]
Tulsishah: fixed

vadlakondaswetha (changes requested):

internal/storage/bucket_handle.go
Tulsishah: yes because req.range is returning in uint64 but in reader method the parameter type is int64.

internal/storage/bucket_handle.go

internal/storage/bucket_handle.go
```diff
@@ -41,9 +41,15 @@ func (bh *bucketHandle) NewReader(
 	ctx context.Context,
 	req *gcs.ReadObjectRequest) (rc io.ReadCloser, err error) {
 	// Initialising the starting offset and the length to be read by the reader.
-	start := int64((*req.Range).Start)
-	end := int64((*req.Range).Limit)
-	length := int64(end - start)
+	start := int64(0)
+	length := int64(-1)
+	// If Range equals to nil, it has a default call which has start and length as above values otherwise it will take mentioned range.
```
vadlakondaswetha, line 46: rephrase it to  Following the semantics of NewReader method. Passing start, length as 0,-1 reads the entire file.

vadlakondaswetha (approved):

PR #851, opened 2022-11-23
Handling the order of lock for implicit dir node in lookupinode flow

vadlakondaswetha:
We need to follow this order for fetching locks -> inode and then filesystem level lock. Incase of implicitDir flow, we are fetching taking fs lock and then inode lock which is conflicting with other flow in the code that follows locking order. This conflict is creating deadlocks.

This PR fixes the deadlock by following the lock order in case of implicit dir flow.

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -649,6 +649,9 @@ func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Ino
 			if !ok {
 				in = fs.mintInode(ic)
 				fs.implicitDirInodes[in.Name()] = in.(inode.DirInode)
+				// Since we are creating inode here, there is no chance that something else
+				// is holding the lock for inode. Hence its safe to take lock on inode 
+				// without releasing fs.mu.lock.
 				in.Lock()
 				return
 			}
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -642,7 +642,8 @@ func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Ino
 		}
 
 		var ok bool
-		for {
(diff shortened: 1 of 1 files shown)


internal/fs/fs.go
```diff
@@ -872,16 +895,16 @@ func (fs *fileSystem) unlockAndDecrementLookupCount(in inode.Inode, N uint64) {
 //
 // Typical usage:
 //
-//     func (fs *fileSystem) doFoo() (err error) {
```
raj-prince, line 875: Looks like some editor formatting changes, please revert.

internal/fs/fs.go
```diff
@@ -642,14 +642,37 @@ func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Ino
 		}
 
 		var ok bool
-		in, ok = fs.implicitDirInodes[ic.FullName]
-		// If we don't have an entry, create one.
-		if !ok {
-			in = fs.mintInode(ic)
-			fs.implicitDirInodes[in.Name()] = in.(inode.DirInode)
+		var maxTriesToCreateInode = 3
```
raj-prince, line 645: Just want to confirm, with this code changes it will handle the case if we perform 3 concurrent operations (2 forget iNode and CreateInode in progress). If we add one more concurrent forget inode operations, it'll crash. Right? This will not go to the dead-lock at least in this workflow.

raj-prince (approved):
Need minor confirmation.

internal/fs/fs.go
vadlakondaswetha: Lint is not complaining, so i feel this is right formatting.

internal/fs/fs.go
vadlakondaswetha: I am returning nil incase max tries are exhausted. So we will end up returning error to kernel and no crash will happen.

PR #854, opened 2022-11-28
Adding Name method to get name of bucket

Tulsishah:


internal/storage/storage_handle.go
```diff
@@ -100,11 +100,12 @@ func NewStorageHandle(ctx context.Context, clientConfig StorageClientConfig) (sh
...
 	storageBucketHandle := sh.client.Bucket(bucketName)
-	_, err = storageBucketHandle.Attrs(context.Background())
+	obj, err := storageBucketHandle.Attrs(context.Background())
 	if err != nil {
 		return
 	}
 
-	bh = &bucketHandle{bucket: storageBucketHandle}
+	storageBucketName := obj.Name
+	bh = &bucketHandle{bucket: storageBucketHandle, bucketName: storageBucketName}
```
vadlakondaswetha, line 109: you can directly use bucketName: obj.Name. Avoids using extra variable

vadlakondaswetha (approved):

internal/storage/storage_handle.go
Tulsishah: fixed

PR #859, opened 2022-11-30
[Don't Review] Support for Frequency Based Log Rotation

raj-prince:
>> You need to change the code to change the frequency.
>> time.Sleep(<time_period>) you can change the frequency inside HandleReInitLogFile, This will start create different file during the run time.

For example,
time_period = 20sec and file_name is logs.txt
t = 0 -> logs.txt
t = 20 -> logs0.txt
t = 40 -> logs1.txt

You need to clone raj-prince/gcsfuse repo and branch_name = log_rotation.

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -15,9 +15,14 @@
 package logger
 
 import (
+	"fmt"
 	"io"
 	"log"
 	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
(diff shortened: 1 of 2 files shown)


raj-prince:
Created the PR with the same content through branch in the GoogleCloudPlatform (upstream): https://github.com/GoogleCloudPlatform/gcsfuse/pull/890

PR #867, opened 2022-12-15
Fixing the Performance Drop in Random Reads - Implemented using Go Storage Client

Tulsishah:
Issue:

Because of NopCloser conversion  for converting Reader object to ReadCloser, we got less bandwidth and more latencies for random reads.
For 50mb File size
Bandwidth - 224 MiB/s
Latency - 12084.52 ms

Fixes:

After rechecking http_client [NewRangeReader](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L771) method we got  to know that the object is ReadCloser type only so we don't need extra conversion of NopCloser.
For 50mb File size
Bandwidth - 1184MiB/s
Latency - 2199.80ms

Reason:

This could be because of we are copying heavy object many times during conversion.

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -64,6 +64,7 @@ func (bh *bucketHandle) NewReader(
 	}
 
 	// Returning NewRangeReader instance.
+	// "storage.Reader" is a io.ReadCloser, since it contains both Read() and Close() method.
 	return obj.NewRangeReader(ctx, start, length)
 }
 func (b *bucketHandle) DeleteObject(ctx context.Context, req *gcs.DeleteObjectRequest) error {
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -64,15 +64,12 @@ func (bh *bucketHandle) NewReader(
 	}
 
 	// Creating a NewRangeReader instance.
-	r, err := obj.NewRangeReader(ctx, start, length)
+	rc, err = obj.NewRangeReader(ctx, start, length)
 	if err != nil {
 		err = fmt.Errorf("error in creating a NewRangeReader instance: %v", err)
 		return
 	}
 
-	// Converting io.Reader to io.ReadCloser by adding a no-op closer method
-	// to match the return type interface.
-	rc = io.NopCloser(r)
 	return
 }
 func (b *bucketHandle) DeleteObject(ctx context.Context, req *gcs.DeleteObjectRequest) error {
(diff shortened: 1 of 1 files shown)


internal/storage/bucket_handle.go
```diff
@@ -64,15 +64,12 @@ func (bh *bucketHandle) NewReader(
 	}
 
 	// Creating a NewRangeReader instance.
-	r, err := obj.NewRangeReader(ctx, start, length)
+	rc, err = obj.NewRangeReader(ctx, start, length)
 	if err != nil {
```
raj-prince, line 68: We don't need this check now. we can remove it.

internal/storage/bucket_handle.go
```diff
@@ -64,15 +64,12 @@ func (bh *bucketHandle) NewReader(
 	}
 
 	// Creating a NewRangeReader instance.
-	r, err := obj.NewRangeReader(ctx, start, length)
+	rc, err = obj.NewRangeReader(ctx, start, length)
```
raj-prince, line 67: We can return directly, no need to create a "rc", "err" variable.

raj-prince (changes requested):
Also, please add a proper description.
(a) What was the issue before?
(b) How this change fixed the issue?

internal/storage/bucket_handle.go
raj-prince: Also, please add a one line comment that - "storage.Reader" is a io.ReadCloser, since it contains both Read() and Close() method.

internal/storage/bucket_handle.go

internal/storage/bucket_handle.go
```diff
@@ -63,17 +63,9 @@ func (bh *bucketHandle) NewReader(
...
-	if err != nil {
-		err = fmt.Errorf("error in creating a NewRangeReader instance: %v", err)
-		return
-	}
-
-	// Converting io.Reader to io.ReadCloser by adding a no-op closer method
-	// to match the return type interface.
-	rc = io.NopCloser(r)
-	return
+	// Returning NewRangeReader instance.
```
raj-prince, line 66: Instead we can club both comments in one - 
"NewRangeReader creates a "storage.Reader" object which is also a io.ReadCloser since it contains both Read() and Close() methods present in io.ReadCloser interface."

raj-prince (approved):
Please incorporate the changes before submission.

internal/storage/bucket_handle.go

PR #868, opened 2022-12-20
Pass the appname flag as part of user agent to Go storage library

Tulsishah:
Passing AppName flag as part of user agent to Go Storage library. Checking Environment variable for DLVM and DLC and adding that into UserAgent.

### DIFF
--- a/internal/storage/storage_handle_test.go
+++ b/internal/storage/storage_handle_test.go
@@ -34,6 +34,7 @@ func getDefaultStorageClientConfig() (clientConfig StorageClientConfig) {
 		HttpClientTimeout:   800 * time.Millisecond,
 		MaxRetryDuration:    30 * time.Second,
 		RetryMultiplier:     2,
+		UserAgent:           "AppName",
 	}
 }
 
@@ -97,6 +98,13 @@ func (t *StorageHandleTest) TestNewStorageHandleHttp2Enabled() {
 func (t *StorageHandleTest) TestNewStorageHandleWithZeroMaxConnsPerHost() {
 	sc := getDefaultStorageClientConfig()
 	sc.MaxConnsPerHost = 0
+
+	t.invokeAndVerifyStorageHandle(sc)
+}
+
+func (t *StorageHandleTest) TestNewStorageHandleWithDifferentUserAgent() {
+	sc := getDefaultStorageClientConfig()
+	sc.UserAgent = "UserAgent"
 
 	t.invokeAndVerifyStorageHandle(sc)
 }
--- a/internal/storage/storage_handle_test.go
+++ b/internal/storage/storage_handle_test.go
@@ -34,7 +34,7 @@ func getDefaultStorageClientConfig() (clientConfig StorageClientConfig) {
 		HttpClientTimeout:   800 * time.Millisecond,
 		MaxRetryDuration:    30 * time.Second,
 		RetryMultiplier:     2,
-		UserAgent:           "AppName",
+		UserAgent:           "gcsfuse/unknown (Go version go1.20-pre3 cl/474093167 +a813be86df)",
 	}
 }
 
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -139,12 +139,26 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
 ////////////////////////////////////////////////////////////////////////
 
 // Mount the file system according to arguments in the supplied context.
+
+func getUserAgent(AppName string) string {
+	var userAgent string
+
+	if os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE") != "" {
```
sethiay, line 146: Do you also think we can simplify it by setting user agent like:
userAgent = strings.TrimSpace(fmt.Sprintf("gcsfuse/%s %s %s", getVersion(), appName, os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE")))

main.go
```diff
@@ -139,12 +139,26 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
 ////////////////////////////////////////////////////////////////////////
 
 // Mount the file system according to arguments in the supplied context.
+
+func getUserAgent(AppName string) string {
```
sethiay, line 143: AppName should be in camelCase ? i.e. appName

main_test.go
```diff
@@ -25,3 +26,34 @@ func (t *MainTest) TestCreateStorageHandleEnableStorageClientLibraryIsTrue() {
...
+func (t *MainTest) TestGetUserAgentWhereEnvironmentVariableEqualDLVM() {
+	os.Setenv("GCSFUSE_METADATA_IMAGE_TYPE", "DLVM")
+	defer os.Unsetenv("GCSFUSE_METADATA_IMAGE_TYPE")
+
+	userAgent := getUserAgent("AppName")
+
+	ExpectEq("gcsfuse/unknown (Go version go1.20-pre3 cl/474093167 +a813be86df) AppName DLVM", userAgent)
+}
+
+func (t *MainTest) TestGetUserAgentEqualNil() {
```
sethiay, line 55: UserAgent is not Nil but the environment variable is not set. So, can we change the name of function to reflect this case ?

main.go
```diff
@@ -153,6 +167,7 @@ func createStorageHandle(flags *flagStorage) (storageHandle storage.StorageHandl
 		HttpClientTimeout:   flags.HttpClientTimeout,
 		MaxRetryDuration:    flags.MaxRetryDuration,
 		RetryMultiplier:     flags.RetryMultiplier,
+		UserAgent:           getUserAgent(flags.AppName),
```
sethiay, line 170: Your current changes takes care of user agent in Storage client library. We also need to take care when used with jacobsa/gcloud: https://github.com/GoogleCloudPlatform/gcsfuse/blob/5a4355196a0a551505f1175195eafece577a82e4/main.go#L98

internal/storage/storage_handle_test.go
```diff
@@ -34,6 +34,7 @@ func getDefaultStorageClientConfig() (clientConfig StorageClientConfig) {
 		HttpClientTimeout:   800 * time.Millisecond,
 		MaxRetryDuration:    30 * time.Second,
 		RetryMultiplier:     2,
+		UserAgent:           "AppName",
```
sethiay, line 37: Can we set some close to real-world UserAgent ?

internal/storage/storage_handle_test.go
```diff
@@ -100,3 +101,10 @@ func (t *StorageHandleTest) TestNewStorageHandleWithZeroMaxConnsPerHost() {
 
 	t.invokeAndVerifyStorageHandle(sc)
 }
+
+func (t *StorageHandleTest) TestNewStorageHandleWithDifferentUserAgent() {
+	sc := getDefaultStorageClientConfig()
+	sc.UserAgent = "UserAgent"
```
sethiay, line 107: Same: Can we set some close to real-world UserAgent ?

sethiay (changes requested):

main.go
Tulsishah: fixed

main_test.go
Tulsishah: fixed

main.go
Tulsishah: fixed

internal/storage/storage_handle_test.go
Tulsishah: fixed

main.go
Tulsishah: fixed

internal/storage/storage_handle_test.go
Tulsishah: fixed

PR #870, opened 2022-12-23
Return `ENAMETOOLONG` when name is too long

gaul:
This allows more pjdfstests to pass.  References #590.


gaul:
Seems like this was cancelled after 10 minutes: Error: The operation was canceled.

internal/fs/inode/dir.go
```diff
@@ -421,6 +422,10 @@ const ConflictingFileNameSuffix = "\n"
 
 // LOCKS_REQUIRED(d)
 func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error) {
+	if index := strings.LastIndex(name, "/"); index != -1 && len(name)-index > 255 {
```
vadlakondaswetha, line 425: can you add unit tests for this.

PR #887, opened 2023-01-16
copying crc by value instead of by-reference from attrs object

vadlakondaswetha:
Crc returned from go-client library is copied by reference because of which entire attrs object returned by go-client library is not getting garbage collected. Fixing it to copy by value instead of reference


Tulsishah (commented):
Unit test for the same is not passing

Tulsishah (changes requested):
Unit test for the same is not passing

vadlakondaswetha:
> 

fixed it

Tulsishah (approved):

PR #896, opened 2023-01-19
create scorecard.yml

rsaksham:
Currently it is set to run once on monday in every week and on push to master branch
Currently upload artifact is set , if we remove this the results would only be visible to maintainers

PAT tokens aren't being setup due to risks they impose 
To see all checks visit https://github.com/ossf/scorecard#scorecard-checks

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -2,7 +2,7 @@ name: Scorecard supply-chain security
 on:
   branch_protection_rule:
   schedule:
-    # runs periodically on 12:01 utc (5:31pm) ist on monday
+    # runs periodically on 12:01 utc (5:31pm ist) on monday
     - cron: '01 12 * * 1'
   push:
     branches: [ "master" ]
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,52 @@
+name: Scorecard supply-chain security
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '01 12 * * 1'
+  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,52 @@
+name: Scorecard supply-chain security
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '01 12 * * 1'
```
vadlakondaswetha, line 5: can you add a comment what schedule does it represent

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,52 @@
+name: Scorecard supply-chain security
```
vadlakondaswetha, line 1: are there any existing security concerns that needs to be fixed.

.github/workflows/scorecard.yml
rsaksham: done

.github/workflows/scorecard.yml
rsaksham: I ran a scorecard run using docker run -e GITHUB_AUTH_TOKEN=token gcr.io/openssf/scorecard:stable --show-details --repo=https://github.com/GoogleCloudPlatform/gcsfuse
Overall score is 6.1/10. Please find the results here https://drive.google.com/file/d/1mHM4NWukKBjC504dK17nIM1xnH1Cvo8a/view?usp=sharing&resourcekey=0-fceWBJ2PKYwMSZI9Acjelg

vadlakondaswetha (approved):

Issue #891, opened 2023-01-18
GCSFUSE hangs when verifying the content of a bucket.


marwoodandrew:
We have had a number of occurrences of gcsfuse hanging in our production environment and have now managed to replicate the behavior in development. We retrieve a list filepaths/file names from a database or text file traverse the list and issue an ls -i, file and md5sum against each.

**Here is the process to make the gcsfuse hang:**

Running as root :-

**1. Mount gcsfuse:**

umount /usr/dot/dc/device/dev_aap/dev97
umount /usr/dot/dc/device/dev_aap/dev97
umount: /usr/dot/dc/device/dev_aap/dev97: not mounted
mount ph-dev-images-bkt /usr/dot/dc/device/dev_aap/dev97 -t gcsfuse -o rw,_netdev,allow_other,uid=1111,gid=1111,key_file=/etc/google-svc-acnt-key.dev,log_file=/dot/dc/logs/aap/gcsfuse.log,log_format=text,debug_fuse_errors,debug_fuse,debug_fs,debug_gcs,debug_invariants,debug_mutex
Calling gcsfuse with arguments: --gid 1111 --log-file /dot/dc/logs/aap/gcsfuse.log --debug_fuse_errors --debug_fs --key-file /etc/google-svc-acnt-key.dev --debug_gcs -o allow_other --log-format text --debug_fuse --debug_invariants -o rw --uid 1111 --debug_mutex ph-dev-images-bkt /usr/dot/dc/device/dev_aap/dev97
2023/01/18 10:31:12.338265 Start gcsfuse/0.41.10 (Go version go1.18.4) for app "" using mount point: /usr/dot/dc/device/dev_aap/dev97

**2. Make test data**
./GCS-TESTER.sh /usr/dot/dc/device/dev_aap/dev97 make-test-files
… wait an hour or two

**3. Check test data**
./GCS-TESTER.sh /usr/dot/dc/device/dev_aap/dev97 check-test-files
…

After several minutes the checking may hang after printing something like the following :
[...]

PR #898, opened 2023-01-20
Support experimental_enable_storage_client_library flag in mount command

sethiay:
This change was identified in https://github.com/GoogleCloudPlatform/gcsfuse/issues/891#issuecomment-1397905072

### DIFF
--- a/tools/mount_gcsfuse/main.go
+++ b/tools/mount_gcsfuse/main.go
@@ -83,6 +83,7 @@ func makeGcsfuseArgs(
 		case "implicit_dirs",
 			"disable_http2",
 			"experimental_local_file_cache",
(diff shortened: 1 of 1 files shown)


tools/mount_gcsfuse/main.go
```diff
@@ -83,6 +83,7 @@ func makeGcsfuseArgs(
 		case "implicit_dirs",
 			"disable_http2",
 			"experimental_local_file_cache",
+			"experimental_enable_storage_client_library",
```
vadlakondaswetha, line 86: Can you create a bug to add a g3doc mentionning what needs to be done when a new flag is added

vadlakondaswetha (approved):

PR #901, opened 2023-01-23
Updating newer version of cloud storage library

Tulsishah:


raj-prince:
General Comments:
(a) Please provide the description - why do we want to update this? What are the changes in the new library - please link the update release?
(b) Always try to work from the forked branch not in the main repository.
(c) Check from where we are getting the linting issue - if it's coming from the vendor directory - please make change to ignore that folder while linting.
(d) Please do a sanity over the normal file-system operation and fio load test for read write operations, just to make sure there shouldn't be any significant performance drop with this change.

Tulsishah:
Created same PR with forked branch

PR #909, opened 2023-01-31
Adding custom retry mechanism in Go Storage Client

raj-prince:
>> Added support for custom error retry in go-client workflow.
>> Adding 401 as a retryable error - specially this 401 retryable error will be reverted after the resolution of the issue: https://github.com/golang/oauth2/issues/623

### DIFF
--- a/internal/storage/storageutil/custom_retry.go
+++ b/internal/storage/storageutil/custom_retry.go
@@ -1,0 +1,78 @@
+package storageutil
+
+import (
+	"io"
+	"net"
+	"net/url"
+
+	"google.golang.org/api/googleapi"
+)
+
+func ShouldRetry(err error) (b bool) {
+	// HTTP 50x errors.
+	if typed, ok := err.(*googleapi.Error); ok {
+		if typed.Code >= 500 && typed.Code < 600 {
+			b = true
+			return
+		}
+	}
+
+	// HTTP 429 errors (GCS uses these for rate limiting).
+	if typed, ok := err.(*googleapi.Error); ok {
+		if typed.Code == 429 {
+			b = true
+			return
+		}
+	}
+
+	// HTTP 401 errors - Invalid Credentials
+	// This is a work-around to fix the rare issue, we get while executing the
+	// long ml-based test on gcsfuse. The better fix would be to expire the
+	// cached token a little earlier which is used to access the GCS resources
+	// through API.
+
+	// TODO: please replace this implementation with the correction one after the
+	// resolution of this issue: https://github.com/golang/oauth2/issues/623
+	if typed, ok := err.(*googleapi.Error); ok {
+		if typed.Code == 401 {
+			b = true
+			return
+		}
+	}
+
+	// Network errors, which tend to show up transiently when doing lots of
+	// operations in parallel. For example:
+	//
+	//     dial tcp 74.125.203.95:443: too many open files
+	//
+	if _, ok := err.(*net.OpError); ok {
+		b = true
+		return
+	}
+
+	// The HTTP package returns ErrUnexpectedEOF in several places. This seems to
+	// come up when the server terminates the connection in the middle of an
+	// object read.
+	if err == io.ErrUnexpectedEOF {
+		b = true
+		return
+	}
+
+	// The HTTP library also appears to leak EOF errors from... somewhere in its
+	// guts as URL errors sometimes.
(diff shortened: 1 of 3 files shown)


internal/storage/storageutil/custom_retry.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"io"
+	"net"
+	"net/url"
+
+	"google.golang.org/api/googleapi"
+)
+
+func ShouldRetry(err error) (b bool) {
+	// HTTP 50x errors.
```
vadlakondaswetha, line 12: add a comment above that this code is replica of retry_bucket.go

internal/storage/storageutil/custom_retry.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// HTTP 429 errors (GCS uses these for rate limiting).
+	if typed, ok := err.(*googleapi.Error); ok {
+		if typed.Code == 429 {
+			b = true
+			return
+		}
+	}
+
+	// HTTP 401 errors - Invalid Credentials
```
vadlakondaswetha, line 28: move this extra logic to end after retry_bucket.go logic is done

internal/storage/storageutil/custom_retry.go
```diff
@@ -0,0 +1,78 @@
...
+	"net"
+	"net/url"
+
+	"google.golang.org/api/googleapi"
+)
+
+func ShouldRetry(err error) (b bool) {
+	// HTTP 50x errors.
+	if typed, ok := err.(*googleapi.Error); ok {
+		if typed.Code >= 500 && typed.Code < 600 {
```
vadlakondaswetha, line 14: did you explicitly validate each error from the go-client library and that its getting retried. It might happen that we are expecting googleapi.error in case 5xx and go-client library is returned something else.

internal/storage/storageutil/custom_retry.go
```diff
@@ -0,0 +1,78 @@
...
+	if typed, ok := err.(*googleapi.Error); ok {
+		if typed.Code == 429 {
+			b = true
+			return
+		}
+	}
+
+	// HTTP 401 errors - Invalid Credentials
+	// This is a work-around to fix the rare issue, we get while executing the
+	// long ml-based test on gcsfuse. The better fix would be to expire the
```
vadlakondaswetha, line 30: this issue might happen for non ml-based as well if gcsfuse is running for a long time and never stopped. Just mention that this is a workaround to fix the corner case where <mention method which returns token as valid with code link> returns the token as valid but GCS says invalid. This might happen because of client-server timer issues.

Actual fix will be refresh the token earlier than 1hr. Changes will be done post resolution of this issue:

internal/storage/storageutil/custom_retry_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func TestCustomRetry(t *testing.T) { RunTests(t) }
+
+type customRetryTest struct {
+}
+
+func init() { RegisterTestSuite(&customRetryTest{}) }
+
+func (t customRetryTest) TestGoogleAPIErrorRetry() {
```
vadlakondaswetha, line 20: lets use consistent naming for the new tests we write in GCSFuse
say ShouldRetryWithGoogleApiErrorReturnsTrue
or
ShouldRetryReturnsTrueWithGoogleApiError

vadlakondaswetha (changes requested):

internal/storage/storageutil/custom_retry.go
raj-prince: Added this logic in the last after the go-client logic.

internal/storage/storageutil/custom_retry_test.go
```diff
@@ -0,0 +1,70 @@
...
+		Code: 429,
+		Body: "API rate limit exceeded",
+	}
+
+	ExpectEq(ShouldRetry(&err401), true)
+	ExpectEq(ShouldRetry(&err502), true)
+	ExpectEq(ShouldRetry(&err429), true)
+}
+
+func (t customRetryTest) ShouldRetryReturnsFalseWithGoogleApiError() {
```
vadlakondaswetha, line 42: isn't this conflicting with above test case
say here ShouldRetryReturnsFalseWithGoogleApiError400

vadlakondaswetha (approved):

PR #910, opened 2023-01-31
Fixing 412 precondition error in CreateObject method(Write flow)

Tulsishah:
As createObject method was throwing googleapi error instead gcs precondition error, sync function was not catching the error properly. convert the error in gcs precondition error in createobject method.


internal/fs/inode/file.go
```diff
@@ -545,10 +547,12 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 
 	// Special case: a precondition error means we were clobbered, which we treat
 	// as being unlinked. There's no reason to return an error in that case.
-	var preconditionErr *gcs.PreconditionError
-	if errors.As(err, &preconditionErr) {
-		err = nil
-		return
+	var gErr *googleapi.Error
+	if errors.As(err, &gErr) {
```
vadlakondaswetha, line 551: does it work for jacobsa flow as well?

Tulsishah: In jacobsa flow it is not throwing 412 error

Tulsishah: fixed

internal/storage/bucket_handle.go
```diff
@@ -133,8 +133,20 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
...
 	if err = wc.Close(); err != nil {
-		err = fmt.Errorf("error in closing writer: %v", err)
-		return
+		if err != nil {
+			switch ee := err.(type) {
+			case *googleapi.Error:
+				if ee.Code == http.StatusPreconditionFailed {
+					err = &gcs.PreconditionError{Err: ee}
+				}
+				if ee.Code == http.StatusNotFound {
```
vadlakondaswetha, line 142: just curious when will createObject throw notfound error

Tulsishah: Sorry my bad, for create object it will not throw not found error.

internal/storage/bucket_handle.go
```diff
@@ -133,8 +133,17 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 	// We can't use defer to close the writer, because we need to close the
 	// writer successfully before calling Attrs() method of writer.
 	if err = wc.Close(); err != nil {
-		err = fmt.Errorf("error in closing writer: %v", err)
-		return
+		if err != nil {
```
vadlakondaswetha, line 136: why do you need err!=nil condition here when you already have it in line 135??

internal/storage/bucket_handle.go
```diff
@@ -133,8 +133,17 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
...
+			switch ee := err.(type) {
```
vadlakondaswetha, line 137: do you need a switch condition for one case or can you use if-else as well?

vadlakondaswetha (changes requested):

internal/storage/bucket_handle.go

internal/storage/bucket_handle.go

internal/storage/bucket_handle.go
```diff
@@ -133,7 +134,14 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 	// We can't use defer to close the writer, because we need to close the
 	// writer successfully before calling Attrs() method of writer.
 	if err = wc.Close(); err != nil {
-		err = fmt.Errorf("error in closing writer: %v", err)
+		var gErr *googleapi.Error
+		if errors.As(err, &gErr) {
+			if gErr.Code == http.StatusPreconditionFailed {
+				err = &gcs.PreconditionError{Err: err}
```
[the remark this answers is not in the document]
Tulsishah: fixed

vadlakondaswetha (approved):

PR #932, opened 2023-02-13
Adding logrotate configuration as a post-installation for gcsfuse package (deb/rpm)

raj-prince:
### Description
This change will introduce a script, which will be responsible for making the configuration changes for logrotate and rsyslog linux package. All these configuration required to support log-rotation for gcsfuse logs. So, by default gcsfuse will write the logs to /var/log/gcsfuse.log with the help of rsyslog package.

This script will be used while creating the gcsfuse-package for installation. In the process of installation, this script will be executed before the installation of gcsfuse.

### Testing details
1. Manual
Tested the overall system on - all the supported linux systems. (a) Ubuntu (22, 20, 18) (b) Deb (11, 10) (c) Centos - (9, 8) (d) RHEL-7 (e) SLES-12

2. Unit tests - NA
3. Integration tests - NA


tools/logrotate_config/package_after_install_script.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Create the log file for gcsfuse package.
+mkdir -p /var/log/gcsfuse/
+
+# Installs logrotate configuration files.
+mkdir -p /etc/logrotate.d/
+
+cat > /etc/logrotate.d/gcsfuse <<EOF
+/var/log/gcsfuse/*.log {
+    rotate ${LOGROTATE_FILES_MAX_COUNT:-5}
```
raj-prince, line 11: Pass max count as build_arg and convert into a environment variable - LOGROTATE_FILES_MAX_COUNT, same for LOGROTATE_MAX_SIZE.

vadlakondaswetha: will these changes be done after this PR is merged?

tools/logrotate_config/package_after_install_script.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
vadlakondaswetha, line 1: rename the directory to log_rotate/install_script.sh

tools/logrotate_config/package_after_install_script.sh
```diff
@@ -0,0 +1,22 @@
...
+# Create the log file for gcsfuse package.
+mkdir -p /var/log/gcsfuse/
+
+# Installs logrotate configuration files.
+mkdir -p /etc/logrotate.d/
+
+cat > /etc/logrotate.d/gcsfuse <<EOF
+/var/log/gcsfuse/*.log {
+    rotate ${LOGROTATE_FILES_MAX_COUNT:-5}
+    copytruncate
```
vadlakondaswetha, line 12: can you add comments on what these are doing.

vadlakondaswetha (changes requested):

tools/logrotate_config/package_after_install_script.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Create the log file for gcsfuse package.
+mkdir -p /var/log/gcsfuse/
```
vadlakondaswetha, line 4: what happens to the log-file flag ?

sethiay, line 4: Ensure this directory is created with non root user. Also check for path starting with /etc/ below.

tools/logrotate_config/package_after_install_script.sh
```diff
@@ -0,0 +1,22 @@
...
+mkdir -p /etc/logrotate.d/
+
+cat > /etc/logrotate.d/gcsfuse <<EOF
+/var/log/gcsfuse/*.log {
+    rotate ${LOGROTATE_FILES_MAX_COUNT:-5}
+    copytruncate
+    missingok
+    notifempty
+    compress
+    maxsize ${LOGROTATE_MAX_SIZE:-100M}
```
sethiay, line 16: Can we rename it to LOGROTATE_MAX_FILE_SIZE ?

tools/package_gcsfuse_docker/Dockerfile
```diff
@@ -52,7 +52,8 @@ RUN fpm \
     -d fuse \
     --vendor "" \
     --url "https://$GCSFUSE_REPO" \
-    --description "A user-space file system for Google Cloud Storage."
+    --description "A user-space file system for Google Cloud Storage." \
+    --after-install "${GCSFUSE_PATH}/tools/logrotate_config/package_after_install_script.sh"
```
sethiay, line 56: As discussed, confirm if this installation is done at the time package is installed on a machine and what happens if the machine/docker container on which this package is installed doesn't have logrotate library installed.

tools/logrotate_config/package_after_install_script.sh
raj-prince: done.

tools/logrotate_config/package_after_install_script.sh
raj-prince: Yes, done.

tools/logrotate_config/package_after_install_script.sh
raj-prince: For now, we have fixed the value and user can change the config, we will provide documents for that.

tools/logrotate_config/package_after_install_script.sh
raj-prince: I'll create a log-rotation docs to explain all the log-rotation related informations useful for user.

tools/package_gcsfuse_docker/Dockerfile
raj-prince: I checked and found gcsfuse gets installed irrespective of what verdict we get as an execution of post-script. Hence, started applying all these configuration as before-script so in case of failure we will abort the gcsfuse installation.

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
+
+# Create directory, which keeps all logrotate-config to rotate hourly.
+mkdir /etc/logrotate.hourly.d
```
sethiay, line 4: Though it's rare but it's possible that this directory already exist, so can we use mkdir -p option here?

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,73 @@
...
+
+cat << EOF | tee /etc/logrotate.hourly.conf
+# packages drop hourly log rotation information into this directory
+include /etc/logrotate.hourly.d
+EOF
+
+chmod 644 /etc/logrotate.hourly.conf
+
+cat << EOF | tee /etc/cron.hourly/logrotate
+#!/bin/bash
```
sethiay, line 14: similar to above comment, assuming this will not interfere with existing configurations in logrotate file (if there are any)

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
+
+# Create directory, which keeps all logrotate-config to rotate hourly.
+mkdir /etc/logrotate.hourly.d
+
+cat << EOF | tee /etc/logrotate.hourly.conf
```
sethiay, line 6: I think it would be really helpful to add one line comment about the logic here. Same for next cat block.

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,73 @@
...
+if [ $? -eq 0 ]; then
+        echo "Logrotate config for gcsfuse updated successfully!"
+else
+        echo "Please install linux package - logrotate"
+        exit 1
+fi
+
+# Syslog configuration to filter and redirect the logs from /var/log/syslog to
+# /var/log/gcsfuse.log.
+cat > /etc/rsyslog.d/08-gcsfuse.conf <<EOF
```
sethiay, line 49: just for my understanding - why is specifically 08 here?

tools/log_rotate/install_script.sh
raj-prince: Done.

tools/log_rotate/install_script.sh
raj-prince: Added.

tools/log_rotate/install_script.sh
raj-prince: Yes.

tools/log_rotate/install_script.sh
raj-prince: This number is for config file ordering and precedence. I.e. if the same parameter configuration is present in 10-x.conf and 20-y.conf the latter will overwrite first one. 08 is just a random number, we can increase this number.

tools/logrotate_config/package_after_install_script.sh
raj-prince: Nope, all the changes will go in the same PR.

vadlakondaswetha (approved):
Approving it based on Ayush review

tools/log_rotate/install_script.sh
sethiay: we can write a comment for this.

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Change the ownership of create log file through rsyslog.
+# Change the ownership of create log file through rsyslog.
```
sethiay, line 64: nit: delete the same comment.

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,84 @@
...
+# hourly execution of logrotate, we have created logrotate config-files specific
+# hourly-execution.
+
+# This file describes about the hourly execution of logrotate. All the logrotate
+# -config files present in /etc/logrotate.hourly.d will be executed hourly.
+cat << EOF | tee /etc/logrotate.hourly.conf
+# This to enforce logrotate to run as root/adm. In ubuntu distributions, the
+# ownership of the file created via rsyslog is syslog hence, we need to add
+# this explicitly.
+su root adm
```
sethiay, line 17: as discussed, to confirm this doesn't cause any issues.

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Add a shell script which will be run hourly, which eventually executes the
+# command to rotate the logs according to config present in /etc/logrotate.hourly.d
+cat << EOF | tee /etc/cron.hourly/logrotate
+#!/bin/bash
+test -x /usr/sbin/logrotate || exit 0
+/usr/sbin/logrotate /etc/logrotate.hourly.conf
+EOF
+chmod 775 /etc/cron.hourly/logrotate
+
```
sethiay, line 32: As discussed, add similar check as below to here as well for /etc/cron.hourly

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,84 @@
...
+su root adm
+
+# packages drop hourly log rotation information into this directory
+include /etc/logrotate.hourly.d
+EOF
+chmod 644 /etc/logrotate.hourly.conf
+
+# Add a shell script which will be run hourly, which eventually executes the
+# command to rotate the logs according to config present in /etc/logrotate.hourly.d
+cat << EOF | tee /etc/cron.hourly/logrotate
```
sethiay, line 26: As discussed, please check if renaming logrotate to gcsfuse is better.

tools/log_rotate/install_script.sh
raj-prince: By default, log-rotation is done via root user. In case of the owner of the file is 'syslog', if we don't change we get the below error info, which say to explicitly convey the user/group using which logrotate will rotate the log.


Error Info: Parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell log-rotate which user/group should be user for rotation.

tools/log_rotate/install_script.sh
```diff
@@ -0,0 +1,94 @@
...
+        echo "Logrotate config for gcsfuse updated successfully!"
+else
+        echo "Please install linux package - logrotate"
+        exit 1
+fi
+
+# Syslog configuration to filter and redirect the logs from /var/log/syslog to
+# /var/log/gcsfuse.log. The prefix-number 08 in the gcsfuse.conf is just for
+# file ordering and precedence. Like, if the same parameter configuration is
+# present in 10-x.conf and 20-y.conf the latter will overwrite first one.
```
sethiay, line 71: nit:
Like -> e.g.
will overwrite first one - take precedence over the first one.

sethiay (approved):

PR #934, opened 2023-02-14
Removing http client timeout

Tulsishah:
Removing timeout from storage_handle_test file.


flags.go
```diff
@@ -219,12 +219,6 @@ func newApp() (app *cli.App) {
 					"inodes.",
 			},
 
-			cli.DurationFlag{
```
vadlakondaswetha, line 222: aren't these changes already done.

Tulsishah: I did it from the same branch that's why whole changes came.

PR #944, opened 2023-02-20
Commenting out all fs tests

vadlakondaswetha:
Unit tests are getting stuck intermittently. 
The issue is kernel stops sending messages even though there are no errors returned for previous messages from kernel.
Right now, mounting and unmounting is done for every test case. Instead of that, we will move to mounting/umounting at struct level (class level in java terms). This seems to work.

Changes to convert initialisation to struct level are huge, hence will be done iteratively.

### DIFF
--- a/internal/fs/all_buckets_test.go
+++ b/internal/fs/all_buckets_test.go
@@ -14,24 +14,11 @@
 
 package fs_test
 
-import (
-	"io"
-	"io/ioutil"
-	"os"
-	"path"
-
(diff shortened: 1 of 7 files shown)


sethiay (approved):
Approving as it's not a breaking change. Please check if using build tags can be a better option here.

vadlakondaswetha:
[Using](https://github.com/GoogleCloudPlatform/gcsfuse/pull/945)

PR #949, opened 2023-02-21
GCSFuse code change to write the logs to syslog files

raj-prince:
Going forward:
(a) when we don't pass any log-file, this will write the logs to syslog files.
(b) when we pass the log-file, it will start writing the logs to the specified file path.

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -24,9 +24,10 @@ import (
 	"github.com/jacobsa/daemonize"
 )
 
-// ProgrammeName constant is used while writing the logs to syslog file, it is
-// used while filtering the gcsfuse log-message from the syslog file, since
-// syslog file contains all the system related logs from other programmes too.
+// Syslog file contains logs from all different programmes running on the VM.
+// ProgrammeName is prefixed to all the logs written to syslog. This constant is
+// used to filter the logs from syslog and write it to respective log files -
+// gcsfuse.log in case of GCSFuse.
 const ProgrammeName string = "gcsfuse"
 
 var (
@@ -55,9 +56,9 @@ func InitLogFile(filename string, format string) error {
 	} else {
 		// Priority consist of facility and severity, here facility to specify the
 		// type of system that is logging the message to syslog and severity is log-level.
-		// Here, we are using facility as syslog.LOG_LOCAL7 - is a custom facility syslog
-		// provides for the user and severity as syslog.LOG_DEBUG which is for debug
-		// level messages.
+		// User applications are allowed to take facility value between LOG_LOCAL0
+		// to LOG_LOCAL7. We are using LOG_LOCAL7 as facility and LOG_DEBUG to write
+		// debug messages.
 		sysWriter, err = syslog.New(syslog.LOG_LOCAL7|syslog.LOG_DEBUG, ProgrammeName)
 		if err != nil {
 			return fmt.Errorf("error while creating syswriter: %w", err)
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -15,12 +15,19 @@
 package logger
 
 import (
+	"fmt"
 	"io"
 	"log"
+	"log/syslog"
 	"os"
 
 	"github.com/jacobsa/daemonize"
 )
+
+// ProgrammeName constant is used while writing the logs to syslog file, it is
+// used while filtering the gcsfuse log-message from the syslog file, since
+// syslog file contains all the system related logs from other programmes too.
+const ProgrammeName string = "gcsfuse"
 
 var (
 	defaultLoggerFactory *loggerFactory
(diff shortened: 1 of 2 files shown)


internal/logger/logger.go
```diff
@@ -114,26 +136,37 @@ func (f *loggerFactory) newLogger(level, prefix string) *log.Logger {
...
+		if f.format == "json" {
+			return &jsonWriter{
+				w:     f.sysWriter,
+				level: level,
+			}
+		} else {
+			return &textWriter{
+				w:     f.sysWriter,
+				level: level,
+			}
```
vadlakondaswetha, line 160: please extract this logic into private method and use it for both file and syswriter

internal/logger/logger.go
```diff
@@ -114,26 +136,37 @@ func (f *loggerFactory) newLogger(level, prefix string) *log.Logger {
...
+		}
+	} else {
```
vadlakondaswetha, line 162: Earlier this is not an else case, why did you make it else?

internal/logger/logger.go
```diff
@@ -15,34 +15,55 @@
...
+		f, err = os.OpenFile(
+			filename,
+			os.O_WRONLY|os.O_CREATE|os.O_APPEND,
+			0644,
+		)
+		if err != nil {
+			return err
+		}
+	} else {
+		sysWriter, err = syslog.New(syslog.LOG_ALERT, ProgrammeName)
```
vadlakondaswetha, line 56: why do we need log_alert and why does lower levels not work?

vadlakondaswetha (changes requested):

internal/logger/logger.go
raj-prince: This method (https://github.com/GoogleCloudPlatform/gcsfuse/blob/adeea67dbd30253431c9a052e2efb44d4a3c7996/internal/logger/logger.go#L53) is called in the start, at that time both file/syslog writers are not initialised and flow goes to the else block.

internal/logger/logger.go
raj-prince: LOG_DEBUG was also working but logs were going to a different file - /var/log/debug.log. I changed the syslog configuration to get debug level logs to /var/log/gcsfuse.log instead of /var/log/debug.log file.

internal/logger/logger.go
vadlakondaswetha: you have return statement in every clause. you dont need else conditions.

if f.file != nil {
    return
}

if f.syswriter != nil {
   return
}

switch level

internal/logger/logger.go
```diff
@@ -103,37 +124,43 @@ func Info(v ...interface{}) {
...
-				w:     f.file,
-				level: level,
-			}
+func (f *loggerFactory) createJsonOrTextWriter(level string, writer io.Writer) io.Writer {
+	if f.format == "json" {
+		return &jsonWriter{
+			w:     writer,
+			level: level,
+		}
+	} else {
```
vadlakondaswetha, line 143: same here also. we can make the default as textWriter.
if f.format == "json" {
		return &jsonWriter{
			w:     writer,
			level: level,
		}
	}


return &textWriter {}

internal/logger/logger.go
```diff
@@ -15,34 +15,55 @@
...
+		f, err = os.OpenFile(
+			filename,
+			os.O_WRONLY|os.O_CREATE|os.O_APPEND,
+			0644,
+		)
+		if err != nil {
+			return err
+		}
+	} else {
+		sysWriter, err = syslog.New(syslog.LOG_LOCAL7|syslog.LOG_DEBUG, ProgrammeName)
```
vadlakondaswetha, line 56: add a comment what is log_local7 mean and why do you need both log_local7 and log_debug

vadlakondaswetha (changes requested):

internal/logger/logger.go
```diff
@@ -15,34 +15,60 @@
...
 	"log"
+	"log/syslog"
 	"os"
 
 	"github.com/jacobsa/daemonize"
 )
 
+// ProgrammeName constant is used while writing the logs to syslog file, it is
+// used while filtering the gcsfuse log-message from the syslog file, since
+// syslog file contains all the system related logs from other programmes too.
```
vadlakondaswetha, line 29: rephrase it.

Syslog file contains logs from all different programmes running on the VM. Programme name is prefixed to all logs written to syslog. This constant is used to filter the logs from syslog and write it to respective log files(gcsfuse.log for gcsfuse.)

internal/logger/logger.go
```diff
@@ -15,34 +15,60 @@
...
+			os.O_WRONLY|os.O_CREATE|os.O_APPEND,
+			0644,
+		)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Priority consist of facility and severity, here facility to specify the
+		// type of system that is logging the message to syslog and severity is log-level.
+		// Here, we are using facility as syslog.LOG_LOCAL7 - is a custom facility syslog
```
vadlakondaswetha, line 58: instead method user applications are allowed to use any value between local0 to local7 and we are using local7.

vadlakondaswetha (approved):

PR #950, opened 2023-02-21
Making other fs tests to use struct level init and teardown

vadlakondaswetha:
Earlier cl: https://github.com/GoogleCloudPlatform/gcsfuse/pull/947

### DIFF
--- a/internal/fs/fs_test.go
+++ b/internal/fs/fs_test.go
@@ -43,9 +43,10 @@ import (
 )
 
 const (
-	filePerms      os.FileMode = 0740
-	dirPerms                   = 0754
-	RenameDirLimit             = 5
+	filePerms            os.FileMode = 0740
+	dirPerms                         = 0754
+	RenameDirLimit                   = 5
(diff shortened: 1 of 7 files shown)


internal/fs/fs_test.go
```diff
@@ -141,7 +144,7 @@ func (t *fsTest) SetUpTestSuite() {
 		appendThreshold: 0,
 		tmpObjectPrefix: ".gcsfuse_tmp/",
 	}
-	t.serverCfg.RenameDirLimit = 5
+	t.serverCfg.RenameDirLimit = RenameDirLimit
 	t.serverCfg.SequentialReadSizeMb = 200
```
sethiay, line 148: Can we create constant for SequentialReadSizeMb as well ?

sethiay (approved):

internal/fs/fs_test.go
vadlakondaswetha: done.

PR #952, opened 2023-02-21
Use RetryAlways policy in conjunction with custom ShouldRetry

sethiay:
Currently, the default WithPolicy is [RetryIdempotent](https://github.com/googleapis/google-cloud-go/blob/storage/v1.29.0/storage/storage.go#L1945) which only causes idempotent request to be retried in case of errors. Because of this, we recently observed failure in CopyObject operation(which is not idempotent) due to intermittent backend err(503) from GCS as retry was not performed.
With this change, we are shifting to RetryAlways policy which will try to retry for all requests regardless of idempotency considerations.

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -90,16 +90,19 @@ func NewStorageHandle(ctx context.Context, clientConfig StorageClientConfig) (sh
 		return
 	}
 
+	// ShouldRetry function checks if an operation should be retried based on the
+	// response of operation (error.Code).
+	// RetryAlways causes all operations to be checked for retries using
+	// ShouldRetry function.
+	// Without RetryAlways, only those operations are checked for retries which
+	// are idempotent.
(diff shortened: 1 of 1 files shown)


internal/storage/storage_handle.go
```diff
@@ -90,13 +90,16 @@ func NewStorageHandle(ctx context.Context, clientConfig StorageClientConfig) (sh
 		return
 	}
 
-	// RetryAlways causes all operations to be retried when the service returns a transient error, regardless of
-	// idempotency considerations.
 	sc.SetRetry(
```
vadlakondaswetha, line 93: Adding seperate comments is little confusing. One says always retried and other says based on error.code

can you simplify and add a comment on the top saying that every operation will be retried if storageutil.shouldRetry returns true. Without storage.RetryAlways only idempotency operations will be checked for retry. Add link to the code where this logic is present.

vadlakondaswetha (approved):

internal/storage/storage_handle.go
sethiay: Done

PR #953, opened 2023-02-22
Making allbuckets and caching tests use struct level init and tear down

vadlakondaswetha:
Cleanup of unused variables in fs_test.go will be done after all tests are fixed.

### DIFF
--- a/internal/fs/fs_test.go
+++ b/internal/fs/fs_test.go
@@ -228,7 +228,7 @@ func (t *fsTest) TearDown() {
 	// for next test run.
 
 	// ReadDirPicky throws error incase of allbuckets_test. That is expected since
-	// we can list buckets when bucket-name is not specified during mount.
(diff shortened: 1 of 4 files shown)


internal/fs/fs_test.go
```diff
@@ -219,8 +223,16 @@ func (t *fsTest) TearDown() {
 	}
 
 	// Remove all contents for mntDir. This helps to keep the directory clean
-	// for next test run
-	os.RemoveAll(mntDir)
+	// for next test run.
+
+	// ReadDirPicky throws error incase of allbuckets_test. That is expected since
+	// we can list buckets when bucket-name is not specified during mount.
```
sethiay, line 229: can list -> can't list ?

sethiay (approved):

internal/fs/fs_test.go
vadlakondaswetha: done

PR #954, opened 2023-02-22
Fixing Test Failing in CI

Tulsishah:
Fixing this [Issue](https://b.corp.google.com/issues/270090832) 
Created test_cred.json file and passing it as key file in tests.


fakecreds.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"os"
+)
+
+// Create a fake creds.json file for testing
+func CreateFakeCreds(creds string) (err error) {
+	content := "{\n  \"type\": \"service_account\"," +
```
vadlakondaswetha, line 9: create a test file called test_creds.json under testdata folder and use that in main_test.go

Tulsishah: fixed

vadlakondaswetha (approved):

PR #984, opened 2023-03-02
GCSFuse code change to write the logs to syslog files

raj-prince:
### Description
1. Reverts GoogleCloudPlatform/gcsfuse#977
2. Modified log behavior - If user passes --log-file in case all the logs will be written to given log file. Otherwise in background mode gcsfuse will write the log to syslog and in foreground mode logs will be printed to console as stdout/err.

### Testing
1. Manual - Checked all the four behaviors manually.
2. Integration Test - NA
3. Unit Test - NA

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -15,7 +15,6 @@
 package logger
 
 import (
-	"fmt"
 	"io"
 	"log"
 	"log/syslog"
@@ -65,7 +64,7 @@ func InitLogFile(filename string, format string) error {
 			// debug messages.
 			sysWriter, err = syslog.New(syslog.LOG_LOCAL7|syslog.LOG_DEBUG, ProgrammeName)
 			if err != nil {
-				fmt.Printf("error while creating syswriter: %w", err)
+				Infof("error while creating sys writer: %w", err)
 				sysWriter = nil
 			}
 		} else {
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -15,12 +15,20 @@
 package logger
 
 import (
+	"fmt"
 	"io"
 	"log"
+	"log/syslog"
 	"os"
 
 	"github.com/jacobsa/daemonize"
 )
+
+// Syslog file contains logs from all different programmes running on the VM.
+// ProgrammeName is prefixed to all the logs written to syslog. This constant is
+// used to filter the logs from syslog and write it to respective log files -
+// gcsfuse.log in case of GCSFuse.
+const ProgrammeName string = "gcsfuse"
 
 var (
 	defaultLoggerFactory *loggerFactory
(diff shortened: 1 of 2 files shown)


internal/logger/logger.go
```diff
@@ -17,32 +17,66 @@ package logger
...
+			// type of system that is logging the message to syslog and severity is log-level.
+			// User applications are allowed to take facility value between LOG_LOCAL0
+			// to LOG_LOCAL7. We are using LOG_LOCAL7 as facility and LOG_DEBUG to write
+			// debug messages.
+			sysWriter, err = syslog.New(syslog.LOG_LOCAL7|syslog.LOG_DEBUG, ProgrammeName)
+			if err != nil {
+				Infof("error while creating sys writer: %v", err)
+				sysWriter = nil
+			}
+		} else {
```
vadlakondaswetha, line 70: syswriter is by default nil. why do you need to set it again?

internal/logger/logger.go
```diff
@@ -17,32 +17,66 @@ package logger
...
+		if _, ok := os.LookupEnv(GCSFuseAsBackground); ok {
+			// Priority consist of facility and severity, here facility to specify the
+			// type of system that is logging the message to syslog and severity is log-level.
+			// User applications are allowed to take facility value between LOG_LOCAL0
+			// to LOG_LOCAL7. We are using LOG_LOCAL7 as facility and LOG_DEBUG to write
+			// debug messages.
+			sysWriter, err = syslog.New(syslog.LOG_LOCAL7|syslog.LOG_DEBUG, ProgrammeName)
+			if err != nil {
+				Infof("error while creating sys writer: %v", err)
+				sysWriter = nil
```
vadlakondaswetha, line 68: why do you need to set it to nil?

internal/logger/logger.go
```diff
@@ -17,32 +17,66 @@ package logger
...
+	} else {
+		if _, ok := os.LookupEnv(GCSFuseAsBackground); ok {
+			// Priority consist of facility and severity, here facility to specify the
+			// type of system that is logging the message to syslog and severity is log-level.
+			// User applications are allowed to take facility value between LOG_LOCAL0
+			// to LOG_LOCAL7. We are using LOG_LOCAL7 as facility and LOG_DEBUG to write
+			// debug messages.
+			sysWriter, err = syslog.New(syslog.LOG_LOCAL7|syslog.LOG_DEBUG, ProgrammeName)
+			if err != nil {
+				Infof("error while creating sys writer: %v", err)
```
vadlakondaswetha, line 67: where does this line gets printed since logger is not initialized?

internal/logger/logger.go
```diff
@@ -17,32 +17,66 @@ package logger
...
 
 	"github.com/jacobsa/daemonize"
 )
 
+// Syslog file contains logs from all different programmes running on the VM.
+// ProgrammeName is prefixed to all the logs written to syslog. This constant is
+// used to filter the logs from syslog and write it to respective log files -
+// gcsfuse.log in case of GCSFuse.
+const ProgrammeName string = "gcsfuse"
+const GCSFuseAsBackground string = "GCSFUSE_AS_BACKGROUND"
```
vadlakondaswetha, line 31: Rename to GCSFuseInBackgroundMode

vadlakondaswetha (changes requested):

internal/logger/logger.go
raj-prince: By default, logger is initialized with the below based on label:
 	switch level {
	case "NOTICE":
		return daemonize.StatusWriter
	case "ERROR":
		return os.Stderr
	default:
		return os.Stdout
	}

When we change it to NewNotice, we start getting the info from the daemon.

Will raise another PR to change the behavior by default. Also, one of the asks of GKE team was - we get some logs as std::err. The reason is same we use NewNotice which returns the writer as daemonize.StatusWriter (which is initialized with stderr, and controlled by environment variable).

internal/logger/logger.go
raj-prince: Added just because of safety, since I din't find any doc/comment related to this which says in case of error, writer will be nil.

Removing this.

vadlakondaswetha (approved):

PR #986, opened 2023-03-03
Permission issue mounting a bucket in go client library

Tulsishah:
### Description
Before [Bucket Handle](https://github.com/GoogleCloudPlatform/gcsfuse/blob/ComponentCount_Not_Set/internal/storage/storage_handle.go#L109) method was making [bucket.Attrs](https://pkg.go.dev/cloud.google.com/go/storage#BucketHandle.Attrs) call which needs storage.object.get permission to check if the bucket does exist or not. I skip this call as we are already checking if the bucket exists [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/4859983c583c9f09f3edd26651c8de7f4e490439/internal/gcsx/bucket_manager.go#L248). Before this line only threw an error; I made it throw an error and closed mounting.

### Testing
Manual 
- Did sanity check(list, create, copy, delete operations)
-  What permissions does SA has - Storage Object Viewer
      - no storage.buckets.get permission
      - has storage.objects.get and storage.objects.list
      -  Output -Bucket mounted (read only operations are working)
          In case if bucket does not exist it is throwing error - Error in iterating through objects: storage: bucket doesn't exist
-  What permissions does SA has - Storage Admin
      - has storage.buckets.*  storage.objects.*
      - Output - Mounted Successfully. (all operations are working)
-  What permissions does SA has - Storage Legacy Bucket Reader
      - has storage.buckets.get, storage.objects.list 
      - Output -  Bucket mounted (read only operations are working)
-  What permissions does SA has - Storage Legacy Bucket Write
      - has storage.buckets.get 
      - Output - Bucket mounted (Write only operations are working)
-  What permissions does SA has - Storage Legacy Bucket Owner
      - Output - Bucket mounted (all operations are working)
-  What permissions does SA has - Storage Legacy Object Reader,  Storage Legacy Object Owner
     - no  storage.object.list permission
     - no storage.buckets.get
     - Output -We got this error - Error in iterating through objects: googleapi: Error 403: tulsishah@gcs-fuse-test.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist)., forbidden
mountWithArgs: mountWithConn: fs.NewServer: create file system: SetUpBucket: Error in iterating through objects: googleapi: Error 403: tulsishah@gcs-fuse-test.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist)., forbidden
    - In this case Jacobsa flow also throwing the same error.

Please refer this for [permissions ](https://cloud.google.com/storage/docs/access-control/iam-roles#legacy-roles)
       
Unit test - Added unit test for bucket does not exist in bucket_manager_test.go file
Integration test - N/A

### DIFF
--- a/internal/gcsx/bucket_manager.go
+++ b/internal/gcsx/bucket_manager.go
@@ -17,7 +17,6 @@ package gcsx
 import (
 	"errors"
 	"fmt"
-	"os"
 	"path"
 	"time"
 
@@ -155,7 +154,7 @@ func setUpRateLimiting(
 // bucket as described in that package.
 func (bm *bucketManager) SetUpGcsBucket(ctx context.Context, name string) (b gcs.Bucket, err error) {
 	if bm.config.EnableStorageClientLibrary {
-		b, err = bm.storageHandle.BucketHandle(name)
+		b = bm.storageHandle.BucketHandle(name)
 		if err != nil {
 			return
 		}
@@ -245,9 +244,9 @@ func (bm *bucketManager) SetUpBucket(
 	// Check whether this bucket works, giving the user a warning early if there
 	// is some problem.
 	{
-		_, err := b.ListObjects(ctx, &gcs.ListObjectsRequest{MaxResults: 1})
+		_, err = b.ListObjects(ctx, &gcs.ListObjectsRequest{MaxResults: 1})
 		if err != nil {
-			fmt.Fprintln(os.Stdout, "WARNING, bucket doesn't appear to work: ", err)
+			return
 		}
 	}
 
--- a/internal/gcsx/bucket_manager.go
+++ b/internal/gcsx/bucket_manager.go
@@ -155,9 +155,6 @@ func setUpRateLimiting(
 func (bm *bucketManager) SetUpGcsBucket(ctx context.Context, name string) (b gcs.Bucket, err error) {
 	if bm.config.EnableStorageClientLibrary {
 		b = bm.storageHandle.BucketHandle(name)
-		if err != nil {
-			return
-		}
 
 		if reqtrace.Enabled() {
 			b = gcs.GetWrappedWithReqtraceBucket(b)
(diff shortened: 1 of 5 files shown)


internal/gcsx/bucket_manager.go
```diff
@@ -155,7 +154,7 @@ func setUpRateLimiting(
 // bucket as described in that package.
 func (bm *bucketManager) SetUpGcsBucket(ctx context.Context, name string) (b gcs.Bucket, err error) {
 	if bm.config.EnableStorageClientLibrary {
-		b, err = bm.storageHandle.BucketHandle(name)
+		b = bm.storageHandle.BucketHandle(name)
 		if err != nil {
 			return
 		}
```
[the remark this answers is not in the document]
Tulsishah: fixed

internal/gcsx/bucket_manager_test.go
```diff
@@ -120,3 +119,33 @@ func (t *BucketManagerTest) TestSetUpBucketMethod() {
...
+	bm.config = bucketConfig
+	bm.gcCtx = ctx
+	bm.conn = &Connection{
+		wrapped: gcsfake.NewConn(timeutil.RealClock()),
+	}
+
+	bucket, err := bm.SetUpBucket(context.Background(), invalidBucketName)
+
+	// Error in iterating through objects: storage: bucket doesn't exist
+	ExpectNe(nil, err)
```
vadlakondaswetha, line 149: are we not going to validate the error message?

Tulsishah: It's because it is a string, and if [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/bucket_handle.go#L234) message changes in the future, the test can fail; that's why I didn't compare the string error message and just checked if it is non-nil.

Tulsishah: fixed

vadlakondaswetha (approved):

PR #988, opened 2023-03-06
Adding condition when project team is nil in ObjectAttr conversion

Tulsishah:
### Description
Throwing nil pointer exception when the project team was nil. 
### Testing
- Manual
     - Did sanity check(list, create, copy, delete operations)
     - Project team is nil(in User), it will set Project Team nil in Object (see Unit test)
     - If Access = Uniform: No object-level ACLs enabled - ACL=nil 
     - If Access = Fine-grained: Object-level ACLs enabled - It has 4 ACLs (3 Project Related and 1 User) - for the first three project team will not equal to nil so that It will set, and for the fourth(User) ProjectTeam will be nil so It will set nil.
- Unit-test -
     - Created unit test when the project team was nil
- Integration-test - N/A

### DIFF
--- a/internal/storage/storageutil/object_attrs.go
+++ b/internal/storage/storageutil/object_attrs.go
@@ -24,31 +24,45 @@ import (
 )
 
 func convertObjectAccessControlToACLRule(obj *storagev1.ObjectAccessControl) storage.ACLRule {
-	return storage.ACLRule{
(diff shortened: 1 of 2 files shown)


Tulsishah (commented):
Added all the Details.

vadlakondaswetha (approved):

PR #990, opened 2023-03-07
Making createObject idempotent

vadlakondaswetha:
1. Rightnow CreateObject is not idempotent because when generation is zero in the request, generation is not set in the call to GCS. To make it idempotent, we need to set storage.conditions.DoesNotExist = true when generation is zero.
Generation is zero means create the object only when object doesn't exist.
2. In the rename flow, when the destination directory already exists, precondition error is ignored. But rightn ow it also checks if the error has string **object exists** which is not returned by gcs. Hence removing it.

Testing:
Unit Tests: Added
Integration Tests: N/A
Manual Testing: Done for all scenarios where CreateObject is called using linux commands. Retry is simulated by calling CreateObject twice. Also added some custom log statements to log what is happening - whether the call is successful or failed

- CreateFile/MkNode method -> These are used to create a file if the file doesn't exist. Working as intended. If the file already exists, getting preconditions not met error.
- CreateSymlink -> Same as createFile. Failed if the object already exists. Passed otherwise
- MkDir -> Failed if the directory already exists. Passed otherwise
- RenameDir -> Checks if the destination directory already exists. If the directory already exists, it will ignore the error and move on. Verified for both destination directory exists and not exists.
- FullObjectCreator -> Used to overwrite the entire object when an edit is done. Validates if the generation matches. Verified that its returning success when generation matches and precondition failure when generation does not match. Overall edit will succeed since precondition error is ignored.
- AppendObjectCreator -> In the append flow, creates a temp object with the appended content with generation zero. Validated when object exists and not exists. Overall edit will succeed since precondition error is ignored.

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -375,8 +375,5 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
 }
 
 func isStorageConditionsNotEmpty(conditions storage.Conditions) bool {
-	if conditions != (storage.Conditions{}) {
-		return true
-	}
-	return false
+	return conditions != (storage.Conditions{})
 }
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -112,12 +112,24 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 	// MetaGenerationPrecondition - If non-nil, the object will be created/overwritten
 	// only if the current metaGeneration for the object name is equal to the given value.
 	// Zero means the object does not exist.
-	if req.GenerationPrecondition != nil && *req.GenerationPrecondition != 0 && req.MetaGenerationPrecondition != nil && *req.MetaGenerationPrecondition != 0 {
-		obj = obj.If(storage.Conditions{GenerationMatch: *req.GenerationPrecondition, MetagenerationMatch: *req.MetaGenerationPrecondition})
-	} else if req.GenerationPrecondition != nil && *req.GenerationPrecondition != 0 {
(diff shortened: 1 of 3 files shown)


internal/storage/bucket_handle.go
```diff
@@ -361,3 +373,10 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
 
 	return
 }
+
+func isStorageConditionsNotEmpty(conditions storage.Conditions) bool {
```
sethiay, line 377: I think 'isComplete' is better than 'isNotComplete'. Can we change this to 'isStorageConditionsEmpty' and then use '!isStorageConditionsEmpty(preconditions)' in bucket_handle.go ?

sethiay (commented):
- it would be helpful if you can write how you tested the tests in Manual sections. i.e. did you call the function individually or the ran linux commands ? How did you simulate retry ?
- did the application program/command failed when in case of precondition error ?

internal/storage/bucket_handle.go
vadlakondaswetha: the purpose of creating this method is to simply the check, if we are going to put not condition in every flow, its better to keep the method as not empty.

vadlakondaswetha:
> * it would be helpful if you can write how you tested the tests in Manual sections. i.e. did you call the function individually or the ran linux commands ? How did you simulate retry ?
> * did the application program/command failed when in case of precondition error ?



> * it would be helpful if you can write how you tested the tests in Manual sections. i.e. did you call the function individually or the ran linux commands ? How did you simulate retry ?
> * did the application program/command failed when in case of precondition error ?



> * it would be helpful if you can write how you tested the tests in Manual sections. i.e. did you call the function individually or the ran linux commands ? How did you simulate retry ?
> * did the application program/command failed when in case of precondition error ?

Done.

sethiay (approved):

PR #991, opened 2023-03-07
Compose object idempotency when dst object doesn't exist.

sethiay:
Description: Make ComposeObject method (when storage client library is enabled) idempotent when destination object doesn't exist. Currently, ComposeObject is called at the time of appending to file where dstObj = original object & srcObjects = [original object, temp object]. Hence, ComposeObject is always called with some GenerationPrecondition set. With these changes, ComposeObject will make idempotent requests even when dstObject doesn't exist and caller of ComposeObject sets 0 as GenerationPrecondition.

Testing:
Manual: 
1. Tried appending to a file of size 2mb and checked the debug_gcs logs. ComposeObject requests were successful.
2. Tried simulating retry of requests in 1# case by duplicating [this statement](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L711).
    - Application program doesn't fail.
    - GCS returns 404 error (StatusNotFound). This happens because one of the src object is dst obj also. So, in the duplicate request, src object is not found because it was changed in first request.

Unit tests:
1. Ran all the unit tests present in GCSFuse.
2. Added two unit tests for ComposeObject:
    - When destination object doesn't exist with 0 GenerationPrecondition (Note: fake-gcs-server doesn't respect the precondition checks).
    - When destination object is one of the source object. (which is close to what happens at the time of appending to file).

Integration tests:
1. Ran integration tests in tools/integration_tests/implicit-dirs

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -350,8 +350,8 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
 	// Only set conditions on dstObj if there is at least one condition in
 	// dstObjConds. Otherwise, storage client library gives empty conditions error.
 	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/7ad451c6f2ead7992e030503e5b66c555b2ebf71/vendor/cloud.google.com/go/storage/storage.go#L1739
-	// Note: Requests are not idempotent in this case and only idempotent if
-	// GenerationMatch or DoesNotExist is set.
+	// Note: Requests are not idempotent when conditions are empty and only
+	// idempotent if GenerationMatch or DoesNotExist is set.
 	if dstObjConds != (storage.Conditions{}) {
 		dstObj = dstObj.If(dstObjConds)
 	}
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -338,8 +338,6 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
 	// DstGenerationPrecondition or DoesNotExist should be set in dstObj
 	// preconditions to make requests Idempotent.
 	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/7ad451c6f2ead7992e030503e5b66c555b2ebf71/vendor/cloud.google.com/go/storage/copy.go#L230
-	// When DstGenerationPrecondition is 0, set DoesNotExist as true.
-	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/7ad451c6f2ead7992e030503e5b66c555b2ebf71/vendor/cloud.google.com/go/storage/storage.go#L1710
 	if req.DstGenerationPrecondition != nil {
 		if *req.DstGenerationPrecondition == 0 {
 			dstObjConds.DoesNotExist = true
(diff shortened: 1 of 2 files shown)


internal/storage/bucket_handle_test.go
```diff
@@ -765,3 +765,186 @@ func (t *BucketHandleTest) TestNameMethod() {
...
+			CacheControl:       CacheControl,
+			ContentDisposition: ContentDisposition,
+			CustomTime:         CustomTime,
+			EventBasedHold:     true,
+			StorageClass:       StorageClass,
+			Acl:                nil,
+		})
+
+	AssertEq(nil, err)
+	// Validation of srcObject1 to ensure that it is not effected.
```
vadlakondaswetha, line 822: contents are compared in the below test but not here.

internal/storage/bucket_handle.go
```diff
@@ -319,12 +319,29 @@ func (b *bucketHandle) UpdateObject(ctx context.Context, req *gcs.UpdateObjectRe
...
-		dstObj = dstObj.If(storage.Conditions{MetagenerationMatch: *req.DstMetaGenerationPrecondition})
+	dstObjConds := storage.Conditions{}
+	if req.DstMetaGenerationPrecondition != nil {
+		dstObjConds.MetagenerationMatch = *req.DstMetaGenerationPrecondition
+	}
+	// DstGenerationPrecondition or DoesNotExist should be set in dstObj
+	// preconditions to make requests Idempotent.
+	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/7ad451c6f2ead7992e030503e5b66c555b2ebf71/vendor/cloud.google.com/go/storage/copy.go#L230
+	// When DstGenerationPrecondition is 0, set DoesNotExist as true.
+	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/7ad451c6f2ead7992e030503e5b66c555b2ebf71/vendor/cloud.google.com/go/storage/storage.go#L1710
```
vadlakondaswetha, line 330: we shouldn't add too many comments in the code. If you click on dstObjConds.DoesNotExist its clear when to set

internal/storage/bucket_handle.go
```diff
@@ -319,12 +319,29 @@ func (b *bucketHandle) UpdateObject(ctx context.Context, req *gcs.UpdateObjectRe
...
+			dstObjConds.DoesNotExist = true
+		} else {
+			dstObjConds.GenerationMatch = *req.DstGenerationPrecondition
+		}
+	}
+	// Only set conditions on dstObj if there is at least one condition in
+	// dstObjConds. Otherwise, storage client library gives empty conditions error.
+	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/7ad451c6f2ead7992e030503e5b66c555b2ebf71/vendor/cloud.google.com/go/storage/storage.go#L1739
+	// Note: Requests are not idempotent when conditions are empty and only
+	// idempotent if GenerationMatch or DoesNotExist is set.
```
vadlakondaswetha, line 342: this line is already mentioned above. remove it.

vadlakondaswetha (approved):

internal/storage/bucket_handle_test.go
sethiay: Added comparison here as well.

internal/storage/bucket_handle.go
sethiay: removed this comment.

internal/storage/bucket_handle.go
sethiay: removed.

internal/storage/bucket_handle.go
```diff
@@ -319,12 +319,25 @@ func (b *bucketHandle) UpdateObject(ctx context.Context, req *gcs.UpdateObjectRe
...
+		if *req.DstGenerationPrecondition == 0 {
+			dstObjConds.DoesNotExist = true
+		} else {
+			dstObjConds.GenerationMatch = *req.DstGenerationPrecondition
+		}
+	}
+	// Only set conditions on dstObj if there is at least one condition in
+	// dstObjConds. Otherwise, storage client library gives empty conditions error.
+	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/7ad451c6f2ead7992e030503e5b66c555b2ebf71/vendor/cloud.google.com/go/storage/storage.go#L1739
+	if dstObjConds != (storage.Conditions{}) {
```
sethiay, line 339: use isStorageConditionsNotEmpty function from https://github.com/GoogleCloudPlatform/gcsfuse/pull/990/files, once merged.

PR #998, opened 2023-03-08
Make delete object idempotent

sethiay:
Description: DeleteObject method is **currently** not idempotent for the cases where generation is not passed or passed as 0. In those cases, DeleteObject makes delete GCS request without setting [generation](https://github.com/GoogleCloudPlatform/gcsfuse/blob/d5b9b44a99eebcc4c4c85af47b4dcd1356932d0a/internal/storage/bucket_handle.go#L74) in it and generation/ifGenerationMatch should be set to make delete GCS request idempotent as mentioned [here](https://cloud.google.com/storage/docs/retry-strategy#conditional-idempotency). With these changes, we always pass generation. If generation is not passed from caller then by default it puts 0 which signifies deleting latest version of object.

Testing
Manual:
For checking whether request is idempotent in each workflow, I added print statement [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/d5b9b44a99eebcc4c4c85af47b4dcd1356932d0a/vendor/cloud.google.com/go/storage/storage.go#L988). And to simulate retry of request, I added duplicate statement [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/d5b9b44a99eebcc4c4c85af47b4dcd1356932d0a/vendor/cloud.google.com/go/storage/http_client.go#L397).
Workflows tested:
1. RmDir: Using rmdir command.
    - rmdir abc/: Idempotent, Retry GCS call fails with 404 error and application program fails with I/O error (directory is deleted).
2. Unlink: Using rm abc.txt
    - rm abc.txt: Idempotent, Retry GCS call fails with 404 error and application program fails with I/O error (file is deleted).
3. Rename (directory): Using mv abc/ xyz/
    - mv abc ./xyz (abc is not empty & xyz is empty): It requires --rename-dir-limit flag, Idempotent, Retry GCS call fails with 404 error and application program fails with I/O error (files in abc are deleted, abc is not deleted, abc is created inside xyz and files are copied to xyz/abc/)
    - mv abc ./xyz (both abc and xyz are empty): Idempotent, Retry GCS call fails with 404 error and application program fails with no such abc/ directory (abc is deleted & xyz/abc is created)
    - mv abc ./xyz (xyz doesn't exist): Idempotent, Retry GCS call fails with 404 error and application program fails with no such abc/ directory (abc is deleted & xyz is created)
4. Rename (file): Using mv abc.txt xyz.txt
    - mv abc.txt xyz.txt (xyz doesn't exist):  Idempotent, Retry GCS call fails with 404 error and application program fails with no abc.txt file found (file is moved). 
    - mv abc.txt xyz.txt (xyz exist):  Idempotent, Retry GCS call fails with 404 error and application program fails with no abc.txt file found (file is moved).    
5. Compose (append): echo "hey" >> abc.txt (size of abc.txt is 2MiB exactly)
    - Idempotent, Retry GCS call fails with 404 error and application program doesn't fail.
6. Garbage collection: Make staleness threshold less than 1 minute and then see if objects starting with .gcsfuse_temp/ prefix are deleted. 
    - Idempotent, GCS calls fails with 404 and garbage collection is done i.e files are deleted.
7. Precondition error: Deliberately put wrong MetaGeneration in rename method.
   - Idempotent, GCS calls with 412 precondition error and application code fails with I/O error.
   
Unit tests: 
1. Ran all the unit tests present in under gcsfuse.
2. Added test case for DeleteObject method in bucket_handle_test.go:
    - Call DeleteObject without passing generation.

Integration tests:
Ran integration tests in tools/integration_tests/implicit-dirs


internal/storage/bucket_handle.go
```diff
@@ -70,16 +70,34 @@ func (bh *bucketHandle) NewReader(
 func (b *bucketHandle) DeleteObject(ctx context.Context, req *gcs.DeleteObjectRequest) error {
 	obj := b.bucket.Object(req.Name)
 
-	// Switching to the requested generation of the object.
-	if req.Generation != 0 {
-		obj = obj.Generation(req.Generation)
-	}
+	// Switching to the requested generation of the object. By default, generation
+	// is 0 which signifies the latest generation.
```
vadlakondaswetha, line 74: Add a comment that even if we dont set generation it will delete live object. With explicitly setting generation=0 we are satisfying idempotency condition.

internal/storage/bucket_handle_test.go
```diff
@@ -187,7 +187,19 @@ func (t *BucketHandleTest) TestDeleteObjectMethodWithMissingObject() {
 			MetaGenerationPrecondition: nil,
 		})
 
-	AssertEq("storage: object doesn't exist", err.Error())
+	AssertEq("gcs.NotFoundError: storage: object doesn't exist", err.Error())
+}
+
+func (t *BucketHandleTest) TestDeleteObjectMethodWithMissingGeneration() {
```
vadlakondaswetha, line 193: add a test case for explicitly passing generation as zero.

vadlakondaswetha (approved):

internal/storage/bucket_handle_test.go
sethiay: I thought of it but the fake storage serve doesn't respect that: https://github.com/GoogleCloudPlatform/gcsfuse/blob/bb041a9ee55fd338fe134b49a9fdfd2a92852a7c/internal/storage/bucket_handle_test.go#L40. 

But I agree that we should still add it. Please let me know if you think otherwise.

vadlakondaswetha: so missing generation is not a valid test case then?

sethiay: Missing generation as a unit test is valid to ensure that the DeleteObject doesn't fail with missing generation (and it not only specific to this change). 
From an idempotency check point of view, it is not valid.

internal/storage/bucket_handle_test.go
```diff
@@ -187,7 +187,19 @@ func (t *BucketHandleTest) TestDeleteObjectMethodWithMissingObject() {
...
+
+	err := t.bucketHandle.DeleteObject(context.Background(),
+		&gcs.DeleteObjectRequest{
+			Name: TestObjectName,
+			//Generation:                 1234, // Missing generation
```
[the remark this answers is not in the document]
sethiay: removed.

internal/storage/bucket_handle.go
sethiay: added.

vadlakondaswetha: Add one more line saying explicitly setting generation as zero to satisfy idempotency condition.

vadlakondaswetha (approved):

PR #1001, opened 2023-03-08
Fix list in garbage collection

sethiay:
**Issue:** Before deleting the stale objects, garbage collector first list all the objects with configured prefix (.gcsfuse_temp/) and it is [not passing](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/gcsutil/list_prefix.go#L32) 'MaxResults' in [ListObjectsRequest](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/gcsx/garbage_collect.go#L41). And the ListObject method of Go storage client library [lists](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/bucket_handle.go#L225) number of objects <= MaxResults which is 0 by default. So, no objects are listed.

**Fix** Modify the condition to list all objects when MaxResults is not passed (or when 0 is passed). This behaviour is similar to that in jacobsa/gcloud client librray (see [this condition](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L164)). Note: This fix is in the ListObject method to return list objects when MaxResults is 0.

Testing
Manual:
1. Increased the garbage collection frequency to remove stale objects every 10 secs and created some objects manually that starts with configured prefix: It deleted the stale objects as expected.

Unit tests:
1. Added test cases for ListObject method in bucket_handle_test.go to test when MaxResults is missing or 0.
2. Ran all unit tests in gcsfuse.

Integration tests:
1. Ran integration tests present in tools/integration_tests/implicit-dirs

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -221,8 +221,11 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 	// Iterating through all the objects in the bucket and one by one adding them to the list.
 	for {
 		var attrs *storage.ObjectAttrs
-		// itr.next returns all the objects present in the bucket. Hence adding a check to break after required number of objects are returned.
-		if len(list.Objects) == req.MaxResults {
+		// itr.next returns all the objects present in the bucket. Hence adding a
+		// check to break after required number of objects are returned.
+		// If req.MaxResults is 0, then wait till iterator is done. This is similar
+		// to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L164
+		if req.MaxResults != 0 && (len(list.Objects) == req.MaxResults) {
 			break
 		}
 		attrs, err = itr.Next()
--- a/internal/storage/bucket_handle_test.go
+++ b/internal/storage/bucket_handle_test.go
@@ -417,19 +417,30 @@ func (t *BucketHandleTest) TestListObjectMethodForMaxResult() {
 }
 
 func (t *BucketHandleTest) TestListObjectMethodWithMissingMaxResult() {
-	// We have 4 objects in the fakestorage.
-	fourObjWithoutMaxResults, err := t.bucketHandle.ListObjects(context.Background(),
+	// Validate that ee have 4 objects in fakeserver
+	fourObjWithMaxResults, err := t.bucketHandle.ListObjects(context.Background(),
 		&gcs.ListObjectsRequest{
 			Prefix:                   "",
 			Delimiter:                "",
 			IncludeTrailingDelimiter: true,
 			ContinuationToken:        "",
-			//MaxResults:               2, // Missing MaxResults
-			ProjectionVal: 0,
+			MaxResults:               100,
+			ProjectionVal:            0,
+		})
+	AssertEq(nil, err)
+	AssertEq(4, len(fourObjWithMaxResults.Objects))
+
+	fourObjWithoutMaxResults, err2 := t.bucketHandle.ListObjects(context.Background(),
+		&gcs.ListObjectsRequest{
+			Prefix:                   "",
+			Delimiter:                "",
+			IncludeTrailingDelimiter: true,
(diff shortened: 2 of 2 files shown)


internal/storage/bucket_handle.go
```diff
@@ -221,8 +221,11 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 	// Iterating through all the objects in the bucket and one by one adding them to the list.
 	for {
 		var attrs *storage.ObjectAttrs
-		// itr.next returns all the objects present in the bucket. Hence adding a check to break after required number of objects are returned.
-		if len(list.Objects) == req.MaxResults {
```
vadlakondaswetha, line 225: if req.MaxResults=0, why not set it as req.MaxResults=Integer.Max or something. It will avoid this extra condition in loop

internal/storage/bucket_handle_test.go
```diff
@@ -407,13 +408,99 @@ func (t *BucketHandleTest) TestListObjectMethodForMaxResult() {
...
+			IncludeTrailingDelimiter: true,
+			ContinuationToken:        "",
+			//MaxResults:               2, // Missing MaxResults
+			ProjectionVal: 0,
+		})
+
+	// Validate that 4 objects are listed when MaxResults = 100, because there
+	// are total of 4 objects in fakebucket.
+	AssertEq(nil, err)
+	AssertEq(4, len(fourObjWithMaxResults.Objects))
```
vadlakondaswetha, line 444: if this is validate the data before invoking the test please move it before test invocation.  why not just validate the data you received in previous call vs test invocation call directly.

internal/storage/bucket_handle_test.go
```diff
@@ -407,13 +408,99 @@ func (t *BucketHandleTest) TestListObjectMethodForMaxResult() {
...
+			Delimiter:                "",
+			IncludeTrailingDelimiter: true,
+			ContinuationToken:        "",
+			MaxResults:               0,
+			ProjectionVal:            0,
+		})
+
+	// Validate that 4 objects are listed when MaxResults = 100, because there
+	// are total of 4 objects in fakebucket.
+	AssertEq(nil, err)
```
vadlakondaswetha, line 485: same here also

internal/storage/bucket_handle_test.go
sethiay: Removed the first invocation and validation block and directly validating the test results with the expected data.

internal/storage/bucket_handle_test.go
sethiay: Removed the first invocation and validation block and directly validating the test results with the expected data.

internal/storage/bucket_handle.go
vadlakondaswetha: mybad. i thought 225 line is not removed. LG

internal/storage/bucket_handle_test.go
vadlakondaswetha: we need to check if there are 4 results before the test invocation.

internal/storage/bucket_handle_test.go
```diff
@@ -407,13 +408,59 @@ func (t *BucketHandleTest) TestListObjectMethodForMaxResult() {
...
 
+func (t *BucketHandleTest) TestListObjectMethodWithMissingMaxResult() {
+	// We have 4 objects in the fakestorage.
+	fourObjWithoutMaxResults, err := t.bucketHandle.ListObjects(context.Background(),
+		&gcs.ListObjectsRequest{
+			Prefix:                   "",
+			Delimiter:                "",
+			IncludeTrailingDelimiter: true,
+			ContinuationToken:        "",
+			//MaxResults:               2, // Missing MaxResults
```
vadlakondaswetha, line 427: we dont add lines and then comment it out. the testcase name itselft tell max result is missing.

sethiay: removed.

vadlakondaswetha (approved):

PR #1005, opened 2023-03-10
Print Panic message when mounting crash

Tulsishah:
### Description
Added handlepanic method in error_mapping to catch panic and get panic message in log file. In case logfile is not mentioned , it will logged in syslogs.

### Testing
1. Manual - Tested manually by giving panic message
2. Unit Test - N/A
3. IntegrationTes - N/A


internal/fs/wrappers/error_mapping.go
```diff
@@ -90,7 +90,17 @@ type errorMapping struct {
...
+func (em *errorMapping) handlePanic() {
+	// detect if panic occurred or not
+	a := recover()
+	if a != nil {
+		em.logger.Fatal("Panic: ", a)
+	}
+}
+
 func (em *errorMapping) mapError(op string, err error) error {
+	defer em.handlePanic()
```
vadlakondaswetha, line 102: why do you need it in mapError?

Tulsishah: fixed

vadlakondaswetha (approved):

PR #1006, opened 2023-03-13
[Don't Review] Printing logs in logfile

Tulsishah:
Not working in case of syslog. verified with @raj-prince


internal/logger/logger.go
```diff
@@ -177,3 +178,21 @@ func (f *loggerFactory) writer(level string) io.Writer {
 		return os.Stdout
 	}
 }
+
+func ErrorLogs(filename string) (writer io.Writer, err error) {
```
vadlakondaswetha, line 182: why are we doing seperate initialization of logs and not using the exisiting initialization?

Tulsishah: If we return io.writer form [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/logger/logger.go#L45) it may need major code changes that'w why prince suggested this way.

PR #1009, opened 2023-03-14
StatObject and ListObject call return input/output error in case of permission denied error.

Tulsishah:
### Description

Before, due to error formation in ListObject and StatObject methods, Go-Client-Library returned input/output error instead of exact error. I fixed it by replacing the original error.
Caught when I was testing permissions on the bucket. So when bucket does not have storage.object.get permission....
1. Go-Storage-Client-Library -  Input/Output Error
2. Jacobsa/gcloud - Permission denied error

As of Go 1.13 (or earlier if you use [golang.org/x/xerrors](https://godoc.org/golang.org/x/xerrors)), you can use the %w verb, only for error values, which wraps the error such that it can later be unwrapped with [errors.Unwrap](https://golang.org/pkg/errors/#Unwrap) ~ StackOverFlow

### Testing
1. Manual - Tested manually (giving Permission denied error)
2. UnitTest - N/A
4. IntegrationTest - N/A


sethiay (commented):
Please verify that unit tests are running fine with this change.

Tulsishah:
> Please verify that unit tests are running fine with this change.

Yes updated

internal/storage/bucket_handle.go
```diff
@@ -246,7 +245,6 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 			break
 		}
 		if err != nil {
-			err = fmt.Errorf("Error in iterating through objects: %v", err)
```
sethiay, line 249: Also, I see this error is given in other methods also like CopyObject. Can we change them also to keep it consistent?

sethiay (approved):
Looks good.

PR #1010, opened 2023-03-15
Crash during mounting

Tulsishah:
### Description
Background, with logfile - Panic message printed in Logfile
Background, without logfile - Panic message printed in Syslog
Foreground, with logfile - Panic message printed in Logfile
Foreground, without logfile - Panic message printed on Console

### Testing
1. Manual - Tested manually by putting panic message before mounting
2. UnitTest - N/A
3. IntegrationTest - N/A

### DIFF
--- a/main.go
+++ b/main.go
@@ -423,7 +423,7 @@ func run() (err error) {
 }
 
 func handlePanic() {
-	// detect if panic occurred or not
+	// Detect if panic happens in main go routine.
 	a := recover()
 	if a != nil {
 		logger.Info("Panic: ", a)
--- a/main.go
+++ b/main.go
@@ -176,10 +176,10 @@ func handlePanic() {
 }
 
 func mountWithArgs(
-		bucketName string,
-		mountPoint string,
-		flags *flagStorage,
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -422,7 +422,18 @@ func run() (err error) {
 	return
 }
 
+func handlePanic() {
+	// detect if panic occurred or not
+	a := recover()
+	if a != nil {
+		logger.Info("Panic: ", a)
+		os.Exit(1)
```
raj-prince, line 430: What happens if we don't keep os.Exit(1)?

Tulsishah: Nothing, but I thought exit status 1 would show an error, as I found it in many places. It Fixed.

main.go
```diff
@@ -422,7 +422,17 @@ func run() (err error) {
 	return
 }
 
+func handlePanic() {
+	// detect if panic occurred or not
```
raj-prince, line 426: Please fix the comment - "Detect if panic happens in main go routine."

raj-prince (approved):

main.go
Tulsishah: Fixed

Issue #1014, opened 2023-03-20
mount_gcsfuse not respecting bool values when passed to bool flags e.g. enable_storage_client_library


sethiay:
**Description:**
We add entries to the /etc/fstab file using mount_gcsfuse however it doesn't look as though that handles being able to set the value =false to disable it

[gcsfuse/tools/mount_gcsfuse/main.go](https://github.com/GoogleCloudPlatform/gcsfuse/blob/13d73e27d22a51ba5d71286f745ef32e2964cac8/tools/mount_gcsfuse/main.go#L90)

Line 90 in [13d73e2](https://github.com/GoogleCloudPlatform/gcsfuse/commit/13d73e27d22a51ba5d71286f745ef32e2964cac8)

 args = append(args, "--"+strings.Replace(name, "_", "-", -1)) 
I tried

    mount_line = ('%s /sftp/%s/upload gcsfuse rw,' '_netdev,allow_other,uid=%u,gid=%u,enable_storage_client_library=false' % (fuse_bucket, account['user'], uid, gid))

which just ignored the flag

    mount_line = ('%s /sftp/%s/upload gcsfuse rw,' '_netdev,allow_other,uid=%u,gid=%u,--enable-storage-client-library=false' % (fuse_bucket, account['user'], uid, gid))

and this didn't work, it add -o --enable-storage-client-library=false

Am I doing something wrong or is it an issue ? Any advice will be much appreciated

**Additional Information:**
This issue is reported by @WobblyRobbly [here](https://github.com/GoogleCloudPlatform/gcsfuse/issues/982#issuecomment-1462945792).

vadlakondaswetha:
Boolean flags are set as true by default and actual values are not considered.

https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/mount_gcsfuse/main.go#L89

PR #1017, opened 2023-03-20
Fixing: mount_gcsfuse not respecting bool values when passed to bool flags e.g. enable_storage_client_library

Tulsishah:
### Description 
Followed by this [issue](https://github.com/GoogleCloudPlatform/gcsfuse/issues/1014) 
Pass the value if explicitly define in flag, or pass it as  it is

### Testing
1. Manual - tested manually on VM
2. UnitTest - N/A
3. IntegrationTest - N/A

### DIFF
--- a/tools/mount_gcsfuse/main.go
+++ b/tools/mount_gcsfuse/main.go
@@ -81,12 +81,15 @@ func makeGcsfuseArgs(
 
 		// Special case: support mount-like formatting for gcsfuse bool flags.
 		case "implicit_dirs",
-			"foreground",
 			"experimental_local_file_cache",
 			"enable_storage_client_library",
 			"reuse_token_from_url",
 			"enable_nonexistent_type_cache":
-			args = append(args, "--"+strings.Replace(name, "_", "-", -1))
+			if value == "" {
+				args = append(args, "--"+strings.Replace(name, "_", "-", -1))
+			} else {
+				args = append(args, "--"+strings.Replace(name, "_", "-", -1)+"="+value)
+			}
 
 		// Special case: support mount-like formatting for gcsfuse string flags.
 		case "dir_mode",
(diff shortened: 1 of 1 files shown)


tools/mount_gcsfuse/main.go
```diff
@@ -121,7 +124,8 @@ func makeGcsfuseArgs(
 			"debug_gcs",
 			"debug_http",
 			"debug_invariants",
-			"debug_mutex":
+			"debug_mutex",
+			"foreground":
```
Tulsishah, line 128: putting foreground in this case because, when I passed foreground=false in argument, VM was crashing.

tools/mount_gcsfuse/main.go
```diff
@@ -81,12 +81,15 @@ func makeGcsfuseArgs(
 
 		// Special case: support mount-like formatting for gcsfuse bool flags.
 		case "implicit_dirs",
-			"foreground",
 			"experimental_local_file_cache",
 			"enable_storage_client_library",
 			"reuse_token_from_url",
 			"enable_nonexistent_type_cache":
-			args = append(args, "--"+strings.Replace(name, "_", "-", -1))
+			if value == "" {
```
Tulsishah, line 88: when we will not pass the value, it will take default value(true), when we pass some value it will pass that value.

sethiay: Can we do something like:
// value should be true in case '--flag' & '--flag=true'
if value != "false" {
   value = "true"
}
args = append(args, "--"+strings.Replace(name, "_", "-", -1)+"="+value)

tools/mount_gcsfuse/main.go
sethiay: As discussed offline, we should look into why VM is crashing when foreground is put in the above case.

Tulsishah: checked on another VM. It was working fine. May be it was VM issue

tools/mount_gcsfuse/main.go
Tulsishah: Fixed

sethiay (approved):

PR #1021, opened 2023-03-27
Adding -o ro in gcsfuse help

Tulsishah:
### Description
Adding -o ro in gcsfuse help
### Link to the issue in case of a bug fix
https://b.corp.google.com/issues/265181889
### Testing
N/A

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -86,7 +86,7 @@ func newApp() (app *cli.App) {
 
 			cli.StringSliceFlag{
 				Name:  "o",
-				Usage: "Additional system-specific mount options. Be careful!",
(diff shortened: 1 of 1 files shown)


flags.go
```diff
@@ -86,7 +86,7 @@ func newApp() (app *cli.App) {
 
 			cli.StringSliceFlag{
 				Name:  "o",
-				Usage: "Additional system-specific mount options. Be careful!",
+				Usage: "Additional system-specific mount options. Multiple options can be passed as comma separated. For readonly, use add --o ro",
```
vadlakondaswetha, line 89: use and add are same things. just mention
use --o ro

vadlakondaswetha (approved):

flags.go
Tulsishah: fixed

PR #1022, opened 2023-03-27
Removing TmpDir from IntegrationTests

Tulsishah:
### Description

Earlier - we were creating tmpDir in mountedDir, when [CreateTempFile](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/setup/setup.go#L114) called. it was creating tmpFile in tmpDir.
Eg . /mnt/tmpDir/tmpFile345678

Because of [the creation of tmpDir](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/setup/setup.go#L202), for the read-only mount, it was throwing an error. 

I removed tmpDir creation now CreateTempFile will create file in mntDir only.
Eg. mnt/tmpFile345678

### Testing
1. Manual - tested manually ( all the tests are working fine in both the cases - testbucket and mountDir )
2. UnitTest - N/A
3. IntegrationTest - N/A


tools/integration_tests/setup/setup.go
```diff
@@ -57,14 +56,6 @@ func TestDir() string {
 	return testDir
 }
 
-func SetTmpDir(tmpDirValue string) {
```
sethiay, line 60: I see this function is deleted but its usages are not. Is it the case that this setter is not used anywhere?

Tulsishah: There are no uses remains for this function

PR #1032, opened 2023-03-29
Performance doc

Tulsishah:
### Description
check here for [view](https://github.com/GoogleCloudPlatform/gcsfuse/blob/0aae7227cc04d640bd813ea59ecb8e5f33cfbb53/docs/performance.md)

### Testing
N/A


docs/performance.md
```diff
@@ -0,0 +1,48 @@
...
+
+All rate limiting is approximate, and is performed over an 8-hour window. By default, there are no limits applied.
+
+# Upload procedure control
+
+An upload procedure is implemented as a retry loop with exponential backoff for failed requests to the Cloud Storage backend. Once the backoff duration exceeds this limit, the retry stops. Flag ```--max-retry-sleep``` controls such behavior. The default is 1 minute. A value of 0 disables retries.
+
+# Cloud Storage round trips
+
+By default, Cloud Storage FUSE uses two forms of caching to save round trips to Cloud Storage, at the cost of consistency guarantees. These caching behaviors can be controlled with the flags ```--stat-cache-capacity```, ```--stat-cache-ttl``` and ```--type-cache-ttl```. See semantics.md for more information.
```
raj-prince, line 44: It would be good, if you can add the link for semantics.md you referred.

raj-prince (approved):
One minor comment, please incorporate.

docs/performance.md
Tulsishah: fixed

PR #1036, opened 2023-03-30
Fix typos in upgrade instructions

zchenyu:


Tulsishah (approved):
Thanks for correction @zchenyu .

PR #1044, opened 2023-04-04
Setup for readonly integration tests

Tulsishah:
### Description
1. Setup to run readonly integration tests.
2. Added cleanup function to clean the bucket
3. Added SetupBucket method to set the bucket.
4. Make public methods private if not needed.
### Testing
1. Manual - All the existing integration tests are working fine for both the cases ( testBucket and mountedDir)
5. UnitTes - N/A
6. IntegrationTest - N/A


tools/integration_tests/readonly/readonly_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"flag"
+	"log"
+	"os"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/setup"
+)
+
+func TestMain(m *testing.M) {
+	flag.Parse()
```
vadlakondaswetha, line 28: lets try to avoid putting same code at every place. We should reuse the code as much as possible.
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/implicitdir/implicitdir_test.go#L27

create a seperate function in setup with all this code and that takes flags as argument.

Tulsishah: fixed

PR #1058, opened 2023-04-12
Skip integration test

Tulsishah:
### Description
- Skip Integration tests execution in git CI tests as we can't authenticate bucket permission.
- To run the integration tests Pass --integrationTest flag.
- Returning os.Exit(1) when testbucket or mountDirectory both flags passed or none of them are passed.

### Testing details
1. Manual - Integration tests are not executing in git actions.
2. Unit tests - N/A
3. Integration tests - N/A


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         go install ./tools/build_gcsfuse
         build_gcsfuse . /tmp ${GITHUB_SHA}
     - name: Test
-      run: go test -p 1 -count 1 -v -cover ./...
+      run: go test -p 1 -count 1 -v -cover ./... $(go list ./... | grep -v /tools/integration_tests/)
```
vadlakondaswetha, line 33: can you add comments what this command does.

Tulsishah: fixed

vadlakondaswetha: did you look at this: https://stackoverflow.com/questions/25965584/separating-unit-tests-and-integration-tests-in-go

Tulsishah: Introduce  --integrationTest flag to run the tests.

tools/integration_tests/setup/setup.go
```diff
@@ -211,12 +212,17 @@ func ExecuteTestForFlags(flags [][]string, m *testing.M) (successCode int) {
 func RunTests(flags [][]string, m *testing.M) {
 	flag.Parse()
 
+	if !*integrationTest {
```
sethiay, line 215: Given that integration tests will only run if --integrationTest flag is passed after this change, we need to let CSI driver developer know about this change as these tests are used in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/examples/gcsfuse-e2e-test/pod.yaml#L49 ? 

(Ideally, we should let them know about these changes and they should make changes in CSI driver repo but since this is very small change, so I would suggest to simply raise PR with this change in their repo and mention there why that change is required)

Tulsishah: Should I do it once this PR gets merged?

sethiay: Yes

Tulsishah: Sure. Will do that once this merged

vadlakondaswetha (approved):

sethiay (approved):

Issue #1054, opened 2023-04-11
Large directory listing returns wrong results with --enable-storage-client-library=true


MattIrv:
**Describe the bug**
Please provide a clear description of what you were trying to achieve along with the details of the flags that you passed.

I'm listing a bucket that has a lot of files in it. It returns incorrect results. When I try it with `--enable-storage-client-library=false` I get the right results.

```
$ gsutil ls gs://<bucket> | wc -l
54979
```

Using gcsfuse:
```
$ ls -f mountdir | wc -l
49979
```

**To Collect more Debug logs**
1. Please make sure you have no other security, monitoring, background processes which can offend the FUSE process running. Possibly reproduce under a fresh/clean installation.
2. Please rerun with --debug_fuse --debug_fs --debug_gcs --debug_http --foreground as additional flags to enable debug logs.
3. Monitor the logs and please capture screenshots or copy the relevant logs to a file (can use --log-format and --log-file as well).
4. Attach the screenshot or the logs file to the bug report here.
5. If you're using gcsfuse with any other library/tool/process please list out the steps you took to reproduce the issue.

```
$ gcsfuse -debug_fuse --debug_fs --debug_gcs --debug_http --foreground <bucket> mountdir
Start gcsfuse/0.42.3 (Go version go1.19.5) for app "" using mount point: /home/mirvine/mountdir
Opening GCS connection...
Creating a mount at "/home/mirvine/mountdir"
Creating a new server...
Set up root directory for bucket <bucket>
gcs: Req              0x0: <- ListObjects("")
gcs: Req              0x0: -> ListObjects("") (331.22039ms): OK
Mounting file system "<bucket>"...
fuse_debug: Beginning the mounting kickoff process
fuse_debug: Parsing fuse file descriptor
fuse_debug: Preparing for direct mounting
fuse_debug: Successfully opened the /dev/fuse in blocking mode
fuse_debug: Starting the unix mounting
fuse_debug: Directmount failed. Trying fallback.
fuse_debug: Creating a socket pair
fuse_debug: Creating files to wrap the sockets
fuse_debug: Starting fusermount/os mount
fuse_debug: Wrapping socket pair in a connection
fuse_debug: Checking that we have a unix domain socket
fuse_debug: Read a message from socket
fuse_debug: Successfully read the socket message.
fuse_debug: Converting FD into os.File
fuse_debug: Completed the mounting kickoff process
fuse_debug: Creating a connection object
[...]

PR #1060, opened 2023-04-13
List obj fix

sethiay:
### Description
Fix bug in list object method due to which some files are not listed if number of files in a directory are more than 5000 and contains one or more nested directory.

#### More details:

- The listing of objects (includes both files and directories as directories are also object in GCS) happens one by one using the next function of [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/bucket_handle.go#L263) iterator. 
- The next function returns object, next page token and other attributes. E.g. if the listing is happening for 1st page, then the next function returns page token of 2nd page. 

Before fix: 
- This [for](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/bucket_handle.go#L254) loop runs until the number of files [are not equal](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/bucket_handle.go#L260) to MaxResults (5000) which is problematic. E.g. if we are listing for 1st page which consists of 5000 objects including 4000 files and 1000 directories, then this for loop will [not stop](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/bucket_handle.go#L278) after listing 4000 files and 1000 directories but will go on till it lists 5000 files, which will go into 2nd page. Now assume 2nd page consists of 5000 files and 0 directories.  At the end of listing 5000 files (4000 from 1st page and 1000 from 2nd page), the next page token will be of 3rd page. So the next ListObject method call from downstream methods will start from 3rd page and hence skipping 4000 files from 2nd page.

After fix: 
- This for loop will run until all the objects in requested page are fetched using (itr.Next() method). So, taking the same example as before, the first call to ListObject will list objects only from 1st page and return token of 2nd page. Hence will not miss 4000 files/objects of 2nd page.

This bug was identified in this issue: https://github.com/GoogleCloudPlatform/gcsfuse/issues/1054

### Testing details
1. Manual
   a. Verified that correct number of files are listed for the exact issue specified by customer: This bug was identified in this issue: https://github.com/GoogleCloudPlatform/gcsfuse/issues/1054
   b. Reproduced the issue for lesser number of files i.e. 5500 and verified the fix against that many files as well.
   c. Also ran [listing benchmarks](https://github.com/GoogleCloudPlatform/gcsfuse/tree/master/perfmetrics/scripts/ls_metrics) and got similar performance.
3. Unit tests
   a. Can't add unit test for testing this functionality specifically because the fake storage doesn't support ContinuationToken.
5. Integration tests

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -276,7 +276,7 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 		}
 
 		// itr.next returns all the objects present in the bucket. Hence adding a
-		// check to break after required number of objects are returned.
+		// check to break after iterating over the current page.
 		// If req.MaxResults is 0, then wait till iterator is done. This is similar
 		// to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L164
 		if req.MaxResults != 0 && (pi.Remaining() == 0) {
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -44,8 +44,8 @@ func (bh *bucketHandle) Name() string {
 }
 
 func (bh *bucketHandle) NewReader(
-	ctx context.Context,
-	req *gcs.ReadObjectRequest) (io.ReadCloser, error) {
+		ctx context.Context,
+		req *gcs.ReadObjectRequest) (io.ReadCloser, error) {
 	// Initialising the starting offset and the length to be read by the reader.
 	start := int64(0)
 	length := int64(-1)
@@ -250,17 +250,10 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 	pi.Token = req.ContinuationToken
 	var list gcs.Listing
 
-	var resultsCount int = 0
 	// Iterating through all the objects in the bucket and one by one adding them to the list.
 	for {
 		var attrs *storage.ObjectAttrs
-		// itr.next returns all the objects present in the bucket. Hence adding a
-		// check to break after required number of objects are returned.
-		// If req.MaxResults is 0, then wait till iterator is done. This is similar
-		// to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L164
-		if req.MaxResults != 0 && (resultsCount == req.MaxResults) {
-			break
-		}
+
 		attrs, err = itr.Next()
 		if err == iterator.Done {
 			err = nil
@@ -281,7 +274,14 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 			currObject := storageutil.ObjectAttrsToBucketObject(attrs)
 			list.Objects = append(list.Objects, currObject)
 		}
-		resultsCount++
+
+		// itr.next returns all the objects present in the bucket. Hence adding a
+		// check to break after required number of objects are returned.
+		// If req.MaxResults is 0, then wait till iterator is done. This is similar
+		// to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L164
+		if req.MaxResults != 0 && (pi.Remaining() == 0) {
+			break
+		}
 	}
 
 	list.ContinuationToken = itr.PageInfo().Token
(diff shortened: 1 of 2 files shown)


internal/storage/bucket_handle_test.go
```diff
@@ -365,6 +365,8 @@ func (t *BucketHandleTest) TestGetProjectValueWhenGcloudProjectionIsDefault() {
 }
 
 func (t *BucketHandleTest) TestListObjectMethodWithPrefixObjectExist() {
+	// Note: ContinuationToken doesn't work with fake storage and hence we can't
+	// add unit tests to test its functionality.
```
vadlakondaswetha, line 369: why can't we add a test case to list 10 objects when bucket has 5 files + 5 folders + 5 files and how the data will be returned. when we set the max results to 10 , we should get 5 files + 5 folders, earlier it must be returning all.

internal/storage/bucket_handle.go
```diff
@@ -250,14 +250,15 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
...
 
+	var resultsCount int = 0
 	// Iterating through all the objects in the bucket and one by one adding them to the list.
 	for {
 		var attrs *storage.ObjectAttrs
 		// itr.next returns all the objects present in the bucket. Hence adding a
 		// check to break after required number of objects are returned.
 		// If req.MaxResults is 0, then wait till iterator is done. This is similar
 		// to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L164
-		if req.MaxResults != 0 && (len(list.Objects) == req.MaxResults) {
```
raj-prince, line 260: Do we need to maintain this extra variable?
Can't we change the check (len(list.objects) + len(list.CollapsedRuns)) == req.MaxResults?

internal/storage/bucket_handle_test.go
sethiay: Thanks for this idea. I tried this and it worked. I will add a test case for this.

internal/storage/bucket_handle.go
sethiay: I don't have a strong preference on this. 
Computation wise, I think both the options are similar ( len(list.objects) + len(list.CollapsedRuns) might be little less optimised because it has extra step of fetching length )
Memory wise also, I believe both the cases are similar ? One has defined variable and other will take space to store  len(list.objects) + len(list.CollapsedRuns) in stack.
Readability wise, I prefer resultsCount. 

Please let me know if I am missing any point here.

internal/storage/bucket_handle_test.go
sethiay: With the new fix, this is not possible because the fake storage will always return all the objects and the new depends on the storage to return correct number of objects according to the value of maxResults.

internal/storage/bucket_handle.go
```diff
@@ -280,6 +274,14 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 			currObject := storageutil.ObjectAttrsToBucketObject(attrs)
 			list.Objects = append(list.Objects, currObject)
 		}
+
+		// itr.next returns all the objects present in the bucket. Hence adding a
+		// check to break after required number of objects are returned.
```
raj-prince, line 279: We can change the comment here -

"after required number of objects are returned" ->  something like "after iterating over the current page"

internal/storage/bucket_handle_test.go
```diff
@@ -473,12 +473,17 @@ func (t *BucketHandleTest) TestListObjectMethodForMaxResult() {
 	AssertEq(TestObjectName, fourObj.Objects[3].Name)
 	AssertEq(nil, fourObj.CollapsedRuns)
 
-	// Validate that 2 objects are listed when MaxResults is passed 2.
+	// Note: Thw behavior is different in real GCS storage JSON API. In real API,
```
raj-prince, line 476: nit: The

raj-prince (approved):

internal/storage/bucket_handle.go
```diff
@@ -280,6 +274,14 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 			currObject := storageutil.ObjectAttrsToBucketObject(attrs)
 			list.Objects = append(list.Objects, currObject)
 		}
+
+		// itr.next returns all the objects present in the bucket. Hence adding a
+		// check to break after iterating over the current page.
+		// If req.MaxResults is 0, then wait till iterator is done. This is similar
+		// to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/jacobsa/gcloud/gcs/bucket.go#L164
+		if req.MaxResults != 0 && (pi.Remaining() == 0) {
```
vadlakondaswetha, line 282: can you also add a comment on what pi.remaining and why is it added at the end and not in the begining of the loop

vadlakondaswetha (approved):

Issue #1040, opened 2023-04-01
http connections are not properly closed without a http.Client Timeout property


BogdanEAM:
**Describe the bug**
We are starting a Java applications from a gcsfuse mount folder.
The app has 139 lib jars and many more prop files and classes.

With the gcsfuse version v0.41.12 it starts fine. However with versions v0.42.2 and v0.42.3 it does not start.
The difference that is causing the issue between these version is the removal of "Timeout" property from the http.Client created in "internal\storage\storage_handle.go" method NewStorageHandle.

***Runtime flags***
used for all versions:
`-o noexec --dir-mode 700 --stackdriver-export-interval 60s --type-cache-ttl 20h -o ro -o nosuid -o nodev --file-mode 400 --enable-storage-client-library=true --implicit-dirs --enable-nonexistent-type-cache --stat-cache-ttl 20h --stat-cache-capacity 40960`

***Behaviour***

> with versions v0.42.2 and v0.42.3

The startup of the app always hits a dead stop when gcsfuse is reading a certain (always the same) file from the lib folder. With the _gcs flag we can see that it is always the same file.

By doing a netstat in parallel with the application startup we can see that the number of "ESTABLISHED" connections for the respective gcsfuse process grows until it reaches the number of 101 connections which is the default value for "max-conns-per-host".

****Workaround 1****
If we add the flag "--max-conns-per-host 300" the application start fine and the number of connections reported by netstat caps at around 115 connections.

****Workaround 2****
I have rebuild the code for v0.42.3 by reading the "Timeout" flag in "internal\storage\storage_handle.go" in the NewStorageHandle with the previous default value of 800 milliseconds.
The neststat in this case shows that  connections are not pilling up, their number increases and decreases while application starts and never reaches a number higher than 15 connections.

If I set the Timeout to 10 seconds the max number of connections observed was 35 and again the app starts fine.

```        // Custom http client for Go Client.
        httpClient := &http.Client{
                Transport: &oauth2.Transport{
                        Base:   transport,
                        Source: clientConfig.TokenSrc,
                },
                Timeout: 800 * time.Millisecond,
        }
```

> with versions v0.41.12

The application start fine.
Running netstat in parallel as described above shows that connections are not pilling up, their number increases and decreases while application starts and never reaches a number higher than 20 connections.

**To Collect more Debug logs**
[...]

PR #1061, opened 2023-04-13
Add http timeout

sethiay:
### Description
Add flag for http client timeout when mounted with Go storage client library enabled. Without timeout, if a user tries to open file handles more than value of --max-conns-per-host, then the GCSFuse mounts get stuck. Introducing timeout solves this issue as mentioned by customer in the issue and also verified using a python script which tries to read files more than value of --max-conns-per-host. Since, keeping timeout can cause [other errors](https://github.com/GoogleCloudPlatform/gcsfuse/pull/922) at the time of writing files, so by default we will not add any timeout.

https://github.com/GoogleCloudPlatform/gcsfuse/issues/1040

### Testing details
1. Manual
    - Ran python script reading more than --max-conns-per-host files (i.e. 101) with timeout and without timeout. Without timeout, the script and gcsfuse hangs and with timeout, it succeeds.
    - Ran fio load test with 100mb and 1gb files using 512 and 100 number of processes respectively. Got ~52-5300 MiB/sec in both cases with and without these changes without timeout.
    -  Ran fio load test with 100mb and 1gb files using 512 and 100 number of processes respectively. Got ~50-5100 MiB/sec in both cases with these changes with timeout of 800ms. Some decrease in bandwidth is expected due to overhead of closing and creating connections again and again.
2. Unit tests
    - Added unit test for flag.
    - NA for testing functionality.
3. Integration tests
   - TBD
4. Perf tests
   - Added execute-perf-test label to run fio perf tests.


docs/mounting.md
```diff
@@ -141,50 +141,51 @@ If the user mounting the Cloud Storage FUSE is not root then the ```allow_other`
...
-|--debug_fuse|Enable fuse-related debugging output.|
-|--debug_fs|Enable file system debugging output.|
-|--debug_gcs|Print GCS request and timing information.|
-|--debug_http| Dump HTTP requests and responses to/from GCS, doesn't work when --enable-storage-client-library flag is true.|
-|--debug_invariants|Panic when internal invariants are violated.|
-|--debug_mutex|Print debug messages when a mutex is held too long.|
-|--enable-storage-client-library|If true, will use go storage client library otherwise jacobsa/gcloud|
-|--help, -h|Show help|
-|--version, -v|Print version|
+| Option                                               | Description                                                                                                                                                                                                                                                                                                                                                                                  |
```
sethiay, line 144: For reviewer: After adding --http-client-timeout flag in the doc and saving the doc, these formatting changes happened automatically. You can open the readme doc in new tab and verify.

internal/storage/storage_handle.go
```diff
@@ -82,6 +83,7 @@ func NewStorageHandle(ctx context.Context, clientConfig StorageClientConfig) (sh
 			Base:   transport,
 			Source: clientConfig.TokenSrc,
 		},
+		Timeout: clientConfig.HttpClientTimeout,
```
vadlakondaswetha, line 86: do you want to add a test case which passes timeout to ensure nothing breaks.

https://github.com/GoogleCloudPlatform/gcsfuse/pull/934/files#diff-2d5c5e3691214ba273ac9050fc100dcfff9a5981d5844f74e653718f05ae10dd

sethiay: Sure, added

vadlakondaswetha (approved):

docs/mounting.md
raj-prince: Try editing by removing the last `|` delimiter and add `|` back in the last after the complete change. It worked for me.

Tulsishah (approved):
Please test for large file sizes (e.g., 100Mb or 1Gb) with a large number of threads(e.g., 512) also if we are getting input-output errors for write or not.
Also, fix lint and Linux tests before submitting.

internal/storage/storage_handle.go
```diff
@@ -77,11 +78,20 @@ func NewStorageHandle(ctx context.Context, clientConfig StorageClientConfig) (sh
...
-	httpClient := &http.Client{
-		Transport: &oauth2.Transport{
-			Base:   transport,
-			Source: clientConfig.TokenSrc,
-		},
+	oauth2Transport := oauth2.Transport{
+		Base:   transport,
+		Source: clientConfig.TokenSrc,
+	}
+	var httpClient *http.Client
```
raj-prince, line 85: As per the documentation, a time-out as 0 mean no timeout.

```Go
	// Timeout specifies a time limit for requests made by this
	// Client. The timeout includes connection time, any
	// redirects, and reading the response body. The timer remains
	// running after Get, Head, Post, or Do return and will
	// interrupt reading of the Response.Body.
	//
	// A Timeout of zero means no timeout.
	//
	// The Client cancels requests to the underlying Transport
	// as if the Request's Context ended.
	//
	// For compatibility, the Client will also use the deprecated
	// CancelRequest method on Transport if found. New
	// RoundTripper implementations should use the Request's Context
	// for cancellation instead of implementing CancelRequest.
	Timeout time.Duration
```

Can you please check if it works, in that case we don't need special check?

raj-prince (changes requested):
Please check.

internal/storage/storage_handle.go
sethiay: Yes, that worked. Made changes accordingly.

raj-prince (approved):

vadlakondaswetha (approved):

Tulsishah (approved):

PR #1063, opened 2023-04-14
Change maintainer from jacobsa to gcs-fuse-dev

sethiay:
### Description
Change maintainer from jacobsa to gcs-fuse-dev in package_gcsfuse

Internal

### Testing details: N/A
1. Manual
2. Unit tests
3. Integration tests

### DIFF
--- a/tools/package_gcsfuse/package.go
+++ b/tools/package_gcsfuse/package.go
@@ -37,7 +37,7 @@ func packageFpm(
 		"-v", version,
 		"-d", "fuse",
 		"--vendor", "",
-		"--maintainer", "Aaron Jacobs <jacobsa@google.com>",
+		"--maintainer", "GCSFuse dev <gcs-fuse-dev@google.com>",
 		"--url", "https://github.com/googlecloudplatform/gcsfuse",
 		"--description", "A user-space file system for interacting with Google Cloud Storage.",
 	)


tools/package_gcsfuse/package.go
```diff
@@ -37,7 +37,7 @@ func packageFpm(
 		"-v", version,
 		"-d", "fuse",
 		"--vendor", "",
-		"--maintainer", "Aaron Jacobs <jacobsa@google.com>",
+		"--maintainer", "GCSFuse dev <gcs-fuse-dev@google.com>",
```
sethiay, line 40: **For reviewer:** I couldn't find any usage of "package_gcsfuse". It would be helpful to mention its usage here if you know of any and if there exist any. We can add the same in the comments/internal docs.

vadlakondaswetha: I am not aware either. Did you check the readme/description deb/rmp packages or on cloud rapture.

vadlakondaswetha (approved):

tools/package_gcsfuse/package.go
sethiay: From reading the description, I understand that this is used for creating gcsfuse package. However, in our current release, we don't use code under package_gcsfuse. Rather we use code under package_gcsfuse_docker. Anyways, I will merge these changes.

PR #1064, opened 2023-04-17
Setup for readonly test

Tulsishah:
### Description
- Adding script to create objects in given bucket for readonly tests
     - Test/a.txt
     - Test1.txt
     - Test/b/b.txt
- Adding script to delete object after testing from bucket.
- return successCode from setup to perform delete operation and then perform os.Exit(successCode)
- Setting Test-bucket env variable to use it in script.
- Now we can use any bucket and no need to take care of separate cleanup.

### Testing details
1. Manual - All the tests are working fine for both the flags
2. Unit tests - N/A
3. Integration tests - N/A


tools/integration_tests/readonly/readonly_test.go
```diff
@@ -0,0 +1,54 @@
...
+	os.Setenv("TEST_BUCKET", setup.TestBucket())
+
+	// Create objects in bucket for testing.
+	cmd := exec.Command("/bin/bash", "create_objects.sh")
+	_, err := cmd.Output()
+	if err != nil {
+		panic(err)
+	}
+
+	successCode := setup.RunTests(flags, m)
```
vadlakondaswetha, line 41: os.RemoveAll(mntDir) in setup.go will it not throw an error?

tools/integration_tests/readonly/readonly_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/setup"
+)
+
+func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+
+	flags := [][]string{{"--o=ro", "--implicit-dirs=true"}}
+
+	// Set environment variable to use testBucket in creating objects.
+	os.Setenv("TEST_BUCKET", setup.TestBucket())
```
vadlakondaswetha, line 32: what is the testbucket value here?

tools/integration_tests/readonly/delete_objects.sh
```diff
@@ -0,0 +1,3 @@
+gsutil rm gs://$TEST_BUCKET/Test/a.txt
```
vadlakondaswetha, line 1: why not just delete everything from bucket

tools/integration_tests/readonly/create_objects.sh
```diff
@@ -0,0 +1,6 @@
+echo "This is from file a" >> a.txt
```
vadlakondaswetha, line 1: can you move this scripts to readonly/testdata/

tools/integration_tests/readonly/readonly_test.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	successCode := setup.RunTests(flags, m)
+
+	// Delete objects from bucket after testing.
+	cmd = exec.Command("/bin/bash", "delete_objects.sh")
+	_, err = cmd.Output()
+	if err != nil {
+		panic(err)
+	}
```
vadlakondaswetha, line 48: this has been coming up multiple times in your PRs. Always follow DRY (dont repeat yourself). if there are more 1 lines to be repeated create a seperate function for it.

vadlakondaswetha (changes requested):

tools/integration_tests/readonly/readonly_test.go
Tulsishah: what we will pass through --testBucket flag.

tools/integration_tests/readonly/delete_objects.sh
Tulsishah: I though we should remove the objects only which we have created.

tools/integration_tests/readonly/create_objects.sh
Tulsishah: fixed

tools/integration_tests/readonly/readonly_test.go
Tulsishah: Sorry, will keep this thing in mind. fixed

tools/integration_tests/readonly/readonly_test.go
Tulsishah: No it is not throwing any error.

tools/integration_tests/readonly/readonly_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/setup"
+)
+
+func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+
+	flags := [][]string{{"--o=ro", "--implicit-dirs=true"}}
```
sethiay, line 28: Please check if same tests can be run with --file-mode 444 --dir-mode 444 flags (without -o ro). If yes, we can include these flags as well.

tools/integration_tests/readonly/readonly_test.go
sethiay, line 31: Can we pass test bucket name as argument to create and delete scripts instead of setting and unsetting environment variables ?

vadlakondaswetha (approved):

sethiay (approved):

PR #1067, opened 2023-04-19
Readonly Integration Tests - Create, Delete, List

Tulsishah:
### Description
Integration tests 
1. CreateObject
    1. CreateFile - CreateFileInMainDir, CreateFileInSubDir
    2. CreateDir - CreateDirInMainDir, CreateDirInSubDir
2. DeleteObject
    1. DeleteFile -DeleteFileInMainDir, DeleteFileInSubDir
    2. DeleteDir - DeleteDirInMainDir, DeleteDirInSubDir
3.  ListObject
    1. ListObject from Main Dir
    2. ListObject from Subdirectory
 
If there are already some objects in the bucket, we will get incorrect results in ListTests. That's why first I add to clear a bucket.
Fixing  - When the bucket was empty, it threw commadException. Fix that by putting it if condition.

### Testing details
1. Manual  - N/A
5. Unit tests - N/A
6. Integration tests - all the tests are passing for any testbucket


tools/integration_tests/readonly/list_objects_test.go
```diff
@@ -0,0 +1,74 @@
...
+	obj, err := os.ReadDir(setup.MntDir())
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if len(obj) != NumberOfObjectsInTestBucket {
+		t.Errorf("The number of objects in the current directory doesn't match.")
+	}
+
+	if obj[0].Name() != "Test" && obj[0].IsDir() != true {
```
Tulsishah, line 42: For reviewer: Other properties were FileInfo -> which has property fileName and Type and FileMode which is fileType only, both are included in this two comparison that's why I didn't mention.

tools/integration_tests/readonly/create_and_delete_object_test.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"io/fs"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/setup"
+)
+
+func CreateFile(filePath string, fileName string, t *testing.T) {
```
sethiay, line 27: Given CreateFile is not used in other packages, so it should start with lowecase.

Tulsishah: fixed for all the functions

tools/integration_tests/readonly/create_and_delete_object_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"io/fs"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/setup"
+)
+
+func createFile(filePath string, t *testing.T) {
+	file, err := os.OpenFile(filePath, os.O_CREATE, setup.FilePermission_0600)
```
raj-prince, line 28: What happens if file already exist and trying to open it to append some data?

Shouldn't we add os.O_APPEND flag?

Tulsishah: Shouldn't we do that in write tests?

Tulsishah: I am about to adding more tests for read and write also

sethiay (approved):

tools/integration_tests/readonly/create_and_delete_object_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"io/fs"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/setup"
+)
+
+func createFile(filePath string, t *testing.T) {
```
raj-prince, line 27: Please rename "createFile" -> "ensureFileSystemLockedForFileCreation"

 This will take parent directory as input, and internally attempt to create a temp file and throws error in case of successful creation.

tools/integration_tests/readonly/create_and_delete_object_test.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func TestCreateDir(t *testing.T) {
+	dirPath := path.Join(setup.MntDir(), "test")
+	createDir(dirPath, t)
+}
+
+func TestCreateDirInSubDirectory(t *testing.T) {
+	dirPath := path.Join(setup.MntDir(), "Test", "test")
+	createDir(dirPath, t)
```
raj-prince, line 64: Here also, please renamed createDir -> "ensureFileSystemLockedForDirCreation".

tools/integration_tests/readonly/create_and_delete_object_test.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func deleteObjects(objPath string, t *testing.T) {
+	err := os.RemoveAll(objPath)
```
raj-prince, line 68: What happens if object already doesn't exist?

tools/integration_tests/readonly/create_and_delete_object_test.go
Tulsishah: fixed for all the functions

tools/integration_tests/readonly/create_and_delete_object_test.go
Tulsishah: fixed

tools/integration_tests/readonly/list_objects_test.go
```diff
@@ -0,0 +1,74 @@
...
```
raj-prince, line 42: Since we have a fixed hierarchy, we should keep these names in a variable. For, "Test", "Test1.txt".

tools/integration_tests/readonly/list_objects_test.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+func TestListObjectsInBucketSubDirectory(t *testing.T) {
+	// ** SubDirectory structure **
+	// testBucket/Test
+	// testBucket/Test/a.txt -- File
+	// testBucket/Test/b/    -- Dir
+
+	Dir := path.Join(setup.MntDir(), "Test")
```
raj-prince, line 57: Use that variable name, instead of value.

raj-prince (changes requested):
Added a few comments.

tools/integration_tests/readonly/list_objects_test.go
Tulsishah: fixed

tools/integration_tests/readonly/list_objects_test.go
```diff
@@ -0,0 +1,74 @@
...
```
raj-prince, line 57: Please use the variable instead of value.

raj-prince (approved):

tools/integration_tests/readonly/list_objects_test.go
Tulsishah: fixed

PR #1072, opened 2023-04-25
Add "(GPN: gcsfuse-<client-identifier>)" to useragent for partner data tracking.

ashmeenkaur:
### Description
Appended "(GPN: gcsfuse-<client-identifier>)" to useragent to enable partner data tracking. 
Removed redundant app-name/METADATA_IMAGE_TYPE from user agent if it has been appended to the partner name.

### Testing details
1. Manual: Verified the useragent received is of the format 'gcsfuse/x.xx.x (Go version gox.xx.x) (GPN: gcsfuse),gzip(gfe)' by querying the bigstore_weblog table
2. Unit tests: Updated
3. Integration tests: NA


Tulsishah (commented):
Are we supposed to test UserAgent [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/storage_handle_test.go#L120)? Please confirm.

main.go
```diff
@@ -76,7 +76,7 @@ func registerSIGINTHandler(mountPoint string) {
 }
 
 func getUserAgent(appName string) string {
-	return strings.TrimSpace(fmt.Sprintf("gcsfuse/%s %s %s", getVersion(), appName, os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE")))
+	return strings.TrimSpace(fmt.Sprintf("gcsfuse/%s %s %s", getVersion(), appName, os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE"))) + " (GPN: gcsfuse)"
```
vadlakondaswetha, line 79: can you add it as part of "gcsfuse/%s %s %s GPN: gcsfuse" instead of doing another append

ashmeenkaur: Tried doing this initially but TestGetUserAgentWhenMetadataImageTypeEnvVarIsNotSet was failing because of an extra space for empty environment variable. I've updated the logic to remove any redundant spaces.

internal/storage/storage_handle_test.go
```diff
@@ -119,7 +119,6 @@ func (t *StorageHandleTest) TestNewStorageHandleWithZeroMaxConnsPerHost() {
 
 func (t *StorageHandleTest) TestNewStorageHandleWhenUserAgentIsSet() {
 	sc := getDefaultStorageClientConfig()
-	sc.UserAgent = "gcsfuse/unknown (Go version go1.20-pre3 cl/474093167 +a813be86df)"
```
Tulsishah, line 122: Sorry why did you remove this, and if this is necessary then I think this test case will become useless.

ashmeenkaur: getDefaultStorageClientConfig() method sets the same user agent. I've removed the test as it was redundant.

main.go
```diff
@@ -76,7 +76,14 @@ func registerSIGINTHandler(mountPoint string) {
 }
 
 func getUserAgent(appName string) string {
-	return strings.TrimSpace(fmt.Sprintf("gcsfuse/%s %s %s", getVersion(), appName, os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE")))
+	if len(os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE")) > 0 {
```
Tulsishah, line 79: What if len(appName) = 0 here ?

main_test.go
```diff
@@ -54,14 +54,21 @@ func (t *MainTest) TestGetUserAgentWhenMetadataImageTypeEnvVarIsSet() {
 	defer os.Unsetenv("GCSFUSE_METADATA_IMAGE_TYPE")
 
 	userAgent := getUserAgent("AppName")
-	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s %s %s", getVersion(), "AppName", os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE")))
+	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s AppName (GPN:gcsfuse-DLVM)", getVersion()))
```
Tulsishah, line 57: I think you should use os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE") instead of "DLVM" directly as we are testing environment variable also.

Tulsishah (commented):
I think you should add unit tests for  len(appName) = 0 also.

main_test.go
ashmeenkaur: Please correct me if I am wrong, but, os.Getenv() is being used in the getUserAgent() method. Using os.Getenv here won't catch any failures if something goes wrong in the os.Getenv method as if it will have the wrong value in both expected and tested user agent.

Tulsishah (approved):

PR #1074, opened 2023-04-25
Removing all the logs which is used under --debug_fs flag

raj-prince:
### Description
Removing the logs which was used under `--debug_fs` flag. Keeping the flag for the future, in case we want to log in gcsfuse code.

Why?

- Logging via `--debug_fs` was subset of logging via `--debug_fuse`. So, we can use `--debug_fuse` flag while logging.
- At FS level gcsfuse only converts from syscall (fuse) to GCS api call. 

### Testing details
1. Manual -

- `debug_fs` flag is not working any more.
- Mouted with other remaining flags, logging is working fine.

3. Unit tests - automation.
4. Integration tests - automation.


vadlakondaswetha (commented):
In future, if we need anything to log in file-system. We can use the same flag --debug_fuse.

if we want to log anything in gcsfuse, how to log it?

>> [Prince] As per my understanding, gcsfuse is working right now as a thin connector b/w fuse and gcs. So, most of the logs will be part of these two flags. In some parts of GCSFuse, E.g. ContentCache, logs are written with the default logger, not guarded by any special debug flag. And most of the logs (E.g. auth, stackdriver) in gcsfuse  are one time setup log, which are also written with the default logger. 

So, I don't think we need another flag to guard gcsfuse related logs unless until we are introducing a complex feature related to caching or some optimisations.

perfmetrics/scripts/ml_tests/tf/resnet/setup_scripts/setup_container.sh
```diff
@@ -17,7 +17,7 @@ cd -
 
 # Mount the bucket and run in background so that docker doesn't keep running after resnet_runner.py fails
 echo "Mounting the bucket"
-nohup gcsfuse/gcsfuse --foreground --implicit-dirs --enable-storage-client-library --debug_fs --debug_gcs --max-conns-per-host 100 --log-format "text" --log-file /home/logs/gcsfuse.log --stackdriver-export-interval 60s ml-models-data-gcsfuse myBucket > /home/output/gcsfuse.out 2> /home/output/gcsfuse.err &
```
Tulsishah, line 20: We are adding --debug_fuse flag instead of this?

Tulsishah (commented):
Changes are remains here.
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/flags.go#L314

perfmetrics/scripts/ml_tests/tf/resnet/setup_scripts/setup_container.sh
raj-prince: Thanks, added --debug_fuse flag.

flags.go
```diff
@@ -313,7 +313,7 @@ func newApp() (app *cli.App) {
 
 			cli.BoolFlag{
 				Name:  "debug_fs",
-				Usage: "Enable file system debugging output.",
+				Usage: "This flag is currently unused.",
```
sethiay, line 316: Does using Hidden:true better here ?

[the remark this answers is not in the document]
raj-prince: I'll correct the doc for tf_resnet. Other than that, I don't think we should remove or not. Otherwise, there is no point to keep it in the code.

vadlakondaswetha (approved):

sethiay (approved):

PR #1080, opened 2023-04-27
Read, Write, Copy, Rename tests [Readonly Filesystem]

Tulsishah:
### Description
Adding testcase
1. Read
    -  ReadFile, ReadSubDirectoryFile
    -  ReadDir is covered in list tests
2. Open
    -  Open File to Write, append or update data
    -  Write, Update, and Compose tests are skipped because we need to open the file first.
3. CopyFile 
   -  Copy one file to another file
   -  Copy one Directory to another dir. 
4. Rename
   -  Rename file 
   -  Rename Directory  to another dir. 
  
### Testing details
1. Manual - All the tests are passing.
6. Unit tests - N/A
7. Integration tests - All the inetgration tests are passing from integration tests directory


tools/integration_tests/readonly/rename_object_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if _, err := os.Stat(newObjPath); err == nil {
+		t.Errorf("Renamed file found in read-only file system.")
+	}
+}
+
+// Rename testBucket/Test1.txt to testBucket/Rename.txt
+func TestRenameFile(t *testing.T) {
+	oldFilePath := path.Join(setup.MntDir(), FileNameInTestBucket)
+	newFilePath := path.Join(setup.MntDir(), "Rename.txt")
```
sethiay, line 57: Can we create constant for "Rename.txt" as well ?

tools/integration_tests/readonly/rename_object_test.go
```diff
@@ -0,0 +1,92 @@
...
+		t.Errorf("File renamed in read-only file system.")
+	}
+
+	checkErrorForReadOnlyFileSystem(err, t)
+
+	if _, err := os.Stat(oldObjPath); err != nil {
+		t.Errorf("SrcFile is deleted in read-only file system.")
+	}
+	if _, err := os.Stat(newObjPath); err == nil {
+		t.Errorf("Renamed file found in read-only file system.")
```
sethiay, line 50: In case of renaming directory, I think we also need to ensure that none of its child is deleted.

tools/integration_tests/readonly/copy_object_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Copy srcFile in testBucket/Test/b/b.txt destination.
+func checkIfFileCopyFailed(srcFilePath string, t *testing.T) {
+	source, err := os.OpenFile(srcFilePath, syscall.O_DIRECT, setup.FilePermission_0600)
+	if err != nil {
+		t.Errorf("Error in the opening file: %v", err)
+	}
+
+	// Checking if destination object exist.
+	copyFile := path.Join(setup.MntDir(), DirectoryNameInTestBucket, SubDirectoryNameInTestBucket, FileNameInSubDirectoryTestBucket)
```
sethiay, line 37: Why is it necessary to have destination file exist for copy to work here ? We don't need in linux cp operation.

tools/integration_tests/readonly/read_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"syscall"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/setup"
+)
+
+func readFile(filePath string, t *testing.T) (content []byte) {
+	file, err := os.OpenFile(filePath, os.O_RDONLY|syscall.O_DIRECT, setup.FilePermission_0600)
+	if err != nil {
+		setup.LogAndExit(fmt.Sprintf("Error in the opening the file %v", err))
```
sethiay, line 31: why are we using setup.LogAndExit here and not throwing errors like other places?

tools/integration_tests/readonly/copy_object_test.go
Tulsishah: As it is readonly file system we can't create the file if it does not exist.

sethiay: In that case, I would suggest to add a comment that cp without destination file creates a destination file and create workflow is already covered separately.

tools/integration_tests/readonly/copy_object_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	destination, err := os.OpenFile(copyFile, syscall.O_DIRECT, setup.FilePermission_0600)
+	if err != nil {
+		t.Errorf("File %s opening error: %v", destination.Name(), err)
+	}
+	defer destination.Close()
+
+	// File copying with io.Copy() utility.
+	_, err = io.Copy(destination, source)
+	// It will throw an error "copy_file_range: bad file descriptor"
```
sethiay, line 50: Let's keep this comment above io.Copy call and change it to something like: In read only filesystem, io.Copy throws "copy_file_range: bad file descriptor" error.

sethiay (approved):

PR #1085, opened 2023-05-03
Adding vendor folder to gitignore file

raj-prince:
### Description
Removing the vendor directory from the gcsfuse codebase. Change contains:
(a) Added vendor directory in the `.gitignore` file.
(b) Deleted vendor directory using `git rm --cached -r vendor` command.

#### Check Required
1. **Building the codebase:** We can build the codebase using the same command. There is not change at all. So, when we build the first time it fetches all the dependencies in the "GOCACHE" (go env) folder, which takes around 30 seconds. After that it uses the same cache to build which takes around 3-4 seconds. Hence, we don't require any change in any of the scripts.
2. **Truly build reproducible:** By default go module contains `go.mod` and `go.sum` file to maintain an immutable database, which makes the build process truly build reproducible. Ref - https://go.dev/blog/module-mirror-launch
3. **Test code locally in changing the dependency code:** You can use https://go.dev/ref/mod#go-mod-file-replace to test while changing in the dependency code.
4. I'll raise another PR to setup cache in the CI/CD workflow.

### Testing details
1. Manual - (a) Manually verified all the basic file system operations. (b) Verified the build process by cleaning the go-cache `go clean -cache`.
5. Unit tests - Automation
6. Integration tests - Automation


sethiay (approved):
Looks good.

1. I think the perf tests are not executed even with execute-perf-test label. I hope that's expected.
2. If it takes 30-40 seconds to build without vendor directory, then i think it's fine to not add cache in CI/CD pipeline ? Let me know if I am missing something.
>> [raj-prince] Yeah, we can skip for now. In future,  if it requires then we can add.

Tulsishah:
> Looks good.
> 
> 1. I think the perf tests are not executed even with execute-perf-test label. I hope that's expected.
> 2. If it takes 30-40 seconds to build without vendor directory, then i think it's fine to not add cache in CI/CD pipeline ? Let me know if I am missing something.

Perf tests are not executed because @raj-prince attached a label and hasn't pushed any commit after that. You can run it through test fusion directly, also.

raj-prince:
Performance test number:
+--------+------------+--------------+------------+--------------+--------------+
+--------+------------+--------------+------------+--------------+--------------+
| Master |  0.25MiB   | 469.83MiB/s  | 2.43MiB/s  |  33.32MiB/s  |  2.79MiB/s   |
|   PR   |  0.25MiB   |  465.8MiB/s  | 2.39MiB/s  |  32.72MiB/s  |  3.26MiB/s   |
| Master | 48.828MiB  | 4775.22MiB/s | 85.0MiB/s  | 1055.44MiB/s |  82.43MiB/s  |
|   PR   | 48.828MiB  | 4634.68MiB/s | 82.05MiB/s | 1030.05MiB/s |  82.3MiB/s   |
| Master | 976.562MiB | 4736.92MiB/s | 33.41MiB/s | 456.48MiB/s  |  34.53MiB/s  |
|   PR   | 976.562MiB | 4697.66MiB/s | 21.09MiB/s | 558.46MiB/s  |  33.13MiB/s  |
+--------+------------+--------------+------------+--------------+--------------+

PR #1091, opened 2023-05-05
Adding the gcsfuse stuck issue in troubleshoot doc

raj-prince:
### Description
Adding the gcsfuse stuck issue in troubleshoot doc.

Only appended an entry in the table. Changes are shown due to reformatting of the table.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


docs/troubleshooting.md
```diff
@@ -1,18 +1,19 @@
...
+| Cloud Storage FUSE gets stuck while doing file operation in high concurrency.                                                                                                                        | This happens when client holds the connection while doing file read/write operation. You might try using `--http-client-timeout=<value>`, this will break the connection if that is hold by client application for too long, and `--max-conns-per-host=<concurrency>`, this will increase the parallel connection.                                                                                                                                                                                                                                                                                                     |
```
sethiay, line 19: Cloud Storage FUSE gets stuck while doing file operation in high concurrency -> Cloud Storage FUSE gets stuck when using it to concurrently work with a large number of opened files (reference issue [here](https://github.com/GoogleCloudPlatform/gcsfuse/issues/1040))

This happens when client holds the connection while doing file read/write operation -> This happens when gcsfuse is mounted with http1 client (default) and the application using gcsfuse tries to keep more than value of ```--max-conns-per-host``` number of files opened.  

You might try using `--http-client-timeout=<value>`, this will break the connection if that is hold by client application for too long, and `--max-conns-per-host=<concurrency>`, this will increase the parallel connection. -> You can try (a) Passing  a value higher than the number of files you want to keep open to ```--max-conns-per-host``` flag. (b) Adding some timeout for http client connections using ```--http-client-timeout``` flag.

sethiay (approved):

docs/troubleshooting.md
raj-prince: Thanks!

PR #1092, opened 2023-05-05
Log crash during mounting

Tulsishah:
Description
Background, with logfile - Panic message printed in Logfile
Background, without logfile - Panic message printed in Syslog
Foreground, with logfile - Panic message printed in Logfile
Foreground, without logfile - Panic message printed on Console

Fatal print simple statement with os.Exit(1)
E.g. I0517 10:38:22.851246 Panic: Test

Testing
Manual - Tested manually by putting panic message before mounting
UnitTest - N/A
IntegrationTest - N/A


main.go
```diff
@@ -423,7 +423,18 @@ func run() (err error) {
 	return
 }
 
+// Handle panic if the crash occurs during mounting.
+func handlePanic() {
+	// Detect if panic happens in main go routine.
+	a := recover()
+	if a != nil {
+		logger.Info("Panic: ", a)
```
vadlakondaswetha, line 431: why is panic an info message?

Tulsishah: As panic is will do exit code 1, before that we are printing the cause of panic.

vadlakondaswetha: message type - info or error represents the type of log. Yes panic will crash the process, that doesn't mean we print the source of panic as info message. 
https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

FYI, here you made panic log as fatal: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

Tulsishah: Yes I thought that but [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/wrappers/error_mapping.go#L93) em errormapping logger have Fatal method but our [logger](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/logger/logger.go) does not have Fatal method.

Should I add that method in our logger in proceed ?

vadlakondaswetha: yes

internal/logger/logger.go
```diff
@@ -133,6 +133,10 @@ func Info(v ...interface{}) {
 	defaultInfoLogger.Println(v...)
 }
 
+func Fatal(v ...interface{}) {
```
raj-prince, line 136: Can you please write a one-liner as like in the other exported method?

Also, can you please update the description how fatal logs look like?

main.go
```diff
@@ -423,7 +423,18 @@ func run() (err error) {
 	return
 }
 
+// Handle panic if the crash occurs during mounting.
+func handlePanic() {
```
raj-prince, line 427: We can rename this method as it looks like so generic - handlePanicWhileMounting().

raj-prince (approved):

internal/logger/logger.go
```diff
@@ -133,6 +133,11 @@ func Info(v ...interface{}) {
 	defaultInfoLogger.Println(v...)
 }
 
+// Fatal calls the default info logger to print the message followed by a call to os.Exit(1).
```
vadlakondaswetha, line 136: logger is not making a call to os.exit(1)

vadlakondaswetha (approved):

internal/logger/logger.go
vadlakondaswetha: also will it create any problem in doing os.exit(1) before the crash is happening?

PR #1102, opened 2023-05-09
Upgrading the fuse go module to suppress error logs for GetXAttrs call

raj-prince:
### Description
Upgrading the fuse go module to accommodate [this](https://github.com/jacobsa/fuse/pull/142) change.

**Mount using the below command on cloud top machine:**
` gcsfuse --foreground --debug_fuse --debug_fuse_errors <bucket_name> <mount_dir>`

**Before this change:**
```
{"name":"root","levelname":"ERROR","severity":"ERROR","message":"fuse: *fuseops.GetXattrOp error: function not implemented\n","timestampSeconds":1683638463,"timestampNanos":315054543}
```
**After this change:**
We don't get the error log when we mount with the above mentioned command.

### Testing details
1. Manual - Manually verified.
2. Unit tests - Automation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/fsouza/fake-gcs-server v1.38.4
 	github.com/googleapis/gax-go/v2 v2.7.0
 	github.com/jacobsa/daemonize v0.0.0-20160101105449-e460293e890f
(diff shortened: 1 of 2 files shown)


sethiay (approved):
Regarding the Description - Do we not get those error logs at all or do we get it only one time ?

raj-prince:
image

As expected, we don't see any performance regression.

PR #1103, opened 2023-05-09
Rename directory integration test

Tulsishah:
### Description
Tested with --rename-dir-limit=3 
- TestRenameDirectoryWithThreeFiles
- TestRenameDirectoryWithTwoFiles
- TestRenameDirectoryWithFourFiles
- TestRenameDirectoryWithTwoFilesAndOneEmptyDirectory
- TestRenameDirectoryWithTwoFilesAndOneNonEmptyDirectory

### Testing details
Manual - Manually tested all the tests with both the flags(testBucket and mountedDir)
Unit tests - N/A
Integration tests - All the tests together are working fine.

### DIFF
--- a/tools/integration_tests/rename_dir_limit/rename_dir_limit_test.go
+++ b/tools/integration_tests/rename_dir_limit/rename_dir_limit_test.go
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Provides integration tests when --rename-dir-limit flag is set.
-package rename_directory_test
+package rename_dir_limit_test
 
 import (
 	"log"
--- a/tools/integration_tests/rename_dir_limit/rename_dir_test.go
+++ b/tools/integration_tests/rename_dir_limit/rename_dir_test.go
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Provides integration tests when --rename-dir-limit flag is set.
-package rename_directory_test
+package rename_dir_limit_test
 
 import (
 	"os"


tools/integration_tests/rename_directory/rename_dir_limit_test.go
```diff
@@ -0,0 +1,221 @@
...
+		t.Errorf("Error in creating file: %v", err)
+	}
+}
+
+// Directory structure
+// testBucket/directoryWithThreeFiles           -- Dir
+// testBucket/directoryWithThreeFiles/a.txt     -- File
+// testBucket/directoryWithThreeFiles/b.txt     -- File
+// testBucket/directoryWithThreeFiles/c.txt     -- File
+func createDirectoryWithThreeFiles(t *testing.T) {
```
sethiay, line 45: Instead of creating different function for directory with three files, directory with four files etc. Can we create a generic function something like createDirectoryWithNFiles ?

tools/integration_tests/rename_directory/rename_dir_limit_test.go
```diff
@@ -0,0 +1,168 @@
+// Copyright 2023 Google Inc. All Rights Reserved.
```
sethiay, line 1: since this test is specific to --rename-dir-limit, so should we rename the package to rename_dir_limit instead of rename_directory_test ?

sethiay (approved):

PR #1115, opened 2023-05-15
Removing condition in setup mounting 

Tulsishah:
### Description
Before if mounting output is throwing more then one line it was throwing an error.
but if I pass --key-file as parameter, in the first line it is fetching absolute path of key file and then throw success message.
but because of [this condition](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L160) it was failing before.

Remove the condition to fix the failure

### Testing details
1. Manual
- All the integration tests are working fine for both the flags testBucket and mountedDirectoy.
- Tested for both foreground and background 
           - No permission from key file to mount the bucket (mounting is failing) 
           - Without key file and gcloud credentials
           - Panic  (mounting is failing) 
           - Crash during mounting
           - Normal error while mounting
3. Unit tests - N/A
4. Integration tests- All the integration tests are working fine together.

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -1,7 +1,6 @@
 package setup
 
 import (
-	"bytes"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


raj-prince:
Can you please check the below behaviour for both `foreground` and `background`: 
1. Crash while mounting?
2. Normal error while mounting? 
Also, could you please go through the mounting workflow to ensure we are handling the error appropriately?

raj-prince:
We can also see the history of this change? Why was this condition added?

Tulsishah:
> We can also see the history of this change? Why was this condition added?

Yeah, I already checked that, @lezh added it, and there is no description on why to add that.

Tulsishah:
> Can you please check the below behaviour for both `foreground` and `background`:
> 
> 1. Crash while mounting?
> 2. Normal error while mounting?
>    Also, could you please go through the mounting workflow to ensure we are handling the error appropriately?

Tested

raj-prince (approved):

PR #1122, opened 2023-05-16
Fix lint failures

ashmeenkaur:
### Description
This CL fixes some of the lint failures we see today on the master branch.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/benchmarks/write_locally/main.go
+++ b/benchmarks/write_locally/main.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"os"
 	"time"
@@ -48,7 +47,7 @@ func run() (err error) {
 	// Create a temporary file.
 	log.Printf("Creating a temporary file in %s.", *fDir)
 
-	f, err := ioutil.TempFile(*fDir, "write_locally")
+	f, err := os.CreateTemp(*fDir, "write_locally")
 	if err != nil {
 		err = fmt.Errorf("TempFile: %w", err)
 		return
--- a/benchmarks/write_locally/main.go
+++ b/benchmarks/write_locally/main.go
@@ -58,11 +58,11 @@ func run() (err error) {
 	// Make sure we clean it up later.
 	defer func() {
 		log.Printf("Truncating and closing %s.", path)
-		f.Truncate(0)
-		f.Close()
+		_ = f.Truncate(0)
+		_ = f.Close()
 
 		log.Printf("Deleting %s.", path)
-		os.Remove(path)
+		_ = os.Remove(path)
 	}()
 
 	// Extend to the initial size.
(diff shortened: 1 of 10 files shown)


benchmarks/write_locally/main.go
```diff
@@ -95,11 +94,6 @@ func run() (err error) {
 		// Overwrite.
 		var n int64
 		for n < *fFileSize && time.Since(start) < *fDuration {
-			toWrite := *fFileSize - n
```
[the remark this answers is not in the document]
ashmeenkaur: Hey, if you'll notice, we're just assigning toWrite variable some value but not using it anywhere. So, this piece of code was redundant.

internal/perf/cpu.go
```diff
@@ -44,7 +43,10 @@ func HandleCPUProfileSignals() {
 		}()
 
 		// Profile.
-		pprof.StartCPUProfile(f)
+		err = pprof.StartCPUProfile(f)
+		if err != nil {
+			logger.Infof("StartCPUProfile failed: %v", err)
```
vadlakondaswetha, line 48: can we make it a warning or error log.

vadlakondaswetha: also does it make sense to return from here instead of executing further steps?

vadlakondaswetha (commented):
minor comments. other changes looks good.

internal/perf/cpu.go
ashmeenkaur: func StartCPUProfile returns an error if profiling is already enabled. So, I think we should not break the flow here. I've changed the log level to ERROR.

internal/perf/cpu.go
```diff
@@ -44,7 +43,11 @@ func HandleCPUProfileSignals() {
 		}()
 
 		// Profile.
-		pprof.StartCPUProfile(f)
+		err = pprof.StartCPUProfile(f)
+		if err != nil {
+			errorLogger := logger.NewError("")
+			errorLogger.Printf("StartCPUProfile failed: %v", err)
```
vadlakondaswetha, line 49: you should create a method like this to log errors incase error is not available on available logger. https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/logger/logger.go#L138

Creating a newError logger for every log is not correct way to do.

internal/perf/cpu.go
vadlakondaswetha: how does it make sense to run these 2 steps if profile start itself failed:
time.Sleep(duration)
pprof.StopCPUProfile()

internal/perf/cpu.go
ashmeenkaur: Updated! Thanks!

internal/logger/logger.go
```diff
@@ -33,6 +33,8 @@ const GCSFuseInBackgroundMode string = "GCSFUSE_IN_BACKGROUND_MODE"
 var (
 	defaultLoggerFactory *loggerFactory
 	defaultInfoLogger    *log.Logger
+
```
vadlakondaswetha, line 36: nit: remove this extra line

vadlakondaswetha (approved):

Tulsishah (approved):

PR #1134, opened 2023-05-19
fixing kokoro script

vadlakondaswetha:
### Description
Unmount is being done before deleting the old logs. fixing it by moving the umount command to post all clean up.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
+++ b/perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
@@ -54,15 +54,15 @@ gcsfuse $GCSFUSE_FLAGS $BUCKET_NAME $MOUNT_POINT
 # Executing perf tests
 chmod +x run_load_test_and_fetch_metrics.sh
 ./run_load_test_and_fetch_metrics.sh
-sudo umount $MOUNT_POINT
 
 # Copying gcsfuse logs to bucket
 gsutil -m cp $LOG_FILE gs://periodic-perf-tests/fio-gcsfuse-logs/
 
 # Deleting logs older than 10 days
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
```diff
@@ -54,15 +54,15 @@ gcsfuse $GCSFUSE_FLAGS $BUCKET_NAME $MOUNT_POINT
...
 chmod +x run_load_test_and_fetch_metrics.sh
 ./run_load_test_and_fetch_metrics.sh
-sudo umount $MOUNT_POINT
 
 # Copying gcsfuse logs to bucket
 gsutil -m cp $LOG_FILE gs://periodic-perf-tests/fio-gcsfuse-logs/
 
 # Deleting logs older than 10 days
 python3 utils/metrics_util.py gcs/fio-gcsfuse-logs/ 10
+sudo umount $MOUNT_POINT
```
sethiay, line 63: 1. I believe unmounting shouldn't affect log file which is already created in the VM's disk. So, whether unmounting before or after copying log file shouldn't make difference.
2. If failure occurs in running ```./run_load_test_and_fetch_metrics.sh``` then the script fails and flow doesn't reach to gsutil command. So, i think we need to fix that.

vadlakondaswetha: 1. Logs are copied using gsutil. But the logs folder fio-gcsfuse-logs is inside the mounted directory. hence cleaning up of files older than 10 days is done using gcsfuse mount.
2. The fix for that is being tracked seperately and will be fixed in a different PR.

sethiay (approved):

PR #1138, opened 2023-05-24
exp1:fetching  selected attributes

anushka567:
### Description
Fetching selective attributes during listObjectRequest

### Testing details
1. Manual - NA
2. Unit tests - [Test](https://github.com/anushka567/gcsfuse/blob/experiments-ls-latency/internal/storage/bucket_handle_test.go#L73)
3. Integration tests - NA


anushka567:
@vadlakondaswetha @gargnitingoogle

internal/storage/bucket_handle.go
```diff
@@ -244,6 +244,8 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 		IncludeTrailingDelimiter: req.IncludeTrailingDelimiter,
 		//MaxResults: , (Field not present in storage.Query of Go Storage Library but present in ListObjectsQuery in Jacobsa code.)
 	}
+	query.SetAttrSelection([]string{"Name","Size","Generation","MetaGeneration","Updated","Metadata"})
```
vadlakondaswetha, line 247: Here we are fetching only fewer attributes but creating gcs.Object with more empty properties.

Can we do an experiment with this flow and then returning smaller object(only with 5 properties) from this method and see if they is any change in the latency?

PR #1147, opened 2023-05-29
Upgrade ai ml tests to successfully run for long duration on another VM

sethiay:
### Description
Upgrade ai ml tests to successfully run for long duration on VM other than the VM on which the kokoro tests are triggered.  This is required because of intermittent issues that comes while running long duration tests on Kokoro VM directly.

Notes for reviewer:
1. I have used `sudo gcloud` instead of `gcloud` because `gcloud` is giving errors while doing ssh from Kokoro VM and I couldn't find a solution.
2. I have changed the base image for Pytorch dino model because the nvidia one doesn't have `gsutil` preinstalled.
3. I will take care of to-dos before merging.
4. To-do: Take care of deleting logs for older build in GCS buckets, similar to periodic perf tests.
5. The existing build.sh files for both tf and pytorch are moved under scripts/ml_tests/tf/resnet/ and scripts/ml_tests/pytorch/dino by the name setup_host_and_run_model.sh

NA

### Testing details
1. Manual - Tested the setup for tf model with less number of epochs. To-do/In progress: Test the setup for pytorch and with relatively higher number of epochs.
6. Unit tests - NA
7. Integration tests - NA


perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
```diff
@@ -0,0 +1,192 @@
...
+    # gsutil always gives error (even the files are copied) while uploading
+    # files that are changing while uploading and gcsfuse logs are changing
+    # when the test is running.
+    set +e
+    echo "Copying GCSFuse logs to GCS bucket"
+    sudo gcloud compute ssh $VM_NAME --zone $ZONE_NAME --internal-ip --command "gsutil rsync -R -d \$HOME/github/gcsfuse/container_artifacts/ $ARTIFACTS_BUCKET_PATH/$1/container_artifacts"
+    echo "\n"
+  )
+  sudo gcloud compute ssh $VM_NAME --zone $ZONE_NAME --internal-ip --command "gsutil cp \$HOME/build.out $ARTIFACTS_BUCKET_PATH/$1/build.out"
+  sudo gcloud compute ssh $VM_NAME --zone $ZONE_NAME --internal-ip --command "gsutil cp \$HOME/build.err $ARTIFACTS_BUCKET_PATH/$1/build.err"
```
vadlakondaswetha, line 91: why not use kokoro artifacts functionality instead of manually copying the files?

vadlakondaswetha: got it. may be its not viable since every time you try to check the status it will create a seperate folder.

vadlakondaswetha (approved):

perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
sethiay: Also, the test and gcsfuse logs are on different VM than Kokoro VM, so we will have to first copy to Kokoro VM to use that functionality. So, instead of first copying to Kokoro and then to bucket, we can directly copy to bucket.

PR #1149, opened 2023-05-30
List Integration tests

Tulsishah:
### Description
Added list directory tests. recursively listing directories.

NA

### Testing details
1. Manual - All the integration tests are working fine for both the flags testBucket and mountedDirectoy.
3. Unit tests - N/A
4. Integration tests- All the integration tests are working fine together.


tools/integration_tests/operations/operations_test.go
```diff
@@ -37,6 +37,12 @@ const SubDirInNonEmptyDestCopyDirectory = "subDestCopyDir"
 const DestCopyDirectoryNotExist = "notExist"
 const NumberOfObjectsInSrcCopyDirectory = 2
 const NumberOfObjectsInDestCopyDirectory = 2
+const DirectoryForListTest = "directoryForListTest"
+const FirstSubDirectoryForListTest = "firstSubDirectoryForListTest"
+const FileInDirectoryForListTest = "fileInDirectoryForListTest"
+const FileInFirstSubDirectoryForListTest = "fileInFirstSubDirectoryForListTest"
+const SecondSubDirectoryForListTest = "secondSubDirectoryForListTest"
+const FileInSecondSubDirectoryForListTest = "fileInSecondSubDirectoryForListTest"
```
sethiay, line 45: can we create third subdirectory which is empty ?

tools/integration_tests/operations/list_dir_test.go
```diff
@@ -0,0 +1,143 @@
...
+func listDirectory(path string, t *testing.T) {
+	//Reading contents of the directory
+	objs, err := os.ReadDir(path)
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for _, obj := range objs {
+		checkIfListedCorrectDirectory(path, obj, t)
```
sethiay, line 100: This checks if the listed object is under correct directory or not. However, let's say there is a bug in code and some object in a directory doesn't appear in "objs" itself then this test case will still pass. Can we cover that case as well ?

tools/integration_tests/operations/operations_test.go
Tulsishah: added

tools/integration_tests/operations/list_dir_test.go
```diff
@@ -0,0 +1,208 @@
...
+		{
+			// testBucket/directoryForListTest/secondSubDirectoryForListTest/fileInSecondSubDirectoryForListTest   -- File
+			if obj.Name() != FileInSecondSubDirectoryForListTest && obj.IsDir() == true {
+				t.Errorf("Listed incorrect object")
+			}
+		}
+	}
+}
+
+func checkIfListedDirectoryHasCorrectNumberOfObjects(dirPath string, numberOfObjects int, t *testing.T) {
```
sethiay, line 102: It is getting a little complicated. Instead of creating functions like getDirNameFromPath, checkIfListedDirectoryHasCorrectNumberOfObjects etc, can we keep it simple i.e. a function to create the directory structure (with files) and then list and check if results are expected in TestListDirectoryRecursively (remove recursive from function name) ?

Tulsishah: Do you mean
Create directory structure.
then list directory one by one and check in the test it self if directory structure is matching or not ?

If yes then this will not be a recursive test right as we added in our test design?

sethiay: Yes,

for recursively listing a directory, you can check https://pkg.go.dev/path/filepath#Walk or similar

tools/integration_tests/operations/list_dir_test.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+}
+
+func createDirectoryStructureForTest(t *testing.T) {
+	// Directory structure
+	// testBucket/directoryForListTest                                                                     -- Dir
+	// testBucket/directoryForListTest/fileInDirectoryForListTest		                                       -- File
+	// testBucket/directoryForListTest/firstSubDirectoryForListTest                                        -- Dir
+	// testBucket/directoryForListTest/firstSubDirectoryForListTest/fileInFirstSubDirectoryForListTest     -- File
+	// testBucket/directoryForListTest/secondSubDirectoryForListTest                                       -- Dir
```
sethiay: do you think adding a test case for a sub directory with more than one file would also be helpful ?

Tulsishah: I think it is covered in the listing the main directory. so currently, how is it working
mntDir
    - dir1
         - subdir1
         - subdir2
         - emptysubdir
         - file
         
We are started listing from mntDir. So first, dir1 will get listed. and dir1 has all types of data.

Yet you think we should add, I will add for the same.

sethiay: Yes agree, mntDir has more than one file collectively in its child directories. But I was thinking there is not dir/subDir which immediately has more than one file.

tools/integration_tests/operations/list_dir_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"log"
+	"os"
+	"path"
+	"path/filepath"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
+)
+
+func createDirectoryWithFile(dirPath string, filePath string, t *testing.T) {
```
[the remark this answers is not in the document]
Tulsishah: I don't think so because here we are creating dir and file with given name.
We already have one function [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/rename_dir_limit/rename_dir_test.go#L26) on what you are talking about and I thought to using it. but here it is creating temp file with random suffix at the end in name which is not useful in our case.

sethiay (approved):

PR #1157, opened 2023-06-01
Fixing integration tests

Tulsishah:
### Description
1. Fixing - building gcsfuse again for every mounting.
2. Fixing - log.Printf to log.Print
3. Before I was running all the tests on cloud top and there it was taking long time to run. on GCE VM it is working fine.

NA

### Testing details
1. Manual - All the integration tests are working fine for both the flags testBucket and mountedDirectoy.
2. Unit tests - N/A
3. Integration tests- All the integration tests are working fine together.


tools/integration_tests/operations/operations_test.go
```diff
@@ -49,14 +49,16 @@ func TestMain(m *testing.M) {
...
-		log.Printf("Both --testbucket and --mountedDirectory can't be specified at the same time.")
+		log.Print("Both --testbucket and --mountedDirectory can't be specified at the same time.")
 		os.Exit(1)
 	}
 
 	// Run tests for mountedDirectory only if --mountedDirectory flag is set.
 	setup.RunTestsForMountedDirectoryFlag(m)
 
 	// Run tests for testBucket
+	setup.SetUpTestDirForTestBucketFlag()
```
vadlakondaswetha, line 60: what is this change for??

Tulsishah: Before...we were building gcsfuse again for every mounting e.g. static and only dir...

Put it in beginning place only ...so that now gcsfuse is building only once for every test package
Before gcsfuse was building twice for every test package...

I made it one

vadlakondaswetha (approved):

PR #1165, opened 2023-06-05
Fixing kokoro perf test issue

Tulsishah:
### Description
Kokoro perf test was failing due to space issue. Fix it by running integration tests after fio tests.
We are going with this fix till the disk issue is not resolved in Kokoro.

NA

### Testing details
1. Manual - Tests were passing on kokoro.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
+++ b/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
@@ -52,9 +52,6 @@ git fetch origin
 echo checkout PR branch
 git checkout pr/$KOKORO_GITHUB_PULL_REQUEST_NUMBER
 
-# Executing integration tests
-GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/... -p 1 --integrationTest -v --testbucket=gcsfuse-integration-test
-
 # Executing perf tests
 echo Mounting gcs bucket from pr branch
 mkdir -p gcs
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
```diff
@@ -65,5 +62,8 @@ chmod +x perfmetrics/scripts/presubmit/run_load_test_on_presubmit.sh
 ./perfmetrics/scripts/presubmit/run_load_test_on_presubmit.sh
 sudo umount gcs
 
+# Executing integration tests
+GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/... -p 1 --integrationTest -v --testbucket=gcsfuse-integration-test
```
sethiay, line 66: should also run e2e tests at the end in case of periodic tests as well ?

sethiay, line 66: Also, can we move these tests after printing the perf test results ?

Tulsishah: Ideally it should be before fio test only as swetha mentioned earlier as if E2E tests will fail it will terminate only from that point.

sethiay: okay, if we get similar errors in periodic test then we can change.

vadlakondaswetha, line 66: how are GCSFuse logs useful here. I dont see we are storing them anywhere? can we clean the disk used by integration tests?

Tulsishah: If integration tests are failing, we can check the logs. Also, I checked size of the package. It was around 30 MB only. While building gcsfuse, I guess some hidden or other files are getting installed which is taking more space.

vadlakondaswetha: where do we check the logs? if its on sponge2, then it has nothing to do with clearing the space consumed by package/integration tests.

Tulsishah: By doing ssh on kokoro VM. The VM is not getting deleted instantly. So one can check from there.

vadlakondaswetha: thats the wrong way of checking logs. if someone doesn't keep track of execution, then there is no way to check the logs.

vadlakondaswetha (approved):

PR #1167, opened 2023-06-06
Integration tests --key-file flag and GOOGLE_APPLICATION_CREDENTIALS env with admin permission tests

Tulsishah:
### Description
Created integration tests for admin credentials. 

1. Test without key-file and with GOOGLE_APPLICATION_CREDENTIALS env
2. Test with key-file and GOOGLE_APPLICATION_CREDENTIALS env
3. Test with key-file and without GOOGLE_APPLICATION_CREDENTIALS env
4. Test Without –key-file flag and GOOGLE_APPLICATION_CREDENTIALS
This case will not get covered as gcsfuse internally authenticates from a metadata server on GCE VM. [ref](https://github.com/golang/oauth2/blob/master/google/default.go#L160)

Running the Creds test with operations_test package only.

NA

### Testing details
11. Unit tests - NA


tools/integration_tests/util/creds_tests/admin_creds.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/mounting/static_mounting"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
+)
+
+func RunTestsForKeyFileAndGoogleApplicationCredentialsEnvVarSet(testFlagSet [][]string, m *testing.M) (successCode int) {
+	// Saving testBucket value for setting back after testing.
+	testBucket := setup.TestBucket()
+
+	// Set the testBucket value to the bucket belonging to a different project for testing credentials.
+	setup.SetTestBucket("integration-test-gcsfuse")
```
sethiay, line 36: Can you try if this works to avoid hard coding the bucket here:
1. Before test, remove storage permission for default SA only for the passed testBucket using https://cloud.google.com/storage/docs/gsutil/commands/iam
2. After test, give the permissions back to testBucket.

tools/integration_tests/util/creds_tests/admin_creds.go
```diff
@@ -0,0 +1,99 @@
...
+		flags := testFlagSet[i]
+		flags = append(flags, setup.TestBucket())
+		flags = append(flags, setup.MntDir())
+		err := mounting.MountGcsfuse(flags)
+		if err == nil {
+			log.Print("Error: Mounting successful without key file.")
+		}
+	}
+
+	// Get credential from bucket.
```
sethiay, line 52: Instead of keeping the service account json in a bucket, do you think we can create, download and delete json: https://cloud.google.com/iam/docs/keys-create-delete

tools/integration_tests/util/creds_tests/testdata/delete_creds.sh
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
Tulsishah: For deleting creds at the end after testing.

tools/integration_tests/util/creds_tests/admin_creds.go
Tulsishah: I checked we cant't do this as we can't change the permissions of project service account. you can check it on pantheon also the edit button is disable for the same.

tools/integration_tests/util/creds_tests/admin_creds.go
Tulsishah: We can't do this as we can't download key file of the service account belongs to different project.

tools/integration_tests/util/creds_tests/testdata/provide_permission.sh
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# create bucket in gcs-fuse-test project and add permissions.
+TEST_BUCKET=$1
+SERVICE_ACCOUNT=$2
+PERMISSION=$3
+
+gsutil iam ch serviceAccount:$SERVICE_ACCOUNT:$PERMISSION gs://$TEST_BUCKET
```
raj-prince, line 20: Does this command also create the bucket?

tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_delete_service_account_and_creds.sh
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Delete service account after testing
+SERVICE_ACCOUNT=$1
+CREDS=$2
+gcloud auth revoke $SERVICE_ACCOUNT
+gcloud iam service-accounts delete $SERVICE_ACCOUNT
+rm $CREDS
```
raj-prince, line 20: You can rename the file name -> revoke_permission_and_delete_service_account_and_key_file.sh

and CREDS -> KEY_FILE

tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -0,0 +1,100 @@
...
+	// Fetching project-id to get service account id.
+	id, err := metadata.ProjectID()
+	if err != nil {
+		log.Printf("Error in fetching project id: %v", err)
+	}
+
+	// Service account id format is name@project-id.iam.gserviceaccount.com
+	serviceAccount := NameOfServiceAccount + "@" + id + ".iam.gserviceaccount.com"
+
+	cred_file_path := path.Join(os.Getenv("HOME"), "creds.json")
```
raj-prince, line 51: cred_file_path -> key_file or key_file_path

tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Testing with --key-file flag only
+	successCode = static_mounting.RunTests(testFlagSet, m)
+
+	if successCode != 0 {
+		return
+	}
+
+	return successCode
```
raj-prince, line 99: Could you please write a comment, why haven't we written the test when both KEY_FILE and ENV_VAR not present?

tools/integration_tests/operations/operations_test.go
```diff
@@ -107,5 +108,10 @@ func TestMain(m *testing.M) {
 		successCode = only_dir_mounting.RunTests(flags, m)
 	}
 
+	if successCode == 0 {
```
raj-prince, line 111: Since, we are running the complete integration test three more times. So, integration test runtime would have been 4x. Could you please check, how much time it takes to run the complete integration test right now?

tools/integration_tests/util/setup/setup.go
```diff
@@ -98,6 +98,10 @@ func MntDir() string {
 	return mntDir
 }
 
+func SetTestBucket(testBucketValue string) {
```
raj-prince, line 101: Just to remind, please revert if not used.

raj-prince (commented):
Have added few comments.

tools/integration_tests/util/creds_tests/creds.go
Tulsishah: I already added it in design doc, did the same here.

tools/integration_tests/operations/operations_test.go
Tulsishah: Around 32 minutes

raj-prince: As we discussed offline, we should skip some non-important tests here:
(a) We can only run the integration test for operations_test package.
(b) We can remove the test case, when both KEY_FILE and ENV_VAR is set. We can cover that under unit test (checking priority of key-file is greater than ENV_VAR) of this method - https://docs.google.com/spreadsheets/d/1ZnMW4zrCTISQBTLmdaDRkbyO1NexVwFXQk5RoQ-hu70/edit?resourcekey=0-K7wXCjQJ4CHhgxPm6q6q2Q#gid=647607290

CC: @sethiay

Tulsishah: Done

raj-prince (approved):
Please update the description, that we are running the Creds test with operations_test package only.

PR #1169, opened 2023-06-07
Adding logs to print all the fio job timing

raj-prince:
### Description
Printing the start and end time of all the fio jobs during the daily kokoro test. This will help to ensure the total run time of fio job is constant. If that is the case, then we have to figure out for the shift, for which we can reach to open-centsos team.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/fetch_metrics.py
+++ b/perfmetrics/scripts/fetch_metrics.py
@@ -63,6 +63,17 @@ if __name__ == '__main__':
   # waiting for 360 secs instead of 240 secs
   time.sleep(360)
 
+  # Print the start and end time of all the fio-jobs
+  # This is just make sure, the total run time of fio jobs are constant over
+  # different execution.
+  for ind, job in enumerate(temp):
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/fetch_metrics.py
```diff
@@ -63,6 +63,17 @@ def _parse_arguments(argv):
...
+  # Print the start and end time of all the fio-jobs
+  # This is just make sure, the total run time of fio jobs are constant over
+  # different execution.
+  for ind, job in enumerate(temp):
+    start_time_sec = job[fio_metrics.consts.START_TIME]
+    end_time_sec = job[fio_metrics.consts.END_TIME]
+
+    # Print start and end time of jobs
+    print("Start time: ", start_time_sec)
+    print("End time: ", end_time_sec)
```
vadlakondaswetha, line 75: print the job index also

PR #1174, opened 2023-06-09
List large dir integration test

Tulsishah:
### Description
Adding tests for listing large directory.
1. Test with a bucket with twelve thousand files.
2. Test with a bucket with twelve thousand files and hundred explicit directories.
3. Test with a bucket with twelve thousand files, hundred explicit directories, and hundred implicit directories.

NA

### Testing details
11. Unit tests - NA


raj-prince:
This doesn't trigger the pagination flow, can we increase the number greater than 50000? Assuming, we list 50000 objects in one page.

Tulsishah:
> Change wise, looks good to me. This doesn't trigger the pagination flow, can we increase the number greater than 50000? Assuming, we list 50000 objects in one page.

It is 5000
https://github.com/GoogleCloudPlatform/gcsfuse/blob/tulsishah/internal/fs/inode/dir.go#L39

tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
```diff
@@ -0,0 +1,141 @@
...
+	// Checking if all the object is File type.
+	for i := 0; i < len(objs); i++ {
+		if objs[i].IsDir() {
+			t.Errorf("Listes object is incorrect.")
+		}
+	}
+
+	// Checking if correct file present in bucket.
+	var fileName []string
+	for i := 0; i < len(objs); i++ {
```
sethiay, line 68: nit: 

IIUC
(a) This code runs in 0(nlogn) ?
(b) It doesn't check duplicate file names which should be fine because there can't be duplicate files in filesystem ?
(c) File name is of type <filenameprefix>{i} where i >=1 & i <= 12000 ?

if above is correct, then can we reduce the complexity to O(n) ?

tools/integration_tests/list_large_dir/list_large_dir_test.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Printf("Both --testbucket and --mountedDirectory can't be specified at the same time.")
+		os.Exit(1)
+	}
+
+	// Run tests for mountedDirectory only if --mountedDirectory flag is set.
+	setup.RunTestsForMountedDirectoryFlag(m)
+
+	testBucket := setup.TestBucket()
+
+	setup.SetTestBucket("integration-test-data-gcsfuse")
```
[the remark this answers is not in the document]
Tulsishah: Around 15 to 20 mins for One Test.
If we are going to create three different directory for three tests- it will take 50 to 60mins.
We can may be append directory structure like first 12000 files , then 100 explicit dirs and then for third test 100 implicit dirs. In that case may be it will take 20 mins to pass all the tests.

tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
Tulsishah: Fixed

tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
```diff
@@ -0,0 +1,158 @@
...
+	objNumberStr := strings.ReplaceAll(objName, prefix, "")
+	objNumber, err := strconv.Atoi(objNumberStr)
+	if err != nil {
+		t.Errorf("Error in extracting file number.")
+	}
+	return
+}
+
+func throwErrorForIncorrectFileNumber(fileNumber int, t *testing.T) {
+	if fileNumber < 1 || fileNumber > 12000 {
```
sethiay, line 38: you can pass the max number (100 & 12000 in these cases) as argument to checkIfObjNameIsCorrect and add this check there. That way you don't need to have throwErrorForIncorrectFileNumber & throwErrorForIncorrectDirNumber

tools/integration_tests/list_large_dir/list_large_dir_test.go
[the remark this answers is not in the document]
sethiay: are GCE VM and bucket in same zone ?

tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	if objNumber < 1 || objNumber > maxNumber {
+		t.Errorf("Correct object does not exist.")
+	}
+}
+
+func createTwelveThousandFilesAndUploadOnTestBucket(t *testing.T) {
+	// Creating twelve thousand files on disk to upload them on a bucket for testing.
+	for i := 1; i <= NumberOfFilesInDirectoryWithTwelveThousandFiles; i++ {
+		filePath := path.Join(os.Getenv("HOME"), PrefixFileInDirectoryWithTwelveThousandFiles+strconv.Itoa(i))
```
sethiay, line 43: I would suggest to create files inside a directory in $HOME.

tools/integration_tests/list_large_dir/testdata/create_implicit_dir.sh
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+a=1
+#Iterate the loop until a greater than 100
+touch testFile.txt
+while [ $a -lt 101 ]
+do
+   dir="implicitDirInLargeDirListTest"$a
```
sethiay, line 20: should we pass the name of this directory from test module ?

Tulsishah: I don't think so it is required as we are using this script for this tests only like creating 100 implicit directories. and if we want to do the same we required another runScript function which takes more arguments. but at this point I don't think so it is required.

sethiay: We are defining constant variable inside the go test package for implicitDirInLargeDirListTest. So if someone wants to change the value of that constant they would have to change it at two places.
Can the runScript function take dynamic list of arguments and pass the same to script ?

tools/integration_tests/util/operations/dir_operations.go
```diff
@@ -60,7 +61,7 @@ func RenameDir(dirName string, newDirName string) (err error) {
 }
 
 func CreateDirectoryWithNFiles(numberOfFiles int, dirPath string, prefix string, t *testing.T) {
-	err := os.Mkdir(dirPath, FilePermission_0600)
+	err := os.Mkdir(dirPath, FilePermission_0777)
```
Tulsishah, line 64: local directory does not give permission to move files from directory to bucket with 0600 permission. This permission is working with all tests.

sethiay: you can use `gsutil cp`

tools/integration_tests/list_large_dir/testdata/create_implicit_dir.sh
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+a=1
+#Iterate the loop until a greater than 100
+touch testFile.txt
+while [ $a -lt 101 ]
```
sethiay, line 18: 101 & implicitDirInLargeDirListTest can be passed from calling test package.

tools/integration_tests/list_large_dir/testdata/upload_twelve_thousand_files_to_bucket.sh
```diff
@@ -0,0 +1,24 @@
+# Copyright 2023 Google Inc. All Rights Reserved.
```
sethiay, line 1: can we rename it to upload_files_to_bucket.sh ? because this script is not doing anything special particularly to upload "12000" files.

tools/integration_tests/util/operations/dir_operations.go
Tulsishah: for that also it is having 
``` dir_operations.go:74: Create file at "/home/tulsishah_google_com/directoryWithTwelveThousandFiles": open /home/tulsishah_google_com/directoryWithTwelveThousandFiles/fileInDirectoryWithTwelveThousandFiles1: permission denied```

sethiay (approved):

PR #1181, opened 2023-06-13
fix integration test

ashmeenkaur:
### Description
while running integration tests with go version 1.21.x, we were seeing the following error: 
```
invalid value "$HOME/gcsfuse" for flag -C: -C flag must be first flag on command line
usage: go build [-o output] [build flags] [packages]
Run 'go help build' for details.
```
Fix: made -C the first flag on command line. These changes are compatible with go version 1.20.x and newer.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/build_gcsfuse/main.go
+++ b/tools/build_gcsfuse/main.go
@@ -42,6 +42,7 @@ import (
 	"os/exec"
 	"path"
 	"runtime"
+	"strings"
 )
 
 // Build release binaries according to the supplied settings, setting up the
@@ -164,6 +165,9 @@ func buildBinaries(dstDir, srcDir, version string, buildArgs []string) (err erro
 		output, err = cmd.CombinedOutput()
 		if err != nil {
 			err = fmt.Errorf("%v: %w\nOutput:\n%s", cmd, err, output)
+			if strings.Contains(string(output), "flag provided but not defined: -C") {
+				err = fmt.Errorf("%v: %w\nOutput:\n%s\nPlease upgrade to go version 1.20 or higher", cmd, err, output)
+			}
 			return
 		}
 	}
--- a/tools/build_gcsfuse/main.go
+++ b/tools/build_gcsfuse/main.go
@@ -133,10 +133,10 @@ func buildBinaries(dstDir, srcDir, version string, buildArgs []string) (err erro
 		cmd := exec.Command(
 			"go",
 			"build",
-			"-o",
-			path.Join(dstDir, bin.outputPath),
 			"-C",
-			srcDir)
+			srcDir,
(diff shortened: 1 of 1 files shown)


Tulsishah (approved):

tools/build_gcsfuse/main.go
```diff
@@ -133,10 +133,10 @@ func buildBinaries(dstDir, srcDir, version string, buildArgs []string) (err erro
 		cmd := exec.Command(
 			"go",
 			"build",
-			"-o",
-			path.Join(dstDir, bin.outputPath),
```
vadlakondaswetha, line 137: are daily tests also failing because of this?

Tulsishah: This issue visible only for cloud top. On VM it is working fine

Tulsishah (commented):
@ashmeenkaur Hope you have tested on GCE VM also. that won't break.

ashmeenkaur:
> @ashmeenkaur Hope you have tested on GCE VM also. that won't break.

Running into errors while running this on GCE VM. Closing this PR

tools/build_gcsfuse/main.go
ashmeenkaur: Daily tests are not failing because they use go version 1.20.4
We'll see this error only after the go version on our test script is updated to 1.21.x

Tulsishah: Please mention these thing in description.

vadlakondaswetha:
> > @ashmeenkaur Hope you have tested on GCE VM also. that won't break.
> 
> Running into errors while running this on GCE VM with go version < 1.20.

Is it saying in the error that issue is with the go version? if not can you add a log line

ashmeenkaur:
> I dont see its mentioning about go version anywhere. Can you add a log line mentioning that go version has to greater than specified version.

It doesn't specifically mentions that the go version has to be greater than some specific version. I came across this while experimenting. 

`go help build` command shows the following documentation.

For go version 1.21.x
```
The build flags are shared by the build, clean, get, install, list, run,
and test commands:

        -C dir
                Change to dir before running the command.
                Any files named on the command line are interpreted after
                changing directories.
                If used, this flag must be the first one in the command line.

```

For go version 1.20.x
```
The build flags are shared by the build, clean, get, install, list, run,
and test commands:

	-C dir
		Change to dir before running the command.
		Any files named on the command line are interpreted after
		changing directories.
```
Documentation of the flag missing in go version 1.19.x

vadlakondaswetha (approved):

PR #1197, opened 2023-06-22
Handle the error variables and filehandles properly across all integration tests

Tulsishah:
### Description
Adding ignored file handler and error.
1. Added clean mounted directory command to clean the bucket before testing.
2. Added file close for closing file.
3. After os.RemoveAll(mntDir) at the end ignore because anyway we are cleaning in the test starting.

NA

### Testing details
11. Unit tests - NA


tools/integration_tests/util/setup/setup.go
```diff
@@ -169,17 +174,12 @@ func UnMount() error {
...
-	os.RemoveAll(mntDir)
-
 	return successCode
 }
 
 func ExecuteTestForFlagsSet(flags []string, m *testing.M) (successCode int) {
 	var err error
 
-	// Clean the mountedDirectory before running any tests.
-	os.RemoveAll(mntDir)
```
Tulsishah, line 181: Removed this because for readonly tests it is throwing an error -> can't delete from readonly mount.
We are already cleaning mountedDirectory in most of the tests before testing, so I added it in each tests.

tools/integration_tests/util/setup/setup.go
```diff
@@ -116,7 +116,12 @@ func CreateTempFile() string {
 	if err != nil {
 		LogAndExit(fmt.Sprintf("Error in the opening the file %v", err))
 	}
-	defer file.Close()
+
+	defer func() {
```
sethiay, line 120: We should use the CloseFile defined in operations. If you are not using it because of cyclic dependency between setup.go and *_operations.go, then we shouldn't have operations package dependent on setup.

tools/integration_tests/util/mounting/mounting.go
```diff
@@ -34,7 +34,12 @@ func MountGcsfuse(flags []string) error {
 	if err != nil {
 		fmt.Println("Could not open logfile")
 	}
-	defer file.Close()
+	// Closing file at the end.
+	defer func() {
```
sethiay, line 38: Similar to the CloseFile comment.

tools/integration_tests/operations/move_file_test.go
```diff
@@ -78,8 +88,6 @@ func TestMoveFileWithinSameDirectory(t *testing.T) {
 	}
 
 	checkIfFileMoveOperationSucceeded(filePath, destDirPath, t)
-
-	os.RemoveAll(dirPath)
```
sethiay, line 82: why are we removing this ? (here and at all other places)

tools/integration_tests/util/setup/setup.go
```diff
@@ -234,3 +234,18 @@ func LogAndExit(s string) {
 	log.Print(s)
 	os.Exit(1)
 }
+
+// Clean the mounted directory
+func CleanMntDir() {
```
sethiay, line 239: what's the difference between CleanMntDir and os.removeall ? why are we making this shift now?

tools/integration_tests/util/setup/setup.go
sethiay: instead of making changes in every file, isn't it better to 
(a) remove os.RemoveAll from here, create a wrapper function over ExecuteTestForFlagsSet which does os.RemoveAll and then call ExecuteTestForFlagsSet. After that use wrapper function where you want to do cleanup and use ExecuteTestForFlagsSet where don't.
OR
(b) Pass a flag to ExecuteTestForFlagsSet which do cleanup only when true.

tools/integration_tests/operations/move_file_test.go
Tulsishah: So now we are already cleaning before running any tests. so there is no need of cleaning after running tests(that's what I feel)

Before we were cleaning for some set of tests where we required like (list, rename, move)

tools/integration_tests/util/setup/setup.go
Tulsishah: So before we were not catching the error for os.removeall. os.RemoveAll(mntDir) will remove content of directory and directory itself, but as we can't remove mntDir as we mounted it(it was cleaning the mntDir but at the end throwing error that we can't remove mntDir)

In CleanMntDir() --> I am just cleaning mntDir content and not mntDir itself.

Tulsishah: Swetha suggested that we should not ignore any error or file handler. That's why I am capturing it every place.

tools/integration_tests/operations/copy_dir_test.go
```diff
@@ -115,9 +121,6 @@ func TestCopyDirectoryInNonExistingDirectory(t *testing.T) {
 	}
 
 	checkIfCopiedDirectoryHasCorrectData(destDir, t)
-
-	os.RemoveAll(srcDir)
```
sethiay, line 119: why are we not cleaning up here ?

sethiay (approved):
As discussed, please check about getting all test functions of a test package programmatically and then calling cleanup before calling each test function.

tools/integration_tests/util/operations/dir_operations.go
```diff
@@ -68,9 +69,18 @@ func CreateDirectoryWithNFiles(numberOfFiles int, dirPath string, prefix string,
...
 		if err != nil {
 			t.Errorf("Create file at %q: %v", dirPath, err)
 		}
+
+		// Closing file at the end.
+		defer CloseFile(file)
+	}
+}
+
+func RemoveDir(dirPath string) {
```
vadlakondaswetha, line 82: can we put all these methods in one file. its difficult to find for someone new to find what methods exist

tools/integration_tests/operations/read_test.go
```diff
@@ -24,6 +24,9 @@ import (
 )
 
 func TestReadAfterWrite(t *testing.T) {
+	// Clean the mountedDirectory before running test.
+	setup.CleanMntDir()
```
vadlakondaswetha, line 28: can we do it at common place, instead of doing it in every test method.

tools/integration_tests/operations/copy_dir_test.go
```diff
@@ -219,7 +219,4 @@ func TestCopyDirectoryInNonEmptyDirectory(t *testing.T) {
 
 	destSrc := path.Join(destDir, SrcCopyDirectory)
 	checkIfCopiedDirectoryHasCorrectData(destSrc, t)
-
-	os.RemoveAll(srcDir)
```
vadlakondaswetha, line 223: why are these steps removed completely.

Tulsishah: As we are cleaning the bucket before every test begins, we don't need these steps now.

tools/integration_tests/util/operations/dir_operations.go
Tulsishah: It is in one file right. All the directory operations functions operations/dir_operations are in one place and all the file operations are in one place operations/file_operations

tools/integration_tests/operations/read_test.go
Tulsishah: I did POC for the same, but didn't get any thing concrete. Like we can add it I TestMain but it will execute once before whole package. Also I raised yaqs question for the same they were suggesting subtests, but it will require major code changes. and also we need to add more code lines compare to this one line cleanMntDir function. I discussed with ayush for the same.

vadlakondaswetha (approved):

PR #1220, opened 2023-07-11
Get dir attribute integration tests

Tulsishah:
### Description
Created tests for 
1. TestEmptyDirAttributes
2. TestNonEmptyDirAttributes
3. TestSubDirAttributes

[TestFileAttributes](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/operations/file_attributes_test.go#L27) test function was already there, just created common function to use for directory testing.

NA

### Testing details
11. Unit tests - NA


tools/integration_tests/operations/file_and_dir_attributes_test.go
```diff
@@ -0,0 +1,90 @@
...
+	preCreateTime := time.Now()
+	fileName := setup.CreateTempFile()
+	postCreateTime := time.Now()
+
+	// In the setup function we are writing 14 bytes.
+	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L124
+	fStat := checkIfObjectAttrIsCorrect(fileName, preCreateTime, postCreateTime, t)
+
+	// The file size in createTempFile() is 14 bytes
+	if fStat.Size() != 14 {
```
sethiay, line 63: Any reason for not checking this condition in checkIfObjectAttrIsCorrect ? Also, can you please create constant for 14 ?

sethiay: I meant we can also check this condition inside checkIfObjectAttrIsCorrect, right ?

Tulsishah: This is for file only. For directory it will be 0 only. As File test is only one, I include in that only.

sethiay: We can check what value we get for directory and then add check for that ?

Tulsishah: Fixed const changes

sethiay (approved):

PR #1231, opened 2023-07-18
Added new flag to read config file

vadlakondaswetha:
### Description
Added new flag to read config flag and logic to parse the config file

NA

### Testing details
1. Manual - Done testing for all the below combinations:
         1. with and without --foreground flag 
         2. No config file param
         3. full path and relative path
4. Unit tests - Added new unit tests for yaml parser
5. Integration tests - NA


internal/config/yaml_parser.go
```diff
@@ -0,0 +1,30 @@
+package config
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+func ReadConfigFile(fileName string) (mountConfig *MountConfig, err error) {
```
sethiay, line 10: nit: we can rename it to ParseConfigFile ?

main.go
```diff
@@ -272,6 +275,11 @@ func runCLIApp(c *cli.Context) (err error) {
 		return fmt.Errorf("parsing flags failed: %w", err)
 	}
 
+	mountOptions, err := config.ReadConfigFile(flags.ConfigFile)
```
sethiay, line 278: We can use  `mountConfig` instead of `mountOptions` like other places?

main.go
```diff
@@ -215,6 +217,7 @@ func mountWithArgs(
 		bucketName,
 		mountPoint,
 		flags,
+		mountConfig,
```
sethiay, line 220: Not sure on this but mountConfig sounds like (flags + bucketName + mountPoint). So, naming it like configFileFlags or configFileOptions might be better.

internal/config/yaml_parser.go
vadlakondaswetha: done

main.go
vadlakondaswetha: done

main.go
vadlakondaswetha: configFile doesn't sound appropriate. I went back and forth on the naming mountOptions, mountConfig etc. Went with mountConfig just because it is the config specified at mount time. 

Let me know if you have any other suggestion.

sethiay: Even I don't have any other suggestion.

sethiay (approved):

PR #1238, opened 2023-07-20
Minor update to CD pipeline test script

ashmeenkaur:
### Description
minor changes to the test script as follows:

- yum check-update command's exit code will be 100  when  there  are  updates available and is not necessary for yum updates.

- Logs are not captured incase a failure with non zero exit code occurs while running e2e tests.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -72,7 +72,7 @@ then
 else
 #  For rhel and centos
     sudo yum makecache
-    sudo yum check-update
+    sudo yum -y update
 
     #Install fuse
     sudo yum -y install fuse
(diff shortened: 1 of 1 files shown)


tools/cd_scripts/e2e_test.sh
```diff
@@ -107,6 +107,7 @@ cd gcsfuse
 git checkout $(sed -n 2p ~/details.txt) |& tee -a ~/logs.txt
 
 #run tests with testbucket flag
+set +e
```
sethiay, line 110: with this change, would the problem be solved if FAIL didn't some in logs.txt ?

ashmeenkaur: Changed the if condition to use exit status of the previous command (e2e test)

sethiay (approved):

PR #1239, opened 2023-07-21
Kokoro perf test fix - Integration tests

Tulsishah:
### Description
mount.gcsfuse was searching gcsfuse executables in the mentioned [places](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/mount_gcsfuse/find.go#L59). As gcsfuse was not installed in the machine where we are doing perf tests, it was failing.
Solution: Copied executables in /usr/local/bin dir 

It can run only with sudo permission so added function to copy with sudo permission.

NA

### Testing details
11. Unit tests - NA

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -175,7 +175,8 @@ func SetUpTestDir() error {
 // Removing bin file after testing.
 func RemoveBinFileCopiedForTesting() {
 	if !TestInstalledPackage() {
-		err := os.Remove("/usr/local/bin/gcsfuse")
+		cmd := exec.Command("sudo", "rm", "/usr/local/bin/gcsfuse")
+		err := cmd.Run()
 		if err != nil {
 			log.Printf("Error in removing file:%v", err)
 		}
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -148,6 +148,11 @@ func SetUpTestDir() error {
 		}
 		binFile = path.Join(TestDir(), "bin/gcsfuse")
 		sbinFile = path.Join(TestDir(), "sbin/mount.gcsfuse")
+
+		err := operations.CopyDir(binFile, "/usr/local/bin")
+		if err != nil {
+			log.Printf("Error in copying bin file:%v", err)
+		}
 	} else {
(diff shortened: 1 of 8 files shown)


tools/integration_tests/util/setup/setup.go
```diff
@@ -148,6 +148,14 @@ func SetUpTestDir() error {
 		}
 		binFile = path.Join(TestDir(), "bin/gcsfuse")
 		sbinFile = path.Join(TestDir(), "sbin/mount.gcsfuse")
+
+		// mount.gcsfuse will find gcsfuse executable in mentioned locations.
+		// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/mount_gcsfuse/find.go#L59
+		// Copying the executable to /usr/local/bin
+		err := operations.CopyDirWithRootPermission(binFile, "/usr/local/bin")
```
vadlakondaswetha, line 155: how was it working all these days then?

Tulsishah: [persistent mounting](https://cloud.google.com/storage/docs/gcsfuse-mount) integration tests got merged yesterday only. Tests are getting fail because of that.

tools/integration_tests/util/setup/setup.go
```diff
@@ -164,6 +172,17 @@ func SetUpTestDir() error {
 	return nil
 }
 
+// Removing bin file after testing.
+func RemoveBinFileCopiedForTesting() {
```
vadlakondaswetha, line 176: can we not call this method from unmount or executeTest below?

Tulsishah: No, because we are [building](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/operations/operations_test.go#L104) gcsfuse for every test package once only, unmount getting called after running test for every flag set. 
e.g.
[here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/operations/operations_test.go#L88) four times unmount getting called.
executeTest is getting called to run  test for flags for different mountings.
e.g [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/operations/operations_test.go#L106) for static mounting, only_dir_mounting,... - execute tests getting called
here

vadlakondaswetha (approved):

PR #1240, opened 2023-07-24
Sequential write large file integration tests

Tulsishah:
### Description
Adding tests to write 500MB files sequentially with 20MB chunk size.

NA

### Testing details
11. Unit tests - NA


tools/integration_tests/util/operations/file_operations.go
```diff
@@ -192,3 +194,38 @@ func ReadFileSequentially(filePath string, chunkSize int64) (content []byte, err
...
+	file, err := os.OpenFile(filePath, os.O_RDWR|syscall.O_DIRECT|os.O_CREATE, FilePermission_0600)
+	if err != nil {
+		log.Printf("Error in opening file:%v", err)
+	}
+
+	// Closing file at the end.
+	defer CloseFile(file)
+
+	for offset < fileSize {
+		// Get random 200MB or remaining filesize data into chunk.
```
raj-prince, line 211: 200 MB?

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -192,3 +194,38 @@ func ReadFileSequentially(filePath string, chunkSize int64) (content []byte, err
...
+		chunkSize = int64(math.Min(float64(chunkSize), float64(fileSize-offset)))
```
raj-prince, line 212: Why are we converting back and forth (int to float & float to int)?
Instead:
if ((fileSize - offset) < chunkSize) {
    chunkSize = (fileSize - offset);
}

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -192,3 +194,38 @@ func ReadFileSequentially(filePath string, chunkSize int64) (content []byte, err
 	}
 	return
 }
+
+func WriteFileSequentially(filePath string, fileSize int64, chunkSize int64) (err error) {
+	var numberOfBytes int
```
raj-prince, line 199: Why do we need to declare the variable on the top? We should always initialise, where the variable will be used.

tools/integration_tests/util/operations/file_operations.go
Tulsishah: chunk size is 200MB so we are first getting chunk size data or the data remaining to write in file(<200MB)

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -192,3 +194,38 @@ func ReadFileSequentially(filePath string, chunkSize int64) (content []byte, err
...
+		if err != nil {
+			log.Fatalf("error while generating random string: %s", err)
+		}
+
+		// Writes random 200MB or remaining filesize data into file.
+		numberOfBytes, err = file.Write(chunk)
+		if err != nil {
+			return
+		}
+		if int64(numberOfBytes) != chunkSize {
```
raj-prince, line 224: Ideally, here also we should throw error.

tools/integration_tests/write_large_files/write_one_large_file_sequentially_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"path"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
+)
+
+const FiveHundredMB = 500 * 1024 * 1024
+const FiveHundredMBFile = "fiveHundredMBFile.txt"
+const chunkSize = 200 * 1024 * 1024
```
raj-prince, line 28: Please always declare const in capital case.

Tulsishah: You mean all capital or start should be capital?

tools/integration_tests/write_large_files/write_one_large_file_sequentially_test.go
```diff
@@ -0,0 +1,51 @@
...
+		t.Errorf("Error in writing file: %v", err)
+	}
+
+	// Check if 500MB data written in the file.
+	fStat, err := os.Stat(filePath)
+	if err != nil {
+		t.Errorf("Error in stating file:%v", err)
+	}
+
+	if fStat.Size() != FiveHundredMB {
```
raj-prince, line 48: Is there any way to compare checksum to validate the written content? If possible we can fix the content written to the file. Currently we are writing random data so we can't keep a fixed checksum for that.

raj-prince (changes requested):
Suggested few changes.

raj-prince:
Also, is there any reason to choose the chunkSize as 200MB?

Tulsishah:
> Also, is there any reason to choose the chunkSize as 200MB?

In the read tests we did the same.

Chunks of 200MB from gcs, so better that we have file size > (200 * 2.5) MiB so that we read atleast 2 times and 0.5 so that we know what happens when we don't require full chunk, so 500 MiB should also work.

tools/integration_tests/write_large_files/write_one_large_file_sequentially_test.go
Tulsishah: We can compare two file content, but the problem is where will we keep the fix content of 500MB ? I think comparing filesize may be enough for this.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -192,3 +193,41 @@ func ReadFileSequentially(filePath string, chunkSize int64) (content []byte, err
...
+		if (fileSize - offset) < chunkSize {
+			chunkSize = (fileSize - offset)
+		}
+		chunk := make([]byte, chunkSize)
+		_, err = rand.Read(chunk)
+		if err != nil {
+			log.Fatalf("error while generating random string: %s", err)
+		}
+
+		var numberOfBytes int
```
raj-prince, line 219: Can't we do in the next line?
numberOfBytes, err **:=** file.Write(chunk)

tools/integration_tests/util/operations/file_operations.go
raj-prince: Okay, we can remove 200MB from the comment. As it's a generic method, only knows about chunkSize.

tools/integration_tests/write_large_files/write_one_large_file_sequentially_test.go
raj-prince: Okay.

tools/integration_tests/util/operations/file_operations.go
Tulsishah: No it is giving error.
```
../util/operations/file_operations.go:224:4: result parameter err not in scope at return
	../util/operations/file_operations.go:222:18: inner declaration of var err error
```

tools/integration_tests/write_large_files/write_one_large_file_sequentially_test.go
raj-prince: start should be capital. -> ChunkSize

Tulsishah: done

raj-prince:
> > Also, is there any reason to choose the chunkSize as 200MB?
> 
> In the read tests we did the same.
> 
> Chunks of 200MB from gcs, so better that we have file size > (200 * 2.5) MiB so that we read atleast 2 times and 0.5 so that we know what happens when we don't require full chunk, so 500 MiB should also work.

For read, default value of sequential-read-size-mb is 200MB. That's why we keep 200MB as chunk size. Here, we can reduce the chunk size around 20MB. As here, it depends on the end user, how much data wants to write before sync call.

Tulsishah:
> > > Also, is there any reason to choose the chunkSize as 200MB?
> > 
> > 
> > In the read tests we did the same.
> > Chunks of 200MB from gcs, so better that we have file size > (200 * 2.5) MiB so that we read atleast 2 times and 0.5 so that we know what happens when we don't require full chunk, so 500 MiB should also work.
> 
> For read, default value of sequential-read-size-mb is 200MB. That's why we keep 200MB as chunk size. Here, we can reduce the chunk size around 20MB. As here, it depends on the end user, how much data wants to write before sync call.

Done

raj-prince (approved):

PR #1241, opened 2023-07-24
Creds test error fixing Integration tests

Tulsishah:
### Description
As we are hardcoded [service account name](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/creds_tests/creds.go#L31), when two VMs belongs to the same projects tries to run the tests at a time it is failing due to service account creation and deletion inconsistency. 

Made changes to create a service account and ignore the error if "already exist." 

The flow is.
1. Create service account will create service account name creds-test-gcsfuse in the project. If the service account already exist it will use existing one only.
2. create_key_file script will create credentials in that service account for testing.
3. revoke_and_delete_creds will revoke permission and delete the crede which we have created in create_key_file script.

here we are not deleting service account at any place, but accidentally if anyone deleted or we want to run the tests in different project, it will create the service account with given name. as we are not deleting service account, it can be possible that creation of credentials may exceeded greater than limit(10). which can fail the tests. 

NA

### Testing details
1. Manual - Run all the tests for testbucket( Parallal on two VMs belongs to same project)  and mountedDirectory flags.(status:SUCCESS)
11. Unit tests - NA

### DIFF
--- a/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
+++ b/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set +e
 SERVICE_ACCOUNT=$1
 SERVICE_ACCOUNT_ID=$2
-# Create service account if already exist.
 gcloud iam service-accounts create $SERVICE_ACCOUNT --description="$SERVICE_ACCOUNT" --display-name="$SERVICE_ACCOUNT"
+if [ $? -eq 1 ]; then
+  echo "Service account exist."
+fi
--- a/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
+++ b/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
@@ -15,11 +15,13 @@
 
 SERVICE_ACCOUNT=$1
 
-gcloud iam service-accounts create $SERVICE_ACCOUNT --description="$SERVICE_ACCOUNT" --display-name="$SERVICE_ACCOUNT" &>> output.txt
+gcloud iam service-accounts create $SERVICE_ACCOUNT --description="$SERVICE_ACCOUNT" --display-name="$SERVICE_ACCOUNT" &>> ~/output.txt
 if [ $? -eq 1 ]; then
   if grep "already exists within project" output.txt; then
     echo "Service account exist."
+    rm ~/output.txt
   else
+    rm ~/output.txt
     exit 1
   fi
 fi
--- a/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
+++ b/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
@@ -14,6 +14,7 @@
 # Create service account if does not exist.
 
 SERVICE_ACCOUNT=$1
+
 gcloud iam service-accounts create $SERVICE_ACCOUNT --description="$SERVICE_ACCOUNT" --display-name="$SERVICE_ACCOUNT" &>> output.txt
 if [ $? -eq 1 ]; then
   if grep "already exists within project" output.txt; then
--- a/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
+++ b/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
@@ -14,7 +14,6 @@
 # Create service account if does not exist.
 
 SERVICE_ACCOUNT=$1
-SERVICE_ACCOUNT_ID=$2
 gcloud iam service-accounts create $SERVICE_ACCOUNT --description="$SERVICE_ACCOUNT" --display-name="$SERVICE_ACCOUNT" &>> output.txt
 if [ $? -eq 1 ]; then
   if grep "already exists within project" output.txt; then
--- a/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
+++ b/tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+# Create service account if does not exist.
 
 SERVICE_ACCOUNT=$1
 SERVICE_ACCOUNT_ID=$2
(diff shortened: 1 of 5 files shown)


tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh
```diff
@@ -12,9 +12,19 @@
...
+# Crete key file output
+# e.g. created key [key_id] of type [json] as [key_file_path] for [service_account]
+# capturing third word from the file to get key-id
+# e.g. capture [KEY_ID]
+KEY_ID=$(cat key_id.txt | cut -d " " -f 3)
+# removing braces
+# e.g. capture KEY_ID
+KEY_ID=${KEY_ID:1:40}
+
+gcloud iam service-accounts keys delete $KEY_ID --iam-account=$SERVICE_ACCOUNT
```
vadlakondaswetha, line 29: isn;t just deleting service account from the project sufficient?

vadlakondaswetha, line 29: isn;t just deleting service account from the project sufficient? why do you need to revoke access and then delete

Tulsishah: We can create at max 10 keys in service account, so if the limit exceeded it will not create credentials.

Tulsishah: And the command is deleting only key of the service account not the service account it self.

vadlakondaswetha: what is the flow here - are we creating a new service account or using the same one in all tests?
Max 10 keys gets applied to deleted keys as well?

tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh
```diff
@@ -12,9 +12,20 @@
...
 SERVICE_ACCOUNT=$1
 KEY_FILE=$2
+
 gcloud auth revoke $SERVICE_ACCOUNT
-gcloud iam service-accounts delete $SERVICE_ACCOUNT
+# Crete key file output
+# e.g. created key [key_id] of type [json] as [key_file_path] for [service_account]
+# capturing third word from the file to get key-id
+# e.g. capture [KEY_ID]
+KEY_ID=$(cat ~/key_id.txt | cut -d " " -f 3)
```
gargnitingoogle, line 24: I'm not aware of this code, so naive question.. Is `~/key_id.txt` supposed to be replaced by `$KEY_FILE` here ?

tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh
Tulsishah: The flow is.
1. Create service account will create service name creds-test-gcsfuse in the project. If the service account already exist it will use existing one only.
2. create_key_file script will create credentials in that service account for testing.
3. revoke_and_delete_creds will revoke permission and delete the crede which we have created in create_key_file script.

here we are not deleting service account at any place, but accidentally if anyone deleted or we want to run the tests in different project it will create the service account with given name. as we are not deleting service account it can be possible that creation of credentials may exceeded greater than limit(10). which can fail the tests.

tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh
gargnitingoogle, line 24: Shouldn't we guard here against absence of `~/key_id.txt` ?

```
if [ -f "~/.key_id.txt" ]; then
    ...
fi
```
```

Comment (gargnitingoogle) [reply tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh:24]: From the code, it looks like I'm wrong. Never mind.

File: tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh (hunk 1 above)
Comment (ashmeenkaur) [reply tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh:29]: Please correct me if I am wrong here @Tulsishah 
I think we are doing this so that the service account always exists (not deleted during, before or after the test execution). 
We're creating a key file associated with the service account, using it and then deleting the key file after use (instead of the service account itself) so that it doesn't effect other tests running on other VMs.

File: tools/integration_tests/util/creds_tests/testdata/create_service_account.sh [hunk 3]
```diff
@@ -11,12 +11,15 @@
...
 
 SERVICE_ACCOUNT=$1
-SERVICE_ACCOUNT_ID=$2
-# Delete service account if already exist.
-gcloud iam service-accounts delete $SERVICE_ACCOUNT_ID
+
+gcloud iam service-accounts create $SERVICE_ACCOUNT --description="$SERVICE_ACCOUNT" --display-name="$SERVICE_ACCOUNT" &>> output.txt
 if [ $? -eq 1 ]; then
-  echo "Service account does not exist."
+  if grep "already exists within project" output.txt; then
```
gargnitingoogle, line 20: I'm not sure I understand the logic here. `already exists within project` sounds like it's a case where we don't want to do anything further.. But why are we exiting with 1 (error) in the else case (not already existing account) ?

ashmeenkaur: I think we're exiting with 1 when `gcloud iam service-accounts create` fails with some other error apart from when the service account already exists within project.

Tulsishah: Yes, if we get any other error we will exit with exit code 1

tools/integration_tests/util/creds_tests/testdata/revoke_permission_and_creds.sh
Tulsishah: > what is the flow here - are we creating a new service account or using the same one in all runs? Max 10 keys gets applied to deleted keys as well?

No It will not apply to deleted keys.

tools/integration_tests/util/creds_tests/testdata/create_service_account.sh
gargnitingoogle: Okay. Makes sense.

gargnitingoogle (approved):
LGTM

vadlakondaswetha (approved):

PR #1250, opened 2023-07-28
 Random write integration tests

Tulsishah:
### Description
Adding random write integration tests.
Also making change to compare file content with the file downloaded from gcs bucket for both sequential and random writes.
rename file [write_one_large_file_sequentially_test.go] to seq_write_large_file

NA

### Testing details
2. Unit tests - NA
3. Integration tests - Run all the tests for testbucket and mountedDirectory flags.(status:SUCCESS)


tools/integration_tests/util/operations/download_file_from_bucket.sh
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+TEST_BUCKET=$1
+OBJECT_NAME=$2
+OBJECT_PATH_IN_DISK=$3
+
+gcloud alpha storage cp gs://$TEST_BUCKET/$OBJECT_NAME $OBJECT_PATH_IN_DISK 2>&1 | tee ~/output.txt
```
raj-prince, line 19: Why are we using alpha storage?

tools/integration_tests/util/operations/download_file_from_bucket.sh
```diff
@@ -0,0 +1,23 @@
...
+if grep "The following URLs matched no objects or files" ~/output.txt; then
+  echo "Object does not exist."
+  rm ~/output.txt
```
raj-prince, line 22: Why are we only deleting this file in case object doesn't exist?

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,94 @@
...
+			t.Errorf("Error in syncing file:%v", err)
+		}
+
+		writtenContent, err := operations.ReadChunkFromFile(filePath, ChunkSize, offset)
+		if err != nil {
+			t.Errorf("Error in reading file.")
+		}
+
+		// Download the file from a bucket in which we write the content.
+		fileInBucket := path.Join(os.Getenv("HOME"), FileDownloadedFromBucket)
```
raj-prince, line 78: For every random write call, we download the complete 500 MB object?

Is there any way to get the content of specific range through CLI?

tools/integration_tests/write_large_files/seq_write_large_file_test.go
```diff
@@ -39,13 +41,35 @@ func TestWriteLargeFileSequentially(t *testing.T) {
 		t.Errorf("Error in writing file: %v", err)
 	}
 
+	// Download the file from a bucket in which we write the content.
```
raj-prince, line 44: Create a method to compare two files and use that at appropriate places. You can refer to optimise it - https://github.com/udhos/equalfile/blob/90a7461a2a226fa62eca6e350b5207ae9c78fff8/equalfile.go#L105

tools/integration_tests/write_large_files/write_large_files_test.go
```diff
@@ -31,8 +31,8 @@ func TestMain(m *testing.M) {
 
 	setup.ExitWithFailureIfBothTestBucketAndMountedDirectoryFlagsAreNotSet()
```
raj-prince, line 32: We can simplify the logic here - bucket is mandatory now, checking that will be enough.

raj-prince (changes requested):
Please incorporate.

tools/integration_tests/util/operations/download_file_from_bucket.sh
Tulsishah: Because gcloud storage is not working on kokoro VMs

tools/integration_tests/write_large_files/write_large_files_test.go
Tulsishah: For operation package we can still pass only mountedDirectory flag.

tools/integration_tests/write_large_files/random_write_large_file_test.go
Tulsishah: I didn't find anyway which can download only chunk from the file. I don't think so we  can optimise it further. 
We can compare at end after all random write attempts, but checking it every time is good according to me, please let me know your views.

tools/integration_tests/util/operations/download_file_from_bucket.sh
Tulsishah: Script is not getting used anymore.

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,87 @@
...
+		t.Errorf("Open file for write at start: %v", err)
+		return
+	}
+
+	for i := 0; i < NumberOfRandomWriteCalls; i++ {
+		offset := rand2.Int63n(MaxWritableByteFromFile - MinWritableByteFromFile)
+
+		// Generate chunk with random data.
+		chunk := make([]byte, ChunkSize)
+		_, err := rand.Read(chunk)
```
gargnitingoogle, line 56: Shouldn't we check if the read-size matches chunkSize?

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			log.Fatalf("error while generating random string: %s", err)
```
gargnitingoogle, line 58: Why `log.Fatalf` and not `t.Fatalf` ?

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			t.Errorf("Error in writing randomly in file:%v", err)
+		}
+		if n != ChunkSize {
+			t.Errorf("Incorrect number of bytes written in the file.")
+		}
+
+		err = f.Sync()
+		if err != nil {
+			t.Errorf("Error in syncing file:%v", err)
```
gargnitingoogle, line 72: This this be `t.Fatalf(...)` . Why go ahead if file-sync failed?

tools/integration_tests/write_large_files/write_large_files_test.go
```diff
@@ -16,23 +16,44 @@
...
+func compareFileFromGCSBucketAndMntDir(gcsFile string, mntDirFile string, localFilePathToDownloadGcsFile string) (err error) {
+	err = operations.DownloadGcsObject(gcsFile, localFilePathToDownloadGcsFile)
+	if err != nil {
+		err = fmt.Errorf("Error in downloading object:%w", err)
+		return
+	}
+
+	diff, err := operations.DiffFiles(mntDirFile, localFilePathToDownloadGcsFile)
+	if diff != 0 {
+		err = fmt.Errorf("Download of GCS object %s) didn't match the Mounted local file (%s): %v", localFilePathToDownloadGcsFile, mntDirFile, err)
```
gargnitingoogle, line 38: Mismatched closing bracket after first %s.

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"syscall"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
+)
+
+const NumberOfRandomWriteCalls = 20
+const DirForRandomWrite = "dirForRandomWrite"
+const MinWritableByteFromFile = 0
```
gargnitingoogle, line 32: Is this variable really needed? All I see this being used for is being subtracted from file-size.

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+		// Generate chunk with random data.
+		chunk := make([]byte, ChunkSize)
+		_, err := rand.Read(chunk)
+		if err != nil {
+			log.Fatalf("error while generating random string: %s", err)
+		}
+
+		// Write data in the file.
+		n, err := f.WriteAt(chunk, offset)
```
gargnitingoogle, line 62: I am not sure if WriteAt can beyond the existing file-size. If it can't then this can fail unnecessarily if `MaxWritableByteFromFile + chunkSize` >= 500M.

tools/integration_tests/write_large_files/write_large_files_test.go
```diff
@@ -16,23 +16,44 @@
...
+		return
+	}
+
+	// Remove file after testing.
+	operations.RemoveFile(localFilePathToDownloadGcsFile)
```
gargnitingoogle, line 43: Move this up right after the successful download with a defer (`defer operations.RemoveFile(localFilePathToDownloadGcsFile)`), to avoid leaving this file behind in case the diff fails.

tools/integration_tests/write_large_files/seq_write_large_file_test.go
```diff
@@ -23,29 +23,36 @@ import (
...
 	if err != nil {
 		t.Errorf("Error in writing file: %v", err)
 	}
 
-	// Check if 500MB data written in the file.
-	fStat, err := os.Stat(filePath)
+	// Download the file from a bucket in which we write the content and compare with
+	// the file content we wrote in mntDir.
+	filePathInGcsBucket := path.Join(setup.TestBucket(), DirForSeqWrite, FiveHundredMBFile)
+	localFilePath := path.Join(os.Getenv("HOME"), FiveHundredMBFileForSeqWriteInLocalSystem)
```
gargnitingoogle, line 50: Avoid putting temporary files in home directory directly. Either use `/tmp` or a sub-directory within `/tmp` or on `$HOME`).

tools/integration_tests/write_large_files/seq_write_large_file_test.go
```diff
@@ -23,29 +23,36 @@ import (
...
 	if err != nil {
-		t.Errorf("Error in stating file:%v", err)
+		t.Errorf("Error:%v", err)
 	}
 
-	if fStat.Size() != FiveHundredMB {
-		t.Errorf("Expecred file size %v found %d", FiveHundredMB, fStat.Size())
-	}
+	// Clean up.
+	operations.RemoveDir(seqWriteDir)
```
gargnitingoogle, line 57: Move this as a deferred call right after this directory is successfully created to avoid unintentionally failing to delete, in case of other failures.

gargnitingoogle (commented):
Left some comments.

tools/integration_tests/write_large_files/write_large_files_test.go
```diff
@@ -16,23 +16,44 @@
...
 )
 
+func compareFileFromGCSBucketAndMntDir(gcsFile string, mntDirFile string, localFilePathToDownloadGcsFile string) (err error) {
+	err = operations.DownloadGcsObject(gcsFile, localFilePathToDownloadGcsFile)
+	if err != nil {
+		err = fmt.Errorf("Error in downloading object:%w", err)
+		return
+	}
+
+	diff, err := operations.DiffFiles(mntDirFile, localFilePathToDownloadGcsFile)
```
gargnitingoogle, line 36: Nit: DiffFiles loads the entire files into memory. These are both 500 MiB files, hence would have a 1 GiB requirement just for this step.

Tulsishah: So what changes you are suggesting?

tools/integration_tests/write_large_files/random_write_large_file_test.go
Tulsishah: This is nit reading the file, It is creating random data for writing in the file.

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func TestWriteLargeFileRandomly(t *testing.T) {
+	randomWriteDir := path.Join(setup.MntDir(), DirForRandomWrite)
+	err := os.Mkdir(randomWriteDir, setup.FilePermission_0600)
+	if err != nil {
+		t.Errorf("Error in creating directory:%v", err)
+	}
+	filePath := path.Join(randomWriteDir, FiveHundredMBFile)
+
+	f, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|syscall.O_DIRECT, setup.FilePermission_0600)
```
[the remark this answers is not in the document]
Tulsishah: We are not reading the file, we are creating random data in the chunk array and coping that array into file.

tools/integration_tests/write_large_files/random_write_large_file_test.go
gargnitingoogle: It's not about reading file. The function [rand.Read()](https://pkg.go.dev/crypto/rand#Read) also returns the bytes filled in the byte slice which we should verify. But the function also header also says: `On return, n == len(b) if and only if err == nil.`, so I guess err check is sufficient.

tools/integration_tests/write_large_files/random_write_large_file_test.go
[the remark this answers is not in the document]
gargnitingoogle: Oh yeah, the `rand.Read()` threw me off. As we're only writing to the file, do we need `os.O_RDWR` mode instead of os.O_WRONLY ?

tools/integration_tests/write_large_files/write_large_files_test.go
gargnitingoogle: No solution at the moment. Just wanted to point this out to know the high memory requirement for this operation.

gargnitingoogle: May be capture this in a comment for now, and we can address this later if it causes an issue.

gargnitingoogle (changes requested):

tools/integration_tests/write_large_files/random_write_large_file_test.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		t.Fatalf("Open file for write at start: %v", err)
+	}
+
+	// Clean up.
+	defer operations.RemoveDir(randomWriteDir)
```
gargnitingoogle, line 50: Move this up before the `os.OpenFile(filePath,...` so that even if file open fails, this directory gets deleted.

tools/integration_tests/write_large_files/seq_write_large_file_test.go
```diff
@@ -0,0 +1,59 @@
...
+	filePath := path.Join(seqWriteDir, FiveHundredMBFile)
+
+	// Sequentially read the data from file.
+	err = operations.WriteFileSequentially(filePath, FiveHundredMB, ChunkSize)
+	if err != nil {
+		t.Fatalf("Error in writing file: %v", err)
+	}
+
+	// Clean up.
+	defer operations.RemoveDir(seqWriteDir)
```
gargnitingoogle, line 49: Move this before `operations.WriteFileSequentially(filePath, FiveHundredMB, ChunkSize)` so that even if the file-write fails, this directory gets deleted.

gargnitingoogle (changes requested):

tools/integration_tests/write_large_files/random_write_large_file_test.go
gargnitingoogle: In general, when you do an operation and have a deferred undo, then those should not be separated by other operations.

E.g.

```golang
doSomething(A)
checkIfWorkedOrReturn(A)
defer undoSomething(A) // undoSomething(A) if put here, will be called even if B failed.

doSomething(B)
checkIfWorkedOrReturn(B)
defer undoSomething(A) // undoSomething(A) if put here, will not be called if B failed.
defer undoSomething(B)
...
```

gargnitingoogle (dismissed):
LGTM

gargnitingoogle (approved):
LGTM

vadlakondaswetha (approved):

Tulsishah:
> Please incorporate.

Done

tools/integration_tests/write_large_files/write_large_files_test.go
```diff
@@ -16,23 +16,50 @@
...
+		err = fmt.Errorf("Error in downloading object:%w", err)
+		return
+	}
+
+	// Remove file after testing.
+	defer operations.RemoveFile(localFilePathToDownloadGcsFile)
+
+	// DiffFiles loads the entire files into memory. These are both 500 MiB files, hence would have a 1 GiB
+	// requirement just for this step
+	diff, err := operations.DiffFiles(mntDirFile, localFilePathToDownloadGcsFile)
```
raj-prince, line 46: Ideally, we should compare chunk by chunk. 1 GiB RAM seems a little higher for RAM.

raj-prince: We can created another method, which takes a chunk size to compare. We can keep chunk-size as 16 MiB by default.

Tulsishah: We need to anyways downloading 500MB file from bucket to compare the content. Other 500MB file will be in mntDir.
In this DiffFile function we are just comparing two files. How chunk size will make any. difference?

Tulsishah: As discussed I'll raise bug for this.

raj-prince (approved):
Approving this, as DiffFiles (chunk-size) changes will taken as part of another PR.

PR #1254, opened 2023-08-02
gzip read compressed support and integration tests

gargnitingoogle:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/object.go
+++ b/internal/storage/object.go
@@ -31,4 +31,5 @@ type MinObject struct {
 	MetaGeneration int64
 	Updated        time.Time
 	Metadata       map[string]string
+	IsGzipEncoded  bool
 }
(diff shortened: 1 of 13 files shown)


internal/storage/object.go
```diff
@@ -31,4 +31,5 @@ type MinObject struct {
 	MetaGeneration int64
 	Updated        time.Time
 	Metadata       map[string]string
+	IsGzipEncoded  bool
```
gargnitingoogle, line 34: I first wondered if `IsGzipEncoded` should be a separate member or an item in `Metadata` dict. But as `Metadata` and `ContentEncoding` are separate in struct `gcs.Object`, it is clear that the same should be done here.

PR #1260, opened 2023-08-07
Updated SetInodeAttributes and GetInodeAttributes operations to support local files

vadlakondaswetha:
### Description
1. Updated SetInodeAttributes and GetInodeAttributes operations to support local files
2. Changed package of file_test.go to inode instead of inode_test to be able to read private variables for better validations.

NA

### Testing details
1. Manual - Tested manually by adding print statements in getAttributes flow to check if attributes are being returned correctly
3. Unit tests - Added whereever applicable. We cannot add unit tests for fs.go since fs.go tests do real mounting using fake gcs bucket. For any file operation, multiple operations will be invoked by kernel like lookUpInode, CreateFile, writeFile etc. Without modifying all these operations we will not be able to write unit tests. Hence unit tests will be added in a seperate cl when all operations are modified in fs.go
4. Integration tests - Existing tests are running fine. New tests will be added after all operations are modified.

### DIFF
--- a/internal/fs/inode/dir_test.go
+++ b/internal/fs/inode/dir_test.go
@@ -1052,7 +1052,7 @@ func (t *DirTest) DeleteChildFile_WrongGeneration() {
 	o, err := gcsutil.CreateObject(t.ctx, t.bucket, objName, []byte("taco"))
 	AssertEq(nil, err)
 
-	// Call the inode. inode with the wrong generation. No error should be returned.
+	// Call the inode with the wrong generation. No error should be returned.
 	err = t.in.DeleteChildFile(t.ctx, name, o.Generation+1, &o.MetaGeneration)
 	AssertEq(nil, err)
 
@@ -1072,7 +1072,7 @@ func (t *DirTest) DeleteChildFile_WrongMetaGeneration() {
 	o, err := gcsutil.CreateObject(t.ctx, t.bucket, objName, []byte("taco"))
 	AssertEq(nil, err)
 
-	// Call the inode. inode with the wrong meta-generation. No error should be
+	// Call the inode with the wrong meta-generation. No error should be
 	// returned.
 	precond := o.MetaGeneration + 1
 	err = t.in.DeleteChildFile(t.ctx, name, o.Generation, &precond)
--- a/internal/fs/inode/dir_test.go
+++ b/internal/fs/inode/dir_test.go
@@ -749,7 +749,7 @@ func (t *DirTest) CreateChildFile_DoesntExist() {
 	const name = "qux"
 	objName := path.Join(dirInodeName, name)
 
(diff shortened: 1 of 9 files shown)


internal/fs/inode/dir_test.go
```diff
@@ -68,7 +67,7 @@ func (t *DirTest) SetUp(ti *TestInfo) {
 		1, // Append threshold
 		".gcsfuse_tmp/",
 		bucket)
-	// Create the inode. No implicit dirs by default.
+	// Create the  No implicit dirs by default.
```
sethiay, line 70: nit: I think this happened by mistake ?

internal/fs/inode/file_test.go
```diff
@@ -514,6 +513,55 @@ func (t *FileTest) TruncateUpwardThenSync() {
...
+
+	AssertEq(nil, err)
+	// The inode should return the new size.
+	attrs, err = t.in.Attributes(t.ctx)
+	AssertEq(nil, err)
+	AssertEq(6, attrs.Size)
+	// Data shouldn't be updated to GCS.
+	statReq := &gcs.StatObjectRequest{Name: t.in.Name().GcsObjectName()}
+	_, err = t.bucket.StatObject(t.ctx, statReq)
+	AssertNe(nil, err)
```
sethiay, line 536: should we assert equal with Object not found error ?

internal/fs/inode/file_test.go
```diff
@@ -514,6 +513,55 @@ func (t *FileTest) TruncateUpwardThenSync() {
...
+	var err error
+	var attrs fuseops.InodeAttributes
+	t.createInodeWithLocalParam("test", true)
+	err = t.in.CreateEmptyTempFile()
+	AssertEq(nil, err)
+	err = t.in.Write(t.ctx, []byte("burrito"), 0)
+	AssertEq(nil, err)
+	// Validate the new data is written correctly.
+	attrs, err = t.in.Attributes(t.ctx)
+	AssertEq(nil, err)
```
sethiay, line 549: nit: I think the logical spacing can be improved here

sethiay (approved):

internal/fs/inode/dir_test.go
vadlakondaswetha: Reverted.

internal/fs/inode/file_test.go
vadlakondaswetha: I am trying to seperate the arrange, act, assert blocks by new line. Adding new lines with each block will not seperate out the blocks clearly. Added more comments to make it clear.

internal/fs/inode/file_test.go
vadlakondaswetha: done.

PR #1258, opened 2023-08-04
Deprecate --enable-storage-client-library flag 

Tulsishah:
### Description
Deprecating from enable storage client library flag 

### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - Did sanity check(create, read, delete, copy, move,... operations)
2. Unit tests - NA
3. Integration tests - Run perf tests using execute-perf-test label (kokoro tests)

PR #1264, opened 2023-08-10
Enabling --endpoint flag for go-storage-client library

raj-prince:
### Description
* Adding `--custom-endpoint` flag to connect with non-prod similar to GCS endpoint.
* Create client.go with unit test to separate all custom client creation logic.

NA

### Testing details
1. Manual - tested all the basic operations by setting up the [storage-testbench](https://github.com/googleapis/storage-testbench)
3. Unit tests - Added unit test; running successfully.
4. Integration tests - Run successfully.


flags.go
```diff
@@ -136,7 +136,8 @@ func newApp() (app *cli.App) {
 			cli.StringFlag{
 				Name:  "endpoint",
 				Value: "https://storage.googleapis.com:443",
-				Usage: "The endpoint to connect to.",
```
Tulsishah, line 139: Please add unit test for the same. E.g. correct format, Incorrect format.

internal/storage/storage_handle.go
```diff
@@ -91,8 +92,10 @@ func NewStorageHandle(ctx context.Context, clientConfig StorageClientConfig) (sh
 		wrapped:   httpClient.Transport,
 		UserAgent: clientConfig.UserAgent,
 	}
+
+	clientOpts := append(clientConfig.ClientOptions, option.WithHTTPClient(httpClient))
 	var sc *storage.Client
-	sc, err = storage.NewClient(ctx, option.WithHTTPClient(httpClient))
+	sc, err = storage.NewClient(ctx, clientOpts...)
```
Tulsishah, line 98: I think we need to add this parameter in [storage_handle_test ](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/storage_handle_test.go#L39)also

main_test.go
```diff
@@ -21,6 +23,64 @@ type MainTest struct {
 
 func init() { RegisterTestSuite(&MainTest{}) }
 
+func (t *MainTest) TestHandleCustomEndpointWithProdGCSEndpoint() {
+	url, err := url.Parse("https://storage.googleapis.com:443")
+	AssertEq(nil, err)
```
Tulsishah, line 28: nit: Empty line before and after assert statement

main_test.go
```diff
@@ -21,6 +23,64 @@ type MainTest struct {
...
+	err = handleCustomEndpoint(&fs, &storageClientConfig)
+	AssertEq(nil, err)
+
+	ExpectNe(nil, &storageClientConfig.TokenSrc)
+	ExpectEq(0, len(storageClientConfig.ClientOptions))
+}
+
+func (t *MainTest) TestHandleCustomEndpointWithNonProdGCSEndpoint() {
+	url, err := url.Parse("http://localhost:443")
+	AssertEq(nil, err)
```
Tulsishah, line 48: Here also

main_test.go
raj-prince: Ideally, we separate the arrange, act and assert block by new line. 

That assert is part of arrange block, that is why I haven't added any newline.

Adding empty line after the assert block. And adding comment to separate the `arrange`, `act` and `assert` block.

main_test.go
raj-prince: this will be handled as stated above.

flags.go
raj-prince: We can't add unit-test for that. Ideally, both behaviours are correct. I have notified to @tritone to check, if it's by design or need to fix.

In the end, this will be taken care during the retry integration test with the help of https://github.com/googleapis/storage-testbench.

internal/storage/storage_handle.go
raj-prince: Haven't added, since default value will be nil only. 

But I'll add unit for this also in storage_handle_test.go class 
Thanks!

main_test.go
Tulsishah: Got it thanks

main.go
```diff
@@ -148,24 +148,45 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
...
+		}
+	} else {
+
+		// Do not use OAuth with non-Google hosts.
+		tokenSrc = oauth2.StaticTokenSource(&oauth2.Token{})
+		opts = append(opts, option.WithEndpoint(flags.Endpoint.String()))
 	}
 
+	config.TokenSrc = tokenSrc
+	config.ClientOptions = opts
```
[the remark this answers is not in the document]
raj-prince: But that type is publicly exposed.

Anyway, the reason behind adding this instead of Endpoint url is we can pass multiple options from here. For e.g. we are passing WithJSONReads in this PR-https://github.com/GoogleCloudPlatform/gcsfuse/pull/1265 (incomplete). In future, also we don't have to pass different parameter for such options.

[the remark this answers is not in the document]
sethiay: Got it, in that case WithHTTPClient should also be passed in option.ClientOption instead of creating that in NewStorageHandle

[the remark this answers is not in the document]
raj-prince: Yeah, that will make the NewStorageHandle method clean, also simplify the StorageClientConfig object.

main_test.go
```diff
@@ -21,6 +25,98 @@ type MainTest struct {
...
+	ExpectNe(nil, httpClient)
+	ExpectNe(nil, httpClient.Transport)
+	ExpectEq(20, httpClient.Timeout)
+}
+
+func (t *MainTest) TestCreateHttpClientObjWithHttp2() {
+	fs := flagStorage{
+		KeyFile:           DummyKeyFile,
+		ClientProtocol:    mountpkg.HTTP2,
+		HttpClientTimeout: 20,
```
Tulsishah, line 108: Should we create const for HttpClientTimeout?

main_test.go
```diff
@@ -21,6 +25,98 @@ type MainTest struct {
...
+
+	ExpectEq(nil, err)
+	ExpectNe(nil, httpClient)
+	ExpectNe(nil, httpClient.Transport)
+	ExpectEq(20, httpClient.Timeout)
+}
+
 func (t *MainTest) TestCreateStorageHandleEnableStorageClientLibraryIsTrue() {
 	storageHandle, err := createStorageHandle(&flagStorage{
 		EnableStorageClientLibrary: true,
```
Tulsishah, line 122: Please update with DummyKeyFile const.

Tulsishah (approved):

main.go
```diff
@@ -148,22 +151,83 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
...
+
+	return httpClient, err
+}
+
+func isProdEndpoint(endpoint *url.URL) bool {
+	return (endpoint == nil) || (endpoint.Hostname() == GcsProdHostName)
+}
+
+func createTokenSource(flags *flagStorage) (tokenSrc oauth2.TokenSource, err error) {
+	if isProdEndpoint(flags.Endpoint) {
```
tritone, line 204: Why is this necessary? When would we expect to explicitly pass in the prod endpoint?

raj-prince: The default value of Endpoint is (`https://storage.googleapis.com/storage/v1`). See - [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/bb315266bb684dfd3ea9bed54cf530b467b1c023/flags.go#L136)

I'll make default value as nil, then we can remove that.

tritone: Sounds good, yes there is some other things that can go on under the hood with endpoints in the client (e.g. using MTLS) so it's better to not pass the endpoint unless you want a non-prod one.

main.go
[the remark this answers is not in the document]
raj-prince: Create client_helper to separate all client creation logic at one place.

internal/storage/storage_handle.go
```diff
@@ -42,57 +38,30 @@ type storageClient struct {
...
-			DisableKeepAlives: true,
-			MaxConnsPerHost:   clientConfig.MaxConnsPerHost,
-			ForceAttemptHTTP2: true,
+func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClientConfig) (sh StorageHandle, err error) {
+
+	var clientOpts []option.ClientOption
+
+	// Add WithHttpClient option.
+	if clientConfig.ClientProtocol == mountpkg.HTTP1 || clientConfig.ClientProtocol == mountpkg.HTTP2 {
+		httpClient, clientErr := storageutil.CreateHttpClientObj(&clientConfig)
```
vadlakondaswetha, line 50: why not use err directly here. why do you need seperate variable.

internal/storage/storageutil/client_helper.go
```diff
@@ -0,0 +1,110 @@
...
+		UserAgent: storageClientConfig.UserAgent,
+	}
+
+	return httpClient, err
+}
+
+// IsProdEndpoint GCSFuse assumes an endpoint as prod endpoint
+// if user haven't specified via --endpoint flag.
+// Also, we don't encourage use --endpoint flag to pass actual gcs url
+// in that case, some configuration is not set. E.g. using MTLS.
```
vadlakondaswetha, line 99: I didnt follow this comment. why can't users pass actual endpoint?

internal/storage/storageutil/client_helper.go
```diff
@@ -0,0 +1,110 @@
...
+	UserAgent           string
+	Endpoint            *url.URL
+	KeyFile             string
+	TokenUrl            string
+	ReuseTokenFromUrl   bool
+}
+
+// GetDefaultStorageClientConfig is only for test, making the default endpoint
+// non-nil, so that we can create dummy tokenSource while unit test.
+func GetDefaultStorageClientConfig() (clientConfig StorageClientConfig) {
```
vadlakondaswetha, line 35: dont put test and non-test code in the same file.

internal/storage/storageutil/client_helper_test.go
```diff
@@ -0,0 +1,95 @@
...
+	ExpectNe(nil, &tokenSrc)
+}
+
+func (t *clientHelperTest) TestCreateTokenSrcWithProdEndpoint() {
+	sc := GetDefaultStorageClientConfig()
+	sc.Endpoint = nil
+
+	tokenSrc, err := createTokenSource(&sc)
+
+	// This will fail with file not found error, since dummy file doesn't exist.
```
vadlakondaswetha, line 68: what are we testing with this test case?

internal/storage/storageutil/client_helper_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	res := IsProdEndpoint(url)
+
+	ExpectTrue(res)
+}
+
+func (t *clientHelperTest) TestCreateTokenSrcWithCustomEndpoint() {
+	url, err := url.Parse(CustomEndpoint)
+	AssertEq(nil, err)
+
```
vadlakondaswetha, line 52: [go/unit-testing-practices#structure](http://goto.google.com/unit-testing-practices#structure)

No new lines in arrange block

internal/storage/storage_handle.go
```diff
@@ -42,57 +38,30 @@ type storageClient struct {
...
-		Timeout: clientConfig.HttpClientTimeout,
+		clientOpts = append(clientOpts, option.WithHTTPClient(httpClient))
 	}
 
-	// Setting UserAgent through RoundTripper middleware
-	httpClient.Transport = &userAgentRoundTripper{
-		wrapped:   httpClient.Transport,
-		UserAgent: clientConfig.UserAgent,
+	// Add Custom endpoint option.
+	if !storageutil.IsProdEndpoint(clientConfig.Endpoint) {
```
vadlakondaswetha, line 60: what will happen if we always pass the endpoint?

main.go
```diff
@@ -149,21 +149,18 @@ func getConnWithRetry(flags *flagStorage) (c *gcsx.Connection, err error) {
 }
 
 func createStorageHandle(flags *flagStorage) (storageHandle storage.StorageHandle, err error) {
-	tokenSrc, err := auth.GetTokenSource(context.Background(), flags.KeyFile, flags.TokenUrl, true)
```
vadlakondaswetha, line 152: just curious why so much of refactoring being done just to construct auth?

vadlakondaswetha (changes requested):

internal/storage/storageutil/client_helper.go
raj-prince: Ref: https://screenshot.googleplex.com/89HEJFnx2HtvSuz

internal/storage/storage_handle.go
raj-prince: Yeah, we can do like this:
```go
var httpClient *http.Client
httpClient, err = storageutil.CreateHttpClientObj(&clientConfig)
if err != nil {
	err = fmt.Errorf("while creating http endpoint: %w", err)
	return
}
```

Other we use `:` in that case, it gives the below error: 
```internal/storage/storage_handle.go:54:4: result parameter err not in scope at return
        internal/storage/storage_handle.go:51:15: inner declaration of var err error
```

internal/storage/storage_handle.go
raj-prince: Ref: https://screenshot.googleplex.com/89HEJFnx2HtvSuz

internal/storage/storageutil/client_helper.go
raj-prince: Moving this method to test-file.

internal/storage/storageutil/client_helper_test.go
raj-prince: I just wanted to check, if it triggers the path of creation of actual token. As we create dummy token for custom-endpoint case.

main.go
raj-prince: Not only just to construct auth:
1. We need `IsProdEndPoint` check while creating Auth token as well as setting WithEndpoint() option. One of the motivation for this refactoring is just to keep this logic at same place.
2. The size of `main.go` was growing if we would kept all these logic in main.go file. Hence, move it together to storageutil package.
3. Also, we will support grpc client. I did all these refactoring in keeping those changes also in mind. And created a separate file for client creation in storageutil.

internal/storage/storageutil/client_helper.go
raj-prince: Ohh, we can't do, since this is used by test of storage package.

internal/storage/storageutil/client_helper.go
vadlakondaswetha: I still didnt follow the comment - why does MTLS will get missed if prod endpoint is passed.

internal/storage/storage_handle.go
vadlakondaswetha: I am sorry. didnt follow.

flags.go
```diff
@@ -134,9 +134,9 @@ func newApp() (app *cli.App) {
 			/////////////////////////
 
 			cli.StringFlag{
-				Name:  "endpoint",
```
vadlakondaswetha, line 137: can we rename the flag to non-prod-endpoint or custom-endpoint and mention that auth will be skipped for custom endpoints?

internal/storage/storageutil/client_helper.go
vadlakondaswetha: You can't put test and non-test code in the same file.

flags.go
raj-prince: Yeah, that would be good. Thanks!

internal/storage/storageutil/client_helper.go
```diff
@@ -0,0 +1,111 @@
...
+		RetryMultiplier:     2,
+		UserAgent:           "gcsfuse/unknown (Go version go1.20-pre3 cl/474093167 +a813be86df) (GCP:gcsfuse)",
+		Endpoint:            &url.URL{},
+		KeyFile:             DummyKeyFile,
+		TokenUrl:            "",
+		ReuseTokenFromUrl:   true,
+	}
+}
+
+func CreateHttpClientObj(storageClientConfig *StorageClientConfig) (httpClient *http.Client, err error) {
```
sethiay, line 51: nit: I believe we can omit Obj from the function name ?

internal/storage/storageutil/client_helper.go
```diff
@@ -0,0 +1,111 @@
...
+		UserAgent: storageClientConfig.UserAgent,
+	}
+
+	return httpClient, err
+}
+
+// IsProdEndpoint -  If user pass any url including (GCS prod) using --endpoint flag
+// GCSFuse assumes it as a custom endpoint. Hence, we don't encourage to use --endpoint
+// flag to pass actual GCS prod url, as in that case we also need to take care mTLS
+// option, which hasn't been handled properly in go-client-lib (by design of WithEndpoint
```
sethiay, line 99: can we add any reference link here where it is mentioned that the option is not handled properly ?

internal/storage/storageutil/client_helper_test.go
```diff
@@ -0,0 +1,96 @@
...
+func (t *clientHelperTest) TestCreateHttpClientObjWithHttp1() {
+	sc := GetDefaultStorageClientConfig() // By default http1 enabled
+
+	// Act: this method add tokenSource and clientOptions.
+	httpClient, err := CreateHttpClientObj(&sc)
+
+	ExpectEq(nil, err)
+	ExpectNe(nil, httpClient)
+	ExpectNe(nil, httpClient.Transport)
+	ExpectEq(sc.HttpClientTimeout, httpClient.Timeout)
```
sethiay, line 83: any particular reason for just comparing HttpClientTimeout ?

internal/storage/storageutil/client_helper.go
```diff
@@ -0,0 +1,111 @@
+package storageutil
```
sethiay, line 1: nit: given that client_helper.go is already in a package/directory that has util in its name i.e. storage"util". So do you also think we can simply call this file "client.go" ?

internal/storage/storageutil/client_helper.go
```diff
@@ -0,0 +1,111 @@
...
+// IsProdEndpoint -  If user pass any url including (GCS prod) using --endpoint flag
+// GCSFuse assumes it as a custom endpoint. Hence, we don't encourage to use --endpoint
+// flag to pass actual GCS prod url, as in that case we also need to take care mTLS
+// option, which hasn't been handled properly in go-client-lib (by design of WithEndpoint
+// option used by GCSFuse).
+func IsProdEndpoint(endpoint *url.URL) bool {
+	return endpoint == nil
+}
+
+func createTokenSource(storageClientConfig *StorageClientConfig) (tokenSrc oauth2.TokenSource, err error) {
```
sethiay, line 105: you can add brief comment describing what this function do

internal/storage/storageutil/client_helper.go
raj-prince: I didn't look into much details but - somewhere we override the MTLS - https://github.com/googleapis/google-api-go-client/blob/f7f1c7b4c610f6a41c130c3bb85d6a609e99fb37/internal/cba.go#L98

by default MTLS endpoint - https://storage.mtls.googleapis.com/storage/v1/ but this gets override with endpoint.

internal/storage/storage_handle.go
raj-prince: As per guidance by tritone@. We also need to take care of mtls endpoint if we set prod url WithEndpoint option.

Somewhere we override the MTLS endpoint with the provided endpoint - https://github.com/googleapis/google-api-go-client/blob/f7f1c7b4c610f6a41c130c3bb85d6a609e99fb37/internal/cba.go#L98

internal/storage/storageutil/client_helper_test.go
raj-prince: We can't validate other transport related configurations. Since http.Transport.Base is a RoundTripper interface. We can't access. E.g. MaxConnsPerHost etc.

internal/storage/storageutil/client_helper.go
raj-prince: Added a new file (test.go) containing test related method.

internal/storage/storageutil/client_helper.go
raj-prince: Removed this method, as renamed the endpoint to custom-endpoint.

internal/storage/storageutil/client.go
```diff
@@ -0,0 +1,102 @@
...
+	"net/url"
+	"time"
+
+	"github.com/googlecloudplatform/gcsfuse/internal/auth"
+	mountpkg "github.com/googlecloudplatform/gcsfuse/internal/mount"
+	"golang.org/x/net/context"
+	"golang.org/x/oauth2"
+)
+
+const CustomEndpoint = "https://localhost:9000"
```
sethiay, line 30: should this be moved to test.go ?

flags.go
```diff
@@ -134,9 +134,10 @@ func newApp() (app *cli.App) {
...
-				Value: "https://storage.googleapis.com:443",
-				Usage: "The endpoint to connect to.",
+				Name: "custom-endpoint",
+				Usage: "The endpoint to connect to - endpoint should be equivalent to the base endpoint " +
```
vadlakondaswetha, line 138: The end point to connect to for fetching data. The endpoint should be equivalent to the base endpoint of GCS JSON API.

internal/storage/storageutil/client.go
vadlakondaswetha, line 30: Are these test data. Move it from this file.

internal/storage/storageutil/client_test.go
```diff
@@ -0,0 +1,78 @@
...
+	sc := GetDefaultStorageClientConfig()
+	sc.CustomEndpoint = url
+
+	tokenSrc, err := createTokenSource(&sc)
+
+	ExpectEq(nil, err)
+	ExpectNe(nil, &tokenSrc)
+}
+
+func (t *clientTest) TestCreateTokenSrcWithProdEndpoint() {
```
vadlakondaswetha, line 44: TestCreateTokenSrcWhencustomEndpointIsNil

internal/storage/storageutil/test.go
```diff
@@ -0,0 +1,40 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storageutil
+
+import (
```
vadlakondaswetha, line 17: rename it to test_util.go. the name makes it clear that there are no tests but only utility methods.

internal/storage/storage_handle_test.go
```diff
@@ -98,27 +84,27 @@ func (t *StorageHandleTest) TestBucketHandleWhenBucketDoesNotExistWithNonEmptyBi
...
 
 func (t *StorageHandleTest) TestNewStorageHandleWithZeroMaxConnsPerHost() {
-	sc := getDefaultStorageClientConfig()
+	sc := storageutil.GetDefaultStorageClientConfig()
 	sc.MaxConnsPerHost = 0
 
 	t.invokeAndVerifyStorageHandle(sc)
 }
 
 func (t *StorageHandleTest) TestNewStorageHandleWhenUserAgentIsSet() {
```
sethiay, line 106: should we also add test for new fields added in StorageClientConfig?

main.go
```diff
@@ -89,7 +89,7 @@ func getUserAgent(appName string) string {
 
 func getConn(flags *flagStorage) (c *gcsx.Connection, err error) {
 	var tokenSrc oauth2.TokenSource
-	if flags.Endpoint.Hostname() == "storage.googleapis.com" {
+	if flags.CustomEndpoint.Hostname() == "storage.googleapis.com" {
```
sethiay, line 92: I think this `if` will not be entered by default because we are no longer keeping storage.googleapis.com. 
Anyways, we are deprecating jacobsa/gcloud. I would suggest you to merge your changes after that PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1258

internal/storage/storage_handle_test.go
raj-prince: Didn't add, since those are being tested as part of client_test.go.

sethiay: I think no because the StorageHandle is not created there.

main.go
raj-prince: Yeah, this change was just to ensure build success. I believe this function will not be exist after merging the mentioned PR. I'll wait for this.

flags.go
```diff
@@ -134,9 +134,10 @@ func newApp() (app *cli.App) {
 			/////////////////////////
 
 			cli.StringFlag{
-				Name:  "endpoint",
-				Value: "https://storage.googleapis.com:443",
-				Usage: "The endpoint to connect to.",
+				Name: "custom-endpoint",
+				Usage: "The end point to connect to for fetching data. The endpoint should be equivalent " +
```
vadlakondaswetha, line 138: Sorry for multiple iterations. can you rephrase like this:

Alternate endpoint for fetching data. Should be used only for testing purposes. The endpoint should be equivalent to the base endpoint of GCS JSON API (https://storage.googleapis.com/storage/v1). If not specified GCS endpoint will be used. Auth will be skipped for custom endpoint.

vadlakondaswetha (approved):

internal/storage/storage_handle_test.go
```diff
@@ -98,28 +86,69 @@ func (t *StorageHandleTest) TestBucketHandleWhenBucketDoesNotExistWithNonEmptyBi
...
+}
+
+func (t *StorageHandleTest) TestNewStorageHandleWhenReuseTokenUrlFalse() {
+	sc := storageutil.GetDefaultStorageClientConfig()
+	sc.ReuseTokenFromUrl = false
+
+	t.invokeAndVerifyStorageHandle(sc)
+}
+
+func (t *StorageHandleTest) TestNewStorageHandleWhenTokenUrlIsSet() {
```
sethiay, line 149: We can add one test to show that NewStorageHandle fails when client protocol is neither HTTP1 or HTTP2 until we support GRPC ?

raj-prince: As discussed offline, we can skip this. Right now, if user pass other than HTTP1 or HTTP2 it creates default client doesn't give any error.

raj-prince:
image

No regression.

flags.go
```diff
@@ -466,11 +466,19 @@ func resolvePathForTheFlagsInContext(c *cli.Context) (err error) {
...
-		return
+	customEndpointStr := c.String("custom-endpoint")
+	var customEndpoint *url.URL
+
+	if customEndpointStr == "" {
+		customEndpoint = nil
+	} else {
+		customEndpoint, err = url.Parse(customEndpointStr)
+		if err != nil {
+			fmt.Printf("Could not parse endpoint")
```
vadlakondaswetha, line 477: we shouldn't use fmt.Printf statements in code.

PR #1270, opened 2023-08-10
updated deb package architecture to all

ashmeenkaur:
### Description
Changed architecture in DEBIAN/control file to all so that it builds for both arm64 and amd64 architectures.
Changed package name in dockerfile based on the passed build argument. (By default amd64)

NA

### Testing details
1. Manual - Manual mount, file creation, deletion, update succeeds on the created arm64 and amd64 package.
2. Unit tests - NA
3. Integration tests - @Tulsishah verified and fixed the integration tests.

### DIFF
--- a/DEBIAN/control
+++ b/DEBIAN/control
@@ -3,7 +3,7 @@ Source: gcsfuse
 Maintainer: GCSFuse Team <gcs-fuse-maintainers@google.com>
 Homepage: https://github.com/GoogleCloudPlatform/gcsfuse
 Package: gcsfuse
-Architecture: amd64
+Architecture: all
 Depends: fuse
 Description: User-space file system for Google Cloud Storage.
  GCSFuse is a FUSE adapter that allows you to mount and access Cloud Storage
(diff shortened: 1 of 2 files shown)


vadlakondaswetha (commented):
Please update the description that existing packages - amd64 are tested and working fine

Tulsishah (commented):
It is breaking on [this](https://github.com/GoogleCloudPlatform/gcsfuse/tree/master/tools/integration_tests/mounting) integration test only with below error on arm64 machine.

```
../go/pkg/mod/github.com/jacobsa/fuse@v0.0.0-20230509090321-7263f3a2b474/fusetesting/stat_linux.go:31:9: cannot use sys.(*syscall.Stat_t).Nlink (variable of type uint32) as uint64 value in return statement
```

I raised a fix on fuse repo - https://github.com/jacobsa/fuse/pull/149

Tulsishah (approved):

vadlakondaswetha (approved):

PR #1272, opened 2023-08-10
Removing jacobsa gcloud dependencies

Tulsishah:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/gcloud/gcs/gcsutil/read_object.go
```diff
@@ -0,0 +1,62 @@
...
+	// Don't forget to close.
+	defer func() {
+		closeErr := rc.Close()
+		if closeErr != nil && err == nil {
+			err = fmt.Errorf("Close: %v", closeErr)
+		}
+	}()
+
+	// Read the contents.
+	n, err := rc.Read(contents)
```
Tulsishah, line 52: For reviewer: Made changes to here as io package is depricated and lint was failing.

PR #1275, opened 2023-08-13
Modified Sync, forgetInode  operations to handle localFiles

vadlakondaswetha:
### Description
Modified SyncFile, ForgetInode operations to handle local Files.

NA

### Testing details
1. Manual - Done manual testing and verified that file is getting uploaded to GCS correctly. Also verified forgetInode by manually throwing an error in writeFile flow. 
2. Unit tests - Added wherever applicable. Fs.go unit tests will be added after lookupinode operation is updated.
3. Integration tests -Yet to run the existing tests. New tests will be added in later PRs.

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -174,6 +174,7 @@ func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool) (o *g
 	if errors.As(err, &notFoundErr) {
 		err = nil
 		if f.IsLocal() {
+			// For localFile, it is expected that object doesn't exist in GCS.
 			return
 		}
 
@@ -582,6 +583,7 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 	// If we wrote out a new object, we need to update our state.
 	if newObj != nil && !f.localFileCache {
 		f.src = convertObjToMinObject(newObj)
+		// Convert localFile to nonLocalFile after it is synced to GCS.
 		if f.IsLocal() {
 			f.local = false
 		}
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -173,6 +173,10 @@ func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool) (o *g
 	var notFoundErr *gcs.NotFoundError
 	if errors.As(err, &notFoundErr) {
 		err = nil
+		if f.IsLocal() {
+			return
+		}
+
 		b = true
 		return
 	}
@@ -559,7 +563,7 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 	// Write out the contents if they are dirty.
 	// Object properties are also synced as part of content sync. Hence, passing
 	// the latest object fetched from gcs which has all the properties populated.
-	newObj, err := f.bucket.SyncObject(ctx, latestGcsObj, f.content)
+	newObj, err := f.bucket.SyncObject(ctx, f.Name().GcsObjectName(), latestGcsObj, f.content)
 
 	// Special case: a precondition error means we were clobbered, which we treat
 	// as being unlinked. There's no reason to return an error in that case.
@@ -578,6 +582,9 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 	// If we wrote out a new object, we need to update our state.
 	if newObj != nil && !f.localFileCache {
 		f.src = convertObjToMinObject(newObj)
+		if f.IsLocal() {
+			f.local = false
+		}
 		f.content.Destroy()
 		f.content = nil
 	}
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -385,7 +385,9 @@ func (f *FileInode) Attributes(
 	attrs.Atime = attrs.Mtime
 	attrs.Ctime = attrs.Mtime
 
-	// Avoid clobbered check and return the attributes for local files.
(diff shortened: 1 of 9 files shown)


internal/fs/inode/file.go
```diff
@@ -380,7 +385,9 @@ func (f *FileInode) Attributes(
 	attrs.Atime = attrs.Mtime
 	attrs.Ctime = attrs.Mtime
 
-	// Avoid clobbered check and return the attributes for local files.
+	// Clobbered check makes a stat call to GCS. Avoiding stat calls to GCS for
```
ashmeenkaur, line 388: Should we avoid this stat call incase of local file inode in Sync menthod (line 558) as well?

vadlakondaswetha: Before syncing file to GCS, we check if the object is still valid (i.e, generation/metageneration). In case of local files, we will check if the object is still not present on GCS(not created via any other flow). Hence stat call is required.

internal/gcsx/syncer.go
```diff
@@ -81,31 +82,48 @@ type fullObjectCreator struct {
...
+		MetadataMap[MtimeMetadataKey] = mtime.UTC().Format(time.RFC3339Nano)
+	}
+
+	var req *gcs.CreateObjectRequest
+	if srcObject == nil {
+		var precond int64
+		req = &gcs.CreateObjectRequest{
+			Name:                   fileName,
+			Contents:               r,
+			GenerationPrecondition: &precond,
```
ashmeenkaur, line 109: NIT: shall we pass 0 here instead?

internal/gcsx/append_object_creator.go
```diff
@@ -125,7 +126,9 @@ func (oc *appendObjectCreator) Create(
 		MetadataMap[key] = value
 	}
 
-	MetadataMap[MtimeMetadataKey] = mtime.Format(time.RFC3339Nano)
+	if mtime != nil {
```
ashmeenkaur, line 129: Please correct me if I am wrong here but I believe this flow is not modified by local file changes. Why did we add this check?

internal/gcsx/append_object_creator_test.go
```diff
@@ -89,8 +89,9 @@ func (t *AppendObjectCreatorTest) SetUp(ti *TestInfo) {
 func (t *AppendObjectCreatorTest) call() (o *gcs.Object, err error) {
 	o, err = t.creator.Create(
 		t.ctx,
+		t.srcObject.Name,
 		&t.srcObject,
-		t.mtime,
+		&t.mtime,
```
ashmeenkaur, line 94: Shall we modify CallsComposeObjects() and CallsComposeObjectsWithObjectProperties() test in this file to use local mtime instead and assert that the received metadata gcsfuse_mtime is in UTC format?

internal/gcsx/syncer_test.go
```diff
@@ -161,6 +162,63 @@ func (t *FullObjectCreatorTest) CallsCreateObjectsWithObjectProperties() {
 	ExpectEq("test_value", req.Metadata["test_key"])
 }
 
+func (t *FullObjectCreatorTest) CallsCreateObjectWhenSrcObjectIsNil() {
```
sethiay, line 165: nit: From a unit testing POV, should we add a test case where we pass non-nill srcObject and a fileName different from srcObject.Name ?

ashmeenkaur (approved):

internal/gcsx/append_object_creator_test.go
vadlakondaswetha: I didnt follow.  CallsComposeObjects() and CallsComposeObjectsWithObjectProperties() are already validating that mtime is being set correctly.

internal/gcsx/append_object_creator.go
vadlakondaswetha: Earlier mtime is a value now its a pointer which can be nil. hence validation is required.

internal/gcsx/append_object_creator_test.go
ashmeenkaur: We modified the Create method to use UTC time. However in the test we are still using this:
```
t.mtime = time.Now().Add(123 * time.Second).UTC()
// other code
ExpectEq(t.mtime.Format(time.RFC3339Nano), req.Metadata["gcsfuse_mtime"])
```

Shall we change this to:
```
t.mtime = time.Now().Add(123 * time.Second)
// other code
ExpectEq(t.mtime.UTC().Format(time.RFC3339Nano), req.Metadata["gcsfuse_mtime"])
```

internal/gcsx/syncer.go
```diff
@@ -81,31 +82,48 @@ type fullObjectCreator struct {
...
-		StorageClass:               srcObject.StorageClass,
+	// Any existing mtime value will be overwritten with new value.
+	if mtime != nil {
+		MetadataMap[MtimeMetadataKey] = mtime.UTC().Format(time.RFC3339Nano)
+	}
+
+	var req *gcs.CreateObjectRequest
+	if srcObject == nil {
+		var precond int64
+		req = &gcs.CreateObjectRequest{
```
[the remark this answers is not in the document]
sethiay: Yes, my bad.

internal/gcsx/syncer_test.go
vadlakondaswetha: I am not sure why should we add a test case for a flow which doesn't exist and check its not getting triggered.

sethiay (approved):

PR #1279, opened 2023-08-14
Modified lookUpInode operation to check localFileInodes

vadlakondaswetha:
### Description
Modified lookUpInode operation to check localFileInodes first before calling GCS. Added unit tests for fs.go

NA

### Testing details
1. Manual - Tested manually by mounting gcsfuse and checking if lookUpInode is returning correct inodeId. 
2. Unit tests - Added unit tests for fs.go
3. Integration tests - Yet to run

### DIFF
--- a/internal/fs/local_file_test.go
+++ b/internal/fs/local_file_test.go
@@ -23,6 +23,7 @@ func init() {
 }
 
 func (t *LocalFileTest) SetUpTestSuite() {
+	t.serverCfg.ImplicitDirectories = true
 	t.serverCfg.MountConfig = &config.MountConfig{
 		WriteConfig: config.WriteConfig{
 			CreateEmptyFile: false,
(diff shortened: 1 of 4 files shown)


internal/fs/local_file_test.go
```diff
@@ -0,0 +1,212 @@
...
+			map[string]string{
+				// File
+				"implicitFoo/bar": "",
+			}))
+
+	t.newFileShouldGetSyncedToGCSAtClose("implicitFoo/foo")
+}
+
+func (t *LocalFileTest) newFileShouldGetSyncedToGCSAtClose(fileName string) {
+	// Creating a file shouldn't create file on GCS.
```
ashmeenkaur, line 63: Shall we extract line 63 - 71 into a separate helper method and use it in other tests?

vadlakondaswetha: done

ashmeenkaur (approved):

sethiay (approved):

PR #1287, opened 2023-08-17
Support local files for ReadDir operations

ashmeenkaur:
### Description
Added support to list local files present in a directory.

NA

### Testing details
1. Manual - 
 * ls command on mounted directory with local files
 * ls command on explicit dir with local files
 * ls command on implicit dir with local files
 * ls -R command

here 
`l1, l2, l3, l4, l5` and `l6` are local files.
`c` and `d` are explicit directories
`foo` is implicit directory
error `ls: cannot access 'l*': No such file or directory` is because changes are required in lookupInode command.

 ```
 ashmeen@ashmeen:~/Documents/mnt$ ls -R
.:
ls: cannot access 'l1': No such file or directory
ls: cannot access 'l2': No such file or directory
c  d  foo  l1  l2

./c:
ls: cannot access './c/l3': No such file or directory
l3

./d:
ls: cannot access './d/l4': No such file or directory
ls: cannot access './d/l5': No such file or directory
l4  l5

./foo:
ls: cannot access './foo/l6': No such file or directory
bar  l6

 ```
3. Unit tests - Added
4. Integration tests - NA

### DIFF
--- a/internal/fs/dir_handle.go
+++ b/internal/fs/dir_handle.go
@@ -16,6 +16,7 @@ package fs
 
 import (
 	"fmt"
+	"path"
 	"sort"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/fs/inode"
@@ -58,8 +59,8 @@ type dirHandle struct {
 
 // Create a directory handle that obtains listings from the supplied inode.
 func newDirHandle(
-	in inode.DirInode,
-	implicitDirs bool) (dh *dirHandle) {
+		in inode.DirInode,
+		implicitDirs bool) (dh *dirHandle) {
 	// Set up the basic struct.
 	dh = &dirHandle{
 		in:           in,
@@ -67,7 +68,7 @@ func newDirHandle(
 	}
 
 	// Set up invariant checking.
-	dh.Mu = locker.New("DH." + in.Name().GcsObjectName(), dh.checkInvariants)
+	dh.Mu = locker.New("DH."+in.Name().GcsObjectName(), dh.checkInvariants)
 
 	return
 }
@@ -159,8 +160,10 @@ func fixConflictingNames(entries []fuseutil.Dirent) (err error) {
 //
 // LOCKS_REQUIRED(in)
 func readAllEntries(
-	ctx context.Context,
-	in inode.DirInode) (entries []fuseutil.Dirent, err error) {
+		ctx context.Context,
+		in inode.DirInode,
+		localFileInodes map[inode.Name]inode.Inode) (entries []fuseutil.Dirent, err error) {
+	//Read entries from GCS
 	// Read one batch at a time.
 	var tok string
 	for {
@@ -181,6 +184,14 @@ func readAllEntries(
 			break
 		}
 	}
+
+	//Append local file entries (not synced to GCS)
+	localEntries, err := localFileEntries(in, localFileInodes)
+	if err != nil {
+		err = fmt.Errorf("ReadLocalFileEntries: %w", err)
+		return
+	}
+	entries = append(entries, localEntries...)
 
 	// Ensure that the entries are sorted, for use in fixConflictingNames
 	// below.
@@ -218,15 +229,30 @@ func readAllEntries(
 	return
 }
 
+// localFileEntries lists the local files present in the directory.
+// Local means that the file is not yet present on GCS.
+func localFileEntries(d inode.DirInode, localFileInodes map[inode.Name]inode.Inode) (localEntries []fuseutil.Dirent, err error) {
+	for localInodeName, _ := range localFileInodes {
+		if localInodeName.IsDirectChildOf(d.Name()) {
+			entry := fuseutil.Dirent{
+				Name: path.Base(localInodeName.LocalName()),
+				Type: fuseutil.DT_File,
+			}
+			localEntries = append(localEntries, entry)
+		}
+	}
+	return
+}
+
 // LOCKS_REQUIRED(dh.Mu)
 // LOCKS_EXCLUDED(dh.in)
-func (dh *dirHandle) ensureEntries(ctx context.Context) (err error) {
+func (dh *dirHandle) ensureEntries(ctx context.Context, localFileInodes map[inode.Name]inode.Inode) (err error) {
 	dh.in.Lock()
 	defer dh.in.Unlock()
 
 	// Read entries.
 	var entries []fuseutil.Dirent
-	entries, err = readAllEntries(ctx, dh.in)
+	entries, err = readAllEntries(ctx, dh.in, localFileInodes)
 	if err != nil {
 		err = fmt.Errorf("readAllEntries: %w", err)
 		return
@@ -252,8 +278,9 @@ func (dh *dirHandle) ensureEntries(ctx context.Context) (err error) {
 // LOCKS_REQUIRED(dh.Mu)
 // LOCKS_EXCLUDED(du.in)
 func (dh *dirHandle) ReadDir(
-	ctx context.Context,
-	op *fuseops.ReadDirOp) (err error) {
+		ctx context.Context,
+		op *fuseops.ReadDirOp,
+		localFileInodes map[inode.Name]inode.Inode) (err error) {
 	// If the request is for offset zero, we assume that either this is the first
 	// call or rewinddir has been called. Reset state.
 	if op.Offset == 0 {
@@ -263,7 +290,7 @@ func (dh *dirHandle) ReadDir(
 
 	// Do we need to read entries from GCS?
 	if !dh.entriesValid {
-		err = dh.ensureEntries(ctx)
+		err = dh.ensureEntries(ctx, localFileInodes)
 		if err != nil {
 			return
 		}
--- a/internal/fs/handle/dir_handle.go
+++ b/internal/fs/handle/dir_handle.go
@@ -12,11 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package fs
+package handle
 
 import (
 	"fmt"
-	"path"
 	"sort"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/fs/inode"
@@ -27,8 +26,8 @@ import (
 	"golang.org/x/net/context"
 )
 
-// State required for reading from directories.
-type dirHandle struct {
+// DirHandle is the state required for reading from directories.
+type DirHandle struct {
 	/////////////////////////
 	// Constant data
 	/////////////////////////
@@ -57,12 +56,12 @@ type dirHandle struct {
 	entriesValid bool
 }
 
-// Create a directory handle that obtains listings from the supplied inode.
-func newDirHandle(
-		in inode.DirInode,
-		implicitDirs bool) (dh *dirHandle) {
+// NewDirHandle creates a directory handle that obtains listings from the supplied inode.
+func NewDirHandle(
+	in inode.DirInode,
+	implicitDirs bool) (dh *DirHandle) {
 	// Set up the basic struct.
-	dh = &dirHandle{
+	dh = &DirHandle{
 		in:           in,
 		implicitDirs: implicitDirs,
 	}
@@ -77,14 +76,14 @@ func newDirHandle(
 // Helpers
 ////////////////////////////////////////////////////////////////////////
 
-// Dirents, sorted by name.
+// Directory entries, sorted by name.
 type sortedDirents []fuseutil.Dirent
 
 func (p sortedDirents) Len() int           { return len(p) }
 func (p sortedDirents) Less(i, j int) bool { return p[i].Name < p[j].Name }
 func (p sortedDirents) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
 
-func (dh *dirHandle) checkInvariants() {
+func (dh *DirHandle) checkInvariants() {
 	// INVARIANT: For each i, entries[i+1].Offset == entries[i].Offset + 1
 	for i := 0; i < len(dh.entries)-1; i++ {
 		if !(dh.entries[i+1].Offset == dh.entries[i].Offset+1) {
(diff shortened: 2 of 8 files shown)


internal/fs/dir_handle.go
```diff
@@ -16,6 +16,7 @@ package fs
 
 import (
 	"fmt"
+	"path"
```
vadlakondaswetha, line 19: can you try moving dir_handle.go into handle package and write unit tests.
Having dir_handle in fs package doesn't make sense.

internal/fs/dir_handle.go
```diff
@@ -159,8 +160,10 @@ func fixConflictingNames(entries []fuseutil.Dirent) (err error) {
 //
 // LOCKS_REQUIRED(in)
 func readAllEntries(
-	ctx context.Context,
-	in inode.DirInode) (entries []fuseutil.Dirent, err error) {
+		ctx context.Context,
+		in inode.DirInode,
+		localFileInodes map[inode.Name]inode.Inode) (entries []fuseutil.Dirent, err error) {
+	//Read entries from GCS
```
vadlakondaswetha, line 166: nit: For all comments add space at the begining and full stop at the end.

internal/fs/local_file_test.go
```diff
@@ -0,0 +1,146 @@
...
+	f1, err := os.Create(path.Join(mntDir, "localFile1"))
+	AssertEq(nil, err)
+	f2, err := os.Create(path.Join(mntDir, "localFile2"))
+	AssertEq(nil, err)
+
+	// Attempt to list mntDir.
+	entries, err := os.ReadDir(mntDir)
+	AssertEq(nil, err)
+	info, err := entries[0].Info()
+	fmt.Println(entries[0].IsDir(), entries[0].Name(), entries[0].Type(), info)
```
vadlakondaswetha, line 65: nit: remove fmt.Println statements when you raise PR

internal/fs/local_file_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *LocalFileTest) TestReadDirWithEmptyLocalFile() {
+	// Create local files.
+	f1, err := os.Create(path.Join(mntDir, "localFile1"))
+	AssertEq(nil, err)
+	f2, err := os.Create(path.Join(mntDir, "localFile2"))
```
vadlakondaswetha, line 58: Please validate that file isnot created in GCS and created when file.close is called.

reference: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1279/files

internal/fs/local_file_test.go
```diff
@@ -0,0 +1,146 @@
...
+				// File
+				"implicitFoo/bar": "",
+			}))
+	f1, err := os.Create(path.Join(mntDir, "implicitFoo/file1"))
+	AssertEq(nil, err)
+	f2, err := os.Create(path.Join(mntDir, "implicitFoo/file2"))
+	AssertEq(nil, err)
+
+	// Attempt to list explicit directory.
+	entries, err := os.ReadDir(path.Join(mntDir, "implicitFoo/"))
```
vadlakondaswetha, line 128: do you want to extract all this common logic into seperate method.

vadlakondaswetha (changes requested):
Please fix all lint errors

internal/fs/dir_handle.go
```diff
@@ -218,15 +229,30 @@ func readAllEntries(
 	return
 }
 
+// localFileEntries lists the local files present in the directory.
+// Local means that the file is not yet present on GCS.
+func localFileEntries(d inode.DirInode, localFileInodes map[inode.Name]inode.Inode) (localEntries []fuseutil.Dirent, err error) {
```
vadlakondaswetha, line 234: Thinking loud,  when we implement that rmDir and renameDir doesnt support when localfiles are present, we might need this logic in fs.go. See if you need this logic and if yes move this to fs.go

ashmeenkaur: I initially added this method to DirInode, but realised that all the methods in DirInode today deals with GCS objects. That will be the most appropriate location from code reusability POV as well. Do you think it will be okay to move this to DirInode?

vadlakondaswetha: Yeah should be good, if that operations have dirInode with them.

internal/fs/handle/dir_handle_test.go
```diff
@@ -0,0 +1,138 @@
...
+		&t.clock,
+		true) // localFile
+	return
+}
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *DirHandleTest) EnsureEntriesWithLocalAndGCSFiles() {
```
vadlakondaswetha, line 108: Test cases when 
1. local fileinode and GCS object has same name
2. local fileinode and GCS directory has same name

internal/fs/inode/dir.go
```diff
@@ -126,6 +126,10 @@ type DirInode interface {
 		ctx context.Context,
 		name string,
 		isImplicitDir bool) (err error)
+
+	// localFileEntries lists the local files present in the directory.
+	// Local means that the file is not yet present on GCS.
+	LocalFileEntries(localFileInodes map[Name]Inode) (localEntries []fuseutil.Dirent)
```
vadlakondaswetha, line 132: should we call it getLocalFileEntries or ListLocalFileEntries. what is the go recommendation?

internal/fs/inode/dir_test.go
```diff
@@ -1210,3 +1229,30 @@ func (t *DirTest) CreateLocalChildFile_ShouldnotCreateObjectInGCS() {
...
+
+func (t *DirTest) LocalFileEntriesEmpty() {
+	localFileInodes := map[Name]Inode{}
+
+	entries := t.in.LocalFileEntries(localFileInodes)
+
+	AssertEq(0, len(entries))
+}
+
+func (t *DirTest) LocalFileEntriesWith2LocalChildFiles() {
```
vadlakondaswetha, line 1241: Add unit test when all non child files are passed. this is to ensure that empty list is returned and not error or nil

internal/fs/local_file_test.go
```diff
@@ -0,0 +1,159 @@
...
+		WriteConfig: config.WriteConfig{CreateEmptyFile: false},
+	}
+	t.fsTest.SetUpTestSuite()
+}
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *LocalFileTest) TestReadDirWithEmptyLocalFile() {
```
vadlakondaswetha, line 56: can you add test case for listing non-empty files and checkinng if the size is returned correctly.
This is our main usecase. The purpose of supporting ls is to show the size of the file.

internal/fs/local_file_test.go
```diff
@@ -0,0 +1,159 @@
...
+	// Verify entries received successfully.
+	AssertEq(2, len(entries))
+	t.verifyLocalFileEntry(entries[0], fileName1)
+	t.verifyLocalFileEntry(entries[1], fileName2)
+	// Close the local files.
+	t.closeFileAndValidateObjectContents(f1, fileName1, "")
+	t.closeFileAndValidateObjectContents(f2, fileName2, "")
+}
+
+func (t *LocalFileTest) TestReadDirForExplicitDirWithLocalFile() {
```
vadlakondaswetha, line 75: is it possible to write test for ls -R?

vadlakondaswetha (changes requested):

internal/fs/inode/dir.go
ashmeenkaur: The recommendation is to avoid "Get" prefix and to prefer starting the name with the noun directly go/go-style/decisions#getters

internal/fs/local_file_test.go
ashmeenkaur: This test case requires lookUpInode to be supported for local files. I'll add this test in a separate PR.

ashmeenkaur: added a TODO for the same. [link](https://github.com/GoogleCloudPlatform/gcsfuse/pull/1287/files#diff-7800585ad28fc6c239f5a8d71b053eab502f7f40d5c0c7d6b794d4a4c3386afcR211)

internal/fs/local_file_test.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+////////////////////////////////////////////////////////////////////////
+// Helpers
+////////////////////////////////////////////////////////////////////////
+
+func (t *LocalFileTest) createLocalFile(filePath string) (file *os.File) {
+	file, err := os.Create(filePath)
+	AssertEq(nil, err)
+	t.validateObjectNotFoundErr(filePath)
```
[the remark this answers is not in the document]
ashmeenkaur: path to mounted directory (mntDir) is not required in the path for GCS objects. It is only required for local files here.

I noticed that i was using path.join in every call to createLocalFile, so moving that logic to this method.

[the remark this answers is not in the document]
vadlakondaswetha: I didnt follow this. I am saying
os.Create should have /mntDir also in the path
and t.validateObjectNotFoundErr(filePath) shouldn't have /mntDir in the path.

Right now you are using same parameter fileName in both methods.

internal/fs/handle/dir_handle_test.go
```diff
@@ -0,0 +1,183 @@
...
+	t.validateEntry(t.dh.entries[0], "gcsObject1", fuseutil.DT_File)
+	t.validateEntry(t.dh.entries[1], "gcsObject2", fuseutil.DT_File)
+	t.validateEntry(t.dh.entries[2], "localFile1", fuseutil.DT_File)
+	t.validateEntry(t.dh.entries[3], "localFile2", fuseutil.DT_File)
+}
+
+func (t *DirHandleTest) EnsureEntriesWithSameNameLocalAndGCSFile() {
+	var err error
+	// Set up empty GCS objects.
+	_, err = gcsutil.CreateObject(t.ctx, t.bucket, "mnt/file1", nil)
```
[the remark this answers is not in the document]
ashmeenkaur: Mount is not the mounted directory. I should have used a better name here. Updating to use "directory" here instead.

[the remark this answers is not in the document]
ashmeenkaur: "directory" is a directory present on gcs. 
DirHandle holds a DirInode pointing to this "directory"

internal/fs/local_file_test.go
ashmeenkaur: Added the test

internal/fs/local_file_test.go
```diff
@@ -177,29 +245,135 @@ func (t *LocalFileTest) RandomWritesToLocalFile() {
...
-	AssertEq(nil, err)
-	t.validateObjectNotFoundErr(fileName)
+	// Recursively list mntDir/subDir/ directory.
+	cmd := exec.Command("ls", "-R", path.Join(mntDir, "subDir/"))
+	out, err := cmd.Output()
 
-	return
+	// Verify output.
+	AssertEq(nil, err)
+	AssertTrue(strings.Contains(string(out), "/subDir:\nbaseLocalFile\nexplicitFoo\nimplicitFoo"))
```
vadlakondaswetha, line 373: can you try using filepath.walk or filepath.walkDir to recursively all files under directory.  Creating one cache for all buckets
Refactoring the code to use same file-cache implementation

We can avoid assertion on a string

internal/fs/local_file_test.go
```diff
@@ -177,29 +255,177 @@ func (t *LocalFileTest) RandomWritesToLocalFile() {
...
+	t.closeFileAndValidateObjectContents(f1, "implicitFoo/"+FileName, "")
+	t.closeFileAndValidateObjectContents(f2, "implicitFoo/"+FileName2, "")
+}
 
-	return
+func (t *LocalFileTest) TestRecursiveListingWithLocalFiles() {
+	/* Structure
+	mntDir/
+		- baseLocalFile 			--- file
+		- explicitFoo/ 				--- directory
```
vadlakondaswetha, line 343: implicit??

internal/fs/local_file_test.go
```diff
@@ -177,29 +255,177 @@ func (t *LocalFileTest) RandomWritesToLocalFile() {
...
+			return err
+		}
+		// The object type is not directory.
+		if !dir.IsDir() {
+			return nil
+		}
+
+		objs, err := os.ReadDir(path)
+		if err != nil {
+			log.Fatal(err)
```
vadlakondaswetha, line 382: why only this log.Fatal?

internal/fs/local_file_test.go
```diff
@@ -177,29 +255,177 @@ func (t *LocalFileTest) RandomWritesToLocalFile() {
...
+
+		// Check if mntDir/implicitFoo/ has correct objects.
+		if path == mntDir+"/implicitFoo" {
+			// numberOfObjects = 2
+			if len(objs) != 2 {
+				return fmt.Errorf("incorrect number of objects in mntDir/implicitFoo/")
+			}
+
+			if objs[0].Name() != "bar" || objs[0].IsDir() != false ||
+				objs[1].Name() != implicitLocalFileName || objs[1].IsDir() != false {
```
vadlakondaswetha, line 419: can we use Assert statements here similar to your other statements?

internal/fs/local_file_test.go
ashmeenkaur: Changed to assert instead

internal/fs/local_file_test.go
ashmeenkaur: done

vadlakondaswetha (approved):

vadlakondaswetha (dismissed):

internal/fs/handle/dir_handle.go
```diff
@@ -182,11 +184,26 @@ func readAllEntries(
...
 	// below.
 	sort.Sort(sortedDirents(entries))
 
 	// Fix name conflicts.
+	// When a local file is synced to GCS but not removed from the local file map,
+	// the entries list will have two duplicate entries.
+	// To handle this scenario, we had 2 options:
+	// Option 1: [Selected]
+	// Throw an error while fixing conflicting names. The error will be fixed in
+	// subsequent ls calls.
```
vadlakondaswetha, line 201: this error will be fixed in subsequent ls calls assuming that entry will be removed from localFileInodes.

vadlakondaswetha (dismissed):

vadlakondaswetha (approved):

PR #1293, opened 2023-08-22
Added mandatory codeowners file for gcsfuse

raj-prince:
### Description
- Added Swetha & Ayush as mandatory reviewer for gcsfuse codebase. One approval is enough to merge the codebase.

Note: code owners must have write permissions for the repository.

NA

### Testing details
1. Manual - NA
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 # These owners will be the default owners for everything in
-# the repo. Unless a later match takes precedence.
+# the repo.
 *           @sethiay @vadlakondaswetha
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 # These owners will be the default owners for everything in
-# the repo. Unless a later match takes precedence,
+# the repo. Unless a later match takes precedence.
 *           @sethiay @vadlakondaswetha
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,4 @@
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
-*           @ashmeenkaur @gargnitingoogle @raj-prince @sethiay @vadlakondaswetha
-
-# All changes to Go files must be reviewed by the designated reviewer.
-*.go        @sethiay @vadlakondaswetha
+*           @sethiay @vadlakondaswetha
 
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# Global rule:
+*           @sethiay @vadlakondaswetha
```
vadlakondaswetha, line 2: does this work on all branches or only master?

raj-prince: We need to add a branch protection rule to configure it for other branches. Currently, only rule is for master branch.

.github/CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+*           @ashmeenkaur @gargnitingoogle @raj-prince @sethiay @vadlakondaswetha
+
+# All changes to Go files must be reviewed by the designated reviewer.
+*.go        @sethiay @vadlakondaswetha
```
vadlakondaswetha, line 6: Mandatory review by Ayush and Swetha should be for all files in the repo

.github/CODEOWNERS
vadlakondaswetha: can you add it for sept_release branch also

sethiay: Is it possible to specify branches using regex (e.g. master|*_release) ?

raj-prince: Yeah, we can specify the regex. There is no right way to combine both branches to a single rule.

Hence, I have created two rules:
(a) master
(b) *_release - (we need to make sure to create a branch ending with _release).

sethiay (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence.
```
vadlakondaswetha, line 2: do you want to update the comment?
Unless a later match takes precedence. - this doesn;t make sense now.

.github/CODEOWNERS
vadlakondaswetha: i am sorry, i can't where you added rule for sept_release branch also?

.github/CODEOWNERS
raj-prince: Removed that part.

.github/CODEOWNERS
raj-prince: https://screenshot.googleplex.com/9RRt9CmEPmXny5T

vadlakondaswetha (approved):

PR #1311, opened 2023-08-25
Unlink and RmDir changes to support local files

ashmeenkaur:
### Description
Changes to support unlink and remove directory operation on local files. 

These changes are in sync with non-local file system behaviour, i.e., 
- When FlushFile call is made on an unlinked file, 
  - the file is considered clobbered.
  - we don't return any error in this case.
  - we don't write the file to GCS.
  - ref: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/file.go#L168C35-L168C35
- When WriteFile call is made on an unlinked file, 
   - the write operation fails while writing. 
   - throws IO error while opening a reader.
   - ref: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/file.go#L217
   
<br> <br>

Changes for local files:
* Unlink will remove the local file entry from localFileInodes map.
* RmDir will check for entries in localFileInodes map first (before GCS) to check whether the directory is empty or not.
* syncFile will not write to GCS if the local file has been unlinked.
* WriteFile operation will throw IO error if the local file has been unlinked.

NA

### Testing details
1. Manual - Verified manually
- if empty local file is deleted before sync, no error is thrown at file.close() and file is not written to GCS for both local and synced files.
- all write calls to unlinked local file fails with IO error and file is not written to GCS on file.close() for both local and synced files.
3. Unit tests - Added unit tests for fs.go
4. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -977,11 +977,14 @@ func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
 	// Do not sync local file if it has been deleted from the localFileInodes map.
+	fs.mu.Lock()
 	_, ok := fs.localFileInodes[f.Name()]
 	if f.IsLocal() && !ok {
-		// Return without any error. This is in sync with POSIX file behaviour.
+		// Do not return any error. This is in sync with non-local file behaviour.
+		fs.mu.Unlock()
 		return
 	}
+	fs.mu.Unlock()
 
 	// Sync the inode.
 	err = f.Sync(ctx)
@@ -2062,6 +2065,16 @@ func (fs *fileSystem) WriteFile(
 
 	in.Lock()
 	defer in.Unlock()
+
+	// If inode is a local file inode and has been unlinked (deleted), do not serve
+	// WriteFile op and return IO error. This is in sync with non-local file behaviour.
+	fs.mu.Lock()
+	_, ok := fs.localFileInodes[in.Name()]
+	if in.IsLocal() && !ok {
+		fs.mu.Unlock()
+		return fmt.Errorf("no such file or directory: %w", syscall.EIO)
+	}
+	fs.mu.Unlock()
 
 	// Serve the request.
 	if err := in.Write(ctx, op.Data, op.Offset); err != nil {
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -934,6 +934,12 @@ func (fs *fileSystem) lookUpLocalFileInode(parent inode.DirInode, childName stri
 			child.Unlock()
 			continue
 		}
+		// Check if the local file inode has been unlinked?
+		fileInode, ok := child.(*inode.FileInode)
+		if ok && fileInode.IsUnlinked() {
+			child.Unlock()
+			continue
+		}
 
 		return
 	}
@@ -976,15 +982,12 @@ func (fs *fileSystem) lookUpOrCreateChildDirInode(
 func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
-	// Do not sync local file if it has been deleted from the localFileInodes map.
-	fs.mu.Lock()
-	_, ok := fs.localFileInodes[f.Name()]
-	if f.IsLocal() && !ok {
-		// Do not return any error. This is in sync with non-local file behaviour.
-		fs.mu.Unlock()
+	// syncFile can be triggered for unlinked files if the fileHandle is open by
+	// another user/terminal. Hence, ignoring the syncFile call for local file.
+	if f.IsLocal() && f.IsUnlinked() {
+		// Silently ignore the syncFile call. This is in sync with non-local file behaviour.
 		return
 	}
-	fs.mu.Unlock()
 
 	// Sync the inode.
 	err = f.Sync(ctx)
@@ -997,7 +1000,7 @@ func (fs *fileSystem) syncFile(
 	// Delete the entry from localFileInodes map and add it to generationBackedInodes.
 	fs.mu.Lock()
 	delete(fs.localFileInodes, f.Name())
-	_, ok = fs.generationBackedInodes[f.Name()]
+	_, ok := fs.generationBackedInodes[f.Name()]
 	if !ok {
 		fs.generationBackedInodes[f.Name()] = f
 	}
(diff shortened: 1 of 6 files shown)


internal/fs/fs.go
```diff
@@ -976,6 +976,13 @@ func (fs *fileSystem) lookUpOrCreateChildDirInode(
 func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
```
vadlakondaswetha, line 978: The call shouldn't reach here if the inode is not valid. In what scenario does it reach here?

internal/fs/fs.go
```diff
@@ -1884,6 +1901,16 @@ func (fs *fileSystem) Unlink(
 	parent.Lock()
 	defer parent.Unlock()
 
+	inodeName := inode.NewFileName(parent.Name(), op.Name)
+	fs.mu.Lock()
+	_, ok := fs.localFileInodes[inodeName]
```
vadlakondaswetha, line 1906: delete it from fs.inodes as well.

internal/fs/fs.go
ashmeenkaur: file.close() on local file makes a FlushFile call.

ashmeenkaur: 1. unlinking a synced file
  The calls made to gcsfuse are
    ```
    Unlink()
    SetInodeAttributes()
    LookUpInode()
    ForgetInode()
    ```

2. process 1: hold a file handle f.
 process 2: unlink the file from another process
 process 1: f.close()
    The calls made to gcsfuse are
    ```
    Unlink()
    SetInodeAttributes()
    LookUpInode()
    FlushFile()
    ReleaseFileHandle()
    ForgetInode()
    ```

2. process 1: hold a file handle f.
 process 2: unlink the file from another process
 process 1: f.write("content")
 process 1: f.close()
    The calls made to gcsfuse are
    ```
    Unlink()
    SetInodeAttributes()
    LookUpInode()
    WriteFile() -> returns Error: "no such file or directory"
    SetInodeAttributes()
    FlushFile()
    ReleaseFileHandle()
    ForgetInode()
    ```
Please also take a look at the updated PR description as well.

internal/fs/fs.go
ashmeenkaur: The ForgetInode() call will ensure the deletion from fs.inodes

internal/fs/fs.go
```diff
@@ -976,6 +976,16 @@ func (fs *fileSystem) lookUpOrCreateChildDirInode(
 func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
+	// Do not sync local file if it has been deleted from the localFileInodes map.
+	fs.mu.Lock()
+	_, ok := fs.localFileInodes[f.Name()]
+	if f.IsLocal() && !ok {
+		// Do not return any error. This is in sync with non-local file behaviour.
```
vadlakondaswetha, line 983: instead add a comment saying deletion from localFileInodes is done during unlink operation. SyncFile will be triggered for unlinked file if the fileHandle is open by another user/terminal. Hence ignoring the syncFile call.

internal/fs/fs.go
```diff
@@ -1629,6 +1639,16 @@ func (fs *fileSystem) RmDir(
 	//     https://github.com/GoogleCloudPlatform/gcsfuse/issues/9
 	//
 	//
+
+	// Check for local file entries.
+	localFileEntries := childDir.LocalFileEntries(fs.localFileInodes)
+	// Are there any local entries?
+	if len(localFileEntries) != 0 {
```
vadlakondaswetha, line 1646: This is just a validation right. when does localFileEntries be present in dir if unlink on files is called first?

internal/fs/local_file_test.go
```diff
@@ -113,11 +113,15 @@ func (t *LocalFileTest) validateObjectNotFoundErr(fileName string) {
...
 
-func (t *LocalFileTest) closeFileAndValidateObjectContents(f **os.File, fileName string, contents string) {
+func (t *LocalFileTest) closeLocalFile(f **os.File) error {
 	err := (*f).Close()
-	AssertEq(nil, err)
 	*f = nil
+	return err
+}
 
+func (t *LocalFileTest) closeFileAndValidateObjectContents(f **os.File, fileName string, contents string) {
```
vadlakondaswetha, line 122: why do you need double pointer here?

internal/fs/fs.go
vadlakondaswetha: Thanks for the details Ashmeen. Like i mentioned in earlier writeFlow will not always throw an error.

internal/fs/local_file_test.go
ashmeenkaur: for test cleanup, we need to mark *t.f1, *t.f2 as nil, hence we are passing a reference to pointer here. 
example: https://stackoverflow.com/a/8769095/14369827

internal/fs/fs.go
ashmeenkaur: When `os.RemoveAll` is called, RmDir call is made first followed by unlink calls.
Whereas in case of `rm -r` unlink calls are made first. 
Ref: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1311/files/3053d4a168dd5ccb7bcaceba933fd4451f21cd8e#diff-ef03141dd8338143c09a5b30ceb3d079d25015a96f8b5292ee29daa23348df2eR1590

internal/fs/fs.go
```diff
@@ -976,6 +976,16 @@ func (fs *fileSystem) lookUpOrCreateChildDirInode(
 func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
+	// Do not sync local file if it has been deleted from the localFileInodes map.
+	fs.mu.Lock()
+	_, ok := fs.localFileInodes[f.Name()]
```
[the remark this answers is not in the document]
ashmeenkaur: done

internal/fs/fs.go
ashmeenkaur: done

internal/fs/fs.go
```diff
@@ -2036,6 +2066,16 @@ func (fs *fileSystem) WriteFile(
 	in.Lock()
 	defer in.Unlock()
 
+	// If inode is a local file inode and has been unlinked (deleted), do not serve
+	// WriteFile op and return IO error. This is in sync with non-local file behaviour.
+	fs.mu.Lock()
```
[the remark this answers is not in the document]
ashmeenkaur: Thanks for the insights, Swetha. Updated the PR to ensure the same behaviour.

internal/fs/fs.go
```diff
@@ -934,6 +934,12 @@ func (fs *fileSystem) lookUpLocalFileInode(parent inode.DirInode, childName stri
 			child.Unlock()
 			continue
 		}
+		// Check if the local file inode has been unlinked?
+		fileInode, ok := child.(*inode.FileInode)
+		if ok && fileInode.IsUnlinked() {
+			child.Unlock()
+			continue
+		}
```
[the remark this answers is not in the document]
ashmeenkaur: Done

internal/fs/fs.go
```diff
@@ -914,13 +913,21 @@ func (fs *fileSystem) lookUpLocalFileInode(parent inode.DirInode, childName stri
...
 
 		if child == nil {
+			fs.mu.Unlock()
 			return
 		}
 
 		// If the inode already exists, we need to follow the lock ordering rules
 		// to get the lock. First get inode lock and then fs lock.
 		fs.mu.Unlock()
 		child.Lock()
```
vadlakondaswetha, line 923: I would suggest taking fs.mu.lock here instead of carefuly unlocking in every exit which can get skipped in future edits.

Just add a comment that we are taking the lock early to be able to use common defer function.

ashmeenkaur: Done

vadlakondaswetha (approved):

PR #1315, opened 2023-08-28
Create separate service account for each integration test

Tulsishah:
### Description
Create separate service account for each integration test. Create key file provide permission, wait for 10s sleep time, run integration test, delete key file and delete service account. 

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Run on kokoro machines


tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -42,23 +47,39 @@ func RunTestsForKeyFileAndGoogleApplicationCredentialsEnvVarSet(testFlagSet [][]
...
-	// Create service account
-	setup.RunScriptForTestData("../util/creds_tests/testdata/create_service_account.sh", NameOfServiceAccount)
+	// Service account id format is name@project-id.iam.gserviceaccount.com
+	serviceAccount := serviceAccountName + "@" + id + ".iam.gserviceaccount.com"
+	description := serviceAccount
+	displayName := serviceAccount
 
-	key_file_path := path.Join(os.Getenv("HOME"), "creds.json")
+	// Delete service account after testing.
+	defer operations.DeleteServiceAccount(serviceAccount)
```
sethiay, line 58: I believe even this is not fully robust because there is a limit of 100 SA per project. And it is possible that SA created as part of these tests are not deleted. 

So to keep things simple, would it possible to do the below:
1. Used a fixed SA as we are using now.
2. Not create new SA key everytime tests are run and just maintain one SA key.
This should not be a problem because I am assuming that whenever integration tests are run parallely,  they are run with different buckets which means we can grant and revoke permissions from the SA just for the particular bucket for which tests are run. E.g. if integrations tests are running from two Machines A & B parallely with Bucket b1 & b2 then the A's integration tests will grant and revoke permissions of bucket b1 only and not interfere with bucket b2. Similarly, B will only interfere with b2. Even though they will use same credentials.
And we can download the creds using https://cloud.google.com/iam/docs/keys-list-get.

sethiay: Sorry my bad, just found that GCP doesn't allow redownloading json key. So this won't work.

Tulsishah:
Closing it as @sethiay is working on this fix.

PR #1329, opened 2023-09-01
Flaky test fix

sethiay:
### Description
This PR is to fix the flakiness in integration tests by
1. Create fixed SA accounts in gcs-fuse-test and gcs-fuse-test-ml and store their credential keys in secret managers in gcs-fuse-test-ml & gcs-fuse-test projects. This credentials key can be accessed by default SA of both the GCP projects.  (Currently, we create and delete credentials everytime tests are run)
2. If the integration tests are not run from the gcsfuse test GCP projects, then the credentials test will simply not be executed and a message will be printed for the same.
3. Use 2 minutes of sleep time after the command for setting permissions to SA is executed. 2 mins is required because it usually takes 2 mins for access propagation as mentioned here: https://cloud.google.com/iam/docs/access-change-propagation

1 and 2 will avoid failures due to limit on number of credential keys a SA can have. 3 will fix the intermittent permission denied errors.

NA

### Testing details
1. Manual - Ran the integration tests locally 2 times.
2. Unit tests - NA
3. Integration tests - Got success in the Kokoro presubmit integration tests.


Tulsishah (commented):
We are creating "gcsfuse-integration-tests-data" bucket in both the projects and store the key right?

sethiay:
> We are creating "gcsfuse-integration-tests-data" bucket in both the projects and store the key right?

Yes

tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -21,50 +21,75 @@ import (
...
+	gcsKeyFilePath := path.Join(CredentialsBucketName, id+".json")
 
-	key_file_path := path.Join(os.Getenv("HOME"), "creds.json")
-
-	// Create credentials
-	setup.RunScriptForTestData("../util/creds_tests/testdata/create_key_file.sh", key_file_path, serviceAccount)
+	// Download credentials
+	err = operations.DownloadGcsObject(gcsKeyFilePath, localKeyFilePath)
+	if err != nil {
+		log.Printf("Error while downloading key file %v", err)
```
Tulsishah, line 64: Should't it be log.Fatalf as if key will not download it will exit from that stat itself with exit code 1.

tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -21,50 +21,75 @@ import (
...
+	if err != nil {
+		log.Printf("Error while setting permissions to SA: %v", err)
+		return 1
+	}
+	// Waiting for 2 minutes as it usually takes within 2 minutes for policy
+	// changes to propagate: https://cloud.google.com/iam/docs/access-change-propagation
+	time.Sleep(120 * time.Second)
+
+	// Revoke the permission after testing.
+	defer func(format string) {
```
Tulsishah, line 79: Should we keep this thing in function for better readability and call that function defer function.

raj-prince:
`Create fixed SA accounts in gcs-fuse-test and gcs-fuse-test-ml and store their credential keys in a bucket in gcs-fuse-test-ml project.` - I am not sure, bucket is a good place to keep the credential. Have we considered using Secret Manager?

tools/integration_tests/util/creds_tests/creds.go
sethiay: As discussed, we are returning successCode=1, so the tests will fail.

Tulsishah (dismissed):

tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -18,53 +18,87 @@ package creds_tests
...
 
-const NameOfServiceAccount = "creds-test-gcsfuse"
+const NameOfServiceAccount = "creds-integration-tests"
+const CredentialsSecretName = "gcsfuse-integration-tests"
 
-func setPermission(permission string, serviceAccount string) {
-	// Provide permission to the bucket.
-	setup.RunScriptForTestData("../util/creds_tests/testdata/provide_permission.sh", setup.TestBucket(), serviceAccount, permission)
-}
+var WhitelistedGcpProjects = []string{"gcs-fuse-test", "gcs-fuse-test-ml"}
```
raj-prince, line 39: What's the rationale behind adding this whitelisting logic?

tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
```diff
@@ -101,6 +101,8 @@ func generateRandomString(length int) string {
 }
 
 func RunTests(flags [][]string, m *testing.M) (successCode int) {
+	log.Println("Running dynamic monting tests...")
```
raj-prince, line 104: nit: mounting*

tools/integration_tests/util/creds_tests/creds.go
sethiay: GCSFuse has only two GCP projects where test can be run. gcs-fuse-test is currently used for running CD tests and gcs-fuse-test-ml is used for CI tests.
If the tests are run on other projects, not the tests will not fail but a message will be printed in stdout that credentials tests are not run.

raj-prince (approved):

Tulsishah (approved):

vadlakondaswetha (approved):

PR #1331, opened 2023-09-04
Add yaml config for logging changes

ashmeenkaur:
### Description
Added yaml config for log severity.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA


internal/config/yaml_parser.go
```diff
@@ -17,10 +17,25 @@ package config
...
 	"os"
+	"strings"
 
 	"gopkg.in/yaml.v3"
 )
 
+func IsValidLogSeverity(severity string) bool {
+	switch severity {
+	case
+		"TRACE",
```
vadlakondaswetha, line 28: can we define an go-lang enum type instead of using as string everywhere. I am assuming when you write logs, you need to refer to these strings.

internal/config/testdata/valid_log_config.yaml
```diff
@@ -0,0 +1,4 @@
+write:
+  create-empty-file: true
+log:
```
vadlakondaswetha, line 3: lets use common file for all valid config and common testcase
https://github.com/GoogleCloudPlatform/gcsfuse/blob/tmp_sept_release/internal/config/testdata/valid_config.yaml

vadlakondaswetha (changes requested):

internal/config/mount_config.go
```diff
@@ -18,8 +18,13 @@ type WriteConfig struct {
 	CreateEmptyFile bool `yaml:"create-empty-file"`
 }
 
+type LogConfig struct {
+	Severity string `yaml:"severity"`
```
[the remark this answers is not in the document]
ashmeenkaur: cloud logging uses severity, so I thought we should keep it same here. (https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry)

Please let me know what will be a better name here.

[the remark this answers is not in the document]
vadlakondaswetha: SG

internal/config/yaml_parser.go
ashmeenkaur: done

internal/config/testdata/valid_log_config.yaml
ashmeenkaur: done

vadlakondaswetha (approved):

PR #1348, opened 2023-09-06
Update architecture in DEBIAN control file while building docker

Tulsishah:
### Description
Updating building package docker file to add architecture in control file
Currently due to [Architecture](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/DEBIAN/control#L6) value all, the deb package gets uploaded on  all the cloud [repos](https://packages.cloud.google.com/apt/dists/gcsfuse-jessie/main)(amd64, arm64)

NA

### Testing details
1. Manual - Tested locally by building docker image and uploading it on artifact registary.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/package_gcsfuse_docker/Dockerfile
+++ b/tools/package_gcsfuse_docker/Dockerfile
@@ -37,7 +37,7 @@ ARG DEBFULLNAME="GCSFuse Team"
 # Build Arg for building through a particular branch/commit. By default, it uses
 # the tag corresponding to passed GCSFUSE VERSION
 ARG BRANCH_NAME="v${GCSFUSE_VERSION}"
-# RUN git checkout "${BRANCH_NAME}"
+RUN git checkout "${BRANCH_NAME}"
 
 # Install fpm package using bundle
 RUN bundle install --gemfile=${GCSFUSE_PATH}/tools/gem_dependency/Gemfile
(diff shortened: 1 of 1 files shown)


tools/package_gcsfuse_docker/Dockerfile
```diff
@@ -60,6 +60,10 @@ RUN mv ${GCSFUSE_BIN}/DEBIAN/copyright ${GCSFUSE_DOC} &&  \
 # Update gcsfuse version in changelog and control file
 RUN sed -i "1s/.*/gcsfuse (${GCSFUSE_VERSION}) stable; urgency=medium/" ${GCSFUSE_DOC}/changelog && \
     sed -i "1s/.*/Version: ${GCSFUSE_VERSION}/" ${GCSFUSE_BIN}/DEBIAN/control
+
+# Update ARCHITECTURE in control file.
+RUN sed -i "6s/.*/Architecture: ${ARCHITECTURE}/" ${GCSFUSE_BIN}/DEBIAN/control
```
ashmeenkaur, line 65: I think we should add a check here to ensure that the passed architecture is arm64 or amd64 only.

Tulsishah: I don't think so it is require, what ever architecture user will pass it will update accordingly. In future if we support i386 architecture, we can pass that also.

ashmeenkaur: what if someone makes a typo in the passed architecture? 
Although the process is fairly automated now, but, i still think it is still good to add a check.

ashmeenkaur (approved):

sethiay (approved):

PR #1351, opened 2023-09-06
use default logger instead of separate debug_mutex logger [slog PR-7]

ashmeenkaur:
### Description
As we are enabling log-levels, removed debug_mutex logger from locker.go and logged via default logger methods instead.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/locker/locker.go
+++ b/internal/locker/locker.go
@@ -91,7 +91,7 @@ func (d *debugger) Lock() {
 	d.holder = string(buf)
 
 	d.timer = time.AfterFunc(5*time.Second, func() {
-		logger.Debugf("debug_mutex: Potential dead lock detected for a lock %q held by: %v\n", d.name, d.holder)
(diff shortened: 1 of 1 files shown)


internal/locker/locker.go
```diff
@@ -94,7 +91,7 @@ func (d *debugger) Lock() {
 	d.holder = string(buf)
 
 	d.timer = time.AfterFunc(5*time.Second, func() {
-		d.logger.Printf("Potential dead lock detected for a lock %q held by: %v\n", d.name, d.holder)
+		logger.Debugf("debug_mutex: Potential dead lock detected for a lock %q held by: %v\n", d.name, d.holder)
```
vadlakondaswetha, line 94: I am approving these PRs for now. But we need a better way to append prefixes. Please create a bug and track this.

vadlakondaswetha (approved):

vadlakondaswetha (approved):

PR #1355, opened 2023-09-06
Updating script to test arm64 machine 

Tulsishah:
### Description
1. Adding check to run installtion for amr64 on latest version as arm64 does not have any previous versions
2. Used `apt-transport-artifact-registry-unstable main` as stable one does not supports arm64, will switch back to stable once they support
3. Used `dnf-plugin-artifact-registry` as dnf supports both architecture.

NA

### Testing details
1. Manual - tested on louhi pipeline.
5. Unit tests - NA
6. Integration tests - NA


tools/cd_scripts/install_test.sh
```diff
@@ -22,12 +22,20 @@ gsutil cp  gs://gcsfuse-release-packages/version-detail/details.txt .
...
+  architecture="arm64"
+fi
+
 # Based on the os type(from vm instance name) in detail.txt, run the following commands to install apt-transport-artifact-registry
 if grep -q ubuntu details.txt || grep -q debian details.txt;
 then
 #  For ubuntu and debian os
     curl https://us-central1-apt.pkg.dev/doc/repo-signing-key.gpg | sudo apt-key add - && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
-    echo 'deb http://packages.cloud.google.com/apt apt-transport-artifact-registry-stable main' | sudo tee -a /etc/apt/sources.list.d/artifact-registry.list
+    echo 'deb http://packages.cloud.google.com/apt apt-transport-artifact-registry-unstable main' | sudo tee -a /etc/apt/sources.list.d/artifact-registry.list
```
sethiay, line 38: add the comment you added in the description here also about unstable

tools/cd_scripts/install_test.sh
```diff
@@ -62,39 +72,41 @@ else
...
-# Uninstall gcsfuse latest version and install old version
-if grep -q ubuntu details.txt || grep -q debian details.txt;
-then
-    sudo apt remove -y gcsfuse
-    sudo apt install -y gcsfuse=0.42.5 -t gcsfuse-$(lsb_release -cs) |& tee -a ~/logs.txt
-else
-    sudo yum -y remove gcsfuse
-    sudo yum -y install gcsfuse-0.42.5-1 |& tee -a ~/logs.txt
-fi
+if [[ $architecture == amd64 ]]; then
```
sethiay, line 75: use then consistently, either use in the same line or in different line. I prefer in different line

sethiay, line 75: why do we have this if ? (there is no else to it also)

Tulsishah: This if block will execute only for amd64 machines as we don't have any previous release for arm64.

Tulsishah: added comment for the same.

sethiay: This part will also not execute https://github.com/GoogleCloudPlatform/gcsfuse/blob/a5f5782386123da077a7d34d9615b02cf22eda66/tools/cd_scripts/install_test.sh#L98? And we should execute this part ?

Tulsishah: Once we will release one version, we will add that. At this point I don't think so it is require.

sethiay: my concern is that customers can still run upgrade command. So, we should ensure that the command works (even though the version already exists)
Also, please make note of things on internal bug what code changes we should do from next release.

Tulsishah: Yes will add.

sethiay (approved):

PR #1359, opened 2023-09-07
Fix comment for local file sync  call

ashmeenkaur:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -986,9 +986,8 @@ func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
 	// syncFile can be triggered for unlinked files if the fileHandle is open by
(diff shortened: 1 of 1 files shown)


internal/fs/fs.go
```diff
@@ -986,9 +986,8 @@ func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
 	// syncFile can be triggered for unlinked files if the fileHandle is open by
-	// another user/terminal. Hence, ignoring the syncFile call for local file.
```
vadlakondaswetha, line 989: filehandle can be open in same terminal as well. Just mention same user or another user

ashmeenkaur: done

vadlakondaswetha (approved):

PR #1361, opened 2023-09-08
 [execute-integration-tests] Disabling CreateEmptyFile flag by default - opposite of the current behaviour

vadlakondaswetha:
### Description
Disabling createEmptyFile flag by default i.e, zero byte files will not be created.
Also fixed setting HardLink count for local files as part of unit test fixes.

NA

### Testing details
1. Manual - Test manually by enabling/disabling and with the default values of the flag.
2. Unit tests - All unit tests will run with createEmptyFile=false now. Fixed the unit tests as required.
4. Integration tests - All integration will run with createEmptyFile=false now.

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -397,13 +397,6 @@ func (f *FileInode) Attributes(
 	attrs.Atime = attrs.Mtime
 	attrs.Ctime = attrs.Mtime
 
-	// Clobbered check makes a stat call to GCS. Avoiding stat calls to GCS for
-	// local files when fetching attributes. Any validations will be done during
-	// sync call.
-	if f.IsLocal() {
-		return
-	}
-
 	// If the object has been clobbered, we reflect that as the inode being
 	// unlinked.
 	_, clobbered, err := f.clobbered(ctx, false)
@@ -412,8 +405,11 @@ func (f *FileInode) Attributes(
 		return
 	}
 
-	if !clobbered {
-		attrs.Nlink = 1
+	attrs.Nlink = 1
+
+	// For local files, also checking if file is unlinked locally.
+	if clobbered || (f.IsLocal() && f.IsUnlinked()) {
+		attrs.Nlink = 0
 	}
 
 	return
--- a/internal/fs/local_modifications_test.go
+++ b/internal/fs/local_modifications_test.go
@@ -2211,15 +2211,16 @@ func (t *FileTest) ContentTypes() {
 		// Create a file.
 		f, err := os.Create(p)
 		AssertEq(nil, err)
-		f.Close()
-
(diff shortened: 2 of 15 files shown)


internal/fs/inode/file.go
```diff
@@ -397,13 +397,6 @@ func (f *FileInode) Attributes(
 	attrs.Atime = attrs.Mtime
 	attrs.Ctime = attrs.Mtime
 
-	// Clobbered check makes a stat call to GCS. Avoiding stat calls to GCS for
-	// local files when fetching attributes. Any validations will be done during
-	// sync call.
-	if f.IsLocal() {
```
raj-prince, line 403: I didn't understand, why are we removing this?

vadlakondaswetha: Earlier i thought of removing call to GCS for local files. But now it is required to check if the file is clobbered or not to ensure Nlink is set correctly.

internal/fs/local_modifications_test.go
```diff
@@ -222,6 +217,10 @@ func (t *OpenTest) NonExistent_CreateFlagSet() {
 	// Close the file.
 	AssertEq(nil, t.f1.Close())
 	t.f1 = nil
+	// The object should now be present in the bucket, with empty contents.
```
raj-prince, line 220: Please correct the comment, also add new line before.

vadlakondaswetha: done.

internal/fs/local_modifications_test.go
```diff
@@ -2040,6 +2039,9 @@ func (t *FileTest) Sync_NotDirty() {
 	// Create a file.
 	t.f1, err = os.Create(path.Join(mntDir, "foo"))
 	AssertEq(nil, err)
+	// Sync the file.
```
raj-prince, line 2043: nit: New line before.

internal/fs/inode/file.go
```diff
@@ -412,8 +405,11 @@ func (f *FileInode) Attributes(
 		return
 	}
 
-	if !clobbered {
-		attrs.Nlink = 1
+	attrs.Nlink = 1
```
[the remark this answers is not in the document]
vadlakondaswetha: As discussed offline, we dont use the attributes passed in the request. We construct that attributes, update the request and send it back.

Nlink will be only 0 or 1 in the new flow. and 0 if its clobbered.

internal/fs/local_modifications_test.go
vadlakondaswetha: sync and create are one operation here to create file in GCS. Fixed the new lines in further steps to group operations.

raj-prince (approved):

PR #1362, opened 2023-09-08
add old logging flags to config file

ashmeenkaur:
### Description
added file and format to logging configs in config file.
made changes such that it over writes the log flag if both are set.

NA

### Testing details
1. Manual - Verified that the new log configs for format and file are working.
2. Unit tests - added
3. Integration tests - NA


main.go
```diff
@@ -186,6 +186,40 @@ func populateArgs(c *cli.Context) (
...
+	// severity to TRACE.
+	if flags.DebugFuse || flags.DebugGCS || flags.DebugMutex {
+		mountConfig.LogConfig.Severity = config.TRACE
+	}
+}
+
+func resolveFilePath(filePath string, configKey string) (resolvedPath string, err error) {
+	resolvedPath, err = getResolvedPath(filePath)
+	if filePath == resolvedPath || err != nil {
+		return
```
vadlakondaswetha, line 208: looks like this is all duplicate of code in flags.go. can we common logic into some utility file.

internal/config/mount_config.go
```diff
@@ -19,7 +19,9 @@ type WriteConfig struct {
 }
 
 type LogConfig struct {
-	Severity LogSeverity `yaml:"severity"`
+	Severity  LogSeverity `yaml:"severity"`
+	LogFormat string      `yaml:"format"`
```
vadlakondaswetha, line 23: is it possible to keep the property name same as yaml field name. In that case you need to refer to logConfig as mountConfig.LogConfig.format. In that case do we need to add nil check on LogConfig everytime accessed via mountConfig?

Going forward if we need to add same properties in different configs, it will be easier to refer using their sub config names.

ashmeenkaur: In golang, the zero-value of a struct type is the struct with all its elements set to their zero values so it will never be nil. 

Made the changes to use the subconfig (full path).

main.go
ashmeenkaur: Refactored the code.

config.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+import "github.com/googlecloudplatform/gcsfuse/internal/config"
+
+func overrideWithLoggingFlags(mountConfig *config.MountConfig, flags *flagStorage) {
```
vadlakondaswetha, line 5: can we move everything into internal/config package and also merge config.go into util.go file.

util.go
```diff
@@ -0,0 +1,54 @@
+package main
```
vadlakondaswetha, line 1: please confirm if this is all copy pasted code and nothing is modified

util.go and util_test.go

ashmeenkaur: Yes this is just copy pasted.

config.go
```diff
@@ -0,0 +1,27 @@
...
+	// if log file is not set in config file, override it with flag value.
+	if mountConfig.LogConfig.File == "" {
+		mountConfig.LogConfig.File = flags.LogFile
+	}
+	// if log format is not set in config file, override it with flag value.
```
vadlakondaswetha, line 10: nit: ensure that all comments start with capital letter

internal/util/util.go
```diff
@@ -0,0 +1,64 @@
...
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/googlecloudplatform/gcsfuse/internal/config"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
+)
+
+const GCSFUSE_PARENT_PROCESS_DIR = "gcsfuse-parent-process-dir"
```
ashmeenkaur, line 14: FYI: this is moved from flags.go

internal/util/util.go
```diff
@@ -0,0 +1,64 @@
...
+	if filePath == resolvedPath || err != nil {
+		return
+	}
+
+	logger.Infof("Value of [%s] resolved from [%s] to [%s]\n", configKey, filePath, resolvedPath)
+	return resolvedPath, nil
+}
+
+// ResolveConfigFilePaths resolved the config file paths specified in the config file.
+func ResolveConfigFilePaths(config *config.MountConfig) (err error) {
```
ashmeenkaur, line 58: FYI, this is a new method added

internal/util/util_test.go
```diff
@@ -0,0 +1,145 @@
...
+	os.Setenv(GCSFUSE_PARENT_PROCESS_DIR, gcsFuseParentProcessDir)
+	defer os.Unsetenv(GCSFUSE_PARENT_PROCESS_DIR)
+
+	resolvedPath, err := GetResolvedPath("/var/dir/test.txt")
+
+	AssertEq(nil, err)
+	ExpectEq("/var/dir/test.txt", resolvedPath)
+}
+
+func (t *UtilTest) TestResolveConfigFilePaths() {
```
ashmeenkaur, line 133: Apart from this test, everything is copied from flags_test.go

internal/util/util.go
vadlakondaswetha, line 58: Ideally util shouldn't refer to any other package to avoid cyclic dependencies in future. can we put this also in config.go

internal/config/mount_config.go
```diff
@@ -20,6 +20,8 @@ type WriteConfig struct {
 
 type LogConfig struct {
 	Severity LogSeverity `yaml:"severity"`
+	Format   string      `yaml:"format"`
+	File     string      `yaml:"file"`
```
vadlakondaswetha, line 24: file-path

internal/config/config.go
```diff
@@ -0,0 +1,20 @@
+package config
+
+// OverrideWithLoggingFlags overwrites the configs with the flag values if the
+// config values are empty.
+func OverrideWithLoggingFlags(mountConfig *MountConfig, logFile string, logFormat string,
```
vadlakondaswetha, line 5: rename this file to config_util.go

internal/util/util.go
ashmeenkaur: Putting this in config.go is causing cyclic dependency as then we need to import util (ResolveFilePath method) which in turn depends on logger, and logger depends on config values.

dependency:

vadlakondaswetha (approved):

vadlakondaswetha (approved):

PR #1365, opened 2023-09-11
Revert "Write failure experience"

vadlakondaswetha:
Reverts GoogleCloudPlatform/gcsfuse#1316


raj-prince (approved):

tools/integration_tests/util/setup/setup.go
```diff
@@ -111,7 +110,7 @@ func CompareFileContents(t *testing.T, fileName string, fileContent string) {
 	}
 
 	if got := string(content); got != fileContent {
-		t.Errorf("File content doesn't match. Expected: %q, Actual: %q", fileContent, got)
```
gargnitingoogle, line 114: Can we not undo this fix and leave it as it is?

vadlakondaswetha: what is undo without reverting the PR?

PR #1333, opened 2023-09-04
Added unittests for symlink flow

vadlakondaswetha:
### Description
Added unit tests for symlink flow
Creating a symlink to local file and reading it
Creating symlink to local file, deleting it and reading the symlink

### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - Only unit tests are added
2. Unit tests - NA
3. Integration tests - NA

PR #1367, opened 2023-09-11
Reverting the unlink changes for local files as part of multi-writer fix revert

vadlakondaswetha:
Sync after unlink for local file is made similar to multi-writer fix flow. Throwing error is not letting user to close the file handle. Hence reverting it as part of multi-write fix revert.

Reverts: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1358
Manual revert is done since automatic revert is not possible.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -988,7 +988,7 @@ func (fs *fileSystem) syncFile(
 	// syncFile can be triggered for unlinked files if the fileHandle is open by same
 	// or another user. Hence, returning an error when syncFile is called for local file.
 	if f.IsLocal() && f.IsUnlinked() {
-		err = fmt.Errorf("file is unlinked")
+		// Silently ignore the syncFile call. This is in sync with non-local file behaviour.
 		return
 	}
 
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -985,10 +985,10 @@ func (fs *fileSystem) lookUpOrCreateChildDirInode(
 func (fs *fileSystem) syncFile(
 	ctx context.Context,
 	f *inode.FileInode) (err error) {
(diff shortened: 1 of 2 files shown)


internal/fs/fs.go
```diff
@@ -988,7 +988,7 @@ func (fs *fileSystem) syncFile(
 	// syncFile can be triggered for unlinked files if the fileHandle is open by same
 	// or another user. Hence, returning an error when syncFile is called for local file.
```
ashmeenkaur, line 989: Need an update here as well

ashmeenkaur, line 989: Hence, ignoring the syncFile call for local file.

vadlakondaswetha: fixed.

internal/fs/local_file_test.go
```diff
@@ -788,7 +781,7 @@ func (t *LocalFileTest) TestReadSymlinkForDeletedLocalFile() {
 	err = os.Remove(filePath)
 	AssertEq(nil, err)
 	err = t.closeLocalFile(&t.f1)
-	t.validateIOError(err)
+	AssertEq(nil, err)
```
raj-prince, line 784: Not seeing this change in this PR - https://github.com/GoogleCloudPlatform/gcsfuse/pull/1358

vadlakondaswetha: Symlink changes were added in a follow PR on top of this.

https://github.com/GoogleCloudPlatform/gcsfuse/pull/1333

ashmeenkaur (approved):

raj-prince (approved):

PR #1375, opened 2023-09-14
[execute-integration-tests]Fixing mtime for empty local files synced to GCS

vadlakondaswetha:
### Description
Any modification to the file will result in a setInodeAttributes call to GCSFuse to update mtime. But when a file is created and synced without writing any content, then mtime is not set. Fixing it by setting mtime on file when it is created.

NA

### Testing details
1. Manual - Tested manually by mounting.
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/internal/fs/inode/file_test.go
+++ b/internal/fs/inode/file_test.go
@@ -47,6 +47,7 @@ const gid = 456
 const fileInodeID = 17
 const fileName = "foo/bar"
 const fileMode os.FileMode = 0641
+const Delta = 30 * time.Minute
 
 type FileTest struct {
 	ctx    context.Context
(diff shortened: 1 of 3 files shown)


internal/fs/inode/file_test.go
```diff
@@ -829,11 +833,15 @@ func (t *FileTest) TestSetMtimeForLocalFileShouldUpdateLocalFileAttributes() {
 	var attrs fuseops.InodeAttributes
 	// Create a local file inode.
 	t.createInodeWithLocalParam("test", true)
+	createTime := t.in.mtimeClock.Now()
 	err = t.in.CreateEmptyTempFile()
+	// Validate the attributes on an empty file.
+	attrs, err = t.in.Attributes(t.ctx)
 	AssertEq(nil, err)
+	ExpectThat(attrs.Mtime, timeutil.TimeNear(createTime, 30*time.Minute))
```
raj-prince, line 841: We can create a const for 30 * time.Minute.

raj-prince (approved):

raj-prince (approved):

ashmeenkaur (approved):

PR #1376, opened 2023-09-14
Increasing timeout in e2e tests

Tulsishah:
### Description
On kokoro continuous test we used n2-standard-16 and t2a-standard-16 machines. On these machines operation package is taking around 30 minutes to run all the tests, so I increase the timeout from 24m to 40m.

As discussed @vadlakondaswetha I'll investigate about why it is taking long time and merge this for now. 

NA

### Testing details
1. Manual - Tested on local VM
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/run_e2e_tests.sh
+++ b/perfmetrics/scripts/run_e2e_tests.sh
@@ -18,7 +18,7 @@
 set -e
 sudo apt-get update
 
-readonly INTEGRATION_TEST_TIMEOUT=33m
+readonly INTEGRATION_TEST_TIMEOUT=40m
 readonly PROJECT_ID="gcs-fuse-test-ml"
 readonly BUCKET_LOCATION="us-west1"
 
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -18,7 +18,7 @@
 set -e
 sudo apt-get update
 
-readonly INTEGRATION_TEST_TIMEOUT=24m
```
raj-prince, line 21: As discussed: 
1. this is a periodic test, so it's okay to use low-computation (standard-16) power VM.
2. For presubmit test, we will be using standard-32 machines, which took lesser time.

We can increase further to 40m, so that we don't have to change the timeout frequently as we add more tests.

CC: @vadlakondaswetha

Tulsishah: Increased to 40m.

raj-prince (approved):

PR #1380, opened 2023-09-19
Copying logs on kokoro artifacts incase of integration test failure.

Tulsishah:
### Description
Copying logs on kokoro artifacts incase of integration test failure.
O_DIRECT flag was giving error of invalid argument while io.Copy.

NA

### Testing details
1. Manual - tested on pr perf test kokoro trigger.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/util/operations/file_operations.go
+++ b/tools/integration_tests/util/operations/file_operations.go
@@ -24,9 +24,11 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path"
 	"strconv"
 	"strings"
 	"syscall"
+	"testing"
 )
 
 func copyFile(srcFileName, dstFileName string, allowOverwrite bool) (err error) {
@@ -160,14 +162,14 @@ func MoveFile(srcFilePath string, destDirPath string) (err error) {
 
 func CloseFile(file *os.File) {
 	if err := file.Close(); err != nil {
-		log.Printf("error in closing: %v", err)
+		log.Fatalf("error in closing: %v", err)
 	}
 }
 
 func RemoveFile(filePath string) {
 	err := os.Remove(filePath)
 	if err != nil {
-		log.Printf("Error in removing file:%v", err)
+		log.Printf("os.Remove(%s): %v", filePath, err)
 	}
 }
 
@@ -177,7 +179,7 @@ func ReadFileSequentially(filePath string, chunkSize int64) (content []byte, err
 
 	file, err := os.OpenFile(filePath, os.O_RDONLY|syscall.O_DIRECT, FilePermission_0600)
 	if err != nil {
-		log.Printf("Error in opening file:%v", err)
+		log.Printf("Error in opening file: %v", err)
 	}
 
 	// Closing the file at the end.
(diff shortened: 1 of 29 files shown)


tools/integration_tests/util/setup/setup.go
```diff
@@ -211,6 +211,15 @@ func UnMountAndThrowErrorInFailure(flags []string, successCode int) {
 	if successCode != 0 {
 		f := strings.Join(flags, " ")
 		log.Print("Test Fails on " + f)
+
+		logData, err := os.ReadFile(LogFile())
+		if err != nil {
+			log.Fatalf("Error in printing logs: %v", err)
+			return
```
gargnitingoogle, line 218: Nit: A return after log.Fatalf is probably meaningless.

gargnitingoogle (approved):

tools/integration_tests/util/setup/setup.go
```diff
@@ -211,6 +211,14 @@ func UnMountAndThrowErrorInFailure(flags []string, successCode int) {
 	if successCode != 0 {
 		f := strings.Join(flags, " ")
 		log.Print("Test Fails on " + f)
+
+		logData, err := os.ReadFile(LogFile())
```
vadlakondaswetha, line 215: Log file will have logs for all tests. What do you mean by printing logs of failed tests?

vadlakondaswetha: Also why are we printing in console and not uploading to kokoro?

Tulsishah: > Also why are we printing in console and not uploading to kokoro?

We are printing in kokoro console right?

Tulsishah: > Log file will have logs for all tests. What do you mean by printing logs of failed tests?

Log file will have log of that mount on which test fails.

vadlakondaswetha: We are printing in kokoro console right?
It will print logs on test fusion.

>> why are we not uploading to kokoro. wont the test fusion get bloated up logs considering there are lot of tests under each mount?

Tulsishah: You mean to upload the logs on bucket ?

vadlakondaswetha: sorry, yes bucket. I meant use as kokoro artifact

Tulsishah: fixed

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -50,9 +50,9 @@ func copyFile(srcFileName, dstFileName string, allowOverwrite bool) (err error)
 
 	var destination *os.File
 	if allowOverwrite {
-		destination, err = os.OpenFile(dstFileName, os.O_WRONLY|os.O_CREATE|syscall.O_DIRECT|os.O_TRUNC, FilePermission_0600)
```
vadlakondaswetha, line 53: why is this syscall.O_DIRECT and os.O_DIRECT?

Also why was O_DIRECT required here?

tools/integration_tests/util/setup/setup.go
```diff
@@ -236,6 +249,15 @@ func UnMountAndThrowErrorInFailure(flags []string, successCode int) {
 	if successCode != 0 {
 		f := strings.Join(flags, " ")
 		log.Print("Test Fails on " + f)
+
+		// Logfile name will be gcsfuse-failed-integration-test-log-xxxxx
+		failedlogsFileName := "gcsfuse-failed-integration-test-logs-" + GenerateRandomString(5)
```
vadlakondaswetha, line 254: why do you need such a big name. I am assumig the log is different for every package. how will you which log corresponds which failure?

vadlakondaswetha (commented):
Please update the description properly

tools/integration_tests/util/setup/setup.go
Tulsishah: We are printing log file name after test failure. If test will fail one can find that file name in artifacts.
logfile is different for every mount.

tools/integration_tests/util/operations/file_operations.go
Tulsishah:
```
Currently we are using linux command to clear the cache, instead we should replace with the O_DIRECT flag while reading any file.

os.OpenFile(name, syscall.O_DIRECT)
```

@raj-prince suggested this before.

But while executing some operations it is throwing invalid argument error.
I raised question on yaqs, they suggested to remove this flag `filesystem doesn't support this where you are reading the file.`

tools/integration_tests/util/setup/setup.go
vadlakondaswetha: Got it. Also why do you have this logic of generating random string ? why can't you directly copy the log file with its original name into some destination folder and use that as regex_match for artifacts.

Tulsishah: Logfile name is gcsfuse.log only, it is in the directory /tmp/gcsfuse_readwrite_test_xxxxxxxx/gcsfuse.log. You mean to copy whole path? If yes, than we have to first create directory /tmp/gcsfuse_readwrite_test_xxxxxxxx/ in kokoro artifacts. 

I think what I did previously by creating file name with random string seems okay to me. Let me know what correction is required. I'll work accordingly.

vadlakondaswetha: Okay. 
Atleast remove gcsfuse from this: gcsfuse-failed-integration-test-logs-

name looks too big.

vadlakondaswetha (dismissed):

vadlakondaswetha (approved):

PR #1381, opened 2023-09-19
[ready for review] local file tests with implicit dir flag

ashmeenkaur:
Add tests to verify that file is not created on GCS until local files are synced to run with implicit_dir flag.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - 
Ran implicit_dir and explicit_dir tests locally on a VM with 
- test-bucket flag
- The following commands:
```
# package implicit_dir
# Run tests with static mounting. (flags: --implicit-dirs)
gcsfuse --implicit-dirs $TEST_BUCKET_NAME $MOUNT_DIR
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/implicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME
sudo umount $MOUNT_DIR

# Run test with persistent mounting. (flags: --implicit-dirs)
mount.gcsfuse $TEST_BUCKET_NAME $MOUNT_DIR -o implicit_dirs
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/implicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME
sudo umount $MOUNT_DIR

# Run tests with static mounting. (flags: --implicit-dirs, --only-dir testDir)
gcsfuse --only-dir testDir --implicit-dirs $TEST_BUCKET_NAME $MOUNT_DIR
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/implicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME/testDir
sudo umount $MOUNT_DIR

# Run test with persistent mounting. (flags: --implicit-dirs,--only-dir=testDir)
mount.gcsfuse $TEST_BUCKET_NAME $MOUNT_DIR -o only_dir=testDir,implicit_dirs
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/implicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME/testDir
sudo umount $MOUNT_DIR
```
```
# package explicit_dir
# Run tests with static mounting. (flags: --implicit-dirs=false)
gcsfuse --implicit-dirs=false $TEST_BUCKET_NAME $MOUNT_DIR
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/explicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME
sudo umount $MOUNT_DIR

# Run test with persistent mounting. (flags: --implicit-dirs=false)
mount.gcsfuse $TEST_BUCKET_NAME $MOUNT_DIR -o implicit_dirs=false
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/explicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME
sudo umount $MOUNT_DIR

# Run tests with static mounting. (flags: --implicit-dirs=false, --only-dir testDir)
gcsfuse --only-dir testDir  --implicit-dirs=false $TEST_BUCKET_NAME $MOUNT_DIR
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/explicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME/testDir
sudo umount $MOUNT_DIR

# Run test with persistent mounting. (flags: --implicit-dirs=false, --only-dir=testDir)
mount.gcsfuse $TEST_BUCKET_NAME $MOUNT_DIR -o only_dir=testDir,implicit_dirs=false
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/explicit_dir/...  -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME/testDir
sudo umount $MOUNT_DIR

```

### DIFF
--- a/tools/integration_tests/implicit_dir/implicit_dir_test.go
+++ b/tools/integration_tests/implicit_dir/implicit_dir_test.go
@@ -16,8 +16,15 @@
 package implicit_dir_test
 
 import (
+	"context"
+	"log"
+	"os"
+	"path"
 	"testing"
+	"time"
 
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/client"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
 	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup/implicit_and_explicit_dir_setup"
 )
 
@@ -29,7 +36,27 @@ const NumberOfFilesInExplicitDirInImplicitSubDir = 1
 const NumberOfFilesInExplicitDirInImplicitDir = 1
 
 func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+
+	ctx = context.Background()
+	var cancel context.CancelFunc
+	var err error
+
+	// Create storage client before running tests.
+	ctx, cancel = context.WithTimeout(ctx, time.Minute*15)
+	storageClient, err = client.CreateStorageClient(ctx)
+	if err != nil {
+		log.Fatalf("client.CreateStorageClient: %v", err)
+	}
+
 	flags := [][]string{{"--implicit-dirs"}}
 
-	implicit_and_explicit_dir_setup.RunTestsForImplicitDirAndExplicitDir(flags, m)
+	successCode := implicit_and_explicit_dir_setup.RunTestsForImplicitDirAndExplicitDir(flags, m)
+
+	// Close storage client and release resources.
+	storageClient.Close()
+	cancel()
+	// Clean up test directory created.
+	setup.CleanupDirectoryOnGCS(path.Join(setup.TestBucket(), testDirName))
+	os.Exit(successCode)
 }
--- a/tools/integration_tests/implicit_dir/local_file_test.go
+++ b/tools/integration_tests/implicit_dir/local_file_test.go
@@ -15,30 +15,19 @@ package implicit_dir_test
 
 import (
 	"io/fs"
-	"os"
 	"path"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"cloud.google.com/go/storage"
-	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/client"
+	. "github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/client"
 	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/operations"
 	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
 	"golang.org/x/net/context"
 )
 
 const (
-	FileName1         = "foo1"
-	FileName2         = "foo2"
-	ImplicitDirName   = "implicit"
-	ImplicitFileName1 = "implicitFile1"
-	ExplicitDirName   = "explicit"
-	ExplicitFileName1 = "explicitFile1"
-	FilePerms         = 0644
-	FileContents      = "teststring"
-	ReadSize          = 1024
-	testDirName       = "ImplicitDirTest"
+	testDirName = "ImplicitDirTest"
 )
 
 var (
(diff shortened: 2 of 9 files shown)


tools/integration_tests/implicit_dir/implicit_dir_test.go
```diff
@@ -29,7 +36,27 @@ const NumberOfFilesInExplicitDirInImplicitSubDir = 1
...
+	}
+
 	flags := [][]string{{"--implicit-dirs"}}
 
-	implicit_and_explicit_dir_setup.RunTestsForImplicitDirAndExplicitDir(flags, m)
+	successCode := implicit_and_explicit_dir_setup.RunTestsForImplicitDirAndExplicitDir(flags, m)
+
+	// Close storage client and release resources.
+	storageClient.Close()
+	cancel()
```
vadlakondaswetha, line 58: what is cancel doing if storageClient is already closed?

tools/integration_tests/util/setup/setup.go
```diff
@@ -262,17 +287,43 @@ func LogAndExit(s string) {
...
+
+// SetupTestDirectory creates a testDirectory in the mounted directory and cleans up
+// any content present in it.
+func SetupTestDirectory(testDirName string) string {
+	testDirPath := path.Join(MntDir(), testDirName)
+	err := os.Mkdir(testDirPath, DirPermission_0755)
+	if err != nil && !strings.Contains(err.Error(), "file exists") {
+		log.Printf("Error while setting up directory %s for testing: %v", testDirPath, err)
+	}
+	CleanUpDir(testDirPath)
```
vadlakondaswetha, line 318: Is this step really necessary. won't os.Mkdir() throw error if directory already exists?

tools/integration_tests/implicit_dir/local_file_test.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func validateObjectNotFoundErrOnGCS(fileName string, t *testing.T) {
+	_, err := client.ReadObjectFromGCS(storageClient, path.Join(testDirName, fileName), ReadSize, ctx)
+	if err == nil || !strings.Contains(err.Error(), "storage: object doesn't exist") {
+		t.Fatalf("Incorrect error returned from GCS for file %s: %v", fileName, err)
+	}
+}
+
+func validateObjectContentsFromGCS(fileName string, expectedContent string, t *testing.T) {
```
vadlakondaswetha, line 72: Are these functions not required by other tests?

tools/integration_tests/implicit_dir/implicit_dir_test.go
ashmeenkaur: In case the timeout exceeds, the cancel will be called by itself. Here we are explicitly calling cancel to release any resources (before the timeout) since we are done with the operations.

tools/integration_tests/util/setup/setup.go
ashmeenkaur: Background: @Tulsishah suggested that in future, in order to parallelise the tests, we need to run them in separate directories.

This step is necessary because we are not sure of the order in which the tests are called. I tried creating the test directory before executing the tests as well but we have many different kinds of mount - for eg: dynamic, only-dir, etc where we are not sure of the location where we should create this directory. Even if we do, we still need to update the path, so this change seemed reasonable.

os.Mkdir will throw "file exists" error if directory already exists which we are ignoring here.

tools/integration_tests/implicit_dir/local_file_test.go
ashmeenkaur: These tests check specifically in the testDir for implicit_dir. I have created similar functions in local_file_tests.
Some code repetition is present here, but it is saving us from passing additional parameters like storage client, context and test_dir. Let me know if you think they should be in a util file instead.

tools/integration_tests/implicit_dir/implicit_dir_test.go
vadlakondaswetha: Understood the timeout flow.
Can you give an example where resources needs to be cleaned up using cancel eventhough storage.client is closed?

tools/integration_tests/implicit_dir/local_file_test.go
vadlakondaswetha: Yeah i would prefer them to be in single place to avoid repeated bugs.

But if this is cleaner lets keep like this in this PR

tools/integration_tests/util/setup/setup.go
vadlakondaswetha: got it.

tools/integration_tests/implicit_dir/implicit_dir_test.go
ashmeenkaur: client.Close just sets the fields to nil so that subsequent uses of client will panic. It does not cancels the context. (Code copied below)

```
func (c *Client) Close() error {
	// Set fields to nil so that subsequent uses will panic.
	c.hc = nil
	c.raw = nil
	c.creds = nil
	if c.tc != nil {
		return c.tc.Close()
	}
	return nil
}
```

cancelling the context will stop execution of any in progress operations, so the two are slightly different I believe.

tools/integration_tests/implicit_dir/local_file_test.go
ashmeenkaur: Very valid concern. I do see some scope of refactoring in file_operations.go and dir_operations.go. I'll create a backlog bug for it and take it up when I am on-call next time.

tools/integration_tests/implicit_dir/implicit_dir_test.go
vadlakondaswetha: Thanks.

vadlakondaswetha (approved):

PR #1382, opened 2023-09-19
Integration tests to verify that file creation does not creates file on GCS until synced.

ashmeenkaur:
### Description
Added integration tests to verify that file creation does not creates file on GCS until synced.

NA

### Testing details
1. Manual - 
- Manually ran on GCP VM using the following commands:
```
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/local_file/...  -p 1 --integrationTest -v --testbucket=$TEST_BUCKET_NAME
``` 
```
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/local_file/...  -p 1 --integrationTest -v --testbucket=$TEST_BUCKET_NAME --testInstalledPackage
``` 
- Ran the following commands for mounted directory tests
```
# package local_file
# Run test with static mounting. (flags: --implicit-dirs=true)
gcsfuse --implicit-dirs=true --rename-dir-limit=3 $TEST_BUCKET_NAME $MOUNT_DIR
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/local_file/... -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME
sudo umount $MOUNT_DIR

# Run test with static mounting. (flags: --implicit-dirs=false)
gcsfuse --implicit-dirs=false --rename-dir-limit=3 $TEST_BUCKET_NAME $MOUNT_DIR
GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/local_file/... -p 1 --integrationTest -v --mountedDirectory=$MOUNT_DIR --testbucket=$TEST_BUCKET_NAME
sudo umount $MOUNT_DIR
```
2. Unit tests - NA
3. Integration tests - NA


tools/integration_tests/local_file/create_file_test.go
```diff
@@ -0,0 +1,55 @@
...
+	// Create a local file.
+	_, fh := CreateLocalFileInTestDir(ctx, storageClient, testDirPath, FileName1, t)
+
+	// Create a file on GCS with the same name.
+	CreateObjectInGCSTestDir(ctx, storageClient, testDirName, FileName1, GCSFileContent, t)
+
+	// Write to local file.
+	operations.WriteWithoutClose(fh, FileContents, t)
+	// Close the local file and ensure that the content on GCS is not overwritten.
+	CloseFileAndValidateObjectContentsFromGCS(ctx, storageClient, fh, testDirName, FileName1, GCSFileContent, t)
```
vadlakondaswetha, line 54: Can you split this method just to keep the flow clear.
Closing local file should not update content on GCS. Ideally it should throw an error which will be fixed with multi-writer failure fix.

2nd. The existing object on GCS is not modified.

tools/integration_tests/local_file/local_file_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+
+	ctx = context.Background()
+	var cancel context.CancelFunc
+	var err error
+
+	setup.ExitWithFailureIfBothTestBucketAndMountedDirectoryFlagsAreNotSet()
+
+	if setup.MountedDirectory() != "" && setup.TestBucket() == "" {
```
vadlakondaswetha, line 55: The error says both of them has to be present.
what about other condition when mountedDirectory="" and testBucket != "".

I would suggest adding a helper method in setup to validate if both of them are present.

tools/integration_tests/local_file/local_file_test.go
```diff
@@ -0,0 +1,105 @@
...
+		log.Print("Both --testbucket and --mountedDirectory should be specified to run mounted directory tests.")
+		os.Exit(1)
```
vadlakondaswetha, line 57: why not log.fatalf here?

tools/integration_tests/local_file/local_file_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// Create storage client before running tests.
+	ctx, cancel = context.WithTimeout(ctx, time.Minute*15)
+	storageClient, err = client.CreateStorageClient(ctx)
+	if err != nil {
+		log.Fatalf("client.CreateStorageClient: %v", err)
+	}
+
+	// Run tests for mountedDirectory only if --mountedDirectory flag is set.
+	setup.RunTestsForMountedDirectoryFlag(m)
```
vadlakondaswetha, line 68: its not clear if you need both mountedDirectory and testbucket flags to be set.

vadlakondaswetha, line 68: no need to set the config here since the default value is false. If yes why are we setting it below explicitly. Creates confusion?

tools/integration_tests/local_file/local_file_test.go
ashmeenkaur: Refactored as suggested.

Just to clarify:
With testBucket flag, we run tests line 70 onwards.
With mountedDirectory flag, we run the tests on mounted directory. Usually in other tests, we don't need the testBucket to be specified in this case but since. local file tests are validating from GCS, we need test bucket to be specified. Hence this check.

tools/integration_tests/util/setup/setup.go
```diff
@@ -327,3 +327,13 @@ func CleanupDirectoryOnGCS(directoryPathOnGCS string) {
 			directoryPathOnGCS, err)
 	}
 }
+
+func RunMountedDirectoryTestsWithTestBucket() bool {
```
vadlakondaswetha, line 331: can you rename this method AreBothMountedDirectoryAndTestBucketFlags are set. Lets keep the methods generic

tools/integration_tests/local_file/local_file_test.go
```diff
@@ -0,0 +1,93 @@
...
+	// flags to be set, as local_file tests validates content from the bucket.
+	if setup.RunMountedDirectoryTestsWithTestBucket() {
+		setup.RunTestsForMountedDirectoryFlag(m)
+	}
+
+	// Else run tests for testBucket.
+	// Set up test directory.
+	setup.SetUpTestDirForTestBucketFlag()
+
+	// Set up flags to run tests on.
```
vadlakondaswetha, line 71: just add a comment that createEmptyFile is false by default. Hence not adding it as seperate config.

vadlakondaswetha (approved):

PR #1392, opened 2023-09-22
Write test failure on some machines because of gsutil download

Tulsishah:
### Description
While running e2e tests on louhi pipeline I found that some e2e tests are failing while coping file using gsutil.

These thing is happing only on some machines and not all machines.

The error is 
==> NOTE: You are downloading one or more large file(s), which would            
run significantly faster if you enabled sliced object downloads. This
feature is enabled by default but requires that compiled crcmod be
installed (see "gsutil help crcmod").

CommandException: 
Downloading this composite object requires integrity checking with CRC32c,
but your crcmod installation isn't using the module's C extension, so the
hash computation will likely throttle download performance. For help
installing the extension, please see "gsutil help crcmod".

To download regardless of crcmod performance or to skip slow integrity
checks, see the "check_hashes" option in your boto config file.

NOTE: It is strongly recommended that you not disable integrity checks. Doing so
could allow data corruption to go undetected during uploading/downloading.

This requires to install crcmod python library.

NA

### Testing details
1. Manual - Tested on louhi pipeline
2. Unit tests - NA
3. Integration tests - NA


tools/cd_scripts/e2e_test.sh
```diff
@@ -71,9 +71,8 @@ then
     #install git
     sudo apt install -y git
 
-    # Downloading one or more large file(s), which would run significantly faster if
-    # you enabled sliced object downloads. This feature is enabled by default but
-    # requires that compiled crcmod be installed.
+    # Downloading composite object requires integrity checking with CRC32c in gsutil.
```
raj-prince, line 74: All these utilities do already present on other machines?

Tulsishah: On some machines
```
gsutil version -l
compiled crcmod: False
```

and on some machines it's by default true

tools/cd_scripts/e2e_test.sh
```diff
@@ -100,6 +105,13 @@ else
     #install git
     sudo yum -y install git
 
+    # install python3-setuptools tools and python3-pip
+    sudo yum -y install gcc python3-devel python3-setuptools redhat-rpm-config
+    sudo yum -y install python3-pip
+    # Downloading composite object requires integrity checking with CRC32c in gsutil.
+    # it requires to install crcmod.
+    sudo pip3 install --no-cache-dir -U crcmod
```
raj-prince, line 113: Can't we install crcmod without pip3?

Otherwise you need to follow --required-hashes path to install it.
Ref: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1096

raj-prince (changes requested):
Please check.

tools/cd_scripts/e2e_test.sh
Tulsishah: Fixed

Tulsishah (commented):
Please check @raj-prince

tools/cd_scripts/e2e_test.sh
```diff
@@ -71,6 +71,11 @@ then
     #install git
     sudo apt install -y git
 
+    # Downloading composite object requires integrity checking with CRC32c in gsutil.
+    # it requires to install crcmod.
+    sudo apt-get install -y gcc python3-dev python3-setuptools
+    sudo apt install -y python3-crcmod
```
raj-prince, line 77:
```
CentOS, RHEL, and Fedora
  ------------------------

  To compile and install crcmod:

    yum install gcc python3-devel python3-setuptools redhat-rpm-config
    sudo pip3 uninstall crcmod
    sudo pip3 install --no-cache-dir -U crcmod

  Debian and Ubuntu
  -----------------

  To compile and install crcmod:

    sudo apt-get install gcc python3-dev python3-setuptools
    sudo pip3 uninstall crcmod
    sudo pip3 install --no-cache-dir -U crcmod
```

Why are we not following the installation method provided by `gsutil help crcmod` command for Debian and Ubuntu?

Tulsishah: It is getting installed with python3 library, so why should we do extra effort for requirements and all.

Tulsishah: this is alternative of pip3 only for debian and ubuntu.

perfmetrics/scripts/run_e2e_tests/run_e2e_tests.sh
```diff
@@ -31,6 +31,16 @@ echo "Installing go-lang 1.21.0..."
 wget -O go_tar.tar.gz https://go.dev/dl/go1.21.0.linux-${architecture}.tar.gz -q
 sudo rm -rf /usr/local/go && tar -xzf go_tar.tar.gz && sudo mv go /usr/local
 export PATH=$PATH:/usr/local/go/bin
+# Downloading composite object requires integrity checking with CRC32c in gsutil.
+# It requires to install crcmod.
+sudo apt-get install -y gcc python3-dev python3-setuptools
```
raj-prince, line 36: Duplicate installation.

perfmetrics/scripts/run_e2e_tests/run_e2e_tests.sh
```diff
@@ -31,6 +31,16 @@ echo "Installing go-lang 1.21.0..."
...
+sudo apt install -y python3-crcmod
```
raj-prince, line 37: Now, this is not required.

raj-prince (approved):

PR #1396, opened 2023-09-26
Add configs for minimal file buffering

ashmeenkaur:
### Description
added `enable-write-streaming` and `buffer-size` configs for minimal file buffering changes.

NA

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA


internal/config/mount_config.go
```diff
@@ -31,6 +33,8 @@ type MountConfig struct {
 
 func NewMountConfig() *MountConfig {
 	mountConfig := &MountConfig{}
+	mountConfig.WriteConfig.EnableStreamingWrites = false
```
vadlakondaswetha, line 36: isn't the default value false?
Also if enableStreamigWRites is not enabled by default, why are we setting it to 16. 

Lets wait for Marco to comeup with these values and then set it.

ashmeenkaur: * default value is false. I will remove line 36.

* We won't use bufferSize value until EnableStreamingWrites is enabled. 16 MB is the default chunk size of go client so I thought it should be okay to keep similar value here for start. I'll verify with Marco and update.

vadlakondaswetha: If you are waiting for Marco to verify these values, dont block this PR. 

Remove the default value and proceed.

ashmeenkaur: Done

vadlakondaswetha (approved):

internal/config/yaml_parser.go
```diff
@@ -42,9 +42,16 @@ func IsValidLogSeverity(severity LogSeverity) bool {
...
 		ERROR,
 		OFF:
-		return true
+		return nil
 	}
-	return false
+	return fmt.Errorf("error parsing config file: log severity should be one of [trace, debug, info, warning, error, off]")
+}
+
+func validateWriteConfig(writeConfig WriteConfig) error {
```
gargnitingoogle, line 50: Nit: The name is a bit redundant as it's taking in a WriteConfig, so it's obvious that it'll validate a WriteConfig. This will lead to code like 

```go
WriteConfig writeConfig = ...
validateWriteConfig(writeConfig)
```

 We could switch this to `func (writeConfig WriteConfig)Validate() error {` , which will reduce the above to

```go
writeConfig.validate()
```

Same goes for validateLogSeverity .

internal/config/yaml_parser.go
```diff
@@ -42,9 +42,16 @@ func IsValidLogSeverity(severity LogSeverity) bool {
...
+	if writeConfig.EnableStreamingWrites && writeConfig.BufferSizeMB <= 0 {
+		return fmt.Errorf("buffer size should be greater than 0")
```
gargnitingoogle, line 52: `"buffer-size-mb should be greater than 0 if enable-streaming-writes is set"` .

internal/config/yaml_parser.go
```diff
@@ -67,8 +74,11 @@ func ParseConfigFile(fileName string) (mountConfig *MountConfig, err error) {
 	}
 	// convert log severity to upper-case
 	mountConfig.LogConfig.Severity = LogSeverity(strings.ToUpper(string(mountConfig.LogConfig.Severity)))
-	if !IsValidLogSeverity(mountConfig.LogConfig.Severity) {
-		err = fmt.Errorf("error parsing config file: log severity should be one of [trace, debug, info, warning, error, off]")
+	if err = validateLogSeverity(mountConfig.LogConfig.Severity); err != nil {
+		return
```
gargnitingoogle, line 78: Don't we want to add additional error string about the specific error here, e.g.

```go
return fmt.Errorf("invalid log severity in config-file: %w", err)
```

The same goes for return from the validateWriteConfig error.

internal/config/yaml_parser.go
ashmeenkaur: Raised separate PR for all the suggested changes https://github.com/GoogleCloudPlatform/gcsfuse/pull/1412. PTAL

Thanks!

PR #1402, opened 2023-09-28
Adding --disable-auth flag changes

Tulsishah:
### Description
Adding new flag --disable-auth.
1. when we will pass --disable-auth flag it will not authenticate.(creates dummy token)
2. Customer who are using custom-endpoint for testing - they need to pass --disable-auth flag additionally.
3. Customer who wants to use custom-endpoint with authentication - they can authenticate now, as default value of --disable-auth is false.

NA

### Testing details
1. Manual - Tested using storage test bench server.
                   1. with --disable-auth flag, with credentials  and with custom endpoint - (gcsfuse mounts successfully)
                   2. without --disable-auth flag, without credentials and with custom endpoint - (throwing auth error as expected)
                   3. without --disable-auth flag, with credentials and with custom endpoint - (gcsfuse mounts successfully)
                   4. with --disable-auth flag, without credentials and with custom endpoint - (gcsfuse mounts successfully)
                   5. with --disable-auth flag, with credentials, without custom endpoint - (gcsfuse mount stuck as we don't have necessary permisisons on the bucket, it stucks on first list call itself.)
                   6. Able to mount public readonly bucket(able to do list and cat)(In create, write it stucks due to retry issue)
7. Unit tests - Added unit tests
8. Integration tests - integration tests are working with kokoro test label

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -187,6 +187,12 @@ func newApp() (app *cli.App) {
 				Usage: "File chunk size to read from GCS in one call. Need to specify the value in MB. ChunkSize less than 1MB is not supported",
 			},
 
+			cli.BoolFlag{
+				Name: "disable-auth",
+				Usage: "Authorization will be enabled by default. With this flag, authorization will be disabled. If one " +
+					"wants to use a custom endpoint for testing purposes; they need to pass --disable-auth flag.",
+			},
+
 			/////////////////////////
 			// Tuning
 			/////////////////////////
@@ -378,6 +384,7 @@ type flagStorage struct {
 	EgressBandwidthLimitBytesPerSecond float64
 	OpRateLimitHz                      float64
 	SequentialReadSizeMb               int32
+	DisableAuth                        bool
 
 	// Tuning
 	MaxRetrySleep              time.Duration
@@ -488,6 +495,7 @@ func populateFlags(c *cli.Context) (flags *flagStorage, err error) {
 		EgressBandwidthLimitBytesPerSecond: c.Float64("limit-bytes-per-sec"),
 		OpRateLimitHz:                      c.Float64("limit-ops-per-sec"),
 		SequentialReadSizeMb:               int32(c.Int("sequential-read-size-mb")),
+		DisableAuth:                        c.Bool("disable-auth"),
 
 		// Tuning,
 		MaxRetrySleep:              c.Duration("max-retry-sleep"),
--- a/flags.go
+++ b/flags.go
@@ -189,8 +189,8 @@ func newApp() (app *cli.App) {
 
 			cli.BoolFlag{
 				Name: "disable-auth",
-				Usage: "Authorization will be enabled by default. With this flag, authorization will be disabled. If one " +
-					"wants to use a custom endpoint for testing purposes; they need to pass --disable-auth flag.",
+				Usage: "Authentication is enabled by default. The --disable-auth flag disables authentication. For users of the --custom-endpoint flag," +
+					" please pass --disable-auth flag explicitly if you do not want authentication enabled for your workflow.",
 			},
 
 			/////////////////////////
--- a/internal/storage/storage_handle_test.go
+++ b/internal/storage/storage_handle_test.go
@@ -49,8 +49,14 @@ func (t *StorageHandleTest) TearDown() {
 	t.fakeStorage.ShutDown()
 }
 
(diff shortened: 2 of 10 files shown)


flags.go
```diff
@@ -187,6 +187,12 @@ func newApp() (app *cli.App) {
 				Usage: "File chunk size to read from GCS in one call. Need to specify the value in MB. ChunkSize less than 1MB is not supported",
 			},
 
+			cli.BoolFlag{
+				Name: "disable-auth",
```
raj-prince, line 191: =>
Authentication is enabled by default. The --disable-auth flag disables authentication. For users of the --custom-endpoint flag, please pass --disable-auth flag explicitly if you do not want authentication enabled for your workflow.

internal/storage/storageutil/test_util.go
```diff
@@ -41,5 +41,6 @@ func GetDefaultStorageClientConfig() (clientConfig StorageClientConfig) {
 		TokenUrl:                   "",
 		ReuseTokenFromUrl:          true,
 		ExperimentalEnableJsonRead: false,
+		DisableAuth:                true,
```
raj-prince, line 44: Why default value is true? It should be true. Right?

Tulsishah: Updated

internal/storage/storage_handle_test.go
```diff
@@ -121,10 +122,33 @@ func (t *StorageHandleTest) TestNewStorageHandleWithCustomEndpoint() {
 	t.invokeAndVerifyStorageHandle(sc)
 }
 
-// This will fail while fetching the token-source, since key-file doesn't exist.
 func (t *StorageHandleTest) TestNewStorageHandleWhenCustomEndpointIsNil() {
 	sc := storageutil.GetDefaultStorageClientConfig()
 	sc.CustomEndpoint = nil
+	sc.DisableAuth = true
```
sethiay, line 128: Because the name TestNewStorageHandleWhenCustomEndpointIsNil doesn't specify if we are changing anything other than CustomEndpoint, so we should use default value of other parameters. So DisableAuth shouldn't be set here.

TestNewStorageHandleWhenCustomEndpointIsNilAndDisableAuthIsFalse should be changed to TestNewStorageHandleWhenCustomEndpointIsNilAndDisableAuthIsTrue

internal/storage/storage_handle_test.go
```diff
@@ -121,10 +122,33 @@ func (t *StorageHandleTest) TestNewStorageHandleWithCustomEndpoint() {
...
+
+	AssertEq(nil, err)
+	AssertNe(nil, handleCreated)
+}
+
+func (t *StorageHandleTest) TestNewStorageHandleWhenCustomEndpointIsNilAndDisableAuthIsFalse() {
+	url, err := url.Parse(storageutil.CustomEndpoint)
+	AssertEq(nil, err)
+	sc := storageutil.GetDefaultStorageClientConfig()
+	sc.CustomEndpoint = url
```
sethiay, line 140: The name says custom-endpoint is nil but it isn't. Please correct that.

internal/storage/storage_handle_test.go
```diff
@@ -121,10 +122,33 @@ func (t *StorageHandleTest) TestNewStorageHandleWithCustomEndpoint() {
...
+
+	handleCreated, err := NewStorageHandle(context.Background(), sc)
+
+	AssertNe(nil, err)
+	ExpectThat(err, oglematchers.Error(oglematchers.HasSubstr("no such file or directory")))
+	AssertEq(nil, handleCreated)
+}
+
+func (t *StorageHandleTest) TestNewStorageHandleWhenCustomEndpointIsNotNilAndDisableAuthIsFalse() {
```
sethiay, line 149: Should we also test TestNewStorageHandleWhenCustomEndpointIsNotNilAndDisableAuthIsTrue ?

internal/storage/storageutil/client_test.go
```diff
@@ -55,7 +81,7 @@ func (t *clientTest) TestCreateTokenSrcWhenCustomEndpointIsNil() {
 
 func (t *clientTest) TestCreateHttpClientWithHttp1() {
 	sc := GetDefaultStorageClientConfig() // By default http1 enabled
-
+	sc.DisableAuth = true
```
sethiay, line 84: TestCreateHttpClientWithHttp1 doesn't say about DisableAuth. So, we shouldn't set DisableAuth inside.

Should we write tests for all possible combinations ((http1 and http2) * (disable auth and enable auth)) ?

PR #1405, opened 2023-09-29
Create in memory buffer when buffersize <= 50MB

ashmeenkaur:
### Description
Create an in-memory buffer of 2*configured buffer size during write when configured buffersize <= 50 mb

NA

### Testing details
1. Manual - Manually verified that buffer is creating only once.
2. Unit tests - Added unit tests for file.go and memorybuffer. unit tests for fs.go will be added in follow up PR (along with write changes.)
3. Integration tests - NA

### DIFF
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -12,12 +12,14 @@ type InMemoryWriteBuffer struct {
 	chunkSize int
 }
 
-func (b *InMemoryWriteBuffer) Create(sizeInMB int) WriteBuffer {
-	b.chunkSize = sizeInMB * 1024 * 1024
+// NewInMemoryWriteBuffer creates a buffer of 2*sizeInMB passed as a parameter.
+func NewInMemoryWriteBuffer(sizeInMB int) *InMemoryWriteBuffer {
+	b := &InMemoryWriteBuffer{}
+	b.chunkSize = sizeInMB * MiB
 	b.buffer = bytes.NewBuffer(make([]byte, 0, 2*b.chunkSize))
 	return b
 }
 
-func (b *InMemoryWriteBuffer) Write(data []byte, offset int64) error {
+func (b *InMemoryWriteBuffer) WriteAt(data []byte, offset int64) error {
 	return nil
 }
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -19,19 +19,27 @@ import (
 )
 
 type InMemoryWriteBuffer struct {
-	// buffer will hold the data to be written to GCS. Buffer holds the data for 2
-	// GCS write calls.
+	// Holds the data to be written to GCS. At any time, Buffer holds the data for
+	// 2 GCS write calls.
 	buffer *bytes.Buffer
-	// chunkSize is the size of data to be written in one GCS call.
-	chunkSize int
 }
 
-// NewInMemoryWriteBuffer creates a buffer of 2*sizeInMB passed as a parameter.
-func NewInMemoryWriteBuffer(sizeInMB int) *InMemoryWriteBuffer {
+// EmptyInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
+// set to nil. Memory is allocated to the buffer when the first write call comes.
+// This avoids unnecessarily bloating GCSFuse memory consumption.
+//
+// To allocate memory to the buffer, use WriteBuffer.AllocateBuffer.
+func EmptyInMemoryWriteBuffer() *InMemoryWriteBuffer {
 	b := &InMemoryWriteBuffer{}
-	b.chunkSize = sizeInMB * MiB
-	b.buffer = bytes.NewBuffer(make([]byte, 0, 2*b.chunkSize))
+	// TODO: set mtime attribute.
 	return b
+}
+
+func (b *InMemoryWriteBuffer) AllocateBuffer(sizeInMB int) {
+	ChunkSize = sizeInMB * MiB
+	if b.buffer == nil {
+		b.buffer = bytes.NewBuffer(make([]byte, 0, 2*sizeInMB*MiB))
+	}
 }
 
 func (b *InMemoryWriteBuffer) WriteAt(data []byte, offset int64) error {
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -24,18 +24,18 @@ type InMemoryWriteBuffer struct {
 	buffer *bytes.Buffer
 }
 
-// EmptyInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
+// CreateInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
 // set to nil. Memory is allocated to the buffer when the first write call comes.
 // This avoids unnecessarily bloating GCSFuse memory consumption.
 //
-// To allocate memory to the buffer, use WriteBuffer.AllocateBuffer.
-func EmptyInMemoryWriteBuffer() *InMemoryWriteBuffer {
+// To allocate memory to the buffer, use WriteBuffer.InitializeBuffer.
+func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
 	b := &InMemoryWriteBuffer{}
 	// TODO: set mtime attribute.
 	return b
 }
 
-func (b *InMemoryWriteBuffer) AllocateBuffer(sizeInMB int) {
+func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
 	ChunkSize = sizeInMB * MiB
 	if b.buffer == nil {
 		b.buffer = bytes.NewBuffer(make([]byte, 0, 2*sizeInMB*MiB))
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 )
 
-type MemoryBuffer struct {
+type InMemoryWriteBuffer struct {
 	// buffer will hold the data to be written to GCS. Buffer holds the data for 2
 	// GCS write calls.
 	buffer *bytes.Buffer
@@ -12,25 +12,12 @@ type MemoryBuffer struct {
 	chunkSize int
 }
 
-func (b *MemoryBuffer) Create(sizeInMB int) WriteBuffer {
+func (b *InMemoryWriteBuffer) Create(sizeInMB int) WriteBuffer {
 	b.chunkSize = sizeInMB * 1024 * 1024
 	b.buffer = bytes.NewBuffer(make([]byte, 0, 2*b.chunkSize))
 	return b
 }
 
-func (b *MemoryBuffer) Write(data []byte, offset int64) error {
+func (b *InMemoryWriteBuffer) Write(data []byte, offset int64) error {
 	return nil
-}
-
-func (b *MemoryBuffer) Upload() {
-}
-
-func (b *MemoryBuffer) Status() {
-}
-
-func (b *MemoryBuffer) Destroy() {
-}
-
-func (b *MemoryBuffer) Read(p []byte) (n int, err error) {
-	return
 }
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -1,3 +1,17 @@
+// Copyright 2023 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package buffer
 
 import (
--- a/internal/fs/buffer/inmemory_write_buffer_test.go
+++ b/internal/fs/buffer/inmemory_write_buffer_test.go
@@ -27,7 +27,7 @@ func TestMemoryBuffer(t *testing.T) { RunTests(t) }
 ////////////////////////////////////////////////////////////////////////
 
 type MemoryBufferTest struct {
-	mb *MemoryBuffer
+	mb *InMemoryWriteBuffer
 }
 
 var _ SetUpInterface = &MemoryBufferTest{}
@@ -36,7 +36,7 @@ var _ TearDownInterface = &MemoryBufferTest{}
 func init() { RegisterTestSuite(&MemoryBufferTest{}) }
 
 func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
-	t.mb = &MemoryBuffer{}
+	t.mb = &InMemoryWriteBuffer{}
 }
 
 func (t *MemoryBufferTest) TearDown() {}
--- a/internal/fs/buffer/inmemory_write_buffer_test.go
+++ b/internal/fs/buffer/inmemory_write_buffer_test.go
@@ -46,17 +46,17 @@ func (t *MemoryBufferTest) TearDown() {}
 // //////////////////////////////////////////////////////////////////////
 
 func (t *MemoryBufferTest) TestCreateEmptyInMemoryBuffer() {
-	t.mb = EmptyInMemoryWriteBuffer()
+	t.mb = CreateInMemoryWriteBuffer()
 
 	AssertEq(nil, t.mb.buffer)
 	AssertEq(0, ChunkSize)
 }
 
-func (t *MemoryBufferTest) TestAllocateInMemoryBuffer() {
+func (t *MemoryBufferTest) TestInitializeInMemoryBuffer() {
 	bufferSizeMB := 10
-	t.mb = EmptyInMemoryWriteBuffer()
+	t.mb = CreateInMemoryWriteBuffer()
 
-	t.mb.AllocateBuffer(bufferSizeMB)
+	t.mb.InitializeBuffer(bufferSizeMB)
 
 	AssertEq(bufferSizeMB*MiB, ChunkSize)
 	AssertEq(2*bufferSizeMB*MiB, t.mb.buffer.Cap())
--- a/internal/fs/buffer/inmemory_write_buffer_test.go
+++ b/internal/fs/buffer/inmemory_write_buffer_test.go
@@ -46,11 +46,11 @@ func (t *MemoryBufferTest) TearDown() {}
 // //////////////////////////////////////////////////////////////////////
 
 func (t *MemoryBufferTest) TestCreateInMemoryBuffer() {
-	bufferSize := 10
+	bufferSizeMB := 10
 
-	t.mb.Create(bufferSize)
+	t.mb = NewInMemoryWriteBuffer(bufferSizeMB)
 
-	AssertEq(bufferSize*1024*1024, t.mb.chunkSize)
-	AssertEq(2*bufferSize*1024*1024, t.mb.buffer.Cap())
+	AssertEq(bufferSizeMB*MiB, t.mb.chunkSize)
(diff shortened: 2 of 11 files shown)


internal/fs/fs.go
```diff
@@ -2077,13 +2077,31 @@ func (fs *fileSystem) WriteFile(
 	defer in.Unlock()
 
 	// Serve the request.
-	if err := in.Write(ctx, op.Data, op.Offset); err != nil {
-		return err
+	if fs.shouldTriggerBufferEnabledWrite(in) {
+		// Trigger write file buffering flow.
+		err := in.WriteWithBuffer(fs.mountConfig.WriteConfig.BufferSizeMB, op.Data, op.Offset)
```
vadlakondaswetha, line 2082: rename it to writeToBuffer

internal/fs/fs.go
```diff
@@ -2077,13 +2077,31 @@ func (fs *fileSystem) WriteFile(
...
+			return err
+		}
 	}
 
 	return
 }
 
+func (fs *fileSystem) shouldTriggerBufferEnabledWrite(in *inode.FileInode) bool {
+	if fs.mountConfig.WriteConfig.EnableStreamingWrites &&
+		!fs.mountConfig.WriteConfig.CreateEmptyFile &&
```
vadlakondaswetha, line 2098: do we need these two conditions?
in.IsLocal is true only when createEmptyFile=false.

internal/fs/writebuffer/buffer.go
```diff
@@ -0,0 +1,18 @@
+package writebuffer
```
vadlakondaswetha, line 1: the filename and interface name has to match.

Make the package name as buffer.
interface as WriteBuffer and file name as writebuffer.go.

memory buffer as Inmemorywritebuffer.go.

internal/fs/writebuffer/memoryBuffer.go
```diff
@@ -0,0 +1,32 @@
+package writebuffer
```
vadlakondaswetha, line 1: dont use camelCase in file names
memoryBuffer.go -> memorybuffer.go

vadlakondaswetha (commented):
I didn't follow why there are no unit tests in this PR. Atleast for the methods added in file.go should have unit tests.

internal/fs/writebuffer/buffer.go
ashmeenkaur: should filename and struct name also match? shall I rename MemoryBuffer to InMemoryWriteBuffer?

internal/fs/fs.go
ashmeenkaur: removed 	`!fs.mountConfig.WriteConfig.CreateEmptyFile` condition

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,18 @@
+package buffer
+
+type WriteBuffer interface {
+	// Create creates a buffer of 2*size passed as a parameter.
+	Create(size int) WriteBuffer
```
gargnitingoogle, line 5: I would suggest avoiding Create function from the interface as it'll vary a lot between the implementations. Interfaces should only define their core functionality, which in this case are Write, Upload, Status etc.

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,18 @@
...
+
+	// Write writes to the buffer.
+	Write(data []byte, offset int64) error
+
+	// Upload asynchronously uploads written data to GCS.
+	Upload()
+
+	// Status fetches the stat details of buffer.
+	Status()
```
gargnitingoogle, line 14: It should return something. Probably enumerate constants above ?

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,18 @@
+package buffer
+
+type WriteBuffer interface {
+	// Create creates a buffer of 2*size passed as a parameter.
+	Create(size int) WriteBuffer
+
+	// Write writes to the buffer.
+	Write(data []byte, offset int64) error
```
gargnitingoogle, line 8: It should also return number of bytes written.

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,18 @@
...
+
+	// Upload asynchronously uploads written data to GCS.
+	Upload()
```
gargnitingoogle, line 11: It should return error at least, and probably some kind of status value/struct.

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,18 @@
...
+
+	// Status fetches the stat details of buffer.
+	Status()
+
+	// Destroy destorys the buffer when upload to GCS is finalized.
+	Destroy()
```
gargnitingoogle, line 17: I'd suggest returning error here at least.

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,18 @@
+package buffer
+
+type WriteBuffer interface {
```
gargnitingoogle, line 3: Add brief description the purpose and functionality of this interface.

gargnitingoogle (changes requested):
Reviewed the WriteBuffer interface. Need to review the other files too.

internal/fs/buffer/writebuffer.go
gargnitingoogle, line 3: This interface should extend the ioReader interface, example [here](https://stackoverflow.com/questions/48643906/how-to-extend-a-go-interface-to-another-interface) .

gargnitingoogle (commented):
Added another comment.

internal/fs/buffer/writebuffer.go
ashmeenkaur: It is hard to predict all the parameters and values that will be required for specific methods. This PR is just for buffer creation. The exact function parameters will be taken care of in their respective separate PRs and would probably require more iterations.

internal/fs/buffer/writebuffer.go
ashmeenkaur: This PR is just for buffer creation. This will be taken care of in a separate PR.

internal/fs/buffer/writebuffer.go
ashmeenkaur: This PR is just for buffer creation. This will be taken care of in a separate PR.

internal/fs/buffer/writebuffer.go
ashmeenkaur: This PR is just for buffer creation. This will be taken care of in a separate PR.

internal/fs/buffer/writebuffer.go
ashmeenkaur: On it. 
I believe you need it for go client methods. Could you please share function signatures for go client changes? Asking so that I can include any such changes which might be missing here. :)

internal/fs/buffer/writebuffer.go
ashmeenkaur: I would suggest let's keep it like this for now. We can remove it from the interface later incase that doesn't fit while working on onDisk buffer.

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,26 @@
+package buffer
+
+import "io"
+
+// WriteBuffer is an interface that buffers the data to be written to GCS during
+// the write flow.
+// WriteBuffer is used only in create new file flow with sequential writes and
+// at any point in time, only 2x of the configured buffer size is stored in the
+// write buffer.
+type WriteBuffer interface {
```
gargnitingoogle, line 10: @ashmeenkaur as discussed offline, let's keep only those changes in the PR and in the shared branch which have matured enough to be shared and reviewed. For now, I think you only need a Write function in the WriteBuffer interbace, let's remove the others.

gargnitingoogle (changes requested):
Added a comment

internal/fs/writebuffer/buffer.go
vadlakondaswetha: yes. atleast to begin with they should match. If we add more structs in the same file in future, its okay

vadlakondaswetha: Also updated the comment, please use underscores in filenames. thats the pattern we are following in gcsfuse.

internal/fs/buffer/writebuffer.go
```diff
@@ -0,0 +1,26 @@
...
+	io.Reader
```
vadlakondaswetha, line 11: still not clear to me why we need this. add it when go-client library methods are added/

internal/fs/buffer/writebuffer.go
ashmeenkaur: as discussed offline, I am not adding the ioReader interface as we may not require it. As @vadlakondaswetha suggested, we can add it later if required while integrating with go storage client interface.

internal/fs/buffer/write_buffer.go
```diff
@@ -0,0 +1,14 @@
+package buffer
+
+// WriteBuffer is an interface that buffers the data to be written to GCS during
+// the write flow.
+// WriteBuffer is used only in create new file flow with sequential writes and
+// at any point in time, only 2x of the configured buffer size is stored in the
+// write buffer.
+type WriteBuffer interface {
+	// Create creates a buffer of 2*size passed as a parameter.
+	Create(size int) WriteBuffer
```
gargnitingoogle, line 10: Is the size here in MiBs or in bytes ?

internal/fs/buffer/write_buffer.go
```diff
@@ -0,0 +1,14 @@
...
+
+	// Write writes to the buffer.
+	Write(data []byte, offset int64) error
```
gargnitingoogle, line 13: Should be called WriteAt as it's taking an offet.

internal/fs/buffer/write_buffer.go
gargnitingoogle: Also, can't it fail ? If it can, return an error also.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -0,0 +1,23 @@
...
+type InMemoryWriteBuffer struct {
+	// buffer will hold the data to be written to GCS. Buffer holds the data for 2
+	// GCS write calls.
+	buffer *bytes.Buffer
+	// chunkSize is the size of data to be written in one GCS call.
+	chunkSize int
+}
+
+func (b *InMemoryWriteBuffer) Create(sizeInMB int) WriteBuffer {
+	b.chunkSize = sizeInMB * 1024 * 1024
```
gargnitingoogle, line 16: Pull out the 1024*1024 out as a constant

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -0,0 +1,23 @@
...
+
+type InMemoryWriteBuffer struct {
+	// buffer will hold the data to be written to GCS. Buffer holds the data for 2
+	// GCS write calls.
+	buffer *bytes.Buffer
+	// chunkSize is the size of data to be written in one GCS call.
+	chunkSize int
+}
+
+func (b *InMemoryWriteBuffer) Create(sizeInMB int) WriteBuffer {
```
gargnitingoogle, line 15: I'm unclear about this signature. I thought InMemoryWriteBuffer was supposed to implement the WriteBuffer interface, wasn't it? If yes, why is it returning a WriteBuffer from an instance of InMemoryWriteBuffer ?

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -0,0 +1,56 @@
...
+func (t *MemoryBufferTest) TearDown() {}
+
+// //////////////////////////////////////////////////////////////////////
+// Tests
+// //////////////////////////////////////////////////////////////////////
+
+func (t *MemoryBufferTest) TestCreateInMemoryBuffer() {
+	bufferSize := 10
+
+	t.mb.Create(bufferSize)
```
gargnitingoogle, line 51: As Create returns something, we should assert on that return value.

internal/fs/fs.go
```diff
@@ -2077,13 +2077,30 @@ func (fs *fileSystem) WriteFile(
...
+		if err := in.Write(ctx, op.Data, op.Offset); err != nil {
+			return err
+		}
 	}
 
 	return
 }
 
+func (fs *fileSystem) shouldTriggerBufferEnabledWrite(in *inode.FileInode) bool {
+	if fs.mountConfig.WriteConfig.EnableStreamingWrites &&
```
gargnitingoogle, line 2097: `return fs.mountConfig.WriteConfig.EnableStreamingWrites &&
		in.IsLocal()`

instead of the if-else.

internal/fs/fs.go
```diff
@@ -2077,13 +2077,30 @@ func (fs *fileSystem) WriteFile(
...
+		// Trigger the temp file flow.
+		if err := in.Write(ctx, op.Data, op.Offset); err != nil {
+			return err
+		}
 	}
 
 	return
 }
 
+func (fs *fileSystem) shouldTriggerBufferEnabledWrite(in *inode.FileInode) bool {
```
gargnitingoogle, line 2096:
```suggestion
func (fs fileSystem) shouldTriggerBufferEnabledWrite(in *inode.FileInode) bool {
```

internal/fs/inode/file.go
```diff
@@ -77,6 +78,10 @@ type FileInode struct {
 	// authoritative.
 	content gcsx.TempFile
 
+	// Represents the buffer associated with inode which stores the data to be
+	// written to GCS.
+	buffer buffer.WriteBuffer
```
gargnitingoogle, line 83:
```suggestion
	writeBuffer buffer.WriteBuffer
```

to avoid confusion

internal/fs/inode/file.go
```diff
@@ -471,6 +476,29 @@ func (f *FileInode) Write(
 	return
 }
 
+func (f *FileInode) ensureBuffer(bufferSize int) {
+	if f.buffer != nil {
+		return
+	}
+	if bufferSize <= 50 {
```
gargnitingoogle, line 483: We really should rename bufferSize to bufferSizeInMb to make it unambiguous.

internal/fs/inode/file.go
```diff
@@ -471,6 +476,29 @@ func (f *FileInode) Write(
...
+		f.buffer = &buffer.InMemoryWriteBuffer{}
```
gargnitingoogle, line 484: Will this throw away the data accumulated in the previous write-buffer?

internal/fs/inode/file.go
```diff
@@ -471,6 +476,29 @@ func (f *FileInode) Write(
...
+	}
+	// TODO: else assign on-disk buffer to f.buffer.
+	f.buffer.Create(bufferSize)
```
gargnitingoogle, line 487: Need to store the returned value in something

gargnitingoogle (changes requested):
Added some comments.

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: The function signatures should match for struct to implement an interface, that is why we were returning the interface instead of struct previously. 

However I have refactored to remove create method as suggested previously from the interface. PTAL again. Resolving all the comments related to this change.

internal/fs/fs.go
ashmeenkaur: I didn't follow this change. Can you please elaborate why we should not use pointer receiver?

internal/fs/inode/file.go
ashmeenkaur: this is stored at inode level. so until the inode is retained, the data will not be thrown.

internal/fs/inode/file.go
```diff
@@ -471,6 +476,28 @@ func (f *FileInode) Write(
 	return
 }
 
+func (f *FileInode) ensureBuffer(bufferSizeMB int) {
+	if f.writeBuffer != nil {
+		return
+	}
+	if bufferSizeMB <= 50 {
```
gargnitingoogle, line 483: @ashmeenkaur we should put 50 in some config somewhere e.g maxMemSize

gargnitingoogle (changes requested):
Added a minor comment

gargnitingoogle (approved):
LGTM

This is based on our discussion of what's needed next, and to unblock the progress of the feature.

internal/fs/fs.go
```diff
@@ -2077,13 +2077,26 @@ func (fs *fileSystem) WriteFile(
 	defer in.Unlock()
 
 	// Serve the request.
-	if err := in.Write(ctx, op.Data, op.Offset); err != nil {
-		return err
+	if fs.shouldTriggerBufferEnabledWrite(in) {
```
vadlakondaswetha, line 2080: for local inodes, we are creating temp file in createFile flow: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/fs.go#L1491C34-L1491C34

This is to cover scenarios where setInodeAttributes is called after createFile and before writeFile. At this point of time, its not clear how will you handle setInodeAttributes, add a TODO atleast saying we need to move to createLocalFile if setInodeattributes doesn't work.

internal/fs/inode/file_test.go
```diff
@@ -928,3 +928,22 @@ func (t *FileTest) UnlinkLocalFile() {
 	AssertNe(nil, err)
 	AssertEq("gcs.NotFoundError: Object test not found", err.Error())
 }
+
+func (t *FileTest) TestMultipleCallsToWriteToBufferCreatesBufferOnce() {
```
vadlakondaswetha, line 932: unit tests for ensureBuffer to validate that inmemory buffer is created for <50MB

internal/fs/fs.go
ashmeenkaur: As discussed over chat, created an empty buffer object at the time of create file operation and allocated memory to it before write.
Also added a TODO for mtime attribute.

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
ashmeenkaur: changed to a global variable.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -0,0 +1,47 @@
...
+	// 2 GCS write calls.
+	buffer *bytes.Buffer
+}
+
+// EmptyInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
+// set to nil. Memory is allocated to the buffer when the first write call comes.
+// This avoids unnecessarily bloating GCSFuse memory consumption.
+//
+// To allocate memory to the buffer, use WriteBuffer.AllocateBuffer.
+func EmptyInMemoryWriteBuffer() *InMemoryWriteBuffer {
```
vadlakondaswetha, line 32: instead of empty call it CreateInMemoryWriteBuffer or InitInMemoryWriteBuffer.

Also InMemoryWriteBuffer is an public struct- why do you need seperate method for initialization. I am assuming you need it for setting mtime attributes.

internal/fs/buffer/write_buffer.go
```diff
@@ -0,0 +1,41 @@
...
+// WriteBuffer is used only in create new file flow with sequential writes and
+// at any point in time, only 2x of the configured buffer size is stored in the
+// write buffer.
+type WriteBuffer interface {
+	// WriteAt writes at an offset to the buffer.
+	WriteAt(data []byte, offset int64) error
+
+	// AllocateBuffer assigns a buffer of 2*configured buffer size to the WriteBuffer.
+	// AllocateBuffer should be called before any write calls to the buffer.
+	AllocateBuffer(sizeInMB int)
```
vadlakondaswetha, line 40: nit: move this first followed by WriteAt. It makes the definition clear while reading.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -0,0 +1,47 @@
...
+	b := &InMemoryWriteBuffer{}
+	// TODO: set mtime attribute.
+	return b
+}
+
+func (b *InMemoryWriteBuffer) AllocateBuffer(sizeInMB int) {
```
vadlakondaswetha, line 38: allocatebuffer gives the wrong impression that we can update the buffer size. instead call it InitializeBuffer.

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: > why do you need seperate method for initialization. I am assuming you need it for setting mtime attributes.
Yes, added a TODO for the same (line 34)

vadlakondaswetha (approved):

PR #1406, opened 2023-09-29
Adding cache config

sethiay:
### Description
Adding cache config for file cache.

NA

### Testing details
1. Manual - Mounted GCSFuse with file cache attributes (they are just ignored)
5. Unit tests - Added unit tests wherever applicable.
6. Integration tests - NA

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -235,6 +235,11 @@ func newApp() (app *cli.App) {
 				Usage: "Param for exponential backoff algorithm, which is used to increase waiting time b/w two consecutive retries.",
 			},
 
+			cli.BoolFlag{
+				Name:  "experimental-local-file-cache",
+				Usage: "Experimental: Cache GCS files on local disk for reads.",
+			},
+
 			cli.StringFlag{
 				Name:  "temp-dir",
 				Value: "",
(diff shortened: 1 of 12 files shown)


internal/config/mount_config.go
```diff
@@ -24,9 +24,29 @@ type LogConfig struct {
 	FilePath string      `yaml:"file-path"`
 }
 
+type FileCacheConfig struct {
+	Path                    string `yaml:"path"`
+	Size                    int64  `yaml:"size-mb"`
```
vadlakondaswetha, line 29: we should keep keep the yaml names and properties names similar as much as possible.
SizeMB
TTLInSec etc

internal/config/mount_config.go
```diff
@@ -24,9 +24,29 @@ type LogConfig struct {
...
+	TTL                     int64  `yaml:"ttl-sec"`
+	CacheFileForRandomReads bool   `yaml:"cache-file-for-random-reads"`
+}
+
+type MetadataCacheConfig struct {
+	Capacity int64 `yaml:"capacity"`
+	TTL      int64 `yaml:"ttl-sec"`
+}
+
+type TypeCacheConfig struct {
```
vadlakondaswetha, line 39: dont do these extra changes. We are going to merge typecache and stat-cache ttl in this release

internal/config/mount_config.go
```diff
@@ -24,9 +24,29 @@ type LogConfig struct {
...
+	TTL              int64 `yaml:"ttl-sec"`
+	CacheNonexistent bool  `yaml:"cache-nonexistent"`
```
vadlakondaswetha, line 41: We want to avoid changes around flags as much as possible. Until we decide on a plan for type-cache dont move around anything

internal/config/mount_config.go
```diff
@@ -24,9 +24,29 @@ type LogConfig struct {
...
 
+type FileCacheConfig struct {
+	Path                    string `yaml:"path"`
+	Size                    int64  `yaml:"size-mb"`
+	TTL                     int64  `yaml:"ttl-sec"`
+	CacheFileForRandomReads bool   `yaml:"cache-file-for-random-reads"`
+}
+
+type MetadataCacheConfig struct {
+	Capacity int64 `yaml:"capacity"`
```
vadlakondaswetha, line 35: this is going to be sizeInMB. Also why are you touching metadata cache configs now.
Focus only on filecache

internal/config/mount_config.go
```diff
@@ -24,9 +24,29 @@ type LogConfig struct {
...
+	TTL              int64 `yaml:"ttl-sec"`
+	CacheNonexistent bool  `yaml:"cache-nonexistent"`
+}
+
 type MountConfig struct {
-	WriteConfig `yaml:"write"`
-	LogConfig   `yaml:"logging"`
+	WriteConfig         `yaml:"write"`
+	LogConfig           `yaml:"logging"`
+	FileCacheConfig     `yaml:"file-cache"`
```
vadlakondaswetha, line 47: can you check what is preferred in yaml underscore or hypen?

vadlakondaswetha (commented):
I would suggest focus only on filecache config for now.

Dont do any of these breaking flag removal changes until required.

internal/config/mount_config.go
```diff
@@ -24,9 +24,17 @@ type LogConfig struct {
 	FilePath string      `yaml:"file-path"`
 }
 
+type FileCacheConfig struct {
+	Path                    string `yaml:"path"`
```
vadlakondaswetha, line 28: path and cachefileForRandomReads - I would suggest to use the names mentioned in design doc until we finalize the names.

flags.go
```diff
@@ -234,11 +235,6 @@ func newApp() (app *cli.App) {
 				Usage: "Param for exponential backoff algorithm, which is used to increase waiting time b/w two consecutive retries.",
 			},
 
-			cli.BoolFlag{
-				Name:  "experimental-local-file-cache",
```
vadlakondaswetha, line 238: why are you removing this? Dont make any changes to existing flow

internal/config/mount_config.go
sethiay: As discussed offline, this is dependent on the language around which the yaml is used. And if we consider that, the naming should be like jobName. However, that will not be suitable as all our flags are like job-name. So, we will go with hypens job-name. Also, we will go with hyphens at all levels in config file i.e. 
config.yaml

- first-level:
  - second-level: some_value

The above makes things consistent.

vadlakondaswetha (approved):

PR #1411, opened 2023-10-03
[Async object upload 2] Add struct chunkUploader

gargnitingoogle:
Adds a new struct chunkUploader which implements the ChunkUploader interface.

### Description
A chunkUploader is used to write content to a new GCS object asynchronously.

This is followed up in and [Async object upload 3](https://github.com/GoogleCloudPlatform/gcsfuse/pull/1417).

NA

### Testing details
1. Manual - NA
2. Unit tests - All existing unit tests passed. New tests needed for chunkUploader.
3. Integration tests - NA

### DIFF
--- a/internal/storage/chunk_uploader.go
+++ b/internal/storage/chunk_uploader.go
@@ -156,7 +156,7 @@ func (uploader *chunkUploader) Upload(ctx context.Context, contents io.Reader) e
 
 	select {
 	case <-ctx.Done():
-		return fmt.Errorf("uploader.Upload() timed out")
+		return fmt.Errorf("uploader.Upload() failed: %v", ctx.Err())
 	case <-ioCopyComplete:
 		if err != nil {
 			uploader.state = UploadError
@@ -213,7 +213,7 @@ func (uploader *chunkUploader) Close(ctx context.Context) (*gcs.Object, error) {
 
 	select {
 	case <-ctx.Done():
-		return nil, fmt.Errorf("uploader.Close() timed out")
+		return nil, fmt.Errorf("uploader.Close() failed: %v", ctx.Err())
 	case <-writerCloseComplete:
 		if err != nil {
 			var gErr *googleapi.Error
--- a/internal/storage/chunk_uploader.go
+++ b/internal/storage/chunk_uploader.go
@@ -1,0 +1,233 @@
+package storage
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"sync"
+	"sync/atomic"
+
+	"cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
+	"github.com/googlecloudplatform/gcsfuse/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/internal/storage/storageutil"
+	"google.golang.org/api/googleapi"
+)
+
+type UploaderState int
+
+const (
+	// Initialized is the state of successful initialization, no uploads so far.
+	Initialized UploaderState = iota
+	// Uploading is the state when an asynchronous upload is in progress.
+	Uploading
+	// UploadError is the state when an Upload failed.
+	UploadError
+	// Closed is the state of an uploader which has been finalized.
+	Closed
+)
+
+// A chunkUploader is an implementation of ChunkUploader
+// interface, which uses storage.Writer from go-storage-client
+// for resumable upload.
+//
+// It also stores the current state of the uploader.
+type chunkUploader struct {
+	// Internal objects for business logic.
+	writer     *storage.Writer
+	objectName string
+	mu         sync.Mutex
+
+	// Attributes for providing updates to user.
+	totalUploadInitiatedSoFar int64
+	// Total number of bytes successfully written by this writer so far.
+	totalBytesUploadSuccessfullySoFar atomic.Int64
+	userProgressFunc                  func(int64)
+
+	// Internal state for lifecycle management.
+	state UploaderState
+}
+
+// NewChunkUploader creates a new instance of chunkUploader,
+// for the given inputs.
+func NewChunkUploader(ctx context.Context, obj *storage.ObjectHandle, req *gcs.CreateObjectRequest, chunkSize int, progressFunc func(int64)) (gcs.ChunkUploader, error) {
+	if ctx == nil {
+		return nil, fmt.Errorf("ctx is nil")
+	}
+	if obj == nil || req == nil {
+		return nil, fmt.Errorf("nil ObjectHandle or CreateObjectRequest")
+	}
+
+	if obj.ObjectName() != req.Name {
+		return nil, fmt.Errorf("names of passed ObjectHandle and CreateObjectRequest don't match: ObjectHandle.Name=%s createObjectRequest.Name=%s", obj.ObjectName(), req.Name)
+	}
+
+	if req.GenerationPrecondition != nil && *req.GenerationPrecondition != 0 {
+		return nil, fmt.Errorf("request received for pre-existing object %s, supported only for new objects", req.Name)
+	}
+
+	if chunkSize <= 0 {
+		return nil, fmt.Errorf("chunkSize <= 0")
+	}
+
+	// Raw initialization.
+	uploader := chunkUploader{}
+
+	// Store references to necessary parameters.
+	uploader.objectName = obj.ObjectName()
+	uploader.userProgressFunc = progressFunc
+
+	// Create a new writer with the requested attributes, using Go Storage Client.
+	// NewWriter never returns nil, so no nil-check is needed on it.
+	wc := obj.NewWriter(ctx)
+	wc = storageutil.SetAttrsInWriter(wc, req)
+	wc.ChunkSize = chunkSize
+	wc.ProgressFunc = func(n int64) {
+		uploader.totalBytesUploadSuccessfullySoFar.Store(n)
+		logger.Debugf("%d bytes copied so far for object/file %s. chunk-size = %d", n, req.Name, wc.ChunkSize)
+
+		if uploader.userProgressFunc != nil {
+			uploader.userProgressFunc(n)
+		}
+	}
+
+	uploader.writer = wc
+	uploader.state = Initialized
+
+	return &uploader, nil
+}
+
+// BytesUploadedSoFar returns the total number of bytes successfully uploaded
+// so far using this uploader.
+// This is thread-safe against the
+// in-progress calls to BytesUploadedSoFar/Upload/Close
+// invoked from other threads/go-routines.
+func (uploader *chunkUploader) BytesUploadedSoFar() int64 {
+	return uploader.totalBytesUploadSuccessfullySoFar.Load()
+}
+
+func (uploader *chunkUploader) readyToUpload() bool {
+	switch uploader.state {
+	case Initialized, Uploading:
+		return true
+	default:
+		return false
+	}
+}
+
+func (uploader *chunkUploader) readyToClose() bool {
+	switch uploader.state {
+	case Initialized, Uploading, UploadError:
+		return true
+	default:
+		return false
+	}
+}
+
+// Upload uploads the passed content to GCS.
+// This waits (using mutex) until the completion of
+// in-progress calls to BytesUploadedSoFar/Upload/Close
+// invoked from other threads/go-routines.
+// If contents contains more than n chunks, then it synchronizes on the
+// upload of the first n chunks, and queues up the last chunk,
+// before returning. Ideally, the caller would want to invoke this
+// function in a goroutine in case of contents containing multiple chunks.
+func (uploader *chunkUploader) Upload(ctx context.Context, contents io.Reader) error {
+	if ctx == nil {
+		return fmt.Errorf("ctx is nil")
+	}
+
+	uploader.mu.Lock()
+	defer uploader.mu.Unlock()
+
+	if !uploader.readyToUpload() {
+		return fmt.Errorf("uploader not ready to upload: object: %s, status = %v, writer: %v", uploader.objectName, uploader.state, uploader.writer)
+	}
+
+	var n int64
+	var err error
+	ioCopyComplete := make(chan bool)
+	go func() {
+		n, err = io.Copy(uploader.writer, contents)
+		ioCopyComplete <- true
+	}()
+
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("uploader.Upload() timed out")
+	case <-ioCopyComplete:
+		if err != nil {
+			uploader.state = UploadError
+			return fmt.Errorf("upload failed for object %s: totalSizeUploaded-so-far=%d, successfully-uploaded-in-last-upload=%d, chunk-size=%d, %v", uploader.objectName, uploader.BytesUploadedSoFar(),
+				n, uploader.writer.ChunkSize,
+				err)
+		}
+		if n == 0 {
+			return nil
+		}
+
+		uploader.totalUploadInitiatedSoFar += n
+		uploader.state = Uploading
+		return nil
+	}
+}
+
+// Close finalizes the chunk uploads and returns the
+// created GCS object.
+// If a chunk upload was in progress at the time of call,
+// it will be waited on and be completed before going ahead
+// with the finalization.
+// This waits (using mutex) until the completion of
+// in-progress calls to BytesUploadedSoFar/Upload/Close
+// invoked from other threads/go-routines.
+func (uploader *chunkUploader) Close(ctx context.Context) (*gcs.Object, error) {
+	if ctx == nil {
+		return nil, fmt.Errorf("ctx is nil")
+	}
+
+	if uploader == nil {
+		return nil, fmt.Errorf("uploader is nil")
+	}
+
+	uploader.mu.Lock()
+	defer uploader.mu.Unlock()
+
+	defer func() {
+		uploader.state = Closed
+		uploader.writer = nil
+	}()
+
+	if !uploader.readyToClose() {
+		return nil, fmt.Errorf("improper state (%v) for finalizing object %s", uploader.state, uploader.objectName)
+	}
+
+	var err error
+	writerCloseComplete := make(chan bool, 1)
+
+	go func() {
+		err = uploader.writer.Close()
+		writerCloseComplete <- true
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil, fmt.Errorf("uploader.Close() timed out")
+	case <-writerCloseComplete:
+		if err != nil {
+			var gErr *googleapi.Error
+			if errors.As(err, &gErr) {
+				if gErr.Code == http.StatusPreconditionFailed {
+					return nil, &gcs.PreconditionError{Err: err}
+				}
+			}
+			return nil, fmt.Errorf("error in closing : %w", err)
+		}
+
+		// Retrieving the attributes of the created object.
+		attrs := uploader.writer.Attrs()
+		// Converting attrs to type *Object.
+		return storageutil.ObjectAttrsToBucketObject(attrs), nil
+	}
+}
--- a/internal/storage/chunk_uploader_test.go
+++ b/internal/storage/chunk_uploader_test.go
@@ -334,6 +334,10 @@ func (t *ChunkUploaderTest) TestUploadWithTimeout() {
 	defer cancel()
 	err := uploader.Upload(ctx, strings.NewReader(content))
 	AssertNe(nil, err)
+	expectedErrorSubstring := "context deadline exceeded"
+	if !strings.Contains(err.Error(), expectedErrorSubstring) {
+		AddFailure("Error message (%s) doesn't contain expected sustring(%s)", err.Error(), expectedErrorSubstring)
+	}
 }
 
 func (t *ChunkUploaderTest) TestUploadEmptyContent() {
@@ -437,6 +441,10 @@ func (t *ChunkUploaderTest) TestCloseWithTimeout() {
 	defer cancel()
 	_, err := uploader.Close(ctx)
 	AssertNe(nil, err)
+	expectedErrorSubstring := "context deadline exceeded"
+	if !strings.Contains(err.Error(), expectedErrorSubstring) {
+		AddFailure("Error message (%s) doesn't contain expected sustring(%s)", err.Error(), expectedErrorSubstring)
+	}
 }
 
 func (t *ChunkUploaderTest) TestDoubleClosure() {
--- a/internal/storage/chunk_uploader_test.go
+++ b/internal/storage/chunk_uploader_test.go
@@ -1,0 +1,521 @@
+package storage
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+	"io"
+	"math"
+	"strings"
+	"testing"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/internal/storage/gcs"
+	. "github.com/jacobsa/ogletest"
+	"google.golang.org/api/googleapi"
+)
+
+func TestChunkUploader(t *testing.T) { RunTests(t) }
+
+////////////////////////////////////////////////////////////////////////
+// Boilerplate
+////////////////////////////////////////////////////////////////////////
+
+type ChunkUploaderTest struct {
+	fakeStorage FakeStorage
+	content     string
+	req         *gcs.CreateObjectRequest
+	obj         *storage.ObjectHandle
+}
+
+var _ SetUpInterface = &ChunkUploaderTest{}
+var _ TearDownInterface = &ChunkUploaderTest{}
+
+func init() { RegisterTestSuite(&ChunkUploaderTest{}) }
+
+func (t *ChunkUploaderTest) SetUp(ti *TestInfo) {
+	t.fakeStorage = NewFakeStorage()
+	storageHandle := t.fakeStorage.CreateStorageHandle()
+	bucketHandle := storageHandle.BucketHandle(TestBucketName, "")
+
+	AssertNe(nil, bucketHandle)
+
+	t.content = "Creating a new object"
+	t.req = &gcs.CreateObjectRequest{
+		Name:     "test_object",
+		Contents: strings.NewReader(t.content),
+	}
+
+	t.obj = bucketHandle.bucket.Object(t.req.Name)
+
+	AssertNe(nil, t.obj)
+	AssertEq("test_object", t.obj.ObjectName())
+}
+
+func (t *ChunkUploaderTest) TearDown() {
+	t.fakeStorage.ShutDown()
+}
+
+////////////////////////////////////////////////////////////////////////
+// Helper functions
+////////////////////////////////////////////////////////////////////////
+
+// testNewChunkUploader tests a newly created ChunkUploader and asserts
+// if it hasn't been created properly.
+// If properly created, it casts it to chunkUploader pointer and returns.
+func (t *ChunkUploaderTest) testNewChunkUploader(ctx context.Context,
+	obj *storage.ObjectHandle,
+	req *gcs.CreateObjectRequest,
+	chunkSize int,
+	progressFunc func(int64)) *chunkUploader {
+	uploader, err := NewChunkUploader(ctx, obj, req, chunkSize, progressFunc)
+	AssertNe(nil, uploader)
+	AssertEq(nil, err)
+
+	chunkUploader := uploader.(*chunkUploader)
+
+	AssertNe(nil, chunkUploader)
+	AssertEq(Initialized, chunkUploader.state)
+	AssertEq(0, chunkUploader.totalUploadInitiatedSoFar)
+	AssertEq(0, chunkUploader.BytesUploadedSoFar())
+	AssertNe(nil, chunkUploader.writer)
+	AssertEq(req.Name, chunkUploader.objectName)
+	AssertEq(chunkSize, chunkUploader.writer.ChunkSize)
+
+	return chunkUploader
+}
+
+func (t *ChunkUploaderTest) testUpload(uploader *chunkUploader, contents io.Reader) {
+	err := uploader.Upload(context.Background(), contents)
+	AssertEq(nil, err)
+	AssertEq(Uploading, uploader.state)
+}
+
+func (t *ChunkUploaderTest) testUploadEmptyContent(uploader *chunkUploader, contents io.Reader) {
+	err := uploader.Upload(context.Background(), contents)
+	AssertEq(nil, err)
+}
+
+func (t *ChunkUploaderTest) testFailedUpload(uploader *chunkUploader, contents io.Reader) {
+	err := uploader.Upload(context.Background(), contents)
+	AssertNe(nil, err)
+	AssertEq(UploadError, uploader.state)
+}
+
+func (t *ChunkUploaderTest) testClose(uploader *chunkUploader, objName string, len int) {
+	o, err := uploader.Close(context.Background())
+	AssertEq(nil, err)
+	AssertNe(nil, o)
+	AssertEq(Closed, uploader.state)
+	AssertEq(objName, o.Name)
+	AssertEq(len, o.Size)
+}
+
+func (t *ChunkUploaderTest) testFailedClose(uploader *chunkUploader, expectedErrStr string) {
+	_, err := uploader.Close(context.Background())
+	AssertNe(nil, err)
+	AssertTrue(uploader.state == Closed)
+	if !strings.Contains(err.Error(), expectedErrStr) {
+		AddFailure("chunkUploader.Close() failed with wrong error. "+
+			"Expected error-substring: \"%s\", Actual error-string: \"%s\"",
+			expectedErrStr, err.Error())
+	}
+}
+
+type failingContentReader struct{}
+
+func (ilcr *failingContentReader) Read(p []byte) (n int, err error) {
+	return 0, fmt.Errorf("read failed as intended")
+}
+
+func genRandomContent(size int) string {
+	b := make([]byte, size)
+
+	_, err := rand.Read(b)
+	AssertEq(nil, err)
+
+	return string(b)
+}
+
+func numWholeChunksInContentSize(contentSize, chunkSize int) int {
+	return int(math.Floor(float64(contentSize) / float64(chunkSize)))
+}
+
+func contentSizeFromChunkSizeMultiplier(chunkSize int, chunkSizeMultiplier float32) int {
+	return int(float32(chunkSize) * chunkSizeMultiplier)
+}
+
+func createVarsForMultipleUploads(chunkSize int,
+	chunkSizeMultipliersForUploads []float32) (contentsForNthUploads []string,
+	numWholeChunksUptoNthUploads []int) {
+	var totalContentSizesUptoNthUploads []int
+	for i := range chunkSizeMultipliersForUploads {
+		contentSizeForUpload := contentSizeFromChunkSizeMultiplier(chunkSize,
+			chunkSizeMultipliersForUploads[i])
+		content := genRandomContent(contentSizeForUpload)
+		contentsForNthUploads = append(contentsForNthUploads, content)
+
+		if i == 0 {
+			totalContentSizesUptoNthUploads = append(totalContentSizesUptoNthUploads,
+				contentSizeForUpload)
+		} else {
+			totalContentSizesUptoNthUploads = append(totalContentSizesUptoNthUploads,
+				totalContentSizesUptoNthUploads[i-1]+contentSizeForUpload)
+		}
+
+		numWholeChunksUptoNthUploads =
+			append(numWholeChunksUptoNthUploads,
+				numWholeChunksInContentSize(totalContentSizesUptoNthUploads[i], chunkSize))
+	}
+
+	return
+}
+
+// This function invokes multiple uploads of different sizes,
+// and different random contents using a chunkUploader.
+//
+// It verifies that all the upload calls succeeded, and
+// for all of them, upload callbacks were called on every
+// instance of content-size reaching multiples of chunkSize
+// and that the BytesUploadedSoFar was updated correctly
+// for the uploader.
+//
+// For example, let's take the following sequence of uploads.
+//
+// 1. In the first upload, we upload .25*chunkSize data,
+// then no callback will be received during/after that upload.
+//
+// 2. In the next (second) upload,  we upload 0.5*chunkSize data,
+// then again no callback will be received during/after that upload,
+// as only 0.75*chunkSize data has been received so far by the uploader
+// and a complete chunk has not been received yet.
+//
+// 3. In the next (third) callback, 1.5*chunkSize data is sent,
+// so total 2.25 chunks were uploaded upto that upload.
+// So, after this upload, we
+// expect 2 callbacks to be received, if we wait long enough.
+//
+// 4. In the next (fourth) callback, .775*chunkSize data is sent,
+// so total 3.025 chunks were uploaded upto that upload.
+// So, after this upload, we
+// expect 1 more callbacks to be received, if we wait long enough.
+//
+// This above test generalizes this scenario by taking any general
+// set of uploads through chunk-size-multipliers for uploaders.
+//
+// Keep the multiplier values as multiples of 1/chunkSize for simplicity of floating-point
+// calculations and comparisons.
+func (t *ChunkUploaderTest) testUploadMultipleUploads(chunkSize int, chunkSizeMultipliersForUploads []float32) {
+	// Set up inputs for uploads.
+	numUploads := len(chunkSizeMultipliersForUploads)
+	contentsForNthUploads, numWholeChunksUptoNthUploads :=
+		createVarsForMultipleUploads(chunkSize, chunkSizeMultipliersForUploads)
+	totalNumberOfWholeChunks := numWholeChunksUptoNthUploads[numUploads-1]
+	numBytesSuccessfullyUploadedSoFar := make(chan int64, totalNumberOfWholeChunks)
+	defer close(numBytesSuccessfullyUploadedSoFar)
+	ctx := context.Background()
+	var numCallbacks int
+	uploader := t.testNewChunkUploader(ctx, t.obj, t.req, chunkSize,
+		func(n int64) {
+			numCallbacks++
+			numBytesSuccessfullyUploadedSoFar <- n
+		})
+
+	chunkCallbackIndex := 0
+	progressFuncCallbackTimeout := 10 * time.Millisecond
+	for i := 0; i < numUploads; i++ {
+		t.testUpload(uploader, strings.NewReader(contentsForNthUploads[i]))
+
+		// For all the chunks completed during/after ith upload, confirm that its
+		// callbacks were received and with right values of n.
+		for ; chunkCallbackIndex < numWholeChunksUptoNthUploads[i]; chunkCallbackIndex++ {
+			select {
+			case n := <-numBytesSuccessfullyUploadedSoFar:
+				AssertEq((chunkCallbackIndex+1)*chunkSize, n)
+			case <-time.After(progressFuncCallbackTimeout):
+				AddFailure("Did not not receive write progressFunc callback for "+
+					"chunkSize=%v,chunkSizeMultipliersForUploads=%v,i=%v,chunkCallbackIndex=%v in %v",
+					chunkSize, chunkSizeMultipliersForUploads, i, chunkCallbackIndex, progressFuncCallbackTimeout)
+			}
+		}
+
+		AssertEq(chunkCallbackIndex, numCallbacks)
+		AssertEq(chunkCallbackIndex*chunkSize, uploader.BytesUploadedSoFar())
+	}
+}
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *ChunkUploaderTest) TestCreateWithImproperInputs() {
+	properReq := t.req
+	properObj := t.obj
+	diffCreateReq := &gcs.CreateObjectRequest{}
+	*diffCreateReq = *properReq
+	diffCreateReq.Name = "FakeName"
+	inputs := []struct {
+		ctx          context.Context
+		obj          *storage.ObjectHandle
+		req          *gcs.CreateObjectRequest
+		chunkSize    int
+		progressFunc func(int64)
+		errStr       string
+	}{{
+		errStr: "ctx is nil",
+	}, {
+		ctx:    context.Background(),
+		errStr: "nil ObjectHandle or CreateObjectRequest",
+	}, {
+		ctx:    context.Background(),
+		obj:    properObj,
+		req:    properReq,
+		errStr: "chunkSize <= 0",
+	}, {
+		ctx:       context.Background(),
+		obj:       properObj,
+		req:       diffCreateReq,
+		errStr:    "names of passed ObjectHandle and CreateObjectRequest don't match",
+		chunkSize: 100,
+	},
+	}
+
+	for _, input := range inputs {
+		uploader, err := NewChunkUploader(input.ctx, input.obj, input.req,
+			input.chunkSize, input.progressFunc)
+
+		AssertEq(nil, uploader)
+		AssertNe(nil, err)
+		AssertTrue(strings.Contains(err.Error(), input.errStr))
+	}
+}
+
+func (t *ChunkUploaderTest) TestCreateWithProperInputs() {
+	inputs := []struct {
+		ctx          context.Context
+		obj          *storage.ObjectHandle
+		req          *gcs.CreateObjectRequest
+		chunkSize    int
+		progressFunc func(int64)
+	}{{
+		ctx:       context.Background(),
+		obj:       t.obj,
+		req:       t.req,
+		chunkSize: 1,
+	}, {
+		ctx:       context.Background(),
+		obj:       t.obj,
+		req:       t.req,
+		chunkSize: 2 << 24,
+	},
+		{
+			ctx:          context.Background(),
+			obj:          t.obj,
+			req:          t.req,
+			chunkSize:    2 << 24,
+			progressFunc: func(int64) {},
+		},
+	}
+
+	for _, input := range inputs {
+		t.testNewChunkUploader(input.ctx, input.obj, input.req, input.chunkSize, input.progressFunc)
+	}
+}
+
+func (t *ChunkUploaderTest) TestUploadWithTimeout() {
+	obj := t.obj
+	chunkSize := googleapi.MinUploadChunkSize
+	contentSize := 10 * chunkSize
+	content := genRandomContent(contentSize)
+
+	uploader := t.testNewChunkUploader(context.Background(), obj, t.req, chunkSize, nil)
+	ctx, _ := context.WithTimeout(context.Background(), time.Millisecond)
+	err := uploader.Upload(ctx, strings.NewReader(content))
+	AssertNe(nil, err)
+}
+
+func (t *ChunkUploaderTest) TestUploadEmptyContent() {
+	// setup
+	obj := t.obj
+	chunkSize := googleapi.MinUploadChunkSize
+	content := ""
+
+	uploader := t.testNewChunkUploader(context.Background(), obj, t.req, chunkSize, nil)
+	t.testUploadEmptyContent(uploader, strings.NewReader(content))
+
+	AssertEq(0, uploader.totalUploadInitiatedSoFar)
+}
+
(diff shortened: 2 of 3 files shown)


internal/storage/streamed_object_writer_test.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (t *StreamedObjectWriterTest) TearDown() {
+}
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *StreamedObjectWriterTest) DummyWriter() {
```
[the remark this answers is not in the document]
gargnitingoogle: Done

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+	Uninitialized UploaderState = iota
+	// State of successful initialization, no writes so far.
+	Initialized
+	// No write is in progress, but there have been writes in the past.
+	Waiting
+	// Asynchronous write is in progress.
+	Writing
+	// Last write failed.
+	WriteError
+	// Close has been failed.
```
gargnitingoogle, line 30: Fix typo

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	sow.wc = wc
+	sow.state = Initialized
+
+	return &sow, nil
+}
+
+func (sow StreamedChunkUploader) ChunkSize() (int, error) {
+	if nil == sow.wc {
+		return 0, fmt.Errorf("writer not initiated")
```
gargnitingoogle, line 108:
```suggestion
		return 0, fmt.Errorf("writer not initialized")
```

internal/storage/streamed_chunk_uploader_test.go
```diff
@@ -0,0 +1,329 @@
...
+	// time.Sleep(1000 * time.Millisecond)
+	if numCallbacks > 0 {
+		AssertGe(numChunks, numCallbacks)
+		AssertGe(contentSize, numBytesUploaded)
+	}
+	AssertEq(numBytesUploaded, uploader.BytesUploadedSoFar())
+	AssertEq(contentSize, uploader.BytesUploadInitiatedSoFar())
+}
+
+func (t *StreamedChunkUploaderTest) TestUploadMultipleMultichunksUploads() {
```
gargnitingoogle, line 204:
```suggestion
func (t *StreamedChunkUploaderTest) TestUploadMultipleMultichunkUploads() {
```

internal/storage/streamed_chunk_uploader_test.go
```diff
@@ -0,0 +1,329 @@
...
+
+	for _, input := range inputs {
+		uploader, err := NewStreamedChunkUploader(input.ctx, input.obj, input.req, input.writeChunkSize, input.progressFunc)
+		AssertNe(nil, uploader)
+		AssertEq(nil, err)
+		AssertEq(Initialized, uploader.State())
+		AssertEq(0, uploader.BytesUploadInitiatedSoFar())
+		AssertEq(0, uploader.BytesUploadedSoFar())
+		chunkSize, err := uploader.ChunkSize()
+		AssertEq(nil, err)
```
gargnitingoogle, line 126: Duplicate line. remove!

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+// A StreamedChunkUploader is an implementation of ChunkUploader
+// interface, which use storage.Writer from go-storage-client
+// for resumable upload.
+//
+// It also stores state of how many bytes have been sent to it
+// to write, how many have been successfully written by it so far,
+// and the current state of the uploader.
+type StreamedChunkUploader struct {
```
vadlakondaswetha, line 41: Why should this struct be public. As the responsibility of this struct it to implement the chunkUploader interface, the init method should just return the interface type.

vadlakondaswetha, line 41: why is this called StreamedChunkUpload and not chunkUploader simply?

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+	// Implements the following interfaces.
```
vadlakondaswetha, line 42: If the struct implements all the interface methods, it is assumed that struct implements the interface. no need of this line.

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+	sow.state = Uninitialized
+
+	// store references to necessary parameters
+	sow.objectName = obj.ObjectName()
+	sow.userProgressFunc = progressFunc
+
+	// Create a NewWriter with the requested attributes, using Go Storage Client.
+	// Chuck size for resumable upload is default i.e. 16MB.
+	wc := obj.NewWriter(ctx)
+	// NewWriter never returns nil, so no nil-check here.
```
vadlakondaswetha, line 88: nit: add comments on top of line not below.

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+	// raw initialization
+	sow := StreamedChunkUploader{}
+	sow.state = Uninitialized
+
+	// store references to necessary parameters
+	sow.objectName = obj.ObjectName()
+	sow.userProgressFunc = progressFunc
+
+	// Create a NewWriter with the requested attributes, using Go Storage Client.
+	// Chuck size for resumable upload is default i.e. 16MB.
```
vadlakondaswetha, line 86: nit: chuck -> chunk

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+	sow.wc = wc
+	sow.state = Initialized
+
+	return &sow, nil
+}
+
+func (sow StreamedChunkUploader) ChunkSize() (int, error) {
+	if nil == sow.wc {
```
vadlakondaswetha, line 107: nit: sow.wc == nil

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+func (sow StreamedChunkUploader) State() UploaderState {
+	return sow.state
+}
+
+func (sow StreamedChunkUploader) BytesUploadInitiatedSoFar() int64 {
+	return sow.totalWriteInitiatedSoFar
+}
+
+func (sow StreamedChunkUploader) BytesUploadedSoFar() int64 {
+	return sow.totalWriteSucceededSoFar
```
vadlakondaswetha, line 122: why are these methods public, how will the caller use them?
I assume caller should use chunkUploader interface to upload and query the status.

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (sow *StreamedChunkUploader) readyToWrite() bool {
+	switch sow.state {
+	case Initialized:
```
vadlakondaswetha, line 127: why dont you combine everything as below:

switch sow.state {
	case
		Initialized,
		Waiting,
		Writing:
		return true
	}
	return false

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+func (sow *StreamedChunkUploader) readyToClose() bool {
+	switch sow.state {
+	case Initialized:
+		fallthrough
```
vadlakondaswetha, line 141: same here also.

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+	case Writing:
+		fallthrough
+	case WriteError:
+		return true
+	default:
+		return false
+	}
+}
+
+// Write uploads the given content
```
vadlakondaswetha, line 153: the comment is not clear. Please rephrase in simpler terms

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+// writer have been already by a previous calls to CreateObject
+// which created the object-writer.
+func (sow *StreamedChunkUploader) UploadChunkAsync(contents io.Reader) error {
+	if !sow.readyToWrite() || (sow.wc == nil) {
+		return fmt.Errorf("writer not ready to write: object: %s, status = %v, writer: %v", sow.objectName, sow.state, sow.wc)
+	}
+
+	n, err := io.Copy(sow.wc, contents)
+	sow.totalWriteInitiatedSoFar += n
+	if n == 0 || err == io.EOF {
```
vadlakondaswetha, line 166: why are you converting n==0 to io.EOF? just make it a nop when no data is passed?

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (sow *StreamedChunkUploader) readyToClose() bool {
+	switch sow.state {
+	case Initialized:
+		fallthrough
+	case Waiting:
+		fallthrough
+	case Writing:
+		fallthrough
```
vadlakondaswetha, line 145: what happens if the chunk upload is in progress and we call the close. Will the chunk gets uploaded successfully before finalizing the object?
If yes can you document it on the close method of chunkUploader interface.

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+		sow.state = WriteError
+		return fmt.Errorf("totalSizeUploaded-so-far=%d, successfully-uploaded-in-last-upload=%d, chunk-size=%d for object %s: %v", sow.totalWriteSucceededSoFar, n, sow.wc.ChunkSize,
+			sow.objectName, err)
+	}
+
+	sow.state = Writing
+	return nil
+}
+
+func (sow *StreamedChunkUploader) Close() (o *gcs.Object, err error) {
```
vadlakondaswetha, line 181: the purpose of assigning variable names to the object is to set the values using these names.


For eg: 
o =storageutil.ObjectAttrsToBucketObject(attrs)
return

If you are not doing that ever in the code, dont use variable names in the method definition.

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if req.GenerationPrecondition != nil && *req.GenerationPrecondition != 0 {
+		return nil, fmt.Errorf("request received for pre-existing object %s, supported only for new objects", req.Name)
+	}
+
+	if writeChunkSize <= 0 {
+		return nil, fmt.Errorf("writeChunkSize <= 0 not supported")
+	}
+
+	// raw initialization
```
vadlakondaswetha, line 77: nit: Please start all comments with capital letter and end with fullstop;.

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// store references to necessary parameters
+	sow.objectName = obj.ObjectName()
+	sow.userProgressFunc = progressFunc
+
+	// Create a NewWriter with the requested attributes, using Go Storage Client.
+	// Chuck size for resumable upload is default i.e. 16MB.
+	wc := obj.NewWriter(ctx)
+	// NewWriter never returns nil, so no nil-check here.
+
```
vadlakondaswetha, line 89: nit: remove new line here.

internal/storage/streamed_chunk_uploader.go
vadlakondaswetha: Are these methods only required in unit tests. If yes you should be able to access private methods since both the files are in same package. refer to bucket_handle_test

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+	if !sow.readyToWrite() || (sow.wc == nil) {
+		return fmt.Errorf("writer not ready to write: object: %s, status = %v, writer: %v", sow.objectName, sow.state, sow.wc)
+	}
+
+	n, err := io.Copy(sow.wc, contents)
+	sow.totalWriteInitiatedSoFar += n
+	if n == 0 || err == io.EOF {
+		// for io.EOF, don't change/append the error (message) as it makes it difficult to identify EOF for the caller.
+		sow.state = Waiting
+		return io.EOF
```
vadlakondaswetha, line 169: what happens if you try to write data < chunksize, will the writer wait till the next set of data comes?

internal/storage/streamed_chunk_uploader.go
```diff
@@ -0,0 +1,206 @@
...
+		sow.state = Destroyed
+		sow.wc = nil
+	}()
+
+	if !sow.readyToClose() {
+		return nil, fmt.Errorf("improper state for finalizing object %s", sow.objectName)
+	}
+
+	// We can't use defer to close the writer, because we need to close the
+	// writer successfully before calling Attrs() method of writer.
```
vadlakondaswetha, line 192: this comment doesn't make any sense. this method is triggered to finalize the upload and it has to be done synchronously.

vadlakondaswetha (commented):
Unit tests are not reviewed.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Made private.

gargnitingoogle: Renamed to chunkUploader.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: The null check was needed later, so I added the comment later, to keep the flow for the reader. Have moved before the line now.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: This comment line was outdated, so removed it.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Removed unused names.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Done.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: `BytesUploadedSoFar()` is declared inthe ChunkUploader interface, so I've kept it public, and have made the non-interface functions private.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Done.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Done. Simplified to one sentence.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Done. Removed the `n==0` case.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Done.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Done.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Removed the unnecessary and confusing comment.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: No, it doesn't wait, it just returns with io.EOF, which is a harmless error.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Done.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: @vadlakondaswetha Do you mean just the comment or the interface declaration on the next line as well is not needed? 

I know the interface declaration isn't necessary, but I've made it explicit to make it easier for the reader, e.g. in https://github.com/GoogleCloudPlatform/gcsfuse/blob/66b8100310ff701a7f4fba5f801615c3480228dc/internal/gcsx/content_type_bucket.go#L32 .

gargnitingoogle:
> Unit tests are not reviewed.

Understood.

internal/storage/streamed_chunk_uploader.go
vadlakondaswetha: The reason for adding inside content_type_bucket struct is to store the reference of original implementation and call it in other methods:

https://github.com/GoogleCloudPlatform/gcsfuse/blob/66b8100310ff701a7f4fba5f801615c3480228dc/internal/gcsx/content_type_bucket.go#L28
https://github.com/GoogleCloudPlatform/gcsfuse/blob/66b8100310ff701a7f4fba5f801615c3480228dc/internal/gcsx/content_type_bucket.go#L44

 I dont think there is any other definition which uploadChunker wants to call.

internal/storage/streamed_chunk_uploader.go
vadlakondaswetha: When does the upload happen of that data happens? when we write the next chunk or call close?

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	AssertEq(nil, err)
+
+	AssertEq(Writing, uploader.state)
+
+	// time.Sleep(1000 * time.Millisecond)
+	if numCallbacks > 0 {
+		AssertGe(numChunks, numCallbacks)
+		AssertGe(contentSize, numBytesUploaded)
+	}
+	AssertEq(numBytesUploaded, uploader.BytesUploadedSoFar())
```
vadlakondaswetha, line 201: Isn't this wrong?  you are trying to compare with values generated from the test progress function instead of absolute values. Basically numBytesUploaded and uploader.BytesUploadedSoFar() are generated from the same function

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	AssertEq(contentSize, uploader.bytesUploadInitiatedSoFar())
+}
+
+func (t *ChunkUploaderTest) TestUploadMultipleMultichunkUploads() {
```
vadlakondaswetha, line 205: Same comments for all test methods 
1. formatting
2. comparision part
3. adding comments.

vadlakondaswetha, line 205: What about testcase for TEstMultipleSingleChunkUploads ?

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+		AssertGe(numChunks, numCallbacks)
+		AssertGe(totalContentSize, numBytesUploaded)
+	}
+	AssertEq(numBytesUploaded, uploader.BytesUploadedSoFar())
+	AssertEq(totalContentSize, uploader.bytesUploadInitiatedSoFar())
+}
+
+func (t *ChunkUploaderTest) TestCloseWithoutWrite() {
+	obj := t.obj
+
```
vadlakondaswetha, line 250: nit: remove new line

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	uploader, err := NewChunkUploader(context.Background(), obj, t.req, 1000, nil)
+	AssertNe(nil, uploader)
+	AssertEq(nil, err)
+
+	o, err := uploader.Close()
```
vadlakondaswetha, line 255: nit: add new line after uploader.Close().

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	AssertNe(nil, o)
+	AssertEq(nil, err)
+	AssertEq(o.Name, "test_object")
+	AssertEq(contentSize, o.Size)
+}
+
+func (t *ChunkUploaderTest) TestCloseWithMultipleSmallUploads() {
```
vadlakondaswetha, line 281: can you define what does small means in the test method itself.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	o, err := uploader.Close()
+	AssertNe(nil, o)
+	AssertEq(nil, err)
+	AssertEq(o.Name, "test_object")
+	AssertEq(contentSize, o.Size)
+}
+
+func (t *ChunkUploaderTest) TestCloseWithMultipleBigUploads() {
```
vadlakondaswetha, line 304: what does big mean? add it in the test method name itself.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	obj := t.obj
+	chunkSize := 4
+	contents := [6]string{"hfuerj3ifj3920",
+		"yrf8934h9or329",
+		"iru328ryo9dj320j3",
+		"iy3rh34r489y89j309",
+		"ie32hr83hr43rt9y8do93j9",
+		"i9r38ry2u9j9orj3"}
+	var contentSize int
```
vadlakondaswetha, line 313: Negative test cases are missing
when empty contents are passed to uploadChunkAsync?
when io.EOF is returned
Can you trigger an error scenario - invalid io.Reader and what happens when you call close?

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	AssertNe(nil, uploader)
+	AssertEq(nil, err)
+
+	for _, content := range contents {
+		err = uploader.UploadChunkAsync(strings.NewReader(content))
+		AssertEq(nil, err)
+		contentSize += len(content)
+	}
+
+	o, err := uploader.Close()
```
vadlakondaswetha, line 325: Please validate the uploader state after close in all test cases

internal/storage/chunk_uploader_test.go
gargnitingoogle: > when empty contents are passed to uploadChunkAsync?

Added test for passing empty content to uploadChunkAsync.

> when io.EOF is returned

io.EOF is never returned, so removed that check from implementation.

> invalid io.Reader and what happens when you call close?

This needs work.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,186 @@
...
+	// Has been finalized.
+	Destroyed
+)
+
+// A chunkUploader is an implementation of ChunkUploader
+// interface, which uses storage.Writer from go-storage-client
+// for resumable upload.
+//
+// It also stores the current state of the uploader.
+type chunkUploader struct {
```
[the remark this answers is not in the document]
gargnitingoogle: Added mutex and locks on all public calls to chunkUplader.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+		obj: proper_obj,
+		req: proper_req,
+		// chunkSize is 0
+	},
+	}
+
+	for _, input := range inputs {
+		uploader, err := NewChunkUploader(input.ctx, input.obj, input.req, input.writeChunkSize, input.progressFunc)
+		AssertEq(nil, uploader)
+		AssertNe(nil, err)
```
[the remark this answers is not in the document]
gargnitingoogle: Added check on err.Error()

internal/storage/chunk_uploader_test.go
gargnitingoogle: Done

internal/storage/chunk_uploader_test.go
gargnitingoogle: Done

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Ok. Removed.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: It happens during the next upload call(s) whenever a full chunk is available.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,186 @@
...
+	totalWriteSucceededSoFar int64
+	userProgressFunc         func(int64)
+
+	// Internal state for lifecycle management.
+	state UploaderState
+}
+
+// NewChunkUploader creates a new instace of StreamedChunkUploader,
+// for the given inputs.
+func NewChunkUploader(ctx context.Context, obj *storage.ObjectHandle, req *gcs.CreateObjectRequest, writeChunkSize int, progressFunc func(int64)) (*chunkUploader, error) {
```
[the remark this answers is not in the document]
gargnitingoogle: Done.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,186 @@
...
+		return 0, fmt.Errorf("writer not initialized")
+	}
+	return uploader.wc.ChunkSize, nil
+}
+
+func (uploader chunkUploader) bytesUploadInitiatedSoFar() int64 {
+	return uploader.totalWriteInitiatedSoFar
+}
+
+func (sow chunkUploader) BytesUploadedSoFar() int64 {
```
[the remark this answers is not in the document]
gargnitingoogle: Done.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,186 @@
...
+	// Attributes for providing updates to user.
+	totalWriteInitiatedSoFar int64
+	totalWriteSucceededSoFar int64
+	userProgressFunc         func(int64)
+
+	// Internal state for lifecycle management.
+	state UploaderState
+}
+
+// NewChunkUploader creates a new instace of StreamedChunkUploader,
```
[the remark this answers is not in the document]
gargnitingoogle: Done.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+var _ SetUpInterface = &ChunkUploaderTest{}
+var _ TearDownInterface = &ChunkUploaderTest{}
+
+func init() { RegisterTestSuite(&ChunkUploaderTest{}) }
+
+func (t *ChunkUploaderTest) SetUp(ti *TestInfo) {
+	t.fakeStorage = NewFakeStorage()
+	t.storageHandle = t.fakeStorage.CreateStorageHandle()
+	t.bucketHandle = t.storageHandle.BucketHandle(TestBucketName, "")
```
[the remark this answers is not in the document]
gargnitingoogle: Removed both storageHandle and bucketHandle from struct ChunkUploaderTest .

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	AssertNe(nil, uploader)
+	AssertEq(nil, err)
+
+	err = uploader.UploadChunkAsync(strings.NewReader(content))
+	AssertEq(nil, err)
+
+	AssertEq(Writing, uploader.state)
+	AssertEq(contentSize, uploader.bytesUploadInitiatedSoFar())
+
+	// time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
gargnitingoogle: No, removed. I've synchronized based on a channel now.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	if numCallbacks > 0 {
+		AssertGt(1, numCallbacks)
+		AssertGt(contentSize, numBytesSuccessfullyUploaded)
+	}
+	AssertEq(numBytesSuccessfullyUploaded, uploader.BytesUploadedSoFar())
```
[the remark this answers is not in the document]
gargnitingoogle: Uploaded the code now. I've synchronized based on a channel now.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	AssertNe(nil, uploader)
+	AssertEq(nil, err)
+
+	err = uploader.UploadChunkAsync(strings.NewReader(content))
+	AssertEq(nil, err)
+
+	AssertEq(Writing, uploader.state)
+
+	// time.Sleep(1000 * time.Millisecond)
+	if numCallbacks > 0 {
```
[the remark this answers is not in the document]
gargnitingoogle: Uploaded the code now. I've synchronized based on a channel now.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+	AssertNe(nil, uploader)
+	AssertEq(nil, err)
+
+	err = uploader.UploadChunkAsync(strings.NewReader(content))
+	AssertEq(nil, err)
+
+	AssertEq(Writing, uploader.state)
+
+	// time.Sleep(1000 * time.Millisecond)
```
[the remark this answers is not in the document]
gargnitingoogle: Removed.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+	err = uploader.UploadChunkAsync(strings.NewReader(content))
+	AssertEq(nil, err)
+
+	AssertEq(Writing, uploader.state)
+	AssertEq(contentSize, uploader.bytesUploadInitiatedSoFar())
+
+	// time.Sleep(100 * time.Millisecond)
+	if numCallbacks > 0 {
+		AssertGt(1, numCallbacks)
```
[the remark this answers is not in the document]
gargnitingoogle: Fixed.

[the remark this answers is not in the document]
gargnitingoogle: Fixed.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,330 @@
...
+		AssertEq(0, uploader.bytesUploadInitiatedSoFar())
+		AssertEq(0, uploader.BytesUploadedSoFar())
+		AssertEq(t.req.Name, uploader.objectName)
+		chunkSize, err := uploader.chunkSize()
+		AssertEq(nil, err)
+		AssertEq(input.writeChunkSize, chunkSize)
+	}
+}
+
+func (t *ChunkUploaderTest) TestUploadSingleChunkSingleUpload() {
```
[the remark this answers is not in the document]
gargnitingoogle: Fixed.

internal/storage/chunk_uploader_test.go
gargnitingoogle: Fixed.

internal/storage/chunk_uploader_test.go
gargnitingoogle: Done.

gargnitingoogle: I've added a single generic test for multiple uploads of mixed sizes, which covers all cases.

internal/storage/chunk_uploader_test.go
gargnitingoogle: Small stands for where content-size < chunkSize.

internal/storage/chunk_uploader_test.go
gargnitingoogle: Big stands for where content-size > chunkSize.

internal/storage/streamed_chunk_uploader.go
gargnitingoogle: Yeah, I was wrong earlier, io.EOF is never returned from io.Copy in successful case unless size is specified. Since we're specifying size here, we can't get io.EOF. So, I've removed any special handling for io.EOF.

As for n==0, I've treated that a NOP now.

internal/storage/chunk_uploader_test.go
gargnitingoogle: Added test for failed uploader and another test for showing its sucessful closure.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,592 @@
...
+		numFullChunksByUploadsCumulative =
+			append(numFullChunksByUploadsCumulative, int(math.Floor(float64(contentSizesForUploadsCumulative[i])/float64(chunkSize))))
+	}
+	progressFuncCallbackTimeout := time.Millisecond
+	var numCallbacks int
+	numBytesSuccessfullyUploadedSoFar := make(chan int64, numFullChunksByUploadsCumulative[numUploads-1])
+	ctx := context.Background()
+
+	// cleanup
+	defer func() {
```
gargnitingoogle, line 392: No need for func() here.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,592 @@
...
+		func(n int64) {
+			numCallbacks++
+			numBytesSuccessfullyUploadedSoFar <- n
+		})
+
+	// test
+	AssertNe(nil, uploader)
+	AssertEq(nil, err)
+
+	// test payload
```
gargnitingoogle, line 407: Move this comment down.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// NewChunkUploader creates a new instance of chunkUploader,
+// for the given inputs.
+func NewChunkUploader(ctx context.Context, obj *storage.ObjectHandle, req *gcs.CreateObjectRequest, writeChunkSize int, progressFunc func(int64)) (gcs.ChunkUploader, error) {
+	if ctx == nil {
+		return nil, fmt.Errorf("ctx is nil")
+	}
+	if obj == nil || req == nil {
+		return nil, fmt.Errorf("nil obj or req")
```
ashmeenkaur, line 60: NIT: nil ObjectHandle or CreateObjectRequest

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	if obj.ObjectName() != req.Name {
+		return nil, fmt.Errorf("names of passed object-handle and createObjectRequest.Name don't match: object-handle=%s createObjectRequest.Name=%s", obj.ObjectName(), req.Name)
```
ashmeenkaur, line 64: let's use camelcase for ObjectHandle as well

ashmeenkaur (commented):
added some comments. Need to review unit tests as well.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,588 @@
+package storage
```
ashmeenkaur, line 1: there are a lot of common assertions in this test file, can we move them to a common helper function?

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
+package storage
```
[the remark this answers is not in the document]
gargnitingoogle: Yes, good idea.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
...
+func (uploader *chunkUploader) BytesUploadedSoFar() int64 {
+	uploader.mu.Lock()
+	defer uploader.mu.Unlock()
+
+	return uploader.totalWriteSucceededSoFar
+}
+
+func (uploader *chunkUploader) readyToWrite() bool {
+	switch uploader.state {
+	case Initialized, Waiting, Writing:
```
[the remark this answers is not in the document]
gargnitingoogle: @ashmeenkaur  Yes, the default state of a non-nil chunkUploader will be Initialized. I didn't understand what you meant by 'shall we change it' ?

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
...
+	uploader.objectName = obj.ObjectName()
+	uploader.userProgressFunc = progressFunc
+
+	// Create a NewWriter with the requested attributes, using Go Storage Client.
+	// NewWriter never returns nil, so no nil-check is needed on it.
+	wc := obj.NewWriter(ctx)
+	wc = storageutil.SetAttrsInWriter(wc, req)
+	wc.ChunkSize = writeChunkSize
+	wc.ProgressFunc = func(n int64) {
+		uploader.totalWriteSucceededSoFar = n
```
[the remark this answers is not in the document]
gargnitingoogle: This callback doesn't indicate if there is any more data upload in progress or not, so not a good idea to set it here. Plus changing it here will require a mutex lock which can slow things down unnecessary, or even a deadlock.

internal/storage/chunk_uploader.go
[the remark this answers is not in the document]
gargnitingoogle: Done.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
...
+	default:
+		return false
+	}
+}
+
+// UploadChunkAsync uploads the passed content to GCS.
+// This waits (using mutex) until the completion of
+// in-progress calls to BytesUploadedSoFar/UploadChunkAsync/Close
+// invoked from other threads/go-routines.
+func (uploader *chunkUploader) UploadChunkAsync(contents io.Reader) error {
```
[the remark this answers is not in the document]
gargnitingoogle: > This method has no logic related to async upload

The word async in function name indicates that this function doesn't block on the upload to complete. This is a common convention.

> I am a little concerned about retries and timeouts for this call since it doesn't take context. Context aware reader is one option but I was wondering if go client has something related?

Let's take this separately in another PR. This needs discussion and possibly interface change.

internal/storage/chunk_uploader_test.go
gargnitingoogle: Clubbed all the common assertions around the new/upload/close calls into helper functions. 

I am not sure it's good idea to try to optimize/share the test logic easily without making it harder to change them later if needed.

internal/storage/chunk_uploader.go
gargnitingoogle: Done.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
...
+	uploader.mu.Lock()
+	defer uploader.mu.Unlock()
+
+	if !uploader.readyToWrite() || (uploader.wc == nil) {
```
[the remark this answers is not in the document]
gargnitingoogle: This checke was redundant. Removed now.

internal/storage/chunk_uploader.go
gargnitingoogle: Done.

internal/storage/chunk_uploader.go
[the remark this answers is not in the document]
gargnitingoogle: I have now removed the `Waiting` state altogether as it didn't make much sense after more thought. So, this is now a moot point.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+	}
+
+	uploader.wc = wc
+	uploader.state = Initialized
+
+	return &uploader, nil
+}
+
+func (uploader *chunkUploader) chunkSize() (int, error) {
```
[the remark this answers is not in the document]
gargnitingoogle: Was redundant. Removed.

gargnitingoogle:
@ashmeenkaur I addressed your comments. There are follow-ups on a few of them, PTAL.

internal/storage/chunk_uploader.go
ashmeenkaur: there is one more object-handle

internal/storage/chunk_uploader.go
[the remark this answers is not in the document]
ashmeenkaur: I was concerned if someone can run into a case where they create a zero value instance of chunk with Initialized state but upon thinking more, I believe we can never run into that case considering the struct is private and the only way to create an instance is via NewChunkUploader.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,214 @@
...
+	// Uploading is the state when an asynchronous upload is in progress.
+	Uploading
+	// UploadError is the state when an Upload failed.
+	UploadError
+	// Closed is the state of an uploader which has been finalized.
+	Closed
+	// StorageCopyTimeoutDuration is how long the chunkUploader waits for the
+	// copy call to the uploader to return. If it takes longer, the
+	// upload call times out.
+	StorageCopyTimeoutDuration time.Duration = time.Second
```
ashmeenkaur, line 34: As discussed offline, let's remove the timeout logic from asyncChunkUpload and this const

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,214 @@
...
+	return &uploader, nil
+}
+
+// BytesUploadedSoFar returns the total number of bytes successfully uploaded
+// so far using this uploader.
+// This is thread-safe against the
+// in-progress calls to BytesUploadedSoFar/UploadChunkAsync/Close
+// invoked from other threads/go-routines.
+func (uploader *chunkUploader) BytesUploadedSoFar() int64 {
+	return atomic.LoadInt64(&uploader.totalWriteSucceededSoFar)
```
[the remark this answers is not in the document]
gargnitingoogle: `atomic.LoadInt64` is basically for thread-safety. It's needed because this variable is updated in the callback (different goroutine) and is read in the goroutine (calling the `uploader.Close()` or `uploader.Upload...`).

I didn't know that there was a clearner alternative to it. I'll switch to `atomic.Int64.Load()` .

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,214 @@
...
+)
+
+// A chunkUploader is an implementation of ChunkUploader
+// interface, which uses storage.Writer from go-storage-client
+// for resumable upload.
+//
+// It also stores the current state of the uploader.
+type chunkUploader struct {
+	// Internal objects for business logic.
+	wc         *storage.Writer
```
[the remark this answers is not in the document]
gargnitingoogle: No specific reason, the seed code for this struct was taken by `bucket_handle.CreateObject()`, and also the names. I'll change it.

internal/storage/chunk_uploader.go
gargnitingoogle: ok

internal/storage/chunk_uploader.go
gargnitingoogle: Removed.

internal/storage/chunk_uploader.go
gargnitingoogle: Done.

internal/storage/chunk_uploader_test.go
gargnitingoogle: @ashmeenkaur reminder.

ashmeenkaur: Thank you!

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,463 @@
...
+	AssertEq(nil, err)
+	AssertNe(nil, o)
+	AssertEq(Closed, uploader.state)
+	AssertEq(objName, o.Name)
+	AssertEq(len, o.Size)
+}
+
+func (t *ChunkUploaderTest) testFailedClose(uploader *chunkUploader) {
+	_, err := uploader.Close()
+	AssertNe(nil, err)
```
ashmeenkaur, line 113: Should we assert on the error string as well?

ashmeenkaur (commented):
Minor comments, LGTM otherwise.
Also noticed some error statements going beyond 80 characters and comments < 80 characters. Can we make it consistent?

internal/storage/chunk_uploader_test.go
gargnitingoogle: Added assert on error substring.

gargnitingoogle (commented):
Addressed comments.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,463 @@
...
+		select {
+		case n := <-numBytesSuccessfullyUploadedSoFar:
+			AssertEq(chunkCallbackIndex*chunkSize, n)
+		case <-time.After(progressFuncCallbackTimeout):
+			AddFailure("Did not not receive %d-th write progressFunc callback in %v", chunkCallbackIndex, progressFuncCallbackTimeout)
+		}
+	}
+}
+
+func (t *ChunkUploaderTest) TestUploadMultipleUploads() {
```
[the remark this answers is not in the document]
gargnitingoogle: Done. Added a comment explaining the flow, and also added a few functions to simplify the code flow.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,463 @@
...
+			select {
+			case n := <-numBytesSuccessfullyUploadedSoFar:
+				AssertEq(chunkCallbackIndex*chunkSize, n)
+			case <-time.After(progressFuncCallbackTimeout):
+				AddFailure("Did not not receive %d-th write progressFunc callback in %v", chunkCallbackIndex, progressFuncCallbackTimeout)
+			}
+		}
+	}
+}
+
```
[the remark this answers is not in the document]
gargnitingoogle: Also added another test TestMultipleHomogeneousUploads .

gargnitingoogle:
@ashmeenkaur all comments addressed. PTAL.

internal/storage/chunk_uploader_test.go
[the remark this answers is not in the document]
ashmeenkaur: I see the timeout is 100ms. Is this intended?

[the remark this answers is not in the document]
gargnitingoogle: Yes, I increased the rest of timeouts to 10ms to avoid some flakiness with 1ms. So, increased this one to 100ms.

ashmeenkaur (approved):

gargnitingoogle:
@sethiay @raj-prince please review and approve.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,193 @@
...
+// It also stores the current state of the uploader.
+type chunkUploader struct {
+	// Internal objects for business logic.
+	writer     *storage.Writer
+	objectName string
+	mu         sync.Mutex
+
+	// Attributes for providing updates to user.
+	totalWriteInitiatedSoFar int64
+	// Total number of bytes successfully written by this writer so far.
```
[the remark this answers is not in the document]
gargnitingoogle: Yes. Changed all places.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,193 @@
...
+	default:
+		return false
+	}
+}
+
+// UploadChunkAsync uploads the passed content to GCS.
+// This waits (using mutex) until the completion of
+// in-progress calls to BytesUploadedSoFar/UploadChunkAsync/Close
+// invoked from other threads/go-routines.
+// The call times out if it is taking longer than StorageCopyTimeoutDuration.
```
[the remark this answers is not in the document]
gargnitingoogle: Stale comment line.. Removed.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,193 @@
...
+func (uploader *chunkUploader) UploadChunkAsync(contents io.Reader) error {
```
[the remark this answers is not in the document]
gargnitingoogle: No, the timeout will be in the caller.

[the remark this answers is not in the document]
gargnitingoogle: The function is asynchronous as it queues up the GCS uploads and returns right after queueing up the last chunk provided in the content, rather than blocking on that chunk-upload finishing. When the uploaded content is a single chunk (which is the main use case), then it just queues up the gcs upload and returns. There will not be much value to calling this in a goroutine if we stick to single-chunk uploads.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,193 @@
...
+	uploader.mu.Lock()
+	defer uploader.mu.Unlock()
+
+	if !uploader.readyToUpload() {
+		return fmt.Errorf("writer not ready to write: object: %s, status = %v, writer: %v", uploader.objectName, uploader.state, uploader.writer)
+	}
+
+	n, err := io.Copy(uploader.writer, contents)
```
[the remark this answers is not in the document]
gargnitingoogle: No, as per the documentation of io.Copy, io.EOF is not considered an error and is just retuned as `err: nil`.

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,497 @@
...
+		uploader, err := NewChunkUploader(input.ctx, input.obj, input.req,
+			input.chunkSize, input.progressFunc)
+
+		AssertEq(nil, uploader)
+		AssertNe(nil, err)
+		AssertTrue(strings.Contains(err.Error(), input.errStr))
+	}
+}
+
+func (t *ChunkUploaderTest) CreateWithProperInputs() {
```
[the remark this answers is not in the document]
gargnitingoogle: Yes. It's a test. Renamed.

internal/storage/gcs/bucket.go
```diff
@@ -31,9 +31,6 @@ type ChunkUploader interface {
 	// Progress should be tracked using the progress func
 	// passed during ChunkUploader instance creation.
 	//
-	// io.EOF is an expected error in case the reader has less
```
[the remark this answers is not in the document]
gargnitingoogle: @sethiay  You mean what the comment says isn't true for GCS? If so, I've removed that comment.

internal/storage/chunk_uploader.go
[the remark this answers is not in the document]
sethiay: When the buffer size is greater than 50 MiB, chunk uploader will get `contents` which has more than 1 chunk size of data. So in that case, this will be a blocking call for caller. Should we do something to handle this ?

internal/storage/chunk_uploader.go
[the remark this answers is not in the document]
sethiay: let's please change the code comments also.

internal/storage/gcs/bucket.go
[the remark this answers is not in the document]
sethiay: Sorry, I meant "removing the comment" isn't true for real GCS (or the comment which was there is correct for real GCS).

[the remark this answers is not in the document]
gargnitingoogle: Resolving as discussed offline as this is a stale comment removed.

internal/storage/chunk_uploader.go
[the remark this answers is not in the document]
gargnitingoogle: @sethiay As discussed offline, if the uploader has to be used for uploading multiple chunks, then this function should be called in a goroutine on the caller side. 
I'll add the context for timing out the call in a goroutine as you suggested though.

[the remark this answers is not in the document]
gargnitingoogle: Done.

gargnitingoogle:
@sethiay addressed all your comments.

Changes:
1. Passed context into chunkUploader.{Upload(), Close()} for allowing them to be cancelled by the caller. Added additional unit tests for them.
2. Renamed ChunkUploader.UploadChunkAsync() to Upload to generalize it.

cc @ashmeenkaur PTAL.

internal/storage/chunk_uploader.go
```diff
@@ -0,0 +1,233 @@
...
+	var err error
+	ioCopyComplete := make(chan bool)
+	go func() {
+		n, err = io.Copy(uploader.writer, contents)
+		ioCopyComplete <- true
+	}()
+
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("uploader.Upload() timed out")
```
sethiay, line 159: nit: whether it's timed out or cancelled because of other reasons, we don't know. So, can we use ctx.Err() here instead of hard coding timed out?

internal/storage/chunk_uploader_test.go
```diff
@@ -0,0 +1,523 @@
...
+	obj := t.obj
+	chunkSize := googleapi.MinUploadChunkSize
+	contentSize := 10 * chunkSize
+	content := genRandomContent(contentSize)
+
+	uploader := t.testNewChunkUploader(context.Background(), obj, t.req, chunkSize, nil)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+	defer cancel()
+	err := uploader.Upload(ctx, strings.NewReader(content))
+	AssertNe(nil, err)
```
sethiay, line 336: please also assert with error message.

sethiay (approved):

internal/storage/chunk_uploader.go
gargnitingoogle: Done.

internal/storage/chunk_uploader_test.go
gargnitingoogle: Done.

sethiay (approved):

PR #1423, opened 2023-10-05
Weighted LRU Cache

raj-prince:
### Description
1. This PR contains changes over [this](https://github.com/jacobsa/util/tree/master/lrucache) package
2. Added support for weighted entry, which can be used as normal lru cache by returning ValueType.Size() 1 every time.
3. Added support of mutex so that only one method of lrucache instance will run at a time.
4. Added unit test.

NA

### Testing details
1. Manual - Validated normal operation.
6. Unit tests - Automation
7. Integration tests - NA.


internal/cache/lrucache/lrucache_test.go
```diff
@@ -0,0 +1,248 @@
...
+	defer c.Wrapped.CheckInvariants()
+
+	return c.Wrapped.LookUp(key)
+}
+
+////////////////////////////////////////////////////////////////////////
+// Boilerplate
+////////////////////////////////////////////////////////////////////////
+
+const maxWeight = 50
```
sethiay: let's keep this maxSize to be consistent with the maxSize in Cache.

internal/cache/lrucache/lrucache.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// Return the value.
+	return e.Value.(entry).Value
+}
+
+////////////////////////////////////////////////////////////////////////
+// Gob encoding
+////////////////////////////////////////////////////////////////////////
+
+func (c *Cache) GobEncode() (b []byte, err error) {
```
sethiay, line 218: Can we remove this to reduce code if not required ? We can include whenever there is a requirement.

go.mod
```diff
@@ -23,8 +23,12 @@ require (
 	go.opencensus.io v0.24.0
 	golang.org/x/net v0.10.0
 	golang.org/x/oauth2 v0.8.0
+	golang.org/x/sync v0.2.0
+	golang.org/x/sys v0.8.0
+	golang.org/x/text v0.9.0
 	golang.org/x/time v0.3.0
 	google.golang.org/api v0.126.0
+	gopkg.in/yaml.v2 v2.4.0
```
sethiay, line 31: Could you please help me understand why we are moving this from indirect to direct ?

sethiay (commented):
Given this package will be used at multiple places, can we check we can write tests to check concurrency of exposed public methods ?

internal/cache/lrucache/lrucache.go
```diff
@@ -0,0 +1,294 @@
...
+	return evictedValues
+}
+
+// Erase any entry for the supplied key, also returns the value of erased key.
+// LOCK_EXCLUDED(c.mu)
+func (c *Cache) Erase(key string) (value ValueType) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	e := c.index[key]
```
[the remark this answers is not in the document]
raj-prince: Both are good, but making in consistent. Using two value assignment at every places.

internal/cache/lrucache/lrucache.go
```diff
@@ -0,0 +1,294 @@
...
+	// INVARIANT: Each element is of type entry
+	entries list.List
+
+	// Index of elements by name.
+	//
+	// INVARIANT: For each k, v: v.Value.(entry).Key == k
+	// INVARIANT: Contains all and only the elements of entries
+	index map[string]*list.Element
+
+	// All exposed methods are guarded by this Mutex. This means that only one
```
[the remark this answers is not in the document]
raj-prince: Done, with this change (updating -> accessing/updating)

internal/cache/lrucache/lrucache.go
```diff
@@ -0,0 +1,294 @@
...
+		// Add the entry if already doesn't exist.
+		e := c.entries.PushFront(entry{key, value})
+		c.index[key] = e
+		c.currentSize += value.Size()
+	}
+
+	var evictedValues []ValueType
+	// Evict until we're at or below maxSize.
+	for c.currentSize > c.maxSize {
+		evictedValues = append(evictedValues, c.evictOne())
```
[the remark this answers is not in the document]
raj-prince: As discussed offline - Insertion in such cases will return a pre-defined error, which clients need to handle.

go.mod
raj-prince: We are using directly [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/66b8100310ff701a7f4fba5f801615c3480228dc/tools/integration_tests/util/setup/yaml-config.go#L9).

internal/cache/lrucache/lrucache.go
```diff
@@ -0,0 +1,294 @@
...
+// all values are registered using gob.Register.
+//
+// May be used directly as a field in a larger struct. Must be created with New
+// or initialized using gob decoding.
+type Cache struct {
+	/////////////////////////
+	// Constant data
+	/////////////////////////
+
+	// INVARIANT: maxSize > 0
```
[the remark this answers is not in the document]
raj-prince: As discussed offline, will let the client handle to create an infinite size cache. Client needs to explicitly pass MAX_uint64 to create infinite size cache.

raj-prince:
> Given this package will be used at multiple places, can we check we can write tests to check concurrency of exposed public methods ?



> Given this package will be used at multiple places, can we check we can write tests to check concurrency of exposed public methods ?

Went through with posts - https://stackoverflow.com/questions/19189542/how-to-test-concurrency-and-locking-in-golang, which mentions this is hard to test. This should be covered mostly by design and code review.

For now, I have added test to catch  race condition, but to automate this. Either we need to run our ci linux-test to run with `-race` flag. 

Running all the test with `race` flag is also not good. So, we need to mark individually test to not run with `race` flag. This required changes in all the package-test. 

For now I test with -race tag it is running successfully. I tried by commenting out the lock, it fails.

```
==================
[       OK ] CacheTest.TestRaceCondition
[----------] Finished with tests from CacheTest
    testing.go:1465: race detected during execution of test
--- FAIL: TestCache (0.00s)

```

Please let me know, what's your view?

internal/cache/lru/lrucache_test.go
```diff
@@ -0,0 +1,239 @@
...
+	DataSize uint64
+}
+
+func (td testData) Size() uint64 {
+	return td.DataSize
+}
+
+// insertAndAssert inserts the given key,value in the cache and assert based on
+// the expected eviction and error.
+func (t *CacheTest) insertAndAssert(key string, val lru.ValueType, expectedEviction int, evictedValues []int64, expectedError error) {
```
sethiay, line 84: can we skip using expectedEviction as length can be determined using evictedValues?

internal/cache/lru/lrucache_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+// This will detect race if we run the test with `-race` flag.
+// We get the race condition failure if we remove lock from Insert or Erase method.
+func (t *CacheTest) TestRaceCondition() {
+	var wg sync.WaitGroup
+	wg.Add(3)
+
+	go func() {
+		defer wg.Done()
+		for i := 0; i < MaxSize; i++ {
```
sethiay, line 214: Is there any reason why we running till MaxSize ? I believe we just want to run for multiple iterations, is that's true then we should use different name like numIterations, count.

sethiay (commented):
Left few minor comments

internal/cache/lru/lrucache.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+type entry struct {
+	Key   string
+	Value ValueType
+}
+
+// New initialize a cache with the supplied maxSize, which must be greater than
+// zero.
+func New(maxSize uint64) (c Cache) {
```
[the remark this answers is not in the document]
raj-prince: Done.

sethiay (approved):

PR #1427, opened 2023-10-09
Write to buffer for minimal file buffering flow

ashmeenkaur:
### Description
Write to buffer when file buffering flow is enabled. 

NA

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -15,18 +15,19 @@
 package buffer
 
 import (
-	"bytes"
 	"fmt"
 )
 
 type InMemoryWriteBuffer struct {
 	// Holds the data to be written to GCS. At any time, Buffer holds the data for
 	// 2 GCS write calls.
-	buffer *bytes.Buffer
-	// [ currStartOffset, currEndOffset] is the range of file offset for which the
+	buffer []byte
+	// [ bufferStartOffset, bufferEndOffset] is the range of file offset for which the
 	// data is currently stored in buffer.
-	currStartOffset int64
-	currEndOffset   int64
+	bufferStartOffset int64
+	bufferEndOffset   int64
+
+	fileSize int64
 }
 
 // CreateInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
@@ -43,68 +44,63 @@ func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
 func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
 	ChunkSize = int64(sizeInMB * MiB)
 	BufferSize = 2 * ChunkSize
-	b.currEndOffset = ChunkSize
+	b.bufferEndOffset = ChunkSize
 	if b.buffer == nil {
-		b.buffer = bytes.NewBuffer(make([]byte, 0, BufferSize))
+		b.buffer = make([]byte, 0, BufferSize)
 	}
 }
 
-func (b *InMemoryWriteBuffer) WriteAt(data []byte, startOffset int64) error {
-	unreadBuffer := b.buffer.Bytes()
-	size := int64(len(data))
-	endOffset := startOffset + size
+func (b *InMemoryWriteBuffer) WriteAt(data []byte, dataStartOffset int64) error {
+	dataSize := int64(len(data))
+	dataEndOffset := dataStartOffset + dataSize
 
-	// If startOffset is not in the range of current offset in buffer, trigger
-	// temp file flow for writes.
-	if startOffset < b.currStartOffset {
-		// todo: finalise the write and trigger temp file flow.
-		return nil
+	// If dataStartOffset is not in the range of offsets stored in buffer block,
+	// trigger temp file flow for writes.
+	if dataStartOffset < b.bufferStartOffset || b.bufferEndOffset < dataStartOffset {
+		// TODO: finalise write and trigger temp file flow.
+		return fmt.Errorf(NonSequentialWriteError)
 	}
 
-	// get slice of unread buffer that we are currently writing to.
-	bufferStartOffset := b.currStartOffset % BufferSize
-	bufferEndOffset := b.currEndOffset % BufferSize
-	unreadBuffer = unreadBuffer[bufferStartOffset:bufferEndOffset]
-
-	// if data can be completely written without wrapping to previously written
-	// buffer offsets (beyond the current block of buffer), write it.
-	if endOffset <= b.currEndOffset {
-		fmt.Println("startOffset = ", startOffset)
-		fmt.Println("startOffset%BufferSize = ", startOffset%BufferSize)
-		bufferEndOffset = 20
-		fmt.Println("bufferEndOffset = ", bufferEndOffset)
-		n := copy(unreadBuffer[startOffset%BufferSize:bufferEndOffset], data)
-		if int64(n) != size {
-			return fmt.Errorf("could not write all the data to buffer. "+
-				"Expected bytes to be written: %d, actually written: %d", size, n)
-		}
-		fmt.Println("unreadBuffer = ", unreadBuffer[startOffset%BufferSize:bufferEndOffset])
-		fmt.Println("data = ", data)
+	// If data can be completely written without wrapping to previously written
+	// buffer block (beyond the current block of buffer), write it.
+	if dataEndOffset <= b.bufferEndOffset {
+		return b.copyDataToBuffer(dataStartOffset, dataEndOffset, data)
 	} else {
-		// Else write the chunk of data that can be written, check status of last
-		// write to GCS, trigger write call for current buffer block, and write
-		// remaining data by wrapping.
-		l := b.currEndOffset - startOffset
-		n := copy(unreadBuffer[startOffset%BufferSize:bufferEndOffset], data[:l])
-		if int64(n) != l {
-			return fmt.Errorf("could not write all the data to buffer. "+
-				"Expected bytes to be written: %d, actually written: %d", l, n)
+		// Below logic is written with the assumption that the data received in a
+		// single write call from the kernel will never exceed beyond buffer block size.
+
+		// Write the chunk of data that can be written to the current buffer block.
+		l := b.bufferEndOffset - dataStartOffset
+		if err := b.copyDataToBuffer(dataStartOffset, b.bufferEndOffset, data[:l]); err != nil {
+			return err
 		}
-		//TODO: logic to get status of last write and trigger async write.
 
-		// Update buffer offsets.
-		b.currStartOffset = b.currEndOffset + 1
-		b.currEndOffset = b.currEndOffset + ChunkSize
-		bufferStartOffset = b.currStartOffset % BufferSize
-		bufferEndOffset = b.currEndOffset % BufferSize
+		//TODO: get status of last write to GCS, and trigger async write for current block.
 
+		// Update buffer offsets for next buffer block.
+		b.bufferStartOffset = b.bufferEndOffset
+		b.bufferEndOffset += ChunkSize
 		// Write the remaining data by wrapping over to the next buffer block.
-		l = size - l
-		n = copy(unreadBuffer[(startOffset+l)%BufferSize:bufferEndOffset], data[:l])
-		if int64(n) != l {
-			return fmt.Errorf("could not write all the data to buffer. "+
-				"Expected bytes to be written: %d, actually written: %d", l, n)
-		}
+		return b.copyDataToBuffer(b.bufferStartOffset, dataEndOffset, data[l:])
 	}
+}
+
+func (b *InMemoryWriteBuffer) copyDataToBuffer(dataStartOffset, dataEndOffset int64, data []byte) error {
+	dataSize := int64(len(data))
+	si := dataStartOffset % BufferSize
+	ei := si + dataSize
+
+	n := copy(b.buffer[si:ei], data)
+	if int64(n) != dataSize {
+		return fmt.Errorf(DataNOtWrittenCompletelyError, dataSize, n)
+	}
+	b.updateFileSize(dataEndOffset)
 	return nil
+}
+
+// After successful copy of data to buffer, increments the bytes written so far.
+func (b *InMemoryWriteBuffer) updateFileSize(endOffset int64) {
+	if endOffset > b.fileSize {
+		b.fileSize = endOffset
+	}
 }
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -43,14 +43,14 @@ type InMemoryWriteBuffer struct {
 // set to nil. Memory is allocated to the buffer when the first write call comes.
 // This avoids unnecessarily bloating GCSFuse memory consumption.
 //
-// To allocate memory to the buffer, use WriteBuffer.InitializeBuffer.
+// To allocate memory to the buffer, use WriteBuffer.Initialize.
 func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
 	b := &InMemoryWriteBuffer{}
 	// TODO: set mtime attribute.
 	return b
 }
 
-func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
+func (b *InMemoryWriteBuffer) Initialize(sizeInMB int) {
 	if ChunkSize == 0 {
 		ChunkSize = int64(sizeInMB * MiB)
 	}
@@ -110,7 +110,7 @@ func (b *InMemoryWriteBuffer) copyDataToBuffer(contentStartOffset, contentEndOff
 
 	n := copy(b.currentBuffer[si:ei], content)
 	if int64(n) != dataSize {
-		return fmt.Errorf(DataNOtWrittenCompletelyError, dataSize, n)
+		return fmt.Errorf(DataNotWrittenCompletelyError, dataSize, n)
 	}
 	b.updateFileSize(contentEndOffset)
 	return nil
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -16,12 +16,17 @@ package buffer
 
 import (
 	"bytes"
+	"fmt"
 )
 
 type InMemoryWriteBuffer struct {
 	// Holds the data to be written to GCS. At any time, Buffer holds the data for
 	// 2 GCS write calls.
 	buffer *bytes.Buffer
+	// [ currStartOffset, currEndOffset] is the range of file offset for which the
+	// data is currently stored in buffer.
+	currStartOffset int64
+	currEndOffset   int64
 }
 
 // CreateInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
@@ -36,12 +41,70 @@ func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
 }
 
 func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
-	ChunkSize = sizeInMB * MiB
+	ChunkSize = int64(sizeInMB * MiB)
+	BufferSize = 2 * ChunkSize
+	b.currEndOffset = ChunkSize
 	if b.buffer == nil {
-		b.buffer = bytes.NewBuffer(make([]byte, 0, 2*sizeInMB*MiB))
+		b.buffer = bytes.NewBuffer(make([]byte, 0, BufferSize))
 	}
 }
 
-func (b *InMemoryWriteBuffer) WriteAt(data []byte, offset int64) error {
+func (b *InMemoryWriteBuffer) WriteAt(data []byte, startOffset int64) error {
+	unreadBuffer := b.buffer.Bytes()
+	size := int64(len(data))
+	endOffset := startOffset + size
+
+	// If startOffset is not in the range of current offset in buffer, trigger
+	// temp file flow for writes.
+	if startOffset < b.currStartOffset {
+		// todo: finalise the write and trigger temp file flow.
+		return nil
+	}
+
+	// get slice of unread buffer that we are currently writing to.
+	bufferStartOffset := b.currStartOffset % BufferSize
+	bufferEndOffset := b.currEndOffset % BufferSize
+	unreadBuffer = unreadBuffer[bufferStartOffset:bufferEndOffset]
+
+	// if data can be completely written without wrapping to previously written
+	// buffer offsets (beyond the current block of buffer), write it.
+	if endOffset <= b.currEndOffset {
+		fmt.Println("startOffset = ", startOffset)
+		fmt.Println("startOffset%BufferSize = ", startOffset%BufferSize)
+		bufferEndOffset = 20
+		fmt.Println("bufferEndOffset = ", bufferEndOffset)
+		n := copy(unreadBuffer[startOffset%BufferSize:bufferEndOffset], data)
+		if int64(n) != size {
+			return fmt.Errorf("could not write all the data to buffer. "+
+				"Expected bytes to be written: %d, actually written: %d", size, n)
+		}
+		fmt.Println("unreadBuffer = ", unreadBuffer[startOffset%BufferSize:bufferEndOffset])
+		fmt.Println("data = ", data)
+	} else {
+		// Else write the chunk of data that can be written, check status of last
+		// write to GCS, trigger write call for current buffer block, and write
+		// remaining data by wrapping.
+		l := b.currEndOffset - startOffset
+		n := copy(unreadBuffer[startOffset%BufferSize:bufferEndOffset], data[:l])
+		if int64(n) != l {
+			return fmt.Errorf("could not write all the data to buffer. "+
+				"Expected bytes to be written: %d, actually written: %d", l, n)
+		}
+		//TODO: logic to get status of last write and trigger async write.
+
+		// Update buffer offsets.
+		b.currStartOffset = b.currEndOffset + 1
+		b.currEndOffset = b.currEndOffset + ChunkSize
+		bufferStartOffset = b.currStartOffset % BufferSize
+		bufferEndOffset = b.currEndOffset % BufferSize
+
+		// Write the remaining data by wrapping over to the next buffer block.
+		l = size - l
+		n = copy(unreadBuffer[(startOffset+l)%BufferSize:bufferEndOffset], data[:l])
+		if int64(n) != l {
+			return fmt.Errorf("could not write all the data to buffer. "+
+				"Expected bytes to be written: %d, actually written: %d", l, n)
+		}
+	}
 	return nil
 }
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -16,111 +16,150 @@ package buffer
 
 import (
 	"fmt"
+
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 )
 
-// InMemoryWriteBuffer implements WriteBuffer and is stored in memory.
-type InMemoryWriteBuffer struct {
-	// Holds the incoming data from kernel write calls.
-	currentBuffer []byte
-	// [ currentBufferStartOffset, currentBufferEndOffset) is the range of file offset for which the
-	// data is currently stored in currentBuffer.
-	currentBufferStartOffset int64
-	currentBufferEndOffset   int64
+type offset struct {
+	// [ start, end) is the range of file offset for which the
+	// data is currently stored in the buffer.
+	start, end int64
+}
 
-	// Holds the data that is sent to go storage client for upload. Data in this buffer
-	// will persist until the entire data is successfully uploaded to GCS.
-	flushedBuffer []byte
-	// [ flushedBufferStartOffset, flushedBufferEndOffset) is the range of file offset for which the
-	// data is currently stored in flushedBuffer.
-	flushedBufferStartOffset int64
-	flushedBufferEndOffset   int64
+type inMemoryBuffer struct {
+	buffer []byte
+	offset offset
+}
+
+// InMemoryWriteBuffer implements WriteBuffer and is stored in memory.
+// Refer to buffer.WriteBuffer for sample code usage.
+type InMemoryWriteBuffer struct {
+	// Holds the incoming data from kernel write calls, currently being buffered
+	// for upload in the future.
+	current inMemoryBuffer
+
+	// Holds the data that is currently being uploaded asynchronously to GCS. Data
+	// in this buffer will persist until the entire data is successfully uploaded.
+	flushed inMemoryBuffer
 
 	// Holds the last offset to which data is written.
 	fileSize int64
+
+	// Holds the buffer's size (in bytes) to be allocated to currentBuffer and flushedBuffer.
+	bufferSize int64
 }
 
-// CreateInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
-// set to nil. Memory is allocated to the buffer when the first write call comes.
-// This avoids unnecessarily bloating GCSFuse memory consumption.
-//
-// To allocate memory to the buffer, use WriteBuffer.Initialize.
-func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
-	b := &InMemoryWriteBuffer{}
+// CreateInMemoryWriteBuffer returns a new InMemoryWriteBuffer.
+// Memory is lazily allocated to current and flushed buffer on their first usage.
+func CreateInMemoryWriteBuffer(bufferSizeMB uint) (*InMemoryWriteBuffer, error) {
+	if bufferSizeMB == 0 {
+		return nil, fmt.Errorf(ZeroSizeBufferError)
+	}
+	b := &InMemoryWriteBuffer{
+		bufferSize: int64(bufferSizeMB * MiB),
+	}
 	// TODO: set mtime attribute.
-	return b
+	logger.Debugf("TODO: set mtime attribute.")
+	return b, nil
 }
 
-func (b *InMemoryWriteBuffer) Initialize(sizeInMB int) {
-	if ChunkSize == 0 {
-		ChunkSize = int64(sizeInMB * MiB)
+// Allocates memory to currentBuffer if it hasn't been allocated already.
+func (b *InMemoryWriteBuffer) ensureCurrentBuffer() error {
+	if b.bufferSize == 0 {
+		return fmt.Errorf(ZeroSizeBufferError)
 	}
-	if b.currentBuffer == nil {
-		b.currentBuffer = make([]byte, 0, ChunkSize)
-		b.currentBufferEndOffset = ChunkSize
+	if b.current.buffer == nil {
+		b.current.buffer = make([]byte, 0, b.bufferSize)
+		b.current.offset.end = b.bufferSize
 	}
-	if b.flushedBuffer == nil {
-		b.flushedBuffer = make([]byte, 0, ChunkSize)
-	}
+	return nil
 }
 
-func (b *InMemoryWriteBuffer) WriteAt(content []byte, receivedContentStartOffset int64) error {
+// Allocates memory to flushedBuffer if it hasn't been allocated already.
+func (b *InMemoryWriteBuffer) ensureFlushedBuffer() error {
+	if b.bufferSize == 0 {
+		return fmt.Errorf(ZeroSizeBufferError)
+	}
+	if b.flushed.buffer == nil {
+		b.flushed.buffer = make([]byte, 0, b.bufferSize)
+	}
+	return nil
+}
+
+func (b *InMemoryWriteBuffer) WriteAt(content []byte, offset int64) error {
 	dataSize := int64(len(content))
-	receivedContentEndOffset := receivedContentStartOffset + dataSize
+	receivedContentEndOffset := offset + dataSize
+	if dataSize == 0 {
+		return nil
+	}
+
+	// Ensure b.currentBuffer != nil
+	if err := b.ensureCurrentBuffer(); err != nil {
+		return err
+	}
 
 	// If receivedContentStartOffset is not in the range of offsets stored in current
 	// buffer, trigger temp file flow for writes.
-	if receivedContentStartOffset < b.currentBufferStartOffset ||
-		b.currentBufferEndOffset < receivedContentStartOffset {
+	if offset < b.current.offset.start ||
+		offset > b.current.offset.end {
 		// TODO: finalise write and trigger temp file flow.
+		logger.Debugf("TODO: finalise write and trigger temp file flow.")
 		return fmt.Errorf(NonSequentialWriteError)
 	}
 
-	// If data can be completely written to the current buffer block, write it.
-	if receivedContentEndOffset <= b.currentBufferEndOffset {
-		return b.copyDataToBuffer(receivedContentStartOffset, receivedContentEndOffset, content)
-	} else {
-		// Below logic is written with the assumption that the data received in a
-		// single write call from the kernel will never exceed beyond buffer block size.
-		// This should always hold true because kernel writes never exceed 1 MiB size.
-
-		// Write the chunk of data that can be written to the current buffer block.
-		l := b.currentBufferEndOffset - receivedContentStartOffset
-		if err := b.copyDataToBuffer(receivedContentStartOffset, b.currentBufferEndOffset, content[:l]); err != nil {
-			return err
-		}
-
-		//TODO: get status of last write to GCS.
-		// If unsuccessful, trigger temp file flow.
-		// Else:
-		b.swapBuffers()
-		// TODO: trigger async upload to GCS for flushedBuffer.
-
-		// Update buffer offsets for next buffer block.
-		b.currentBufferStartOffset = b.currentBufferEndOffset
-		b.currentBufferEndOffset += ChunkSize
-		// Write the remaining data to currentBuffer.
-		return b.copyDataToBuffer(b.currentBufferStartOffset, receivedContentEndOffset, content[l:])
+	// If data can be completely written to the current buffer, write it.
+	if receivedContentEndOffset <= b.current.offset.end {
+		return b.copyDataToBuffer(offset, receivedContentEndOffset, content)
 	}
+	// Else, write the chunk of data that can be written to the current buffer block.
+	l := b.current.offset.end - offset
+	err := b.copyDataToBuffer(offset, b.current.offset.end, content[:l])
+	if err != nil {
+		return err
+	}
+	//TODO: get status of last write to GCS. If unsuccessful, trigger temp file flow.
+	logger.Debugf("TODO: get status of last write to GCS and if " +
+		"unsuccessful, trigger temp file flow.")
+	if err = b.advanceToNextChunk(); err != nil {
+		return err
+	}
+	// TODO: trigger async upload to GCS for flushedBuffer.
+	logger.Debugf("TODO: trigger async upload to GCS for flushedBuffer.")
+
+	// Write the remaining data to currentBuffer.
+	return b.copyDataToBuffer(b.current.offset.start, receivedContentEndOffset, content[l:])
 }
 
+// Copies received content to currentChunk if it has capacity.
 func (b *InMemoryWriteBuffer) copyDataToBuffer(contentStartOffset, contentEndOffset int64, content []byte) error {
 	dataSize := int64(len(content))
-	si := contentStartOffset % ChunkSize
+	si := contentStartOffset % b.bufferSize
 	ei := si + dataSize
-
-	n := copy(b.currentBuffer[si:ei], content)
-	if int64(n) != dataSize {
-		return fmt.Errorf(DataNotWrittenCompletelyError, dataSize, n)
+	if ei > b.bufferSize {
+		return fmt.Errorf(NotEnoughSpaceInCurrentBuffer)
 	}
+
+	copy(b.current.buffer[si:ei], content)
 	b.updateFileSize(contentEndOffset)
 	return nil
 }
 
-func (b *InMemoryWriteBuffer) swapBuffers() {
-	b.flushedBufferStartOffset = b.currentBufferStartOffset
-	b.flushedBufferEndOffset = b.currentBufferEndOffset
-	b.flushedBuffer, b.currentBuffer = b.currentBuffer, b.flushedBuffer
-	clear(b.currentBuffer)
+func (b *InMemoryWriteBuffer) advanceToNextChunk() error {
+	// ensure b.flushed.buffer != nil
+	if err := b.ensureFlushedBuffer(); err != nil {
+		return err
+	}
+
+	// Move current buffer to flushed buffer.
+	b.flushed.buffer, b.current.buffer = b.current.buffer, b.flushed.buffer
+	b.flushed.offset.start = b.current.offset.start
+	b.flushed.offset.end = b.current.offset.end
+
+	// Make current buffer ready for new writes coming from kernel.
+	clear(b.current.buffer)
+	b.current.offset.start = b.current.offset.end
+	b.current.offset.end += b.bufferSize
+	return nil
 }
 
 // After successful copy of data to buffer, increments the bytes written so far.
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -18,15 +18,24 @@ import (
 	"fmt"
 )
 
+// InMemoryWriteBuffer implements WriteBuffer and is stored in memory.
 type InMemoryWriteBuffer struct {
-	// Holds the data to be written to GCS. At any time, Buffer holds the data for
-	// 2 GCS write calls.
-	buffer []byte
-	// [ bufferStartOffset, bufferEndOffset] is the range of file offset for which the
-	// data is currently stored in buffer.
-	bufferStartOffset int64
-	bufferEndOffset   int64
+	// Holds the incoming data from kernel write calls.
+	currentBuffer []byte
+	// [ currentBufferStartOffset, currentBufferEndOffset) is the range of file offset for which the
+	// data is currently stored in currentBuffer.
+	currentBufferStartOffset int64
+	currentBufferEndOffset   int64
 
+	// Holds the data that is sent to go storage client for upload. Data in this buffer
+	// will persist until the entire data is successfully uploaded to GCS.
+	flushedBuffer []byte
+	// [ flushedBufferStartOffset, flushedBufferEndOffset) is the range of file offset for which the
+	// data is currently stored in flushedBuffer.
+	flushedBufferStartOffset int64
+	flushedBufferEndOffset   int64
+
+	// Holds the last offset to which data is written.
 	fileSize int64
 }
 
@@ -42,60 +51,76 @@ func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
 }
 
 func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
-	ChunkSize = int64(sizeInMB * MiB)
-	BufferSize = 2 * ChunkSize
-	b.bufferEndOffset = ChunkSize
-	if b.buffer == nil {
-		b.buffer = make([]byte, 0, BufferSize)
+	if ChunkSize == 0 {
+		ChunkSize = int64(sizeInMB * MiB)
+	}
+	if b.currentBuffer == nil {
+		b.currentBuffer = make([]byte, 0, ChunkSize)
+		b.currentBufferEndOffset = ChunkSize
+	}
+	if b.flushedBuffer == nil {
+		b.flushedBuffer = make([]byte, 0, ChunkSize)
 	}
 }
 
-func (b *InMemoryWriteBuffer) WriteAt(data []byte, dataStartOffset int64) error {
-	dataSize := int64(len(data))
-	dataEndOffset := dataStartOffset + dataSize
+func (b *InMemoryWriteBuffer) WriteAt(content []byte, receivedContentStartOffset int64) error {
+	dataSize := int64(len(content))
+	receivedContentEndOffset := receivedContentStartOffset + dataSize
 
-	// If dataStartOffset is not in the range of offsets stored in buffer block,
-	// trigger temp file flow for writes.
-	if dataStartOffset < b.bufferStartOffset || b.bufferEndOffset < dataStartOffset {
+	// If receivedContentStartOffset is not in the range of offsets stored in current
+	// buffer, trigger temp file flow for writes.
+	if receivedContentStartOffset < b.currentBufferStartOffset ||
+		b.currentBufferEndOffset < receivedContentStartOffset {
 		// TODO: finalise write and trigger temp file flow.
 		return fmt.Errorf(NonSequentialWriteError)
 	}
 
-	// If data can be completely written without wrapping to previously written
-	// buffer block (beyond the current block of buffer), write it.
-	if dataEndOffset <= b.bufferEndOffset {
-		return b.copyDataToBuffer(dataStartOffset, dataEndOffset, data)
+	// If data can be completely written to the current buffer block, write it.
+	if receivedContentEndOffset <= b.currentBufferEndOffset {
+		return b.copyDataToBuffer(receivedContentStartOffset, receivedContentEndOffset, content)
 	} else {
 		// Below logic is written with the assumption that the data received in a
 		// single write call from the kernel will never exceed beyond buffer block size.
+		// This should always hold true because kernel writes never exceed 1 MiB size.
 
 		// Write the chunk of data that can be written to the current buffer block.
-		l := b.bufferEndOffset - dataStartOffset
-		if err := b.copyDataToBuffer(dataStartOffset, b.bufferEndOffset, data[:l]); err != nil {
+		l := b.currentBufferEndOffset - receivedContentStartOffset
+		if err := b.copyDataToBuffer(receivedContentStartOffset, b.currentBufferEndOffset, content[:l]); err != nil {
 			return err
 		}
 
-		//TODO: get status of last write to GCS, and trigger async write for current block.
+		//TODO: get status of last write to GCS.
+		// If unsuccessful, trigger temp file flow.
+		// Else:
+		b.swapBuffers()
+		// TODO: trigger async upload to GCS for flushedBuffer.
 
 		// Update buffer offsets for next buffer block.
-		b.bufferStartOffset = b.bufferEndOffset
-		b.bufferEndOffset += ChunkSize
-		// Write the remaining data by wrapping over to the next buffer block.
-		return b.copyDataToBuffer(b.bufferStartOffset, dataEndOffset, data[l:])
+		b.currentBufferStartOffset = b.currentBufferEndOffset
+		b.currentBufferEndOffset += ChunkSize
+		// Write the remaining data to currentBuffer.
+		return b.copyDataToBuffer(b.currentBufferStartOffset, receivedContentEndOffset, content[l:])
 	}
 }
 
-func (b *InMemoryWriteBuffer) copyDataToBuffer(dataStartOffset, dataEndOffset int64, data []byte) error {
-	dataSize := int64(len(data))
-	si := dataStartOffset % BufferSize
+func (b *InMemoryWriteBuffer) copyDataToBuffer(contentStartOffset, contentEndOffset int64, content []byte) error {
+	dataSize := int64(len(content))
+	si := contentStartOffset % ChunkSize
 	ei := si + dataSize
 
-	n := copy(b.buffer[si:ei], data)
+	n := copy(b.currentBuffer[si:ei], content)
 	if int64(n) != dataSize {
 		return fmt.Errorf(DataNOtWrittenCompletelyError, dataSize, n)
 	}
-	b.updateFileSize(dataEndOffset)
+	b.updateFileSize(contentEndOffset)
 	return nil
+}
+
+func (b *InMemoryWriteBuffer) swapBuffers() {
+	b.flushedBufferStartOffset = b.currentBufferStartOffset
+	b.flushedBufferEndOffset = b.currentBufferEndOffset
+	b.flushedBuffer, b.currentBuffer = b.currentBuffer, b.flushedBuffer
+	clear(b.currentBuffer)
 }
 
 // After successful copy of data to buffer, increments the bytes written so far.
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -111,7 +111,7 @@ func (b *InMemoryWriteBuffer) WriteAt(content []byte, offset int64) error {
 	}
 
 	var contentWrittenSoFar int64
-	for true {
+	for {
 		n := b.writePartialContentToBuffer(content[contentWrittenSoFar:], offset+contentWrittenSoFar)
 		contentWrittenSoFar += n
 		if contentWrittenSoFar == int64(len(content)) {
@@ -127,7 +127,6 @@ func (b *InMemoryWriteBuffer) WriteAt(content []byte, offset int64) error {
 		// TODO: trigger async upload to GCS for flushedBuffer.
 		logger.Debugf("TODO: trigger async upload to GCS for flushedBuffer.")
 	}
-	return nil
 }
 
 // Helper method to copy received content to currentBuffer.
--- a/internal/fs/buffer/inmemory_write_buffer.go
+++ b/internal/fs/buffer/inmemory_write_buffer.go
@@ -112,10 +112,7 @@ func (b *InMemoryWriteBuffer) WriteAt(content []byte, offset int64) error {
 
 	var contentWrittenSoFar int64
 	for true {
-		n, err := b.writePartialContentToBuffer(content[contentWrittenSoFar:], offset+contentWrittenSoFar)
-		if err != nil {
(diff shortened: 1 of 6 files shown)


internal/fs/fs.go
```diff
@@ -2083,17 +2084,15 @@ func (fs *fileSystem) WriteFile(
...
 	} else {
 		// Trigger the temp file flow.
-		if err := in.Write(ctx, op.Data, op.Offset); err != nil {
-			return err
-		}
+		err = in.Write(ctx, op.Data, op.Offset)
 	}
-
 	return
 }
```
mustvicky, line 2097: We could write this in a slightly more compressed and perhaps intuitive way: 

if (fs.isWriteBufferEnabled(in) { 
         err := in.WriteToBuffer(fs.mountConfig.WriteConfig.BufferSizeMB, op.Data, op.Offset)
         if (err == nil) 
               return
        log error as warning
        if (err.Error() != NonSequentialWriteError) 
              return
         else log info/warning that we are falling back to file writing
}
err = in.Write(ctx, op.Data, op.Offset)
return

internal/fs/fs.go
```diff
@@ -2083,17 +2084,15 @@ func (fs *fileSystem) WriteFile(
 	// Serve the request.
 	if fs.isWriteBufferEnabled(in) {
```
mustvicky, line 2085: Should we add a TODO that this logic requires unit tests (we can then look at all TODOs before code shipping that all the important TODOs are taken care of or not)

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -36,12 +42,65 @@ func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
...
+		b.bufferStartOffset = b.bufferEndOffset
+		b.bufferEndOffset += ChunkSize
+		// Write the remaining data by wrapping over to the next buffer block.
+		return b.copyDataToBuffer(b.bufferStartOffset, dataEndOffset, data[l:])
+	}
+}
+
+func (b *InMemoryWriteBuffer) copyDataToBuffer(dataStartOffset, dataEndOffset int64, data []byte) error {
+	dataSize := int64(len(data))
+	si := dataStartOffset % BufferSize
```
gargnitingoogle, line 90: Shouldn't this be 

```suggestion
	si := dataStartOffset % ChunkSize
```

? Or we need to track which chunk (first half or second half) of the buffer to write to..

gargnitingoogle (changes requested):
Added some comments. Will follow up offline with @ashmeenkaur for understanding the design a bit better before going forward.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -36,12 +42,65 @@ func CreateInMemoryWriteBuffer() *InMemoryWriteBuffer {
 }
 
 func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
```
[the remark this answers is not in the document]
ashmeenkaur: to move other 3 lines in if condition

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: This is obsolete now as we use 2 buffers

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,13 +15,19 @@
...
 	// Holds the data to be written to GCS. At any time, Buffer holds the data for
 	// 2 GCS write calls.
-	buffer *bytes.Buffer
+	buffer []byte
+	// [ bufferStartOffset, bufferEndOffset] is the range of file offset for which the
+	// data is currently stored in buffer.
+	bufferStartOffset int64
+	bufferEndOffset   int64
+
+	fileSize int64
```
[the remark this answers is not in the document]
ashmeenkaur: This is the total file size up to now. This will be required to serve stat calls.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,13 +15,19 @@
 package buffer
 
 import (
-	"bytes"
+	"fmt"
 )
 
 type InMemoryWriteBuffer struct {
```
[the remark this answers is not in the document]
ashmeenkaur: Added the header, thanks!
Unsure what you mean by sample code usage. This should be used with the interface public methods. Should I add the comments explaining it?

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,13 +15,19 @@
...
 	// Holds the data to be written to GCS. At any time, Buffer holds the data for
 	// 2 GCS write calls.
-	buffer *bytes.Buffer
+	buffer []byte
+	// [ bufferStartOffset, bufferEndOffset] is the range of file offset for which the
+	// data is currently stored in buffer.
+	bufferStartOffset int64
```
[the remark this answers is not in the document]
ashmeenkaur: I've renamed the variables, can you please check again if this is still required

internal/fs/fs.go
ashmeenkaur: added the TODO, thanks

internal/fs/fs.go
ashmeenkaur: Made the suggested change, thanks.

internal/fs/fs.go
```diff
@@ -2083,17 +2084,15 @@ func (fs *fileSystem) WriteFile(
...
 		// Trigger write file buffering flow.
-		err := in.WriteToBuffer(fs.mountConfig.WriteConfig.BufferSizeMB, op.Data, op.Offset)
-		if err != nil {
-			return err
+		err = in.WriteToBuffer(fs.mountConfig.WriteConfig.BufferSizeMB, op.Data, op.Offset)
+		if err != nil && err.Error() == buffer.NonSequentialWriteError {
+			// Trigger the temp file flow in case of non-sequential writes.
```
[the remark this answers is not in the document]
ashmeenkaur: Added the comments, thanks.

internal/fs/buffer/write_buffer.go
```diff
@@ -19,12 +19,15 @@ const (
...
-	InMemoryBufferThresholdMB = 50
+	InMemoryBufferThresholdMB     = 50
+	NonSequentialWriteError       = "non-sequential writes are not supported with buffer"
+	DataNOtWrittenCompletelyError = "could not write all the data to buffer. Expected bytes to be written: %d, actually written: %d"
 )
 
 // ChunkSize (bytes) is the size of data to be written in 1 write call to GCS.
 // Ensure ChunkSize <= 100MB to avoid memory bloat.
-var ChunkSize int
+var ChunkSize int64
```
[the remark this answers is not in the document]
ashmeenkaur: As discussed offline, I've added if condition before updating this variable.

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
gargnitingoogle: > Unsure what you mean by sample code usage.

Sample code usage to how this function/struct is to be used e.g. https://github.com/GoogleCloudPlatform/gcsfuse/blob/3bab5990e4e377dbd0160087cd0c70cbafbd55c9/internal/storage/gcs/bucket.go#L86 . This just makes it easier to understand how this struct/function should be used.

This can refer to the header of the interface (WriteBuffer), but I see that WriteBuffer also didn't have any sample code for usage.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+	flushedBufferStartOffset int64
+	flushedBufferEndOffset   int64
+
+	// Holds the last offset to which data is written.
+	fileSize int64
 }
 
 // CreateInMemoryWriteBuffer creates a buffer with InMemoryWriteBuffer.buffer
 // set to nil. Memory is allocated to the buffer when the first write call comes.
 // This avoids unnecessarily bloating GCSFuse memory consumption.
```
gargnitingoogle, line 44: Suggestion for header (hides internal details and intent):

```
CreateInMemoryWriteBuffer returns a new InMemoryWriteBuffer.
Its buffer is lazily allocated on its first usage.
```

internal/fs/buffer/write_buffer.go
```diff
@@ -19,22 +19,24 @@ const (
 	MiB = 1024 * 1024
 	// InMemoryBufferThresholdMB is the upper limit on the size upto which the buffer should
 	// be created in memory. Beyond this size, buffer should be on disk.
-	InMemoryBufferThresholdMB = 50
+	InMemoryBufferThresholdMB     = 50
+	NonSequentialWriteError       = "non-sequential writes are not supported with buffer"
```
gargnitingoogle, line 23:
```suggestion
	NonSequentialWriteError       = "non-sequential writes are not supported with WriteBuffer"
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+	if b.currentBuffer == nil {
+		b.currentBuffer = make([]byte, 0, ChunkSize)
+		b.currentBufferEndOffset = ChunkSize
+	}
+	if b.flushedBuffer == nil {
+		b.flushedBuffer = make([]byte, 0, ChunkSize)
+	}
+}
+
+func (b *InMemoryWriteBuffer) WriteAt(content []byte, receivedContentStartOffset int64) error {
```
gargnitingoogle, line 66: I think `offset` would do for the argument name as the rest is clear from the context.

```suggestion
func (b *InMemoryWriteBuffer) WriteAt(content []byte, offset int64) error {
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+// InMemoryWriteBuffer implements WriteBuffer and is stored in memory.
 type InMemoryWriteBuffer struct {
-	// Holds the data to be written to GCS. At any time, Buffer holds the data for
-	// 2 GCS write calls.
-	buffer *bytes.Buffer
+	// Holds the incoming data from kernel write calls.
+	currentBuffer []byte
+	// [ currentBufferStartOffset, currentBufferEndOffset) is the range of file offset for which the
+	// data is currently stored in currentBuffer.
+	currentBufferStartOffset int64
```
gargnitingoogle, line 27: Instead of startOffset and endOffset, start and end would suffice. Or you could create an range struct and use that for all sorts of offsets.

```golang
type Range struct {
  start, end int64
}

type ChunkBuffer struct
{
  ...
  contentOffset Range
  ...
}
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+	currentBufferEndOffset   int64
+
+	// Holds the data that is sent to go storage client for upload. Data in this buffer
```
gargnitingoogle, line 30:
```suggestion
	// Holds the data that is currently being uploaded asynchronously.
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
-	"bytes"
+	"fmt"
 )
 
+// InMemoryWriteBuffer implements WriteBuffer and is stored in memory.
 type InMemoryWriteBuffer struct {
-	// Holds the data to be written to GCS. At any time, Buffer holds the data for
-	// 2 GCS write calls.
-	buffer *bytes.Buffer
+	// Holds the incoming data from kernel write calls.
```
gargnitingoogle, line 23:
```suggestion
	// Holds the incoming data from kernel write calls, currently being buffered
	// for upload in future.
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+}
+
+func (b *InMemoryWriteBuffer) WriteAt(content []byte, receivedContentStartOffset int64) error {
+	dataSize := int64(len(content))
+	receivedContentEndOffset := receivedContentStartOffset + dataSize
+
+	// If receivedContentStartOffset is not in the range of offsets stored in current
+	// buffer, trigger temp file flow for writes.
+	if receivedContentStartOffset < b.currentBufferStartOffset ||
+		b.currentBufferEndOffset < receivedContentStartOffset {
```
gargnitingoogle, line 73: Let's maintain the order of operands for comparison, for ease of reader.

```suggestion
		receivedContentStartOffset > b.currentBufferEndOffset {
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+		// TODO: finalise write and trigger temp file flow.
```
gargnitingoogle, line 74: Let's put a debug log here (and for all TODOs) so that we can easily realize the flow by running with debug logs.

internal/fs/buffer/write_buffer.go
```diff
@@ -19,22 +19,24 @@ const (
 	MiB = 1024 * 1024
 	// InMemoryBufferThresholdMB is the upper limit on the size upto which the buffer should
 	// be created in memory. Beyond this size, buffer should be on disk.
```
gargnitingoogle, line 21:
```suggestion
	// be created in memory. Beyond this size, buffering would be on disk.
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+		return fmt.Errorf(NonSequentialWriteError)
+	}
+
+	// If data can be completely written to the current buffer block, write it.
+	if receivedContentEndOffset <= b.currentBufferEndOffset {
+		return b.copyDataToBuffer(receivedContentStartOffset, receivedContentEndOffset, content)
+	} else {
```
gargnitingoogle, line 81: We can open up the else-block here, as the if-case was returning, for ease of reader and for reducing indentation.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+		// Below logic is written with the assumption that the data received in a
```
gargnitingoogle, line 82: Can we put some kind of check to make sure of this, rather than assuming it?

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
gargnitingoogle: Looks better now, but have added some more comments, after the change.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+		// Update buffer offsets for next buffer block.
+		b.currentBufferStartOffset = b.currentBufferEndOffset
+		b.currentBufferEndOffset += ChunkSize
+		// Write the remaining data to currentBuffer.
+		return b.copyDataToBuffer(b.currentBufferStartOffset, receivedContentEndOffset, content[l:])
 	}
 }
 
-func (b *InMemoryWriteBuffer) WriteAt(data []byte, offset int64) error {
+func (b *InMemoryWriteBuffer) copyDataToBuffer(contentStartOffset, contentEndOffset int64, content []byte) error {
```
gargnitingoogle, line 106: Can you please state the assumptions here as a comment, so it makes it easier to identify if there is something wrong?

e.g.
```golang
// assuming
// contentEndOffset - contentStartOffset = len(byte) and
// len(byte) <= ChunkSize - (contentStartOffset % ChunkSize)
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+func (b *InMemoryWriteBuffer) swapBuffers() {
+	b.flushedBufferStartOffset = b.currentBufferStartOffset
+	b.flushedBufferEndOffset = b.currentBufferEndOffset
+	b.flushedBuffer, b.currentBuffer = b.currentBuffer, b.flushedBuffer
+	clear(b.currentBuffer)
+}
+
+// After successful copy of data to buffer, increments the bytes written so far.
+func (b *InMemoryWriteBuffer) updateFileSize(endOffset int64) {
+	if endOffset > b.fileSize {
```
gargnitingoogle, line 128: Will this check ever fail ? If yes, in what case ?

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
 func (t *MemoryBufferTest) TearDown() {}
 
+// //////////////////////////////////////////////////////////////////////
+// Helpers
+// //////////////////////////////////////////////////////////////////////
+
+func generateRandomData(size int64) []byte {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	data := make([]byte, size)
+	r.Read(data)
```
gargnitingoogle, line 56: Pass `t *MemoryBufferTest` to this function, check the output value of r.Read() and assert on that using `t`. Not doing this adds to our already big list of linter/static-check errors.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
 	// TODO: set mtime attribute.
 	return b
 }
 
-func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
-	ChunkSize = sizeInMB * MiB
-	if b.buffer == nil {
-		b.buffer = bytes.NewBuffer(make([]byte, 0, 2*sizeInMB*MiB))
+func (b *InMemoryWriteBuffer) Initialize(sizeInMB int) {
+	if ChunkSize == 0 {
```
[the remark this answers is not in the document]
gargnitingoogle: Also, setting a global variable here in a member method is a design risk. To somewhat mitigate it, could you add an error here if ChunkSize was already > 0 and earlier had a value different from what is to be assigned here, as a safety measure, and then add a test for that.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
 	b := &InMemoryWriteBuffer{}
 	// TODO: set mtime attribute.
 	return b
 }
 
-func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
-	ChunkSize = sizeInMB * MiB
-	if b.buffer == nil {
-		b.buffer = bytes.NewBuffer(make([]byte, 0, 2*sizeInMB*MiB))
+func (b *InMemoryWriteBuffer) Initialize(sizeInMB int) {
```
gargnitingoogle, line 53: What if sizeInMB is passed as 0 or negative? It's a public function, so it should do complete input validations.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
 	t.mb = CreateInMemoryWriteBuffer()
 
-	t.mb.InitializeBuffer(bufferSizeMB)
+	t.mb.InitializeBuffer(sizeInMB)
+
+	AssertEq(sizeInMB*MiB, ChunkSize)
+	AssertEq(sizeInMB*MiB, cap(t.mb.currentBuffer))
+	AssertEq(0, len(t.mb.currentBuffer))
+	AssertEq(0, t.mb.currentBufferStartOffset)
+	AssertEq(sizeInMB*MiB, t.mb.currentBufferEndOffset)
```
gargnitingoogle, line 82: As we have already asserted on `ChunkSize == sizeInMB*MiB`

```suggestion
	AssertEq(ChunkSize, t.mb.currentBufferEndOffset)
```

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
 func (t *MemoryBufferTest) TestInitializeInMemoryBuffer() {
-	bufferSizeMB := 10
+	sizeInMB := 1
 	t.mb = CreateInMemoryWriteBuffer()
 
-	t.mb.InitializeBuffer(bufferSizeMB)
+	t.mb.InitializeBuffer(sizeInMB)
+
+	AssertEq(sizeInMB*MiB, ChunkSize)
+	AssertEq(sizeInMB*MiB, cap(t.mb.currentBuffer))
```
gargnitingoogle, line 79: Nit: There is some duplication in the checks on currentBuffer* and flushedBuffer* . See if it can be reduced/removed.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+func (t *MemoryBufferTest) TestSingleWriteToInMemoryBuffer() {
+	// Allocate a buffer
+	t.TestInitializeInMemoryBuffer()
+	data := []byte("Taco")
+
+	// Write to buffer
+	err := t.mb.WriteAt(data, 0)
+
+	AssertEq(nil, err)
+	AssertEq(t.mb.fileSize, 4)
```
gargnitingoogle, line 99: Let's define 4 a const above in the setup section of the test or use `len(data)` in place of 4 here for easier maintenance.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	AssertEq(true, bytes.Equal(data, t.mb.currentBuffer[0:t.mb.fileSize]))
```
gargnitingoogle, line 100: Somewhat shorter
```suggestion
	AssertTrue(bytes.Equal(data, t.mb.currentBuffer[0:t.mb.fileSize]))
```

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
 func (t *MemoryBufferTest) TestCreateEmptyInMemoryBuffer() {
 	t.mb = CreateInMemoryWriteBuffer()
 
-	AssertEq(nil, t.mb.buffer)
+	AssertEq(nil, t.mb.currentBuffer)
+	AssertEq(nil, t.mb.flushedBuffer)
 	AssertEq(0, ChunkSize)
 }
 
 func (t *MemoryBufferTest) TestInitializeInMemoryBuffer() {
```
gargnitingoogle, line 72: Please add a negative test case of writing to an uninitialized IMWB .

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
-	bufferSizeMB := 10
+	sizeInMB := 1
 	t.mb = CreateInMemoryWriteBuffer()
 
-	t.mb.InitializeBuffer(bufferSizeMB)
+	t.mb.InitializeBuffer(sizeInMB)
```
gargnitingoogle, line 76: DIdn't we rename this function to just Initialize ?

internal/fs/buffer/inmemory_write_buffer_test.go
gargnitingoogle: Same goes for other cases belo.w

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	AssertEq(true, bytes.Equal(data, t.mb.currentBuffer[0:t.mb.fileSize]))
+}
+
+func (t *MemoryBufferTest) TestMultipleSequentialWritesToInMemoryBuffer() {
+	// Allocate a buffer
+	t.TestInitializeInMemoryBuffer()
+
+	// Write to buffer
+	err := t.mb.WriteAt([]byte("Taco"), 0)
+	AssertEq(nil, err)
```
gargnitingoogle, line 109: Let's verify the increasing value of fileSize, all 4 offsets after each write, to verify that they are all getting updated successfully. Not in every test, just in a few.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+func (t *MemoryBufferTest) Test4MBWriteTo2MBInMemoryBuffer() {
+	// Allocate a buffer
+	t.TestInitializeInMemoryBuffer()
+	data1 := generateRandomData(MiB)
+	data2 := generateRandomData(MiB)
+	data3 := generateRandomData(MiB)
+	data4 := generateRandomData(MiB)
+
+	// Write to buffer
+	err := t.mb.WriteAt(data1, 0)
```
gargnitingoogle, line 165: Let's add test(s) for cases where passed buffer slice is empty e.g. `buf[x:x]`.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+func (t *MemoryBufferTest) TestWriteAtChunkSizeOffset() {
+	// Allocate a buffer
+	t.TestInitializeInMemoryBuffer()
+
+	// Write to buffer
+	err := t.mb.WriteAt([]byte("Taco"), MiB)
+	AssertEq(nil, err)
+
+	AssertEq(t.mb.fileSize, MiB+4)
+	AssertEq(true, bytes.Equal([]byte("Taco"), t.mb.currentBuffer[0:4]))
```
gargnitingoogle, line 217: Can we verify the offsets of flushedBuffer here to see that they got updated despite flushedBuffer never being written to ?

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+func (t *MemoryBufferTest) TestWriteJustAfterChunkSizeOffset() {
+	// Allocate a buffer
+	t.TestInitializeInMemoryBuffer()
+
+	// Write to buffer
+	err := t.mb.WriteAt([]byte("Taco"), MiB+1)
+
+	AssertNe(nil, err)
+	AssertEq(NonSequentialWriteError, err.Error())
+	AssertEq(t.mb.fileSize, 0)
```
gargnitingoogle, line 229: Can we verify the offsets of flushedBuffer here to see that they got updated despite flushedBuffer never being written to ?

gargnitingoogle: Let's also check the currentBuffer offsets.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	err := t.mb.WriteAt(data1, 0)
+	AssertEq(nil, err)
+	err = t.mb.WriteAt(data2, MiB)
+	AssertEq(nil, err)
+	err = t.mb.WriteAt(data2, 2*MiB-KiB)
+	AssertEq(nil, err)
+
+	AssertEq(t.mb.fileSize, 2*MiB+KiB)
+	AssertEq(true, bytes.Equal(data2[0:KiB], t.mb.flushedBuffer[MiB-KiB:MiB]))
+	AssertEq(true, bytes.Equal(data2[KiB:], t.mb.currentBuffer[0:KiB]))
```
gargnitingoogle, line 248: Let's also verify that the offsets got updated properly.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +46,240 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+func (t *MemoryBufferTest) TestRandomWriteOnAnAlreadyWrittenBufferBlockShouldFail() {
+	// Allocate a buffer
+	t.TestInitializeInMemoryBuffer()
+	data1 := generateRandomData(MiB)
+	data2 := generateRandomData(2 * KiB)
+
+	// Write to buffer
+	err := t.mb.WriteAt(data1, 0)
+	AssertEq(nil, err)
+	err = t.mb.WriteAt(data2, MiB)
```
gargnitingoogle, line 260: Let's add test(s) for scenarios where two consecutive writes skipped more than a chunk i.e. a chunk in-between got never written to, both with cases of contentSize <= chunkSize and > chunkSize.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
 
-func (b *InMemoryWriteBuffer) WriteAt(data []byte, offset int64) error {
+func (b *InMemoryWriteBuffer) copyDataToBuffer(contentStartOffset, contentEndOffset int64, content []byte) error {
+	dataSize := int64(len(content))
+	si := contentStartOffset % ChunkSize
+	ei := si + dataSize
+
+	n := copy(b.currentBuffer[si:ei], content)
+	if int64(n) != dataSize {
+		return fmt.Errorf(DataNotWrittenCompletelyError, dataSize, n)
```
gargnitingoogle, line 113: I don't any test for this scenario. It is not possible to cause this ?

gargnitingoogle (changes requested):
Completed reviewing the whole code.

internal/fs/fs.go
```diff
@@ -2080,20 +2081,23 @@ func (fs *fileSystem) WriteFile(
...
+		if err == nil {
+			return
 		}
-	} else {
-		// Trigger the temp file flow.
-		if err := in.Write(ctx, op.Data, op.Offset); err != nil {
-			return err
+		logger.Warnf("Failed to complete write with buffer: %v", err)
+		if err.Error() != buffer.NonSequentialWriteError {
+			return
```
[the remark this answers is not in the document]
ashmeenkaur: This is the case of unknown error type and will be logged by error mapper.

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
ashmeenkaur: removed the global variable

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+		b.currentBuffer = make([]byte, 0, ChunkSize)
+		b.currentBufferEndOffset = ChunkSize
+	}
+	if b.flushedBuffer == nil {
+		b.flushedBuffer = make([]byte, 0, ChunkSize)
+	}
+}
+
+func (b *InMemoryWriteBuffer) WriteAt(content []byte, receivedContentStartOffset int64) error {
+	dataSize := int64(len(content))
```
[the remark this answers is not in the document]
ashmeenkaur: I've validated that kernel doesn't send write calls with empty or nil op.data, so this will never happen.

internal/fs/buffer/write_buffer.go
```diff
@@ -19,22 +19,24 @@ const (
...
-	InMemoryBufferThresholdMB = 50
+	InMemoryBufferThresholdMB     = 50
```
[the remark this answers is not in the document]
ashmeenkaur: Renamed to MaxInMemoryBufferSizeMB, as the chunk size and buffer size won't necessarily be same.

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: This will always hold true because we never receive more than 1 MiB data in 1 kernel write call. I don't think there's a need for this check. Let me know what you think.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,117 @@
...
+		// If unsuccessful, trigger temp file flow.
+		// Else:
+		b.swapBuffers()
+		// TODO: trigger async upload to GCS for flushedBuffer.
+
+		// Update buffer offsets for next buffer block.
+		b.currentBufferStartOffset = b.currentBufferEndOffset
+		b.currentBufferEndOffset += ChunkSize
+		// Write the remaining data to currentBuffer.
+		return b.copyDataToBuffer(b.currentBufferStartOffset, receivedContentEndOffset, content[l:])
```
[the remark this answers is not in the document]
ashmeenkaur: This case will never happen as kernel write calls never exceed 1 MiB size and minimum buffer size that a user can configure is 1 MiB as well. So if we are advancing the chunk, we will always have the space to write the data.

internal/fs/buffer/inmemory_write_buffer.go
gargnitingoogle: This is a generic utility, which in this case will get its data from kernel, so that assumption will never be tested in this use case. But as this is a public interface, it can really be used by any other caller. So, let's either (a) state (int header), check (in implementation that that case isn't supported) and test (as a negative unit test) that assumption, or (b) not make that assumption, add check and test.

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
gargnitingoogle: This is a generic utility, which in this case will get its data from kernel, so that assumption will never be tested in this use case. But as this is a public interface, it can really be used by any other caller. 

So, let's explicitly state that assumption in the function/interface-function header, and add check (in implementation that that case isn't supported) and test it (as a negative unit test).

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: Yes, if a user overwrites some previously written content which is currently in buffer.

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
ashmeenkaur: handled the case in copyDataToBuffer

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: range is a reserved keyword so used offset instead, thanks :)

internal/fs/buffer/inmemory_write_buffer_test.go
ashmeenkaur: Added to a few tests

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
ashmeenkaur: Added to the WriteBuffer interface

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,370 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	err = t.mb.WriteAt([]byte(smallContent1), 5*MiB)
+
+	AssertNe(nil, err)
+	AssertEq(NonSequentialWriteError, err.Error())
+	AssertEq(t.mb.fileSize, MiB+2*KiB)
+	t.validateInMemoryBuffer(t.mb.current, sizeInMB*MiB, 0, MiB, 2*MiB)
+	t.validateInMemoryBuffer(t.mb.flushed, sizeInMB*MiB, 0, 0, MiB)
+}
+
+func (t *MemoryBufferTest) TestBigContentWriteWith2ChunksSkippedShouldFail() {
```
sethiay, line 382: isn't this case same as TestRandomWriteBeyondCurrentBufferAfterSequentialWritesShouldFail ?

sethiay (commented):
Few minor comments, overall looks good.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,370 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	err = t.mb.WriteAt(data2, MiB)
+	AssertEq(nil, err)
+	AssertEq(t.mb.fileSize, 2*MiB)
+	err = t.mb.WriteAt(data3, 2*MiB)
+	AssertEq(nil, err)
+	AssertEq(t.mb.fileSize, 3*MiB)
+
+	AssertEq(t.mb.fileSize, 3*MiB)
+	AssertTrue(bytes.Equal(data3, t.mb.current.buffer[0:MiB]))
+	AssertTrue(bytes.Equal(data2, t.mb.flushed.buffer[0:MiB]))
```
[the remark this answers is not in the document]
ashmeenkaur: At any point in time, we will only have 2MiB (2*buffer size) of data. latest 1 MiB data (data3) will be in current buffer and second last 1MiB data (data2) will be in flushed buffer.

internal/fs/buffer/write_buffer.go
```diff
@@ -17,25 +17,29 @@ package buffer
...
+// // Create in-memory write or on-disk write buffer
+// writeBuffer, err := buffer.CreateInMemoryWriteBuffer(bufferSize)
+// // check err
+//
+//	for n times {
+//		err = buffer.WriteAt([]byte("content"), offset)
+//		// check err
+//	}
 type WriteBuffer interface {
-	// InitializeBuffer assigns a buffer of 2*configured buffer size to the WriteBuffer.
```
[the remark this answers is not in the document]
ashmeenkaur: @gargnitingoogle suggested that instead of relying on the ensureWriteBuffer, we should lazily allocate memory to buffers during the write calls itself. So, I removed this from method from ensureWriteBuffer and WriteBuffer interface and added ensureCurrentBuffer and ensureFlushedBuffer helpers to WriteAt method's logic.

internal/fs/buffer/inmemory_write_buffer_test.go
ashmeenkaur: Both are hitting the same condition but are slightly different in the sense that in TestRandomWriteBeyondCurrentBufferAfterSequentialWritesShouldFail, we write in the following manner:
0-1 MiB
1 MiB - 1MiB+2Kib
// no chunk of buffer size skipped
2MiB - 3MiB (fails)

while in TestBigContentWriteWith2ChunksSkippedShouldFail, we write in the following manner:
0-1MiB
1MiB - 1MiB+2KiB
//skips two chunks of buffer size
5MiB - 7MiB - (fails)

CCing @gargnitingoogle since he suggested to add this additional test case.

sethiay:
Looks good to me (and find the code well written ! ) 
As discussed offline, we may make this code usable by abstracting out the buffer logic (inMemoryBuffer) for in memory and on disk cases instead of creating separate classes with similar business logic ( InMemoryWriteBuffer in this PR is the class for in memory case). So, I would suggest to keep this PR open for now.

sethiay (approved):

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+	logger.Debugf("TODO: set mtime attribute.")
+	return b, nil
 }
 
-func (b *InMemoryWriteBuffer) InitializeBuffer(sizeInMB int) {
-	ChunkSize = sizeInMB * MiB
-	if b.buffer == nil {
-		b.buffer = bytes.NewBuffer(make([]byte, 0, 2*sizeInMB*MiB))
+// Allocates memory to currentBuffer if it hasn't been allocated already.
+func (b *InMemoryWriteBuffer) ensureCurrentBuffer() error {
```
gargnitingoogle, line 67: The code for ensureCurrentBuffer and ensureFlushedBuffer is exactly same except for b.current vs. b.flushed. Better to merge them into a single function. This will be shorter calling code as well, and easier to test too.

E.g.

```golang
func (b *inMemoryBuffer) ensure(bufferSizeMB) error {
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+	}
+	// TODO: trigger async upload to GCS for flushedBuffer.
+	logger.Debugf("TODO: trigger async upload to GCS for flushedBuffer.")
+
+	// Write the remaining data to currentBuffer.
+	return b.copyDataToBuffer(b.current.offset.start, receivedContentEndOffset, content[l:])
+}
+
+// Copies received content to currentChunk if it has capacity.
+func (b *InMemoryWriteBuffer) copyDataToBuffer(contentStartOffset, contentEndOffset int64, content []byte) error {
```
gargnitingoogle, line 134: contentEndOffset is redundant. It's same as `contentStartOffset + len(content)` .

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+
+func (b *InMemoryWriteBuffer) advanceToNextChunk() error {
+	// ensure b.flushed.buffer != nil
+	if err := b.ensureFlushedBuffer(); err != nil {
+		return err
+	}
+
+	// Move current buffer to flushed buffer.
+	b.flushed.buffer, b.current.buffer = b.current.buffer, b.flushed.buffer
+	b.flushed.offset.start = b.current.offset.start
```
gargnitingoogle, line 155: Couldn't you just instead do (i) swap b.flushed and b.current, and then (b) b.current.offset.advanceBy(2*bufferSize)? That'll be much simpler and easier to read.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+	b.flushed.offset.end = b.current.offset.end
+
+	// Make current buffer ready for new writes coming from kernel.
+	clear(b.current.buffer)
```
gargnitingoogle, line 159: Why is this clear required ?

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,371 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	t.mb, err = CreateInMemoryWriteBuffer(uint(sizeInMB))
+
+	AssertNe(nil, err)
+	AssertTrue(strings.Contains(err.Error(), ZeroSizeBufferError))
+	AssertEq(nil, t.mb)
+}
+
+func (t *MemoryBufferTest) TestWritingToUninitializedBuffer() {
+	// Write to buffer
+	err := t.mb.WriteAt([]byte(smallContent1), 0)
```
gargnitingoogle, line 110: Isn't InMemoryWriteBuffer supposed to self-allocate its current/flushed buffers as and when used (lazy allocation) ?

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,371 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+}
+
+func (t *MemoryBufferTest) TestEnsureCurrentBuffer() {
+	t.createAndValidateInMemoryBuffer(bufferSizeInMB)
+
+	err := t.mb.ensureCurrentBuffer()
+
+	AssertEq(nil, err)
+	AssertEq(bufferSizeInMB*MiB, t.mb.bufferSize)
+	t.validateInMemoryBuffer(t.mb.current, bufferSizeInMB*MiB, 0, 0, bufferSizeInMB*MiB)
```
gargnitingoogle, line 123: `bufferSizeInMB*MiB` appears too many times in this file. Could you just rename this to `bufferSize` or `bufferSizeInBytes` ?

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,371 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+}
+
+func (t *MemoryBufferTest) Test2MBWriteTo2MBInMemoryBuffer() {
+	t.createAndValidateInMemoryBuffer(bufferSizeInMB)
+	data1 := t.generateRandomData(MiB)
+	data2 := t.generateRandomData(MiB)
+
+	// Write to buffer
+	err := t.mb.WriteAt(data1, 0)
+	AssertEq(nil, err)
```
gargnitingoogle, line 187: Please leave line gaps between WriteAt() calls and corresponding Assert statements.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -15,11 +15,26 @@
...
+	"strings"
 	"testing"
+	"time"
 
 	. "github.com/jacobsa/ogletest"
 )
 
+const (
+	KiB                     = 1024
+	smallContent1           = "Taco"
```
gargnitingoogle, line 29: smallContent{1-3} are always cast to `byte[]` everywhere ? It's a good idea to just create them with that type right here.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,371 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	t.validateInMemoryBuffer(t.mb.current, bufferSizeInMB*MiB, 0, MiB, 2*MiB)
+	t.validateInMemoryBuffer(t.mb.flushed, bufferSizeInMB*MiB, 0, 0, MiB)
+}
+
+func (t *MemoryBufferTest) TestWriteAtChunkSizeOffset() {
+	t.createAndValidateInMemoryBuffer(bufferSizeInMB)
+
+	// Write to buffer
+	err := t.mb.WriteAt([]byte(smallContent1), MiB)
+	AssertEq(nil, err)
```
gargnitingoogle, line 283: Nit: Club this assert statement with the rest of the assertw below.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,371 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	t.createAndValidateInMemoryBuffer(bufferSizeInMB)
+	data1 := t.generateRandomData(MiB)
+	data2 := t.generateRandomData(2 * KiB)
+
+	// Write to buffer
+	err := t.mb.WriteAt(data1, 0)
+	AssertEq(nil, err)
+	err = t.mb.WriteAt(data2, MiB)
+	AssertEq(nil, err)
+	err = t.mb.WriteAt(data2, 2*MiB-KiB)
```
gargnitingoogle, line 314: Can we use a data3 instead of data2 here to see that it actually overwrites the previous write (`t.mb.WriteAt(data2, MiB)`) when in overlapping range ?

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -41,24 +56,371 @@ func (t *MemoryBufferTest) SetUp(ti *TestInfo) {
...
+	err = t.mb.WriteAt([]byte(smallContent1), 5*MiB)
+
+	AssertNe(nil, err)
+	AssertEq(NonSequentialWriteError, err.Error())
+	AssertEq(t.mb.fileSize, MiB+2*KiB)
+	t.validateInMemoryBuffer(t.mb.current, bufferSizeInMB*MiB, 0, MiB, 2*MiB)
+	t.validateInMemoryBuffer(t.mb.flushed, bufferSizeInMB*MiB, 0, 0, MiB)
+}
+
+func (t *MemoryBufferTest) TestBigContentWriteWith2ChunksSkippedShouldFail() {
```
gargnitingoogle, line 384: Another useful test would when the `contentSize > bufferSize` and the final bits of content go even beyond the next current buffer (i.e. the content doesn't fit in the next current buffer even after advancing), because we don't advance more than once in the code today.
I know it's not a problem in real use case because kernel will never send more than 1 MB (which is <= bufferSize), but let's a negative test to confirm it.

gargnitingoogle (changes requested):
Added some more comments. Comments are mostly about code quality, with a few which are about some logical/semantic/testing problems.

Please feel free to go ahead with this merge and address these in a later iteration.

For the record, there is also a lot of scope for refactoring in this, ideally to carve out a much richer interface for sharing with OndiskWriteBuffer. But as discussed offline, that can be taken up in a future iteration, and not block this merge.

internal/fs/buffer/inmemory_write_buffer_test.go
ashmeenkaur: go doesn't allow creation of byte[] const

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: current buffer will have previously written data. In case of random writes on current buffer, the old data will still exist if we don't clear.

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: shall I pass flushed or current buffer as a parameter then?

internal/fs/buffer/inmemory_write_buffer_test.go
ashmeenkaur: The offsets are entirely different. Data 2 is being written at MiB to MiB + 2KiB and MiB - KiB to MiB + KiB. I don't see a need for this change.

internal/fs/buffer/inmemory_write_buffer_test.go
ashmeenkaur: Not making this change to stick with the format of 
setup 
<line break>
test case
<line break>
Assertions

for the ease of reader.

internal/fs/buffer/inmemory_write_buffer_test.go
ashmeenkaur: This is the case where InMemoryWriteBuffer object itself has not been created, so the buffer doesn't have any buffer size assigned.

internal/fs/buffer/inmemory_write_buffer.go
gargnitingoogle: Right, but even if the current buffer will have previously written data, does it matter ? Is there an expectation that if user did not write to some parts of a buffer, then it must have zeros in them ? 

It's just there is a cost to clearing, and I'm trying to understand if it is needed at all ?

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: (ii) -> This will require additional if checks as in the first call to advanceToNextChunk, flushed buffer start and end offset will be 0 which unnecessarily complicates the logic.

I've made some changes to simplify this though, PTAL.

internal/fs/buffer/inmemory_write_buffer_test.go
gargnitingoogle: Okay, not a const may be, but a var can be created as [here](https://stackoverflow.com/questions/13137463/declare-a-constant-array#:~:text=There%20is%20no%20such%20thing,complex%20constants%2C%20and%20string%20constants.)

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
-	if b.buffer == nil {
-		b.buffer = bytes.NewBuffer(make([]byte, 0, 2*sizeInMB*MiB))
+// Allocates memory to currentBuffer if it hasn't been allocated already.
+func (b *InMemoryWriteBuffer) ensureCurrentBuffer() error {
+	if b.bufferSize == 0 {
+		return fmt.Errorf(ZeroSizeBufferError)
 	}
+	if b.current.buffer == nil {
+		b.current.buffer = make([]byte, 0, b.bufferSize)
+		b.current.offset.end = b.bufferSize
```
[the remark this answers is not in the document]
ashmeenkaur: > This is assuming that a flushed buffer has already been filled before this.
This is not current, we will write to current buffer and only after it is filled, we will move to flushed buffer. So in this case, flushed buffer has not been created yet.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+	}
+	if b.flushed.buffer == nil {
+		b.flushed.buffer = make([]byte, 0, b.bufferSize)
+	}
 	return nil
 }
+
+func (b *InMemoryWriteBuffer) WriteAt(content []byte, offset int64) error {
+	dataSize := int64(len(content))
+	receivedContentEndOffset := offset + dataSize
```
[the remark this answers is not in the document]
ashmeenkaur: Moved this to just before its first usage

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+	if dataSize == 0 {
+		return nil
+	}
+
+	// Ensure b.currentBuffer != nil
+	if err := b.ensureCurrentBuffer(); err != nil {
+		return err
```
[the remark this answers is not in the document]
ashmeenkaur: Used the following everywhere (for consistency)
```go
fmt.Errorf("methodName: %v:", err)
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+		return b.copyDataToBuffer(offset, receivedContentEndOffset, content)
+	}
+	// Else, write the chunk of data that can be written to the current buffer block.
+	l := b.current.offset.end - offset
+	err := b.copyDataToBuffer(offset, b.current.offset.end, content[:l])
+	if err != nil {
+		return err
+	}
+	//TODO: get status of last write to GCS. If unsuccessful, trigger temp file flow.
+	logger.Debugf("TODO: get status of last write to GCS and if " +
```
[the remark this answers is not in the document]
ashmeenkaur: Made the change but recommend not nit picking on TODO logs that we are going to remove in future.

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: Yes, this is required to keep this consistent with the temp file flow where zero bytes are present if a user doesn't write the data.

internal/fs/buffer/inmemory_write_buffer_test.go
gargnitingoogle: Ok, do we have any tests targeting overwriting ? If not, we should.

internal/fs/buffer/inmemory_write_buffer.go
gargnitingoogle: I understand. But that just makes me question the clearing in the temp-file flow as well..

For now, could you add a comment/question here to revisit this choice ?

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,156 @@
...
+
+	// Ensure b.currentBuffer != nil
+	if err := b.ensureCurrentBuffer(); err != nil {
+		return err
+	}
+
+	// If receivedContentStartOffset is not in the range of offsets stored in current
+	// buffer, trigger temp file flow for writes.
+	if offset < b.current.offset.start ||
+		offset > b.current.offset.end {
```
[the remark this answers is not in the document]
ashmeenkaur: I've explicitly kept the last offset included in the condition because kernel writes are expected to come at multiples of MiB.

Sample write logs to validate the same:
I wrote hello, hi and ashmeen to a file and we can see kernel sent the old data again just to keep the write as a multiple of 4096.
```
time="09/11/2023 04:39:40.174953" severity=TRACE msg="fuse_debug: Op 0x000001fc        connection.go:416] <- WriteFile (inode 5, PID 0, handle 32, offset 0, 5 bytes)"
...
time="09/11/2023 04:39:42.002566" severity=TRACE msg="fuse_debug: Op 0x0000020c        connection.go:416] <- WriteFile (inode 5, PID 0, handle 33, offset 0, 7 bytes)"
...
time="09/11/2023 04:39:43.454277" severity=TRACE msg="fuse_debug: Op 0x0000021a        connection.go:416] <- WriteFile (inode 5, PID 0, handle 34, offset 0, 14 bytes)"
```

internal/fs/buffer/inmemory_write_buffer.go
ashmeenkaur: Added a todo to revisit this later.

internal/fs/buffer/inmemory_write_buffer_test.go
ashmeenkaur: Updated this test to also validate on empty contents. Also caught an error in the clear syntax usage. Thanks 💯

internal/fs/buffer/inmemory_write_buffer.go
gargnitingoogle: @ashmeenkaur  Sorry for the late response. I somehow missed this earlier. 

Either way will do for now, as long as we reduce duplication of logic. We can come back and refactor later if ever needed.

gargnitingoogle (changes requested):

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
ashmeenkaur: I initially implemented this entire PR with the assumption that kernel sends only 1MiB content in a write that is why this was written in this way. Agreed that this logic would not have worked if we ever had to reuse this logic elsewhere.

I've updated the PR with the following:
-  to consider random writes if current offset goes beyond the range of offsets stored in current buffer and +1 buffer. 
- added the logic to keep on writing until we exhaust the content received (more than 1MiB possible).
- removed not enough space in buffer errors

Please take another look. Thanks!

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,165 @@
...
+	current inMemoryBuffer
+
+	// Holds the data that is currently being uploaded asynchronously to GCS. Data
+	// in this buffer will persist until the entire data is successfully uploaded.
+	flushed inMemoryBuffer
+
+	// Holds the last offset to which data is written.
+	fileSize int64
+
+	// Holds the buffer's size (in bytes) to be allocated to currentBuffer and flushedBuffer.
```
gargnitingoogle, line 53: Nit: comment needs an update. 
```suggestion
	// Holds the buffer's size (in bytes) to be allocated to current and flushed.
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,165 @@
...
+	}
 	// TODO: set mtime attribute.
-	return b
+	logger.Debugf("TODO: set mtime attribute.")
+	return b, nil
+}
+
+// Allocates memory to currentBuffer if it hasn't been allocated already.
+func (b *InMemoryWriteBuffer) ensureCurrentBuffer() error {
+	if b.bufferSize == 0 {
```
gargnitingoogle, line 73: nit: This is technically not possible because CreateInMemoryWriteBuffer can't set it to zero, and that's the only approved way to create an InMemoryWriteBuffer. So, this check is really redundant.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,165 @@
...
+
+func (b *InMemoryWriteBuffer) writePartialContentToBuffer(content []byte, offset int64) int64 {
+	if offset < b.current.offset.start ||
+		offset >= b.current.offset.end {
+		// Nothing to copy to current buffer.
+		return 0
+	}
+
+	// copy content to current buffer
+	capacityOfCurrentBuffer := b.current.offset.end - offset
```
gargnitingoogle, line 175: to make it easier to read

```suggestion
	remainingCapacityOfCurrentBuffer := b.current.offset.end - offset
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,165 @@
...
+	if err := b.ensureFlushedBuffer(); err != nil {
+		return fmt.Errorf("ensureFlushedBuffer: %w", err)
+	}
+
+	// Move current buffer to flushed buffer.
+	b.flushed.buffer, b.current.buffer = b.current.buffer, b.flushed.buffer
+	b.flushed.offset = b.current.offset
+
+	// Make current buffer ready for new writes coming from kernel.
+	// TODO: revisit this when implementing on-disk-write-buffer.
```
gargnitingoogle, line 154: just to make it clear (no-pun intended)

```suggestion
	// TODO: revisit this decision to clear, when implementing on-disk-write-buffer.
```

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,165 @@
...
+		offset >= b.current.offset.end+b.bufferSize {
+		// TODO: finalise write and trigger temp file flow.
+		logger.Debugf("Non-sequential write encountered. TODO: Switch to " +
+			"temp-file flow insteading of erroring out.")
+		return fmt.Errorf(NonSequentialWriteError)
 	}
+
+	var contentWrittenSoFar int64
+	for {
+		n := b.writePartialContentToBuffer(content[contentWrittenSoFar:], offset+contentWrittenSoFar)
```
gargnitingoogle, line 115: @ashmeenkaur  I see a bit of a tricky situation here, which might cause a deadlock.

The updated checks on the limits on the value of `offset` allow the passed `offset` to be in either the current `current` buffer or in the next `current` buffer.

Assuming that the current `current` buffer is not full yet, and the passed `offset` is in `[current.offset.end, current.offset+bufferSize)` range, then on the first iteration of this for-loop, the returned `n` will be 0 and each iteration will just do the same.

Now, contentWrittenSoFar will never be increased, and the terminating condition `contentWrittenSoFar == int64(len(content))` will never be attained.

I might be missing something. Please check and let me know.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,165 @@
...
 }
+
+// After successful copy of data to buffer, increments the bytes written so far.
+func (b *InMemoryWriteBuffer) updateFileSize(endOffset int64) {
+	if endOffset > b.fileSize {
+		b.fileSize = endOffset
+	}
+}
+
+func (b *InMemoryWriteBuffer) writePartialContentToBuffer(content []byte, offset int64) int64 {
```
gargnitingoogle, line 167: nit: for ease of the reader

```suggestion
func (b *InMemoryWriteBuffer) writePartialContentToCurrentBuffer(content []byte, offset int64) int64 {
```

gargnitingoogle (changes requested):
thanks @ashmeenkaur . The code looks much cleaner and closer to the real code now. Added some comments. All but one are about readability, one is about a potential deadlock.

internal/fs/buffer/inmemory_write_buffer.go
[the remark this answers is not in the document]
gargnitingoogle: Sure, I'll open any new questions on the updated design then. Marking this thread closed.

internal/fs/buffer/inmemory_write_buffer.go
```diff
@@ -15,33 +15,165 @@
...
+}
+
+type inMemoryBuffer struct {
+	buffer []byte
+	offset offset
+}
+
+// InMemoryWriteBuffer implements WriteBuffer and is stored in memory.
+// Refer to buffer.WriteBuffer for sample code usage.
 type InMemoryWriteBuffer struct {
```
gargnitingoogle, line 41: This struct need not be public. This struct by itself isn't needed anywhere, only the interface WriteBuffer is. 

This is not very different from https://github.com/GoogleCloudPlatform/gcsfuse/blob/3679762e4ed291a26ae73c471abfc22d57799227/internal/storage/chunk_uploader.go#L37 .
If we just change `CreateInMemoryWriteBuffer` to return WriteBuffer (similar to https://github.com/GoogleCloudPlatform/gcsfuse/blob/3679762e4ed291a26ae73c471abfc22d57799227/internal/storage/chunk_uploader.go#L55), instead of `*InMemoryWriteBuffer`, it'll all fall in place.

gargnitingoogle (changes requested):

internal/fs/inode/file_test.go
```diff
@@ -954,17 +954,3 @@ func (t *FileTest) TestMultipleCallsToWriteToBufferCreatesBufferOnce() {
 	AssertEq(bufferCreated, writeToBuffer(t, bufferSizeMB, content, 13))
 	AssertEq(bufferCreated, writeToBuffer(t, bufferSizeMB, content, 25))
 }
-
-func (t *FileTest) TestEnsureBufferCreatesInMemoryBufferWhenBufferSizeIsLessThan50() {
```
ashmeenkaur, line 958: Unfortunately I had to remove this unit test after making InMemoryWriteBuffer struct un-exported.

gargnitingoogle (dismissed):
LGTM

internal/fs/inode/file_test.go
gargnitingoogle: Discussed this offline that this test is now hard to maintain given  InMemoryWriteBuffer being private, and we'll look to reintroduce it later.

internal/fs/buffer/inmemory_write_buffer_test.go
```diff
@@ -36,29 +58,369 @@ var _ TearDownInterface = &MemoryBufferTest{}
...
+}
+
+func (t *MemoryBufferTest) TestWriteJustBeforeChunkSizeOffset() {
+	t.createAndValidateInMemoryBuffer(bufferSizeInMB)
+
+	// Write to buffer
+	err := t.mb.WriteAt(smallContent1, MiB-1)
+	AssertEq(nil, err)
+
+	//AssertEq(MiB+smallContent1Size-1, t.mb.fileSize)
```
sethiay, line 271: nit: uncomment or remove this?

sethiay (dismissed):

sethiay (approved):

PR #1429, opened 2023-10-10
Adding FileInfo data class

raj-prince:
### Description
Adding data class which will be used as an entry for LRU cache

NA

### Testing details
1. Manual - Basic sanity
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/data/file_info.go
+++ b/internal/cache/data/file_info.go
@@ -21,6 +21,6 @@ type FileInfo struct {
 	FileSize         uint64
 }
 
-func (fi *FileInfo) Size() uint64 {
+func (fi FileInfo) Size() uint64 {
 	return fi.FileSize
 }
--- a/internal/cache/data/file_info.go
+++ b/internal/cache/data/file_info.go
@@ -14,8 +14,32 @@
 
 package data
 
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+const InvalidKeyAttributes = "un-initialized key attributes"
+
+type FileInfoKey struct {
+	BucketName         string
+	BucketCreationTime time.Time
+	ObjectName         string
+}
+
+// Key will return a string, combining all the attributes of FileInfoKey.
+// Returns error in case of uninitialized value.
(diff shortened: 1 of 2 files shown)


internal/cache/data/file_info.go
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package data
+
+type FileInfo struct {
+	Name             string
```
sethiay, line 18: should we create another struct for Name and expose a method that gives complete name (appending bucket, bucket creation, object name) to be used as key for storing in File Cache ?

raj-prince:
``` go
type FileInfoKey struct {
	BucketName string
	BucketCreationTime string
	ObjectName string
}

func (fik FileInfoKey) Key() string {
	return fik.BucketName + fik.BucketCreationTime + fik.ObjectName
}

type FileInfo struct {
	Key             string
	ObjectGeneration string
	Offset           uint64
	FileSize         uint64
}

```

sethiay: Key in FileInfo should be of type FileInfoKey.

internal/cache/data/file_info.go
```diff
@@ -0,0 +1,50 @@
...
+
+package data
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+const InvalidKeyAttributes = "un-initialized key attributes"
```
sethiay, line 23: nit: key attributes not initialised.

sethiay (dismissed):

internal/cache/data/file_info.go
raj-prince: done.

sethiay (approved):

PR #1449, opened 2023-10-18
Using locker with invariant check in LRU cache

raj-prince:
### Description
Changing the normal mutex with the customise locker which supports debugging and invariant check.

NA

### Testing details
1. Manual - NA
2. Unit tests - Automation
3. Integration tests - NA

### DIFF
--- a/internal/cache/lru/lru.go
+++ b/internal/cache/lru/lru.go
@@ -84,7 +84,7 @@ func NewCache(maxSize uint64) *Cache {
 	return c
 }
 
-// CheckInvariants panic if any internal invariants have been violated.
+// checkInvariants panic if any internal invariants have been violated.
 func (c *Cache) checkInvariants() {
 	// INVARIANT: maxSize > 0
 	if !(c.maxSize > 0) {
--- a/internal/cache/lru/lru.go
+++ b/internal/cache/lru/lru.go
@@ -19,7 +19,8 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"sync"
+
+	"github.com/googlecloudplatform/gcsfuse/internal/locker"
 )
 
 // Predefined error message returned by the Cache.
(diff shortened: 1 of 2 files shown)


internal/cache/lru/lru.go
```diff
@@ -70,12 +71,14 @@ type entry struct {
 	Value ValueType
 }
 
-// New initialize a cache with the supplied maxSize, which must be greater than
+// Init initialize a cache with the supplied maxSize, which must be greater than
 // zero.
-func NewCache(maxSize uint64) (c Cache) {
+func (c *Cache) Init(maxSize uint64) {
```
sethiay, line 76: we can keep it NewCache only.

internal/cache/lru/lru.go
```diff
@@ -70,12 +71,14 @@ type entry struct {
...
 	c.maxSize = maxSize
 	c.index = make(map[string]*list.Element)
-	return
+
+	// Set up invariant checking.
+	c.mu = locker.New("LRUCache", c.CheckInvariants)
 }
 
 // CheckInvariants panic if any internal invariants have been violated.
```
sethiay, line 84: we can make checkInvariants private and remove "The careful user can arrange to call this at crucial moments."?

internal/cache/lru/lru.go
```diff
@@ -70,17 +71,21 @@ type entry struct {
...
+		maxSize: maxSize,
+		index:   make(map[string]*list.Element),
+	}
+
+	// Set up invariant checking.
+	c.mu = locker.New("LRUCache", c.checkInvariants)
+	return c
 }
 
 // CheckInvariants panic if any internal invariants have been violated.
```
sethiay, line 87: nit: checkInvariants

sethiay (approved):

sethiay (approved):

PR #1451, opened 2023-10-18
Read cache job 1

sethiay:
### Description
Add structure for File Download Job struct. Currently, I haven't added unimplemented methods for job.Start() and job.Retry as I am figuring out their final function signatures after today's call.

NA

### Testing details
1. Manual - Ran unit tests.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -132,8 +132,8 @@ func (job *Job) Cancel() {
 	defer job.mu.Unlock()
 }
 
-// Download starts async download if not already started and waits till the
-// download is completed for given offset if waitForDownload is true.
+// Download downloads object till the given offset if not already downloaded
+// and waits for download if waitForDownload is true.
 //
 // Acquires and releases LOCK(job.mu)
 func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool) (jobStatus JobStatus) {
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -81,8 +81,6 @@ type JobStatus struct {
 // jobSubscriber represents a subscriber waiting on async download of job to
 // complete downloading at least till the subscribed offset.
 type jobSubscriber struct {
-	notificationC    chan<- JobStatus
-	subscribedOffset int64
 }
 
 func NewJob(object *gcs.MinObject, bucket gcs.Bucket, filePath string,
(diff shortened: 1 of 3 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -0,0 +1,141 @@
...
+	/////////////////////////
+	// Constant data
+	/////////////////////////
+
+	object               *gcs.MinObject
+	bucket               gcs.Bucket
+	filePath             string
+	fileInfoCache        *lru.Cache
+	sequentialReadSizeMb int32
+	filePerm             os.FileMode
```
raj-prince, line 51: Should we club all file_creation parameter as a struct? E.g. LocalFileSpec { filePath, filePerm, fileUid, fileGid}

internal/cache/file/downloader/job.go
```diff
@@ -0,0 +1,141 @@
...
+func (job *Job) Cancel() {
+	job.mu.Lock()
+	defer job.mu.Unlock()
+}
+
+// Download downloads object till the given offset if not already downloaded
+// and waits for download if waitForDownload is true.
+//
+// Acquires and releases LOCK(job.mu)
+func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool) (jobStatus JobStatus) {
```
raj-prince, line 137: Just write a comment TODO, if you want to implement these method later.

internal/cache/file/downloader/job.go
```diff
@@ -0,0 +1,141 @@
...
+	job.init()
+	return
+}
+
+// CheckInvariants panic if any internal invariants have been violated.
+func (job *Job) checkInvariants() {
+	// INVARIANT: Each subscriber is of type jobSubscriber
+	for e := job.subscribers.Front(); e != nil; e = e.Next() {
+		switch e.Value.(type) {
+		case jobSubscriber:
```
raj-prince, line 109: Why switch case? - We will not support types other than jobSubscriber?

raj-prince (changes requested):
Few comments.

internal/cache/file/downloader/job.go
sethiay: I couldn't find a simple/efficient way to do this using if statement. So, will continue with switch.

raj-prince: Tried this - https://go.dev/play/p/0QP63lALIt4

raj-prince (approved):

PR #1452, opened 2023-10-18
Adding logic to add and notify subscriber

sethiay:
### Description
Adding helper functions for adding and notifying subscriber in download job and a function to fail while downloading. Also setup fake GCS object and bucket in job_test.go. I

NA

### Testing details
1. Manual - NA
2. Unit tests - Added unit tests for newly added functions
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -78,6 +78,8 @@ type JobStatus struct {
 // jobSubscriber represents a subscriber waiting on async download of job to
 // complete downloading at least till the subscribed offset.
 type jobSubscriber struct {
+	notificationC    chan<- JobStatus
+	subscribedOffset int64
 }
 
 func NewJob(object *gcs.MinObject, bucket gcs.Bucket, fileInfoCache *lru.Cache,
@@ -122,6 +124,47 @@ func (job *Job) init() {
 func (job *Job) Cancel() {
 	job.mu.Lock()
 	defer job.mu.Unlock()
+}
+
+// addSubscriber adds subscriber for download job and returns channel which is
+// notified when the download is completed at least till the subscribed offset
+// or in case of failure.
+//
+// Not concurrency safe and requires LOCK(job.mu)
+func (job *Job) addSubscriber(subscribedOffset int64) (notificationC <-chan JobStatus) {
+	subscriberC := make(chan JobStatus, 1)
+	job.subscribers.PushBack(jobSubscriber{subscriberC, subscribedOffset})
+	return subscriberC
+}
+
+// notifySubscribers notifies all the subscribers of download job in case of
+// error/cancellation or when download is completed till the subscribed offset.
+//
+// Not concurrency safe and requires LOCK(job.mu)
+func (job *Job) notifySubscribers() {
+	subItr := job.subscribers.Front()
+	for subItr != nil {
+		subItrValue := subItr.Value.(jobSubscriber)
+		nextSubItr := subItr.Next()
+		if job.status.Name == FAILED || job.status.Name == CANCELLED || job.status.Offset >= subItrValue.subscribedOffset {
+			subItrValue.notificationC <- job.status
+			close(subItrValue.notificationC)
+			job.subscribers.Remove(subItr)
+		}
+		subItr = nextSubItr
+	}
+}
+
+// errWhileDownloading changes the status of job to failed and notifies
+// subscribers about the download error.
+//
+// Acquires and releases LOCK(job.mu)
+func (job *Job) failWhileDownloading(downloadErr error) {
+	job.mu.Lock()
+	job.status.Err = downloadErr
+	job.status.Name = FAILED
+	job.notifySubscribers()
+	job.mu.Unlock()
 }
 
 // Download downloads object till the given offset if not already downloaded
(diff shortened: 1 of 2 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -124,6 +126,47 @@ func (job *Job) Cancel() {
 	defer job.mu.Unlock()
 }
 
+// addSubscriber adds subscriber for download job and returns channel which is
+// notified when the download is completed at least till the subscribed offset
+// or in case of failure.
+//
+// Not concurrency safe and requires LOCK(job.mu)
+func (job *Job) addSubscriber(subscribedOffset int64) (notificationC <-chan JobStatus) {
```
raj-prince, line 134: Ideally, we should give a jobSubscriber as an input Parameter to addSubscriber. And accordingly we should not return anything. What do you think?

internal/cache/file/downloader/job.go
```diff
@@ -124,6 +126,47 @@ func (job *Job) Cancel() {
...
+// notifySubscribers notifies all the subscribers of download job in case of
+// error/cancellation or when download is completed till the subscribed offset.
+//
+// Not concurrency safe and requires LOCK(job.mu)
+func (job *Job) notifySubscribers() {
+	subItr := job.subscribers.Front()
+	for subItr != nil {
+		subItrValue := subItr.Value.(jobSubscriber)
+		nextSubItr := subItr.Next()
+		if job.status.Name == FAILED || job.status.Name == CANCELLED || job.status.Offset >= subItrValue.subscribedOffset {
```
raj-prince, line 149: Just thinking out loud, the failed status here we will get after trying all the retry configured.

raj-prince: Can we simplify the looping over list?

Something like this,

``` go

for e := l.Front(); e != nil; e = e.Next() {
        v := e.Value
}
```

internal/cache/file/downloader/job.go
sethiay: Would it work if we change the name from addSubscriber to subscribe and keep input and return types same ?

internal/cache/file/downloader/job.go
```diff
@@ -124,6 +126,47 @@ func (job *Job) Cancel() {
...
+			subItrValue.notificationC <- job.status
+			close(subItrValue.notificationC)
+			job.subscribers.Remove(subItr)
+		}
+		subItr = nextSubItr
+	}
+}
+
+// errWhileDownloading changes the status of job to failed and notifies
```
raj-prince, line 158: nit: failWhileDownloading

internal/cache/file/downloader/job_test.go
```diff
@@ -16,34 +16,87 @@ package downloader
...
 }
 
 func init() { RegisterTestSuite(&jobTest{}) }
 
 func (jt *jobTest) SetUp(*TestInfo) {
 	locker.EnableInvariantsCheck()
-	var fileSpec data.FileSpec
-	jt.job = NewJob(nil, nil, nil, 200, fileSpec)
+
+	jt.bucket = fake.NewFakeBucket(timeutil.RealClock(), TestBucketName)
```
raj-prince, line 57: Should we move

internal/cache/file/downloader/job.go
sethiay: That will not work because we are removing elements while looping (if notified).

raj-prince (changes requested):
Please incorporate.

internal/cache/file/downloader/job.go
raj-prince: Still we can do this, the only thing we need to maintain a variable as we are doing above nextSubItr.

something like this,
``` go
var next *Element
for e := l.Front(); e != nil; e = next {
    next = e.Next()
    l.Remove(e)
}
```

internal/cache/file/downloader/job.go
raj-prince: Honestly speaking, I am more inclined towards addSubscriber(jobSubscriber). In this way we give flexibility to client  to give their notificationChannel.

sethiay: We decided both the options are fine. Decided to go with subscribe option as the other one increase code for test.

internal/cache/file/downloader/job_test.go
sethiay: We decided offline to keep it this way for now. (just use fake storage server bucket instead)

internal/cache/file/downloader/job_test.go
```diff
@@ -16,34 +16,92 @@ package downloader
...
+	}
+
+	fileSpec := data.FileSpec{
+		Path:     path.Join("./", storage.TestBucketName, DefaultObjectName),
+		Perm:     os.FileMode(0666),
+		OwnerUid: uint32(os.Getuid()),
+		OwnerGid: uint32(os.Getgid()),
+	}
+
+	jt.job = NewJob(&defaultMinObject, jt.bucket, jt.cache, 200, fileSpec)
```
raj-prince, line 97: Please create a constant for sequentialReadSizeMb.

raj-prince (approved):

PR #1453, opened 2023-10-18
Adding cache handle class - 1

raj-prince:
### Description
1. Adding cache_handle.go - this will be responsible to keep all the cache related handle, which will be used directly by the client of the cache.
2. Added Read() method - which will read the data from the cache. If data is already downloaded, this will serve the data from the cached path. If not, then it will wait for sequential-read and will not wait for random read to download the data.

TODO: I'll take as next PR.
1. Remove the check to validate the data using fileInfoCache, instead new GetJobStatus and handle based on that.
2. Add/handle job.Download call before reading the data from local downloaded path.
3. Add unit test for job.Download logic.

NA

### Testing details
1. Manual - NA
4. Unit tests - Added unit test.
5. Integration tests - NA


internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,114 @@
...
+		return errors.New(InvalidFileInfoCache)
+	}
+
+	return nil
+}
+
+// Read attempts to read the data from the cached location. This expects a
+// fileInfoCache entry for the current read request, and will wait to download
+// the requested chunk if it is not already present.
+func (fch *CacheHandle) Read(object *gcs.MinObject, bucket gcs.Bucket, offset uint64, dst []byte) (n int, err error) {
```
sethiay, line 70: Please also check if we can handle the below cases now/when download job is ready:
(We are going to check file cache in all cases irrespective of type of read).
1. When all sequential reads turns into random for a file, we need to cancel the ongoing job.
2. When random read and --download-file-random-read is not set.
3. Not required immediately but it's possible that random read doesn't give good performance, in that case, we may want to trigger the download but not wait for it

internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,114 @@
...
+	// Create fileInfoKey to get the existing fileInfoEntry in the cache.
+	fileInfoKey := data.FileInfoKey{ObjectName: object.Name, BucketName: bucket.Name(), BucketCreationTime: bucketCreationTime}
+	fileInfoKeyName, err := fileInfoKey.Key()
+	if err != nil {
+		return
+	}
+
+	// Get the existing fileInfo entry in the cache.
+	fileInfo := fch.fileInfoCache.LookUp(fileInfoKeyName)
+	if fileInfo == nil {
```
sethiay, line 88: it's possible that we are reading a file for the first time and in that case there is no entry in the file info cache. So, we should try to download first in that case.

internal/cache/file/cache_handle.go
sethiay, line 70: We also need to implement close method for this class ? (we should also close fileHandle in that call)

internal/cache/file/cache_handle.go
raj-prince: As per the current design, cacheFileHandler will have the responsibility to create and insert into the cache for the first time.

internal/cache/file/cache_handle.go
raj-prince: I agree, will add.

raj-prince: `When all sequential reads turns into random for a file, we need to cancel the ongoing job.` - As discussed, we will provide this method in CacheHandler, which will cancel the job for the specific fileInfoCache entry.

internal/cache/file/cache_handle_test.go
```diff
@@ -0,0 +1,208 @@
...
+func (t *cacheHandleTest) TestValidateCacheHandleWithNonNilMemberAttributes() {
+	err := t.ch.validateCacheHandle()
+
+	ExpectEq(nil, err)
+}
+
+func (t *cacheHandleTest) TestReadWithFileInfoKeyNotPresentInTheCache() {
+	dst := make([]byte, DstBufferLen)
+
+	// This will return the minGCSObject whose entry is in fileInfoCache.
```
sethiay, line 165: please correct the comment, IIUC fileinfokey is not present in fileinfocache

internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,126 @@
...
+	// Create fileInfoKey to get the existing fileInfoEntry in the cache.
+	fileInfoKey := data.FileInfoKey{ObjectName: object.Name, BucketName: bucket.Name()}
+	fileInfoKeyName, err := fileInfoKey.Key()
+	if err != nil {
+		return
+	}
+
+	// Get the existing fileInfo entry in the cache.
+	fileInfo := fch.fileInfoCache.LookUp(fileInfoKeyName)
+	if fileInfo == nil {
```
sethiay, line 84: Given that we are anyways geting fileInfo here, so should we also compare the generations ? if there is mismatch we can close the file handle too. That will also solve the problem of closing file handles in case of eviction.

internal/cache/file/cache_handle_test.go
```diff
@@ -0,0 +1,208 @@
...
+	minGCSObject.Name = TestObjectNameNotInFileInfoCache
+
+	_, err := t.ch.Read(minGCSObject, testBucket{BucketName: TestBucketName}, TestOffset, dst)
+
+	ExpectEq(InvalidFileInfo, err.Error())
+}
+
+func (t *cacheHandleTest) TestReadWithReadFromLocalCachedFilePath() {
+	file, err := createFileWithContent(TestFilePath, TestFileContent)
+	defer func() {
```
sethiay, line 176: you can write TearDown method to clean up files after every unit test.

internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,130 @@
...
+	fileInfoKey := data.FileInfoKey{ObjectName: object.Name, BucketName: bucket.Name(), BucketCreationTime: bucketCreationTime}
+	fileInfoKeyName, err := fileInfoKey.Key()
+	if err != nil {
+		return
+	}
+
+	// Get the existing fileInfo entry in the cache.
+	fileInfo := fch.fileInfoCache.LookUp(fileInfoKeyName)
+	if fileInfo == nil {
+		err = errors.New(InvalidFileInfo)
```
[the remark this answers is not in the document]
raj-prince: Logged such errors as warning. But return error is simple message so that caller function can handle properly.

internal/cache/file/cache_handle.go
raj-prince: Handled,  but disabled now, will enable when the download method is ready.

internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,232 @@
...
+
+// NewCacheHandle constructs and returns CacheHandle object.
+func NewCacheHandle(localFileHandle *os.File, fileDownloadJob *downloader.Job, fileInfoCache *lru.Cache, initialOffset int64) *CacheHandle {
+	return &CacheHandle{
+		fileHandle:               localFileHandle,
+		fileDownloadJob:          fileDownloadJob,
+		fileInfoCache:            fileInfoCache,
+		isSequential:             initialOffset == 0,
+		prevOffset:               initialOffset,
+		contributesToJobRefCount: true,
```
sethiay, line 77: Just writing down the cases that we need to handle here:
(We need to check file cache in all cases irrespective of type of read)
1. When --download-for-random-read=true:

    -  If read is sequential - wait for download & ignore the job reference count because we don't need to cancel the async job in this case once started.
    - if read is random - don't wait for download & ignore the job reference count because we don't need to cancel the async job in this case once started.
    - if read changes from sequential to random - no handling, follow the random read case.

1. When --download-for-random-read=false:

    - If read is sequential - wait for download & increase the job reference count.
    - if read is random - don't **download** (rather it should check the status of job here - so job should have a function GetJobStatus) & ignore the job reference count because we don't need to cancel the async job in this case.
    - if read changes from sequential to random - don't **download** & decrease the job reference count. if count becomes 0, on file cache handler side, cancel the job. _Note:_ _There is one corner case here - What if count becomes 0 at certain point and then again increases. In that case, file cachehandler should "Reset" the download job i.e. set to NOT_STARTED state._  (I will handle add Reset function in job if all this makes sense to you).
    
If you think handling all the cases above will make things very complicated then - My suggestion is to keep it simple like this:
1. If download-for-random-read = true:
    - if sequential read - wait for download.
    - if random read - don't wait for download.
    - if sequential read changes to random - follow random read.
 
 2. if download-for-random-read = false
    - if sequential read - wait for download.
    - if random read - don't **download** (still requires GetStatus)
    - if sequential read changes to random - follow random read.

internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,232 @@
...
+			n = 0
+			err = fmt.Errorf("while downloading job: %v", err)
+			return
+		}
+
+		// Todo - (raj-prince): to incorporate the newly introduced Invalid flag.
+		if jobStatus.Name == "Invalid" || jobStatus.Name == downloader.NOT_STARTED {
+			return 0, errors.New(InvalidFileDownloadJob)
+		} else {
+			if jobStatus.Offset < int64(requiredOffset) {
```
sethiay, line 154: nit: we can use "else if " here.

internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/data"
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/file/downloader"
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/lru"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
+	"github.com/googlecloudplatform/gcsfuse/internal/storage/gcs"
+)
+
+const (
+	InvalidFileHandle      = "destroy this cache_handle: invalid file handle"
+	InvalidFileDownloadJob = "destroy this cache_handle: invalid download job"
```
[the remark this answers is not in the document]
raj-prince: Invalid means, it can be nil as well as some wrong value. So, in some cases, we might point to INVALID job in that case too we will use this message.

InvalidFileDownloadJobErrMsg ? Or let me know - if I am missing something.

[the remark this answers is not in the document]
sethiay: SG, in that case, InvalidFileInfoCache = "file info cache is nil" -> InvalidFileInfoCache = "invalid file info cache"

internal/cache/file/cache_handle.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return
+	}
+
+	// We need to download the data till offset + len(dst), if not already.
+	bufferLen := uint64(len(dst))
+	requiredOffset := offset + bufferLen
+
+	// Also, need to make sure, it should not exceed the total file-size.
+	if requiredOffset > object.Size {
```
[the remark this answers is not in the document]
raj-prince: Not I think, as buffer len is fixed. Everytime we get the destination buffer of 1 MB.

internal/cache/file/cache_handle.go
raj-prince: Yeah, that's what we are doing here. Handling code will be random_reader.go.

It's simple only, just we need to ignore JobCancelling logic via jobRefCount when download-for-random=true.

`Note: There is one corner case here - What if count becomes 0 at certain point and then again increases. In that case, file cachehandler should "Reset" the download job i.e. set to NOT_STARTED state` - yeah sure.

internal/cache/file/cache_handle_test.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+func (t *cacheHandleTest) TestClose() {
+	err := t.ch.Close()
+	AssertEq(nil, err)
+
+	ExpectEq(nil, t.ch.fileHandle)
+}
+
+func (t *cacheHandleTest) TestIsSequentialWhenReadTypeIsSequential() {
```
sethiay, line 283: nit: TestIsSequentialWhenReadTypeIsNotSequential

sethiay (approved):

internal/cache/file/cache_handle.go
sethiay: As discussed offline, we are going with the simple logic

sethiay (approved):

PR #1454, opened 2023-10-19
Fix behavior and documentation of  max-retry-sleep flag

Tulsishah:
### Description
Fix behavior and documentation of max-retry-sleep flag.

- Replace max-retry-sleep with max-retry-duration
- Fix documentation for both the flag(max-retry-sleep and max-retry-duration)
- git grep MaxRetryDuration
     - only remains in flags file(As we did not implement this functionality for now)
NA

### Testing details
1. Manual - locally tested max-retry-sleep flag.
2. Unit tests - NA
3. Integration tests - NA


flags.go
```diff
@@ -225,8 +224,8 @@ func newApp() (app *cli.App) {
 
 			cli.DurationFlag{
 				Name:  "max-retry-duration",
-				Value: 30 * time.Second,
-				Usage: "The operation will be retried till the value of max-retry-duration.",
+				Value: -1 * time.Second,
+				Usage: "The operation will be retried till the value of max-retry-duration. This flag is currently unused.",
```
mustvicky, line 228: Should we also log a warning when this flag is set? (now that the default is changed, we can identify when the value is different than -1 and log a warning)

mustvicky (commented):
Just one comment, rest all looks good to me.

flags.go
Tulsishah: Yes, I'll raise separate PR for that for info logs

Tulsishah:
> Just one comment, rest all looks good to me.

Will raise separate PR for that.

mustvicky (approved):

flags.go
```diff
@@ -195,9 +195,8 @@ func newApp() (app *cli.App) {
 			cli.DurationFlag{
 				Name:  "max-retry-sleep",
 				Value: time.Minute,
```
sethiay, line 197: Should we change the default Value to 30s here? Because earlier it was 30s and we haven't got any issues with that.

flags.go
Tulsishah: Yeah, I was thinking the same, let's do the same as we did for that debug flag.
`And if we are not warning then we can keep 30s as default too.` --> -1 will make sense I guess as we are not using this flag anywhere.

sethiay: okay, let's just use "This flag is currently unused." as description

sethiay (approved):

PR #1456, opened 2023-10-19
Add create file util , update fileinfocache, remove ownership, bucket creation time

sethiay:
### Description
Changes involve:
1. Added create file util to create file in the file cache at the time of downloading.
2. Helper to update file info cache at the time of downloading.
3. Realized from the design that we don't need to put specific uid and gid for files in cache. So removed that from file spec. Haven't deleted file spec in case we need to add more parameters in future.
4. Ignored check in fileinfokey for bucket creation time till we support bucket creation time via metadata call.

NA

### Testing details
1. Manual - Ran unit tests.
5. Unit tests - Added unit tests wherever applicable.
6. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -128,7 +128,7 @@ func (job *Job) Cancel() {
 
 // subscribe adds subscriber for download job and returns channel which is
 // notified when the download is completed at least till the subscribed offset
-// or in case of failure.
+// or in case of failure and cancellation.
 //
 // Not concurrency safe and requires LOCK(job.mu)
 func (job *Job) subscribe(subscribedOffset int64) (notificationC <-chan JobStatus) {
@@ -177,7 +177,8 @@ func (job *Job) updateFileInfoCache() (err error) {
 	}
 	fileInfoKeyName, err := fileInfoKey.Key()
 	if err != nil {
-		err = fmt.Errorf(fmt.Sprintf("error while calling FileInfoKey.Key() for %s %v", fileInfoKeyName, err))
+		err = fmt.Errorf(fmt.Sprintf("init: error while calling FileInfoKey.Key() for bucket %s and object %s %v",
+			fileInfoKey.BucketName, fileInfoKey.ObjectName, err))
 		return
 	}
 
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -166,6 +166,36 @@ func (job *Job) failWhileDownloading(downloadErr error) {
 	job.mu.Unlock()
 }
 
+// updateFileInfoCache updates the file info cache with latest offset downloaded
+// by job. Returns error in case of failure.
+//
+// Not concurrency safe and requires LOCK(job.mu)
+func (job *Job) updateFileInfoCache() (err error) {
+	fileInfoKey := data.FileInfoKey{
+		BucketName: job.bucket.Name(),
+		ObjectName: job.object.Name,
+	}
+	fileInfoKeyName, err := fileInfoKey.Key()
+	if err != nil {
+		err = fmt.Errorf(fmt.Sprintf("error while calling FileInfoKey.Key() for %s %v", fileInfoKeyName, err))
+		return
+	}
+
+	updatedFileInfo := data.FileInfo{
+		Key: fileInfoKey, ObjectGeneration: job.object.Generation,
+		FileSize: job.object.Size, Offset: uint64(job.status.Offset),
+	}
+
+	// To-Do(raj-prince): We should not call normal insert here as that internally
+	// changes the LRU element which is undesirable given this is not user access.
+	_, err = job.fileInfoCache.Insert(fileInfoKeyName, updatedFileInfo)
+	if err != nil {
+		err = fmt.Errorf(fmt.Sprintf("error while inserting updatedFileInfo to the FileInfoCache %s: %v", updatedFileInfo.Key, err))
+		return
+	}
+	return
+}
+
 // Download downloads object till the given offset if not already downloaded
 // and waits for download if waitForDownload is true.
 // ToDo (sethiay): Implement this function.
--- a/internal/cache/file/downloader/job_test.go
+++ b/internal/cache/file/downloader/job_test.go
@@ -239,6 +239,8 @@ func (jt *jobTest) Test_updateFileInfoCache_UpdateEntry() {
 	ExpectEq(jt.job.object.Size, fileInfo.FileSize)
 }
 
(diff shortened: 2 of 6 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -124,6 +126,76 @@ func (job *Job) Cancel() {
 	defer job.mu.Unlock()
 }
 
+// subscribe adds subscriber for download job and returns channel which is
+// notified when the download is completed at least till the subscribed offset
+// or in case of failure.
```
raj-prince, line 131: Please correct the comment, we also notify the user in case of  Job get cancelled.

internal/cache/util/util_test.go
```diff
@@ -0,0 +1,165 @@
...
+const FileName = "foo.txt"
+
+func init() { RegisterTestSuite(&utilTest{}) }
+
+func (ut *utilTest) SetUp(*TestInfo) {
+	operations.RemoveDir(FileDir)
+	ut.flag = os.O_RDWR
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		panic(fmt.Errorf("error while finding home directory: %v", err))
```
raj-prince, line 49: Can we use t.Fatal() or t.FailNow() instead of panic()?
Ref - https://blog.devgenius.io/write-better-golang-code-by-avoiding-panic-in-tests-f6a7a22937e6#:~:text=So%2C%20to%20summarize%2C%20you',to%20writing%20better%20Go%20code.

internal/cache/util/util_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	ut.fileSpec = data.FileSpec{
+		Path: path.Join(homeDir, FileDir, FileName),
+		Perm: os.FileMode(0644),
```
raj-prince, line 53: Let's create a const for default FileMode.

internal/cache/util/util_test.go
```diff
@@ -0,0 +1,165 @@
...
+	ExpectTrue(strings.Contains(err.Error(), "bad file descriptor"))
+	ExpectEq(nil, file.Close())
+}
+
+func (ut *utilTest) Test_CreateFile_ReadWriteFile() {
+	ut.flag = os.O_RDWR
+
+	file, err := CreateFile(ut.fileSpec, ut.flag)
+
+	ut.assertFileAndDirCreation(file, err)
```
raj-prince, line 116: We should also read from the file to check Read behaviour.

As only os.O_WRONLY also exists.

internal/cache/util/util_test.go
```diff
@@ -0,0 +1,165 @@
...
+func (ut *utilTest) Test_CreateFile_FilePerm0755() {
+	ut.fileSpec.Perm = os.FileMode(0755)
+
+	file, err := CreateFile(ut.fileSpec, ut.flag)
+
+	ut.assertFileAndDirCreation(file, err)
+	ExpectEq(nil, file.Close())
+}
+
+func (ut *utilTest) Test_CreateFile_FilePerm0644() {
```
raj-prince, line 132: Can we add a test case of relative path?

internal/cache/util/util_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/data"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/operations"
+	. "github.com/jacobsa/ogletest"
+)
+
+func TestUtil(t *testing.T) { RunTests(t) }
+
+type utilTest struct {
```
raj-prince, line 32: Can we add a test-case for relative path?

internal/cache/file/downloader/job.go
```diff
@@ -166,6 +166,36 @@ func (job *Job) failWhileDownloading(downloadErr error) {
 	job.mu.Unlock()
 }
 
+// updateFileInfoCache updates the file info cache with latest offset downloaded
+// by job. Returns error in case of failure.
+//
+// Not concurrency safe and requires LOCK(job.mu)
+func (job *Job) updateFileInfoCache() (err error) {
+	fileInfoKey := data.FileInfoKey{
```
raj-prince, line 174: As discussed let's make the fileInfoKeyName as a member variable of Job, so that we don't need to calculate every time.

internal/cache/file/downloader/job_test.go
```diff
@@ -209,3 +208,73 @@ func (jt *jobTest) Test_failWhileDownloading() {
...
+	// confirm fileInfoCache is updated with new offset.
+	lookupResult := jt.cache.LookUp(fileInfoKeyName)
+	ExpectFalse(lookupResult == nil)
+	fileInfo = lookupResult.(data.FileInfo)
+	ExpectEq(1, fileInfo.Offset)
+	ExpectEq(jt.job.object.Generation, fileInfo.ObjectGeneration)
+	ExpectEq(jt.job.object.Size, fileInfo.FileSize)
+}
+
+func (jt *jobTest) Test_updateFileInfoCache_InsertNew() {
```
raj-prince, line 242: Thinking loud - later when we start using overrideKey of lru, inserting a new entry will return an error.

raj-prince (changes requested):
A few comments, please incorporate.

internal/cache/util/util_test.go
sethiay: I also thought of using that but I couldn't find how a similar option in https://github.com/jacobsa/ogletest. Let me know if you know of any similar way.

internal/cache/file/downloader/job.go
sethiay: Even if we make fileInfoKeyName as member of Job class, we will have to create fileInfoKey in this function every time to insert into the fileInfoCache.
So either we have to keep both fileInfoKey and fileInfoKeyName as members or keep fileInfoKey as member and get fileInfoKeyName everytime in this function. 
To keep things simple, we can go ahead with the current approach ?

internal/cache/file/downloader/job_test.go
sethiay: Added a comment to remove this test when we support updateValue in lru.

internal/cache/file/downloader/job.go
raj-prince: Okay, for now we can skip. We will take it separately to optimize fileInfoKey creation part globally.

internal/cache/util/util_test.go
raj-prince: Ohh, we can't take *testing.T as input in SetUp method as per the current implementation. Then we don't have any option other than panic().

raj-prince (approved):

PR #1458, opened 2023-10-20
Read cache job 4 - Adding download async job & test cases

sethiay:
### Description
Adding downloadObjectAsync method to download object into a file in cache sequentially. Their can only be one go routine running this method at a time for a download job object. 
This method can be cancelled by job.Cancel method. I will add job.Cancel and job.Download in the next PR along with test cases for job.Download and job.Cancel.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added unit tests.
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -210,7 +210,7 @@ func (job *Job) updateFileInfoCache() (err error) {
 // Acquires and releases LOCK(job.mu)
 func (job *Job) downloadObjectAsync() {
 	// Create and open cache file for writing object into it.
-	cacheFile, err := util.CreateFile(job.fileSpec, os.O_RDWR)
+	cacheFile, err := util.CreateFile(job.fileSpec, os.O_WRONLY)
 	defer func(cacheFile *os.File) {
 		err = cacheFile.Close()
 		if err != nil {
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -122,17 +122,6 @@ func (job *Job) init() {
 	job.cancelCtx, job.cancelFunc = context.WithCancel(context.Background())
 }
 
-// cancel is helper function for job.Cancel().
-//
-// Not concurrency safe and requires LOCK(job.mu)
-func (job *Job) cancel() {
-	if job.status.Name == DOWNLOADING && job.cancelFunc != nil {
-		job.cancelFunc()
-	}
-	job.status.Name = CANCELLED
-	job.notifySubscribers()
-}
-
 // Cancel changes the state of job to cancelled and cancels the async download
 // job if there. Also, notifies the subscribers of job if any.
 // ToDo (sethiay): Implement this function.
@@ -242,10 +231,6 @@ func (job *Job) downloadObjectAsync() {
 	for {
 		select {
 		case <-job.cancelCtx.Done():
-			job.mu.Lock()
-			job.cancelFunc = nil
-			job.cancel()
-			job.mu.Unlock()
 			return
 		default:
 			if start < end {
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -39,8 +39,7 @@ const (
 	CANCELLED   jobStatusName = "CANCELLED"
 )
 
-const MiB = 1024 * 1024
-const ReadChunkSize = 8 * MiB
+const ReadChunkSize = 8 * util.MiB
 
 // Job downloads the requested object from GCS into the specified local file
 // path with given permissions and ownership.
@@ -60,6 +59,7 @@ type Job struct {
 	/////////////////////////
 
 	// Represents the current status of Job.
+	// status.Offset means data in cache is present in range [0, status.offset)
 	status JobStatus
 
 	// list of subscribers waiting on async download.
@@ -211,6 +211,11 @@ func (job *Job) updateFileInfoCache() (err error) {
 func (job *Job) downloadObjectAsync() {
 	// Create and open cache file for writing object into it.
 	cacheFile, err := util.CreateFile(job.fileSpec, os.O_WRONLY)
+	if err != nil {
+		err = fmt.Errorf("downloadObjectAsync: error in creating cache file: %v", err)
+		job.failWhileDownloading(err)
+		return
+	}
 	defer func(cacheFile *os.File) {
 		err = cacheFile.Close()
 		if err != nil {
@@ -218,15 +223,11 @@ func (job *Job) downloadObjectAsync() {
 			job.failWhileDownloading(err)
 		}
 	}(cacheFile)
-	if err != nil {
-		err = fmt.Errorf("downloadObjectAsync: error in creating cache file: %v", err)
-		job.failWhileDownloading(err)
-	}
 
 	var newReader io.ReadCloser
 	var start, end, sequentialReadSize, newReaderLimit int64
 	end = int64(job.object.Size)
-	sequentialReadSize = int64(job.sequentialReadSizeMb) * MiB
+	sequentialReadSize = int64(job.sequentialReadSizeMb) * util.MiB
 
 	for {
 		select {
@@ -300,6 +301,7 @@ func (job *Job) downloadObjectAsync() {
 
 // Download downloads object till the given offset if not already downloaded
 // and waits for download if waitForDownload is true.
+// The caller of this method should not read from cache if job status is FAILED.
 // ToDo (sethiay): Implement this function.
 //
 // Acquires and releases LOCK(job.mu)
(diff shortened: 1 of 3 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -197,6 +203,101 @@ func (job *Job) updateFileInfoCache() (err error) {
...
+// downloadObjectAsync downloads the backing GCS object into a file as part of
+// file cache using NewReader method of gcs.Bucket.
+//
+// Note: There can only be one async download running for a job at a time.
+// Acquires and releases LOCK(job.mu)
+func (job *Job) downloadObjectAsync() {
+	// Create and open cache file for writing object into it.
+	cacheFile, err := util.CreateFile(job.fileSpec, os.O_WRONLY)
+	defer func(cacheFile *os.File) {
+		err = cacheFile.Close()
```
raj-prince, line 215: We shouldn't directly call cacheFile.Close(). It may panic if it's nil.

internal/cache/file/downloader/job.go
```diff
@@ -197,6 +203,101 @@ func (job *Job) updateFileInfoCache() (err error) {
...
+		if err != nil {
+			err = fmt.Errorf("downloadObjectAsync: error while closing cache file: %v", err)
+			job.failWhileDownloading(err)
+		}
+	}(cacheFile)
+	if err != nil {
+		err = fmt.Errorf("downloadObjectAsync: error in creating cache file: %v", err)
+		job.failWhileDownloading(err)
```
raj-prince, line 223: Also, seems we need to return after failing the job?

internal/cache/file/downloader/job.go
```diff
@@ -197,6 +203,101 @@ func (job *Job) updateFileInfoCache() (err error) {
...
+					err = fmt.Errorf(fmt.Sprintf("downloadObjectAsync: error while seeking file handle, seek %d: %v", start, err))
+					job.failWhileDownloading(err)
+					return
+				}
+
+				// copy the contents from NewReader to cache file.
+				_, readErr := io.CopyN(cacheFile, newReader, maxRead)
+				if readErr != nil && readErr != io.EOF {
+					err = fmt.Errorf("downloadObjectAsync: error at the time of copying content to cache file %v", readErr)
+					job.failWhileDownloading(err)
```
raj-prince, line 269: Thinking loud - should we read from the failed Job?

If yes, how are we validating things are correct when we get any failure while executing io.CopyN?

internal/cache/file/downloader/job.go
```diff
@@ -197,6 +203,101 @@ func (job *Job) updateFileInfoCache() (err error) {
...
+					return
+				}
+				start += maxRead
+				if readErr == io.EOF {
+					newReader = nil
+				}
+
+				job.mu.Lock()
+				job.status.Offset = start
```
raj-prince, line 278: Thinking aloud - start is the position from where we start downloading?

If yes then we need document about the Offset, that it keeps the downloaded data till Offset - 1.

internal/cache/file/downloader/job.go
sethiay: no, the client shouldn't read when job is in failed state. This can hamper the performance in worst case incase the failure occurs at the end of download job. However, I want to keep things simple here. Also, this problem will be handled when and if add responsibility of DownloadJobManager to Retry Failed/Cancelled jobs.
What do you think ?

internal/cache/file/downloader/job.go
sethiay: offset stores length up to which data is downloaded. So, an offset of 5 MiB means 5 MiB of data is downloaded.

start is the position(starts from 0) from which the next download should start. So, initially start = 0, limit = 8MiB, offset = 0. When 8 MiB is downloaded, both start and offset becomes 8MiB and limit becomes 16 MiB.

Does the above need to change ? Maybe I am missing something here?

internal/cache/file/downloader/job.go
```diff
@@ -36,6 +39,9 @@ const (
 	CANCELLED   jobStatusName = "CANCELLED"
 )
 
+const MiB = 1024 * 1024
```
raj-prince, line 42: We can keep these constants in util so that different component can use it.

internal/cache/file/downloader/job_test.go
```diff
@@ -59,46 +65,69 @@ func (jt *jobTest) SetUp(*TestInfo) {
...
 
-	fileSpec := data.FileSpec{
-		Path: path.Join("./", storage.TestBucketName, DefaultObjectName),
-		Perm: os.FileMode(0666),
-	}
+func (jt *jobTest) initJobTest(objectName string, objectContent []byte, sequentialReadSize int32, lruCacheSize uint64) {
+	ctx := context.Background()
+	objects := map[string][]byte{objectName: objectContent}
+	err := storageutil.CreateObjects(ctx, jt.bucket, objects)
+	ExpectEq(nil, err)
```
raj-prince, line 100: I think, we should assert.

internal/cache/file/downloader/job_test.go
```diff
@@ -280,3 +309,90 @@ func (jt *jobTest) Test_updateFileInfoCache_Fail() {
...
 	ExpectTrue(lookupResult == nil)
 }
+
+// Note: We can't test Test_downloadObjectAsync_MoreThanSequentialReadSize as
+// the fake storage bucket/server in the testing environment doesn't support
+// reading ranges (start and limit in NewReader call)
+func (jt *jobTest) Test_downloadObjectAsync_LessThanSequentialReadSize() {
+	// Create new object in bucket and create new job for it.
+	objectName := "path/in/gcs/foo.txt"
+	objectSize := 50 * MiB
```
raj-prince, line 319: ObjectSize = 50 MiB but content is "t". I think we can reduce the object Size further so that objectSize and content should match?

internal/cache/file/downloader/job_test.go
```diff
@@ -280,3 +309,90 @@ func (jt *jobTest) Test_updateFileInfoCache_Fail() {
...
+	ExpectTrue(reflect.DeepEqual(jobStatus, jt.job.status))
+	// verify file is downloaded
+	jt.verifyFile(objectContent)
+	// Verify fileInfoCache update
+	jt.verifyFileInfoEntry(uint64(objectSize))
+}
+
+func (jt *jobTest) Test_downloadObjectAsync_LessThanChunkSize() {
+	objectName := "path/in/gcs/foo.txt"
+	objectSize := 2 * MiB
```
raj-prince, line 339: Same as above comment - we should make the test case realistic. objectSize == len(objectContent)?

internal/cache/file/downloader/job_test.go
```diff
@@ -280,3 +309,90 @@ func (jt *jobTest) Test_updateFileInfoCache_Fail() {
...
+	objectName := "path/in/gcs/foo.txt"
+	objectSize := 50 * MiB
+	objectContent := bytes.Repeat([]byte("t"), objectSize)
+	jt.initJobTest(objectName, objectContent, 100, uint64(objectSize-1))
+
+	// start download
+	jt.job.downloadObjectAsync()
+
+	// check job failed
+	jt.job.mu.Lock()
```
raj-prince, line 393: Why job.mu lock required?

raj-prince (changes requested):
Few comments.

internal/cache/file/downloader/job.go
raj-prince: Got it. But I was thinking offset as index till data has been downloaded. In general also offset treated as index not as length. No, the only thing we need document about the offset properly, as it contains data [0, offset).

internal/cache/file/downloader/job.go
raj-prince: Yeah, sounds good.
Should we write it somewhere as comment, please don't read data for the failed/cancelled Job?

internal/cache/file/downloader/job_test.go
sethiay: Used random bytes instead of just 't' * size.

internal/cache/file/downloader/job.go
```diff
@@ -197,8 +203,105 @@ func (job *Job) updateFileInfoCache() (err error) {
...
+				job.mu.Unlock()
+				// change status of job in case of error while updating file cache.
+				if err != nil {
+					job.failWhileDownloading(err)
+					return
+				}
+			} else {
+				job.mu.Lock()
+				job.status.Name = COMPLETED
+				job.notifySubscribers()
```
raj-prince, line 294: Should we notify everyone in case of job-completion? This is just to handle, if someone subscribes for offset greater than object-size.

sethiay: they will get error if they try to subscribe for offset greater than object size. That check is in Download method in next PR.

raj-prince (approved):

PR #1460, opened 2023-10-22
Adding download job manager 

sethiay:
### Description
Adding file download job manager.

NA

### Testing details
1. Manual - NA
2. Unit tests - added unit tests for new functions
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/downloader.go
+++ b/internal/cache/file/downloader/downloader.go
@@ -33,17 +33,35 @@ func getObjectPath(bucketName string, objectName string) string {
 // JobManager is responsible for maintaining, getting and removing file download
 // jobs.
 type JobManager struct {
-	fileInfoCache        *lru.Cache
-	perm                 os.FileMode
-	cacheLocation        string
-	sequentialReadSizeMb int32
+	/////////////////////////
+	// Constant data
+	/////////////////////////
 
+	// filePerm is passed to Job created by JobManager. filePerm decides the
+	// permission of file in cache created by Job.
+	filePerm os.FileMode
+	// cacheLocation is the path to directory where cache files should be created.
+	cacheLocation string
+	// sequentialReadSizeMb is passed to Job created by JobManager, and it decides
+	// the size of GCS read requests by Job at the time of downloading object to
+	// file in cache.
+	sequentialReadSizeMb int32
+	fileInfoCache        *lru.Cache
+
+	/////////////////////////
+	// Mutable state
+	/////////////////////////
+
+	// jobs contains the reference to Job for a given object path. Object path is
+	// concatenation of bucket name, "/", and object name. e.g. object path for an
+	// object named "a/b/foo.txt" in bucket named "test_bucket" would be
+	// "test_bucket/a/b/foo.txt"
 	jobs map[string]*Job
 	mu   locker.Locker
 }
 
-func NewJobManager(fileInfoCache *lru.Cache, perm os.FileMode, cacheLocation string, sequentialReadSizeMb int32) (jm *JobManager) {
-	jm = &JobManager{fileInfoCache: fileInfoCache, perm: perm,
+func NewJobManager(fileInfoCache *lru.Cache, filePerm os.FileMode, cacheLocation string, sequentialReadSizeMb int32) (jm *JobManager) {
+	jm = &JobManager{fileInfoCache: fileInfoCache, filePerm: filePerm,
 		cacheLocation: cacheLocation, sequentialReadSizeMb: sequentialReadSizeMb}
 	jm.mu = locker.New("JobManager", func() {})
 	jm.jobs = make(map[string]*Job)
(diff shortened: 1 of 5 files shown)


internal/cache/file/downloader/downloader.go
```diff
@@ -0,0 +1,90 @@
...
+	fileInfoCache        *lru.Cache
+	perm                 os.FileMode
+	cacheLocation        string
+	sequentialReadSizeMb int32
+
+	jobs map[string]*Job
+	mu   locker.Locker
+}
+
+func NewJobManager(fileInfoCache *lru.Cache, perm os.FileMode, cacheLocation string, sequentialReadSizeMb int32) (jm *JobManager) {
```
raj-prince, line 45: perm - is FilePerms in the ServerConfig, right?
If this is the case, then let's rename it filePerm. Also, please write a one line comment, what is it used for.

internal/cache/file/downloader/downloader.go
```diff
@@ -0,0 +1,90 @@
...
+
+// JobManager is responsible for maintaining, getting and removing file download
+// jobs.
+type JobManager struct {
+	fileInfoCache        *lru.Cache
+	perm                 os.FileMode
+	cacheLocation        string
+	sequentialReadSizeMb int32
+
+	jobs map[string]*Job
```
raj-prince, line 41: Let's document about jobs, specially, what's the key?

internal/cache/file/downloader/downloader_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/googlecloudplatform/gcsfuse/internal/storage"
+	"github.com/googlecloudplatform/gcsfuse/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/operations"
+	. "github.com/jacobsa/ogletest"
+	"golang.org/x/net/context"
+)
+
+var cacheLocation string = path.Join(os.Getenv("HOME"), "cache/location")
+
+const DefaultPerm = os.FileMode(0644)
```
raj-prince, line 36: Let's move this constant to util, since it will be used at multiple places.

internal/cache/file/downloader/downloader_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	dt.jm.mu.Lock()
+	defer dt.jm.mu.Unlock()
+	dt.verifyJob(job, &dt.object, dt.bucket, dt.jm.sequentialReadSizeMb)
+}
+
+func (dt *downloaderTest) Test_GetJob_Existing() {
+	// first create new job
+	expectedJob := dt.jm.GetJob(&dt.object, dt.bucket)
+	dt.jm.mu.Lock()
```
raj-prince, line 121: Just curious, why are we taking jobManager resource lock, before calling verifyJob? We are not doing anything related to job-manager resources?

internal/cache/file/downloader/downloader_test.go
```diff
@@ -0,0 +1,224 @@
...
+	// remove the job
+	dt.jm.RemoveJob(dt.object.Name, dt.bucket.Name())
+
+	// verify no job existing
+	dt.jm.mu.Lock()
+	objectPath := getObjectPath(dt.bucket.Name(), dt.object.Name)
+	_, ok := dt.jm.jobs[objectPath]
+	ExpectFalse(ok)
+	dt.jm.mu.Unlock()
+	// verify the job is cancelled
```
raj-prince, line 193: nit: cancelled -> invalid

internal/cache/file/downloader/job_test.go
```diff
@@ -249,173 +215,174 @@ func (jt *jobTest) Test_failWhileDownloading() {
...
 
 	// check job failed
-	jt.job.mu.Lock()
-	defer jt.job.mu.Unlock()
-	AssertEq(FAILED, jt.job.status.Name)
-	AssertTrue(strings.Contains(jt.job.status.Err.Error(), "size of the entry is more than the cache's maxSize"))
+	dt.job.mu.Lock()
+	defer dt.job.mu.Unlock()
+	AssertEq(FAILED, dt.job.status.Name)
+	AssertEq(ReadChunkSize, dt.job.status.Offset)
```
raj-prince, line 373: Seems we have removed this in the previous PR? (job.Download)

raj-prince (changes requested):
Few minor comments.

internal/cache/file/downloader/downloader_test.go
sethiay: We are passing dt.jm.sequentialReadSizeMb

internal/cache/file/downloader/downloader.go
sethiay: it will not be exactly filePerms in ServerConfig. The filePerms in ServerConfig defines the permission for file in GCS bucket. This defines the permission of file in cache. We can't keep the filePerms in ServerConfig same because the file in cache will always be created by gcsfuse daemon. So we will keep a default value which restricts any other user from accessing it if allow-other is not passed at the time of mounting. If allow-other is passed then we will allow others to access it.
but the type of both are same, so I changed the name to filePerm and added documentation.

internal/cache/file/downloader/job_test.go
```diff
@@ -621,18 +621,8 @@ func (dt *downloaderTest) Test_Cancel_WhenDownlooading() {
 	AssertEq(nil, err)
```
raj-prince, line 621: Do we need to wait for some time? We can directly cancel after the first job.Download call.

raj-prince (approved):

raj-prince (approved):

PR #1463, opened 2023-10-25
Fix perf test

ashmeenkaur:
### Description
Perf tests were failing because we were using job_start key while fetching the metrics but still using the old fio version which did not support "job_start" key. 
In this PR:
- moved installing requirements before switching to master.
- building fio from source to support job_start key.

NA

### Testing details
Ran perf tests
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
+++ b/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
@@ -68,6 +68,8 @@ function install_requirements() {
   pip install google-cloud-vision
   pip install google-api-python-client
   pip install prettytable
+  # install libaio as fio has a dependency on libaio
+  sudo apt-get install libaio-dev
   # We are building fio from source because of issue: https://github.com/axboe/fio/issues/1640.
   # The fix is not currently released in a package as of 20th Oct, 2023.
   # TODO: install fio via package when release > 3.35 is available.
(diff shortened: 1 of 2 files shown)


perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
```diff
@@ -58,10 +58,36 @@ function execute_perf_test() {
...
+  ./configure && make && sudo make install
+  fio --version
+  cd "${KOKORO_ARTIFACTS_DIR}/github/gcsfuse"
+}
+
 # execute perf tests.
 if [[ "$perfTestStr" == *"$EXECUTE_PERF_TEST_LABEL"* ]];
 then
  # Executing perf tests for master branch
+ install_requirements
```
gargnitingoogle, line 90: @ashmeenkaur Ideally, this call should have gone into the function execute_perf_test (as that's where the equivalent code was originally code i.e. perfmetrics/scripts/presubmit/run_load_test_on_presubmit.sh from) . But I guess this function isn't idempotent and hence placed here to ensure that it's call only once at the most ?

ashmeenkaur: Agreed but I didn't see a need to install the requirements twice. Moreover, now we are building fio from source, which would unnecessarily clone and build fio repo twice and was not required.

PR #1475, opened 2023-10-27
Making sync call in the last for Fio sequential write workflow

raj-prince:
### Description
Currently, in the sequential write operation of FIO workload `SyncFile` call happens after every `WriteFile` call, which shows less bandwidth for write flow.

Now, for sequential workflow, `SyncFile` call will happen in the last.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


raj-prince:
Stale.

PR #1487, opened 2023-10-31
add configs for log rotation

ashmeenkaur:
### Description
Added configs for log rotation:
```
logging:
  log-rotate:
    max-size-in-mb (uint32): maximum size in megabytes of the log file before it gets rotated. It defaults to 200 megabytes.
    backup-count (uint32): maximum number of old log files to retain after they are log rotated. It defaults to 3.
    compress (bool):  determines if the rotated log files should be compressed using gzip. It defaults to true.
```

NA

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/internal/config/mount_config.go
+++ b/internal/config/mount_config.go
@@ -19,9 +19,10 @@ type WriteConfig struct {
 }
 
 type LogConfig struct {
-	Severity LogSeverity `yaml:"severity"`
-	Format   string      `yaml:"format"`
-	FilePath string      `yaml:"file-path"`
+	Severity  LogSeverity `yaml:"severity"`
+	Format    string      `yaml:"format"`
+	FilePath  string      `yaml:"file-path"`
+	LogRotate LogRotate   `yaml:"log-rotate"`
 }
 
 type MountConfig struct {
@@ -29,11 +30,29 @@ type MountConfig struct {
 	LogConfig   `yaml:"logging"`
 }
 
+type LogRotate struct {
+	MaxSizeInMB uint32 `yaml:"max-size-in-mb"`
+	MaxDays     uint32 `yaml:"max-days"`
+	BackupCount uint32 `yaml:"backup-count"`
+	Compress    bool   `yaml:"compress"`
+}
+
+func DefaultLogRotateConfig() LogRotate {
+	return LogRotate{
+		MaxSizeInMB: 200,
+		MaxDays:     28,
+		BackupCount: 3,
+		Compress:    true,
+	}
+}
+
 func NewMountConfig() *MountConfig {
 	mountConfig := &MountConfig{}
 	mountConfig.LogConfig = LogConfig{
 		// Making the default severity as INFO.
 		Severity: INFO,
+		// Setting default values of log rotate config.
+		LogRotate: DefaultLogRotateConfig(),
 	}
 	return mountConfig
 }
--- a/internal/config/mount_config.go
+++ b/internal/config/mount_config.go
@@ -32,7 +32,6 @@ type MountConfig struct {
 
 type LogRotate struct {
 	MaxSizeInMB uint32 `yaml:"max-size-in-mb"`
-	MaxDays     uint32 `yaml:"max-days"`
 	BackupCount uint32 `yaml:"backup-count"`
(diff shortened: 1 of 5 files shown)


internal/config/mount_config.go
```diff
@@ -19,21 +19,38 @@ type WriteConfig struct {
...
+	FilePath  string      `yaml:"file-path"`
+	LogRotate LogRotate   `yaml:"log-rotate"`
 }
 
 type MountConfig struct {
 	WriteConfig `yaml:"write"`
 	LogConfig   `yaml:"logging"`
 }
 
+type LogRotate struct {
```
gargnitingoogle, line 33: Rename this to LogRotateConfig for consistency with other structs.

internal/config/mount_config.go
```diff
@@ -19,21 +19,38 @@ type WriteConfig struct {
...
+	MaxSizeInMB uint32 `yaml:"max-size-in-mb"`
+	BackupCount uint32 `yaml:"backup-count"`
```
gargnitingoogle, line 35: Please add short descriptions here for each as headers.

internal/config/mount_config.go
```diff
@@ -19,21 +19,38 @@ type WriteConfig struct {
...
+	Compress    bool   `yaml:"compress"`
+}
+
+func DefaultLogRotateConfig() LogRotate {
+	return LogRotate{
+		MaxSizeInMB: 200,
```
gargnitingoogle, line 41: 200, 3, true should be defined as const values at the top of the file.

e.g. 
```golang
const MaxLogSizeInMB int = 200
```

internal/config/yaml_parser_test.go
```diff
@@ -81,6 +84,9 @@ func (t *YamlParserTest) TestReadConfigFile_ValidConfig() {
 	AssertEq(ERROR, mountConfig.LogConfig.Severity)
 	AssertEq("/tmp/logfile.json", mountConfig.LogConfig.FilePath)
 	AssertEq("text", mountConfig.LogConfig.Format)
+	AssertEq(100, mountConfig.LogConfig.LogRotate.MaxSizeInMB)
+	AssertEq(5, mountConfig.LogConfig.LogRotate.BackupCount)
+	AssertEq(false, mountConfig.LogConfig.LogRotate.Compress)
```
gargnitingoogle, line 89: Without documentation, it's not clear why this case should fail. Please either add a comment here to say why, or headers where the config are defined, or at both places.

internal/config/yaml_parser_test.go
```diff
@@ -81,6 +84,9 @@ func (t *YamlParserTest) TestReadConfigFile_ValidConfig() {
 	AssertEq(ERROR, mountConfig.LogConfig.Severity)
 	AssertEq("/tmp/logfile.json", mountConfig.LogConfig.FilePath)
 	AssertEq("text", mountConfig.LogConfig.Format)
+	AssertEq(100, mountConfig.LogConfig.LogRotate.MaxSizeInMB)
```
gargnitingoogle, line 87: Is there a lower/upper limit on `mountConfig.LogConfig.LogRotate.MaxSizeInMB` or on `mountConfig.LogConfig.LogRotate.BackupCount` or on combinations of those two ? If yes, we should add tests for those.

internal/logger/logger.go
```diff
@@ -166,6 +168,7 @@ type loggerFactory struct {
 	sysWriter *syslog.Writer
 	format    string
 	level     config.LogSeverity
+	logRotate config.LogRotate
```
gargnitingoogle, line 171:
```suggestion
	logRotateConfig config.LogRotate
```

internal/logger/logger.go
```diff
@@ -46,13 +46,13 @@ var (
 // config.
 // Here, background true means, this InitLogFile has been called for the
 // background daemon.
-func InitLogFile(filename string, format string, level config.LogSeverity) error {
+func InitLogFile(logConfig config.LogConfig) error {
```
gargnitingoogle, line 49: Is this function unit-tested ? If yes, those should be updated, and I don't see them here.

gargnitingoogle (changes requested):

PR #1488, opened 2023-11-01
Adding Cache Handler Class

raj-prince:
### Description
1. Added GetCacheHandle() and InvalidateCache() method.
2. Added unit tests for the implementation.

NA

### Testing details
1. Manual - NA
3. Unit tests -Added unit test.
4. Integration tests - NA

### DIFF
--- a/internal/cache/file/cache_handler.go
+++ b/internal/cache/file/cache_handler.go
@@ -92,13 +92,16 @@ func (chr *CacheHandler) cleanUpEvictedFile(fileInfo *data.FileInfo) error {
 	return nil
 }
 
-// addFileInfoEntryInCache adds a data.FileInfo entry for the given object and bucket
+// addFileInfoEntryToCache adds a data.FileInfo entry for the given object and bucket
 // in the file info cache if it does not already exist. While adding the entry, it also
 // performs the post eviction work (clean up for async job and local cache file) for the
 // evicted entry.
+// In case if the cache contains the stale data.FileInfo entry (generation < object.generation)
+// it cleans up (job and local cache file) for the old entry and adds the new entry with latest
+// generation to the cache.
 
 // Acquires and releases LOCK(CacheHandler.mu)
-func (chr *CacheHandler) addFileInfoEntryInCache(object *gcs.MinObject, bucket gcs.Bucket) error {
+func (chr *CacheHandler) addFileInfoEntryToCache(object *gcs.MinObject, bucket gcs.Bucket) error {
 	fileInfoKey := data.FileInfoKey{
 		BucketName: bucket.Name(),
 		ObjectName: object.Name,
@@ -111,9 +114,29 @@ func (chr *CacheHandler) addFileInfoEntryInCache(object *gcs.MinObject, bucket g
 	chr.mu.Lock()
 	defer chr.mu.Unlock()
 
+	addEntryToCache := false
 	// Todo - (raj-prince) - this lookUp should not change the LRU order.
 	fileInfo := chr.fileInfoCache.LookUp(fileInfoKeyName)
 	if fileInfo == nil {
+		addEntryToCache = true
+	} else {
+		fileInfoData := fileInfo.(data.FileInfo)
+		if fileInfoData.ObjectGeneration < object.Generation {
+			erasedVal := chr.fileInfoCache.Erase(fileInfoKeyName)
+			if erasedVal != nil {
+				fileInfo := erasedVal.(data.FileInfo)
+				err := chr.cleanUpEvictedFile(&fileInfo)
+				if err != nil {
+					return fmt.Errorf("while performing post eviction of %s object error: %v", fileInfo.Key.ObjectName, err)
+				}
+			}
+			addEntryToCache = true
+		} else if fileInfoData.ObjectGeneration > object.Generation {
+			return fmt.Errorf("cache generation %d is more than object generation: %d", fileInfoData.ObjectGeneration, object.Generation)
+		}
+	}
+
+	if addEntryToCache {
 		fileInfo = data.FileInfo{
 			Key:              fileInfoKey,
 			ObjectGeneration: object.Generation,
@@ -152,7 +175,7 @@ func (chr *CacheHandler) addFileInfoEntryInCache(object *gcs.MinObject, bucket g
 
 // Acquires and releases LOCK(CacheHandler.mu)
 func (chr *CacheHandler) GetCacheHandle(object *gcs.MinObject, bucket gcs.Bucket, initialOffset int64) (*CacheHandle, error) {
-	err := chr.addFileInfoEntryInCache(object, bucket)
+	err := chr.addFileInfoEntryToCache(object, bucket)
 	if err != nil {
 		return nil, fmt.Errorf("while adding the entry in the cache: %v", err)
 	}
--- a/internal/cache/file/cache_handler.go
+++ b/internal/cache/file/cache_handler.go
@@ -1,0 +1,162 @@
+// Copyright 2023 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package file
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/data"
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/file/downloader"
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/lru"
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/util"
+	"github.com/googlecloudplatform/gcsfuse/internal/locker"
+	"github.com/googlecloudplatform/gcsfuse/internal/storage/gcs"
+)
+
+// CacheHandler Responsible for managing fileInfoCache as well as fileDownloadManager.
+type CacheHandler struct {
+	// fileInfoCache contains the reference of fileInfo cache.
+	fileInfoCache *lru.Cache
+
+	// jobManager contains reference to a singleton jobManager.
+	jobManager *downloader.JobManager
+
+	// Directory location which would contain local file to keep the downloaded data.
+	cacheLocation string
+
+	// Guards the handling of cache entry insertion while creating the CacheHandle.
+	// Also guards post eviction logic E.g. cancelling and deletion of async download job,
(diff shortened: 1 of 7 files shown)


internal/cache/file/cache_handler.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (chr *CacheHandler) performPostEvictionWork(fileInfo *data.FileInfo) error {
+	key := fileInfo.Key
+	_, err := key.Key()
+	if err != nil {
+		return fmt.Errorf("error while performing post eviction: %v", err)
+	}
+
+	chr.jobManager.RemoveJob(key.ObjectName, key.BucketName)
```
sethiay, line 71: can we add comment here that Removing Job doesn't delete the job object itself but invalidates the job and hence it's possible that some existing cache handle will have reference to older job object but that job object will have INVALID status.

internal/cache/file/cache_handler.go
```diff
@@ -0,0 +1,181 @@
...
+
+	localFileReadHandle, err := chr.createLocalFileReadHandle(object.Name, bucket.Name())
+	if err != nil {
+		return nil, fmt.Errorf("while create local-file read handle: %v", err)
+	}
+
+	return NewCacheHandle(localFileReadHandle, chr.jobManager.GetJob(object, bucket), chr.fileInfoCache, initialOffset), nil
+}
+
+// InvalidateCache removes the entry from the fileInfoCache, cancel the async running job incase,
```
sethiay, line 149: nit: removes download job and deletes the local file in cache for given object.
nit: Remove the TODO.

internal/cache/file/cache_handler.go
```diff
@@ -0,0 +1,181 @@
...
+		fileInfo := erasedVal.(data.FileInfo)
+		err := chr.performPostEvictionWork(&fileInfo)
+		if err != nil {
+			return fmt.Errorf("while performing post eviction of %s object error: %v", fileInfo.Key.ObjectName, err)
+		}
+	}
+	return nil
+}
+
+// Destroy destroys the internal state of CacheHandler correctly specifically closing any fileHandles.
```
sethiay, line 177: I think we may not be able to close filehandles explicitly but we will evict all entries and do clean up for evicted entries

internal/cache/file/cache_handler_test.go
```diff
@@ -0,0 +1,438 @@
...
+	ctx := context.Background()
+	testObjectContent := make([]byte, TestObjectSize)
+	n, err := rand.Read(testObjectContent)
+	AssertEq(TestObjectSize, n)
+	AssertEq(nil, err)
+	objects := map[string][]byte{TestObjectName: testObjectContent}
+	err = storageutil.CreateObjects(ctx, chrT.bucket, objects)
+	AssertEq(nil, err)
+
+	gcsObj, err := chrT.bucket.StatObject(ctx, &gcs.StatObjectRequest{Name: TestObjectName,
```
sethiay, line 76: should we reuse getMinObject method here?

internal/cache/file/cache_handler_test.go
```diff
@@ -0,0 +1,438 @@
...
+	gcsObj, err := chrT.bucket.StatObject(ctx, &gcs.StatObjectRequest{Name: objName,
+		ForceFetchFromGcs: true})
+	AssertEq(nil, err)
+	minObject := storageutil.ConvertObjToMinObject(gcsObj)
+	return &minObject
+}
+
+// isFileExist returns true if the file exists and false otherwise.
+// If an error occurs, the function panics.
+func isFileExist(filePath string) bool {
```
sethiay, line 146: nit: doesFileExist

internal/cache/file/cache_handler_test.go
```diff
@@ -0,0 +1,438 @@
...
+
+	err = chrT.cacheHandler.performPostEvictionWork(&fileInfoData)
+
+	ExpectEq(nil, err)
+	jobStatusAfter := cacheHandle.fileDownloadJob.GetStatus()
+	ExpectEq(jobStatusAfter.Name, downloader.INVALID)
+	ExpectEq(false, isFileExist(chrT.downloadPath))
+}
+
+func (chrT *cacheHandlerTest) Test_performPostEvictionWork_WhenLocalFileNotExist() {
```
sethiay, line 190: should we not throw error in performPostEvictionWork if file is already deleted ? our intention is to delete the file and if it is already deleted, we should be good.

internal/cache/file/cache_handler_test.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+func (chrT *cacheHandlerTest) Test_addFileInfoEntryInTheCacheIfNotAlready_IfAlready() {
+	// Existing cacheHandle.
+	cacheHandle1 := chrT.getCacheHandleForSetupEntryInCache()
+	AssertEq(nil, cacheHandle1.validateCacheHandle())
+
+	err := chrT.cacheHandler.addFileInfoEntryInTheCacheIfNotAlready(chrT.object, chrT.bucket)
+
+	ExpectEq(nil, err)
```
sethiay, line 216: should we check if entry is added here?

internal/cache/file/cache_handler_test.go
```diff
@@ -0,0 +1,438 @@
...
+		wg.Add(1)
+		go InvalidateCacheTestFun(i)
+		wg.Add(1)
+		go getCacheHandleTestFun(i)
+	}
+	wg.Wait()
+
+	jobStatus := cacheHandle1.fileDownloadJob.GetStatus()
+	ExpectEq(downloader.INVALID, jobStatus.Name)
+	ExpectEq(false, isFileExist(chrT.downloadPath))
```
sethiay, line 437: we can't say for sure that InvalidateCacheTestFun is called in the last right ?

internal/cache/file/cache_handler.go
raj-prince: I agree. So, we have two options:
1. We will provide Clear() method in lru cache which will return all the entries, and will perform cleanUpEvictedEntries.
2. We can directly call jobManager.destroy and fileInfoCache.Clear() - but in this case, we will not be able to delete individual entry unless until we delete the path as part of jobManager.Remove(). Let's discuss offline.

internal/cache/file/cache_handler_test.go
raj-prince: That's true, but eviction of existing fileInfo (before the concurrent test started) is sure.

internal/cache/file/cache_handler_test.go
raj-prince: Not throwing error for `file not exist`, but logged this error as warning to catch it in our testing.

internal/cache/file/cache_handler.go
raj-prince: As discussed offline, will go with the jobManager.Destroy() and will delete everything under cacheLocation. But we will call it out.

internal/cache/file/cache_handler.go
```diff
@@ -0,0 +1,181 @@
...
+
+		for _, val := range evictedValues {
+			fileInfo := val.(data.FileInfo)
+			err := chr.performPostEvictionWork(&fileInfo)
+			if err != nil {
+				return fmt.Errorf("while performing post eviction of %s object error: %v", fileInfo.Key.ObjectName, err)
+			}
+		}
+	} else {
+		job := chr.jobManager.GetJob(object, bucket)
```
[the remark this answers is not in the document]
raj-prince: I don't think this check is required here. Although we can add check which will act like as an assert that - fileInfo.generation will always be equal to object.generation.

The reason is for every kernel read fileHandle's tryEnsure is called - 

``` go
// If we already have a reader, and it's at the appropriate generation, we
	// can use it. Otherwise we must throw it away.
	if fh.reader != nil {
		if fh.reader.Object().Generation == fh.inode.SourceGeneration().Object {
			return
		}

		fh.reader.Destroy()
		fh.reader = nil
	}
```

So, we will invalidate the cache as part of destroy method only.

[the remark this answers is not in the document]
sethiay: I think it should be required in the below case
1. Open, Read  and close file A - File A is cached with Gen 1.
2. File A is changed remotely - The backing Object's Gen becomes 2.
3. Stat cache expired
5. Open, Read and close file A again - The above check that you mentioned will not be triggered, right ? (because at the time of opening file, LookUpInode will be called and that will update the object's generation to 2). So, we will end up reading stale data from file in cache ?

internal/cache/file/cache_handler_test.go
sethiay: Yes, however, ExpectEq(false, isFileExist(chrT.downloadPath)) should not always be true if InvalidateCacheTestFun is not called in the last ?

internal/cache/file/cache_handler.go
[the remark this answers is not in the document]
raj-prince: Got it. In this case, object.generation == inode.generation but greater than fileInfo.generation.

internal/cache/file/cache_handler.go
```diff
@@ -0,0 +1,224 @@
...
+	// Removing Job doesn't delete the job object itself but invalidates the job and
+	// hence it's possible that some existing cache handle will have reference to older
+	// job object, but that job object will have INVALID status.
+	chr.jobManager.RemoveJob(key.ObjectName, key.BucketName)
+
+	localFilePath := util.GetDownloadPath(chr.cacheLocation, util.GetObjectPath(key.BucketName, key.ObjectName))
+	err = os.Remove(localFilePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			logger.Warnf("while clean up, filePath should exist: %v", err)
```
sethiay, line 86: cleanUpEvictedFile: file was not present at the time of clean up.

Let's start the error string with function name at all places.

internal/cache/file/cache_handler.go
```diff
@@ -0,0 +1,224 @@
...
+			return fmt.Errorf("while deleting file: %s, error: %v", localFilePath, err)
+		}
+	}
+
+	return nil
+}
+
+// addFileInfoEntryToCache adds a data.FileInfo entry for the given object and bucket
+// in the file info cache if it does not already exist. While adding the entry, it also
+// performs the post eviction work (clean up for async job and local cache file) for the
```
sethiay, line 97: nit: let's just call job and not async job.

//While adding the entry, it also
// performs the post eviction work (clean up for async job and local cache file) for the

It also cleans up clean up for entries that are evicted at the time of adding new entry.

internal/cache/file/cache_handler.go
```diff
@@ -0,0 +1,224 @@
...
+	// Todo - (raj-prince) - this lookUp should not change the LRU order.
+	fileInfo := chr.fileInfoCache.LookUp(fileInfoKeyName)
+	if fileInfo == nil {
+		addEntryToCache = true
+	} else {
+		fileInfoData := fileInfo.(data.FileInfo)
+		if fileInfoData.ObjectGeneration < object.Generation {
+			erasedVal := chr.fileInfoCache.Erase(fileInfoKeyName)
+			if erasedVal != nil {
+				fileInfo := erasedVal.(data.FileInfo)
```
sethiay, line 127: let's use the name erasedFileInfo here to avoid confusion with fileInfo used outside if condition.

sethiay (dismissed):

PR #1499, opened 2023-11-11
Fix experimental periodic perf test

Tulsishah:
### Description
It is failing because writes are happening on the same bucket with the same files parallelly with four experiments. 

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


Tulsishah (commented):
@raj-prince , @sethiay Please take a look.

perfmetrics/scripts/run_load_test_and_fetch_metrics.sh
```diff
@@ -35,7 +35,7 @@ echo "Mounting gcs bucket"
 mkdir -p gcs
 GCSFUSE_FLAGS=$1
 UPLOAD_FLAGS=$2
-BUCKET_NAME=periodic-perf-tests
+BUCKET_NAME="periodic-perf-tests${EXPERIMENT_NUMBER}"
```
raj-prince, line 38: What happens when we set `EXPERIMENT_NUMBER` with empty value?

Can we add a check to fail the job if the environment variable is set but empty, just to make sure to not set this environment variable with empty value?

raj-prince: Or if we don't then please pass the BucketName too as a command line arguments.

Tulsishah: `Can we add a check to fail the job if the environment variable is set but empty, just to make sure to not set this environment variable with empty value?` For [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh) tests `EXPERIMENT_NUMBER` is unset(empty value). when we will remove this setup at that time we can add this condition I guess.

perfmetrics/scripts/continuous_test/gcp_ubuntu/periodic_experiments/build.sh
```diff
@@ -83,7 +82,8 @@ fi
 # Executing perf tests
 LOG_FILE_FIO_TESTS="${KOKORO_ARTIFACTS_DIR}/gcsfuse-logs${EXPERIMENT_NUMBER}.txt"
 GCSFUSE_FIO_FLAGS="$GCSFUSE_FLAGS --log-file $LOG_FILE_FIO_TESTS --log-format \"text\" --stackdriver-export-interval=30s"
-./run_load_test_and_fetch_metrics.sh "$GCSFUSE_FIO_FLAGS" "$UPLOAD_FLAGS"
+BUCKET_NAME="periodic-perf-tests${EXPERIMENT_NUMBER}"
```
raj-prince, line 85: nit: can we name it experimental-periodic-perf-tests-${EXPERIMENT_NUMBER}?

Also, why are we keeping creds.json in the bucket?

Tulsishah: fixed
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/perfmetrics/scripts/run_load_test_and_fetch_metrics.sh#L53

this script is common for both the builds and it is downloading creds.json file to upload data on gsheet.

raj-prince (approved):

perfmetrics/scripts/continuous_test/gcp_ubuntu/periodic_experiments/build.sh
raj-prince: As discussed offline, anyway we are not gonna upload to gsheet as part of periodic_experiments. So, we will not keep creds.json in the experimental-periodic-perf-tests-${EXPERIMENT_NUMBER} bucket.

PR #1500, opened 2023-11-15
Increase linux-tests timeout to 15m

gargnitingoogle:
### Description
Increase linux-tests timeout to 15m
    
This is to address internal issue#311507969 .
The linux-tests in github ci have false failures because of
timeout (currently 10 m) some times, which force the developer to
rerun the test manually. Increasing the timeout duration from
10m to 15m should lower the failure caused by timeout and
should reduce developer pain.

#### Justification especially for the read_cache_release branch:
[screenshot](https://user-images.githubusercontent.com/8597474/284286106-61d562aa-9fda-4d9c-af0f-9ab10c15813c.png)
In the above run, the overall runtime of the linux-tests presubmit run is more than 6 min. But it varies from run to run.

As mentioned [here](https://github.com/GoogleCloudPlatform/gcsfuse/pull/1500#issuecomment-1819102111), about 4 min of runtime is supposed to be added by new cache tests, thus timeout needs to also be increased from 10 min by 4 minutes or more.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         go: [ 1.21.x ]
     runs-on: ubuntu-20.04
-    timeout-minutes: 10
+    timeout-minutes: 15
 
     steps:
     - uses: actions/checkout@v2


raj-prince:
Did we check the average time taken by the successful linux-test? There might be chance, intermittent failure is happening due to some other cause. So, we should fix those test cases. For example, in this PR linux test only took around 2m, with this perspective 10m is enough.

gargnitingoogle:
@raj-prince changed destination branch to read_cache_release.

gargnitingoogle:
@raj-prince please review.

gargnitingoogle:
> You can update the reasoning in the PR description.
> 
> image
> 
> We have added Cache detector test specially for `internal/cache` folder, which takes around 4m. Increasing the timeout to incorporate those test-runs time.

Done. Updated PR description.

raj-prince (approved):

PR #1504, opened 2023-11-16
Integrate with lumberjack for log rotation

ashmeenkaur:
### Description
1. This PR enables log rotation on GCSFuse logs by integrating with lumberjack. 
2. This PR also cleans up `internal/logger/logger.go` and corresponding changes to `internal/fs/fs_test.go` to remove unused logger methods.

Rotated file format: `<log-file-name>-<date in YYYY-MM-DD format>T<time in HH-MM-SS.millisecond format>.<file extension>.<gz if compressed>`
Example: log-2023-11-22T23-22-58.487.txt.gz

In case of remount with no change in the log file name, old logs from the previous mount will be retained and counted against the log rotation limits specified in config file.

NA

### Testing details
1. Manual - Manually verified on VM and docker image that log rotation works as expected.
2. Unit tests - NA
3. Integration tests - Added

Perf Test Results:
log-enabled-perf

### DIFF
--- a/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
+++ b/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
@@ -63,6 +63,8 @@ function install_requirements() {
   echo installing requirements
   echo Installing python3-pip
   sudo apt-get -y install python3-pip
+  echo Installing Bigquery module requirements...
+  pip install --require-hashes -r ./perfmetrics/scripts/bigquery/requirements.txt --user
   echo Installing libraries to run python script
   pip install google-cloud
   pip install google-cloud-vision
--- a/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
+++ b/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
@@ -88,6 +88,7 @@ if [[ "$perfTestStr" == *"$EXECUTE_PERF_TEST_LABEL"* ]];
 then
  # Executing perf tests for master branch
  install_requirements
+ git reset --hard
  git checkout master
  # Store results
  touch result.txt
@@ -97,6 +98,7 @@ then
 
  # Executing perf tests for PR branch
  echo checkout PR branch
+ git reset --hard
  git checkout pr/$KOKORO_GITHUB_PULL_REQUEST_NUMBER
  echo Mounting gcs bucket from pr branch and execute tests
  execute_perf_test
@@ -110,6 +112,7 @@ fi
 if [[ "$integrationTestsStr" == *"$EXECUTE_INTEGRATION_TEST_LABEL"* ]];
 then
   echo checkout PR branch
+  git reset --hard
   git checkout pr/$KOKORO_GITHUB_PULL_REQUEST_NUMBER
 
   echo "Running e2e tests...."
--- a/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
+++ b/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
@@ -91,22 +91,18 @@ if [[ "$perfTestStr" == *"$EXECUTE_PERF_TEST_LABEL"* ]];
 then
  # Executing perf tests for master branch
  install_requirements
-  mkdir /tmp/gcsfuse_integration_test_logs
-  rm -r /tmp/gcsfuse_integration_test_logs
-  echo "logging:
-          file-path: /tmp/gcsfuse_integration_test_logs/log.txt
-          format: text
-          severity: trace
-          log-rotate:
-            max-file-size-mb: 1
-            file-count: 1
-            compress: true
-         " > /tmp/gcsfuse_config.yaml
  git reset --hard
  git checkout master
  # Store results
  touch result.txt
  echo Mounting gcs bucket for master branch and execute tests
+ echo writing config file without log rotation
+  mkdir /tmp/gcsfuse_integration_test_logs
+  echo "logging:
+          file-path: /tmp/gcsfuse_integration_test_logs/log1.txt
+          format: text
+          severity: trace
+         " > /tmp/gcsfuse_config.yaml
  execute_perf_test
 
 
@@ -114,6 +110,16 @@ then
  echo checkout PR branch
  git reset --hard
  git checkout pr/$KOKORO_GITHUB_PULL_REQUEST_NUMBER
+ echo writing config file with log rotation
+ echo "logging:
+          file-path: /tmp/gcsfuse_integration_test_logs/log2.txt
+          format: text
+          severity: trace
+          log-rotate:
+            max-file-size-mb: 200
+            file-count: 10
+            compress: true
+         " > /tmp/gcsfuse_config.yaml
  echo Mounting gcs bucket from pr branch and execute tests
  execute_perf_test
 
--- a/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
+++ b/perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
@@ -47,8 +47,7 @@ git fetch origin -q
 
 function execute_perf_test() {
   mkdir -p gcs
-  echo "running perf tests with trace logs"
-  GCSFUSE_FLAGS="--implicit-dirs --max-conns-per-host 100 --config-file=/tmp/gcsfuse_config.yaml"
+  GCSFUSE_FLAGS="--implicit-dirs --max-conns-per-host 100"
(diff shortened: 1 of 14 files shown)


raj-prince:
As discussed offline:
1. Please update the PR description with rotated-file names format, and also what happens in case of remount.
2. Generate the logs by running concurrent worker (around 10) to reduce the time to generate logs.
3. Please check the GCSFuse performance with and without this changes by enabling all the logs.
4. Please add the integration test for the uncompressed rotated file.

perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
```diff
@@ -47,7 +47,8 @@ git fetch origin -q
 
 function execute_perf_test() {
   mkdir -p gcs
-  GCSFUSE_FLAGS="--implicit-dirs --max-conns-per-host 100"
+  echo "running perf tests with trace logs"
```
raj-prince, line 50: Assuming all these are temporary changes for presubmit perf test. Adding a note to revert before merge.

tools/integration_tests/log_rotation/helpers.go
```diff
@@ -0,0 +1 @@
+package log_rotation
```
raj-prince, line 1: Do we need this file? If not please revert.

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -0,0 +1,139 @@
...
+	// Setup file to write to.
+	filePath := path.Join(testDirPath, fileName)
+	fh := operations.CreateFile(filePath, filePerms, t)
+
+	// Keep performing operations in mounted directory until log file is rotated.
+	var lastLogFileSize int64 = 0
+	for {
+		operations.WriteAt(string(randomData), 0, fh, t)
+
+		fi, err := operations.StatFile(logFilePath)
```
raj-prince, line 66: I am not sure, t.Errorf will work inside go routine or not. Have we checked this by generating the test error.
If not then we can use ErrGroup instead of WaitGroup.

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -0,0 +1,139 @@
...
+	MiB          = 1024 * 1024
+	filePerms    = 0644
+	testFileName = "foo"
+)
+
+////////////////////////////////////////////////////////////////////////
+// Helpers
+////////////////////////////////////////////////////////////////////////
+
+func generateRandomData(t *testing.T, sizeInBytes int64) []byte {
```
raj-prince, line 41: Instead we can create a util to create the randomData. We don't need to couple it with *testing.T.

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -0,0 +1,95 @@
...
+	configFile1 := setup.YAMLConfigFile(
+		getMountConfigForLogRotation(1, 2, true, logFilePath),
+		"config1.yaml")
+	configFile2 := setup.YAMLConfigFile(
+		getMountConfigForLogRotation(1, 2, false, logFilePath),
+		"config2.yaml")
+
+	// Set up flags to run tests on.
+	// Not setting config file explicitly with 'create-empty-file: false' as it is default.
+	flags := [][]string{
```
raj-prince, line 84: Optional: We might disable the metadata-cache to generate more logs per operations.

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -0,0 +1,139 @@
...
+	// Generate random data to write to file.
+	randomData := generateRandomData(t, MiB)
+	// Setup file to write to.
+	filePath := path.Join(testDirPath, fileName)
+	fh := operations.CreateFile(filePath, filePerms, t)
+
+	// Keep performing operations in mounted directory until log file is rotated.
+	var lastLogFileSize int64 = 0
+	for {
+		operations.WriteAt(string(randomData), 0, fh, t)
```
raj-prince, line 64: Optional: I feel read operation will generate more logs in less time.

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -0,0 +1,139 @@
...
+			// Log file got rotated as current log file size < last log file size.
+			break
+		}
+		lastLogFileSize = (*fi).Size()
+	}
+	operations.CloseFileShouldNotThrowError(fh, t)
+}
+
+func runParallelOperationsInMountedDirectoryTillLogRotation(t *testing.T) {
+	// Parallely perform operations on 10 files in-order to generate logs.
```
raj-prince, line 80: nit: Parallelly

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Else run tests for testBucket.
+	// Set up test directory.
+	setup.SetUpTestDirForTestBucketFlag()
+
+	logDirPath = setup.SetUpLogDir(logDirName)
+	logFilePath = path.Join(logDirPath, logFileName)
+
+	configFile1 := setup.YAMLConfigFile(
+		getMountConfigForLogRotation(1, 2, true, logFilePath),
```
raj-prince, line 76: Please make 1 and 2 as const and also use that while validation in validateLogFileSize.

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// Run tests for mountedDirectory only if --mountedDirectory flag is set.
+	logDirPath = path.Join("/tmp", logDirName)
+	logFilePath = path.Join(logDirPath, logFileName)
+	setup.RunTestsForMountedDirectoryFlag(m)
+
+	// Else run tests for testBucket.
+	// Set up test directory.
+	setup.SetUpTestDirForTestBucketFlag()
+
+	logDirPath = setup.SetUpLogDir(logDirName)
```
raj-prince, line 72: Just curious, for mounted directly (line: 64) we didn't create directory (logDirPath), here we need to create the directory (logDirPath).

Anyway we should create the path using join here only. Then we should create the directory via utility method.

raj-prince (changes requested):
Minor comments.

tools/integration_tests/log_rotation/logrotate_logfile_test.go
ashmeenkaur: I've verified that t.Errorf is working with waitgroup.

tools/integration_tests/operations/operations_test.go
```diff
@@ -100,16 +100,20 @@ func TestMain(m *testing.M) {
...
+	// Set up config file with create-empty-file: true.
 	mountConfig := config.MountConfig{
 		WriteConfig: config.WriteConfig{
 			CreateEmptyFile: true,
 		},
 		LogConfig: config.LogConfig{
 			Severity: config.TRACE,
+			LogRotateConfig: config.LogRotateConfig{
+				MaxFileSizeMB: 10,
+				FileCount:     1,
```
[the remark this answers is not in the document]
ashmeenkaur: As per current behaviour, this will retain all the old logs. I will change this in a follow up PR.

tools/integration_tests/log_rotation/logrotate_logfile_test.go
ashmeenkaur: Thank you, this was really helpful!

tools/integration_tests/log_rotation/log_rotation_test.go
ashmeenkaur: The test is taking a few seconds to execute with read operations so I am skipping this :)

perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
ashmeenkaur: Reverted this. The results are in the PR description.

tools/integration_tests/log_rotation/log_rotation_test.go
ashmeenkaur: The problem here was that log file is generated at the time of mount. So for mounted directory tests, the directory should be there beforehand. SO I've added it to run_tests_mounted_directory script.

I will add a check to fail the test if directory is not present. 

The path is different for both the tests. One is in test directory and the other in /tmp but I will move the logic together.

raj-prince (approved):

PR #1505, opened 2023-11-16
add configs for log rotation

ashmeenkaur:
### Description
Added the following configs for log rotation.
1. max-file-size-mb: specifies the maximum size in megabytes that a log file
can reach before it is rotated. The default value is 512 megabytes.
2. file-count: determines the maximum number of log files to retain after they
have been rotated. The default value is 10.
3. compress: indicates whether the rotated log files should be compressed
using gzip. The default value is False.

NA

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/internal/config/mount_config.go
+++ b/internal/config/mount_config.go
@@ -14,14 +14,22 @@
 
 package config
 
+const (
+	// Default log rotation config values.
+	defaultMaxFileSizeMB = 512
+	defaultFileCount     = 10
+	defaultCompress      = false
+)
+
 type WriteConfig struct {
 	CreateEmptyFile bool `yaml:"create-empty-file"`
 }
 
 type LogConfig struct {
-	Severity LogSeverity `yaml:"severity"`
-	Format   string      `yaml:"format"`
-	FilePath string      `yaml:"file-path"`
+	Severity        LogSeverity     `yaml:"severity"`
+	Format          string          `yaml:"format"`
+	FilePath        string          `yaml:"file-path"`
+	LogRotateConfig LogRotateConfig `yaml:"log-rotate"`
 }
 
 type MountConfig struct {
@@ -29,11 +37,35 @@ type MountConfig struct {
 	LogConfig   `yaml:"logging"`
 }
 
+// LogRotateConfig defines the parameters for log rotation. It consists of three
+// configuration options:
+// 1. max-file-size-mb: specifies the maximum size in megabytes that a log file
+// can reach before it is rotated. The default value is 512 megabytes.
+// 2. file-count: determines the maximum number of log files to retain after they
+// have been rotated. The default value is 10.
+// 3. compress: indicates whether the rotated log files should be compressed
+// using gzip. The default value is False.
+type LogRotateConfig struct {
+	MaxFileSizeMB int  `yaml:"max-file-size-mb"`
+	FileCount     int  `yaml:"file-count"`
+	Compress      bool `yaml:"compress"`
+}
+
+func DefaultLogRotateConfig() LogRotateConfig {
+	return LogRotateConfig{
+		MaxFileSizeMB: defaultMaxFileSizeMB,
+		FileCount:     defaultFileCount,
+		Compress:      defaultCompress,
+	}
+}
+
 func NewMountConfig() *MountConfig {
 	mountConfig := &MountConfig{}
 	mountConfig.LogConfig = LogConfig{
 		// Making the default severity as INFO.
 		Severity: INFO,
+		// Setting default values of log rotate config.
+		LogRotateConfig: DefaultLogRotateConfig(),
 	}
 	return mountConfig
 }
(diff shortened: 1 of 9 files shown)


internal/config/mount_config.go
```diff
@@ -14,26 +14,58 @@
...
 	LogConfig   `yaml:"logging"`
 }
 
+// LogRotateConfig defines the parameters for log rotation. It consists of three
+// configuration options:
+// 1. max-file-size-mb: specifies the maximum size in megabytes that a log file
+// can reach before it is rotated. The default value is 512 megabytes.
+// 2. file-count: determines the maximum number of log files to retain after they
+// have been rotated. The default value is 10.
+// 3. compress: indicates whether the rotated log files should be compressed
```
raj-prince, line 46: Should we write by what factor (~) log-file size gets compressed after using this flag?
Not here, but we can update in log-rotate doc.

internal/config/yaml_parser.go
```diff
@@ -49,6 +51,16 @@ func IsValidLogSeverity(severity LogSeverity) bool {
 	return false
 }
 
+func IsValidLogRotateConfig(config LogRotateConfig) error {
+	if config.MaxFileSizeMB <= 0 {
+		return fmt.Errorf("max-file-size-mb cannot be less than 0")
```
raj-prince, line 56: nit: less than or equal to 0. At both places.

internal/config/testdata/invalid_log_rotate_config_1.yaml
```diff
@@ -0,0 +1,7 @@
+write:
+  create-empty-file: true
+logging:
+  severity: error
+  log-rotate:
+    max-file-size-mb: -1
+    file-count: -1
```
raj-prince, line 7: Just curious, what happens if someone assign float value on these options.

internal/config/yaml_parser.go
```diff
@@ -33,6 +33,8 @@ const (
 	WARNING LogSeverity = "WARNING"
 	ERROR   LogSeverity = "ERROR"
 	OFF     LogSeverity = "OFF"
+
+	parseConfigFileError = "error parsing config file: %w"
```
raj-prince, line 37: nit: parseConfigFileErrMsgFormat, %v

internal/config/yaml_parser_test.go
```diff
@@ -89,3 +95,17 @@ func (t *YamlParserTest) TestReadConfigFile_InvalidValidLogConfig() {
 	AssertNe(nil, err)
 	AssertTrue(strings.Contains(err.Error(), "error parsing config file: log severity should be one of [trace, debug, info, warning, error, off]"))
 }
+
+func (t *YamlParserTest) TestReadConfigFile_InvalidValidLogRotateConfig1() {
+	_, err := ParseConfigFile("testdata/invalid_log_rotate_config_1.yaml")
+
+	AssertNe(nil, err)
+	AssertTrue(strings.Contains(err.Error(), "error parsing config file: max-file-size-mb cannot be less than 0"))
```
raj-prince, line 103: I think, we can use `parseConfigFileError` variable here?

raj-prince (changes requested):
Overall looks good. Few minor comments.

internal/config/yaml_parser_test.go
```diff
@@ -34,6 +34,9 @@ func validateDefaultConfig(mountConfig *MountConfig) {
 	AssertEq("INFO", mountConfig.LogConfig.Severity)
 	AssertEq("", mountConfig.LogConfig.Format)
 	AssertEq("", mountConfig.LogConfig.FilePath)
+	AssertEq(512, mountConfig.LogConfig.LogRotateConfig.MaxFileSizeMB)
```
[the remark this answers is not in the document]
ashmeenkaur: Thank you for the input. I will keep this in mind for any future changes. I think for all the current changes, we would like to fail the unit test, if the values are incorrect. Please let me know if you think otherwise :)

internal/config/testdata/invalid_log_rotate_config_1.yaml
ashmeenkaur: It takes the floor integer value in this case.

internal/config/yaml_parser.go
ashmeenkaur: Changed to more than 1 at both places to also take care of floating point values as mentioned in the other comment.

internal/logger/logger_test.go
```diff
@@ -261,3 +262,32 @@ func (t *LoggerTest) TestSetLoggingLevel() {
 		AssertEq(test.programLevel.Level(), test.expectedProgramLevel)
 	}
 }
+
+func (t *LoggerTest) TestInitLogFile() {
```
raj-prince, line 266: Please use Expect in this test also wherever required.

raj-prince (approved):

PR #1508, opened 2023-11-17
Adding checks to catch - if cache is deleted externally

raj-prince:
### Description
1. Handling error the case when someone deletes the file cache externally.
2. Adding support to invalidate cache in case some deletes the file.

Important: What happens if the cache file is deleted externally? This means that the fileInfoCache contains an entry for the cache file, but the cache file is not available locally.
(a) If the fileCacheHandle is nil, in this case it will return a data-inconsistent error.
(b) In the case of an old cacheFileHandle, if the local file is deleted externally, this will return correct data until the handle is closed. This is because Linux does not delete a file until its link count is zero. When the fileHandle is open, it contributes to the link_count and therefore the file is not deleted and serves the data correctly.

NA

### Testing details
1. Manual - NA
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/internal/cache/file/cache_handler_test.go
+++ b/internal/cache/file/cache_handler_test.go
@@ -269,12 +269,10 @@ func (chrT *cacheHandlerTest) Test_addFileInfoEntryToCache_IfNotAlready() {
 	ExpectEq(false, doesFileExist(chrT.downloadPath))
 }
 
-func (chrT *cacheHandlerTest) Test_addFileInfoEntryToCache_IfLocalFileGotDeleted() {
-	{
-		// Existing cacheHandle.
-		cacheHandle1 := chrT.getCacheHandleForSetupEntryInCache()
-		AssertEq(nil, cacheHandle1.validateCacheHandle())
-	}
+func (chrT *cacheHandlerTest) Test_addFileInfoEntryToCache_IfLocalFileGetsDeleted() {
+	// Existing cacheHandle.
+	cacheHandle1 := chrT.getCacheHandleForSetupEntryInCache()
+	AssertEq(nil, cacheHandle1.validateCacheHandle())
 	// Delete the local cache file.
 	err := os.Remove(chrT.downloadPath)
 	AssertEq(nil, err)
@@ -283,7 +281,7 @@ func (chrT *cacheHandlerTest) Test_addFileInfoEntryToCache_IfLocalFileGotDeleted
 	err = chrT.cacheHandler.addFileInfoEntryToCache(chrT.object, chrT.bucket)
 
 	AssertNe(nil, err)
-	ExpectTrue(strings.Contains(err.Error(), "data inconsistent -"))
+	ExpectTrue(strings.Contains(err.Error(), util.DataInconsistentErrMsg))
 }
 
 func (chrT *cacheHandlerTest) Test_GetCacheHandle_WhenCacheContainsStaleEntry() {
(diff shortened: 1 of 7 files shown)


internal/gcsx/random_reader_test.go
```diff
@@ -985,6 +985,74 @@ func (t *RandomReaderTest) Test_ReadAt_CacheHitNextToCacheMissIncaseOfInvalidFil
...
+	rc1 := getReadCloser(testContent)
+	t.mockNewReaderCallForTestBucket(0, objectSize, rc1)
+	ExpectCall(t.bucket, "Name")().WillRepeatedly(Return("test"))
+	buf := make([]byte, objectSize)
+	_, cacheHit, err := t.rr.ReadAt(buf, 0)
+	AssertEq(nil, err)
+	AssertTrue(cacheHit)
+	AssertTrue(reflect.DeepEqual(testContent, buf))
+	AssertNe(nil, t.rr.wrapped.fileCacheHandle)
+	err = t.rr.wrapped.fileCacheHandle.Close()
```
raj-prince, line 1032: Ohh, test was wrong.

internal/gcsx/random_reader_test.go
```diff
@@ -978,6 +978,74 @@ func (t *RandomReaderTest) Test_ReadAt_CacheHitNextToCacheMissIncaseOfInvalidFil
...
+	// Delete the local cache file.
+	filePath := util.GetDownloadPath(t.cacheLocation, util.GetObjectPath(t.bucket.Name(), t.object.Name))
+	err = os.Remove(filePath)
+	AssertEq(nil, err)
+	// Second reader (rc2) is required, since first reader (rc) is completely read.
+	// Reading again will return EOF.
+	rc2 := getReadCloser(testContent)
+	t.mockNewReaderCallForTestBucket(0, objectSize, rc2)
+	_, cacheHit, err = t.rr.ReadAt(buf, 0) // Read from GCS.
+	ExpectEq(nil, err)
```
sethiay, line 1036: can we use assert here as well ?

internal/gcsx/random_reader_test.go
```diff
@@ -978,6 +978,74 @@ func (t *RandomReaderTest) Test_ReadAt_CacheHitNextToCacheMissIncaseOfInvalidFil
...
+	ExpectFalse(cacheHit)
+	ExpectTrue(reflect.DeepEqual(testContent, buf))
+	ExpectEq(nil, t.rr.wrapped.fileCacheHandle)
+	rc3 := getReadCloser(testContent)
```
sethiay, line 1040: I didn't follow quite follow what's happening here ? Could you please write comments what's happening in 2nd read and in 3rd read exactly ?

internal/cache/file/cache_handler_test.go
```diff
@@ -266,6 +269,21 @@ func (chrT *cacheHandlerTest) Test_addFileInfoEntryToCache_IfNotAlready() {
...
 
+func (chrT *cacheHandlerTest) Test_addFileInfoEntryToCache_IfLocalFileGetsDeleted() {
+	// Existing cacheHandle.
+	cacheHandle1 := chrT.getCacheHandleForSetupEntryInCache()
+	AssertEq(nil, cacheHandle1.validateCacheHandle())
+	// Delete the local cache file.
+	err := os.Remove(chrT.downloadPath)
+	AssertEq(nil, err)
+
+	// Insertion will happen and that leads to eviction.
```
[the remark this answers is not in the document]
raj-prince: My bad - this comment was from the different test case.  Corrected it - in this case we will have a fileInfoEntry in the fileInfoCache but not local cache file for this entry. Hence, DataInconsistent error will come.

internal/gcsx/random_reader_test.go
raj-prince: Updated the test and comment.

internal/fs/read_cache_test.go
```diff
@@ -289,20 +295,132 @@ func (t *FileCacheTest) DeletingFileFromCacheShouldReadFromGCS() {
...
+	_, err = file.Read(buf)
+	AssertEq(nil, err)
+	closeFile(file)
+	objectPath := util.GetObjectPath(bucket.Name(), DefaultObjectName)
+	downloadPath := util.GetDownloadPath(CacheLocation, objectPath)
+	file, err = os.OpenFile(filePath, os.O_RDWR|syscall.O_DIRECT, util.DefaultFilePerm)
+	AssertEq(nil, err)
+	defer closeFile(file)
+	AssertEq(nil, err)
+	_, err = file.Read(buf)
```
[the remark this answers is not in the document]
raj-prince: Updated the test - we don't need three reads.

sethiay (approved):

PR #1519, opened 2023-11-22
Add lookup and update without changing order in LRU cache

sethiay:
### Description
- Add lookup and update without changing order in LRU cache
- Also removed one of the unit test in job_test.go as it is expected that GetCacheHandle will insert fileInfo in fileInfoCache at the time of creating cache handle and initializing job.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added unit test.
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"strings"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/cache/data"
 	"github.com/googlecloudplatform/gcsfuse/internal/cache/lru"
@@ -91,7 +92,7 @@ type jobSubscriber struct {
 }
 
 func NewJob(object *gcs.MinObject, bucket gcs.Bucket, fileInfoCache *lru.Cache,
-	sequentialReadSizeMb int32, fileSpec data.FileSpec) (job *Job) {
+		sequentialReadSizeMb int32, fileSpec data.FileSpec) (job *Job) {
 	job = &Job{
 		object:               object,
 		bucket:               bucket,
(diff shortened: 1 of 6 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -175,7 +176,7 @@ func (job *Job) notifySubscribers() {
 	for subItr := job.subscribers.Front(); subItr != nil; subItr = nextSubItr {
 		subItrValue := subItr.Value.(jobSubscriber)
 		nextSubItr = subItr.Next()
-		if job.status.Name == FAILED || job.status.Name == CANCELLED || job.status.Offset >= subItrValue.subscribedOffset {
+		if job.status.Name == FAILED || job.status.Name == CANCELLED || job.status.Name == INVALID || job.status.Offset >= subItrValue.subscribedOffset {
```
raj-prince, line 179: We can change the `Name` to `State`.

internal/cache/lru/lru.go
```diff
@@ -214,3 +218,49 @@ func (c *Cache) LookUp(key string) (value ValueType) {
 	// Return the value.
 	return e.Value.(entry).Value
 }
+
+// LookUpWithoutChangingOrder looks up previously-inserted value for a given key
+// without changing the order of entries in the cache. Return nil if no value
+// is present.
+func (c *Cache) LookUpWithoutChangingOrder(key string) (value ValueType) {
```
raj-prince, line 225: Should we use `sync.RWMutex` to allow multiple readers together for this method?

raj-prince (changes requested):
Looks good. Added a couple of comments.

internal/cache/lru/lru.go
sethiay: That's a good catch. Thanks !

internal/locker/rw_locker.go
```diff
@@ -78,7 +78,9 @@ func (c *rwChecker) RUnlock() {
 	c.locker.RUnlock()
 }
 
-// Note: rwDebugger doesn't check potential deadlock in case of read only lock.
+// Note: rwDebugger doesn't check potential deadlock in case of read only lock as
+// doing that is not straight forward because multiple readers can acquire locks
+// at the time same time and that needs keeping track of every different lock.
```
raj-prince, line 83: nit: time same time -> same time

raj-prince (approved):

PR #1522, opened 2023-11-24
Fix file creation in case of random read when downloadFileForRandomRead is False

sethiay:
### Description
Fix file creation in case of random read when downloadFileForRandomRead is true

NA

### Testing details
1. Manual - NA
2. Unit tests - Added unit tests where applicable.
3. Integration tests - NA

### DIFF
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -25,14 +25,15 @@ import (
 )
 
 const (
-	InvalidFileHandleErrMsg      = "invalid file handle"
-	InvalidFileDownloadJobErrMsg = "invalid download job"
-	InvalidCacheHandleErrMsg     = "invalid cache handle"
-	InvalidFileInfoCacheErrMsg   = "invalid file info cache"
-	ErrInSeekingFileHandleMsg    = "error while seeking file handle"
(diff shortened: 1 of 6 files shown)


internal/cache/util/util.go
```diff
@@ -25,14 +25,15 @@ import (
...
-	FileNotPresentInCacheErrMsg  = "file is not present in cache"
+	InvalidFileHandleErrMsg       = "invalid file handle"
+	InvalidFileDownloadJobErrMsg  = "invalid download job"
+	InvalidCacheHandleErrMsg      = "invalid cache handle"
+	InvalidFileInfoCacheErrMsg    = "invalid file info cache"
+	ErrInSeekingFileHandleMsg     = "error while seeking file handle"
+	ErrInReadingFileHandleMsg     = "error while reading file handle"
+	FallbackToGCSErrMsg           = "read via gcs"
+	FileNotPresentInCacheErrMsg   = "file is not present in cache"
+	CacheMissWhenRandomReadErrMsg = "downloadFileForRandomRead is false, read type random read and fileInfo entry is absent"
```
raj-prince, line 36: nit: We can think of better name - NoCacheHandleRequiredErrMsg?

raj-prince (approved):

PR #1523, opened 2023-11-26
Add destroy method to delete cache during unmounting

sethiay:
### Description
Add destroy method to delete cache post unmounting

NA

### Testing details
1. Manual - NA
2. Unit tests - Added wherever applicable
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/cache_handler.go
+++ b/internal/cache/file/cache_handler.go
@@ -261,9 +261,15 @@ func (chr *CacheHandler) InvalidateCache(objectName string, bucketName string) e
 	return nil
 }
 
-// Destroy destroys the internal state of CacheHandler correctly specifically evict all the
-// entries in the file info cache and clean up for the evicted entries.
-// TODO (raj-prince) to implement.
-func (chr *CacheHandler) Destroy() {
+// Destroy destroys the internal state of CacheHandler correctly destroying all
+// the download jobs and deleting all the files in cache.
+//
+// Acquires and releases Lock(chr.mu)
+func (chr *CacheHandler) Destroy() (err error) {
+	chr.mu.Lock()
+	defer chr.mu.Unlock()
 
+	chr.jobManager.Destroy()
+	err = os.RemoveAll(chr.cacheLocation)
+	return
 }
--- a/internal/cache/file/cache_handler.go
+++ b/internal/cache/file/cache_handler.go
@@ -261,8 +261,10 @@ func (chr *CacheHandler) InvalidateCache(objectName string, bucketName string) e
 	return nil
 }
 
-// Destroy destroys the internal state of CacheHandler correctly destroying all
-// the download jobs and deleting all the files in cache.
+// Destroy destroys the job manager (i.e. invalidate all the jobs) and delete the
+// file cache directory (i.e. deleting all the files in cache). Note: This method
+// is expected to be called at the time of unmounting and because file info cache
+// is in-memory, it is not required to destroy it.
 //
 // Acquires and releases Lock(chr.mu)
 func (chr *CacheHandler) Destroy() (err error) {
--- a/internal/cache/file/cache_handler_test.go
+++ b/internal/cache/file/cache_handler_test.go
@@ -609,7 +609,7 @@ func (chrT *cacheHandlerTest) Test_Destroy() {
 
 	AssertEq(nil, err)
 	// Verify the cacheLocation is deleted.
-	_, err = os.Stat(path.Join(chrT.cacheLocation, util.Cache))
+	_, err = os.Stat(path.Join(chrT.cacheLocation, util.FileCache))
 	AssertNe(nil, err)
 	AssertTrue(strings.Contains(err.Error(), "no such file or directory"))
 	// Verify jobs are invalidated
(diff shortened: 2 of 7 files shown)


internal/cache/file/cache_handler.go
```diff
@@ -227,9 +227,15 @@ func (chr *CacheHandler) InvalidateCache(objectName string, bucketName string) e
...
 
-// Destroy destroys the internal state of CacheHandler correctly specifically evict all the
-// entries in the file info cache and clean up for the evicted entries.
-// TODO (raj-prince) to implement.
-func (chr *CacheHandler) Destroy() {
+// Destroy destroys the internal state of CacheHandler correctly destroying all
+// the download jobs and deleting all the files in cache.
+//
+// Acquires and releases Lock(chr.mu)
+func (chr *CacheHandler) Destroy() (err error) {
```
raj-prince, line 234: Ideally, we should clear the fileInfoCache also, to make this method generic.

internal/cache/file/cache_handler_test.go
```diff
@@ -532,3 +532,35 @@ func (chrT *cacheHandlerTest) Test_InvalidateCacheAndGetHandle_Concurrent() {
...
+	minObject1 := chrT.getMinObject("object_1", []byte("content of object_1"))
+	minObject2 := chrT.getMinObject("object_2", []byte("content of object_2"))
+	cacheHandle1, err := chrT.cacheHandler.GetCacheHandle(minObject1, chrT.bucket, true, 0)
+	AssertEq(nil, err)
+	cacheHandle2, err := chrT.cacheHandler.GetCacheHandle(minObject2, chrT.bucket, true, 1)
+	AssertEq(nil, err)
+	ctx := context.Background()
+	// Read to create and populate file in cache.
+	buf := make([]byte, 3)
+	_, _ = cacheHandle1.Read(ctx, minObject1, 4, buf)
```
raj-prince, line 546: Why are we not handling/asserting the Read() error?

internal/cache/file/cache_handler_test.go
```diff
@@ -532,3 +532,35 @@ func (chrT *cacheHandlerTest) Test_InvalidateCacheAndGetHandle_Concurrent() {
...
+	job1 := cacheHandle1.fileDownloadJob
+	job2 := cacheHandle2.fileDownloadJob
+
+	err = chrT.cacheHandler.Destroy()
+
+	AssertEq(nil, err)
+	// Verify the cacheLocation is deleted.
+	_, err = os.Stat(path.Join(chrT.cacheLocation, util.FileCache))
+	AssertNe(nil, err)
+	AssertTrue(strings.Contains(err.Error(), "no such file or directory"))
```
raj-prince, line 562: Can't we use os.ErrNotExist?

internal/cache/util/util.go
```diff
@@ -41,6 +41,7 @@ const (
 	KiB                    = 1024
 	DefaultFilePerm        = os.FileMode(0600)
 	FilePermWithAllowOther = os.FileMode(0644)
+	FileCache              = "file-cache"
```
raj-prince, line 44: We can rename it to gcsfuse-file-cache, just to make the name more specific?

internal/fs/read_cache_test.go
```diff
@@ -693,10 +694,66 @@ func (t *FileCacheWithDefaultCacheLocation) SetUpTestSuite() {
...
+	err := os.RemoveAll(path.Join(os.TempDir(), util.FileCache))
 	AssertEq(nil, err)
 }
 
 func (t *FileCacheWithDefaultCacheLocation) DefaultLocationIsTempDir() {
-	sequentialReadShouldPopulateCache(&t.fsTest, os.TempDir())
+	sequentialReadShouldPopulateCache(&t.fsTest, path.Join(os.TempDir(), util.FileCache))
+}
+
+// Test to check cache is deleted at the time pf unmounting.
```
raj-prince, line 705: nit: time of unmounting

internal/fs/read_cache_test.go
```diff
@@ -693,10 +694,66 @@ func (t *FileCacheWithDefaultCacheLocation) SetUpTestSuite() {
...
+	}
+	t.serverCfg.AllowOther = true
+	t.fsTest.SetUpTestSuite()
+}
+
+func (t *FileCacheDestroyTest) TearDownTestSuite() {
+	// do nothing as fs is unmounted in the test itself
+}
+
+func (t *FileCacheDestroyTest) TearDown() {
```
raj-prince, line 731: I am not sure, if we need TearDown/TearDownTestSuite here. If not can we remove this?

raj-prince (changes requested):
Minor comments.

internal/cache/file/cache_handler.go
sethiay: Updated the comment explicitly saying that file info cache destroy is not required.

internal/fs/read_cache_test.go
sethiay: We don't need to do anything in TearDown/TestSuite that's why we have defined the function, otherwise it uses the implementation from fsTest.

raj-prince (approved):

PR #1524, opened 2023-11-26
Truncate file on invalidation before deletion

sethiay:
### Description
Truncate file on invalidation before deletion so that space is not taken because of open file handles.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added where applicable.
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/cache_handler.go
+++ b/internal/cache/file/cache_handler.go
@@ -70,8 +70,8 @@ func (chr *CacheHandler) createLocalFileReadHandle(objectName string, bucketName
 }
 
 // cleanUpEvictedFile is a utility method called for the evicted/deleted fileInfo.
-// As part of execution, it stops and removes the download job, and deletes the cache
-// file.
+// As part of execution, it stops and truncates and removes the download job,
+// and deletes the cache file.
 func (chr *CacheHandler) cleanUpEvictedFile(fileInfo *data.FileInfo) error {
 	key := fileInfo.Key
 	_, err := key.Key()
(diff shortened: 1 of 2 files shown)


internal/cache/file/cache_handler.go
```diff
@@ -70,8 +70,8 @@ func (chr *CacheHandler) createLocalFileReadHandle(objectName string, bucketName
 }
 
 // cleanUpEvictedFile is a utility method called for the evicted/deleted fileInfo.
-// As part of execution, it stops and removes the download job, and deletes the cache
-// file.
+// As part of execution, it stops and truncates and removes the download job,
```
raj-prince, line 73: nit: As part of execution, it (a) stops and removes the download job, (b) truncate and deletes the cache file.

internal/cache/file/cache_handler.go
```diff
@@ -85,14 +85,20 @@ func (chr *CacheHandler) cleanUpEvictedFile(fileInfo *data.FileInfo) error {
 	chr.jobManager.RemoveJob(key.ObjectName, key.BucketName)
 
 	localFilePath := util.GetDownloadPath(chr.cacheLocation, util.GetObjectPath(key.BucketName, key.ObjectName))
-	err = os.Remove(localFilePath)
+	// Truncate the file to 0 size, so that even if there are open file handles
+	// and linux doesn't delete the file, the file will not take space.
+	err = os.Truncate(localFilePath, 0)
 	if err != nil {
 		if os.IsNotExist(err) {
 			logger.Warnf("cleanUpEvictedFile: file was not present at the time of clean up: %v", err)
```
raj-prince, line 93: We can return from here only, if file doesn't exist.

raj-prince, line 93: nit: "at the time of clean up" -> "at the time of truncating file"

raj-prince (commented):
Minor comments.

raj-prince (approved):

PR #1534, opened 2023-11-29
Wait for termination of async job during Cancel and Invalidate

sethiay:
### Description
Wait for termination of async job during cancel and invalidation. 
Details:
Termination of Go routine is co-operative and not preemptive which means that when we signal job.downloadAsync to terminate by running job.cancelFunc() then it may not be immediately terminated and run few statements in the default case of select.
When CacheHandler.cleanUpEvictedEntry is called, it executes job.Invalidate + os.remove(file_in_cache) one after the other, because of that, it may happen that job.Invalidate signals the job.downloadAsync but it takes some time to terminate and meanwhile os.remove is called. If that happens, it's possible that job.downloadAsync is writing to the file in cache at that time and hence the file in cache will appear again which is not correct.
This is more graceful handling of cancellation of async job because job.Invalidate will wait until the job.downloadAsync is terminated and hence fixing the above possible issue.

NA

### Testing details
1. Manual - Ran 256KB fio job (40 threads) with read cache size = 5 mb and observed no extra files are being created at any point in time.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -16,6 +16,7 @@ package downloader
 
 import (
 	"container/list"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -296,12 +297,18 @@ func (job *Job) downloadObjectAsync() {
 		}
 	}(cacheFile)
 
+	notifyCancelled := func() {
+		job.mu.Lock()
+		job.status.Name = CANCELLED
+		job.notifySubscribers()
+		job.mu.Unlock()
+	}
+
 	var newReader io.ReadCloser
 	var start, end, sequentialReadSize, newReaderLimit int64
 	end = int64(job.object.Size)
 	sequentialReadSize = int64(job.sequentialReadSizeMb) * cacheutil.MiB
 
-	ctx := context.Background()
 	for {
 		select {
 		case <-job.cancelCtx.Done():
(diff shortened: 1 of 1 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -123,6 +127,26 @@ func (job *Job) init() {
 	job.status = JobStatus{NOT_STARTED, nil, 0}
 	job.subscribers = list.List{}
 	job.cancelCtx, job.cancelFunc = context.WithCancel(context.Background())
```
raj-prince, line 129: Let's call the cancelFunc() when the operation completed successfully.

Ref: https://pkg.go.dev/context#WithCancel

internal/cache/file/downloader/job.go
```diff
@@ -133,13 +157,28 @@ func (job *Job) init() {
 // Acquires and releases LOCK(job.mu)
 func (job *Job) Cancel() {
 	job.mu.Lock()
-	defer job.mu.Unlock()
-	if job.status.Name == DOWNLOADING || job.status.Name == NOT_STARTED {
-		job.cancelFunc()
```
raj-prince, line 138: What happens if we call the cancelFunc() twice?

internal/cache/file/downloader/job.go
```diff
@@ -235,8 +278,13 @@ func (job *Job) updateFileInfoCache() (err error) {
 // Note: There can only be one async download running for a job at a time.
 // Acquires and releases LOCK(job.mu)
 func (job *Job) downloadObjectAsync() {
-	// Create and open cache file for writing object into it.
-	cacheFile, err := cacheutil.CreateFile(job.fileSpec, os.O_WRONLY)
+	// Close the job.doneCh and call job.canceFunc() in any case -
```
raj-prince, line 281: nit: job.cancelFunc()

internal/cache/file/downloader/job.go
sethiay: Nothing. https://pkg.go.dev/context#CancelFunc

raj-prince (approved):

PR #1542, opened 2023-12-11
Reduce epoches in pytorch2.0 model

Tulsishah:
### Description
Reducing the epochs as pytorch2 long haul tests are running on NVIDIA L4 machines, which lack the powerful GPU of
the NVIDIA A100. So it is taking longer time to complete the training. We will set it back to 80 when the NVIDIA A100 GPU machine will be available.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/ml_tests/pytorch/run_model.sh
+++ b/perfmetrics/scripts/ml_tests/pytorch/run_model.sh
@@ -75,7 +75,7 @@ python -c 'import torch;torch.hub.list("facebookresearch/xcit:main")'
 # will be available.
 if [ ${PYTORCH_VESRION} == "v2" ];
 then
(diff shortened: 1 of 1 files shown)


sethiay (approved):

perfmetrics/scripts/ml_tests/pytorch/run_model.sh
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 PYTORCH_VESRION=$1
+EPOCHS=80
```
sethiay, line 17: Let's rename it to NUM_EPOCHS

sethiay (approved):

PR #1544, opened 2023-12-11
fix flaky log rotation test

ashmeenkaur:
### Description
There is an edge case that rarely happens in log rotation integration tests where we try to stat a log file when the logs are getting rotated (the old log file getting renamed and new log file being created). This was causing failure in the integration tests. In order to fix the issue, we will retry the stat call once again before failing the test. If two consecutive  stat calls fail, we will fail the test.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/log_rotation/logrotate_logfile_test.go
+++ b/tools/integration_tests/log_rotation/logrotate_logfile_test.go
@@ -52,6 +52,7 @@ func runOperationsOnFileTillLogRotation(t *testing.T, wg *sync.WaitGroup, fileNa
 
 	// Keep performing operations in mounted directory until log file is rotated.
 	var lastLogFileSize int64 = 0
+	var retryStatLogFile = true
 	for {
 		// Perform Read operation to generate logs.
 		_, err = operations.ReadFile(filePath)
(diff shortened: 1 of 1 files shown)


tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -62,7 +63,12 @@ func runOperationsOnFileTillLogRotation(t *testing.T, wg *sync.WaitGroup, fileNa
 		// Break the loop when log file is rotated.
 		fi, err := operations.StatFile(logFilePath)
 		if err != nil {
-			t.Errorf("stat operation on file %s: %v", logFilePath, err)
+			t.Logf("stat operation on file %s failed: %v", logFilePath, err)
+			if !retryStatLogFile {
```
raj-prince, line 67: The issue might come again in case when rotation will happen twice for a file. 
I think, we should add a check to catch when error will happen back to back (in consecutive loops).

ashmeenkaur: Discussed offline that this is only for one log rotation operation, so this check won't be required.

PR #1540, opened 2023-12-06
Merge type-cache-ttl into stat-cache-ttl

gargnitingoogle:
### Description
Take minimum of type-cache-ttl and stat-cache-ttl for a common ttl for stat-cache and type-cache i.e. `ttl=min(stat-cache-ttl,type-cache-ttl)`

Round up both stat-cache-ttl and type-cache-ttl to the next higher multiples of a second and use the minima of the two for both the TTL values.

This is followed up in https://github.com/GoogleCloudPlatform/gcsfuse/pull/1545 .

### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - Did sanity tests
2. Unit tests - Modified a few tests. All tests passed locally. 
3. Integration tests - ran through presubmit.

PR #1545, opened 2023-12-12
Add metadata-cache:ttl-secs config

gargnitingoogle:
### Description
Add the support for following configuration fields in config-file.

```
metadata-cache:
        ttl-secs:
```
    
`ttl-secs` is an integer values denoting the ttl of stat-cache (aka
metadata cache in seconds). It is `-1` for no-ttl-based-expiry behavior,
`0` for no stat-cache, and `> 0` for normal ttl-based-expiry behavior.
A value of `ttl-secs < -1` will return error.
If ttl-secs is not defined, it will
fall back to the value of gcsfuse mount-flag `stat-cache-ttl`
(rounded up to the next multiple of second) .

If both ttl-secs and `stat-cache-ttl` are absent, then the default
value of 1 minute is used for ttl for stat-cache.

`stat-cache-ttl` will soon be deprecated.

This change also adds the corresponding unit tests etc.

This builds on top of https://github.com/GoogleCloudPlatform/gcsfuse/pull/1540 .

NA

### Testing details
1. Manual - Did sanity testing locally.
2. Unit tests - Added new tests.
3. Integration tests - Through presubmits.


internal/config/mount_config.go
```diff
@@ -31,11 +41,21 @@ type FileCacheConfig struct {
 	DownloadFileForRandomRead bool  `yaml:"download-file-for-random-read"`
 }
 
+type MetadataCacheConfig struct {
+	// TtlInSeconds is the ttl
+	// value in seconds, to be used for stat-cache and type-cache.
+	// It can be set to -1 for no-ttl, 0 for
+	// no cache and > 0 for ttl-controlled metadata-cache.
+	// Any value set below -1 will throw an error.
+	TtlInSeconds int64 `yaml:"ttl-secs,omitempty"`
```
gargnitingoogle, line 50: Add special handling for TtlInSeconds being `-1`

gargnitingoogle: Done.

main.go
```diff
@@ -204,6 +206,19 @@ func runCLIApp(c *cli.Context) (err error) {
...
 		flags.DebugFuse, flags.DebugGCS, flags.DebugMutex)
 
+	// if metadata-cache:ttl-secs has been set in config-file, then
+	// switch metadata cache ttls (stat-cache-ttl and type-cache-tll)
+	// to that as that takes precedence over both stat-cache-ttl and
+	// type-cache-tll.
+	if mountConfig.MetadataCacheConfig.TtlInSeconds != config.TtlInSecsUnset {
+		// if ttl-secs is set to -1, set StatOrTypeCacheTTL to the max possible duration.
+		if mountConfig.MetadataCacheConfig.TtlInSeconds == -1 {
+			flags.StatOrTypeCacheTTL = time.Duration(math.MaxInt64)
```
sethiay, line 216: multiply by time.Second ?

gargnitingoogle: No, StatOrTypeCacheTTL is time.Duration . So, no need to further multiply it with anything. `time.Duration(math.MaxInt64)` is the next duration represetable by time.Duration, ~290 years.

flags.go
```diff
@@ -29,7 +29,14 @@ import (
 )
 
 // Defines the max value supported by sequential-read-size-mb flag.
-const maxSequentialReadSizeMb = 1024
+const (
+	// maxSequentialReadSizeMb is the max value supported by sequential-read-size-mb flag.
+	maxSequentialReadSizeMb = 1024
+	// DefaultStatOrTypeCacheTTL is the default value used for
+	// stat-cache-ttl or type-cache if they have not been set
```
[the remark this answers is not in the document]
gargnitingoogle: Done.

internal/config/mount_config.go
```diff
@@ -14,6 +14,16 @@
...
 package config
 
+import "math"
+
+const (
+	// TtlInSecsUnset is the value internally set for metada-cache:ttl-secs
+	// when it is not set in the gcsfuse mount config file.
+	// The constant value has been chosen deliberately to
+	// to be improbable for a user to explicitly set.
+	TtlInSecsUnset int64 = math.MinInt64
```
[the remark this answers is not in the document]
gargnitingoogle: Done. Added as a suffix.

internal/config/yaml_parser_test.go
```diff
@@ -99,3 +104,18 @@ func (t *YamlParserTest) TestReadConfigFile_InvalidFileCacheMaxSizeConfig() {
...
+	AssertNe(nil, err)
+	AssertThat(err, oglematchers.Error(oglematchers.HasSubstr(MetadataCacheTtlSecsInvalidValueError)))
+}
+
+func (t *YamlParserTest) TestReadConfigFile_MetatadaCacheConfig_TtlNotSet() {
+	mountConfig, err := ParseConfigFile("testdata/metadata_cache_config_ttl-unset.yaml")
+
+	AssertEq(nil, err)
+	AssertNe(nil, mountConfig)
+	AssertEq(math.MinInt64, mountConfig.MetadataCacheConfig.TtlInSeconds)
```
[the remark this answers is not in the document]
gargnitingoogle: I had originally tried that only, but that causes a circular package dependency and I'll have move to variable (or possibly even code) around to make this work. So, I fell back to using the value directly. Let me know if you think I should move things around to accommodate this.

flags.go
```diff
@@ -29,7 +29,14 @@ import (
 )
 
 // Defines the max value supported by sequential-read-size-mb flag.
-const maxSequentialReadSizeMb = 1024
+const (
```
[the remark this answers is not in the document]
gargnitingoogle: Replaced all instances of `stat-cache-ttl` and `type-cache-ttl` with `metadata-cache:ttl-secs` in all places in gcsfuse repo except the code/tests needed for backward compatibility support.

internal/fs/caching_test.go
```diff
@@ -486,17 +486,6 @@ func (t *MultiBucketMountCachingTest) TearDown() {
 	for _, bucketName := range []string{bucket1Name, bucket2Name} {
 		bucket := buckets[bucketName]
 		AssertEq(nil, storageutil.DeleteAllObjects(context.Background(), bucket))
-
```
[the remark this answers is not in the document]
gargnitingoogle: This teardown function was added in my change for single-stat-cache across multiple buckets in dynamic-mount case. 
At the time of writing it, I took the approach of deleting the objects from both the GCS server and the mounted file-system, which is probably aggressive. 
I didn't see any failures at the time, but started seeing the `resource busy` errors during the current change test runs. 
I wasn't able to get details of which resource busy because the messages were probably coming from the kernel. But I was able to trace it to the MultiBucketMountCachingTest tests which invoked this function for tear-down. I suspected the double-deletion as the issue and it worked.

perfmetrics/scripts/ml_tests/pytorch/run_model.sh
```diff
@@ -29,9 +29,15 @@ cd -
 # Create a directory for gcsfuse logs
 mkdir  run_artifacts/gcsfuse_logs
 
+config_filename=/pytorch_dino/gcsfuse/gcsfuse-config.yml
```
gargnitingoogle, line 32: @sethiay PTAL

perfmetrics/scripts/read_cache/generate_yml_config.sh
```diff
@@ -29,4 +29,6 @@ cache-location: ${CACHE_LOCATION:-/tmp/read_cache/}
 file-cache:
   max-size-in-mb: ${MAX_SIZE_IN_MB:-100}
   download-file-for-random-read: ${DOWNLOAD_FOR_RANDOM_READ:-false}
+metadata-cache:
```
gargnitingoogle, line 32: @sethiay PTAL

internal/config/mount_config.go
```diff
@@ -14,6 +14,17 @@
...
 
+import "math"
+
+const (
+	// TtlInSecsUnsetSentinel is the value internally
+	// set for metada-cache:ttl-secs
+	// when it is not set in the gcsfuse mount config file.
+	// The constant value has been chosen deliberately
+	// to be improbable for a user to explicitly set.
+	TtlInSecsUnsetSentinel int64 = math.MinInt64
```
sethiay, line 25: we can just name it TtlInSecsSentinelValue ?

docs/semantics.md
```diff
@@ -82,14 +87,26 @@ Warning: Using stat caching breaks the consistency guarantees discussed in this
...
+To alleviate this, Cloud Storage FUSE supports a "type cache" on directory inodes. When type cache is enabled, each directory inode will maintain a mapping from the name of its children to whether those children are known to be files or directories or both. When a child is looked up, if the parent's cache says that the child is a file but not a directory, only one Cloud Storage object will need to be stated. Similarly if the child is a directory but not a file.
+
+The behavior of type cache is controlled by the following flags/config parameters:
+1. Type-cache capacity: The size of the type cache is **not configurable**. By default the type cache will hold up to 32768 items per directory. If you have folders containing more than 32768 items (folders or files) you may want to increase this, otherwise the caching will not function properly when listing that folder's contents:
```
sethiay, line 100: can we remove /1 for now and directly add whenever we make the type cache configurable ?

docs/semantics.md
```diff
@@ -82,14 +87,26 @@ Warning: Using stat caching breaks the consistency guarantees discussed in this
...
+    - Because there are more objects than cache capacity, the earliest entries will be evicted
+    - The linux kernel then asks for a little more information on each file.
+    - As the earliest cache entries were evicted, this is a fresh List request
+    - This cycle repeats and sends a List request for every page in the folder, as though caching were disabled
+
+2. Type-cache TTL: It controls the duration for which Cloud Storage FUSE allows the kernel to inode type attributes. It can be set in one of the following two ways.
+   1. ```metadata-cache: ttl-secs``` in the config-file. This is set as an integer, which sets the TTL in seconds. If this is -1, TTL is taken as infinite i.e. no-TTL based expirations of entries. If this is 0, that disables the type-cache. If this is <-1, then an error is thrown on mount.
+   2. ```--type-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This will be deprecated in a future version and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--type-cache-ttl``` is ignored.
```
sethiay, line 109: can we also add a note somewhere common to both stat and type cache that
(a) "metadata-cache:ttl-secs" is the recommended way to set TTL. and stat-cache-ttl and type-cache-ttl flags will be deprecated in future releases.
(b) minimum of stat-cache-ttl and type-cache-ttl is used as TTL when metadata-cache: ttl-secs is not passed.

sethiay (commented):
Minor comments.

internal/config/mount_config.go
gargnitingoogle: The word `sentinel` by itself doesn't mean missing/unset value. It means a special value indicating the end of a sequence ([wikipedia](https://en.wikipedia.org/wiki/Sentinel_value)), so replacing `Unset` with `Sentinel` will be misleading. If there is a golang convention in such cases, let me know.

sethiay: Wikipedia says it is end of sequence but that is too strict (it's subjective/debatale): https://stackoverflow.com/questions/56760547/what-is-the-term-for-a-sentinel-that-does-not-denote-the-end-of-a-sequence.
Why I suggested to not use Unset in the name is because sentinal often means a signal value for something. In our case, it's a signal that ttl is not set. 

However, there is no harm in using Unset in the name. So I am fine with that.

docs/semantics.md
gargnitingoogle: Removed.

docs/semantics.md
gargnitingoogle: It's there in the caching section above. It was added in the previous PR. I'll move it down here so make it makes more sense.

internal/config/yaml_parser_test.go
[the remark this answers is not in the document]
sethiay: the constant TtlInSecsUnsetSentinel is in the same package, so i wonder we should be able to use it ?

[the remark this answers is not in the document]
gargnitingoogle: Not sure what changed. But just now when I tried, it worked.

gargnitingoogle:
> Minor comments.

All resolved from my end.

sethiay (approved):

PR #1546, opened 2023-12-13
Add metadata-cache:type-cache-max-size-mb-per-dir config

gargnitingoogle:
### Description

This adds on top of https://github.com/GoogleCloudPlatform/gcsfuse/pull/1545 .

This change adds the following in GCSFuse config file.

```
metadata-cache: # pre-existing
	type-cache-max-size-mb-per-dir: # new
```

This makes the max size of type-cache configurable at per directory level.

This change adds on top of https://github.com/GoogleCloudPlatform/gcsfuse/pull/1545 .

NA

### Testing details
1. Manual - Did sanity testing.
2. Unit tests - Ran locally. Added new tests.
3. Integration tests - Through presubmits.

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -95,7 +95,14 @@ The negative result for ```foo/``` will be cached, but that only helps with the 
 To alleviate this, Cloud Storage FUSE supports a "type cache" on directory inodes. When type cache is enabled, each directory inode will maintain a mapping from the name of its children to whether those children are known to be files or directories or both. When a child is looked up, if the parent's cache says that the child is a file but not a directory, only one Cloud Storage object will need to be stated. Similarly if the child is a directory but not a file.
 
 The behavior of type cache is controlled by the following flags/config parameters:
-1.  Type-cache TTL: It controls the duration for which Cloud Storage FUSE allows the kernel to inode type attributes. It can be set in one of the following two ways.
+1. Type-cache capacity: This is configurable at per-directory level by setting `metadata-cache: type-cache-max-size-mb-per-dir` in config-file. This is the maximum size of type-cache per-directory in MiBs. By default, this is set at 16, which roughly equates to about a million entries. If you have folders containing more than a million items (folders or files) you may want to increase this, otherwise the caching will not function properly when listing that folder's contents:
+    - ListObjects will return information on the items within the folder. Each item's data is cached
+    - Because there are more objects than cache capacity, the earliest entries will be evicted
+    - The linux kernel then asks for a little more information on each file.
+    - As the earliest cache entries were evicted, this is a fresh List request
+    - This cycle repeats and sends a List request for every page in the folder, as though caching were disabled
+
+2.  Type-cache TTL: It controls the duration for which Cloud Storage FUSE allows the kernel to inode type attributes. It can be set in one of the following two ways.
     * ```metadata-cache: ttl-secs``` in the config-file. This is set as an integer, which sets the TTL in seconds. If this is -1, TTL is taken as infinite i.e. no-TTL based expirations of entries. If this is 0, that disables the type-cache. If this is <-1, then an error is thrown on mount.
     * ```--type-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This will be deprecated in a future version and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--type-cache-ttl``` is ignored.
 
--- a/internal/config/mount_config.go
+++ b/internal/config/mount_config.go
@@ -23,12 +23,10 @@ const (
 	// The constant value has been chosen deliberately
 	// to be improbable for a user to explicitly set.
 	TtlInSecsUnsetSentinel int64 = math.MinInt64
-	// TypeCacheMaxSizeInMbPerDirectoryUnset is the value internally set for
+	// DefaultTypeCacheMaxSizeInMbPerDirectory is the value internally set for
 	// metadata-cache:type-cache-max-size-mb-per-dir
 	// when it is not set in the gcsfuse mount config file.
-	// The constant value has been chosen deliberately to
-	// to be improbable for a user to explicitly set.
-	TypeCacheMaxSizeInMbPerDirectoryUnset int = math.MinInt
+	DefaultTypeCacheMaxSizeInMbPerDirectory int = 16
 )
 
 type WriteConfig struct {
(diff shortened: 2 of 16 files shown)


internal/fs/inode/dir_test.go
```diff
@@ -46,6 +46,7 @@ const dirInodeID = 17
 const dirInodeName = "foo/bar/"
 const dirMode os.FileMode = 0712 | os.ModeDir
 const typeCacheTTL = time.Second
+const typeCacheMaxSizeMbPerDirectory = 16
```
sethiay, line 49: is it possible to write unit tests here for type cache working ?

flags.go
```diff
@@ -545,10 +550,6 @@ func validateFlags(flags *flagStorage) (err error) {
 		err = fmt.Errorf("client protocol: %s is not valid", flags.ClientProtocol)
 	}
 
-	if flags.StatCacheTTL != flags.TypeCacheTTL {
```
bjornleffler: Why is this an error?

internal/util/util.go
```diff
@@ -80,3 +80,15 @@ func ResolveConfigFilePaths(config *config.MountConfig) (err error) {
...
 	return
 }
+
+// MibToBytes returns the bytes equivalent
+// of given no.s of MiBs (Mibi Bytes).
+// For reference, each MiB = 2^20 bytes.
+// It supports only upto 2^44-1 MiBs (~4 Tebi MiBs, or ~4 Ebi bytes)
+// as inputs, and panics for higher inputs.
+func MibToBytes(mib uint64) uint64 {
+	if mib > 0xFFFFFFFFFFF {
```
bjornleffler, line 90: Use some pre-defined constant instead of a weird hex constant. MAX_UINT64 or something.

gargnitingoogle: There is no standard golng constant for this. I've defined constant MaxMiBsInMaxUint64 above for it now.

internal/fs/inode/type_cache.go
```diff
@@ -50,28 +57,30 @@ type typeCache struct {
...
 ////////////////////////////////////////////////////////////////////////
 
-// Panic if any internal invariants have been violated. The careful user can
-// arrange to call this at crucial moments.
-func (tc *typeCache) CheckInvariants() {
-	tc.entries.CheckInvariants()
-}
-
 // Insert inserts a record to the cache.
 func (tc *typeCache) Insert(now time.Time, name string, it Type) {
```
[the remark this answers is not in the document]
gargnitingoogle: Done. PTAL.

internal/util/util.go
```diff
@@ -80,3 +80,15 @@ func ResolveConfigFilePaths(config *config.MountConfig) (err error) {
 	}
 	return
 }
+
+// MibToBytes returns the bytes equivalent
+// of given no.s of MiBs (Mibi Bytes).
+// For reference, each MiB = 2^20 bytes.
+// It supports only upto 2^44-1 MiBs (~4 Tebi MiBs, or ~4 Ebi bytes)
+// as inputs, and panics for higher inputs.
+func MibToBytes(mib uint64) uint64 {
```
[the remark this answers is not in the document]
gargnitingoogle: Never mind. renamed it to MiBsToBytes which solves both our problems.

internal/config/mount_config.go
```diff
@@ -22,6 +22,12 @@ const (
 	// The constant value has been chosen deliberately to
 	// to be improbable for a user to explicitly set.
 	TtlInSecsUnset int64 = math.MinInt64
+	// TypeCacheMaxSizeInMbPerDirectoryUnset is the value internally set for
+	// metadata-cache:type-cache-max-size-mb-per-dir
+	// when it is not set in the gcsfuse mount config file.
+	// The constant value has been chosen deliberately to
+	// to be improbable for a user to explicitly set.
+	TypeCacheMaxSizeInMbPerDirectoryUnset int = math.MinInt
```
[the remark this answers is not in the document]
gargnitingoogle: It was for additional logging, whether default value was provided by user, or applied by us because of unset value. Have removed it now.

internal/fs/inode/type_cache.go
```diff
@@ -50,47 +58,58 @@ type typeCache struct {
 	//
 	// INVARIANT: entries.CheckInvariants() does not panic
 	// INVARIANT: Each value is of type cacheEntry
-	entries lrucache.Cache
```
bjornleffler, line 53: Is this a major change? Switching
"github.com/jacobsa/util/lrucache"
to 
"github.com/googlecloudplatform/gcsfuse/internal/cache/lru"?

If so, it should be documented.

docs/semantics.md
```diff
@@ -95,7 +95,14 @@ The negative result for ```foo/``` will be cached, but that only helps with the
...
 
 The behavior of type cache is controlled by the following flags/config parameters:
-1.  Type-cache TTL: It controls the duration for which Cloud Storage FUSE allows the kernel to inode type attributes. It can be set in one of the following two ways.
+1. Type-cache capacity: This is configurable at per-directory level by setting `metadata-cache: type-cache-max-size-mb-per-dir` in config-file. This is the maximum size of type-cache per-directory in MiBs. By default, this is set at 16, which roughly equates to about a million entries. If you have folders containing more than a million items (folders or files) you may want to increase this, otherwise the caching will not function properly when listing that folder's contents:
```
bjornleffler, line 98: 1. It should be very rare to have 1M "files" in a single "directory".
2. The type-cache capacity does not limit overall memory usage, as that will depend on the number of directories.

So for the vast majority of use cases, this change does nothing. Why are we adding code that will never be used?

internal/fs/inode/type_cache.go
gargnitingoogle: > Is this a major change?

In terms of performance, no. In terms of functionality, a minor change. `jacobsa/util/lrucache` was defined in terms of capacity i.e. upper-limit on the number of entries, but `gcsfuse/internal/cache/lru` is limited by size in MBs. This is to make it consistent with the other cache types (file-read-cache and stat-cache) which also takes sizes instead of capacity as limits.

> If so, it should be documented.

Based on the above, do you think that there is a need to document a change here ?
The change from capacity to size is a user-interface-level change, and that'll be documented.

gargnitingoogle: Discussed this offline with @bjornleffler and resolved this issue.

docs/semantics.md
gargnitingoogle: Discussed this offline with @bjornleffler and resolved this issue.

internal/gcsx/bucket_manager.go
```diff
@@ -203,7 +203,7 @@ func (bm *bucketManager) SetUpBucket(
 	}
 
 	// Enable cached StatObject results, if appropriate.
-	if bm.config.StatCacheTTL != 0 && bm.sharedStatCache != nil {
+	if bm.config.StatCacheTTL != 0 && bm.sharedStatCache != nil && bm.config.StatCacheCapacity != 0 {
```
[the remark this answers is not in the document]
gargnitingoogle: Obsolete comment.

gargnitingoogle:
The changes here have been included in https://github.com/GoogleCloudPlatform/gcsfuse/pull/1560 which is a super-set of these changes. Please continue reviewing there.

PR #1554, opened 2023-12-20
Change LRU in file cache on every request from kernel

sethiay:
### Description
Change LRU in file cache on every request from kernel. Also replaced file.Seek + io.ReadFull with file.ReadAt along with a check that number of bytes read into dst buffer is equal to requested number of bytes.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added for LRU order.
3. Integration tests - Ran existing integration tests [here](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/28180e6a-2ac8-4dfe-9ba2-dc092fbf03a9)

### DIFF
--- a/internal/cache/file/cache_handle.go
+++ b/internal/cache/file/cache_handle.go
@@ -120,11 +120,11 @@ func (fch *CacheHandle) checkEntryInFileInfoCache(bucket gcs.Bucket, object *gcs
 		return err
 	}
 	fileInfoData := fileInfo.(data.FileInfo)
-	if requiredOffset > int64(fileInfoData.Offset) {
-		err = fmt.Errorf("%v: required offset: %v is greater than offset in cache: %v", util.FallbackToGCSErrMsg, requiredOffset, fileInfoData.Offset)
-		return err
-	} else if fileInfoData.ObjectGeneration != object.Generation {
+	if fileInfoData.ObjectGeneration != object.Generation {
 		err = fmt.Errorf("%v: generation of cached object: %v is different from required generation: %v", util.InvalidFileInfoCacheErrMsg, fileInfoData.ObjectGeneration, object.Generation)
+		return err
+	} else if requiredOffset > int64(fileInfoData.Offset) {
+		err = fmt.Errorf("%v: required offset: %v is greater than offset in cache: %v", util.FallbackToGCSErrMsg, requiredOffset, fileInfoData.Offset)
 		return err
 	}
 
--- a/internal/cache/file/cache_handle.go
+++ b/internal/cache/file/cache_handle.go
@@ -94,9 +94,8 @@ func (fch *CacheHandle) shouldReadFromCache(jobStatus *downloader.JobStatus, req
 }
 
 // checkEntryInFileInfoCache checks if entry is present for a given object in
-// file info cache with same generation and offset at least equal to
-// requiredOffset. It returns nil if entry is present, otherwise returns an error
-// with appropriate message.
+// file info cache with same generation. It returns nil if entry is present,
+// otherwise returns an error with appropriate message.
 func (fch *CacheHandle) checkEntryInFileInfoCache(bucket gcs.Bucket, object *gcs.MinObject, requiredOffset int64) error {
 	fileInfoKey := data.FileInfoKey{
 		BucketName: bucket.Name(),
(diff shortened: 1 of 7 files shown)


internal/cache/file/cache_handle.go
```diff
@@ -148,28 +149,47 @@ func (fch *CacheHandle) Read(ctx context.Context, object *gcs.MinObject, offset
...
+	n, err = fch.fileHandle.ReadAt(dst, offset)
+	requestedNumBytes := int(requiredOffset - offset)
 	// Returns io.ErrUnexpectedEOF when buffer size is more than content to be read.
 	if err == io.EOF || err == io.ErrUnexpectedEOF {
 		err = nil
 	}
 	if err != nil {
 		errMsg := fmt.Sprintf("%s: while reading from %d offset of the local file: %v", util.ErrInReadingFileHandleMsg, offset, err)
 		return 0, errors.New(errMsg)
+	} else if n != requestedNumBytes {
```
raj-prince, line 161: Do we know, when does it happen?

sethiay: No, I added this check because if this happens, we will never know. And if this never happens, then also there is no harm in having simple additional check.

sethiay: one thing that I thought but didn't created the scenario - if a file is truncated externally such that it contains data till offset + x but x is less than len(dst)

internal/cache/file/cache_handle.go
```diff
@@ -148,28 +149,47 @@ func (fch *CacheHandle) Read(ctx context.Context, object *gcs.MinObject, offset
...
+		return 0, errors.New(errMsg)
 	}
 
-	// The job state may change while reading data from the local downloaded file. This may be
-	// due to a failure in the download job, or cache eviction due to another file. In this case,
-	// we will check the job state again after reading the data, and fall back to the normal flow
-	// if job status is invalid or failed.
-	jobStatus = fch.fileDownloadJob.GetStatus()
-	if err = fch.shouldReadFromCache(&jobStatus, requiredOffset); err != nil {
+	fileInfoKey := data.FileInfoKey{
```
raj-prince, line 166: Can we create a helper method for all these logic?

internal/cache/file/cache_handle.go
```diff
@@ -148,28 +149,47 @@ func (fch *CacheHandle) Read(ctx context.Context, object *gcs.MinObject, offset
 	}
 
 	// We are here means, we have the data downloaded which kernel has asked for.
-	_, err = fch.fileHandle.Seek(offset, 0)
```
raj-prince, line 151: Please update the PR-description with this change.

internal/cache/file/cache_handle.go
raj-prince: I think, it's good to document this case around the check.

raj-prince (changes requested):
Few changes...

internal/cache/file/cache_handle.go
```diff
@@ -148,28 +187,27 @@ func (fch *CacheHandle) Read(ctx context.Context, object *gcs.MinObject, offset
...
+	// dst buffer has fixed size of 1 MiB even when the offset is such that
+	// offset + 1 MiB > object size. In that case, io.ErrUnexpectedEOF is thrown
+	// which should be ignored.
 	if err == io.EOF || err == io.ErrUnexpectedEOF {
 		err = nil
 	}
 	if err != nil {
 		errMsg := fmt.Sprintf("%s: while reading from %d offset of the local file: %v", util.ErrInReadingFileHandleMsg, offset, err)
 		return 0, errors.New(errMsg)
+	} else if n != requestedNumBytes {
```
raj-prince, line 201: Sorry for the to and fro, I think, this check should be inside `line:195 if block`?

internal/cache/file/cache_handle.go
```diff
@@ -92,12 +93,50 @@ func (fch *CacheHandle) shouldReadFromCache(jobStatus *downloader.JobStatus, req
...
+	// The offset and generation checks below are required because it may happen
+	// that file being read is evicted from cache during or after reading the
+	// required offset from local cached file to `dst` buffer in previous code
+	// block.
+	if fileInfo == nil {
+		err = fmt.Errorf("%v: no entry found in file info cache for key %v", util.InvalidFileInfoCacheErrMsg, fileInfoKeyName)
+		return err
+	}
+	fileInfoData := fileInfo.(data.FileInfo)
+	if requiredOffset > int64(fileInfoData.Offset) {
```
raj-prince, line 123: We call this method after reading from the file in cache. Now, if this condition is true that means corresponding job and fileInfoEntry is not consistent. In this case, we should close the fileHandle and then fallback to GCS. For that we need to return different error other than FallBackToGCS.

As randomReader doesn't close cacheHandle for FallBackToGCSErrMsg.

sethiay: As discussed offline, we don't need comparison of offsets as we are already comparing the number of bytes read with numBytesRequested in previous code.

raj-prince (approved):

PR #1559, opened 2023-12-22
[DontReview] Fix behavior for boundary values of stat cache capacity

gargnitingoogle:
### Description
Handle stat-cache-capacity 0 or -1

It does the following.
1. If stat-cache-capacity is passed as -1, it uses
   math.Uint64 as size to make size effectively infinite.
2. If stat-cache-capacity is passed as 0, then the shared
   stat-cache is not 0, which will effectively disable
   it by not creating fastStatBucket for bucket (in case of
   static-mount), or for any of the buckets (in case of dynamic-mount).

NA

### Testing details
1. Manual - Did sanity testing
2. Unit tests - Ran locally, all passing. Adding one new test.
3. Integration tests - Ran through presubmits.

### DIFF
--- a/internal/gcsx/bucket_manager.go
+++ b/internal/gcsx/bucket_manager.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"path"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/gcsx/bucket_manager.go
```diff
@@ -82,19 +83,31 @@ type bucketManager struct {
...
 }
 
-func NewBucketManager(config BucketConfig, storageHandle storage.StorageHandle) BucketManager {
+func createStatCacheLru(statCacheCapacity int) *lru.Cache {
+	var statCacheSize uint64
+	if statCacheCapacity > 0 {
+		// This conversion from capacity to size will not be needed when
+		/// StatCacheCapacity is replaced with StatCacheSizeMB
+		statCacheSize = uint64(statCacheCapacity) * metadata.StatCacheEntrySize()
+	} else if statCacheCapacity == -1 {
```
sethiay, line 92: We can live without supporting -1 for stat-cache-capacity flag because we currently don't support -1 and we are anyways going to deprecate these flags in future.

internal/gcsx/bucket_manager_test.go
```diff
@@ -42,6 +42,30 @@ func (t *BucketManagerTest) TearDown() {
...
+			capacity: -1,
+			isNil:    false,
+		},
+		{
+			capacity: 10,
+			isNil:    false,
+		},
+	}
+
+	for _, input := range inputs {
```
ankitaluthra1, line 64: @gargnitingoogle @sethiay Can someone help me understand a scenario... will this test pass if we add capacity as 10 and output from `createStatCacheLru` is returned 5 (hypothetically if there is bug in the method), if yes, then will it be good idea to test exact output values in three separate asserts or is this the way we should do in GoLang?

internal/gcsx/bucket_manager.go
gargnitingoogle: Removed.

internal/gcsx/bucket_manager_test.go
gargnitingoogle: createStatCacheLru return a lru.Cache instance given the size. 
If size is <=0, it should return nil. 
For size>0, a non-nil object should be returned. But there is no easy way to verify that the return object has the passed-in size as it has no interface for that. So, we're testing only for whether the object got created.
More detailed tests for statCache are in https://github.com/googlecloudplatform/gcsfuse/blob/5c8ef1cd1ed4572033b6c47e5297141d3f2bd64c/internal/cache/metadata/stat_cache_test.go .

gargnitingoogle:
This is an outdated PR. Closing it.

PR #1560, opened 2023-12-22
Make type-cache mount-level

gargnitingoogle:
### Description
This builds on top of https://github.com/GoogleCloudPlatform/gcsfuse/pull/1546 .

It switches type-cache from being per-directory level to mount-level.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


gargnitingoogle:
Integration tests passed: [log](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/23b9f6bc-5fea-467c-aac8-2c153b0c77b8/log)

internal/fs/inode/type_cache.go
```diff
@@ -15,16 +15,34 @@
...
+	"github.com/googlecloudplatform/gcsfuse/internal/cache/lru"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
+	"github.com/googlecloudplatform/gcsfuse/internal/util"
 )
 
+// TypeCache is interface to expose typeCache
+// outside of the inode package, to allow
+// it to be shared at the mount-level.
+type TypeCache interface {
+	Insert(now time.Time, name string, it Type)
```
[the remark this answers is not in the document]
gargnitingoogle: Done.

internal/fs/inode/type_cache.go
```diff
@@ -50,58 +87,126 @@ type typeCache struct {
...
-	// Are we disabled?
-	if tc.ttl == 0 {
-		return
+	if tc.entries != nil {
+		_, err := tc.entries.Insert(name, cacheEntry{
+			expiry:    now.Add(tc.ttl),
+			inodeType: it,
+		})
+		if err != nil {
+			panic(fmt.Errorf("failed to insert entry in typeCache: %v", err))
```
gargnitingoogle, line 125: Not sure if this error should be ignored, or should panic here.

internal/config/mount_config.go
```diff
@@ -49,6 +53,10 @@ type MetadataCacheConfig struct {
 	// no cache and > 0 for ttl-controlled metadata-cache.
 	// Any value set below -1 will throw an error.
 	TtlInSeconds int64 `yaml:"ttl-secs,omitempty"`
+	// TypeCacheMaxEntries is the upper limit
+	// on the maximum size of type-cache map,
+	// which is maintained at gcsfuse-mount-level.
+	TypeCacheMaxSizeMb int `yaml:"type-cache-max-size-mb,omitempty"`
```
gargnitingoogle, line 59: Define and implement upper-limit on the value of type-cache-max-size-mb .

gargnitingoogle:
Integration tests are currently failing because of the failure detailed in https://github.com/GoogleCloudPlatform/gcsfuse/pull/1593#issuecomment-1884816287 .

internal/fs/fs.go
```diff
@@ -926,7 +947,18 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 	getLookupResult := func() (*inode.Core, error) {
 		parent.Lock()
 		defer parent.Unlock()
-		return parent.LookUpChild(ctx, childName)
+
+		core, err := parent.LookUpChild(ctx, childName)
+		if err == nil && parent.IsBaseDirInode() {
```
gargnitingoogle, line 952: @sethiay as discussed offline, this code is specific to multi-bucket mount. This is creating (if not already created) and inserting a type-cache-bucket-view for this bucket (called childName) into the (bucketName -> type-cache-bucket-view') map in fs. This is the code which comes up when a new bucket is mounted in dynamic-mount.

docs/semantics.md
```diff
@@ -95,7 +95,16 @@ The negative result for ```foo/``` will be cached, but that only helps with the
...
 
 The behavior of type cache is controlled by the following flags/config parameters:
-1.  Type-cache TTL: It controls the duration for which Cloud Storage FUSE allows the kernel to inode type attributes. It can be set in one of the following two ways.
+1. Type-cache size: This is configurable at mount-level level by setting `metadata-cache: type-cache-max-size-mb` in config-file. This is the maximum size of type-cache in MiBs across a GCSFuse mount. By default, this is set at 32, which roughly equates to about a million entries. If you have more than a million objects (including files, directories, symlinks) across your bucket/buckets (one bucket in case of static mount and, multiple buckets in case of dynamic-mount), that you are going to access through this mount, then you may want to set this higher, otherwise the caching will not function properly when listing the contents:
```
gargnitingoogle, line 98: Typo. remove duplicate word `level`

internal/fs/inode/type_cache.go
sethiay: yes, we should panic simlar to how we are doing in statcache.

internal/fs/inode/type_cache.go
```diff
@@ -50,58 +87,126 @@ type typeCache struct {
...
-	tc.entries.CheckInvariants()
+	return &typeCache{}
 }
 
 // Insert inserts a record to the cache.
 func (tc *typeCache) Insert(now time.Time, name string, it Type) {
-	// Are we disabled?
-	if tc.ttl == 0 {
-		return
+	if tc.entries != nil {
```
sethiay, line 119: please add comment saying that this check for disabled/enabled type cache.
Add the comment in Erase and Get functions as well.

internal/fs/inode/type_cache.go
```diff
@@ -50,58 +87,126 @@ type typeCache struct {
...
+		_, err := tc.entries.Insert(name, cacheEntry{
+			expiry:    now.Add(tc.ttl),
+			inodeType: it,
+		})
+		if err != nil {
+			panic(fmt.Errorf("failed to insert entry in typeCache: %v", err))
+		}
+		logger.Debugf("TypeCache: Inserted %s as %s", name, it.String())
```
sethiay, line 127: this will increase the size of debug logs a lot and we haven't planned functional testing so far for metadata cache, so please remove this and other debug statements in type cache.

internal/fs/inode/type_cache.go
```diff
@@ -50,58 +87,126 @@ type typeCache struct {
 	//
 	// INVARIANT: entries.CheckInvariants() does not panic
```
sethiay, line 88: we have removed the first invariant check, right ?

internal/fs/inode/type_cache.go
```diff
@@ -50,58 +87,126 @@ type typeCache struct {
...
+
+////////////////////////////////////////////////////////////////////////
+// Public interface
+////////////////////////////////////////////////////////////////////////
+
+// Creates a new typeCacheBucketView wrapping over the
+// given TypeCache and prefix bucketName for prepending
+// in operational keys.
+// This is needed in case of multi-bucket mount (i.e. dynamic-mount).
+func NewTypeCacheBucketView(stc TypeCache, bn string) TypeCache {
```
sethiay, line 192: can we use sharedTypeCache and bucketName instead of stc and bn?

internal/fs/inode/type_cache.go
```diff
@@ -50,58 +87,126 @@ type typeCache struct {
...
+	if stc == nil {
+		panic("The passed shared-type-cache is nil")
+	}
+	return &typeCacheBucketView{sharedTypeCache: stc, bucketName: bn}
+}
+
+// Insert inserts a record to the cache.
```
sethiay, line 199: nit: remove these comments as we have already added them above and the name is implicit

internal/fs/inode/type_cache.go
```diff
@@ -15,16 +15,53 @@
 package inode
 
 import (
+	"fmt"
```
sethiay, line 18: can we move this package to internal/cache/metadata ? similar to how we did for stat cache ?

internal/util/util.go
```diff
@@ -92,3 +95,15 @@ func Stringify(input any) string {
 	}
 	return string(inputBytes)
 }
+
+// MiBsToBytes returns the bytes equivalent
+// of given no.s of MiBs (Mibi Bytes).
```
sethiay, line 100: nit: numbers

internal/fs/inode/core.go
```diff
@@ -32,6 +32,25 @@ var (
 	NonexistentType Type = 5
 )
 
+func (t Type) String() string {
```
sethiay, line 35: we can remove this if we don't add debug logs for type cache ?

docs/semantics.md
sethiay, line 98: let's only write 
1. Type-cache size: This is configurable at mount-level level by setting `metadata-cache: type-cache-max-size-mb` in config-file. This is the maximum size of type-cache in MiBs across a GCSFuse mount. By default, this is set at 32, which roughly equates to about a million entries.

The other information should go in user recommendation doc about configuring type and stat cache.

docs/semantics.md
```diff
@@ -95,7 +95,16 @@ The negative result for ```foo/``` will be cached, but that only helps with the
 To alleviate this, Cloud Storage FUSE supports a "type cache" on directory inodes. When type cache is enabled, each directory inode will maintain a mapping from the name of its children to whether those children are known to be files or directories or both. When a child is looked up, if the parent's cache says that the child is a file but not a directory, only one Cloud Storage object will need to be stated. Similarly if the child is a directory but not a file.
```
sethiay, line 95: can we include "metadata-cache:type-cache-max-size-mb" in existing e2e tests? IMO, we should include in the following packages - operations, implicit-dir, explicit-dir, local-file, readonly.

Should be added similar to https://github.com/GoogleCloudPlatform/gcsfuse/blob/read_cache_release/tools/integration_tests/operations/operations_test.go#L105

sethiay (commented):
This PR is too long and creates lot of files to review and should be split into 2-3 PRs. Please take care for new PRs.

I haven't reviewed internal/fs/inode/dir_test.go and internal/fs/inode/type_cache_test.go yet.

docs/semantics.md
sethiay: Also add average size for an entry in the doc in brackets. 
Also explain here how you found average size of an entry

internal/fs/inode/type_cache.go
gargnitingoogle: Right. that line of comment was obsolete now. Removed.

docs/semantics.md
gargnitingoogle: Done.

internal/fs/fs.go
gargnitingoogle: done

internal/fs/fs.go
```diff
@@ -189,6 +189,8 @@ func NewFileSystem(
 		mountConfig:                cfg.MountConfig,
 		fileCacheHandler:           fileCacheHandler,
 		cacheFileForRangeRead:      cfg.MountConfig.FileCacheConfig.CacheFileForRangeRead,
+		sharedTypeCache:            inode.NewTypeCache(cfg.MountConfig.MetadataCacheConfig.TypeCacheMaxSizeMb, cfg.DirTypeCacheTTL),
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/fs/inode/core.go
gargnitingoogle: No, it also helps during debugging using and IDE or even using Delve, by displaying the name instead of 0,1,2,3 etc.

internal/fs/fs.go
```diff
@@ -926,7 +947,18 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 	getLookupResult := func() (*inode.Core, error) {
 		parent.Lock()
 		defer parent.Unlock()
-		return parent.LookUpChild(ctx, childName)
+
+		core, err := parent.LookUpChild(ctx, childName)
+		if err == nil && parent.IsBaseDirInode() {
+			if _, ok := fs.typeCacheBucketViews[childName]; !ok {
```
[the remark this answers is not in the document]
gargnitingoogle: added lock/unlock

internal/config/yaml_parser_test.go
```diff
@@ -158,3 +159,18 @@ func (t *YamlParserTest) TestReadConfigFile_MetatadaCacheConfig_TtlNotSet() {
 	AssertNe(nil, mountConfig)
 	AssertEq(TtlInSecsUnsetSentinel, mountConfig.MetadataCacheConfig.TtlInSeconds)
 }
+
+func (t *YamlParserTest) TestReadConfigFile_MetatadaCacheConfig_InvalidTypeCacheSize() {
+	_, err := ParseConfigFile("testdata/metadata_cache_config_invalid_type-cache-size.yaml")
+
+	AssertNe(nil, err)
+	AssertThat(err, oglematchers.Error(oglematchers.HasSubstr(TypeCacheMaxSizeMbInvalidValueError)))
```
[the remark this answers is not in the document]
gargnitingoogle: Done

PR #1566, opened 2023-12-27
cache hit log changes

ashmeenkaur:
### Description
This PR includes code changes such that the first read to a file is a cache miss. It also ensures that if a user has to wait for the cache to be populated, it is considered a cache miss.

NA

### Testing details
1. Manual - Manually verified from logs that the first read call is resulting in cache miss.
2. Unit tests - Updated existing unit tests and added test for multiple-reads of size < chunk size.
3. Integration tests - Ran successfully


internal/cache/file/cache_handle_test.go
```diff
@@ -636,8 +650,33 @@ func (cht *cacheHandleTest) Test_Read_FileInfoGenerationChanged() {
...
+	// First read should be cache miss.
+	n, cacheHit, err := cht.cacheHandle.Read(context.Background(), cht.bucket, cht.object, offset, dst)
+
+	jobStatus := cht.cacheHandle.fileDownloadJob.GetStatus()
+	ExpectGe(jobStatus.Offset, offset)
+	ExpectEq(n, ReadContentSize)
+	cht.verifyContentRead(offset, dst[:n])
+	ExpectFalse(cacheHit)
+	ExpectEq(nil, err)
+
```
sethiay, line 676: nit: to maintain consistency with other tests, let's give only one linespace. you can give after calling the second read and before the last three expect statements.

internal/gcsx/random_reader.go
```diff
@@ -256,14 +255,15 @@ func (rr *randomReader) ReadAt(
...
+	// GCS, cached in file and then served from that file. But the cacheHit is
+	// false in that case.
 	n, cacheHit, err = rr.tryReadingFromFileCache(ctx, p, offset)
 	if err != nil {
 		err = fmt.Errorf("ReadAt: while reading from cache: %v", err)
 		return
 	}
-	if cacheHit {
+	// data was served from cache.
+	if n == len(p) {
```
sethiay, line 266: kernel always send 1mib buffer, so a case where offset + 1 mib < object size, this check will fail.
So, can we keep additional check to compare like n == len(p) || (n < len(p) && offset + n == rr.object.Size) ?

internal/gcsx/random_reader_test.go
```diff
@@ -916,7 +915,7 @@ func (t *RandomReaderTest) Test_ReadAt_CacheHitNextToCacheMissIncaseOfInvalidJob
 	buf := make([]byte, objectSize)
 	_, cacheHit, err := t.rr.ReadAt(buf, 0)
 	AssertEq(nil, err)
-	AssertTrue(cacheHit)
+	AssertFalse(cacheHit)
```
sethiay, line 918: Can we rename this test to something like Test_ReadAt_CachePopulatedAndThenCacheMissDueToInvalidJob
Please do similar for Test_ReadAt_CacheHitNextToCacheMissIncaseOfInvalidFileHandle

internal/gcsx/random_reader_test.go
```diff
@@ -1050,7 +1049,7 @@ func (t *RandomReaderTest) Test_tryReadingFromFileCache_CacheHit() {
 
 	_, cacheHit, err := t.rr.wrapped.tryReadingFromFileCache(t.rr.ctx, buf, 0)
```
sethiay, line 1050: please make one more read call to show cacheHit in this case as the test case is to show cache hit

internal/gcsx/random_reader_test.go
```diff
@@ -708,7 +707,7 @@ func (t *RandomReaderTest) Test_ReadAt_SequentialFullObject() {
 
 	_, cacheHit, err := t.rr.ReadAt(buf, 0)
 
-	ExpectTrue(cacheHit)
+	ExpectFalse(cacheHit)
```
sethiay, line 710: since these tests are for cache, we need to show here that reading populates cache. So, we need to make another read request for cacheHit to become true. Please also make similar changes for other test cases where the intention is to show cacheHit.

sethiay (commented):
Looks good overall, few comments for test cases.

internal/gcsx/random_reader_test.go
```diff
@@ -938,7 +937,7 @@ func (t *RandomReaderTest) Test_ReadAt_CacheHitNextToCacheMissIncaseOfInvalidJob
 	_, cacheHit, err = t.rr.ReadAt(buf, 0)
 
 	ExpectEq(nil, err)
-	ExpectTrue(cacheHit)
+	ExpectFalse(cacheHit)
```
ashmeenkaur, line 940: @sethiay I am not sure if this should be a cache miss. Can you check once?

sethiay: i think it should be a cacheMiss because during rc2 read (2nd t.rr.ReadAt) call, the data is served from GCS and cache handle is closed as cache will give error because we have invalidated the cache before that. 
it is during the rc3 call that cache is populated again with new cache handle and because it is populated so should be a cache miss.

ashmeenkaur: Thanks for the explanation!

internal/gcsx/random_reader_test.go
```diff
@@ -796,15 +803,20 @@ func (t *RandomReaderTest) Test_ReadAt_RandomReadNotStartWithZeroOffsetWhenDownl
...
 	_, cacheHit, err := t.rr.ReadAt(buf, int64(start))
-
 	ExpectFalse(cacheHit)
 	ExpectEq(nil, err)
 	ExpectTrue(reflect.DeepEqual(testContent[start:end], buf))
 	job := t.jobManager.GetJob(t.object, t.bucket)
 	jobStatus := job.GetStatus()
 	ExpectTrue(jobStatus.Name == downloader.DOWNLOADING || jobStatus.Name == downloader.COMPLETED)
+
+	time.Sleep(20 * time.Millisecond)
```
sethiay, line 814: so we didn't wait and read again in this test case because we wanted to avoid adding time.sleep. 
the previous check on job.status is sufficient for this test case, so we can remove this.

internal/gcsx/random_reader.go
```diff
@@ -256,14 +255,15 @@ func (rr *randomReader) ReadAt(
...
-	// true in that case.
+	// GCS, cached in file and then served from that file. But the cacheHit is
+	// false in that case.
 	n, cacheHit, err = rr.tryReadingFromFileCache(ctx, p, offset)
 	if err != nil {
 		err = fmt.Errorf("ReadAt: while reading from cache: %v", err)
 		return
 	}
-	if cacheHit {
+	// data was served from cache.
```
sethiay, line 265: @raj-prince Could you please review this check ?

raj-prince: Check looks good to me.

**One more option:**
We can make this second level check, less chances to hit n < len(p).
if cacheHit  then return 
else { 
if n == len(p) || (...) then return 
}

raj-prince, line 265: nit: data -> Data

raj-prince (approved):
Main logic looks good to me, didn't check the test file.

sethiay (approved):

PR #1568, opened 2023-12-28
Fixing 2nd epoch onwards slowness in read-cache fio test

raj-prince:
### Description
Fixing 2nd epoch onwards slowness in read-cache fio test
(a) Increasing the stat-cache capacity to more than 3 millions.
(b) Only clearing the page cache not inodes, just to keep type-cache intact for further epochs. Otherwise, it will try to populate the type-cache again and in the process it will make two stat calls for each in which one of them will be served from the stat-cache and another one will be served GCS since (during list negative entry is not stored in the stat-cache).

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/read_cache/run_read_cache_fio_workload.sh
+++ b/perfmetrics/scripts/read_cache/run_read_cache_fio_workload.sh
@@ -73,7 +73,8 @@ fi
 # Specially for gcsfuse mounted dir: the purpose of this approach is to efficiently
 # populate the gcsfuse metadata cache by utilizing the list call, which internally
 # works like bulk stat call rather than making individual stat calls.
-time ls -R $workload_dir
+# And to reduce the logs moving the command standard-output to /dev/null.
+time ls -R $workload_dir 1> /dev/null
 
 cd $WORKING_DIR/gcsfuse/perfmetrics/scripts/read_cache/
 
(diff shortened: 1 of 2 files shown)


perfmetrics/scripts/read_cache/run_read_cache_fio_workload.sh
```diff
@@ -80,8 +86,16 @@ for i in $(seq $epoch); do
...
+  # To free pagecache.
+  # Intentionally not clearing dentries and inodes: clearing them
+  # will necessitate the repopulation of the type cache in gcsfuse 2nd epoch onwards.
+  # Since we use "ls -R workload_dir" to populate the cache (sort of hack to fill the cache quickly)
+  # efficiently in the first epoch, it does not populate the negative
+  # entry for the stat cache.
+  # So just to stop the execution of  “ls -R workload_dir” command at the start
+  # of every epoch, not clearing the inodes.
+
+  sudo sh -c "/usr/bin/echo 1 > /proc/sys/vm/drop_caches"
```
sethiay, line 98: let's also add sudo sh -c "/usr/bin/echo 3 > /proc/sys/vm/drop_caches" at the start of this script ?

sethiay (approved):

PR #1569, opened 2023-12-29
Moving monitoring bucket bottom in hierarchy

raj-prince:
### Description
Moving the monitor bucket in the bottom of the hierarchy, just before the actual bucket implementation which interacts with GCS.

Why?
Currently, monitoring bucket is on the top of hierarchy, above fast-stat bucket. So, we see all the request/response gcs stackdriver metrics which is served by cache (fast-stat-bucket). Ideally, we should only see the metrics which is served by gcs server. Hence, fixing this by moving the monitoring bucket bottom of hierarchy.

Impact:
We don't get false Stat call in gcs stackdriver metrics. As you can see in the below screenshot containing before and after
metrics graph for the same workload.

image

NA

### Testing details
1. Manual - Verified debug gcs logs and exported stackdriver metrics before and after.
2. Unit tests - Automation.

### DIFF
--- a/internal/gcsx/bucket_manager.go
+++ b/internal/gcsx/bucket_manager.go
@@ -161,6 +161,11 @@ func setUpRateLimiting(
 func (bm *bucketManager) SetUpGcsBucket(name string) (b gcs.Bucket, err error) {
 	b = bm.storageHandle.BucketHandle(name, bm.config.BillingProject)
 
+	// Enable monitoring
+	if bm.config.EnableMonitoring {
+		b = monitor.NewMonitoringBucket(b)
+	}
+
 	b = storage.NewDebugBucket(b)
 	return
 }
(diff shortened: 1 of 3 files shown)


internal/gcsx/bucket_manager.go
```diff
@@ -175,13 +164,17 @@ func (bm *bucketManager) SetUpBucket(
...
-		b, err = bm.SetUpGcsBucket(name)
-		if err != nil {
-			err = fmt.Errorf("OpenBucket: %w", err)
-			return
-		}
+		b = bm.storageHandle.BucketHandle(name, bm.config.BillingProject)
 	}
 
+	// Enable monitoring.
+	if bm.config.EnableMonitoring {
```
sethiay, line 171: any reason for keeping this above "b = storage.NewDebugBucket(b)" ?

raj-prince: Yeah, but not any strong reason.
(a) Keeping the gcs debug logs and stackdriver metrics order same as fuse logs.
(b) Logically, to avoid the logging time in stackdriver metrics (specially latency related metrics).

sethiay (dismissed):

sethiay (approved):

PR #1575, opened 2024-01-04
Update SLO

Tulsishah:
### Description
There was a general agreement that we should change this SLO because we are not a 24X7 supported product, plus we are not a service, so if the customer ran into issues, they can rollback the gcsfuse library as a mitigation and wait for us to respond in business hours.

Changing the SLO to fix this issue.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/ISSUE_TEMPLATE/gcsfuse-bug-report.md
+++ b/.github/ISSUE_TEMPLATE/gcsfuse-bug-report.md
@@ -28,8 +28,4 @@ Steps to reproduce the behavior:
 Add any other context about the problem here.
 
 **SLO:**
-We strive to respond to all bug reports within 24 business hours during our 
-standard business hours of Monday to Friday, 9:00 AM to 5:00 PM [IST Time zone]. 
-We will then work diligently to resolve the issue, providing updates and mitigation
-steps as available. However, the resolution timeframe may vary depending on the 
-complexity of the issue and our development roadmap.
+We strive to respond to all bug reports within 24 business hours.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/gcsfuse-bug-report.md
```diff
@@ -28,4 +28,8 @@ Steps to reproduce the behavior:
 Add any other context about the problem here.
 
 **SLO:**
-24 hrs to respond and 7 days to close the issue.
+We strive to respond to all bug reports within 24 business hours during our 
+standard business hours of Monday to Friday, 9:00 AM to 5:00 PM [IST Time zone]. 
+We will then work diligently to resolve the issue, providing updates and mitigation
+steps as available. However, the resolution timeframe may vary depending on the 
+complexity of the issue and our development roadmap.
```
mustvicky: I think we can omit this part as the business hours can be different if a person is travelling etc.. Usually business hours are well understood. If we get any questions from community, we can then change this and make it more elaborate. For now, my recommendation is to remove this: "during our 
standard business hours of Monday to Friday, 9:00 AM to 5:00 PM [IST Time zone]. 
We will then work diligently to resolve the issue, providing updates and mitigation
steps as available. However, the resolution timeframe may vary depending on the 
complexity of the issue and our development roadmap."

mustvicky (changes requested):

.github/ISSUE_TEMPLATE/gcsfuse-bug-report.md
Tulsishah: Done

mustvicky (approved):
LGTM

sethiay (approved):

PR #1578, opened 2024-01-05
[ReadCache functional tests] Add util methods to parse GSCFuse logs 

ashmeenkaur:
### Description
Added util methods that filters out GCSFuse read operation logs and parse them into a structured map. This will be used for Read Cache functional tests.

NA

### Testing details
1. Manual - Verified manually that parsing is working as intended.
2. Unit tests - Added unit tests for public method ParseLogFile.
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/util/log_parser/json_parser/helpers.go
+++ b/tools/integration_tests/util/log_parser/json_parser/helpers.go
@@ -29,7 +29,7 @@ func parseToInt64(token string) (int64, error) {
 	return res, nil
 }
 
-func readFileLineByLine(reader io.Reader) ([]string, error) {
+func loadLogLines(reader io.Reader) ([]string, error) {
 	content, err := io.ReadAll(reader)
 	if err != nil {
 		return nil, err
@@ -37,6 +37,9 @@ func readFileLineByLine(reader io.Reader) ([]string, error) {
 	return strings.Split(string(content), "\n"), nil
 }
 
+// parseReadFileLog parses a tokenized read file log message and adds details
+// (handle, processId, inodeId) corresponding to the file handle in the
+// structuredLogs map.
 func parseReadFileLog(startTimeStampSec, startTimeStampNanos int64, logs []string,
 	structuredLogs map[int64]*StructuredLogEntry) error {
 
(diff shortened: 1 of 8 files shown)


tools/integration_tests/util/log_parser/json_parser/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+	if logEntry.ObjectName == "" && logEntry.BucketName == "" {
+		bucketAndObjectName := logs[2][10 : len(logs[2])-1] // Remove prefix "FileCache(" and suffix ","
+		// bucketAndObjectName will be stored in format <bucketName>:/<objectName>
+		logEntry.BucketName = strings.Split(bucketAndObjectName, ":")[0]
+		logEntry.ObjectName = strings.Split(bucketAndObjectName, ":")[1][1:] // Remove prefix "/"
+	}
+
+	// Create chunk data entry and append it to the filecache logs.
+	chunkData := ChunkData{
+		StartTimeSeconds: startTimeStampSec,
```
sethiay, line 111: please check/plan if this requires some changes for this extensibility in future -  add error also in this struct in case we want to put checks on errors ?

tools/integration_tests/util/log_parser/json_parser/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+
+func parseToInt64(token string) (int64, error) {
+	res, err := strconv.ParseInt(token, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("could not parse %s to int64: %v", token, err)
+	}
+	return res, nil
+}
+
+func readFileLineByLine(reader io.Reader) ([]string, error) {
```
sethiay, line 32: nit: rename it to loadLogLines, readAllFileLines

sethiay (commented):
I haven't reviewed json_parser_test.go as it is test case of testing helpers.

tools/integration_tests/util/log_parser/json_parser/helpers.go
ashmeenkaur: Added the error string. I will update the parsing logic for error strings in a separate PR.

tools/integration_tests/util/log_parser/json_parser/structure.go
```diff
@@ -0,0 +1,52 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package json_parser
+
+// StructuredLogEntry stores the structured format to be created from logs.
+type StructuredLogEntry struct {
```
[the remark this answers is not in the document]
ashmeenkaur: I don't think the structured log entry will deviate a lot for other operations like write. Do you think it will be a better idea to add operation type in this entry instead?

tools/integration_tests/util/log_parser/json_parser/helpers.go
```diff
@@ -0,0 +1,171 @@
...
+// first Filecache log for the read operation.
+// 2. adds read operation chunk (opId, size, offset) corresponding to the file
+// handle in the structuredLogs map.
+// 3. Stores a reverse mapping of FileCache operation id to file handle and
+// chunk index in a map, to be re-used while mapping file cache response logs to
+// read chunk.
+//
+// Note: It is expected that parseFileCacheLog will always come after ReadFile
+// log. If corresponding ReadFile log is missing, this function throws an error.
+func parseFileCacheLog(startTimeStampSec, startTimeStampNanos int64, logs []string,
```
sethiay, line 89: nit: parseFileCacheRequestLog ?

tools/integration_tests/util/log_parser/json_parser/structure.go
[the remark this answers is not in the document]
sethiay: There are other logical operations like stat apart from read and write. I would prefer to keep "Read" in name and in future when and if we need structure for write and other operations then we can extract the common structure depending upon requirements.

sethiay (commented):
Handle the lint error.

tools/integration_tests/util/log_parser/json_parser/structure.go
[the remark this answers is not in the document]
ashmeenkaur: Makes sense. Refactored to move this into a separate package for reads and also renamed the structs to include Read in the name

tools/integration_tests/util/log_parser/json_parser/helpers.go
ashmeenkaur: I am removing the error string for now as I don't think it will be required and also seeing lint issues if it is unused. I'll add it back if needed. Creating a backlog bug for the same.

sethiay (approved):

PR #1607, opened 2024-01-11
Read cache functional test case 1

ashmeenkaur:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -25,7 +25,7 @@ func readFileAndGetExpectedOutcome(testDirPath, fileName string, t *testing.T) *
 	expected := &Expected{
 		StartTimeStampSeconds: time.Now().Unix(),
 		BucketName:            setup.TestBucket(),
-		ObjectName:            path.Join(testDirName, setup.OnlyDirMounted(), fileName),
+		ObjectName:            path.Join(setup.OnlyDirMounted(), testDirName, fileName),
 	}
 
 	content, err := operations.ReadFileSequentially(path.Join(testDirPath, fileName), MiB)
@@ -68,7 +68,7 @@ func validate(expected *Expected, logEntry *read_logs.StructuredReadLogEntry,
 
 func (s *testStruct) validateFileInCacheDirectory(t *testing.T) {
 	// Validate that the file is present in cache location.
-	expectedPathOfCachedFile := path.Join(cacheLocationPath, cacheSubDirectoryName, setup.TestBucket(), testDirName, testFileName)
+	expectedPathOfCachedFile := path.Join(cacheLocationPath, cacheSubDirectoryName, setup.TestBucket(), setup.OnlyDirMounted(), testDirName, testFileName)
 	fileInfo, err := operations.StatFile(expectedPathOfCachedFile)
 	if err != nil {
 		t.Errorf("Failed to find cached file %s: %v", expectedPathOfCachedFile, err)
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -1,3 +1,17 @@
+// Copyright 2024 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package read_cache
 
 import (
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -1,0 +1,86 @@
+package read_cache
+
+import (
+	"path"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/client"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/log_parser/json_parser/read_logs"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
+)
+
+// Expected is a helper struct that stores list of attributes to be validated from logs.
+type Expected struct {
+	StartTimeStampSeconds int64
+	EndTimeStampSeconds   int64
+	BucketName            string
+	ObjectName            string
+	content               string
+}
+
+func readFileAndGetExpectedOutcome(testDirPath, fileName string, t *testing.T) *Expected {
+	expected := &Expected{
+		StartTimeStampSeconds: time.Now().Unix(),
+		BucketName:            setup.TestBucket(),
+		ObjectName:            path.Join(testDirName, setup.OnlyDirMounted(), fileName),
+	}
+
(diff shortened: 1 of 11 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	expected.EndTimeStampSeconds = time.Now().Unix()
+	expected.content = strings.Trim(string(content), "\x00")
+
+	return expected
+}
+
+func validate(expected *Expected, logEntry *read_logs.StructuredReadLogEntry,
+	isSeq, cacheHit bool, chunkCount int, t *testing.T) {
+	if logEntry.StartTimeSeconds < expected.StartTimeStampSeconds {
```
Tulsishah, line 43: Can it be > ?

ashmeenkaur: expected start time stamp is taken before the read operation. The log entry time stamp should be greater or equal to this expected time stamp.

tools/integration_tests/read_cache/setup_test.go
```diff
@@ -0,0 +1,97 @@
...
+	// Set up test directory.
+	setup.SetUpTestDirForTestBucketFlag()
+
+	// Save mount and root directory variables.
+	mountDir, rootDir = setup.MntDir(), setup.MntDir()
+
+	log.Println("Running static mounting tests...")
+	mountFunc = static_mounting.MountGcsfuseWithStaticMounting
+	successCode := m.Run()
+
```
Tulsishah, line 75: Are we also including persistent mounting tests?

tools/integration_tests/read_cache/setup_test.go
```diff
@@ -0,0 +1,97 @@
+package read_cache
```
Tulsishah, line 1: I think this file should name as read_cache_test.go.

tools/integration_tests/read_cache/setup_test.go
ashmeenkaur: We decided earlier not to include persistent mounting tests

tools/integration_tests/read_cache/read_only_test.go
```diff
@@ -0,0 +1,128 @@
...
+	if setup.MountedDirectory() != "" {
+		t.Log("Skipping setup as tests are running test in GKE environment")
+		return
+	}
+	// Mount GCSFuse.
+	if err := mountFunc(s.flags); err != nil {
+		t.Errorf("Failed to mount GCSFuse: %v", err)
+	}
+	setup.SetMntDir(mountDir)
+	testDirPath = client.SetupTestDirectoryForROMount(s.ctx, s.storageClient, testDirName)
```
Tulsishah, line 45: Without `--ro` mount also testDir and testFile creating by this way?

tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func readFileAndGetExpectedOutcome(testDirPath, fileName string, t *testing.T) *Expected {
+	expected := &Expected{
+		StartTimeStampSeconds: time.Now().Unix(),
+		BucketName:            setup.TestBucket(),
+		ObjectName:            path.Join(testDirName, fileName),
+	}
+
+	content, err := operations.ReadFileSequentially(path.Join(testDirPath, fileName), MiB)
```
charith87, line 31: For Readability - I would add a variable named chunkSizeToRead and then set it as MiB and pass the variable.

tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		t.Errorf("Failed to read file in first iteration: %v", err)
+	}
+	expected.EndTimeStampSeconds = time.Now().Unix()
+	expected.content = strings.Trim(string(content), "\x00")
```
charith87, line 36: "\x00" is this the default EOF marker for a log file always or can it have different ones based on different linux version, configurations.

tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Errorf("Expected Cache Hit: %t, Got from logs: %t", cacheHit, logEntry.Chunks[0].CacheHit)
+	}
+	if isSeq != logEntry.Chunks[0].IsSequential {
+		t.Errorf("Expected Is Sequential: %t, Got from logs: %t", isSeq, logEntry.Chunks[0].IsSequential)
+	}
+}
+
+func (s *testStruct) validateFileInCacheDirectory(t *testing.T) {
+	// Validate that the file is present in cache location.
+	expectedPathOfCachedFile := path.Join(cacheLocationPath, cacheSubDirectoryName, setup.TestBucket(), testDirName, testFileName)
```
charith87, line 71: *nit = Would be good to make this a function separately called getCachedFilePath()

tools/integration_tests/read_cache/read_only_test.go
ashmeenkaur: Yes, for both we are using go storage client. I will rename the method.

tools/integration_tests/util/log_parser/json_parser/read_logs/json_read_log_parser.go
```diff
@@ -46,7 +50,7 @@ func filterAndParseLogLine(logLine string,
 		if err := parseReadFileLog(timestampSeconds, timestampNanos, tokenizedLogs, structuredLogs); err != nil {
 			return fmt.Errorf("parseReadFileLog failed: %v", err)
 		}
-	case strings.Contains(logMessage, "FileCache"):
+	case strings.Contains(logMessage, "FileCache("):
```
charith87, line 53: Feel like this file is going to be the highest maintenance code. As it has some amount of assumptions that log would be in a expected format. If there is a basic sanity test if the logs are in expected format that would help for future. If its already there, then its good.

tools/integration_tests/read_cache/setup_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func createConfigFile() string {
+	cacheLocationPath = path.Join(setup.TestDir(), "cache-dir")
+
+	// Set up config file for file cache.
+	mountConfig := config.MountConfig{
+		FileCacheConfig: config.FileCacheConfig{
+			// Keeping the size as low because the operations are performed on small
+			// files
+			MaxSizeInMB: 10,
```
charith87, line 40: How small are the files that we are testing. I think we would want to test random and sequential both. For Random itself, it gets only if > 8MB etc. So 10MB looks too low IMO

charith87 (changes requested):
Added comments

tools/integration_tests/util/log_parser/json_parser/read_logs/json_read_log_parser.go
ashmeenkaur: Agreed. I will change this to regex match instead of string contains in a separate PR. Adding a TODO and internal bug for the same.

tools/integration_tests/read_cache/setup_test.go
ashmeenkaur: This config file is only for sequential reads where we are planning to test only for small files. We also have plans to test with bigger files for range reads (both sequential and random) which will have a separate config file. I'll make the required parameters configurable in the function.

charith87: Sounds good.

tools/integration_tests/read_cache/helpers_test.go
ashmeenkaur: Yes, \x00 is the default EOF marker on all linux versions. For now, I have made this a constant in util package. I will move it to util if required later as suggested.

tools/integration_tests/read_cache/setup_test.go
ashmeenkaur: Let's call it setup_test.go only as this has the setup for the tests. We can rename later if required.

Tulsishah (approved):

charith87 (approved):

PR #1614, opened 2024-01-16
[ReadCacheFunctionalTest] Test 2: Object modified by another client after read.

ashmeenkaur:
#### Scenario:
R1: Full object read 
Another client has written to make it the file to change object generation
R2: Full object read 
R3: Full object read (after stat cache expiry) 

### Expected Behavior
R1 - cache miss
R2 - cache hit (but stale data)
R3 - cache miss (new data)

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -16,6 +16,9 @@ package read_cache
 
 import (
 	"context"
+	"fmt"
+	"log"
+	"os"
 	"path"
 	"strings"
 	"testing"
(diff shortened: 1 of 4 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -84,21 +89,68 @@ func getCachedFilePath() string {
 	return path.Join(cacheLocationPath, cacheSubDirectoryName, setup.TestBucket(), testDirName, testFileName)
 }
 
-func (s *testStruct) validateFileInCacheDirectory(t *testing.T) {
+func validateFileSizeInCacheDirectory(filesize int64, t *testing.T) (expectedPathOfCachedFile string) {
```
Tulsishah, line 92: Return `expectedPathOfCachedFile ` is required for this function?

tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -84,21 +89,68 @@ func getCachedFilePath() string {
...
-		t.Errorf("Incorrect cached file size. Expected %d, Got: %d", fileSize, (*fileInfo).Size())
+	if (*fileInfo).Size() != filesize {
+		t.Errorf("Incorrect cached file size. Expected %d, Got: %d", filesize, (*fileInfo).Size())
 	}
+	return
+}
+
+func validateFileInCacheDirectory(filesize int64, ctx context.Context, storageClient *storage.Client, t *testing.T) {
+	expectedPathOfCachedFile := validateFileSizeInCacheDirectory(filesize, t)
 	// Validate content of file in cache directory matches GCS.
```
Tulsishah, line 108: Shouldn't we do this?
```
validateFileSizeInCacheDirectory(filesize, t)
expectedPathOfCachedFile := getCachedFilePath()
```

According to me returning filePath in size validation function does not seems correct?

tools/integration_tests/read_cache/small_cache_ttl_test.go
```diff
@@ -0,0 +1,121 @@
...
+	// Append to the file.
+	err := client.WriteToObject(s.ctx, s.storageClient, objectName, smallContent, storage.Conditions{})
+	if err != nil {
+		t.Errorf("Could not append to file: %v", err)
+	}
+
+	// Read same file again immediately.
+	// Stale data would be served from cache in this case.
+	expectedOutcome2 := readFileAndGetExpectedOutcome(testDirPath, testFileName, t)
+	validateFileSizeInCacheDirectory(fileSize, t)
```
Tulsishah, line 81: If data is appended in the file, then how file size will be equal?

ashmeenkaur: in this case, stale data from cache is being served so that is why file size is same

Tulsishah (approved):

PR #1616, opened 2024-01-16
[ReadCacheFunctionalTest] Test 3: Object modified by another client after read with stat cache ttl 0

ashmeenkaur:
#### Scenario:
With stat cache ttl = 0
R1: Full object read 
Another client has written to make it the file to change object generation
R2: Full object read 
R3: Full object read

### Expected Behavior
R1 - cache miss
R2 - cache miss (serves updated data as cache ttl is 0)
R3 - cache hit (serves updated data)

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - In progress

### DIFF
--- a/tools/integration_tests/read_cache/disabled_cache_ttl_test.go
+++ b/tools/integration_tests/read_cache/disabled_cache_ttl_test.go
@@ -16,7 +16,6 @@ package read_cache
 
 import (
 	"context"
-	"path"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


Tulsishah (approved):

Tulsishah (commented):
As agreed offline, we'll replace fixed test file names (testFile1, testFile2, etc.) with randomly generated numbers(testFile-xxx).

PR #1617, opened 2024-01-16
adds bug fix for cache location in case temp dir is defined by user

ankitaluthra1:
b/320416020

### Testing details
1. Manual - NA
2. Unit tests - Yes
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -234,7 +234,6 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
 		} else {
 			cacheLocation = cfg.TempDir
 		}
-
 	}
 	cacheLocation, err := filepath.Abs(cacheLocation)
 	if err != nil {
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -231,7 +231,10 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
 	if cacheLocation == "" {
 		if cfg.TempDir == "" {
 			cacheLocation = os.TempDir()
+		} else {
+			cacheLocation = cfg.TempDir
 		}
+
 	}
 	cacheLocation, err := filepath.Abs(cacheLocation)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/fs/fs.go
```diff
@@ -231,7 +231,10 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
 	if cacheLocation == "" {
 		if cfg.TempDir == "" {
 			cacheLocation = os.TempDir()
+		} else {
+			cacheLocation = cfg.TempDir
 		}
+
```
sethiay, line 237: nit: remove this line ?

internal/fs/read_cache_test.go
```diff
@@ -731,8 +732,13 @@ type FileCacheWithDefaultCacheLocation struct {
 	fsTest
 }
 
+type FileCacheWithUserDefinedTempAsCacheLocation struct {
+	fsTest
+}
+
 func init() {
 	RegisterTestSuite(&FileCacheWithDefaultCacheLocation{})
+	RegisterTestSuite(&FileCacheWithUserDefinedTempAsCacheLocation{})
```
ankitaluthra1, line 741: @sethiay let me know if this needs to be registered in different init.. if its okay, should I add all register statements into one init method (better readable if they are all in one place)?

sethiay (approved):

internal/fs/read_cache_test.go
sethiay: Yes, adding at one place is better.

sethiay (commented):
reverting approval for nit comment.

internal/fs/fs.go
ankitaluthra1: done

internal/fs/read_cache_test.go
ankitaluthra1: done

sethiay: Thank you, nit: can we move init at the top - after constants definition and before first struct ?

ankitaluthra1: done

sethiay (approved):

PR #1624, opened 2024-01-18
[ReadCacheFunctionalTest] Test 19: Test that cache works as intended on remounting GCSFuse and data is cleared between remount.

ashmeenkaur:
#### Scenario:

R1: Read file sequentially
R2: read again
Remount GCSFuse
R3: Read same file again sequentially
R4: Read same file again sequentially

#### Expected Behavior
R1: Cache miss
R2: Cache Hit
Validate cache directory empty
R3: Cache miss
R4: cache hit

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Ran via presubmits

### DIFF
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -88,6 +88,10 @@ func getCachedFilePath(fileName string) string {
 	return path.Join(cacheLocationPath, cacheSubDirectoryName, setup.TestBucket(), testDirName, fileName)
 }
 
+func getCacheDirPath() string {
+	return path.Join(cacheLocationPath, cacheSubDirectoryName)
+}
+
 func validateFileSizeInCacheDirectory(fileName string, filesize int64, t *testing.T) {
 	// Validate that the file is present in cache location.
 	expectedPathOfCachedFile := getCachedFilePath(fileName)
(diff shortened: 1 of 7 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -136,6 +136,22 @@ func unmountGCSFuseAndDeleteLogFile() {
 	}
 }
 
+func remountGCSFuseAndValidateCacheDeleted(flags []string, t *testing.T) {
+	setup.SetMntDir(rootDir)
+	unmountGCSFuseAndDeleteLogFile()
+	mountGCSFuse(flags)
+	setup.SetMntDir(mountDir)
+
+	cacheSize, err := operations.DirSize(cacheLocationPath)
```
Tulsishah, line 145: Shouldn't we use this function `validateFileIsNotCached(testFileName,t)` ?

tools/integration_tests/read_cache/remount_test.go
```diff
@@ -0,0 +1,105 @@
...
+	setup.SetMntDir(rootDir)
+	unmountGCSFuseAndDeleteLogFile()
+}
+
+////////////////////////////////////////////////////////////////////////
+// Test scenarios
+////////////////////////////////////////////////////////////////////////
+
+func (s *remountTest) Test(t *testing.T) {
+	testFileName := testFileName + "1"
```
Tulsishah, line 55: testFileName + generaterandomstring()

tools/integration_tests/read_cache/helpers_test.go
ashmeenkaur: Checking that the directory size is 0 covers all the scenarios I feel. Also we will validate that file is not cached as part of the read operation after remount.

Tulsishah (approved):

PR #1629, opened 2024-01-19
[ReadCacheFunctionalTest] Test 11 and 12: Single file random read of size more than cache capacity

ashmeenkaur:
#### Scenario 1:
With flag cacheFileForRangeRead true
R1: Single file random read of size more than cache capacity
R2: read again

#### Expected Behavior
R1: Cache miss
R2: Cache Miss (because the file won't be cached)

#### Scenario 2:
With flag cacheFileForRangeRead false
R1: Single file random read of size more than cache capacity
R2: read again

#### Expected Behavior
R1: Cache miss
R2: Cache Miss (because the file won't be cached)

This PR also includes refactoring to make flag setup more efficient.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Ran via presubmits

### DIFF
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -203,7 +203,7 @@ func readFileAndValidateCacheWithGCS(ctx context.Context, storageClient *storage
 	return expectedOutcome
 }
 
-func ReadFileAndValidateFileIsNotCached(ctx context.Context, storageClient *storage.Client,
+func readFileAndValidateFileIsNotCached(ctx context.Context, storageClient *storage.Client,
 	filename string, isSeq bool, t *testing.T) (expectedOutcome *Expected) {
 	// Read file via gcsfuse mount.
 	expectedOutcome = readFileAndGetExpectedOutcome(testDirPath, filename, isSeq, t)
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -39,16 +39,25 @@ type Expected struct {
 	content               string
 }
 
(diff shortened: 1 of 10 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -194,15 +203,20 @@ func readFileAndValidateCacheWithGCS(ctx context.Context, storageClient *storage
 	return expectedOutcome
 }
 
-func ReadFileAndValidateFileIsNotCached(ctx context.Context, storageClient *storage.Client, filename string, t *testing.T) (expectedOutcome *Expected) {
+func ReadFileAndValidateFileIsNotCached(ctx context.Context, storageClient *storage.Client,
```
Tulsishah, line 206: Other methods are private, can we also make it private ?

tools/integration_tests/util/client/gcs_helper.go
```diff
@@ -78,6 +78,21 @@ func ValidateObjectContentsFromGCS(ctx context.Context, storageClient *storage.C
 	}
 }
 
+func ValidateObjectChunkFromGCS(ctx context.Context, storageClient *storage.Client,
+	testDirName string, fileName string, offset, size int64, expectedContent string,
+	t *testing.T) {
+	gotContent, err := ReadChunkFromGCS(ctx, storageClient,
```
Tulsishah, line 84: This should be expectedContent and in above should be got content?

ashmeenkaur: Discussed offline, the "expected" and "got" depends on the context of test and what we are considering as our source of truth. We are not changing the variable names here to keep it consistent with other similar methods in gcs_helper.go

Tulsishah (approved):

PR #1634, opened 2024-01-22
Add type-cache composite tests

gargnitingoogle:
### Description
This adds some composite tests for
type-cache along the same lines as the other
tests in internal/fs/*_test.go .
This is not an exhaustive set, but a first
batch, and more might be added later.

This adds on top of https://github.com/GoogleCloudPlatform/gcsfuse/pull/1687 .

NA

### Testing details
1. Manual - Ran and passed the tests locally.
2. Unit tests - Ran and passed locally. No new tests.
3. Integration tests - NA

### DIFF
--- a/internal/cache/metadata/type_cache.go
+++ b/internal/cache/metadata/type_cache.go
@@ -94,6 +94,14 @@ func (ce cacheEntry) Size() (size uint64) {
 	return
 }
 
+// Return the size (rss) of a type-cache entry
+// for a given key-string.
+func SizeOfTypeCacheEntry(key string) uint64 {
+	ce := cacheEntry{}
+	ce.key = key
+	return ce.Size()
+}
+
 // A cache that maps from a name to information about the type of the object
 // with that name. Each name N is in one of the following states:
 //
--- a/internal/cache/metadata/type_cache.go
+++ b/internal/cache/metadata/type_cache.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/cache/lru"
-	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 	"github.com/googlecloudplatform/gcsfuse/internal/util"
 )
 
@@ -141,29 +140,22 @@ func NewTypeCache(maxSizeMB int, ttl time.Duration) TypeCache {
 	return &typeCache{}
 }
 
-func logf(now time.Time, msg string, v ...any) {
-	logger.Infof("TypeCache @{%v}: "+msg, append([]any{now.Format(time.DateTime)}, v...)...)
-}
-
 func (tc *typeCache) Insert(now time.Time, name string, it Type) {
 	if tc.entries != nil { // only if caching is enabled
-		localExpiryVar := now.Add(tc.ttl)
 		_, err := tc.entries.Insert(name, cacheEntry{
-			expiry:    localExpiryVar,
+			expiry:    now.Add(tc.ttl),
 			inodeType: it,
 			key:       name,
 		})
 		if err != nil {
 			panic(fmt.Errorf("failed to insert entry in typeCache: %v", err))
 		}
-		logf(now, "Inserted \"%s\" as %v with ttl=%q", name, it, localExpiryVar.Format(time.DateTime))
 	}
 }
 
 func (tc *typeCache) Erase(name string) {
 	if tc.entries != nil { // only if caching is enabled
 		tc.entries.Erase(name)
-		logf(time.Time{}, "Erased \"%s\"", name)
 	}
 }
 
@@ -174,17 +166,14 @@ func (tc *typeCache) Get(now time.Time, name string) Type {
 
 	val := tc.entries.LookUp(name)
 	if val == nil {
-		logf(now, "Queried \"%s\". Got: %v", name, "UnknownType")
 		return UnknownType
 	}
 
 	entry := val.(cacheEntry)
 	// Has the entry expired?
 	if entry.expiry.Before(now) {
-		logf(now, "Queried \"%s\". Removing entry because of TTL expiry", name)
 		tc.entries.Erase(name)
 		return UnknownType
 	}
-	logf(now, "Queried \"%s\". Got: %v", name, entry.inodeType)
 	return entry.inodeType
 }
--- a/internal/cache/metadata/type_cache.go
+++ b/internal/cache/metadata/type_cache.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/cache/lru"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 	"github.com/googlecloudplatform/gcsfuse/internal/util"
 )
 
@@ -140,22 +141,29 @@ func NewTypeCache(maxSizeMB int, ttl time.Duration) TypeCache {
 	return &typeCache{}
 }
 
+func logf(now time.Time, msg string, v ...any) {
+	logger.Infof("TypeCache @{%v}: "+msg, append([]any{now.Format(time.DateTime)}, v...)...)
+}
+
 func (tc *typeCache) Insert(now time.Time, name string, it Type) {
 	if tc.entries != nil { // only if caching is enabled
+		localExpiryVar := now.Add(tc.ttl)
 		_, err := tc.entries.Insert(name, cacheEntry{
-			expiry:    now.Add(tc.ttl),
+			expiry:    localExpiryVar,
 			inodeType: it,
 			key:       name,
 		})
 		if err != nil {
 			panic(fmt.Errorf("failed to insert entry in typeCache: %v", err))
 		}
+		logf(now, "Inserted \"%s\" as %v with ttl=%q", name, it, localExpiryVar.Format(time.DateTime))
 	}
 }
 
 func (tc *typeCache) Erase(name string) {
 	if tc.entries != nil { // only if caching is enabled
 		tc.entries.Erase(name)
+		logf(time.Time{}, "Erased \"%s\"", name)
 	}
 }
 
@@ -166,14 +174,17 @@ func (tc *typeCache) Get(now time.Time, name string) Type {
 
 	val := tc.entries.LookUp(name)
 	if val == nil {
+		logf(now, "Queried \"%s\". Got: %v", name, "UnknownType")
 		return UnknownType
 	}
 
 	entry := val.(cacheEntry)
 	// Has the entry expired?
 	if entry.expiry.Before(now) {
+		logf(now, "Queried \"%s\". Removing entry because of TTL expiry", name)
 		tc.entries.Erase(name)
 		return UnknownType
 	}
+	logf(now, "Queried \"%s\". Got: %v", name, entry.inodeType)
 	return entry.inodeType
 }
--- a/internal/fs/fs_test.go
+++ b/internal/fs/fs_test.go
@@ -34,7 +34,6 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/internal/locker"
 	"github.com/googlecloudplatform/gcsfuse/internal/logger"
-	"github.com/googlecloudplatform/gcsfuse/internal/mount"
 	"github.com/googlecloudplatform/gcsfuse/internal/perms"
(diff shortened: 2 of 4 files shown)


internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,381 @@
...
+////////////////////////////////////////////////////////////////////////
+// Common
+////////////////////////////////////////////////////////////////////////
+
+type typeCacheTestCommon struct {
+	fsTest
+}
+
+var (
+	// The following should be configued for different tests
```
gargnitingoogle, line 43: Fix typo

internal/fs/type_cache_test.go
```diff
@@ -210,31 +210,38 @@ func (t *TypeCacheTestWithMaxEntries1) TestSizeBasedEviction() {
...
-	ExpectEq(nil, err)
-	AssertNe(nil, fi)
-	ExpectFalse(fi.IsDir())
+	// type-cache-max-size-mb = 1MiB = 1048576,
+	// and each entry-size = 80 Bytes.
+	// let's add another 13107 entries to evict
+	for i := 0; i < 13107; i++ {
+		name := fmt.Sprintf("object%d", i)
+
+		// Create second file object in GCS.
```
gargnitingoogle, line 226: Update obsolete comment

internal/fs/type_cache_test.go
```diff
@@ -94,7 +95,7 @@ type TypeCacheTestWithZeroTTL struct {
 
 func (t *TypeCacheTestWithZeroTTL) SetUpTestSuite() {
 	ttlInSeconds = 0
-	typeCacheMaxEntries = 10
+	typeCacheMaxMb = 10
```
gargnitingoogle, line 98: lower this to 1.

internal/fs/type_cache_test.go
```diff
@@ -105,13 +106,13 @@ type TypeCacheTestWithInfiniteTTL struct {
 
 func (t *TypeCacheTestWithInfiniteTTL) SetUpTestSuite() {
 	ttlInSeconds = -1
-	typeCacheMaxEntries = 10
+	typeCacheMaxMb = 10
```
gargnitingoogle, line 109: lower this to 1.

gargnitingoogle:
@raj-prince please review this as discussed in the meeting.

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,390 @@
...
+	fsTest
+}
+
+var (
+	// The following should be configured for different tests
+	// differently inside SetUpTestSuite as these need to
+	// set for mount itself.
+
+	// ttlInSeconds is equivalent of metadata-cache:ttl-secs in config-file.
+	ttlInSeconds int64
```
raj-prince, line 49: Instead of keeping these values global, can we keep inside typeCacheTestCommon ? Every time specific test's SetpTestSuite method will override these values.

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,390 @@
...
+		ctx,
+		bucket,
+		name1,
+		[]byte(contents))
+
+	ExpectEq(nil, err)
+	ExpectNe(nil, fileObject)
+
+	// Stat-call with file object. It should
+	// be stattable, bypassing type-cache, as a file.
```
raj-prince, line 141: You mean stattable-> statable?

internal/fs/type_cache_test.go
gargnitingoogle: I tried keeping them in the typeCacheTestCommon class at first. But I'm not sure why it didn't work (values kept changing on their own).  It's the same reason there are so many global vars here too: https://github.com/GoogleCloudPlatform/gcsfuse/blob/342c39b1863fedeb3bd947aaad2d0188dd0737fa/internal/fs/fs_test.go#L97

internal/fs/fs_test.go
```diff
@@ -140,6 +141,11 @@ func (t *fsTest) SetUpTestSuite() {
 	t.serverCfg.SequentialReadSizeMb = SequentialReadSizeMb
 	if t.serverCfg.MountConfig == nil {
 		t.serverCfg.MountConfig = config.NewMountConfig()
+	} else {
+		mountConfig := t.serverCfg.MountConfig
```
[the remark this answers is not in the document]
gargnitingoogle: These could have been done by making changes only in type_cache_test.go, but putting these changes here instead has the advantage that with this, any new/existing composite tests can simply set their own mountConfig now (equivalent of config-file if they need) instead of replicating this code in their test files.

internal/fs/type_cache_test.go
gargnitingoogle: There is no dictionary word for this. All we mean here is that we can call `stat` system call on the object. I didn't like use the of the word `statable` as it already has a different [meaning](https://www.google.com/search?q=meaning+of+statable) .

gargnitingoogle: changed.

internal/fs/fs_test.go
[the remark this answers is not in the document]
raj-prince: I just wanted to get ride of changes in the fs_test.SetUpTestSuite(). 

But anyway changing the DirTypeCacheTTL and InodeAttributeCacheTTL based on MountingConfig check doesn't look good.

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,390 @@
...
+		bucket,
+		name1,
+		[]byte(contents))
+
+	ExpectEq(nil, err)
+	ExpectNe(nil, fileObject)
+
+	// Stat-call with file object. It should
+	// pass stat call, bypassing type-cache, as a file.
+	fi, err = os.Stat(path.Join(mntDir, name1))
```
raj-prince, line 142: Does it call, LookUpInode?

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+	// type-cache-max-size-mb = 1MiB = 1048576,
+	// and each entry-size = 92 (80 fixed + 12 key-length) Bytes.
+	// let's add another 11397(1048576/92) entries to evict
+	// the first file object ("foo") entry.
+	for i := 0; i < 11397; i++ {
+		name := fmt.Sprintf("object%06d", i)
+
+		// Create a new file object in GCS.
+		fileObject, err = storageutil.CreateObject(
```
raj-prince, line 228: As discussed offline, it takes around 15 seconds. Can we try with `storageutil.CreateObjects`?

internal/fs/fs_test.go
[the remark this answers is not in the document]
raj-prince: Please make the else code un-conditional. It should always be executed.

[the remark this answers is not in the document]
gargnitingoogle: Will do.

internal/fs/type_cache_test.go
gargnitingoogle: I tried doing this. But realized that `storageutil.CreateObjects` doesn't return the created objects, without which I can't really use it here. I checked, CreateObjects is used in another 8-10 test files, so if I change the signature of storageutil.CreateObjects, then I'll have to make changes in all those files. Should I make the change in all these 9-10 files ?

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,421 @@
...
+	. "github.com/jacobsa/ogletest"
+)
+
+// The following is the control of an os.Stat(name) call in case of GCSFuse,
+// for understanding how the tests work. The stat call
+// 1. comes as fs.LookUpInode() call to gcsfuse, which
+// 2. queries type-cache of the parent directory, without the suffix '/'.
+// 2.1 If entry is found in type-cache, then that is returned as type.
+// 2.2 If entry is not found in type-cache, then its type is queried from GCS, and the returned type is stored in type-cache and is also returned as cache.
+// 3. If the input to os.Stat() had a suffix '/' and the return type is not ExplicitDir or ImplicitDir, then an error containing 'not a directory' is returned.
```
raj-prince, line 44: Can we test this behavior as part of our test? Since, this is the base of our test, we should add a test, where if we add file and directory with the same name, the second stat (either file or directory should throw error).

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,421 @@
...
+	// in this case.
+	fi, err = os.Stat(path.Join(mntDir, name1) + "/")
+
+	ExpectEq(nil, err)
+	AssertNe(nil, fi)
+	ExpectTrue(fi.IsDir())
+}
+
+// //////////////////////////////////////////////////////////////////////
+// Tests for TypeCacheTestWithMaxEntries1
```
raj-prince, line 190: nit: Tests for TypeCacheTestWithMaxSize1MB ?

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,421 @@
...
+	RegisterTestSuite(&TypeCacheTestWithZeroCapacity{})
+	RegisterTestSuite(&TypeCacheTestWithZeroTTL{})
+	RegisterTestSuite(&TypeCacheTestWithInfiniteTTL{})
+}
+
+// //////////////////////////////////////////////////////////////////////
+// helpers
+// //////////////////////////////////////////////////////////////////////
+func (t *typeCacheTestCommon) testNoInsertion() {
+	const name1 = "foo"
```
raj-prince, line 145: I am seeing, these logic repeated:
(a) creation of "foo" object.
(b) Stat call to validate.
(c) Creation of "foo/" object.
(d) Stat call to validate.

Can we create utility method for the above, so that we can remove the duplicacy at every places?

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,421 @@
...
+	// type-cache-max-size-mb = 1MiB.
+	// let's add another 1MiB/perEntrySize entries to evict
+	// the first file object ("foo") entry.
+	numObjectsToBeInserted := int(util.MiBsToBytes(1)) / perEntrySize
+
+	for i := 0; i < numObjectsToBeInserted; i++ {
+		name := nameOfIthObject(i)
+
+		// Create a new file object in GCS.
+		fileObject, err = storageutil.CreateObject(
```
raj-prince, line 259: As discussed offline, if we can parallelize the creation.

raj-prince (changes requested):
Logically, looks good. Few minor comment or refactoring.

internal/fs/type_cache_test.go
gargnitingoogle: done

internal/fs/type_cache_test.go
gargnitingoogle: done.

internal/fs/type_cache_test.go
gargnitingoogle: Added 2 tests for this.

internal/fs/type_cache_test.go
gargnitingoogle: done.

gargnitingoogle:
@raj-prince I addressed all comments.

gargnitingoogle:
@raj-prince no comments open. PTAL.

internal/cache/metadata/type_cache.go
```diff
@@ -94,6 +94,14 @@ func (ce cacheEntry) Size() (size uint64) {
 	return
 }
 
+// Return the size (rss) of a type-cache entry
+// for a given key-string.
+func SizeOfTypeCacheEntry(key string) uint64 {
```
raj-prince, line 99: rss?
shouldn't it be heap size?

gargnitingoogle: No, it's rss now. Look at the code of entry.Size() function.

raj-prince (approved):

internal/fs/type_cache_test.go
```diff
@@ -0,0 +1,406 @@
...
+	// Create a file object in GCS with same name as the directory object.
+	t.createObjectOnGCS(foo)
+
+	// Stat-call with the file object should
+	// pass but it should still report it as a directory
+	// because of its old type-cache entry which says type is directory.
+	t.statAndConfirmIsDir(path.Join(mntDir, foo), true)
+}
+
+func (t *TypeCacheTestWithMaxSize1MB) TestSizeBasedEviction() {
```
raj-prince, line 253: Now, how much time does it take?

gargnitingoogle: The same we discussed.. about 720-730 ms.

PR #1639, opened 2024-01-24
Cache dir perm check

ankitaluthra1:
[buganizer](https://b.corp.google.com/issues/315991900)

### Testing details
1. Manual - WIP
2. Unit tests - Yes
3. Integration tests - NA

### DIFF
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -114,11 +114,15 @@ func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
 	}
 
 	f, err := fsutil.AnonymousFile(dirPath)
-	f.Close()
-
 	if err != nil {
 		return fmt.Errorf(
 			"error creating file at directory (%s), error : (%q)", dirPath, err.Error())
+	}
+
+	tempFileErr := f.Close()
+	if tempFileErr != nil {
+		return fmt.Errorf(
+			"error closing annonymous temp file, error : (%q)", tempFileErr.Error())
 	}
 
 	return nil
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 	"github.com/jacobsa/fuse/fsutil"
-	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
@@ -103,11 +102,11 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
 
-func CheckPermissionsFor(dirPath string, filemode fs.FileMode) bool {
+func CheckCacheDirectoryPermissions(dirPath string) bool {
 	_, statErr := os.Stat(dirPath)
 
 	if statErr != nil {
-		err := os.MkdirAll(dirPath, filemode)
+		err := os.MkdirAll(dirPath, FileDirPerm)
 		if err != nil {
 			logger.Errorf("error in creating directory structure %s: %v", dirPath, err)
 			return false
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -96,12 +96,14 @@ func GetDownloadPath(cacheLocation string, objectPath string) string {
 // for next call onwards.
 func IsCacheHandleInvalid(readErr error) bool {
 	return strings.Contains(readErr.Error(), InvalidFileHandleErrMsg) ||
-		strings.Contains(readErr.Error(), InvalidFileDownloadJobErrMsg) ||
-		strings.Contains(readErr.Error(), InvalidFileInfoCacheErrMsg) ||
-		strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
-		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
+			strings.Contains(readErr.Error(), InvalidFileDownloadJobErrMsg) ||
+			strings.Contains(readErr.Error(), InvalidFileInfoCacheErrMsg) ||
+			strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
+			strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
 
+// Creates directory at given path with 766 permissions in case not already present,
+// returns error in case unable to create directory or directory has insufficient permissions.
 func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
 	_, statErr := os.Stat(dirPath)
 
@@ -116,13 +118,13 @@ func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
 	f, err := fsutil.AnonymousFile(dirPath)
 	if err != nil {
 		return fmt.Errorf(
-			"error creating file at directory (%s), error : (%q)", dirPath, err.Error())
+			"error creating file at directory (%s), error : (%v)", dirPath, err.Error())
 	}
 
 	tempFileErr := f.Close()
 	if tempFileErr != nil {
 		return fmt.Errorf(
-			"error closing annonymous temp file, error : (%q)", tempFileErr.Error())
+			"error closing annonymous temp file, error : (%v)", tempFileErr.Error())
 	}
 
 	return nil
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -105,7 +105,7 @@ func IsCacheHandleInvalid(readErr error) bool {
 func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
 	_, statErr := os.Stat(dirPath)
 
-	if statErr != nil {
+	if os.IsNotExist(statErr) {
 		err := os.MkdirAll(dirPath, FileDirPerm)
 		if err != nil {
 			return fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -16,6 +16,9 @@ package util
 
 import (
 	"fmt"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
+	"github.com/jacobsa/fuse/fsutil"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
@@ -98,4 +101,28 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), InvalidFileInfoCacheErrMsg) ||
 		strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
+}
+
+func CheckPermissionsFor(dirPath string, filemode fs.FileMode) bool {
+	_, statErr := os.Stat(dirPath)
+
+	if statErr != nil {
+		err := os.MkdirAll(dirPath, filemode)
+		if err != nil {
+			logger.Errorf("error in creating directory structure %s: %v", dirPath, err)
+			return false
+		}
+		return true
+	}
+
+	f, err := fsutil.AnonymousFile(dirPath)
+	f.Close()
+
+	if err != nil {
+		logger.Errorf(
+			"error creating file at directory (%s), error : (%q)", dirPath, err.Error())
+		return false
+	}
+
+	return true
 }
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -16,11 +16,12 @@ package util
 
 import (
 	"fmt"
-	"github.com/jacobsa/fuse/fsutil"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
+
+	"github.com/jacobsa/fuse/fsutil"
 
 	"github.com/googlecloudplatform/gcsfuse/internal/cache/data"
 )
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -102,7 +102,7 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
 
-func CheckCacheDirectoryPermissions(dirPath string) (bool, error) {
+func IsCacheDirectoryWriteable(dirPath string) (bool, error) {
 	_, statErr := os.Stat(dirPath)
 
 	if statErr != nil {
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -102,24 +102,24 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
 
-func IsCacheDirectoryWriteable(dirPath string) (bool, error) {
+func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
 	_, statErr := os.Stat(dirPath)
 
 	if statErr != nil {
 		err := os.MkdirAll(dirPath, FileDirPerm)
 		if err != nil {
-			return false, fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
+			return fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
 		}
-		return true, nil
+		return nil
 	}
 
 	f, err := fsutil.AnonymousFile(dirPath)
 	f.Close()
 
 	if err != nil {
-		return false, fmt.Errorf(
+		return fmt.Errorf(
 			"error creating file at directory (%s), error : (%q)", dirPath, err.Error())
 	}
 
-	return true, nil
+	return nil
 }
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -96,10 +96,10 @@ func GetDownloadPath(cacheLocation string, objectPath string) string {
 // for next call onwards.
 func IsCacheHandleInvalid(readErr error) bool {
 	return strings.Contains(readErr.Error(), InvalidFileHandleErrMsg) ||
-			strings.Contains(readErr.Error(), InvalidFileDownloadJobErrMsg) ||
-			strings.Contains(readErr.Error(), InvalidFileInfoCacheErrMsg) ||
-			strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
-			strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
+		strings.Contains(readErr.Error(), InvalidFileDownloadJobErrMsg) ||
+		strings.Contains(readErr.Error(), InvalidFileInfoCacheErrMsg) ||
+		strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
+		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
 
 // Creates directory at given path with 766 permissions in case not already present,
@@ -112,7 +112,6 @@ func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
 		if err != nil {
 			return fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
 		}
-		return nil
 	}
 
 	f, err := fsutil.AnonymousFile(dirPath)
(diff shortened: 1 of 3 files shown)


internal/fs/fs.go
```diff
@@ -246,10 +246,16 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
 
 	// When user passes allow_other flag, then other users should be able to
 	// read from cache
+	isPermissible, err := util.CheckCacheDirectoryPermissions(cacheLocation)
```
ashmeenkaur, line 249: This code change should be before allow_other comment?

internal/cache/util/util.go
```diff
@@ -99,3 +101,25 @@ func IsCacheHandleInvalid(readErr error) bool {
...
+
+	if statErr != nil {
+		err := os.MkdirAll(dirPath, FileDirPerm)
+		if err != nil {
+			return false, fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
+		}
+		return true, nil
+	}
+
+	f, err := fsutil.AnonymousFile(dirPath)
```
ashmeenkaur, line 116: fsutil.AnonymousFile has the following warning:
```
// Warning: this is not production-quality code, and should only be used for
// testing purposes. In particular, there is a race between creating and
// unlinking by name.
```

Why didn't we check for directory permissions instead?

internal/cache/util/util.go
```diff
@@ -99,3 +101,25 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
+
+func CheckCacheDirectoryPermissions(dirPath string) (bool, error) {
+	_, statErr := os.Stat(dirPath)
+
+	if statErr != nil {
```
ashmeenkaur, line 108: Should we check that stat error is particularly because the directory does not exists?

internal/cache/util/util.go
```diff
@@ -99,3 +101,25 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
+
+func CheckCacheDirectoryPermissions(dirPath string) (bool, error) {
```
ashmeenkaur, line 105: (feel free to ignore if you don't agree)
Nit: Since this method is returning bool, I think it is better to name it IsCacheDirectoryWriteable or something similar

internal/cache/util/util.go
```diff
@@ -99,3 +101,25 @@ func IsCacheHandleInvalid(readErr error) bool {
...
+	_, statErr := os.Stat(dirPath)
+
+	if statErr != nil {
+		err := os.MkdirAll(dirPath, FileDirPerm)
```
ashmeenkaur, line 109: We are still creating cache directory in [cache_handler](https://github.com/GoogleCloudPlatform/gcsfuse/blob/aeb8286b98c70cb02c323a0198ed454a85fc0ab4/internal/cache/file/cache_handler.go#L69). Just wanted to confirm that this is as per design? :)

internal/cache/util/util.go
sethiay, line 105: since the function is trying to create cache directory as well, so I would prefer naming it like TryCreatingCacheDirectory and only return err. err should be nil if cache directory is created successfully and non-nil if not.

internal/cache/util/util.go
sethiay: Didn't know about the warning associated with it, however the warning is not applicable to our use case here and fsutil.AnonymousFile is also used in write flow of GCSFuse https://github.com/GoogleCloudPlatform/gcsfuse/blob/2d635761ba0baa584d56be4332835db213b65404/internal/fs/inode/file.go#L258, so we should be fine to use it here.

@ankitaluthra1 can answer to the question put in the comment.

internal/cache/util/util.go
```diff
@@ -99,3 +101,25 @@ func IsCacheHandleInvalid(readErr error) bool {
...
+		if err != nil {
+			return false, fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
+		}
+		return true, nil
+	}
+
+	f, err := fsutil.AnonymousFile(dirPath)
+	f.Close()
```
sethiay, line 117: should we check the error returned from f.Close() as well ?

internal/cache/util/util_test.go
```diff
@@ -233,3 +235,50 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
...
+
+	hasPermissions, err := CheckCacheDirectoryPermissions(dirPath)
+
+	ExpectEq(nil, dirCreationErr)
+	ExpectFalse(hasPermissions)
+	ExpectNe(nil, err)
+	ExpectTrue(strings.Contains(err.Error(), "error creating file at directory ("+dirPath+"), error"))
+}
+
+func generateRandomString() string {
```
sethiay, line 276: we can reuse this method https://github.com/GoogleCloudPlatform/gcsfuse/blob/read_cache_release/internal/util/test_util.go#L22

internal/fs/fs.go
```diff
@@ -246,10 +246,16 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
...
+	if !isPermissible {
+		panic(fmt.Sprintf("Cache Location Error: %v", err.Error()))
```
sethiay, line 251: nit: usually the error string starts with small case (and i would prefer string to be "error while creating file cache directory: %v"

internal/cache/util/util.go
ankitaluthra1: I see that statErr only has error string, dont think its good idea to comapre error string as we dont want to ti with hard code string value. `[ref](https://stackoverflow.com/questions/42035104/how-to-unit-test-go-errors)`

internal/cache/util/util.go
ankitaluthra1: This is being used in checking temp dir access which is live in prod, its okay to ignore this warning imo

ankitaluthra1: permission check is tricky, as we wont be sure if any umask is applied or with what type of user gcsfuse has been  mounted (like same user could be owner or other in different scenario), creating a file is sureshot way to know if we can create a file with current user.

internal/cache/util/util.go
ankitaluthra1: yes we are still creating it, it might be redundant now, we can remove it in future as tech debt, @sethiay please correct me in case I am mistaken

internal/cache/util/util.go
ankitaluthra1: my understanding was even if close returns an error, we would not want to fail mounting hence ignored it, let me know what should be our action for this error

internal/cache/util/util.go
ankitaluthra1: I also agree a method returning bool should start with is, but modifying this method's behaviour as per below comment and naming it accordingly
https://github.com/GoogleCloudPlatform/gcsfuse/pull/1639#discussion_r1465080889

internal/fs/fs.go
ankitaluthra1: done in latest commit

internal/cache/util/util.go
ankitaluthra1: done in latest commit

internal/fs/fs.go
ankitaluthra1: done with latest commit

internal/cache/util/util_test.go
ankitaluthra1: done with latest commit

internal/cache/util/util.go
ashmeenkaur: Can we compare it with syscall.ENOENT?

internal/cache/util/util.go
sethiay: it's redundant but not tech debt because the CreateFile call inside the cache_handler is still required and so is the MkdirAll call inside the CreateFile (because the cahe path for an object is like <user passed path> + gcsfuse-file-cache + <bucket_name> + <object_name> where object name can also be of type a/b/c. So MkdirAll is required for path after gcsfuse-file-cache as well.

internal/cache/util/util.go
sethiay: i think we should return error because it signals that something wrong happened while checking if directory is writable or not.

internal/cache/util/util.go
ankitaluthra1: got it ..

internal/cache/util/util.go
ankitaluthra1: Can you please validate my understanding...If any error occurs while creating annonymous we are already returning error and code panics, in addition to that now we also return error and panic if any error is returned from f.close() i.e. closing the annonymous file handle ? Also verified in temp dir we are ignoring this error
https://github.com/GoogleCloudPlatform/gcsfuse/blob/92dd6ae9d6feddc95397764fd678e5703de8bff3/mount.go#L50C3-L50C12

sethiay: So in general, we shouldn't ignore any error  because that can potentially mask bugs/problems (e.g. https://stackoverflow.com/questions/62594017/error-handing-in-go-when-to-ignore-errors)
So in the above case, either we should change f.close() method itself to not throw error at all or handle the error on our side.

ankitaluthra1: got it... have added changes to handle error in latest commit... ready for final review !

internal/cache/util/util_test.go
```diff
@@ -233,3 +234,37 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
...
+
+func TestCheckPermissionsForShouldNotReturnAnyErrorWhenDirectoryExists(t *testing.T) {
+	base := "./" + string(testutil.GenerateRandomBytes(4))
+	dirPath := base + "/" + "path/cachedir"
+	dirCreationErr := os.MkdirAll(dirPath, 0700)
+	defer os.RemoveAll(base)
+
+	err := CreateCacheDirectoryIfNotPresentAt(dirPath)
+
+	ExpectEq(nil, dirCreationErr)
```
sethiay, line 246: we should use AssertEq instead of ExpectEq. The difference is that AssertEq doesn't execute the unit test further if False. so e.g. in this case it is good to just stop this unit test if ExpectEq(nil, dirCreationErr) is not True.

Also, we should move ExpectEq(nil, dirCreationErr) before err := CreateCacheDirectoryIfNotPresentAt(dirPath) because ExpectEq(nil, dirCreationErr) has to be true for testing err := CreateCacheDirectoryIfNotPresentAt(dirPath)

internal/fs/fs.go
```diff
@@ -244,12 +244,20 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
...
+		panic(fmt.Sprintf("error while creating file cache directory: %v", cacheLocationErr.Error()))
+	}
+
 	// When user passes allow_other flag, then other users should be able to
 	// read from cache
 	filePerm := util.DefaultFilePerm
 	if cfg.AllowOther {
 		filePerm = util.FilePermWithAllowOther
 	}
+	//TODO : Add file permission here
```
sethiay, line 259: nit: remove this TODO

internal/cache/util/util_test.go
```diff
@@ -233,3 +234,37 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
...
+}
+
+func TestCheckPermissionsForShouldNotReturnAnyErrorWhenDirectoryCanBeCreated(t *testing.T) {
+	base := "./" + string(testutil.GenerateRandomBytes(4))
+	dirPath := base + "/" + "path/cachedir"
+	defer os.RemoveAll(base)
+
+	err := CreateCacheDirectoryIfNotPresentAt(dirPath)
+
+	ExpectEq(nil, err)
```
sethiay, line 257: we should also check if the directory is created and has expected dir permission bits set ?

internal/cache/util/util_test.go
```diff
@@ -233,3 +234,37 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
...
+func TestCheckPermissionsForShouldReturnErrorWhenDirectoryDoesNotHavePermissions(t *testing.T) {
+	dirPath := "./" + string(testutil.GenerateRandomBytes(4))
+	dirCreationErr := os.MkdirAll(dirPath, 0444)
+	defer os.RemoveAll(dirPath)
+
+	err := CreateCacheDirectoryIfNotPresentAt(dirPath)
+
+	ExpectEq(nil, dirCreationErr)
+	ExpectNe(nil, err)
+	ExpectTrue(strings.Contains(err.Error(), "error creating file at directory ("+dirPath+"), error"))
```
sethiay, line 269: what's the "error" in the end of string ? is this returned by f, err := fsutil.AnonymousFile(dirPath) ?

internal/fs/fs.go
```diff
@@ -244,12 +244,20 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
 	// gcsfuse related files.
 	cacheLocation = path.Join(cacheLocation, util.FileCache)
 
+	//Panic in case cacheLocation does not have required permissions
+	cacheLocationErr := util.CreateCacheDirectoryIfNotPresentAt(cacheLocation)
+	if cacheLocationErr != nil {
+		panic(fmt.Sprintf("error while creating file cache directory: %v", cacheLocationErr.Error()))
```
sethiay, line 250: can we start the error with "createFileCacheHandler: " ?

internal/cache/util/util.go
```diff
@@ -99,3 +101,29 @@ func IsCacheHandleInvalid(readErr error) bool {
...
+
+func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
+	_, statErr := os.Stat(dirPath)
+
+	if os.IsNotExist(statErr) {
+		err := os.MkdirAll(dirPath, FileDirPerm)
+		if err != nil {
+			return fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
+		}
+		return nil
```
sethiay, line 113: I think we shouldn't return here. Doing that will add additional write permission check which is fine ?

internal/fs/fs.go
```diff
@@ -244,12 +244,20 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
 	// gcsfuse related files.
 	cacheLocation = path.Join(cacheLocation, util.FileCache)
 
+	//Panic in case cacheLocation does not have required permissions
```
sethiay, line 247: nit: space between comment and first character of comment ("// P")

internal/cache/util/util.go
```diff
@@ -99,3 +101,29 @@ func IsCacheHandleInvalid(readErr error) bool {
...
+	}
+
+	f, err := fsutil.AnonymousFile(dirPath)
+	if err != nil {
+		return fmt.Errorf(
+			"error creating file at directory (%s), error : (%q)", dirPath, err.Error())
+	}
+
+	tempFileErr := f.Close()
```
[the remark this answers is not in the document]
ankitaluthra1: added different variable names for better readability of if blocks, looking at variable name only, we figure what that check is about.

internal/cache/util/util.go
```diff
@@ -99,3 +101,29 @@ func IsCacheHandleInvalid(readErr error) bool {
...
+		if err != nil {
+			return fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
+		}
+		return nil
+	}
+
+	f, err := fsutil.AnonymousFile(dirPath)
+	if err != nil {
+		return fmt.Errorf(
+			"error creating file at directory (%s), error : (%q)", dirPath, err.Error())
```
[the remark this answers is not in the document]
ankitaluthra1: done

internal/cache/util/util.go
```diff
@@ -99,3 +101,29 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), ErrInSeekingFileHandleMsg) ||
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
+
```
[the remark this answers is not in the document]
ankitaluthra1: done

internal/cache/util/util_test.go
```diff
@@ -233,3 +234,37 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
 		ExpectFalse(IsCacheHandleInvalid(errors.New(errMsg)))
 	}
 }
+
+func TestCheckPermissionsForShouldNotReturnAnyErrorWhenDirectoryExists(t *testing.T) {
```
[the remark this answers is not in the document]
ankitaluthra1: done

internal/cache/util/util_test.go
```diff
@@ -233,3 +234,37 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
...
+	base := "./" + string(testutil.GenerateRandomBytes(4))
+	dirPath := base + "/" + "path/cachedir"
```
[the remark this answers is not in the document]
ankitaluthra1: done

internal/cache/util/util.go
ankitaluthra1: yes, I was also in dilemma to have a redundant check or not, I chose this one as I wanted to avoid additional check in order to avoid increasing the mount time, but I am okay with both.

internal/fs/fs.go
ankitaluthra1: why not? done

internal/cache/util/util_test.go
ankitaluthra1: yes its returned when permission issue is there, I just wanted to assert that annonymous error string is returned not the create dir error hence added just contains, expected is not complete, since starting string confirms that error is from annoymous file

internal/cache/util/util_test.go
ankitaluthra1: updated as per comments, was trying previously to make tests formatted as per [ref](https://docs.telerik.com/devtools/justmock/basic-usage/arrange-act-assert#:~:text=Arrange%2FAct%2FAssert%20(AAA,more%20natural%20and%20convenient%20way.)

internal/fs/fs.go
ankitaluthra1: done

internal/cache/util/util_test.go
ankitaluthra1: done

internal/fs/fs.go
ankitaluthra1: done

internal/cache/util/util_test.go
```diff
@@ -233,3 +234,40 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
...
+	base := path.Join("./", string(testutil.GenerateRandomBytes(4)))
+	dirPath := path.Join(base, "/", "path/cachedir")
+	defer os.RemoveAll(base)
+
+	err := CreateCacheDirectoryIfNotPresentAt(dirPath)
+
+	AssertEq(nil, err)
+	fileInfo, err := os.Stat(dirPath)
+	AssertEq(nil, err)
+	AssertEq(0755, fileInfo.Mode().Perm())
```
sethiay, line 260: nit: should we reuse the FileDirPerm constant here ?

sethiay (approved):

internal/cache/util/util_test.go
ankitaluthra1: delibaretly added hard coded value so tests fails if FileDirPerm value is changed anytime, to make sure we explicitly know that impact of changing is also on this functionality

internal/cache/util/util.go
```diff
@@ -99,3 +101,30 @@ func IsCacheHandleInvalid(readErr error) bool {
...
+// Creates directory at given path with 766 permissions in case not already present,
```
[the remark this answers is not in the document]
ankitaluthra1: done

internal/cache/util/util_test.go
sethiay: have slight different view here - the functionality of the function is creating directory with permission stated by FileDirPerm and we are unit testing the functionality.

ankitaluthra1: the functionality of this function is to create a writable directory, we used FileDirPerm only for reusability purpose, imo we should test the functionality of writable directory and not link it with FileDirPerm value, does it make sense, what do you think ?

sethiay (approved):

PR #1676, opened 2024-02-01
Allow other cache dir fix

ankitaluthra1:
[Bug](https://b.corp.google.com/issues/322301637)

### Testing details
1. Manual - Done
[With -o allow_option] Tested by adding allow_other, users other than owner were able to view cache folder and files.
[Without -o allow_option] Tested without allow_other option, users other than owner were NOT able to cd into cache folder and was unable to see cache file name or content of cache dir.
3. Unit tests - Done
4. Integration tests - NA

### DIFF
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -104,11 +104,11 @@ func IsCacheHandleInvalid(readErr error) bool {
 
 // Creates directory at given path with FileDirPerm(0755) permissions in case not already present,
 // returns error in case unable to create directory or directory is not writable.
-func CreateCacheDirectoryIfNotPresentAt(dirPath string) error {
+func CreateCacheDirectoryIfNotPresentAt(dirPath string, dirPermissions os.FileMode) error {
 	_, statErr := os.Stat(dirPath)
 
 	if os.IsNotExist(statErr) {
-		err := os.MkdirAll(dirPath, FileDirPerm)
+		err := os.MkdirAll(dirPath, dirPermissions)
 		if err != nil {
 			return fmt.Errorf("error in creating directory structure %s: %v", dirPath, err)
 		}
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -55,7 +55,7 @@ const (
 func CreateFile(fileSpec data.FileSpec, flag int) (file *os.File, err error) {
 	// Create directory structure if not present
 	fileDir := filepath.Dir(fileSpec.Path)
-	err = os.MkdirAll(fileDir, FileDirPerm)
+	err = os.MkdirAll(fileDir, fileSpec.Perm.Perm())
 	if err != nil {
 		err = fmt.Errorf(fmt.Sprintf("error in creating directory structure %s: %v", fileDir, err))
 		return
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -42,8 +42,8 @@ const (
 	FileDirPerm            = os.FileMode(0755) | os.ModeDir
 	MiB                    = 1024 * 1024
 	KiB                    = 1024
-	DefaultFilePerm        = os.FileMode(0600)
-	FilePermWithAllowOther = os.FileMode(0644)
+	DefaultFilePerm        = os.FileMode(0700)
+	FilePermWithAllowOther = os.FileMode(0755)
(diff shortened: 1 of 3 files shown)


internal/cache/util/util_test.go
```diff
@@ -136,18 +149,19 @@ func (ut *utilTest) Test_CreateFile_FilePerm0755() {
...
 
-func (ut *utilTest) Test_CreateFile_FilePerm0544() {
-	ut.fileSpec.Perm = os.FileMode(0544)
-
-	file, err := CreateFile(ut.fileSpec, ut.flag)
-
-	ut.assertFileAndDirCreation(file, err)
-	ExpectEq(nil, file.Close())
-}
+// TODO: should we remove this test as we use same perms for file and dir and dir needs min 7 perm as owner to be able to create file
```
sethiay, line 156: yes we can remove this. Since the function now creates both file and dirs with same permissions. we need to adjust our tests accordingly, so you can write tests (if not already present) 
(a) if directory is already present then different file perms including <700.
(b) if directory is not already present then couple of cases creating dir+files for perm >=700 
(c) if directory is not already present then couple of cases trying to create dir+files for perm <700 should fail.

internal/cache/util/util_test.go
```diff
@@ -64,19 +64,32 @@ func (ut *utilTest) TearDown() {
 }
 
 func (ut *utilTest) assertFileAndDirCreation(file *os.File, err error) {
+	ut.assertFileAndDirStatsWithGivenFileAndDirPermissions(file, err, ut.fileSpec.Perm, 0700)
+}
+
+func (ut *utilTest) assertFileAndDirCreationWithGivenDirMode(file *os.File, err error, dirMode os.FileMode) {
```
sethiay, line 70: can we combine these functions (just keep 3rd one) and use that everywhere ?

internal/cache/util/util.go
```diff
@@ -104,11 +103,11 @@ func IsCacheHandleInvalid(readErr error) bool {
 
 // Creates directory at given path with FileDirPerm(0755) permissions in case not already present,
```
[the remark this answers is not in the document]
ankitaluthra1: done in latest commit

internal/cache/util/util.go
```diff
@@ -55,7 +54,7 @@ const (
 func CreateFile(fileSpec data.FileSpec, flag int) (file *os.File, err error) {
```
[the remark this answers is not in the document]
ankitaluthra1: done in latest commit

internal/cache/util/util_test.go
ankitaluthra1: kept it deliberately like this so the tests using this function (eg: Test_CreateFile_ReadOnlyFile, Test_CreateFile_ReadWriteFile) which assume that directory is already created with default permissions dont need to have additional information that directory should have 700 permissions, its a good LLD practice to segregate responsibilities among different methods/classes as much as possible

internal/cache/util/util_test.go
ankitaluthra1: yes we can remove this. Since the function now creates both file and dirs with same permissions. we need to adjust our tests accordingly, so you can write tests (if not already present) 
(a) if directory is already present then different file perms including <700. _[already present - Test_CreateFile_FilePresentWithLessAccess]_
(b) if directory is not already present then couple of cases creating dir+files for perm >=700 _[multiple scenarios are already present for this]_
(c) if directory is not already present then couple of cases trying to create dir+files for perm <700 should fail. _[modified this commented test for this usecase]_

internal/cache/util/util_test.go
ankitaluthra1: Just an fyi, generalising assertions into a common method is not a good practice, it breaks the AAA format where Assert should also be inside the test, also need for generalising assertions is a hint that we might be writing too many assertions in a single test which again is anti-pattern, ideally we should write one assertion per test but thats not feasible in real world scenarios but at least we can try to minimise and break the tests if needed
 if possible we should avoid these things in new tests
Short read on best practices : https://www.browserstack.com/guide/unit-testing-best-practices#:~:text=accomplish%20its%20purpose.-,Best%20Practices%20for%20Unit%20Testing,-These%20unit%20testing

PR #1682, opened 2024-02-06
Optimize memory by removing async job after completion/failure

sethiay:
### Description
Optimize memory by removing async job after completion/failure
Functional change:
1. The async job is removed from memory once the job is stopped (i.e. complete/failed) or invalidated in case of eviction. We are doing this to reduce the overall memory footprint of GCSFuse with read cache is enabled.

Code changes:
1. Documentation fixes and changes corresponding to code changes.
3. Unit test fixes for test files under internal/cache/. This includes not using the same function in "Arrange" part of unit test that is being tested in unit test.
4. Addition of few more unit test cases.
5. job.go: (a) Added a property called removeJobCallback to Job struct which is call back function to remove job from JobManager. (b) Call removeJobCallback function when the downloadAsyncJob stops (completed/failed/cancelled). (c) Call removeJobCallback function when the job.Invalidate function is called. (d) Once the removeJobCallback function is called, we remove it from Job struct so that it's not called more than once for that job. 
6. downloader.go: (a) Changed the behavior of GetJob function to only return Job if JobManager has it. (b) Added new function CreateJobIfNotExists which creates new job and stores it's reference in JobManager if Job does not already exist, if it exists then the existing job is returned. (c) Add removeJob function which simply removes the job reference from jobs map if present (d) Add a callback function to Job while creating new Job, this callback is nothing but job.removeJob function. (e) Rename the RemoveJob function to InvalidateAndRemoveJob, also this function now releases the jm.mu lock while calling job.Invalidate inside it because job needs to acquire lock on jm.mu while executing jm.removeJob.
7. cache_handle.go: (a) Allow having nil download job. (b) if download job is nil, then check the entry in file info cache (without changing it's order) whether the entry has offset equal to object's size (if offset < object size and there is no job then it means job has failed/invalidated). 
8. cache_handler.go: (a) in the GetCacheHandle function for the case where there is already entry in the file info cache, check whether there is non-nil job that is not failed/invalidated. if there is no such job then invalidate the cache for that file and add new entry to  cache for that file.
9. Misc: (a) Removed job.Cancel function as that was not used anywhere functionally in the code at other places. (b) Used camel case for JobStatus.Name as go suggests to use camel case for constants. (c) Remove Cancel state from Job as that is no longer required.

The PR is longer than what the recommended size should be but it was harder to split it into a way such that every split PR passes all presubmit tests.

NA

### Testing details
1. Manual - Ran tests for 64K 1million files and observed clear memory improvements, results to be shared internally.
7. Unit tests - Added wherever applicable
9. Integration tests - [Link](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/4c337ea4-38e2-4cbd-8fe8-9f555246ecd7)

### DIFF
--- a/internal/cache/file/cache_handle.go
+++ b/internal/cache/file/cache_handle.go
@@ -32,26 +32,27 @@ type CacheHandle struct {
 	// fileHandle to a local file which contains locally downloaded data.
 	fileHandle *os.File
 
-	//	fileDownloadJob is a reference to async download Job.
+	// fileDownloadJob is a reference to async download Job.
 	fileDownloadJob *downloader.Job
 
-	// fileInfoCache contains the reference of fileInfo cache.
+	// fileInfoCache contains the reference to fileInfo cache.
 	fileInfoCache *lru.Cache
 
-	// cacheFileForRangeRead if true, object content will be downloaded for random
-	// reads as well too.
+	// cacheFileForRangeRead if true, async download job will start even for range
+	// reads.
 	cacheFileForRangeRead bool
 
 	// isSequential saves if the current read performed via cache handle is sequential or
 	// random.
 	isSequential bool
 
-	// prevOffset stores the offset of previous cache_handle read call. This is used
+	// prevOffset stores the offset of previous cache handle read call. This is used
 	// to decide the type of read.
 	prevOffset int64
 }
 
-func NewCacheHandle(localFileHandle *os.File, fileDownloadJob *downloader.Job, fileInfoCache *lru.Cache, cacheFileForRangeRead bool, initialOffset int64) *CacheHandle {
+func NewCacheHandle(localFileHandle *os.File, fileDownloadJob *downloader.Job,
+	fileInfoCache *lru.Cache, cacheFileForRangeRead bool, initialOffset int64) *CacheHandle {
 	return &CacheHandle{
 		fileHandle:            localFileHandle,
 		fileDownloadJob:       fileDownloadJob,
@@ -67,10 +68,6 @@ func (fch *CacheHandle) validateCacheHandle() error {
 		return errors.New(util.InvalidFileHandleErrMsg)
 	}
 
-	if fch.fileDownloadJob == nil {
-		return errors.New(util.InvalidFileDownloadJobErrMsg)
-	}
-
 	if fch.fileInfoCache == nil {
 		return errors.New(util.InvalidFileInfoCacheErrMsg)
 	}
@@ -78,12 +75,12 @@ func (fch *CacheHandle) validateCacheHandle() error {
 	return nil
 }
 
-// shouldReadFromCache returns nil if the data should be read from the local,
-// downloaded file. Otherwise, it returns a non-nil error with an appropriate error message.
+// shouldReadFromCache returns nil if the data should be read from the locally
+// downloaded cache file. Otherwise, it returns an appropriate error message.
 func (fch *CacheHandle) shouldReadFromCache(jobStatus *downloader.JobStatus, requiredOffset int64) (err error) {
 	if jobStatus.Err != nil ||
-		jobStatus.Name == downloader.INVALID ||
-		jobStatus.Name == downloader.FAILED {
+		jobStatus.Name == downloader.Invalid ||
+		jobStatus.Name == downloader.Failed {
 		err := fmt.Errorf("%s: jobStatus: %s jobError: %w", util.InvalidFileDownloadJobErrMsg, jobStatus.Name, jobStatus.Err)
 		return err
 	} else if jobStatus.Offset < requiredOffset {
@@ -94,33 +91,41 @@ func (fch *CacheHandle) shouldReadFromCache(jobStatus *downloader.JobStatus, req
 }
 
 // checkEntryInFileInfoCache checks if entry is present for a given object in
-// file info cache with same generation. It returns nil if entry is present,
-// otherwise returns an error with appropriate message.
-func (fch *CacheHandle) checkEntryInFileInfoCache(bucket gcs.Bucket, object *gcs.MinObject, requiredOffset int64) error {
+// file info cache with same generation and at least requiredOffset.
+// It returns nil if entry is present, otherwise returns an appropriate error.
+// Whether to change the order in cache while lookup is controlled via
+// changeCacheOrder.
+func (fch *CacheHandle) checkEntryInFileInfoCache(bucket gcs.Bucket, object *gcs.MinObject, requiredOffset uint64, changeCacheOrder bool) error {
 	fileInfoKey := data.FileInfoKey{
 		BucketName: bucket.Name(),
 		ObjectName: object.Name,
 	}
 	fileInfoKeyName, err := fileInfoKey.Key()
 	if err != nil {
-		return fmt.Errorf("read: while creating key: %v", fileInfoKeyName)
+		return fmt.Errorf("error while creating key for bucket %s and object %s: %w", bucket.Name(), object.Name, err)
 	}
-	// Look up of file being read in file info cache is required to update the LRU
-	// order on every read request from kernel i.e. with every read request from
-	// kernel, the file being read becomes most recently used.
-	fileInfo := fch.fileInfoCache.LookUp(fileInfoKeyName)
 
-	// The generation check below is required because it may happen that file
-	// being read is evicted from cache during or after reading the required offset
-	// from local cached file to `dst` buffer.
+	var fileInfo lru.ValueType
+	if changeCacheOrder {
+		fileInfo = fch.fileInfoCache.LookUp(fileInfoKeyName)
+	} else {
+		fileInfo = fch.fileInfoCache.LookUpWithoutChangingOrder(fileInfoKeyName)
+	}
 	if fileInfo == nil {
 		err = fmt.Errorf("%v: no entry found in file info cache for key %v", util.InvalidFileInfoCacheErrMsg, fileInfoKeyName)
 		return err
 	}
 
+	// The generation check below is required because it may happen that file
+	// being read is evicted from cache during or after reading the required offset
+	// from local cached file to `dst` buffer.
 	fileInfoData := fileInfo.(data.FileInfo)
 	if fileInfoData.ObjectGeneration != object.Generation {
 		err = fmt.Errorf("%v: generation of cached object: %v is different from required generation: %v", util.InvalidFileInfoCacheErrMsg, fileInfoData.ObjectGeneration, object.Generation)
+		return err
+	}
+	if fileInfoData.Offset < requiredOffset {
+		err = fmt.Errorf("%v offset of cached object: %v is less than required offset %v", util.InvalidFileInfoCacheErrMsg, fileInfoData.Offset, requiredOffset)
 		return err
 	}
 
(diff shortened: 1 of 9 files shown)


internal/cache/file/cache_handle.go
```diff
@@ -32,26 +32,27 @@ type CacheHandle struct {
 	// fileHandle to a local file which contains locally downloaded data.
 	fileHandle *os.File
 
-	//	fileDownloadJob is a reference to async download Job.
+	// fileDownloadJob is a reference to async download Job.
```
raj-prince, line 35: Can we write a comment about - it's life-cycle. Like,  this will become nil once Job is finished/cancelled or invalidated.

internal/cache/file/cache_handle.go
```diff
@@ -94,35 +91,43 @@ func (fch *CacheHandle) shouldReadFromCache(jobStatus *downloader.JobStatus, req
...
-	fileInfo := fch.fileInfoCache.LookUp(fileInfoKeyName)
 
-	// The generation check below is required because it may happen that file
-	// being read is evicted from cache during or after reading the required offset
-	// from local cached file to `dst` buffer.
+	var fileInfo lru.ValueType
+	if changeCacheOrder {
+		fileInfo = fch.fileInfoCache.LookUp(fileInfoKeyName)
+	} else {
+		fileInfo = fch.fileInfoCache.LookUpWithoutChangingOrder(fileInfoKeyName)
```
raj-prince, line 112: No action required, just a thought: our code has tightly coupled with LRU cache behavior, we might need to refactor, if we would support different cache policy.

internal/cache/file/cache_handle.go
```diff
@@ -94,35 +91,43 @@ func (fch *CacheHandle) shouldReadFromCache(jobStatus *downloader.JobStatus, req
...
 
 // checkEntryInFileInfoCache checks if entry is present for a given object in
-// file info cache with same generation. It returns nil if entry is present,
-// otherwise returns an error with appropriate message.
-func (fch *CacheHandle) checkEntryInFileInfoCache(bucket gcs.Bucket, object *gcs.MinObject, requiredOffset int64) error {
+// file info cache with same generation and at least requiredOffset.
+// It returns nil if entry is present, otherwise returns an appropriate error.
+// Whether to change the order in cache while lookup is controlled via
+// changeCacheOrder.
+func (fch *CacheHandle) checkEntryInFileInfoCache(bucket gcs.Bucket, object *gcs.MinObject, requiredOffset uint64, changeCacheOrder bool) error {
```
raj-prince, line 98: Can we rename it to checkValidEntryInFileInfoCache, as we do some sort of validation on offset and generation.

internal/cache/file/cache_handle.go
```diff
@@ -161,30 +166,44 @@ func (fch *CacheHandle) Read(ctx context.Context, bucket gcs.Bucket, object *gcs
...
 
-	jobStatus := fch.fileDownloadJob.GetStatus()
-	// If cacheFileForRangeRead is false and readType is random, download will not be initiated.
-	if !fch.cacheFileForRangeRead && !isSequentialRead {
-		if err = fch.shouldReadFromCache(&jobStatus, requiredOffset); err != nil {
-			return 0, false, err
+	// If fileDownloadJob is not nil, it's better to get status of cache file
+	// from the job itself than to use file info cache.
+	if fch.fileDownloadJob != nil {
+		jobStatus := fch.fileDownloadJob.GetStatus()
```
raj-prince, line 172: As discussed offline, let's add a check to make the fileDownloadJob nil, if jobStatus is failed or Invalid.

internal/cache/file/cache_handle_test.go
```diff
@@ -484,7 +590,7 @@ func (cht *cacheHandleTest) Test_Read_RandomWithNoRandomDownload() {
 	n, cacheHit, err := cht.cacheHandle.Read(context.Background(), cht.bucket, cht.object, offset, dst)
 
 	jobStatus := cht.cacheHandle.fileDownloadJob.GetStatus()
-	ExpectEq(jobStatus.Name, downloader.NOT_STARTED)
+	ExpectEq(jobStatus.Name, downloader.NotStarted)
```
raj-prince, line 593: nit: we need to change the order of input parameter, as we did below in line: 606.

internal/cache/file/downloader/job.go
```diff
@@ -37,12 +36,11 @@ import (
 type jobStatusName string
 
 const (
-	NOT_STARTED jobStatusName = "NOT_STARTED"
-	DOWNLOADING jobStatusName = "DOWNLOADING"
-	COMPLETED   jobStatusName = "COMPLETED"
-	FAILED      jobStatusName = "FAILED"
-	CANCELLED   jobStatusName = "CANCELLED"
-	INVALID     jobStatusName = "INVALID"
+	NotStarted  jobStatusName = "NotStarted"
```
raj-prince, line 39: can we rename it to jobStatusName to jobState?

internal/cache/file/downloader/job.go
```diff
@@ -341,10 +318,6 @@ func (job *Job) downloadObjectAsync() {
 							ReadCompressed: job.object.HasContentEncodingGzip(),
 						})
 					if err != nil {
-						if errors.Is(err, context.Canceled) {
```
raj-prince, line 344: Just thinking, if someone will call invalidate anyway we would be removing the job. Hence, we are not handling here?

internal/cache/file/cache_handler_test.go
```diff
@@ -177,161 +176,239 @@ func (chrT *cacheHandlerTest) Test_createLocalFileReadHandle_OnlyForRead() {
...
+	err = chrT.cacheHandler.addFileInfoEntryAndCreateDownloadJob(chrT.object, chrT.bucket)
+
+	ExpectEq(nil, err)
+	ExpectTrue(chrT.isEntryInFileInfoCache(chrT.object.Name, chrT.bucket.Name()))
+	// No new job should be added to job manager
+	actualJob = chrT.jobManager.GetJob(chrT.object.Name, chrT.bucket.Name())
+	ExpectEq(nil, actualJob)
+}
+
+func (chrT *cacheHandlerTest) Test_addFileInfoEntryAndCreateDownloadJob_WhenJobIsInvalidatedAndRemoved() {
```
raj-prince, line 302: Can we rename to - Test_..._WhenUnfinishedJobIsInvalidatedAndRemoved?

If we call InvalidateAndRemove on finished job, in that case nothing will change. right?

raj-prince (commented):
Overall LGTM. Few minor comments.

internal/cache/file/downloader/job.go
raj-prince: We can skip for now, since we are in time-crunch.

sethiay: Also, I would prefer creating a new PR for that since this is already a long PR and focus is on cleaning up completed job.

internal/cache/file/downloader/job.go
sethiay: We have removed the cancellation method from Job so cancellation should not happen and if it ever happens for any reason it should be reported as failure.

raj-prince: When cache is invalidated at that time `job.cancel` is called. Right? And ideally that's not a failure case. 
Or let me know, if I am missing something.

sethiay: There is no harm in that:
1. The error in failure will say context cancelled.
2. When the job is Invalidated or Failed, the behavior is same for caller i.e. to not read from cache. (earlier it was not because the caller can read from Cancelled state).

If you still think it creates confusion logically then we can notify with Invalid instead of Failure.

internal/cache/file/cache_handle.go
sethiay: I think just "Entry" should be fine here because an entry is also valid if it has offset < requiredOffset. What the code is doing here is essentially checking whether the entry is present and if it is present whether it has offset and generation that is required by it. LMK if you think otherwise.

internal/cache/file/cache_handle_test.go
```diff
@@ -506,52 +612,82 @@ func (cht *cacheHandleTest) Test_Read_RandomWithNoRandomDownloadButCacheHit() {
...
-func (cht *cacheHandleTest) Test_Read_RandomWithNoRandomDownloadButCacheHitInCancelledState() {
-	ctx := context.Background()
-	// Download the job till util.MiB
-	jobStatus, err := cht.cacheHandle.fileDownloadJob.Download(ctx, int64(2*util.MiB), true)
-	AssertEq(nil, err)
-	AssertEq(jobStatus.Name, downloader.DOWNLOADING)
-	cht.cacheHandle.fileDownloadJob.Cancel()
-	jobStatus = cht.cacheHandle.fileDownloadJob.GetStatus()
-	AssertEq(jobStatus.Name, downloader.CANCELLED)
+func (cht *cacheHandleTest) Test_Read_Sequential() {
```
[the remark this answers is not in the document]
sethiay: It is an old test, here it is shown as newly added but at the other line also shows it deleted. So I think we are good here.

internal/cache/file/cache_handler_test.go
sethiay: I think it's fine to keep it WhenJobIsInvalidatedAndRemoved because addFileInfoEntryAndCreateDownloadJob doesn't care about whether the job was unfinished or not before it was invalidated. The purpose of this test is to check the behavior of addFileInfoEntryAndCreateDownloadJob when the Job is invalidated and/or removed.

internal/cache/file/cache_handle.go
sethiay: thought again about this and I think it's fine to not add additional check here as file cache handle will automatically be closed here: https://github.com/GoogleCloudPlatform/gcsfuse/blob/read_cache_release/internal/gcsx/random_reader.go#L237 if fileDownloadJob is Failed or Invalid.

internal/cache/file/downloader/job.go
sethiay: Done, notify with Invalid in case context is canceled to avoid giving logger.Errorf in logs (that comes in notify failure) which can create little confusion.

internal/cache/file/downloader/job.go
```diff
@@ -318,6 +326,12 @@ func (job *Job) downloadObjectAsync() {
 							ReadCompressed: job.object.HasContentEncodingGzip(),
 						})
 					if err != nil {
+						// Context is canceled when job.cancel is called at the tine of
```
raj-prince, line 329: nit: tine -> time

internal/cache/file/downloader/job.go
```diff
@@ -336,6 +350,12 @@ func (job *Job) downloadObjectAsync() {
 				// Copy the contents from NewReader to cache file.
 				_, readErr := io.CopyN(cacheFile, newReader, maxRead)
 				if readErr != nil {
+					// Context is canceled when job.cancel is called at the tine of
```
raj-prince, line 353: nit: tine -> time

internal/cache/file/cache_handle.go
raj-prince: Okay, with the assumption user can open max around 1024 files (unless explicitly) which would cost around half MB. We can skip.

internal/cache/file/cache_handler_test.go
raj-prince: Okay, make sense.

internal/cache/file/cache_handle.go
raj-prince: Still feel like, `checkEntryInFileInfoCache` is too generic name. Can we rename it better if not checkValidEntryInFileInfoCache or checkRequiredEntryInFileInCache.

raj-prince (approved):

PR #1685, opened 2024-02-06
CD script changes to support beta release flow

raj-prince:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/package_gcsfuse.sh
+++ b/tools/cd_scripts/package_gcsfuse.sh
@@ -64,6 +64,10 @@ function exit_in_failure() {
 RELEASE_VERSION=$(fetch_meta_data_value "RELEASE_VERSION")
 echo RELEASE_VERSION="$RELEASE_VERSION"
 
+# '~' is not accepted as docker build suffix.
+DOCKER_BUILD_TAG_SUFFIX=$(echo $RELEASE_VERSION | tr '~' '_')
+echo DOCKER_BUILD_TAG_SUFFIX="$DOCKER_BUILD_TAG_SUFFIX"
+
 # Fetch metadata value of the key "UPLOAD_BUCKET"
 UPLOAD_BUCKET=$(fetch_meta_data_value "UPLOAD_BUCKET")
 echo UPLOAD_BUCKET="$UPLOAD_BUCKET"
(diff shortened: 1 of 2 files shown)


tools/cd_scripts/package_gcsfuse.sh
```diff
@@ -80,20 +84,21 @@ sudo apt-get install git -y
 # It is require for multi-arch support
 sudo apt-get install qemu-user-static binfmt-support
 
-git clone https://github.com/GoogleCloudPlatform/gcsfuse.git
+# TODO (raj-prince): to remove this branch once read-cache-release branched is merged into the master branch.
+git clone -b "beta_louhi_cd_script_final_change" https://github.com/GoogleCloudPlatform/gcsfuse.git
```
raj-prince, line 88: Reminder: have to change "read_cache_release" branch.

tools/cd_scripts/package_gcsfuse.sh
```diff
@@ -64,6 +64,10 @@ function exit_in_failure() {
 RELEASE_VERSION=$(fetch_meta_data_value "RELEASE_VERSION")
 echo RELEASE_VERSION="$RELEASE_VERSION"
 
+# '~' is not accepted as docker build tag and git tag. Hence, we will it instead of RELEASE_VERSION.
```
ashmeenkaur, line 67:
```suggestion
# '~' is not accepted as docker build tag and git tag. Hence, we will use '_' instead of RELEASE_VERSION.
```

tools/cd_scripts/package_gcsfuse.sh
```diff
@@ -64,6 +64,10 @@ function exit_in_failure() {
...
+RELEASE_VERSION_TAG=$(echo $RELEASE_VERSION | tr '~' '_')
+echo RELEASE_VERSION_TAG="$DOCKER_BUILD_TAG_SUFFIX"
```
ashmeenkaur, line 69: I didn't follow this change. What is DOCKER_BUILD_TAG_SUFFIX?

tools/cd_scripts/package_gcsfuse.sh
ashmeenkaur: This should be master for non beta release?

tools/package_gcsfuse_docker/Dockerfile
```diff
@@ -37,7 +37,12 @@ ARG DEBFULLNAME="GCSFuse Team"
 # Build Arg for building through a particular branch/commit. By default, it uses
 # the tag corresponding to passed GCSFUSE VERSION
 ARG BRANCH_NAME="v${GCSFUSE_VERSION}"
-RUN git checkout "${BRANCH_NAME}"
+ARG GIT_BUILD_TAG
```
ashmeenkaur, line 40: NIT: Another possible way to do this would have been not making this change and just passing `--build-arg BRANCH_NAME ="v$RELEASE_VERSION_TAG"` while building the dockerfile. We probably could have renamed `BRANCH_NAME` to something more generic like `GIT_CHECKOUT`

sethiay: +1, we should use BRANCH_NAME and rename it later as separate task at all places

tools/cd_scripts/package_gcsfuse.sh
raj-prince: Missed by mistake, removing that.

tools/package_gcsfuse_docker/Dockerfile
raj-prince: Ohh, got it.  "v${GCSFUSE_VERSION}" is the default value. Thanks!

tools/cd_scripts/package_gcsfuse.sh
raj-prince: Yes.

tools/cd_scripts/package_gcsfuse.sh
```diff
@@ -80,20 +84,21 @@ sudo apt-get install git -y
 # It is require for multi-arch support
 sudo apt-get install qemu-user-static binfmt-support
 
-git clone https://github.com/GoogleCloudPlatform/gcsfuse.git
+# TODO (raj-prince): to remove this branch once read-cache-release branched is merged into the master branch.
+git clone -b "read_cache_release" https://github.com/GoogleCloudPlatform/gcsfuse.git
```
ashmeenkaur, line 88:
```suggestion
git clone https://github.com/GoogleCloudPlatform/gcsfuse.git
git checkout <commit hash from details.txt>
```

ashmeenkaur: My bad, just realised that we create details.txt after building the package. We will have to move that stage prior or pass commit hash as metadata to the VM.

ashmeenkaur: Discussed offline that we will do this later to unblock the beta release.

ashmeenkaur (approved):

sethiay (approved):

PR #1686, opened 2024-02-07
Latest allow other cache dir fix

ankitaluthra1:
[Link](https://b.corp.google.com/issues/322301637)

### Testing details
1. Manual - Done
[With -o allow_option] Tested by adding allow_other, users other than owner were able to view cache folder and files.
[Without -o allow_option] Tested without allow_other option, users other than owner were NOT able to cd into cache folder and was unable to see cache file name or content of cache dir.
3. Unit tests - Done
4. Integration tests - Added new card for automation

### DIFF
--- a/internal/cache/file/cache_handle_test.go
+++ b/internal/cache/file/cache_handle_test.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
@@ -82,9 +83,10 @@ func (cht *cacheHandleTest) addTestFileInfoEntryInCache() {
 
 func (cht *cacheHandleTest) verifyContentRead(readStartOffset int64, expectedContent []byte) {
 	fileStat, err := os.Stat(cht.fileSpec.Path)
+	dirStat, err := os.Stat(filepath.Dir(cht.fileSpec.Path))
 	AssertEq(nil, err)
 	AssertEq(cht.fileSpec.FilePerm, fileStat.Mode())
-	AssertEq(cht.fileSpec.DirPerm, 0700)
+	AssertEq(cht.fileSpec.DirPerm, dirStat.Mode().Perm())
 
 	// Create a byte buffer of same len as expectedContent.
 	buf := make([]byte, len(expectedContent))
--- a/internal/cache/file/cache_handle_test.go
+++ b/internal/cache/file/cache_handle_test.go
@@ -84,6 +84,7 @@ func (cht *cacheHandleTest) verifyContentRead(readStartOffset int64, expectedCon
 	fileStat, err := os.Stat(cht.fileSpec.Path)
 	AssertEq(nil, err)
 	AssertEq(cht.fileSpec.Perm, fileStat.Mode())
+	AssertEq(cht.fileSpec.DirPerm, 0700)
 
 	// Create a byte buffer of same len as expectedContent.
 	buf := make([]byte, len(expectedContent))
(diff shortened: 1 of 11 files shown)


internal/cache/file/cache_handle_test.go
```diff
@@ -83,7 +83,8 @@ func (cht *cacheHandleTest) addTestFileInfoEntryInCache() {
 func (cht *cacheHandleTest) verifyContentRead(readStartOffset int64, expectedContent []byte) {
 	fileStat, err := os.Stat(cht.fileSpec.Path)
 	AssertEq(nil, err)
-	AssertEq(cht.fileSpec.Perm, fileStat.Mode())
+	AssertEq(cht.fileSpec.FilePerm, fileStat.Mode())
+	AssertEq(cht.fileSpec.DirPerm, 0700)
```
sethiay, line 87: Since we set cht.fileSpec.DirPerm, so it will not help if we assert that with a fixed value. 
What we need to do is get the directory of fileSpec.Path using path/filepath.Dir(fileSpec.Path), then do os.Stat(dirPath) and then check if it's same as cht.fileSpec.DirPerm.

internal/cache/file/downloader/job_test.go
```diff
@@ -60,7 +60,7 @@ func (dt *downloaderTest) initJobTest(objectName string, objectContent []byte, s
 	err := storageutil.CreateObjects(ctx, dt.bucket, objects)
 	AssertEq(nil, err)
 	dt.object = dt.getMinObject(objectName)
-	dt.fileSpec = data.FileSpec{Path: dt.fileCachePath(dt.bucket.Name(), dt.object.Name), Perm: util.DefaultFilePerm}
+	dt.fileSpec = data.FileSpec{Path: dt.fileCachePath(dt.bucket.Name(), dt.object.Name), FilePerm: util.DefaultFilePerm}
```
sethiay, line 63: add DirPerm to data.FileSpec. I think that's the cause of failure in presubmit tests. (it is taking default as 000 probably)

internal/cache/util/util_test.go
```diff
@@ -85,17 +86,18 @@ func (ut *utilTest) Test_CreateFile_FileDirNotPresent() {
```
sethiay, line 86: I think we should add test case where file dir is not present and we try to create file with dirPerm < 600

internal/cache/file/cache_handle_test.go
ankitaluthra1: done

internal/cache/file/downloader/job_test.go
ankitaluthra1: hmm could be...added default dirPerm it should fail on my local, maybe it takes system's default permission, server machine might have stricter permissions than cloudtop machines.

internal/cache/util/util_test.go
ankitaluthra1: Done

internal/cache/file/cache_handle_test.go
```diff
@@ -82,8 +83,10 @@ func (cht *cacheHandleTest) addTestFileInfoEntryInCache() {
 
 func (cht *cacheHandleTest) verifyContentRead(readStartOffset int64, expectedContent []byte) {
 	fileStat, err := os.Stat(cht.fileSpec.Path)
+	dirStat, err := os.Stat(filepath.Dir(cht.fileSpec.Path))
```
sethiay, line 86: Add AssertEq(nil, err) for "err" in fileStat, err also.

ankitaluthra1: done

sethiay (approved):

PR #1688, opened 2024-02-08
Fix metadata ttl deprecation texts

gargnitingoogle:
### Description
Fix deprecation texts for metadata-cache ttl

Fixing the language used for deprecation for
metadata-cache ttl (i.e. stat-cache-ttl and
type-cache-ttl) from 'will be deprecated' to
'has been deprecated' as the latter is the
correct wording. Deprecation doesn't mean
removal, it only mean strong discouragement
from the use of a feature.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -73,7 +73,7 @@ The behavior of stat cache is controlled by the following flags/config parameter
 2. Stat-cache TTL: It controls the duration for which Cloud Storage FUSE allows the kernel to cache inode attributes. It can be set in one of the following two ways.
    * ```metadata-cache: ttl-secs``` in the config-file. This is set as an integer, which sets the TTL in seconds. If this is -1, TTL is taken as infinite i.e. no-TTL based expirations of entries. If this is 0, that disables the stat-cache.
    If this config variable is missing, then the value of ```--stat-cache-ttl``` is used.
-   * ```--stat-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This will be deprecated in a future version and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--stat-cache-ttl``` is ignored.
+   * ```--stat-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This has been deprecated and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--stat-cache-ttl``` is ignored.
    
    Positive and negative stat results will be cached for the specified amount of time.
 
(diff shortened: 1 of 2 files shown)


docs/semantics.md
```diff
@@ -99,14 +99,14 @@ The behavior of type cache is controlled by the following flags/config parameter
 
 **Type-cache TTL**: It controls the duration for which Cloud Storage FUSE caches an inode's type attribute. It can be set in one of the following two ways.
 * ```metadata-cache: ttl-secs``` in the config-file. This is set as an integer, which sets the TTL in seconds. If this is -1, TTL is taken as infinite i.e. no-TTL based expirations of entries. If this is 0, that disables the type-cache. If this is <-1, then an error is thrown on mount.
-* ```--type-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This will be deprecated in a future version and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--type-cache-ttl``` is ignored.
+* ```--type-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This has been deprecated and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--type-cache-ttl``` is ignored.
```
mustvicky, line 102: Can we say "This has been deprecated starting GCSFuse V2.0.0" or you think that is not important here?

mustvicky (approved):
Thanks. Good catch!. 
I left an additional comment to make it version specific.. without that version specificity also is an improvement from what we have.

docs/semantics.md
gargnitingoogle: > Can we say "This has been deprecated starting GCSFuse V2.0.0"

Probably a bit premature to use a particular version number in the code before it has actually released.

Since this is part of the read-cache branch, it'll be merged to master and be released along
with the actual code change itself.

gargnitingoogle: @mustvicky let me know if you feel strongly about it.. I can add reference to V2.0.0 here.

mustvicky: I feel medium strongly about this. We have only 25% users using V1. V2 will be even lower adoption I am assuming initially and then increase over time. Personally, it is better for understanding which version what is deprecated. 

We can remove V2.0.0 and just say V2? That will make it less specific . .at the same time make it future correct.

mustvicky (approved):

gargnitingoogle:
This isn't needed anymore as this change has already been included in https://github.com/GoogleCloudPlatform/gcsfuse/commit/13ed8af9c27b4c2bcf4c55fa460cb3f1df832c70 .

gargnitingoogle:
Closing this as explained in https://github.com/GoogleCloudPlatform/gcsfuse/pull/1688#issuecomment-1946194928 .

PR #1687, opened 2024-02-08
Switch type cache to size mb

gargnitingoogle:
### Description
Switches type-cache-max-entries to type-cache-max-size-mb

More details of the change.
* Adds utility MibToBytes
* Adds support for metadata-cache:type-cache-max-size-mb in config-file,
  and corresponding unit tests.
* Updates tests for type-cache-configs in dirInode
* Adds documentation for type-cache-max-size-mb in semantics.md

This resurrects  https://github.com/GoogleCloudPlatform/gcsfuse/commits/f35ec5c42e267ea0f90b606d7db4d2355ee3aa0d and https://github.com/GoogleCloudPlatform/gcsfuse/commits/184345e38e1f07668421a2949420f2e47b312ef2.

### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - Did sanity testing.
2. Unit tests - Ran and passed locally. Updated existing tests as needed.
3. Integration tests - Ran through presubmit.

PR #1692, opened 2024-02-12
Switch type-cache-max-entries to type-cache-max-size-mb

gargnitingoogle:
### Description
Switches type-cache-max-entries to type-cache-max-size-mb

More details of the change.
* Replace metadata-cache:type-cache-max-entries metadata-cache:type-cache-max-size-mb in config-file,
  and corresponding unit tests.
* Updates tests for type-cache-configs in dirInode
* Adds documentation for type-cache-max-size-mb in semantics.md

Additional size required for `key string` in type-cache entry struct.
```
For 1,000,0001 entries in type-cache,
RSS before change (with read_cache_release branch) with use of metadata-cache:type-cache-max-entries: 260.87MB
RSS after change (with gargnitin/switch-type-cache-to-size-mb branch) with use of metadata-cache:type-cache-max-size-mb: 276.63MB
Difference: 15.76MB ~= 16.53 bytes/entry.
```

This resurrects https://github.com/GoogleCloudPlatform/gcsfuse/pull/1687, https://github.com/GoogleCloudPlatform/gcsfuse/commits/f35ec5c42e267ea0f90b606d7db4d2355ee3aa0d and https://github.com/GoogleCloudPlatform/gcsfuse/commits/184345e38e1f07668421a2949420f2e47b312ef2.

NA

### Testing details
1. Manual - Did sanity testing.
2. Unit tests - Ran and passed locally. Updated existing tests as needed.
3. Integration tests - Ran through presubmit.


internal/cache/metadata/type_cache.go
```diff
@@ -100,20 +113,20 @@ type typeCache struct {
...
-// When insertion of next entry would cause #entries of cache > maxEntries,
+// When insertion of next entry would cause size of cache > sizeInMB,
 // older entries are evicted according to the LRU-policy.
-// If either of TTL or maxEntries is zero, nothing is ever cached.
-func NewTypeCache(maxEntries int, ttl time.Duration) TypeCache {
-	if ttl > 0 && maxEntries != 0 {
-		var lruMaxEntries uint64 = math.MaxUint64 // default for when maxEntries = -1
-		if maxEntries > 0 {
-			lruMaxEntries = uint64(maxEntries)
+// If either of TTL or sizeInMB is zero, nothing is ever cached.
```
sethiay, line 120: I would prefer using sizeMB as variable name because the flag is max-size-mb and max-size-in-mb.

internal/cache/metadata/type_cache_test.go
```diff
@@ -167,9 +168,6 @@ func (t *TypeCacheTest) TestGetAfterMaxEntriesExpiration() {
 
 	// The first inserted entry should have been evicted by all the later insertions.
 	ExpectEq(UnknownType, t.cache.Get(beforeExpiration, fmt.Sprint(0)))
-
-	// The second entry should not have been evicted
-	ExpectEq(RegularFileType, t.cache.Get(beforeExpiration, fmt.Sprint(1)))
```
sethiay, line 172: why did we remove this ?

internal/config/mount_config.go
```diff
@@ -35,9 +35,10 @@ const (
 	// The maximum multiple of seconds representable by time.Duration.
 	MaxSupportedTtlInSeconds int64 = int64(math.MaxInt64 / int64(time.Second))
 
-	// DefaultTypeCacheMaxEntries is the default vlaue of type-cache-max-entries.
-	// The value is set high at 1,048,476 (2^20) as it is not supposed to be set by users.
-	DefaultTypeCacheMaxEntries int = 2 << 20
+	// DefaultTypeCacheMaxSizeInMb is the default vlaue of type-cache max-size for every directory in MiBs.
```
sethiay, line 38: i think it's better to remove "In" to avoid confusion with the user flag.

internal/config/mount_config.go
```diff
@@ -35,9 +35,10 @@ const (
...
+	// The value is set at the size needed for 1,048,476 (2^20) type-cache entries,
```
sethiay, line 39: please update the default.

internal/config/mount_config.go
```diff
@@ -65,10 +66,11 @@ type MetadataCacheConfig struct {
...
 	TtlInSeconds int64 `yaml:"ttl-secs,omitempty"`
-	// TypeCacheMaxEntries is the upper limit
-	// on the maximum number of type-cache entries,
-	// which are maintained at per-directory level.
-	TypeCacheMaxEntries int `yaml:"type-cache-max-entries,omitempty"`
+	// TypeCacheMaxSizeMb is the upper limit
+	// on the maximum size of type-cache maps,
+	// which are currently
+	// maintained at per-directory level.
+	TypeCacheMaxSizeMb int `yaml:"type-cache-max-size-mb,omitempty"`
```
sethiay, line 73: would prefer using "B" instead of "b" in "Mb". it will also become consistent to MaxSizeInMB in FileCache

internal/fs/inode/dir.go
```diff
@@ -210,7 +210,7 @@ func NewDirInode(
 	bucket *gcsx.SyncerBucket,
 	mtimeClock timeutil.Clock,
 	cacheClock timeutil.Clock,
-	typeCacheMaxEntries int) (d DirInode) {
+	typeCacheSizeInMbPerDirectory int) (d DirInode) {
```
sethiay, line 213: we can remove "PerDirectory" from the name.

sethiay (commented):
Since we are adding additional field in value of cacheEntry, can we please perform a test and add in the description about RSS and heap with and without these changes for 1 million files ?

internal/cache/metadata/type_cache.go
```diff
@@ -64,13 +65,25 @@ type TypeCache interface {
...
+	// accurate size of the type-cache entry on heap.
+	key string
 }
 
-// Size returns the size of cacheEntry.
-// It is currently set to dummy value 1 to avoid
-// the unnecessary  actual size calculation.
+// Size returns the size of cacheEntry, which is
+// 80-bytes (24 for expriy, 8 for inodeType, 3*16 for 3 copies of
+// string structures containing the key ) + length of the key itself,
```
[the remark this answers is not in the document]
gargnitingoogle: One copy here, one in map (as key) and one in lru's linkedlist (https://github.com/GoogleCloudPlatform/gcsfuse/blob/acb8cc5e74a93908170c5b8d3247bbe247e508f9/internal/cache/lru/lru.go#L170) .

internal/cache/metadata/type_cache_test.go
gargnitingoogle: it was difficult to exact cut-off around which the the first entry would be evicted, but 2nd won't be. I have now standardized each entry's size by fixing the key string's size.

internal/fs/inode/dir.go
gargnitingoogle: removed

internal/config/mount_config.go
gargnitingoogle: done

internal/config/mount_config.go
gargnitingoogle: done

internal/config/mount_config.go
gargnitingoogle: done

internal/cache/metadata/type_cache.go
gargnitingoogle: done

internal/cache/metadata/type_cache.go
```diff
@@ -64,13 +65,25 @@ type TypeCache interface {
 type cacheEntry struct {
 	expiry    time.Time
 	inodeType Type
+	// Copy of key string (internally only
+	// points to a copy of the original
```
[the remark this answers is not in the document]
gargnitingoogle: fixed

docs/semantics.md
```diff
@@ -96,8 +96,8 @@ The negative result for ```foo/``` will be cached, but that only helps with the
 To alleviate this, Cloud Storage FUSE supports a "type cache" on directory inodes. When type cache is enabled, each directory inode will maintain a mapping from the name of its children to whether those children are known to be files or directories or both. When a child is looked up, if the parent's cache says that the child is a file but not a directory, only one Cloud Storage object will need to be stated. Similarly if the child is a directory but not a file.
 
 The behavior of type cache is controlled by the following flags/config parameters:
-
-**Type-cache TTL**: It controls the duration for which Cloud Storage FUSE caches an inode's type attribute. It can be set in one of the following two ways.
+1. **Type-cache capacity**: This is configurable at per-directory level by setting `metadata-cache: type-cache-max-size-mb` in config-file. This is the maximum size of type-cache per-directory in MiBs. By default, this is set at 80, which roughly equates to about a million entries (based on 80 bytes per [type-cache entry](https://github.com/GoogleCloudPlatform/gcsfuse/blob/acdb559a7b18e6c863db35b54c463fedd3fd1032/internal/cache/metadata/type_cache.go#L86)).
```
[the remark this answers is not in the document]
gargnitingoogle: done

gargnitingoogle:
> Since we are adding additional field in value of cacheEntry, can we please perform a test and add in the description about RSS and heap with and without these changes for 1 million files ?

working on this..

docs/semantics.md
```diff
@@ -97,9 +97,11 @@ To alleviate this, Cloud Storage FUSE supports a "type cache" on directory inode
...
-* ```--type-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This will be deprecated in a future version and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--type-cache-ttl``` is ignored.
+1. **Type-cache size**: This is configurable at per-directory level by setting `metadata-cache: type-cache-max-size-mb` in config-file. This is the maximum size of type-cache per-directory in MiBs. By default, this is set at 4, which roughly equates to about 42k entries.
+
+1. **Type-cache TTL**: It controls the duration for which Cloud Storage FUSE caches an inode's type attribute. It can be set in one of the following two ways.
+   - ```metadata-cache: ttl-secs``` in the config-file. This is set as an integer, which sets the TTL in seconds. If this is -1, TTL is taken as infinite i.e. no-TTL based expirations of entries. If this is 0, that disables the type-cache. If this is <-1, then an error is thrown on mount.
```
sethiay, line 103: nit: "<-1" should "< -1"

gargnitingoogle: fixed.

gargnitingoogle:
@sethiay this is waiting with all comments resolved. PTAL.

internal/cache/metadata/type_cache.go
```diff
@@ -64,13 +65,33 @@ type TypeCache interface {
...
+// for each entry (including negative ones).
+// This size is specific
+// to 64-bit linux machines, and might differ on other platforms.
+func (ce cacheEntry) Size() (size uint64) {
+	// First calculate size on heap.
+	// Additional 2*util.UnsafeSizeOf(&e.key) is to account for the 2 other copies of string
+	// struct stored in the cache map and in the cache linked-list.
+	size = uint64(util.UnsafeSizeOf(&ce) + 2*util.UnsafeSizeOf(&ce.key) + len(ce.key))
+
+	// Convert heap-size to rss (resident set size).
```
sethiay, line 91: would suggest you to create constant for heap to RSS conversion factor and use it everywhere.

internal/config/mount_config.go
```diff
@@ -35,9 +35,16 @@ const (
...
 
-	// DefaultTypeCacheMaxEntries is the default vlaue of type-cache-max-entries.
-	// The value is set high at 1,048,476 (2^20) as it is not supposed to be set by users.
-	DefaultTypeCacheMaxEntries int = 2 << 20
+	// DefaultTypeCacheMaxSizeMB is the default vlaue of type-cache max-size for every directory in MiBs.
+	// The value is set at the size needed for about 21k type-cache entries,
+	// each of which is about 200 bytes in size.
+	DefaultTypeCacheMaxSizeMB int = 4
+
+	// AverageTypeCacheEntrySize is for internal testing
```
sethiay, line 43: "is for internal testing" -> "internal size calculations"

gargnitingoogle:
> > Since we are adding additional field in value of cacheEntry, can we please perform a test and add in the description about RSS and heap with and without these changes for 1 million files ?
> 
> For 1,000,0001 entries in type-cache, RSS before change (with read_cache_release branch) with use of `metadata-cache:type-cache-max-entries`: 260.87MB RSS after change (with gargnitin/switch-type-cache-to-size-mb branch) with use of `metadata-cache:type-cache-max-size-mb`: 276.63MB Difference: 15.76MB ~= 16.53 bytes/entry.

Also added this to the description.

internal/cache/metadata/type_cache.go
gargnitingoogle: done.

internal/config/mount_config.go
gargnitingoogle: done.

gargnitingoogle:
looking into the test failures..

gargnitingoogle:
> looking into the test failures..

passing after re-run.. only integration tests still running.

PR #1701, opened 2024-02-12
Update semantics about disk-leak on write-failure

gargnitingoogle:
### Description
Update semantics about disk-leak on write-failure

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -8,11 +8,11 @@ Files that have not been modified are read portion by portion on demand. Cloud S
 
 **Writes**
 
-For modifications to existing objects, Cloud Storage FUSE downloads the entire
+For modifications to existing file objects, Cloud Storage FUSE downloads the entire
 backing object's contents from Cloud Storage. The contents are stored in a local
-temporary file whose location is controlled by the flag ```--temp-dir```. Later,
+temporary file (temp-file for short) whose location is controlled by the flag ```--temp-dir```. Later,
 when the file is closed or fsync'd, Cloud Storage FUSE writes the contents of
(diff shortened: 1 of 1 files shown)


docs/semantics.md
```diff
@@ -25,13 +25,21 @@ until they are written out to Cloud Storage, you
...
+    is written to your Cloud Storage bucket, with the existing empty file now
+    reflecting the accurate file size and content. Starting with version 1.2,
+    the default behavior is to not create this zero-byte file, which increases
+    write performance. If needed, it can be re-enabled by setting the
+    `create-empty-file: true` configuration in the config file.
+-   If the application never sends fsync for a file, it will leave behind its
+    temp-file (in temp-dir), which will not be cleared until the user unmounts
+    the bucket. As an example, if you are writing a large file, and temp-dir
+    does not have enough free space available, then you will get 'out of space'
+    error. Then the temp-file will not be deleted until you do an fsync for that
```
gargnitingoogle, line 41: @sethiay @mustvicky since we wrote this together, I had second thoughts about this line. the `the temp-file will not be deleted` seems like a break in flow. IMHO this should read. `Then the output of space error will not go away until ...` . 

PTAL

mustvicky: What you have currently LGTM. I am fine with the change of replacing the last line with `Then the output of space error will not go away until` but since the primary subject of this whole point is temp-file, it is better to keep it this way IMO.

mustvicky (approved):

sethiay (approved):

PR #1704, opened 2024-02-13
[ReadCacheFunctionalTest] Test Concurrent Reads results in non sequential read behavior if download for range read flag is false

ashmeenkaur:
### Description
Scenario:
With cacheFileForRangeRead config false
cache size: 50 MiB:
R1 50 MiB file (parallel)
R2 50MiB file (parallel)

Expected: 
R1 cache evicted, follow up read calls (from R1 operation) will not cache R1 50MiB file as the reads will be considered non sequential. In the end R2 file should be cached.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
+++ b/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
@@ -62,7 +62,6 @@ func readFileAsync(t *testing.T, wg *sync.WaitGroup, testFileName string, expect
 ////////////////////////////////////////////////////////////////////////
 
 func (s *cacheFileForRangeReadFalseTest) TestRangeReadsWithCacheMiss(t *testing.T) {
-	t.SkipNow()
 	testFileName := setupFileInTestDir(s.ctx, s.storageClient, testDirName, fileSizeForRangeRead, t)
 
 	// Do a random read on file and validate from gcs.
--- a/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
+++ b/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
@@ -17,13 +17,14 @@ package read_cache
 import (
 	"context"
 	"log"
+	"sync"
 	"testing"
 
-	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/log_parser/json_parser/read_logs"
-
 	"cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/log_parser/json_parser/read_logs"
 	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/setup"
 	"github.com/googlecloudplatform/gcsfuse/tools/integration_tests/util/test_setup"
+	"github.com/jacobsa/ogletest"
 )
 
 ////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
```diff
@@ -44,11 +45,24 @@ func (s *cacheFileForRangeReadFalseTest) Teardown(t *testing.T) {
...
+	}()
+	return
+}
+
 ////////////////////////////////////////////////////////////////////////
 // Test scenarios
 ////////////////////////////////////////////////////////////////////////
 
 func (s *cacheFileForRangeReadFalseTest) TestRangeReadsWithCacheMiss(t *testing.T) {
+	t.SkipNow()
```
Tulsishah, line 65: Why it is skipped?

tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
```diff
@@ -62,6 +76,38 @@ func (s *cacheFileForRangeReadFalseTest) TestRangeReadsWithCacheMiss(t *testing.
...
+	// the expected outcome objects and file names at positions 0 and 1.
+	if expectedOutcome[0].ObjectName != structuredReadLogs[0].ObjectName {
+		expectedOutcome[0], expectedOutcome[1] = expectedOutcome[1], expectedOutcome[0]
+		testFileNames[0], testFileNames[1] = testFileNames[1], testFileNames[0]
+	}
+	validate(expectedOutcome[0], structuredReadLogs[0], true, false, randomReadChunkCount, t)
+	validate(expectedOutcome[1], structuredReadLogs[1], true, false, randomReadChunkCount, t)
+	for i := 1; i < randomReadChunkCount; i++ {
+		ogletest.ExpectEq(false, structuredReadLogs[0].Chunks[i].IsSequential)
+		ogletest.ExpectEq(false, structuredReadLogs[0].Chunks[i].CacheHit)
```
Tulsishah, line 104: Cache hit false is expected right, as we are reading file on first time?

ashmeenkaur: Cache hit for follow up chunks is usually cache hit (even if we read the file first time). Here, cache hit false confirms that the file is not being cached.

Tulsishah (dismissed):
Overall it looks good but just have concern about flakiness where you are checking cache miss and isSeq false for deleted file.

I think we should run the test for 9-10 times on machine and check if it flaky or not(If possible may be on arm64). Feel free to write your suggestion.

tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
ashmeenkaur: Thanks for pointing out, updated!

ashmeenkaur:
> Overall it looks good but just have concern about flakiness where you are checking cache miss and isSeq false for deleted file.
> 
> I think we should run the test for 9-10 times on machine and check if it flaky or not(If possible may be on arm64). Feel free to write your suggestion.

Ran the test 5 times on ARM machine and 2-3 times on AMD machine. It passed successfully each time . I think we are good to merge this! :)

Tulsishah (approved):

PR #1706, opened 2024-02-13
Removing experimental-local-file-cache flag

Tulsishah:
### Description
Removing experimental-local-file-cache flag, this flag value will be default false and user won't be able to pass this flag after this.

NA

### Testing details
1. Manual - `help` command does not show the flag and while using the flag it was returning error.
2. Unit tests - Automated

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -240,11 +240,6 @@ func newApp() (app *cli.App) {
 				Usage: "Param for exponential backoff algorithm, which is used to increase waiting time b/w two consecutive retries.",
 			},
 
-			cli.BoolFlag{
-				Name:  "experimental-local-file-cache",
-				Usage: "Experimental: Cache GCS files on local disk for reads.",
-			},
-
 			cli.StringFlag{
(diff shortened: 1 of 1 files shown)


flags.go
```diff
@@ -503,7 +498,8 @@ func populateFlags(c *cli.Context) (flags *flagStorage, err error) {
 		HttpClientTimeout:          c.Duration("http-client-timeout"),
 		MaxRetryDuration:           c.Duration("max-retry-duration"),
 		RetryMultiplier:            c.Float64("retry-multiplier"),
-		LocalFileCache:             c.Bool("experimental-local-file-cache"),
+		// This flag is deprecated.
```
sethiay, line 501: "This flag is deprecated and we have plans to remove the implementation related to this flag in next release.

sethiay (commented):
Please add the manual testing you did as part of this PR in the description.

sethiay (approved):

PR #1707, opened 2024-02-13
Fix cache dir permissions

sethiay:
### Description
Problem
1. GCSFuse supports two more permissions related flags that are file-mode and dir-mode which decides the permissions of files and directories inside the gcsfuse mount. Note if file-mode is set 777 but allow-other is not passed then other users except owner will still not be able to access the files via GCSFuse mount. So allow-other tells whether to even respect file-mode and dir-mode for users other than mounting user or not.
2. Mounting gcsfuse with --allow-other flag and setting file-mode 700 is a completely valid thing to do from user's perspective. If user mounts gcsfuse with this config then other users on machine will not be able to access files via GCSFuse mount but will be able to access cache directory directly which is wrong. 

Solution
Always set 700 and 600 permissions for directory and file respectively in cache directory irrespective of allow-other flag. 
With this way, the permissions will be controlled solely via GCSFuse mount and we will not face the above problems described above. 
Also, it doesn't mean that users other than mounting user will not be able to get benefit of cache via GCSFuse mount if allow-other is passed because on read it is gcsfuse daemon that access files and directories in cache directory and not the user accessing GCSFuse mount.

NA

### Testing details
1. Manual 

Without allow-other
 - Mount gcsfuse with uid X at mount point P and cache dir C
    - Check perm of files and dirs in C - 600 & 700 with owner X.
    - Access P and C with uid X - Works.
    - Access P and C with uid Y - both don't work.
 - Mount gcsfuse with uid X at mount point P and cache dir C with file-mode 777 and dir-mode 777
    - Check perm of files and dirs in C - 600 & 700 with owner X.
    - Access P and C with uid X - Works.
    - Access P and C with uid Y - both don't work.

With allow-other
 - Mount gcsfuse with uid X at mount point P and cache dir C
    - Check perm of files and dirs in C - 600 & 700 with owner X.
    - Access P and C with uid X - Works.
    - Access P and C with uid Y - works for P and not for C.
 - Mount gcsfuse with uid X at mount point P and cache dir C with file-mode 777 and dir-mode 777
    - Check perm of files and dirs in C - 600 & 700 with owner X.
    - Access P and C with uid X - Works.
    - Access P and C with uid Y - works for P and not for C.
  - Mount gcsfuse with uid X at mount point P and cache dir C with file-mode 700 and dir-mode 700
    - Check perm of files and dirs in C - 600 & 700 with owner X.
    - Access P and C with uid X - Works.
    - Access P and C with uid Y - don't work.
  - Mount gcsfuse with uid X at mount point P and cache dir C with --file-mode 700 and --dir-mode 700 and --uid Y
    - Check perm of files and dirs in C - 600 & 700 with owner X.
    - Check perm of files in P - owner is Y.
    - Access P and C with uid X - Works.
    - Access P and C with uid Y - don't work.
  - Mount gcsfuse with uid X at mount point P and cache dir C with --file-mode 744 and --dir-mode 744 and --uid Y
    - Check perm of files and dirs in C - 600 & 700 with owner X.
    - Check perm of files in P - owner is Y.
    - Access P and C with uid X - Works.
    - Access P and C with uid Y - works with P and don't work with C.
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/internal/cache/util/util.go
+++ b/internal/cache/util/util.go
@@ -39,15 +39,11 @@ const (
 )
 
 const (
-	FileDirPerm            = os.FileMode(0755) | os.ModeDir
-	MiB                    = 1024 * 1024
-	KiB                    = 1024
-	DefaultFilePerm        = os.FileMode(0600)
-	FilePermWithAllowOther = os.FileMode(0644)
-
-	DefaultDirPerm        = os.FileMode(0700)
-	DirPermWithAllowOther = os.FileMode(0755)
-	FileCache             = "gcsfuse-file-cache"
+	MiB             = 1024 * 1024
+	KiB             = 1024
+	DefaultFilePerm = os.FileMode(0600)
+	DefaultDirPerm  = os.FileMode(0700)
+	FileCache       = "gcsfuse-file-cache"
 )
 
 // CreateFile creates file with given file spec i.e. permissions and returns
@@ -105,8 +101,9 @@ func IsCacheHandleInvalid(readErr error) bool {
 		strings.Contains(readErr.Error(), ErrInReadingFileHandleMsg)
 }
 
-// Creates directory at given path with provided permissions in case not already present,
-// returns error in case unable to create directory or directory is not writable.
+// CreateCacheDirectoryIfNotPresentAt Creates directory at given path with
+// provided permissions in case not already present, returns error in case
+// unable to create directory or directory is not writable.
 func CreateCacheDirectoryIfNotPresentAt(dirPath string, dirPerm os.FileMode) error {
 	_, statErr := os.Stat(dirPath)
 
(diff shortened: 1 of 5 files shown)


ankitaluthra1:
first and third scenario are same as mentioned in description ?

ankitaluthra1:
Mount gcsfuse with uid X at mount point P and cache dir C
Check perm of files and dirs in C - 600 & 700 with owner X.
Access P and C with uid X - Works.
Access P and C with uid Y - works for X and not for Y. - Did you mean P and C instead of X and Y

sethiay:
> first and third, second and fourth manual test scenario are same as mentioned in description ?



> first and third, second and fourth manual test scenario are same as mentioned in description ?

Corrected the identation

ankitaluthra1 (approved):

raj-prince (approved):

PR #1715, opened 2024-02-15
Fix flaky read cache functional test

ashmeenkaur:
### Description

1. Follow up to [#1702](https://github.com/GoogleCloudPlatform/gcsfuse/pull/1702), TestRangeReadsBeyondReadChunkSizeWithoutChunkDownloaded was still flaky after increasing file size to 100MiB. Increasing the file size to 500MiB to ensure cache is miss when reading last chunk of the file immediately after reading the first chunk.
2. TestConcurrentReads_ReadIsTreatedNonSequentialAfterFileIsRemovedFromCache was flaky because we were reading files in parallel and sometimes, 2nd chunk is still cache hit (when read from other file doesn't execute immediately). Removed the logic to check for all chunks 2nd chunk onward, and instead just validating that last chunk is cache miss and is sequential false.
3. Changed read block size to 128KiB to ensure that tests pass on rhel 7 and centos 7 machines (as these machines were not picking read block size of 1MiB)

NA

### Testing details
1. Manual - manually ran on CD pipeline and centos-7 VM
2. Unit tests - NA
3. Integration tests - KOKORO


tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
```diff
@@ -100,11 +100,10 @@ func (s *cacheFileForRangeReadFalseTest) TestConcurrentReads_ReadIsTreatedNonSeq
 	}
 	validate(expectedOutcome[0], structuredReadLogs[0], true, false, randomReadChunkCount, t)
 	validate(expectedOutcome[1], structuredReadLogs[1], true, false, randomReadChunkCount, t)
-	for i := 1; i < randomReadChunkCount; i++ {
-		ogletest.ExpectEq(false, structuredReadLogs[0].Chunks[i].IsSequential)
```
Tulsishah, line 104: As discussed keep for loop here for checking second file cache hit.

Tulsishah (approved):

tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
```diff
@@ -100,11 +100,13 @@ func (s *cacheFileForRangeReadFalseTest) TestConcurrentReads_ReadIsTreatedNonSeq
...
-		ogletest.ExpectEq(false, structuredReadLogs[0].Chunks[i].CacheHit)
-		ogletest.ExpectEq(true, structuredReadLogs[1].Chunks[i].IsSequential)
-	}
+	// validate last chunk was considered non-sequential and cache hit false for first read.
```
raj-prince, line 103: nit: Upper case

raj-prince (approved):

PR #1716, opened 2024-02-15
[PR-1] Add struct for extended object attribute

ashmeenkaur:
### Description
1. This PR adds struct for ExtendedObjectAttributes which contains the missing attributes of Object which are not present in MinObject.
2. This PR also adds following converters to storageutil package:
     1. ConvertObjToExtendedAttributes
     2. ConvertMinObjectAndExtendedAttributesToObject

Internal bug b/325024088

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -111,6 +111,11 @@ func NewFileInode(
 	mtimeClock timeutil.Clock,
 	localFile bool) (f *FileInode) {
 	// Set up the basic struct.
+	var minObj gcs.MinObject
+	minObjPtr := storageutil.ConvertObjToMinObject(o)
+	if minObjPtr != nil {
+		minObj = *minObjPtr
+	}
 	f = &FileInode{
 		bucket:         bucket,
 		mtimeClock:     mtimeClock,
@@ -119,12 +124,9 @@ func NewFileInode(
 		attrs:          attrs,
 		localFileCache: localFileCache,
 		contentCache:   contentCache,
+		src:            minObj,
 		local:          localFile,
 		unlinked:       false,
-	}
-	minObjPtr := storageutil.ConvertObjToMinObject(o)
-	if minObjPtr != nil {
-		f.src = *minObjPtr
 	}
 
 	f.lc.Init(id)
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -119,9 +119,12 @@ func NewFileInode(
 		attrs:          attrs,
 		localFileCache: localFileCache,
 		contentCache:   contentCache,
-		src:            storageutil.ConvertObjToMinObject(o),
 		local:          localFile,
 		unlinked:       false,
+	}
+	minObjPtr := storageutil.ConvertObjToMinObject(o)
+	if minObjPtr != nil {
+		f.src = *minObjPtr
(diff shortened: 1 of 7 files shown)


internal/storage/gcs/object.go
```diff
@@ -90,6 +90,24 @@ type MinObject struct {
...
 }
 
+// ExtendedObjectAttributes contains the missing attributes of Object which are not present in MinObject.
+type ExtendedObjectAttributes struct {
+	ContentType        string
+	ContentLanguage    string
+	CacheControl       string
+	Owner              string
+	MD5                *[md5.Size]byte // Missing for composite objects
+	CRC32C             *uint32         //Missing for CMEK buckets
```
raj-prince, line 100: nit: space b/w `//` and Missing

internal/storage/storageutil/object_attrs_test.go
```diff
@@ -255,3 +255,185 @@ func (t objectAttrsTest) Test_ConvertObjToMinObject_WithValidObject() {
...
+
+	object := ConvertMinObjectToObject(minObject)
+
+	ExpectEq(nil, object)
+}
+
+func (t objectAttrsTest) Test_ConvertMinObjectToObject_WithNonNilMinObject() {
+	var attrMd5 *[16]byte
+	var crc32C *uint32 = nil
+
```
raj-prince, line 404: nit: we can remove empty line.

internal/fs/inode/file.go
```diff
@@ -119,10 +119,13 @@ func NewFileInode(
 		attrs:          attrs,
 		localFileCache: localFileCache,
 		contentCache:   contentCache,
-		src:            storageutil.ConvertObjToMinObject(o),
 		local:          localFile,
 		unlinked:       false,
 	}
+	minObjPtr := storageutil.ConvertObjToMinObject(o)
```
sethiay, line 125: let's do this conversion before creating FileInode

internal/fs/inode/file.go
```diff
@@ -119,10 +119,13 @@ func NewFileInode(
...
+	if minObjPtr != nil {
+		f.src = *minObjPtr
```
sethiay, line 127: given the old code always expects src to be non empty, so we should follow that and panic if minObjPtr is nil, otherwise this can have unexpected results later if someone makes changes.

ashmeenkaur: As discussed offline, when nothing will be assigned to f.src, it will have empty gcs.MinObject (zero value of struct). This is in sync with original behavior of ConvertObjToMinObject as it returned empty gcs.MinObject when o is nil.
But we will still make this change for readability.

internal/fs/inode/file.go
```diff
@@ -517,7 +522,10 @@ func (f *FileInode) SetMtime(
 
 	o, err := f.bucket.UpdateObject(ctx, req)
 	if err == nil {
-		f.src = storageutil.ConvertObjToMinObject(o)
+		minObjPtr := storageutil.ConvertObjToMinObject(o)
```
sethiay, line 525: you will need to use `var minObj gcs.MinObject` here similar to above, otherwise f.src may not be updated if minObjPtr is nil

PR #1718, opened 2024-02-15
Download file using storage client

Tulsishah:
### Description
We were using `gsutil` or `gcloud` command to download file from gcs which is suddenly breaking with crcmod issue on arm64 machines. Changing that approach to download with storageclient instead of `gsutil` or `gcloud`.

CreateStorageClientWithTimeOut function is getting used in read_cache functional tests, we will remove [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/read_cache_release/tools/integration_tests/read_cache/helpers_test.go#L181) function whenever we will back merge.

NA

### Testing details
1. Manual - Tested manually on VM
2. Unit tests - NA
3. Integration tests - louhi and kokoro

### DIFF
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -121,13 +121,13 @@ func createStorageClient(ctx *context.Context, storageClient **storage.Client, t
 }
 
 // Downloads an object to a file.
-func DownloadFileWithStorageClient(gcsFile string, destFileName string,t *testing.T) error {
+func DownloadFileWithStorageClient(gcsFile string, destFileName string, t *testing.T) error {
 	var bucket string
 	setBucketAndObjectBasedOnTypeOfMount(&bucket, &gcsFile)
 
-	ctx:=context.Background()
-	var  storageClient *storage.Client
-	closeStorageClient := createStorageClient(&ctx, &storageClient,t)
+	ctx := context.Background()
+	var storageClient *storage.Client
+	closeStorageClient := createStorageClient(&ctx, &storageClient, t)
 	defer closeStorageClient()
 
 	f, err := os.Create(destFileName)
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -135,7 +135,11 @@ func DownloadObjectFromGCS(gcsFile string, destFileName string, t *testing.T) er
 	if err != nil {
 		return err
 	}
-	defer closeStorageClient()
+	defer func() {
+		if err := closeStorageClient(); err != nil {
+			t.Errorf("Error in closing client (deferred): %v", err)
+		}
+	}()
 
 	f := operations.CreateFile(destFileName, setup.FilePermission_0600, t)
 	defer operations.CloseFile(f)
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -103,7 +103,8 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
 	return nil
 }
 
-func CreateStorageClientWithTimeOut(ctx *context.Context, storageClient **storage.Client, time time.Duration, t *testing.T) func() {
+// Create storage client with a configurable timeout and retrun a function to cancel the storage client
+func CreateStorageClientWithTimeOut(ctx *context.Context, storageClient **storage.Client, time time.Duration) (func(), error) {
 	var err error
 	var cancel context.CancelFunc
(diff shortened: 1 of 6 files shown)


tools/integration_tests/util/client/storage_client.go
```diff
@@ -97,3 +101,53 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
 
 	return nil
 }
+
+func CreateStorageClientWithTimeOut(ctx *context.Context, storageClient **storage.Client, t *testing.T) func() {
+	var err error
+	var cancel context.CancelFunc
+	*ctx, cancel = context.WithTimeout(*ctx, time.Minute*15)
```
ashmeenkaur, line 108: Let's make this time configurable and pass a smaller value if the storage client is not persisted throughout the test suite.

Tulsishah: Let's keep it 5 minutes as for machines with less core it can take more time.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -97,3 +102,47 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
 
 	return nil
 }
+
+func CreateStorageClientWithTimeOut(ctx *context.Context, storageClient **storage.Client, time time.Duration, t *testing.T) func() {
```
raj-prince, line 106: Please write a doc about the responsibility of this function.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -97,3 +102,47 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
...
+	*ctx, cancel = context.WithTimeout(*ctx, time)
+	*storageClient, err = CreateStorageClient(*ctx)
+	if err != nil {
+		log.Fatalf("client.CreateStorageClient: %v", err)
+	}
+	// return func to close storage client and release resources.
+	return func() {
+		err := (*storageClient).Close()
+		if err != nil {
+			t.Log("Failed to close storage client")
```
raj-prince, line 118: We should get rid of t *testing.T from the utility. In that way, this would be a generic method.

We can do that, return error from this method as well as the return function should also return an error.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -97,3 +102,57 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
...
+		defer cancel()
+		if err != nil {
+			return fmt.Errorf("Failed to close storage client: %v", err)
+		}
+		return nil // No error during close
+	}, nil // No error during setup
+}
+
+// Downloads an object to a file.
+func DownloadObjectFromGCS(gcsFile string, destFileName string, t *testing.T) error {
```
raj-prince, line 128: We can remove the dependency of t *testing.T from here also.
It's okay, to directly call open here.

Later, we can use the cache/util.CreateFile here.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -97,3 +102,57 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
 
 	return nil
 }
+
+// Create storage client with a configurable timeout and retrun a function to cancel the storage client
```
ashmeenkaur, line 106:
```suggestion
// Create storage client with a configurable timeout and return a function to cancel the storage client
```

tools/integration_tests/util/client/storage_client.go
Tulsishah: As talked offline we will keep with `t *testing.T` as we are using this function in functional test.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -97,3 +102,48 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
...
+	return func() {
+		err := (*storageClient).Close()
+		if err != nil {
+			t.Log("Failed to close storage client")
+		}
+		defer cancel()
+	}
+}
+
+// Downloads an object to a file.
```
raj-prince, line 125: nit: DownloadObjectFromGCS downloads an object to a local file.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -97,3 +102,48 @@ func CreateObjectOnGCS(ctx context.Context, client *storage.Client, object, cont
 
 	return nil
 }
+
+// Create storage client with a configurable timeout and return a function to cancel the storage client
```
raj-prince, line 106: nit: CreateStorageClientWithTimeOut creates... the storage client.

PR #1722, opened 2024-02-16
Managed folder changes

Tulsishah:
### Description
1. Introduced new flag `enable-managed-folders-listing`
2. Added it in file system structure to pass it till readObjects.
3. Added additional parameter in functions to enableManagedFoldersListing.

**Note: If there are any objects remains in managed folders, then managed folders will be visible even after `enable-managed-folders-listing` flag will be false.**

NA

### Testing details
1. Manual 
      Did manual testing as mentioned [here](https://docs.google.com/document/d/1wxjc7HwILFFIbgY3ZQVXjSv0vSJhO63agXcHpTU0Naw/edit?pli=1&tab=t.0#bookmark=id.izdtl66028sl).

3. Unit tests - Added where required.
FakeGCSServer is not handling IncludeFoldersAsPrefixes check for ListObjects flow.
Hence, we are not writing tests for this flows.
Not adding unit tests related to enableManaedFolders, as currently we don't have any way to create managed folders in fake bucket/mock bucket.


flags.go
```diff
@@ -309,6 +309,12 @@ func newApp() (app *cli.App) {
 					"When this flag is specified, GCSFuse uses the GCS JSON API instead.",
 			},
 
+			cli.BoolTFlag{
+				Name: "enable-managed-folders",
```
raj-prince, line 313: can we rename this flag to enable-managed-folders-listing?

As `enable-managed-folders`  feels like we are controlling the managed folder behavior which is not true.

flags.go
```diff
@@ -309,6 +309,12 @@ func newApp() (app *cli.App) {
...
+				Usage: "GCSFuse automatically lists managed folders found within a bucket. " +
```
raj-prince, line 314: We can enhance the usage statement. We should also say about the `--implicit-dirs` flag something like - customer must pass `--implicit-dirs` flag to make this flag work for true value.

raj-prince (changes requested):
Please incorporate.

Tulsishah:
> Please incorporate.

Done. PTAL

internal/fs/fs.go
```diff
@@ -74,6 +74,12 @@ type ServerConfig struct {
 	// See docs/semantics.md for more info.
 	ImplicitDirectories bool
 
+	// By default, if a managed-folder exist in bucket, this state will list managed-folders in mounted directory provided
+	// --implicit-dirs flag is set to 'true'.
+	//
+	// Setting this bool to false disables the visibility of empty managed folders.
```
raj-prince, line 80: I think, it's good to document clearly.

This flag is specially added to handle the corner case in listing managed folders. There are two corner cases (a) empty managed folder (b) nested managed folder which doesn't contain any descendent as object.  This flag always works in conjunction with `ImplicitDirectories` flag.

(a) If only `ImplicitDirectories` is true, all managed folders are listed other than above two mentioned case.
(b) If  both `ImplicitDirectories` and `EnableManagedFoldersListing` are true, then all the managed folders are listed including the above mentioned corner case.
(c) If `ImplicitDirectories` is false then no managed folders are listed irrespective of `EnableManagedFoldersListing` flag.

internal/fs/inode/dir_test.go
```diff
@@ -86,7 +86,7 @@ func (p DirentSlice) Len() int           { return len(p) }
 func (p DirentSlice) Less(i, j int) bool { return p[i].Name < p[j].Name }
 func (p DirentSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
 
-func (t *DirTest) resetInode(implicitDirs bool, enableNonexistentTypeCache bool) {
+func (t *DirTest) resetInode(implicitDirs bool, enableNonexistentTypeCache bool, enableManagedFolderListing bool) {
```
raj-prince, line 89: nit: enableManagedFolderListing -> enableManagedFoldersListing

internal/fs/inode/dir_test.go
```diff
@@ -512,7 +513,7 @@ func (t *DirTest) LookUpChild_TypeCaching() {
 
 func (t *DirTest) LookUpChild_NonExistentTypeCache_ImplicitDirsDisabled() {
```
raj-prince, line 561: Can we also write a reasoning, why we haven't added unit test corresponding to added new flag in the dir_test.go?

internal/fs/inode/dir.go
```diff
@@ -222,6 +224,7 @@ func NewDirInode(
 		cacheClock:                 cacheClock,
 		id:                         id,
 		implicitDirs:               implicitDirs,
+    enableManagedFoldersListing: enableManagedFoldersListing,
```
raj-prince, line 227: seems formatting issue.

Tulsishah: This arise due to rebasing. Fixed

raj-prince (approved):

PR #1732, opened 2024-02-20
[PR-2] Changes to StatObject to return min object and extended object attributes

ashmeenkaur:
### Description
This PR changes StatObjet method to return MinObject and ExtendedObjectAttributes instead of complete GCS Object. This will enable us to store MinObject in stat cache.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -25,6 +25,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/internal/locker"
 	"github.com/googlecloudplatform/gcsfuse/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/internal/storage/storageutil"
 	"github.com/jacobsa/fuse/fuseops"
 	"github.com/jacobsa/fuse/fuseutil"
 	"github.com/jacobsa/syncutil"
@@ -297,7 +298,8 @@ func findExplicitInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name
 		Name: name.GcsObjectName(),
 	}
 
-	o, err := bucket.StatObject(ctx, req)
+	m, e, err := bucket.StatObject(ctx, req)
+	o := storageutil.ConvertMinObjectAndExtendedAttributesToObject(m, e)
 
 	// Suppress "not found" errors.
 	var gcsErr *gcs.NotFoundError
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -298,8 +298,8 @@ func findExplicitInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name
 		Name: name.GcsObjectName(),
 	}
 
-	m, e, err := bucket.StatObject(ctx, req)
-	o := storageutil.ConvertMinObjectAndExtendedAttributesToObject(m, e)
+	m, _, err := bucket.StatObject(ctx, req)
+	o := storageutil.ConvertMinObjectToObject(m)
 
 	// Suppress "not found" errors.
 	var gcsErr *gcs.NotFoundError
--- a/internal/gcsx/prefix_bucket.go
+++ b/internal/gcsx/prefix_bucket.go
@@ -138,13 +138,13 @@ func (b *prefixBucket) ComposeObjects(
 
 func (b *prefixBucket) StatObject(
 	ctx context.Context,
-	req *gcs.StatObjectRequest) (o *gcs.Object, err error) {
+	req *gcs.StatObjectRequest) (o *gcs.MinObject, e *gcs.ExtendedObjectAttributes, err error) {
 	// Modify the request and call through.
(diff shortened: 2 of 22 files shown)


internal/fs/inode/dir.go
```diff
@@ -297,7 +298,8 @@ func findExplicitInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name
 		Name: name.GcsObjectName(),
 	}
 
-	o, err := bucket.StatObject(ctx, req)
+	m, _, err := bucket.StatObject(ctx, req)
+	o := storageutil.ConvertMinObjectToObject(m)
```
sethiay, line 302: can we also make changes in core.go to use MinObject inside Core struct also and use min object here as well ?

internal/gcsx/prefix_bucket.go
```diff
@@ -138,13 +138,13 @@ func (b *prefixBucket) ComposeObjects(
 
 func (b *prefixBucket) StatObject(
 	ctx context.Context,
-	req *gcs.StatObjectRequest) (o *gcs.Object, err error) {
+	req *gcs.StatObjectRequest) (o *gcs.MinObject, e *gcs.ExtendedObjectAttributes, err error) {
```
sethiay, line 141: let's also use "m" here and everywhere where we are using "o" as variable name for minObject instead of "o" for consistency

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -204,8 +205,10 @@ func (b *fastStatBucket) StatObject(
 			return
 		}
 
-		// Otherwise, return the object.
-		o = entry
+		// Otherwise, return object and ExtendedObjectAttributes.
+		minObj := storageutil.ConvertObjToMinObject(entry)
+		m = &minObj
+		e = storageutil.ConvertObjToExtendedAttributes(entry)
```
sethiay, line 211: we can simply return nil for ExtendedObjectAttributes when cache is hit ?

internal/storage/fake/testing/bucket_tests.go
```diff
@@ -1262,12 +1263,12 @@ func (t *copyTest) DestinationDoesntExist() {
...
 
 	// And stattable.
-	statO, err := t.bucket.StatObject(
+	statMinObj, _, err := t.bucket.StatObject(
 		t.ctx,
 		&gcs.StatObjectRequest{Name: "bar"})
 
 	AssertEq(nil, err)
-	ExpectThat(statO, Pointee(DeepEquals(*dst)))
+	ExpectThat(statMinObj, Pointee(DeepEquals(dstMinObject)))
```
sethiay, line 1271: should we compare both minObject and extAttr ?

internal/storage/caching/fast_stat_bucket.go
ashmeenkaur: FYI: Handling ReturnExtendedObjectAttributes parameter in the stat object request was essential for this change to work as other unit tests started failing. So this PR now includes updates to `fast_stat_bucket.go` and `bucket_handle.go` to properly implement StatObject method.

internal/fs/inode/dir.go
ashmeenkaur: This will require a lot of changes. I'll check and raise a separate PR for this change.

internal/storage/bucket_handle.go
```diff
@@ -106,7 +106,12 @@ func (b *bucketHandle) DeleteObject(ctx context.Context, req *gcs.DeleteObjectRe
 
 }
 
-func (b *bucketHandle) StatObject(ctx context.Context, req *gcs.StatObjectRequest) (o *gcs.Object, err error) {
+func (b *bucketHandle) StatObject(ctx context.Context,
+	req *gcs.StatObjectRequest) (m *gcs.MinObject, e *gcs.ExtendedObjectAttributes, err error) {
+	// If ExtendedObjectAttributes are requested without fetching from gcs enabled, panic.
+	if !req.ForceFetchFromGcs && req.ReturnExtendedObjectAttributes {
+		panic("invalid StatObjectRequest: ForceFetchFromGcs: false and ReturnFull: true")
```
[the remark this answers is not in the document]
sethiay: Also, should we remove this check from here given we have added in fstat bucket ?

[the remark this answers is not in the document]
ashmeenkaur: Makes sense, done

internal/storage/bucket_handle_test.go
```diff
@@ -264,7 +264,7 @@ func (t *BucketHandleTest) TestStatObjectMethodWithValidObject() {
 func (t *BucketHandleTest) TestStatObjectMethodWithMissingObject() {
 	var notfound *gcs.NotFoundError
 
-	_, err := t.bucketHandle.StatObject(context.Background(),
+	_, _, err := t.bucketHandle.StatObject(context.Background(),
```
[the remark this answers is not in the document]
ashmeenkaur: Added the tests!

sethiay (approved):

PR #1734, opened 2024-02-20
Add cache-dir to experiment configuration of read cache tests

sethiay:
### Description
Add cache-dir to experiment configuration

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/continuous_test/gcp_ubuntu/periodic_experiments/experiments_configuration.json
+++ b/perfmetrics/scripts/continuous_test/gcp_ubuntu/periodic_experiments/experiments_configuration.json
@@ -7,7 +7,7 @@
       "end_date": "2024-03-31 05:30:00+00:00"
     },
     {
-      "config_name": "read_cache_cache_file_for_range_read_true",
+      "config_name": "read_cache_cache_for_range_read_true",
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/continuous_test/gcp_ubuntu/periodic_experiments/experiments_configuration.json
```diff
@@ -13,7 +13,8 @@
         "file-cache": {
           "max-size-mb": -1,
           "cache-file-for-range-read": true
-        }
+        },
+        "cache-dir": "/tmp/cache"
```
Tulsishah, line 17: With this change you are not creating new configuration, so although it will not throw an error and run with these configs only but on big query table it will not reflect this parameter. If you want to do so, you have to create new configuration(by giving different config_name)

sethiay: Done, changed the name.

Tulsishah (approved):
Seems good. Just thinking out loud, to easily distinguish this updated data from previous versions in our BigQuery table, should we consider using a descriptive name or timestamp within the new configuration?

Tulsishah (approved):

raj-prince (approved):

PR #1736, opened 2024-02-21
Run tests with static mounting only to reduce run time in e2e tests

Tulsishah:
### Description
Previously one test which recently got updated were taking 7s to complete, which now takes around 20s.
1. 20s*24(sub tests)*3(per mounting) = 1440s
2. 7s*24(sub tests)*3(per mounting) = 504s 

Made those tests to run only for static mounting and not for dynamic and only-dir.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/perfmetrics/scripts/run_e2e_tests.sh
+++ b/perfmetrics/scripts/run_e2e_tests.sh
@@ -18,7 +18,7 @@
 set -e
 sudo apt-get update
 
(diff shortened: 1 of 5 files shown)


ashmeenkaur (dismissed):

sethiay (approved):
.

PR #1742, opened 2024-02-27
Adding EnableManagedFolders flag in config file

Tulsishah:
### Description
Adding EnableManagedFolders flag in config file, by default it will be true and Cx can make it false with 
```
list:
   enable-managed-folders: true
```

NA

### Testing details
1. Manual - Tested all IAM permissions scenarios.
2. Unit tests - NA


internal/config/yaml_parser_test.go
```diff
@@ -33,6 +33,7 @@ func init() { RegisterTestSuite(&YamlParserTest{}) }
 func validateDefaultConfig(mountConfig *MountConfig) {
 	AssertNe(nil, mountConfig)
 	ExpectEq(false, mountConfig.CreateEmptyFile)
+	ExpectEq(true, mountConfig.ListConfig.EnableManagedFolders)
```
sethiay, line 36: please keep this check only once.

sethiay: my comment is not correct.

sethiay (approved):
Approving, however please wait for @raj-prince  approval before merging. Thanks.

raj-prince (approved):

PR #1763, opened 2024-03-04
Replacing panic with error message

raj-prince:
### Description
Problem: we should not use panic statement for the user-controlled behavior. Also, we were getting generic failure while mounting.

Before the change, stdout for the mount with no permission to create cache-dir:
```
daemonize.Run: readFromProcess: Decode: EOF
```

After the change, stdout for the mount with no permission to create cache-dir:
```
daemonize.Run: readFromProcess: sub-process: mountWithArgs: mountWithStorageHandle: fs.NewServer: create file system: while creating NewFileSystem(): createFileCacheHandler: while creating file cache directory: error in creating directory structure /prince/gcsfuse-file-cache: mkdir /prince: permission denied
```

NA

### Testing details
1. Manual - checked the mount failing behavior manually.
2. Unit tests - Decided to go without unit-test for now, since we don't have straight forward way to add mount fail unit test.
3. Integration tests -

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -160,7 +160,7 @@ func NewFileSystem(
 		var err error
 		fileCacheHandler, err = createFileCacheHandler(cfg)
 		if err != nil {
-			return nil, fmt.Errorf("while creating NewFileSystem(): %v", err)
+			return nil, err
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/fs/fs.go
```diff
@@ -157,7 +157,11 @@ func NewFileSystem(
 	// enabled only if cache-dir is not empty and file-cache:max-size-mb is non 0.
 	var fileCacheHandler *file.CacheHandler
 	if config.IsFileCacheEnabled(cfg.MountConfig) {
-		fileCacheHandler = createFileCacheHandler(cfg)
+		var err error
+		fileCacheHandler, err = createFileCacheHandler(cfg)
+		if err != nil {
+			return nil, fmt.Errorf("while creating NewFileSystem(): %v", err)
```
sethiay, line 163: I think it's fine to simply return the error instead of reformatting it because createFileCacheHandler is not used more than once.

internal/fs/fs.go
```diff
@@ -235,7 +239,7 @@ func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHand
 	// Panic in case cacheDir does not have required permissions
 	cacheDirErr := util.CreateCacheDirectoryIfNotPresentAt(cacheDir, dirPerm)
 	if cacheDirErr != nil {
-		panic(fmt.Sprintf("createFileCacheHandler: error while creating file cache directory: %v", cacheDirErr.Error()))
+		return nil, fmt.Errorf("createFileCacheHandler: while creating file cache directory: %v", cacheDirErr)
```
sethiay, line 242: use "%w" instead of "%v"

sethiay (approved):

ankitaluthra1 (commented):
change looks good, test is still pending , can approve after unit test ?

raj-prince:
> change looks good, test is still pending , can approve after unit test ?

Discussed offline, also have added in the description - the reason behind pushing this change without unit test.

PR #1775, opened 2024-03-07
Reduce time for integration tests

Tulsishah:
Today all the tests are running sequentially, because of that we have to wait almost 1 hour to check e2e test success and it is affecting the acceleration of the development cycle.

### Packages in integration tests
Currently there are 14 packages in our integration test directory.

1. explicit_dir
2. implicit_dir
3. list_large_dir
4. operations
5. read_large_files
6. rename_dir_limit
7. managed_folders
8. read_only
9. local_file
10. log_rotation
11. mounting
12. read_cache
13. gzip
14. write_large_files

Packages 1-6: Must run one after the other. Each individual test within these packages also needs to run sequentially. This is because they clean a shared directory (mnt dir).
Packages 7-8 (readonly and managed_folders) Cannot be run in parallel at all. They alter permissions that would break other tests.
Packages 9-14: Can run in parallel with each other (package-level parallelization). However, tests within each package must run sequentially since they work on the same directory.

We can parallelize packages 1 to 6 slowly, it will require effort.
## Approach 
Basic Idea: We're splitting the tests into three groups to make things faster.

1. Three buckets: We're using separate spaces to avoid tests bumping into each other.
2. Group 1 (Non-parallel): These tests must run one after the other. To save time, they'll work quietly in the background. These tests can parallelize with some additional effort and we will do it gradually.
3. Group 2 (Never parallelize): These tests can't be parallelize ever as they are changing bucket permissions.
4. Group 3 (Parallel): These tests can all work at the same time, each handling their own package.

We removed the `setup.RemoveBinFileCopiedForTesting()` function because it caused conflicts between tests. One test would delete a file that another test needed. We addressed this with a script, but running integration tests could [reintroduce the file](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L187). Therefore, removing the function permanently shouldn't cause any problems.

NA

### Testing details
1. Manual - kokoro
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/run_e2e_tests.sh
+++ b/perfmetrics/scripts/run_e2e_tests.sh
@@ -148,35 +148,38 @@ install_packages
 # Create Bucket for non parallel e2e tests
 # The bucket prefix for the random string
 bucket_prefix="gcsfuse-non-parallel-e2e-tests-group-1-"
-bucket_name_non_paraller_group_1=$(create_bucket $bucket_prefix)
-echo "Bucket name for non parallel tests group - 1: "$bucket_name_non_paraller_group_1
+bucket_name_non_parallel_group_1=$(create_bucket $bucket_prefix)
+echo "Bucket name for non parallel tests group - 1: "$bucket_name_non_parallel_group_1
 
 # Test setup
 # Create Bucket for non parallel e2e tests
 # The bucket prefix for the random string
 bucket_prefix="gcsfuse-non-parallel-e2e-tests-group-2-"
-bucket_name_non_paraller_group_2=$(create_bucket $bucket_prefix)
-echo "Bucket name for non parallel tests group - 2 : "$bucket_name_non_paraller_group_2
+bucket_name_non_parallel_group_2=$(create_bucket $bucket_prefix)
+echo "Bucket name for non parallel tests group - 2 : "$bucket_name_non_parallel_group_2
 
 # Create Bucket for parallel e2e tests
 # The bucket prefix for the random string
 bucket_prefix="gcsfuse-parallel-e2e-tests-"
-bucket_name_paraller=$(create_bucket $bucket_prefix)
-echo "Bucket name for parallel tests: "$bucket_name_paraller
+bucket_name_parallel=$(create_bucket $bucket_prefix)
+echo "Bucket name for parallel tests: "$bucket_name_parallel
 
 # Run tests
 set +e
 echo "Running parallel tests..."
 # Run parallel tests
-run_parallel_tests TEST_DIR_PARALLEL $bucket_name_paraller &
+run_parallel_tests TEST_DIR_PARALLEL $bucket_name_parallel &
 parallel_tests_pid=$!
+
 # Run non parallel tests
 echo "Running non parallel tests group-1..."
-run_non_parallel_tests TEST_DIR_NON_PARALLEL_GROUP_1 $bucket_name_non_paraller_group_1 &
+run_non_parallel_tests TEST_DIR_NON_PARALLEL_GROUP_1 $bucket_name_non_parallel_group_1 &
 non_parallel_tests_pid_group_1=$!
 echo "Running non parallel tests group-2..."
-run_non_parallel_tests TEST_DIR_NON_PARALLEL_GROUP_2 $bucket_name_non_paraller_group_2 &
+run_non_parallel_tests TEST_DIR_NON_PARALLEL_GROUP_2 $bucket_name_non_parallel_group_2 &
 non_parallel_tests_pid_group_2=$!
+
+# Wait for all tests to complete.
 wait $parallel_tests_pid
 parallel_tests_exit_code=$?
 wait $non_parallel_tests_pid_group_1
@@ -187,9 +190,9 @@ set -e
 
 # Cleanup
 # Delete bucket after testing.
-gcloud alpha storage rm --recursive gs://$bucket_name_paraller/
-gcloud alpha storage rm --recursive gs://$bucket_name_non_paraller_group_1/
-gcloud alpha storage rm --recursive gs://$bucket_name_non_paraller_group_2/
+gcloud alpha storage rm --recursive gs://$bucket_name_parallel/
+gcloud alpha storage rm --recursive gs://$bucket_name_non_parallel_group_1/
+gcloud alpha storage rm --recursive gs://$bucket_name_non_parallel_group_2/
 
 # Removing bin file after testing.
 if [ $RUN_E2E_TESTS_ON_PACKAGE != true ];


perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -53,28 +53,120 @@ sudo apt-get install -y gcc python3-dev python3-setuptools
...
-bucketPrefix="gcsfuse-integration-test-"
-# The length of the random string
-length=5
-# Generate the random string
-random_string=$(tr -dc 'a-z0-9' < /dev/urandom | head -c $length)
-BUCKET_NAME=$bucketPrefix$random_string
-echo 'bucket name = '$BUCKET_NAME
-# We are using gcloud alpha because gcloud storage is giving issues running on Kokoro
-gcloud alpha storage buckets create gs://$BUCKET_NAME --project=$PROJECT_ID --location=$BUCKET_LOCATION --uniform-bucket-level-access
+function create_bucket() {
```
raj-prince, line 56: Instead pass take bucket_prefix as an argument or $1 and create bucket based on that.

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -53,28 +53,120 @@ sudo apt-get install -y gcc python3-dev python3-setuptools
...
+    fi
+  done
+  return $test_fail_p
+}
 
+# Test setup
+# Create Bucket for non parallel e2e tests
+# The bucket prefix for the random string
+bucketPrefix="gcsfuse-non-parallel-e2e-tests-"
+create_bucket
```
raj-prince, line 109: Call the create_bucket with bucketPrefix as an argument.

raj-prince (commented):
Approach looks good to me, one minor comment.

Our next goal would be make all the non-parallel test package independent (other than two package which required bucket permission change).

sethiay (commented):
Before making changes, could we ensure this doesn't cause failures in GKE pipeline ?

Tulsishah:
> Before making changes, could we ensure this doesn't cause failures in GKE pipeline ?

These changes will not break GKE pipeline as we are not making any change for mountedDirectory flag. Changes are for kokoro e2e script only. Although I'll run that script which we are sharing with kokoro team before merging.

Tulsishah:
> Approach looks good to me, one minor comment.
> 
> Our next goal would be make all the non-parallel test package independent (other than two package which required bucket permission change).

Yes I updated in document.

Fixed comments. PTAL

raj-prince (dismissed):

raj-prince:
Please wait for the @sethiay's approval before merge.

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -53,28 +53,162 @@ sudo apt-get install -y gcc python3-dev python3-setuptools
...
+    if [ $exit_code_non_parallel_group_1 != 0 ]; then
+      test_fail_np_group_1=$exit_code_non_parallel_group_1
+      echo "test fail in non parallel in group 1: " $test_fail_np_group_1
+    fi
+  done
+  return $test_fail_np_group_1
+}
+
+# Non parallel execution of integration tests located within specified test directories for group 2.
+function run_non_parallel_tests_group_2() {
```
Tulsishah, line 85: Created separate function as calling same function with argument in background was overriding the value.

FYI @raj-prince @sethiay

ankitaluthra1: @Tulsishah is $test_fail_np_group_1 value getting overridden if we use same function ? if that's the case then can we append a prefix/suffix of bucket name in this variable name and keep the function same? will it help ?

Tulsishah: `test_dir_np_group_1` the array itself is overriding, so all the tests are not executing that's what I face. It's not about failure code, it's about execution itself.

Tulsishah: It is fixed. Thanks @ankitaluthra1 I was capturing arrays properly in functions that's why it was happening.

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -53,28 +53,152 @@ sudo apt-get install -y gcc python3-dev python3-setuptools
...
+
+  for test_dir_np in "${testArray[@]}"
+  do
+    test_path_non_parallel="./tools/integration_tests/$test_dir_np"
+    # Executing integration tests
+    GODEBUG=asyncpreemptoff=1 go test $test_path_non_parallel -p 1 --integrationTest -v --testbucket=$BUCKET_NAME_NON_PARALLEL --testInstalledPackage=$run_e2e_tests_on_package -timeout $INTEGRATION_TEST_TIMEOUT
+    exit_code_non_parallel=$?
+    if [ $exit_code_non_parallel != 0 ]; then
+      test_fail_np=$exit_code_non_parallel
+      echo "test fail in non parallel: " $test_fail_np
```
sethiay, line 81: should we print test_dir_np instead of test_fail_np?

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -53,28 +53,152 @@ sudo apt-get install -y gcc python3-dev python3-setuptools
...
+test_dir_non_parallel_group_1=(
+  "readonly"
+  "managed_folders"
+)
+
+# Test setup
+# Create Bucket for non parallel e2e tests
+# The bucket prefix for the random string
+bucketPrefix="gcsfuse-non-parallel-e2e-tests-group-2-"
+echo "Bucket name for non parallel tests group - 2 : "$BUCKET_NAME
```
sethiay, line 132: the create bucket is called after this echo statement

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -53,28 +53,152 @@ sudo apt-get install -y gcc python3-dev python3-setuptools
...
+  "log_rotation"
+  "mounting"
+  "read_cache"
+  "gzip"
+  "write_large_files"
+)
 
+# Run tests
+test_fail_p=0
+test_fail_np=0
```
sethiay, line 165: where is test_fail_np used ?

sethiay (commented):
Also, could you please confirm that the gcsfuse debug logs of different tests would not interfere with each ?

perfmetrics/scripts/run_e2e_tests.sh
Tulsishah: On line number 80

Tulsishah:
> Also, could you please confirm that the gcsfuse debug logs of different tests would not interfere with each ?

They should not as every package is building gcsfuse in different directory `/tmp/gcsfuse_readwrite_test_xxx/`, so logfile will be differently stored in these package and in case of failure we are sending logs to kokoro bucket.

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -34,48 +61,144 @@ function upgrade_gcloud_version() {
...
+  random_string=$(tr -dc 'a-z0-9' < /dev/urandom | head -c $length)
+  BUCKET_NAME=$bucketPrefix$random_string
+  # We are using gcloud alpha because gcloud storage is giving issues running on Kokoro
+  gcloud alpha storage buckets create gs://$BUCKET_NAME --project=$PROJECT_ID --location=$BUCKET_LOCATION --uniform-bucket-level-access
+  echo $BUCKET_NAME
+}
+
+# Non parallel execution of integration tests located within specified test directories.
+function run_non_parallel_tests() {
+  local test_dir_np=0
```
sethiay, line 94: correct this

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -34,48 +61,144 @@ function upgrade_gcloud_version() {
...
+  local -n testArray=$1
+  BUCKET_NAME_NON_PARALLEL=$2
```
sethiay, line 96: make all the local variables follow the convention lower_case

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -22,6 +22,33 @@ readonly INTEGRATION_TEST_TIMEOUT=40m
 readonly PROJECT_ID="gcs-fuse-test-ml"
 readonly BUCKET_LOCATION="us-west1"
 
+# Test directory arrays
+test_dir_parallel=(
```
sethiay, line 26: you can make this CAPS_LOCK

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -34,48 +61,144 @@ function upgrade_gcloud_version() {
...
+# Run tests
 set +e
-# Executing integration tests
-GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/... -p 1 --integrationTest -v --testbucket=$BUCKET_NAME --testInstalledPackage=$run_e2e_tests_on_package -timeout $INTEGRATION_TEST_TIMEOUT
-exit_code=$?
 
+echo "Running parallel tests..."
+# Run parallel tests
+run_parallel_tests test_dir_parallel $BUCKET_NAME_PARALLEL &
+my_process_p=$!
```
sethiay, line 173: parallel_tests_pid

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -34,48 +61,144 @@ function upgrade_gcloud_version() {
...
+# Run non parallel tests
+echo "Running non parallel tests group-1..."
+run_non_parallel_tests test_dir_non_parallel_group_1 $BUCKET_NAME_NON_PARALLEL_GROUP_1 &
+my_process_np_group_1=$!
+echo "Running non parallel tests group-2..."
+run_non_parallel_tests test_dir_non_parallel_group_2 $BUCKET_NAME_NON_PARALLEL_GROUP_2 &
+my_process_np_group_2=$!
+wait $my_process_p
+test_fail_p=$?
```
sethiay, line 182: parallel_tests_exit_code

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -34,48 +61,144 @@ function upgrade_gcloud_version() {
...
+    if [ $exit_code_parallel != 0 ]; then
+      test_fail_p=$exit_code_parallel
+      echo "test fail in parallel on package: " $test_dir_p
+    fi
+  done
+  return $test_fail_p
+}
+
 # true or false to run e2e tests on installedPackage
 run_e2e_tests_on_package=$1
```
sethiay, line 141: keep it at top and make it CAPS_LOCK

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -34,48 +62,142 @@ function upgrade_gcloud_version() {
...
-# install python3-setuptools tools.
-sudo apt-get install -y gcc python3-dev python3-setuptools
-# Downloading composite object requires integrity checking with CRC32c in gsutil.
-# it requires to install crcmod.
-sudo apt install -y python3-crcmod
+# Test setup
+# Create Bucket for non parallel e2e tests
+# The bucket prefix for the random string
+bucket_prefix="gcsfuse-non-parallel-e2e-tests-group-1-"
+bucket_name_non_paraller_group_1=$(create_bucket $bucket_prefix)
```
sethiay, line 151: nit: parallel

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -34,48 +62,142 @@ function upgrade_gcloud_version() {
...
+bucket_name_paraller=$(create_bucket $bucket_prefix)
+echo "Bucket name for parallel tests: "$bucket_name_paraller
 
+# Run tests
+set +e
+echo "Running parallel tests..."
+# Run parallel tests
+run_parallel_tests TEST_DIR_PARALLEL $bucket_name_paraller &
+parallel_tests_pid=$!
+# Run non parallel tests
```
sethiay, line 173: nit: Add logical space

sethiay (approved):

sethiay (approved):
Merge after the Kokoro tests pass.

PR #1797, opened 2024-03-18
Delete local file in case of failure

ashmeenkaur:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


sethiay (commented):
For the integration tests, could you also please get it reviewed from @Tulsishah ?

tools/integration_tests/readonly_creds/readonly_creds_test.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			log.Printf("closeStorageClient failed: %v", err)
+		}
+	}()
+	client.SetupTestDirectory(ctx, storageClient, testDirName)
+
+	// To run mountedDirectory tests, we need both testBucket and mountedDirectory
+	// flags to be set, as local_file tests validates content from the bucket.
+	if setup.AreBothMountedDirectoryAndTestBucketFlagsSet() {
+		setup.RunTestsForMountedDirectoryFlag(m)
```
Tulsishah, line 62: If these will run only for creds then, here it should be 

```
if setup.MountedDirectory() != "" {
		t.Logf("These tests will not run with mounted directory..")
		return
	}
```

And it should be on line#44

Tulsishah (approved):

Tulsishah (approved):

sethiay (approved):

PR #1799, opened 2024-03-18
Add crc check to read cache functional tests

ashmeenkaur:
### Description

Instead of comparing content read with GCS and cached file with GCS, calculate CRC of the file and validate that it matches the CRC values on GCS.

NA

### Testing details
1. Manual - Manually ran mounted directory tests.
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -16,6 +16,9 @@ package read_cache
 
 import (
 	"context"
+	"fmt"
+	"hash/crc32"
+	"io"
 	"log"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -119,15 +124,43 @@ func validateFileSizeInCacheDirectory(fileName string, filesize int64, t *testin
...
-		t.Errorf("Failed to read cached file %s: %v", expectedPathOfCachedFile, err)
+		fmt.Println("Error calculating CRC-32:", err)
+		return 0
 	}
-	client.ValidateObjectContentsFromGCS(ctx, storageClient, testDirName, fileName, string(content), t)
+	crc32Value = hasher.Sum32()
+	return
+}
+
+func ValidateCRCWithGCS(gotCRC32Value uint32, fileName string, ctx context.Context, storageClient *storage.Client, t *testing.T) {
```
sethiay, line 148: nit: fileName -> objectName

tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -175,9 +208,9 @@ func readFileAndValidateCacheWithGCS(ctx context.Context, storageClient *storage
 		// Validate cache size within limit.
 		validateCacheSizeWithinLimit(cacheCapacityInMB, t)
 	}
-	// Validate content read via gcsfuse with gcs.
-	client.ValidateObjectContentsFromGCS(ctx, storageClient, testDirName, filename,
-		expectedOutcome.content, t)
+	// Validate CRC32 of content read via gcsfuse with CRC32 value on gcs.
+	gotCRC32Value := crc32.Checksum([]byte(expectedOutcome.content), crc32cTable)
```
sethiay, line 212: would suggest to keep the way of calculating CRC same, so you can create a function which takes reader type/bytes slice and gives crc32 value and use that func here and above

sethiay (approved):

PR #1800, opened 2024-03-19
Adding GRPC client configuration in Mount configuration

raj-prince:
### Description
Adding conn-pool-size as part of grpc client config.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/config/mount_config.go
+++ b/internal/config/mount_config.go
@@ -47,6 +47,7 @@ const (
 
 	DefaultFileCacheMaxSizeMB               int64 = -1
 	DefaultEnableEmptyManagedFoldersListing       = false
+	DefaultGrpcConnectionPoolSize                 = 1
 )
 
 type WriteConfig struct {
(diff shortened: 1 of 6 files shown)


internal/config/mount_config.go
```diff
@@ -71,6 +72,11 @@ type ListConfig struct {
 	EnableEmptyManagedFolders bool `yaml:"enable-empty-managed-folders"`
 }
 
+type GrpcClientConfig struct {
+	// ConnectionPoolSize configures the number of gRPC channel in grpc client.
+	ConnectionPoolSize int `yaml:"conn-pool-size,omitempty"`
```
sethiay, line 77: Should we make it Connection**s**  and conn**s**? similar to max-conns-per-host flag

sethiay (approved):

sethiay (approved):

PR #1801, opened 2024-03-19
Add hns integration test

ankitaluthra1:
### Description

create a main method for all execution steps to be orchestrated at one place

extracted bucket creation and cleanup to a method | add hns bucket creation and clean up 

run operations and implicit package tests for hns bucket

adds create_hns_buckets to create bucket for hns

seggregated check for hns test failures

reafctor variable names to lower case to make it consistent and more readable ([ref](https://stackoverflow.com/questions/673055/correct-bash-and-shell-script-variable-capitalization#:~:text=If%20it%27s%20your%20variable%2C%20lowercase%20it.%20If%20you%20export%20it%2C%20uppercase%20it.))

fix echo statement for hns bucket name

refactoring move actual test calls into functions for flat and hns bu… 

adds bucket for hns in allow list project

b/328727379

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - Done

### DIFF
--- a/perfmetrics/scripts/run_e2e_tests.sh
+++ b/perfmetrics/scripts/run_e2e_tests.sh
@@ -15,47 +15,6 @@
 
 # This will stop execution when any command will have non-zero status.
 
-set -e
-
-# true or false to run e2e tests on installedPackage
-run_e2e_tests_on_package=$1
-readonly integration_test_timeout=40m
-readonly project_id="gcs-fuse-test-ml"
-readonly hns_project_id="gcs-fuse-test"
-readonly bucket_location="us-west1"
-
-# Test directory arrays
-test_dir_parallel=(
-  "local_file"
-  "log_rotation"
-  "mounting"
-  "read_cache"
-  "gzip"
-  "write_large_files"
-)
-
-# These tests never become parallel as it is changing bucket permissions.
-test_dir_non_parallel_group_1=(
-  "readonly"
-  "managed_folders"
-)
-
-# These test packages can be configured to run in parallel once they achieve
-# directory independence.
-test_dir_non_parallel_group_2=(
-  "explicit_dir"
-  "implicit_dir"
-  "list_large_dir"
-  "operations"
-  "read_large_files"
-  "rename_dir_limit"
-)
-
-test_dir_hns_group=(
-  "implicit_dir"
-  "operations"
-)
-
 function upgrade_gcloud_version() {
   sudo apt-get update
   # Upgrade gcloud version.
@@ -71,7 +30,6 @@ function upgrade_gcloud_version() {
   sudo /usr/local/google-cloud-sdk/bin/gcloud components install alpha
 }
 
-# Install packages
 function install_packages() {
   # e.g. architecture=arm64 or amd64
   architecture=$(dpkg --print-architecture)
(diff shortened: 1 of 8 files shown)


ankitaluthra1:
changes related to grpc have come from rebasing local branch with master, can skip from review

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -77,20 +79,29 @@ function install_packages() {
   sudo apt install -y python3-crcmod
 }
 
-# Create bucket for integration tests.
 function create_bucket() {
   bucket_prefix=$1
+  readonly project_id="gcs-fuse-test-ml"
```
sethiay, line 84: it should be "local" than "readonly" ?

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -138,69 +148,146 @@ function run_parallel_tests() {
...
-then
-  echo "The tests failed."
-  exit 1
-fi
+function create_flat_buckets() {
+  # Test setup
+  # Create Bucket for non parallel e2e tests
+  # The bucket prefix for the random string
+  bucketPrefix="gcsfuse-non-parallel-e2e-tests-group-1-"
+  BUCKET_NAME_NON_PARALLEL_GROUP_1=$(create_bucket $bucketPrefix)
```
sethiay, line 156: this shouldn't be in 
(a) CAPS_LOCK because it's not global variable.
(b) It is set here and then used in different function.

Would suggest to just keep one function run_e2e_tests_for_flat_bucket and create buckets inside it.

sethiay: I think you would also need to call clean_up from inside this for the same reason.

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -138,69 +148,146 @@ function run_parallel_tests() {
...
+
+  # Removing bin file after testing.
+  if [ $RUN_E2E_TESTS_ON_PACKAGE != true ];
+  then
+    sudo rm /usr/local/bin/gcsfuse
+  fi
+}
+
+#Main method to run script
+main
```
sethiay, line 291: question - does the script fails if something fails inside main, asking because we are not returning any exit code from main, also "set +e" is set inside main before run_tests is called.

perfmetrics/scripts/run_e2e_tests.sh
ankitaluthra1: done

perfmetrics/scripts/run_e2e_tests.sh
ankitaluthra1: its a bug !! script will not fail if any statement errors out

perfmetrics/scripts/run_e2e_tests.sh
ankitaluthra1: that ll make function very long , want to avoid that

sethiay: Ideally, the bucket names should be passed as argument between functions. However, I feel that will be overkill here. I would maintain consistency and use CAPS_LOCK for constants and allow the function to grow big. Wdyt ?

PR #1802, opened 2024-03-19
GRPC: PR2: Added grpc support in the storage package

raj-prince:
### Description
Added support in the storage package, to create gRPC storage based on Client-protocol value.
- User agent support.
- Custom endpoint support.

NA

### Testing details
1. Manual - NA
2. Unit tests - Automation
3. Integration tests - Automation (gRPC related integration test will run after complete integration).

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -16,9 +16,10 @@ package storage
 
 import (
 	"fmt"
-	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 	"net/http"
 	"os"
+
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 
 	"cloud.google.com/go/storage"
 	"github.com/googleapis/gax-go/v2"
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -15,9 +15,8 @@
 package storage
 
 import (
-	"errors"
 	"fmt"
-	"log"
+	"github.com/googlecloudplatform/gcsfuse/internal/logger"
 	"net/http"
 	"os"
 
@@ -48,12 +47,8 @@ type storageClient struct {
 
 // Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
 func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig) (sc *storage.Client, err error) {
-	if clientConfig.ClientProtocol != mountpkg.GRPC {
-		return nil, errors.New("wrong client-protocol requested")
-	}
-
 	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
-		log.Fatalf("error setting direct path env var: %v", err)
+		logger.Fatal("error setting direct path env var: %v", err)
 	}
 
 	var clientOpts []option.ClientOption
@@ -76,7 +71,7 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 
 	// Unset the environment variable, since it's used only while creation of grpc client.
 	if err := os.Unsetenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS"); err != nil {
-		log.Fatalf("error while unsetting direct path env var: %v", err)
+		logger.Fatal("error while unsetting direct path env var: %v", err)
 	}
 
 	return
@@ -95,8 +90,6 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		}
 
 		clientOpts = append(clientOpts, option.WithHTTPClient(httpClient))
-	} else {
-		return nil, errors.New("wrong client-protocol requested")
 	}
 
 	// Create client with JSON read flow, if EnableJasonRead flag is set.
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -15,8 +15,11 @@
 package storage
 
 import (
+	"errors"
 	"fmt"
+	"log"
 	"net/http"
+	"os"
 
 	"cloud.google.com/go/storage"
 	"github.com/googleapis/gax-go/v2"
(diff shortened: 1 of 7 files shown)


raj-prince:
Perf numbers:
https://screenshot.googleplex.com/6VnnGCJLN67e7uv

internal/storage/storage_handle.go
```diff
@@ -39,22 +46,57 @@ type storageClient struct {
 	client *storage.Client
 }
 
-// NewStorageHandle returns the handle of Go storage client containing
-// customized http client. We can configure the http client using the
-// storageClientConfig parameter.
-func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClientConfig) (sh StorageHandle, err error) {
+// Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
+func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig) (sc *storage.Client, err error) {
+	if clientConfig.ClientProtocol != mountpkg.GRPC {
```
sethiay, line 51: it's fine to remove this check as this func's responsibility is to create GRPC client. if this check has to be added, it should be added in the function calling this function.

we can do the same for http client func as well

internal/storage/storage_handle.go
```diff
@@ -39,22 +46,57 @@ type storageClient struct {
...
+		return nil, errors.New("wrong client-protocol requested")
+	}
+
+	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
+		log.Fatalf("error setting direct path env var: %v", err)
```
sethiay, line 56: we should use logger here ?

internal/storage/storageutil/client.go
```diff
@@ -28,18 +28,27 @@ import (
 )
 
 type StorageClientConfig struct {
-	ClientProtocol             mountpkg.ClientProtocol
+	/** Common client parameters. */
+
+	// ClientProtocol decides the go-sdk client to create. Based on the value,
+	// either http client or grpc client is created.
```
sethiay, line 34: nit: I think we can remove "Based on the value, either http client or grpc client is created."

internal/storage/storage_handle.go
raj-prince: Added it since, it was hard to write unit test for calling method - NewStorageHandle().

internal/storage/storage_handle.go
```diff
@@ -39,16 +46,45 @@ type storageClient struct {
...
 }
 
-// NewStorageHandle returns the handle of Go storage client containing
-// customized http client. We can configure the http client using the
-// storageClientConfig parameter.
-func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClientConfig) (sh StorageHandle, err error) {
+// Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
+func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig) (sc *storage.Client, err error) {
+	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
+		logger.Fatal("error setting direct path env var: %v", err)
```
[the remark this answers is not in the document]
raj-prince: Yes, GCSFuse is crashed with logger.Fatal.

Yes, IMO, we should fail the mounting. 
[Primary] If we are not able to set environment variable, that means there is some issue in customer's programme execution environment.

[Secondary] Not setting this lead the customer to use gRPC.CloudPath, which we don't want (if they have environment to use Direct Path).

internal/storage/storage_handle.go
```diff
@@ -67,8 +103,21 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
...
+// provided StorageClientConfig.ClientProtocol.
+// Please check out the StorageClientConfig to know about the parameters used in
+// http and gRPC client.
+func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClientConfig) (sh StorageHandle, err error) {
 	var sc *storage.Client
-	sc, err = storage.NewClient(ctx, clientOpts...)
+	if clientConfig.ClientProtocol == mountpkg.GRPC {
+		sc, err = createGRPCClientHandle(ctx, &clientConfig)
+	} else {
+		sc, err = createHTTPClientHandle(ctx, &clientConfig)
```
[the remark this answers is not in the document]
raj-prince: This way would be able to write unit test for wrong protocol (other than http1, http2 and grpc) but not to differentiate the http vs grpc.

internal/storage/storage_handle.go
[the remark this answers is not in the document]
raj-prince: Also, we set environment variable during the mount (as part of daemonize.Run()), so in case there is an issue in the gcsfuse run-environment, we will get the error at that time only.

internal/storage/storage_handle.go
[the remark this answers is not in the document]
raj-prince: As discussed offline, we keep the check at both places:
(a) createGRPCClientHandle & createHTTPClientHandle - since it's hard to check the created client type (either http or grpc) as part of NewStorageHandle, hence added check in the createGRPCClientHandle & createHTTPClientHandle. This way, we can ensure, this is being called with right client-protocol.

(b) Also, added check in the NewStorageHandle to ensure, this is not being called with invalid client-protocol (neither of http1, http2 and grpc).

internal/storage/storage_handle.go
raj-prince: Discussed offline to keep the check in both createGRPCClientHandle & createHTTPClientHandle - to make sure, this is being called with right client-protocol.

internal/storage/storage_handle.go
```diff
@@ -39,22 +46,57 @@ type storageClient struct {
...
 }
 
-// NewStorageHandle returns the handle of Go storage client containing
-// customized http client. We can configure the http client using the
-// storageClientConfig parameter.
-func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClientConfig) (sh StorageHandle, err error) {
+// Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
+func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig) (sc *storage.Client, err error) {
+	if clientConfig.ClientProtocol != mountpkg.GRPC {
+		return nil, fmt.Errorf("wrong client-protocol requested: %s", clientConfig.ClientProtocol)
```
sethiay, line 52: nit: you can use message like  "client-protocol requested is not GRPC: %s". This way you will be able to differentiate from error in createHTTPClientHandle

sethiay (approved):

internal/storage/storage_handle.go
raj-prince: Make sense, I'll update the error message in the next PR.

PR #1810, opened 2024-03-26
Managed folder delete create tests with admin permission

Tulsishah:
### Description
We are running these tests for static mounting only as these tests are taking almost 1 hour if we run it on  all the mountings, thus are changed the approach ti run only this file tests with static mounting.

Managed folder has view permission
1. Create object in managed folder - Successful
2. Delete object from managed folder - Successful
3. Delete managed folder - Successful
4. Copy object in managed folder - Successful
5. Copy managed folder(except bucket with view permission) - Successful 
6. Move object in managed folder - Successful 
7. Move managed folder (except bucket with view permission)  -  Successful 
8. Copy managed folder(bucket with view permission) - Operation should fail
9. Move managed folder (bucket with view permission)  -  Operation should fail

NA

### Testing details
1. Manual - VM
2. Unit tests - NA


tools/integration_tests/managed_folders/admin_permissions_test.go
```diff
@@ -33,18 +35,171 @@ import (
 // Boilerplate
 // //////////////////////////////////////////////////////////////////////
 
+const (
+	MoveFile       = "moveFileAdminPerm"
+	MoveDestFile   = "moveDestFileAdminPerm"
```
ashmeenkaur, line 40: MoveDestFile is unused. I also think you can use just two variables for source and destination here.

Tulsishah: It used the same for the destination. In the case of the source, it gives an intermittent error of `Error in stating file` as we create resources before a test and delete them after running the test. It is taking time to reflect. Adding sleep is working, but it will increase run time; that's why I used different sources.

ashmeenkaur:
Can you update the PR description mentioning what should be the expected behavior for each test and why we are running the tests only for static mounting?

tools/integration_tests/managed_folders/admin_permissions_test.go
```diff
@@ -58,43 +205,36 @@ func (s *managedFoldersAdminPermission) TestListNonEmptyManagedFoldersWithAdminP
...
+
+	for i := 0; i < len(permissions); i++ {
+		log.Printf("Running tests with flags, bucket have %s permission and managed folder have %s permissions: %s", permissions[i][0], permissions[i][1], flags)
+		ts.bucketPermission = permissions[i][0]
+		if ts.bucketPermission == ViewPermission {
+			creds_tests.RevokePermission(serviceAccount, AdminPermission, setup.TestBucket())
+			creds_tests.ApplyPermissionToServiceAccount(serviceAccount, ViewPermission)
+			defer creds_tests.RevokePermission(serviceAccount, ViewPermission, setup.TestBucket())
+		}
+		ts.iamPermission = permissions[i][1]
```
ashmeenkaur, line 237: Shall we call this managedFolderPermission instead?

tools/integration_tests/managed_folders/admin_permissions_test.go
```diff
@@ -58,43 +205,36 @@ func (s *managedFoldersAdminPermission) TestListNonEmptyManagedFoldersWithAdminP
...
-	serviceAccount, localKeyFilePath := creds_tests.CreateCredentials()
+	serviceAccount, localKeyFilePath = creds_tests.CreateCredentials()
 	creds_tests.ApplyPermissionToServiceAccount(serviceAccount, AdminPermission)
 	// Revoke permission on bucket.
 	defer creds_tests.RevokePermission(serviceAccount, AdminPermission, setup.TestBucket())
 
-	flags := []string{"--implicit-dirs", "--key-file=" + localKeyFilePath}
+	flags := []string{"--implicit-dirs", "--key-file=" + localKeyFilePath, "--rename-dir-limit=5", "--stat-cache-ttl=0"}
 
+	if setup.OnlyDirMounted() != "" {
```
ashmeenkaur, line 218: This is not required I think if tests are running only for static mounting

tools/integration_tests/managed_folders/admin_permissions_test.go
```diff
@@ -33,18 +35,163 @@ import (
...
 var (
-	bucket  string
-	testDir string
+	bucket           string
+	testDir          string
+	serviceAccount   string
+	localKeyFilePath string
 )
 
 type managedFoldersAdminPermission struct {
```
ashmeenkaur, line 49: struct name is misleading as there are tests for different combination of permissions

tools/integration_tests/managed_folders/admin_permissions_test.go
```diff
@@ -33,18 +35,163 @@ import (
...
+func (s *managedFoldersAdminPermission) TestDeleteObjectInManagedFolder(t *testing.T) {
+	filePath := path.Join(setup.MntDir(), TestDirForManagedFolderTest, ManagedFolder1, FileInNonEmptyManagedFoldersTest)
+
+	err := os.Remove(filePath)
+	if err != nil {
+		t.Errorf("Error in removing file from managed folder: %v", err)
+	}
+}
+
+// Managed folders will not be deleted, but they will become empty. Default empty managed folders will be hidden.
```
ashmeenkaur, line 85: By Default empty managed folders will be hidden - Can we make a stat call to validate this?

tools/integration_tests/managed_folders/admin_permissions_test.go
```diff
@@ -33,18 +35,163 @@ import (
...
+	if err != nil {
+		t.Errorf("Error in stating destination copy dir: %v", err)
+	}
+	_, err = os.Stat(srcDirPath)
+	if err == nil {
+		t.Errorf("SrcDir is not removed after move.")
+	}
+}
+
+func (s *managedFoldersAdminPermission) TestMoveManagedFolderWithViewBucketPermission(t *testing.T) {
```
ashmeenkaur, line 177: Nit: I would suggest create a common test TestMoveManagedFolder and you can create 2 helper methods - validateMoveManagedFolderWithViewBucketPermission and validateManagedFolderWithAdminBucketPermission

Based on the type of permissions, you can do different validations. Setup and operation will be common for both.

Similar for TestCopyManagedFolder also.

Tulsishah: Good Idea. To more optimize it I have create single validator only with permission as argument.

ashmeenkaur (approved):

PR #1816, opened 2024-03-28
Printing logs at the end

Tulsishah:
### Description
Our test logs are currently unorganized, with output from different packages intermixed. This makes it extremely difficult to troubleshoot failures because we can't easily isolate logs from specific tests.

Create a temporary file to store log file names from a background process. Create an array from it and print logs from the given log files, then delete them.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/perfmetrics/scripts/run_e2e_tests.sh
+++ b/perfmetrics/scripts/run_e2e_tests.sh
@@ -166,8 +166,6 @@ function print_test_logs() {
       echo "=== Log for ${test_log_file} ==="
       cat "$log_file"
       echo "========================================="
-      # Remove the log file
-      rm "$log_file"
     fi
   done
 }


perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -145,6 +156,22 @@ function run_parallel_tests() {
...
+  rm "$TEST_LOGS_FILE"
+  for test_log_file in "${TEST_LOGS_ARRAY[@]}"
+  do
+    log_file=${test_log_file}
+    if [ -f "$log_file" ]; then
+      echo "=== Log for ${test_log_file} ==="
+      cat "$log_file"
+      echo "========================================="
+      # Remove the log file
+      rm "$log_file"
```
sethiay, line 170: nit: there is no harm in not removing the log file. incase someone ssh into the VM, they will be able to see the log files.

perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -145,6 +156,22 @@ function run_parallel_tests() {
   return $exit_code
 }
 
+function print_test_logs() {
+  readarray -t TEST_LOGS_ARRAY < "$TEST_LOGS_FILE"
```
sethiay, line 160: nit: TEST_LOGS_ARRAY should be in small as it is a local variable ?

perfmetrics/scripts/run_e2e_tests.sh
Tulsishah: Okay, If we want to keep it then I changed path to `/tmp` for not creating redundant files in current directory.

sethiay: please remove `rm` then ?

Tulsishah: Sorry missed. Done.

sethiay (approved):

PR #1827, opened 2024-04-03
Intro anonymous-access flag

Tulsishah:
### Description
- Introducing new flag `auth-config: anonymous-access: true/false`.
- Authentication is by default on(`auth-config: anonymous-access:false`)
- In case of disabling authentication, We have to externally pass `anonymous-access: skip: true/false`. flag in config file.
- gRPC client 
     - Logging incase of Cx pass `--custom-endpoint` with gRPC flow that `GRPC path is not supported with authentication on custom endpoint.`
    
NA

### Testing details
1. Manual
9fLHX9SSyyv3eTR
3jh8ZHdAYkEpJiZ
3. Unit tests - Added


internal/config/mount_config.go
```diff
@@ -78,6 +79,12 @@ type GrpcClientConfig struct {
 	ConnPoolSize int `yaml:"conn-pool-size,omitempty"`
 }
 
+type Auth struct {
```
raj-prince, line 82: nit: AuthConfig?

internal/config/mount_config.go
```diff
@@ -115,6 +122,7 @@ type MountConfig struct {
 	MetadataCacheConfig `yaml:"metadata-cache"`
 	ListConfig          `yaml:"list"`
 	GrpcClientConfig    `yaml:"grpc"`
+	Auth                `yaml:"auth"`
```
raj-prince, line 125: nit: same here.

internal/storage/storage_handle.go
```diff
@@ -105,6 +108,10 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		return nil, fmt.Errorf("client-protocol requested is not HTTP1 or HTTP2: %s", clientConfig.ClientProtocol)
 	}
 
+	if clientConfig.SkipAuth {
```
raj-prince, line 111: We add tokenSrc while creating http client. Should we also stop the creation of token source while creating httpClient?

internal/storage/storage_handle.go
```diff
@@ -61,6 +61,9 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 
 	// Add Custom endpoint option.
 	if clientConfig.CustomEndpoint != nil {
+		if !clientConfig.SkipAuth {
```
raj-prince, line 64: Skip auth is independent from TPC.

I think, we need to add line: 70 and 80 for gRPC if skipAuth is true.
Also, this should happen irrespective of custom-endpoint.

internal/storage/storageutil/client.go
```diff
@@ -76,39 +77,40 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig) (httpClient *htt
...
+		// Custom http client for Go Client.
+		httpClient = &http.Client{
+			Transport: &oauth2.Transport{
+				Base:   transport,
+				Source: tokenSrc,
+			},
+			Timeout: storageClientConfig.HttpClientTimeout,
+		}
+		// Setting UserAgent through RoundTripper middleware
+		httpClient.Transport = &userAgentRoundTripper{
```
raj-prince, line 101: Why are we removing user-agent, when skip-auth is true? Ideally, we should only remove tokenSrc part in no-auth scenario.

internal/storage/storageutil/client_test.go
```diff
@@ -47,54 +46,57 @@ func (t *clientTest) validateProxyInTransport(httpClient *http.Client) {
...
+	httpClient, err := CreateHttpClient(&sc)
 
 	ExpectEq(nil, err)
-	ExpectNe(nil, &tokenSrc)
+	ExpectNe(nil, httpClient)
+	ExpectEq(sc.HttpClientTimeout, httpClient.Timeout)
 }
 
-func (t *clientTest) TestCreateTokenSrcWhenCustomEndpointIsNil() {
+func (t *clientTest) TestCreateHttpClientWithHttp2() {
```
raj-prince, line 59: We should also write unit test for:

(a) CreateHttpClientWithHttp2 (with and without auth)
(b) CreateHttpClientWithHttp1 (with and without auth)

main_test.go
```diff
@@ -176,19 +176,22 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInMountConfigAsMarshal
 	actual, err := util.Stringify(mountConfig)
 	AssertEq(nil, err)
 
-	expected := "{\"CreateEmptyFile\":false,\"Severity\":\"TRACE\",\"Format\":\"\",\"FilePath\":\"\\\"path\\\"to\\\"file\\\"\",\"LogRotateConfig\":{\"MaxFileSizeMB\":2,\"BackupFileCount\":2,\"Compress\":true},\"MaxSizeMB\":0,\"CacheFileForRangeRead\":false,\"CacheDir\":\"\",\"TtlInSeconds\":0,\"TypeCacheMaxSizeMB\":0,\"StatCacheMaxSizeMB\":0,\"EnableEmptyManagedFolders\":false,\"ConnPoolSize\":0,\"EnableHNS\":true}"
+	expected := "{\"CreateEmptyFile\":false,\"Severity\":\"TRACE\",\"Format\":\"\",\"FilePath\":\"\\\"path\\\"to\\\"file\\\"\",\"LogRotateConfig\":{\"MaxFileSizeMB\":2,\"BackupFileCount\":2,\"Compress\":true},\"MaxSizeMB\":0,\"CacheFileForRangeRead\":false,\"CacheDir\":\"\",\"TtlInSeconds\":0,\"TypeCacheMaxSizeMB\":0,\"StatCacheMaxSizeMB\":0,\"EnableEmptyManagedFolders\":false,\"ConnPoolSize\":0,\"Skip\":false,\"EnableHNS\":true}"
 	AssertEq(expected, actual)
 }
 
-func (t *MainTest) TestEnableHNSFlagFalse() {
+func (t *MainTest) TestEnableHNSFlagFalseAndSkipAuthIsTrue() {
```
raj-prince, line 183: Why do we merge two un-related flags?

Either we need to make a common function for all such boolean flags. Or we should create two different tests independently for different flags.

raj-prince (commented):
Few comments. Please check.

main_test.go
Tulsishah: Removed it. Discussed with @ashmeenkaur and concluded that it is not required to add this type of tests.

internal/storage/storage_handle.go
Tulsishah: > I think, we need to add line: 70 and 80 for gRPC if skipAuth is true.
Added conditions for both CustomEndpoint nil or non-nil
> this should happen irrespective of custom-endpoint.
For GCS endpoints, authentication will be skipped if 'Skip Auth' is enabled; otherwise, a tokenSrc will be added.

internal/storage/storageutil/client.go
Tulsishah: Thanks, I had understanding that tokenSrc is required for that as it added under transport.

Tulsishah: If we add Transport then it will have source: Token. If we pass token nil then while mounting it fails with `oauth2: Transport's Source is nil`.  So we can't add Transport while creating http client.

Tulsishah: @tritone is this expected behavour even if we pass WithOutAuthentication option?

Tulsishah: Checked with @codyoss , In case of defining transport in http client, it will override WithoutAuthentication option.

Tulsishah: WithUserAgent returns a ClientOption that sets the User-Agent. This option is incompatible with the WithHTTPClient option.

And we are setting client with `WithHTTPClient` option.

We can do this via RoundTripper, but it requires oAuth2.Transport initialized in http client with tokenSrc.

internal/storage/storage_handle.go
```diff
@@ -61,18 +61,26 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
-		clientOpts = append(clientOpts, option.WithoutAuthentication())
-		clientOpts = append(clientOpts, option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())))
+		if clientConfig.AnonymousAccess {
+			clientOpts = append(clientOpts, option.WithEndpoint(storageutil.StripScheme(clientConfig.CustomEndpoint.String())))
+			// Explicitly disable auth in case of custom-endpoint, aligned with the http-client.
+			// TODO: to revisit here when supporting TPC for grpc client.
+			clientOpts = append(clientOpts, option.WithoutAuthentication())
+			clientOpts = append(clientOpts, option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())))
+		} else {
+			logger.Fatal("GRPC path is not supported with authentication on custom endpoint.")
```
raj-prince, line 71: nit: `GRPC client doesn't support auth for custom-endpoint. Please set anonymous-access: true via config-file.`

Also, do we see this error while mounting? What happens while mounting in background mode.

internal/storage/storageutil/client.go
```diff
@@ -76,39 +77,39 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig) (httpClient *htt
...
-	// Custom http client for Go Client.
-	httpClient = &http.Client{
-		Transport: &oauth2.Transport{
-			Base:   transport,
-			Source: tokenSrc,
-		},
-		Timeout: storageClientConfig.HttpClientTimeout,
-	}
+	if storageClientConfig.AnonymousAccess {
+		httpClient = &http.Client{
```
raj-prince, line 81: We should write a comment - `why are we not supporting user-agent in case of anonymous-access?`

internal/storage/storageutil/client_test.go
```diff
@@ -47,54 +46,57 @@ func (t *clientTest) validateProxyInTransport(httpClient *http.Client) {
...
-	ExpectNe(nil, httpClient.Transport)
-	t.validateProxyInTransport(httpClient)
-	ExpectEq(sc.HttpClientTimeout, httpClient.Timeout)
+	AssertNe(nil, err)
+	ExpectThat(err, oglematchers.Error(oglematchers.HasSubstr("no such file or directory")))
+	AssertEq(nil, httpClient)
 }
 
-func (t *clientTest) TestCreateHttpClientWithHttp2() {
+func (t *clientTest) TestCreateHttpClientWithHttp2AndSkipAuthFalse() {
```
raj-prince, line 81: nit: We can rename to - `TestCreateHttpClientWithHttp2AndAuthEnabled`

raj-prince (commented):
LGTM, a couple of minor comments.

internal/storage/storage_handle.go
Tulsishah: > do we see this error while mounting?

Yes

> What happens while mounting in background mode.

It requires log file.

Tulsishah:
> LGTM, a couple of minor comments.
> 
> Can we also run performance test - since changes are around custom http client?
Bq6eJshZWHYpsNU

internal/storage/storage_handle.go
Tulsishah: return error instead of fatal. Now it will be visible in background process also.

internal/storage/storageutil/client.go
```diff
@@ -76,39 +77,43 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig) (httpClient *htt
...
-
-	// Custom http client for Go Client.
-	httpClient = &http.Client{
-		Transport: &oauth2.Transport{
-			Base:   transport,
-			Source: tokenSrc,
-		},
-		Timeout: storageClientConfig.HttpClientTimeout,
-	}
+	if storageClientConfig.AnonymousAccess {
```
raj-prince, line 80: Let's correct the comment formatting in the next PR (exposing anonymous-access as cli).

raj-prince (approved):

internal/storage/storageutil/client.go
Tulsishah: Sure.

PR #1828, opened 2024-04-04
Fix exit code issue in script

Tulsishah:
### Description
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/perfmetrics/scripts/run_e2e_tests.sh#L185
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/perfmetrics/scripts/run_e2e_tests.sh#L205

If a test fails, an error is raised/thrown, causing the program to exit that section of code and preventing execution of subsequent lines.
Fixed that by doing return instead of exit. 

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/perfmetrics/scripts/run_e2e_tests.sh
+++ b/perfmetrics/scripts/run_e2e_tests.sh
@@ -272,6 +272,12 @@ function main(){
 
   print_test_logs
 
+  if [ $e2e_tests_flat_bucket_status != 0 ] && [ $e2e_tests_hns_bucket_status != 0 ];
+  then
+    echo "The e2e tests for both flat and hns bucket failed.."
+    exit 1
+  fi
+
   if [ $e2e_tests_flat_bucket_status != 0 ];
   then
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/run_e2e_tests.sh
```diff
@@ -260,14 +262,14 @@ function main(){
...
   e2e_tests_flat_bucket_status=$?
 
   wait $e2e_tests_hns_bucket_pid
   e2e_tests_hns_bucket_status=$?
 
+  set -e
+
   print_test_logs
 
   if [ $e2e_tests_flat_bucket_status != 0 ];
```
sethiay, line 275: the tests can fail on both flat and hns buckets also, so can we handle that case i.e. if both exit codes are non zero then print both flat and hns failed

Tulsishah: Good point. Added thanks.

sethiay (approved):

PR #1840, opened 2024-04-10
Change authentication flow for tpc

Tulsishah:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -24,11 +24,52 @@ import (
 	storagev1 "google.golang.org/api/storage/v1"
 )
 
+const universeDomainDefault = "googleapis.com"
+
+func getUniverseDomain(ctx context.Context, contents []byte, scope string) (string, error) {
+	creds, err := google.CredentialsFromJSON(ctx, contents, scope)
+	if err != nil {
+		err = fmt.Errorf("Error in getting credentials.")
+		return "", err
+	}
+	domain, err := creds.GetUniverseDomain()
+	if err != nil {
+		err = fmt.Errorf("Error in getting universe domain.")
+		return "", err
+	}
+
+	return domain, nil
+}
+
+func getAccessToken(ctx context.Context, domain, scope string, contents []byte) (ts oauth2.TokenSource, err error) {
+	if domain == universeDomainDefault {
+		// By default, a standard OAuth 2.0 token source is created
+		// Create a config struct based on its contents.
+		jwtConfig, err := google.JWTConfigFromJSON(contents, scope)
+		if err != nil {
+			err = fmt.Errorf("JWTConfigFromJSON: %w", err)
+		}
+		// Create the token source.
+		ts = jwtConfig.TokenSource(ctx)
+	} else {
+		// For non-GDU universe domains, token exchange is impossible and services
+		// must support self-signed JWTs with scopes.
+
+		// Create a config struct based on its contents.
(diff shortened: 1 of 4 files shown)


internal/auth/auth.go
```diff
@@ -24,6 +24,24 @@ import (
 	storagev1 "google.golang.org/api/storage/v1"
 )
 
+const universeDomainDefault = "googleapis.com"
+
+func getUniverseDomain(ctx context.Context, contents []byte, scope string) (string, error) {
+	creds, err := google.CredentialsFromJSON(ctx, contents, scope)
+	if err != nil {
+		err = fmt.Errorf("Error in getting credentials: %v", err)
```
raj-prince, line 32: Let's return the error by adding prefix of method name. Also use %w in most of the cases.

err = fmt.Errorf("CredentialsFromJSON(): %w", err)
Similarly for other places.

internal/auth/auth.go
```diff
@@ -37,16 +55,31 @@ func newTokenSourceFromPath(
...
+	if domain != universeDomainDefault {
+		// Create self signed JWT access token.
+		ts, err = google.JWTAccessTokenSourceWithScope(contents, scope)
+		if err != nil {
+			err = fmt.Errorf("JWTConfigFromJSON: %w", err)
+			return
+		}
+	}
 	return
 }
```
raj-prince, line 85: I think, we should update method documentation based on the changed behavior.

internal/auth/auth_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	ExpectEq(nil, err)
+	ExpectEq(universeDomainDefault, domain)
+}
+
+func (t *AuthTest) TestGetUniverseDomainForTPC() {
+	path := "testdata/tpc_creds.json"
+	contents, err := os.ReadFile(path)
+	AssertEq(nil, err)
+	ctx := context.Background()
```
raj-prince, line 64: nit: can't we directly pass to the method instead of creating a new variable?

raj-prince (commented):
Few minor comment, PTAL.

Tulsishah:
> Few minor comment, PTAL.

Fixed. PTAL

internal/auth/auth.go
```diff
@@ -24,6 +24,24 @@ import (
...
+func getUniverseDomain(ctx context.Context, contents []byte, scope string) (string, error) {
+	creds, err := google.CredentialsFromJSON(ctx, contents, scope)
+	if err != nil {
+		err = fmt.Errorf("CredentialsFromJSON(): %w", err)
+		return "", err
+	}
+
+	domain, err := creds.GetUniverseDomain()
+	if err != nil {
+		err = fmt.Errorf("Error in getting universe domain: %w", err)
```
raj-prince, line 38: nit: ("GetUniverseDomain(): %w", err)

internal/auth/auth.go
```diff
@@ -37,21 +55,36 @@ func newTokenSourceFromPath(
...
 		err = fmt.Errorf("JWTConfigFromJSON: %w", err)
-		return
 	}
-
 	// Create the token source.
 	ts = jwtConfig.TokenSource(ctx)
 
+	domain, err := getUniverseDomain(ctx, contents, scope)
+	if err != nil {
+		return ts, err
```
raj-prince, line 69: nit: we can just `return`

raj-prince (approved):

PR #1850, opened 2024-04-17
Making rename dir limit test package parallell

Tulsishah:
### Description
Made rename dir limit package independent, to make it parallel run with other packages.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA


tools/integration_tests/rename_dir_limit/move_dir_test.go
```diff
@@ -194,21 +193,22 @@ func TestMoveDirectoryInEmptyDirectory(t *testing.T) {
 }
 
 func createDestNonEmptyDirectoryForMoveTest(t *testing.T) {
```
ashmeenkaur, line 195: Nit: Is this the only place where we are using testDir? If. yes, we can remove the global variable and pass destDir as parameter to the function. This will also make it consistent with `createSrcDirectoryWithObjectsForMoveDirTest` method.

ashmeenkaur (dismissed):

ashmeenkaur (approved):

sethiay (approved):
Relying on Ashmeen's review.

PR #1855, opened 2024-04-22
Introducing new flag to enable hns flow

Tulsishah:
### Description
A new flag, `--enable-hns`, has been introduced to enable the HNS flow when creating the storage control client for gRPC. The flag's default value is false, and it can be enabled by setting it to true.
config file structure.
```
enable-hns: true
```

NA

### Testing details
1. Manual - Tested manually
2. Unit tests - NA


internal/config/testdata/valid_config.yaml
```diff
@@ -20,3 +20,5 @@ list:
   enable-empty-managed-folders: true
 grpc:
   conn-pool-size: 4
+bucket-flow:
```
ankitaluthra1, line 23: similarly need to update this to bucket-level-feature-toggle

main_test.go
```diff
@@ -170,12 +170,15 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInMountConfigAsMarshal
...
+		},
 	}
 
 	actual, err := util.Stringify(mountConfig)
 	AssertEq(nil, err)
 
-	expected := "{\"CreateEmptyFile\":false,\"Severity\":\"TRACE\",\"Format\":\"\",\"FilePath\":\"\\\"path\\\"to\\\"file\\\"\",\"LogRotateConfig\":{\"MaxFileSizeMB\":2,\"BackupFileCount\":2,\"Compress\":true},\"MaxSizeMB\":0,\"CacheFileForRangeRead\":false,\"CacheDir\":\"\",\"TtlInSeconds\":0,\"TypeCacheMaxSizeMB\":0,\"StatCacheMaxSizeMB\":0,\"EnableEmptyManagedFolders\":false,\"ConnPoolSize\":0}"
+	expected := "{\"CreateEmptyFile\":false,\"Severity\":\"TRACE\",\"Format\":\"\",\"FilePath\":\"\\\"path\\\"to\\\"file\\\"\",\"LogRotateConfig\":{\"MaxFileSizeMB\":2,\"BackupFileCount\":2,\"Compress\":true},\"MaxSizeMB\":0,\"CacheFileForRangeRead\":false,\"CacheDir\":\"\",\"TtlInSeconds\":0,\"TypeCacheMaxSizeMB\":0,\"StatCacheMaxSizeMB\":0,\"EnableEmptyManagedFolders\":false,\"ConnPoolSize\":0,\"EnableHNS\":true}"
 	AssertEq(expected, actual)
 }
```
ankitaluthra1, line 183: we should write one integration test also which covers mounting with this feature toggle true in the config file, should also add a new assertion checking the toggle as false in existing tests

internal/config/mount_config.go
```diff
@@ -77,6 +78,11 @@ type GrpcClientConfig struct {
 	ConnPoolSize int `yaml:"conn-pool-size,omitempty"`
 }
 
+type BucketFlow struct {
```
ankitaluthra1, line 81: Can we rename to BucketLevelFeatureToggle

ankitaluthra1 (changes requested):

main_test.go
Tulsishah: Yes I already added that in my test plan for integration tests. 

> should also add a new assertion checking the toggle as false in existing tests.
Could not able to test it in existing flow hence created new unit test.

internal/config/mount_config.go
sethiay: I am not sure if bucket-flow is easy to understand for end users. Should we rename it to something like "gcs-client: enable-hns", you can take check with Marco & Kislay on this.

Tulsishah: @ankitaluthra1 suggested `BucketLevelFeatureToggle`.

Tulsishah: After discussions with Marco and Kislay, we've decided to keep the feature flag as `enable-hns`. Marco had initially left the decision to the engineering team, as this feature won't be public-facing. Following further discussion with Kislay, we concluded that enable-hns is the best choice, since the likelihood of additional bucket-level features is low.

sethiay (approved):

ankitaluthra1 (approved):

PR #1859, opened 2024-04-23
add prometheus metrics support

songjiaxun:
### Description

Add Prometheus metrics support. Add a new flag `prometheus-port` for users to specify prometheus metrics endpoint port.

NA

### Testing details
1. Manual - Done

- Build the code using the command `go run tools/build_gcsfuse/main.go . . v2.0.0`.
- Mount a bucket: `./bin/gcsfuse --prometheus-port 8080 <bucket-name> <mount-path>`.
- Pull the metrics: `curl http://localhost:8080/metrics`
- See the output example in the docs/metrics.md.

3. Unit tests - NA
4. Integration tests - NA


flags.go
```diff
@@ -291,6 +291,12 @@ func newApp() (app *cli.App) {
 				Usage: "Experimental: Export metrics to the OpenTelemetry collector at this address.",
 			},
 
+			cli.StringFlag{
```
sethiay, line 294: we recommend adding new flag only via config file until we make parity between cli and config file. you can refer to this PR on how to add flag to config file: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1800/files

internal/monitor/exporter.go
```diff
@@ -106,3 +109,59 @@ func CloseOpenTelemetryCollectorExporter() {
...
+	mux.HandleFunc("/metrics", prometheusExporter.ServeHTTP)
+	prometheusServer = &http.Server{
+		Addr:           ":" + port,
+		Handler:        mux,
+		ReadTimeout:    10 * time.Second,
+		WriteTimeout:   10 * time.Second,
+		MaxHeaderBytes: 1 << 20,
+	}
+
+	go func() {
```
sethiay, line 146: Though this should be lightweight but we should do run perf testing of GCSfuse with promotheus server running and also ensure that this doesn't take extra memory and CPU

flags.go
songjiaxun: Done.

internal/monitor/exporter.go
songjiaxun: Hi @sethiay, thanks for the review! I updated the PR to move the mount flag to config file. Do you happen to have a guidance to do the GCSFuse perf test and memory/CPU resource tests?

internal/config/mount_config.go
```diff
@@ -46,6 +46,7 @@ const (
 	DefaultGrpcConnPoolSize                       = 1
 	DefaultAnonymousAccess                        = false
 	DefaultEnableHNS                              = false
+	DefaultPrometheusPport                        = ""
```
hime, line 49: DefaultPrometheusPport > DefaultPrometheusPort

flags.go
```diff
@@ -585,6 +586,7 @@ func populateFlags(c *cli.Context) (flags *flagStorage, err error) {
 		// Monitoring & Logging
 		StackdriverExportInterval:  c.Duration("stackdriver-export-interval"),
 		OtelCollectorAddress:       c.String("experimental-opentelemetry-collector-address"),
+		PrometheusPort:             c.String("prometheus-port"),
```
bjornleffler, line 589: Port number should be Uint, not String.

docs/metrics.md
```diff
@@ -60,5 +63,83 @@ on the Google cloud project.
...
+```bash
+curl http://localhost:8080/metrics
+```
+
+The output is similar to the following:
+
+```text
+# HELP file_cache_read_bytes_count The cumulative number of bytes read from file cache along with read type - Sequential/Random
+# TYPE file_cache_read_bytes_count counter
+file_cache_read_bytes_count{read_type="Random"} 0
```
bjornleffler, line 92: This might be coming from somewhere else, but could we prefix all metrics names with "gcsfuse_", for example 
gcsfuse_file_cache_read_bytes_count ?

flags.go
```diff
@@ -435,6 +435,7 @@ type flagStorage struct {
 	// Monitoring & Logging
 	StackdriverExportInterval  time.Duration
 	OtelCollectorAddress       string
+	PrometheusPort             string
```
hime, line 438: I think we want to remove all mentions of `PrometheusPort` from this file.

flags.go
songjiaxun: Done.

flags.go
songjiaxun: Actually, a recent commit https://github.com/GoogleCloudPlatform/gcsfuse/commit/d0868157afd71d08f2f02f11c802ca3bd97f71bb#diff-84fa0fc3b79e063575d5e65751f76d79c1f961ef42facf47fd225d65e0ff5cfb shows that config file and mount flag are added at the same time, modified this PR accordingly.

docs/metrics.md
songjiaxun: I'd like to keep this PR focusing on the prometheus integration. I can create a follow-up PR to modify the metrics name, but I recommend GCSFuse team make such changes because it will introduce compatibility issues as the current test pipelines may rely on these metrics names.

internal/monitor/exporter.go
songjiaxun: As we discussed offline, our tests show that, with this change included and with the prometheus endpoint enabled, the gcsfuse resource usage does not increase significantly in FIO tests.

kislaykishore (changes requested):

internal/config/testdata/valid_config.yaml
```diff
@@ -31,3 +31,5 @@ enable-hns: true
 file-system:
   ignore-interrupts: true
   disable-parallel-dirops: true
+monitoring:
```
sethiay, line 34: i think this should be called as observability than monitoring.

@kislaykishore wdyt ?

kislaykishore: We are moving stackdriver params under "metrics". So, we should put this too there.

songjiaxun: Done.

cmd/flags.go
```diff
@@ -315,6 +315,12 @@ func newApp() (app *cli.App) {
 				Usage: "Experimental: Export metrics to the OpenTelemetry collector at this address.",
 			},
 
+			cli.IntFlag{
```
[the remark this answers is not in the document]
songjiaxun: Done. But I am not sure if I am missing anything, or adding something unnecessary. Please help take another look. Thanks!

internal/monitor/exporter.go
```diff
@@ -15,11 +15,14 @@
...
 
 import (
+	"context"
 	"fmt"
+	"net/http"
 	"strings"
 	"time"
 
 	"contrib.go.opencensus.io/exporter/ocagent"
+	"contrib.go.opencensus.io/exporter/prometheus"
```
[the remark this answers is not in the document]
songjiaxun: I choose to use this archived repo to make the exporters consistent as other exporters are also using the same archived lib, for example `contrib.go.opencensus.io/exporter/stackdriver`.

I am working on a separate commit to try to migrate the `contrib.go.opencensus.io` lib to `[opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go)`, but it may take longer.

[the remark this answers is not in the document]
songjiaxun: After some attempts to refactor and migrate, I realize that all the existing metrics exporters rely on `internal/monitor/reader.go` and `internal/monitor/bucket.go` to report metrics. These files are using the archived lib `go.opencensus.io` as well. To migrate all of these libs is out of this PR's scope.

Can we accept this PR using the archived package, and create a separate task to migrate all the existing exporters to open telemetry?

internal/monitor/exporter.go
```diff
@@ -106,3 +109,59 @@ func CloseOpenTelemetryCollectorExporter() {
 	}
 	ocExporter = nil
 }
+
+var prometheusExporter *prometheus.Exporter
+var prometheusServer *http.Server
+
+// EnablePrometheusCollectorExporter starts exporting monitoring metrics for
+// the Prometheus to scrape on the given port.
+func EnablePrometheusCollectorExporter(port int) error {
```
kislaykishore, line 118: Calling this method multiple times with different ports will lead to memory leaks. We should probably throw an error if someone calls this method when there are preexisting server and exporter running.

internal/config/config_util.go
```diff
@@ -65,6 +66,14 @@ func OverrideWithKernelListCacheTtlFlag(c cliContext, mountConfig *MountConfig,
 	}
 }
 
+// OverrideWithPrometheusPortFlag overwrites the monitoring:prometheus-port config
```
kislaykishore, line 69:
```suggestion
// OverrideWithPrometheusPortFlag overwrites the metrics:prometheus-port config
```

kislaykishore (changes requested):

internal/config/config_util.go
songjiaxun: Done.

internal/monitor/exporter.go
songjiaxun: Done.

internal/monitor/exporter.go
```diff
@@ -106,3 +109,59 @@ func CloseOpenTelemetryCollectorExporter() {
...
+	if prometheusServer != nil {
+		if err := prometheusServer.Shutdown(context.Background()); err != nil {
+			logger.Errorf("Failed to shutdown Prometheus server: %v", err)
+		}
+	}
+
+	if prometheusExporter != nil {
+		view.UnregisterExporter(prometheusExporter)
+	}
+}
```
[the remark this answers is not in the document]
kislaykishore: I understand that. Given that other exporters are also not thread-safe, it's okay to keep it the way it is and make the changes in the future as needed.

kislaykishore (approved):

sethiay (approved):

internal/config/mount_config.go
```diff
@@ -219,5 +226,13 @@ func NewMountConfig() *MountConfig {
 		MaxRetryAttempts: DefaultMaxRetryAttempts,
 	}
 
+	mountConfig.GCSRetries = GCSRetries{
```
sethiay, line 229: @songjiaxun I think these changes are not part of this PR.

songjiaxun: Good catch! Done.

cmd/legacy_main.go
```diff
@@ -263,6 +263,7 @@ func runCLIApp(c *cli.Context) (err error) {
 	config.OverrideWithIgnoreInterruptsFlag(c, mountConfig, flags.IgnoreInterrupts)
 	config.OverrideWithAnonymousAccessFlag(c, mountConfig, flags.AnonymousAccess)
 	config.OverrideWithKernelListCacheTtlFlag(c, mountConfig, flags.KernelListCacheTtlSeconds)
+	config.OverrideWithPrometheusPortFlag(c, mountConfig, flags.PrometheusPort)
```
kislaykishore, line 267:
```suggestion
	config.OverrideWithKernelListCacheTtlFlag(c, mountConfig, flags.KernelListCacheTtlSeconds)
```

kislaykishore (changes requested):

internal/config/config_util.go
```diff
@@ -74,6 +75,14 @@ func OverrideWithMaxRetryAttemptFlag(c cliContext, mountConfig *MountConfig, ret
...
 }
 
+// OverrideWithPrometheusPortFlag overwrites the metrics:prometheus-port config
+// with the prometheus-port cli-flag value if the cli-flag is set by user.
+func OverrideWithPrometheusPortFlag(c cliContext, mountConfig *MountConfig, port int) {
+	if c.IsSet(PrometheusPortFlagName) {
+		mountConfig.MetricsConfig.PrometheusPort = port
+	}
+}
+
```
kislaykishore, line 85: This function is not needed.
```suggestion
```

cmd/legacy_main.go
songjiaxun: Done.

internal/config/config_util.go
songjiaxun: Removed.

sethiay (commented):
LGTM

kislaykishore (approved):

cmd/mount.go
```diff
@@ -100,7 +100,7 @@ be interacting with the file system.`)
 		OpRateLimitHz:                      newConfig.GcsConnection.LimitOpsPerSec,
 		StatCacheMaxSizeMB:                 statCacheMaxSizeMB,
 		StatCacheTTL:                       metadataCacheTTL,
-		EnableMonitoring:                   newConfig.Metrics.StackdriverExportInterval > 0,
+		EnableMonitoring:                   newConfig.Metrics.StackdriverExportInterval > 0 || newConfig.Metrics.PrometheusPort != 0,
```
songjiaxun, line 103: Hi @sethiay , @kislaykishore I added this line, and then from tests I can see `gcs_reader_count`, `gcs_request_count`, and `gcs_request_latencies` metrics. However, I still don't see `gcs_read_bytes_count` metric. Can you help take a look and see if I am missing anything?

kislaykishore: There is a bug in how this counter is computed - it's not updated unless it's a slightly large file and then too it's updated incorrectly. I'm working on https://github.com/GoogleCloudPlatform/gcsfuse/pull/2167 to fix this.
For your testing, create a 1G file in GCS and used `cat <filename> > /dev/null` to read it. That should move the counter:
image.

kislaykishore (approved):

sethiay (approved):

cmd/mount.go
songjiaxun: Thank you for the fix!

PR #1823, opened 2024-04-01
Enabling gRPC related integration test

raj-prince:
### Description
- Enabling e2e test for gRPC changes.
- Added support of -short flag while running the integration test, this will skip few tests inside test package. 
- Enabling only important tests (operations) for gRPC as part presubmit test. But exhaustive test will be run as part of e2e test.

Currently time taken by pre-submit test - 34m.

### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

PR #1862, opened 2024-04-24
Fix failure for kokoro e2e tests for implicit dir

Tulsishah:
### Description
I've removed the use of gsutil to delete directories. We're already handling directory cleanup with the [SetupDirectory function](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L378).

- This change is necessary because using gsutil was causing test failures due to delays in updating the stat-cache.

2.  Non-parallel test execution

I've removed the & symbol to prevent tests from running in parallel. This change was accidentally introduced in PR #1823

NA

### Testing details
1. Manual - NA
4. Unit tests - NA


ashmeenkaur (approved):

sethiay (approved):
Relying on Ashmeen's review.

PR #1865, opened 2024-04-26
 Initializing new storage control client

Tulsishah:
### Description
- Initializing new storage control client incase of GRPC protocol if enable-hns flag is turned on.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added


internal/storage/storage_handle.go
```diff
@@ -44,13 +45,14 @@ type StorageHandle interface {
...
 
 type storageClient struct {
-	client *storage.Client
+	client               *storage.Client
+	storageControlClient *control.StorageControlClient
 }
 
 // Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
-func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig) (sc *storage.Client, err error) {
+func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig) (sc *storage.Client, controlClient *control.StorageControlClient, err error) {
```
ankitaluthra1, line 53: instead of modifying existing method, we should create a new method which will do same thing i.e creating a grpc client but the client will be control client and it should read from a different flag which should be folder api client protocol

internal/storage/storage_handle.go
```diff
@@ -124,10 +142,13 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 // http and gRPC client.
 func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClientConfig) (sh StorageHandle, err error) {
 	var sc *storage.Client
+	// The default protocol for the Go Storage control client's folders API is gRPC.
+	// gcsfuse will initially mirror this behavior due to the client's lack of HTTP support.
+	var controlClient *control.StorageControlClient
 	if clientConfig.ClientProtocol == mountpkg.GRPC {
-		sc, err = createGRPCClientHandle(ctx, &clientConfig)
+		sc, controlClient, err = createGRPCClientHandle(ctx, &clientConfig)
```
ankitaluthra1, line 149: this should not be under same client protocol check, this if-else block will remain unaltered and a new block can be added which should create grpc control client irrespective of client protocol if enable hns flag is true

internal/storage/storage_handle.go
```diff
@@ -78,7 +69,52 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 	clientOpts = append(clientOpts, option.WithGRPCConnectionPool(clientConfig.GrpcConnPoolSize))
```
ankitaluthra1, line 69: clientConfig.GrpcConnPoolSize needs to be shared between control client and data client i.e if user has set clientConfig.GrpcConnPoolSize to be 100, then control client can have 2 and data clients can have 98, so total number of grpc connections are equal to this value. cc @raj-prince @sethiay

ankitaluthra1 (changes requested):
NA

internal/storage/storage_handle.go
```diff
@@ -78,7 +69,52 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
 	clientOpts = append(clientOpts, option.WithUserAgent(clientConfig.UserAgent))
 
+	return
+}
+
+func createGRPCControlClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig) (sc *control.StorageControlClient, err error) {
+	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
```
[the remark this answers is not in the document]
Tulsishah: It requires for control client also, checked from go sdk code.

internal/storage/storage_handle.go
```diff
@@ -152,7 +197,7 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
 		storage.WithPolicy(storage.RetryAlways),
 		storage.WithErrorFunc(storageutil.ShouldRetry))
```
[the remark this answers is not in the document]
Tulsishah: The Go SDK code appears to lack retries for creating and deleting folders, while retries for getting and renaming folders are infinite. We cannot configure retries for the control client from our side.

https://github.com/googleapis/google-cloud-go/blob/main/storage/control/apiv2/storage_control_client.go#L73
@tritone can you please confirm?

go.mod
```diff
@@ -61,6 +64,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.15.2 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/pkg/xattr v0.4.9 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/prometheus/prometheus v0.35.0 // indirect
 	github.com/rogpeppe/go-internal v1.12.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
```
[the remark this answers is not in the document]
Tulsishah: `cloud.google.com/go/longrunning` package is require to use control library.

main.go
```diff
@@ -114,6 +114,7 @@ func createStorageHandle(flags *flagStorage, mountConfig *config.MountConfig, us
 		ReuseTokenFromUrl:          flags.ReuseTokenFromUrl,
 		ExperimentalEnableJsonRead: flags.ExperimentalEnableJsonRead,
 		GrpcConnPoolSize:           mountConfig.GrpcClientConfig.ConnPoolSize,
+		EnableHNS:                  mountConfig.EnableHNS,
```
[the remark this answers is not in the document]
Tulsishah: Not needed

internal/storage/storage_handle.go
```diff
@@ -137,6 +176,12 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
 		return
 	}
 
+	if clientConfig.EnableHNS {
```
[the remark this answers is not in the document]
Tulsishah: Added.

internal/storage/storage_handle.go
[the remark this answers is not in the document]
tritone: @Tulsishah is correct about the defaults, but you can also configure retries per-call with the GAPIC client. Use https://pkg.go.dev/github.com/googleapis/gax-go/v2#WithRetry to configure; this can be passed to any GAPIC generated RPC.

internal/storage/storage_handle.go
ankitaluthra1: @Tulsishah we can keep this change as-is for now, will refactor once we reach consensus with go sdk team on this setting

internal/storage/storageutil/control_client.go
```diff
@@ -0,0 +1,64 @@
...
+
+func storageControlClientRetries(sc *control.StorageControlClient, clientConfig *StorageClientConfig) {
+	sc.CallOptions.CreateFolder = storageControlClientRetryOptions(clientConfig)
+	sc.CallOptions.DeleteFolder = storageControlClientRetryOptions(clientConfig)
+	sc.CallOptions.RenameFolder = storageControlClientRetryOptions(clientConfig)
+	sc.CallOptions.GetFolder = storageControlClientRetryOptions(clientConfig)
+	sc.CallOptions.GetStorageLayout = storageControlClientRetryOptions(clientConfig)
+}
+
+func CreateGRPCControlClientHandle(ctx context.Context, clientOpts []option.ClientOption, clientConfig *StorageClientConfig) (sc *control.StorageControlClient, err error) {
```
Tulsishah, line 45: @tritone  can you please check if I configured control client correctly with retries?

internal/storage/storageutil/control_client.go
```diff
@@ -0,0 +1,64 @@
...
+
+	control "cloud.google.com/go/storage/control/apiv2"
+	"github.com/googleapis/gax-go/v2"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
+	"google.golang.org/api/option"
+)
+
+func storageControlClientRetryOptions(clientConfig *StorageClientConfig) []gax.CallOption {
+	return []gax.CallOption{gax.WithRetry(func() gax.Retryer {
+		return gax.OnErrorFunc(gax.Backoff{
```
tritone, line 30: I think we should sync with the backend team on retryable codes. I see that they have defined these by default: https://github.com/googleapis/google-cloud-go/blob/storage/v1.40.0/storage/control/apiv2/storage_control_client.go#L79

DeadlineExceeded definitely seems wrong according to the AIP https://google.aip.dev/194

internal/storage/storage_handle.go
```diff
@@ -44,21 +45,11 @@ type StorageHandle interface {
...
-		return nil, fmt.Errorf("client-protocol requested is not GRPC: %s", clientConfig.ClientProtocol)
-	}
-
-	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
-		logger.Fatal("error setting direct path env var: %v", err)
-	}
-
-	var clientOpts []option.ClientOption
-
+func createClientOptionForGRPCClient(clientConfig *storageutil.StorageClientConfig) (clientOpts []option.ClientOption, err error) {
```
tritone, line 52: It would be good to add a comment to clarify whether this is for storage/v2 gRPC or storage/control gRPC or both.

Tulsishah: Sure will add the comment. It is for both.

internal/storage/storageutil/control_client.go
tritone: To clarify, I will do this offline-- not blocking for this PR.

internal/storage/storageutil/control_client.go
```diff
@@ -0,0 +1,64 @@
...
+			Max:        clientConfig.MaxRetrySleep,
+			Multiplier: clientConfig.RetryMultiplier,
+		}, ShouldRetry)
+	})}
+}
+
+func storageControlClientRetries(sc *control.StorageControlClient, clientConfig *StorageClientConfig) {
```
ankitaluthra1, line 37: can be renamed to setRetryConfigForFolderCalls

ankitaluthra1 (changes requested):

ankitaluthra1 (approved):

internal/storage/storage_handle.go
```diff
@@ -150,6 +167,18 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
 		return
 	}
 
+	// TODO: We will implement an additional check for the HTTP control client protocol once the Go SDK supports HTTP.
+	if clientConfig.EnableHNS {
+		clientOpts, err := createClientOptionForGRPCClient(&clientConfig)
+		if err != nil {
+			return nil, fmt.Errorf("Error in getting clientOpts for gRPC client: %w", err)
```
raj-prince, line 174: nit: We should follow this (https://go.dev/wiki/CodeReviewComments#error-strings) while writing error message. Please update at all the places in this change list.

internal/storage/storageutil/control_client.go
```diff
@@ -0,0 +1,64 @@
...
+func storageControlClientRetryOptions(clientConfig *StorageClientConfig) []gax.CallOption {
+	return []gax.CallOption{gax.WithRetry(func() gax.Retryer {
+		return gax.OnErrorFunc(gax.Backoff{
+			Max:        clientConfig.MaxRetrySleep,
+			Multiplier: clientConfig.RetryMultiplier,
+		}, ShouldRetry)
+	})}
+}
+
+func setRetryConfigForFolderAPIs(sc *control.StorageControlClient, clientConfig *StorageClientConfig) {
```
raj-prince, line 37: Do we know the reason, why can't we set the retry configuration to the control_client itself?

Tulsishah: Please take a look https://github.com/GoogleCloudPlatform/gcsfuse/pull/1865#discussion_r1583226442

internal/storage/storage_handle.go
```diff
@@ -87,7 +79,29 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
+	}
+
+	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
+		logger.Fatal("error setting direct path env var: %v", err)
+	}
+
+	var clientOpts []option.ClientOption
+	clientOpts, err = createClientOptionForGRPCClient(clientConfig)
+	if err != nil {
+		return nil, fmt.Errorf("Error in getting clientOpts for gRPC client: %w", err)
```
raj-prince, line 98: nit: here too.

raj-prince (approved):

PR #1867, opened 2024-04-26
Using Stretcher testify package in storage handle tests 

Tulsishah:
### Description
Using [Stretcher testify suites](https://github.com/stretchr/testify?tab=readme-ov-file#suite-package) for writing unit tests.

NA

### Testing details
1. Manual - NA
2. Unit tests - Git run


internal/storage/storage_handle_test.go
```diff
@@ -23,202 +23,200 @@ import (
...
+	suite.Nil(err)
 }
 
-func (t *StorageHandleTest) TearDown() {
-	t.fakeStorage.ShutDown()
+func (suite *StorageHandleTest) TearDownSuite() {
+	suite.fakeStorage.ShutDown()
 }
 
-func (t *StorageHandleTest) invokeAndVerifyStorageHandle(sc storageutil.StorageClientConfig) {
```
ankitaluthra1, line 54: this method is not necessary, it avoids duplication but while writing readable unit tests its better to have assertions directly in the test method, its okay to have repetitive code account of readability, see for [ref](https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests)

internal/storage/storage_handle_test.go
```diff
@@ -23,202 +23,200 @@ import (
...
+func TestStorageHandleSuite(t *testing.T) {
+	suite.Run(t, new(StorageHandleTest))
+}
 
-func (t *StorageHandleTest) SetUp(_ *TestInfo) {
+func (suite *StorageHandleTest) SetupSuite() {
 	var err error
-	t.fakeStorage = NewFakeStorage()
-	AssertEq(nil, err)
+	suite.fakeStorage = NewFakeStorage()
```
ankitaluthra1, line 44: any reason we are using suite. instead of assert. one of the usp of this library is rich assert package, see [ref](https://github.com/stretchr/testify?tab=readme-ov-file#assert-package) here, any reason why we are not leveraging it ? asserts are commonly used in unit tests in all languages if both suite and assert are having same functionality, should prefer asserts.

ankitaluthra1 (changes requested):

internal/storage/storage_handle_test.go
```diff
@@ -23,202 +23,200 @@ import (
...
 
-func (t *StorageHandleTest) TestBucketHandleWhenBucketExistsWithEmptyBillingProject() {
-	storageHandle := t.fakeStorage.CreateStorageHandle()
+func (suite *StorageHandleTest) TestBucketHandleWhenBucketExistsWithEmptyBillingProject() {
+	storageHandle := suite.fakeStorage.CreateStorageHandle()
 	bucketHandle := storageHandle.BucketHandle(TestBucketName, "")
 
-	AssertNe(nil, bucketHandle)
-	AssertEq(TestBucketName, bucketHandle.bucketName)
+	suite.NotNil(bucketHandle)
```
ankitaluthra1, line 62: any reason we are using suite. instead of assert. in all test methods ? one of the usp of this library is rich assert package, see [ref](https://github.com/stretchr/testify?tab=readme-ov-file#assert-package) here, any reason why we are not leveraging it ? asserts are commonly used in unit tests in all languages if both suite and assert are having same functionality, should prefer asserts.

ankitaluthra1 (changes requested):

internal/storage/storage_handle_test.go
Tulsishah: It is require to use suits as we have structure in the tests to create fake storage object, which will get created in setup before each tests. I found suits package is providing functionality same we are expecting in our current package. [ref](https://github.com/stretchr/testify?tab=readme-ov-file#suite-package)

ankitaluthra1 (changes requested):
NA

Tulsishah:
> NA

Please review it. Fixed comments.

ankitaluthra1 (approved):

internal/storage/storage_handle_test.go
```diff
@@ -25,79 +25,68 @@ import (
...
-var _ SetUpInterface = &StorageHandleTest{}
-var _ TearDownInterface = &StorageHandleTest{}
-
-func init() { RegisterTestSuite(&StorageHandleTest{}) }
+func TestStorageHandleSuite(t *testing.T) {
+	suite.Run(t, new(StorageHandleTest))
+}
 
-func (t *StorageHandleTest) SetUp(_ *TestInfo) {
+func (suite *StorageHandleTest) SetupSuite() {
```
sethiay, line 43: is SetUp to SetupSuite intentional ? I thought it will be SetupTest?

internal/storage/storage_handle_test.go
```diff
@@ -25,79 +25,68 @@ import (
...
 	var err error
-	t.fakeStorage = NewFakeStorage()
+	suite.fakeStorage = NewFakeStorage()
```
sethiay, line 45: IMO, it's preferable to rename suite to testSuite or ts or sht as `suite` is also imported package.

internal/storage/storage_handle_test.go
```diff
@@ -25,79 +25,68 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
 	"github.com/jacobsa/oglematchers"
 	. "github.com/jacobsa/ogletest"
```
sethiay, line 27: should we also use replace assertions with https://pkg.go.dev/github.com/stretchr/testify/assert#hdr-Assertions  and get remove ogletest and oglematchers?

internal/storage/storage_handle_test.go
Tulsishah: I took reference from [this](https://github.com/stretchr/testify/issues/1123), anyways rename it to SetupTest

internal/storage/storage_handle_test.go
Tulsishah: Done

internal/storage/storage_handle_test.go
Tulsishah: Sorry, I initially thought so, but then I realized that these assert statements are part of a testing package.
Fixed.

internal/storage/storage_handle_test.go
```diff
@@ -23,254 +23,257 @@ import (
...
 	sc.ClientProtocol = mountpkg.HTTP2
 	sc.AnonymousAccess = false
 
 	handleCreated, err := NewStorageHandle(context.Background(), sc)
 
-	AssertNe(nil, err)
-	ExpectThat(err, oglematchers.Error(oglematchers.HasSubstr("no such file or directory")))
-	AssertEq(nil, handleCreated)
+	assert.NotNil(testSuite.T(), err)
+	assert.True(testSuite.T(), strings.Contains(err.Error(), "no such file or directory"))
```
sethiay, line 99: you can either use https://pkg.go.dev/os#IsNotExist or make use of assert.Contains or assert.ErrorContains https://pkg.go.dev/github.com/stretchr/testify/assert#hdr-Assertions

sethiay (approved):

PR #1876, opened 2024-04-29
Skip empty-directory names in listed prefixes

gargnitingoogle:
GCS supports objects of path-type
 e.g. <bucket>/<path1>//<path2>
 and treats it different from
  <bucket>/<path1>/<path2>.
GCSFuse being a POSIX-compliant file-system
 can only support the latter and
 throws error on former. This error can
 disallow the listing of all directories in the
 parent directory i.e. <path1>.
The current change ignores the listing of
prefixes (directory names) which are empty such
as "/"i above to ignore the error and logs the
above event as a warning.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/storage/bucket_handle.go
```diff
@@ -278,7 +279,11 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 		// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/storage.go#L1304
 		// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L370
 		if attrs.Prefix != "" {
-			list.CollapsedRuns = append(list.CollapsedRuns, attrs.Prefix)
+			if attrs.Prefix != "/" {
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/util/util_test.go
```diff
@@ -278,3 +279,77 @@ func (ts *UtilTest) TestIsolateContextFromParentContext() {
...
+			isUnsupported: true,
+		},
+		{
+			name:          "abc\x00/",
+			isUnsupported: true,
+		},
+	}
+
+	for _, tc := range cases {
+		fmt.Printf("%#v\n", tc)
```
gargnitingoogle, line 307: remove this log

internal/util/util_test.go
```diff
@@ -278,3 +279,77 @@ func (ts *UtilTest) TestIsolateContextFromParentContext() {
...
+			isUnsupported: true,
+		},
+		{
+			name:          "../",
+			isUnsupported: true,
+		},
+	}
+
+	for _, tc := range cases {
+		fmt.Printf("%#v\n", tc)
```
gargnitingoogle, line 352: remove this log

gargnitingoogle:
Presubmit tests passed: [logs](https://fusion2.corp.google.com/ci;prev=s/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/901a5e43-bf82-4bb9-aafa-367d4f7308b3/summary?q=gcsfuse&s=p)

internal/util/util.go
```diff
@@ -25,18 +25,25 @@ import (
...
-const HeapSizeToRssConversionFactor float64 = 2
+	// HeapSizeToRssConversionFactor is a constant factor
+	// which we multiply to the calculated heap-size
+	// to get the corresponding resident set size.
+	HeapSizeToRssConversionFactor float64 = 2
+)
+
+var (
+	UnsupportedObjectNameSubstrings  = []string{"//", "/./", "/../", "\000"}
+	UnsupportedDirectoryNamePrefixes = []string{"/", "./", "../"}
```
ashmeenkaur, line 45: Shouldn't \0 also be here?

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +569,64 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+// Create objects in implicit directories with
+// unsupported names such as ., .., /, \0 and
+// test that stat and ReadDirPicky on the different directories.
+func (t *ImplicitDirsTest) UnsupportedDirNames() {
+	var fi os.FileInfo
+	var entries []os.FileInfo
+	var err error
+
+	// Set up contents.
+	AssertEq(
```
ashmeenkaur, line 582: Let's use stretchr/testify package instead of "github.com/jacobsa/oglematchers" and "github.com/jacobsa/ogletest"

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +569,64 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+	// Statting the mount-directory/a should return a directory entry named "a".
+	fi, err = os.Stat(path.Join(mntDir, "a"))
+	AssertEq(nil, err)
+
+	ExpectEq("a", fi.Name())
+	ExpectTrue(fi.IsDir())
+
+	// ReadDirPicky on mountdir/foo should fail as the unsupported sub-directories in that can not be read.
+	_, err = fusetesting.ReadDirPicky(path.Join(mntDir, "foo"))
+	AssertNe(nil, err)
```
ashmeenkaur, line 618: Isn't readDir supposed to silently ignore unsupported directories?

internal/util/util.go
```diff
@@ -25,18 +25,25 @@ import (
...
-// to get the corresponding resident set size.
-const HeapSizeToRssConversionFactor float64 = 2
+	// HeapSizeToRssConversionFactor is a constant factor
+	// which we multiply to the calculated heap-size
+	// to get the corresponding resident set size.
+	HeapSizeToRssConversionFactor float64 = 2
+)
+
+var (
+	UnsupportedObjectNameSubstrings  = []string{"//", "/./", "/../", "\000"}
```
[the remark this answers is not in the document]
ashmeenkaur: On my linux machine, I am able to create files/directories with \000 or \0 substring

ashmeenkaur (commented):
Still need to take a look at the tests

internal/fs/implicit_dirs_test.go
gargnitingoogle: > Isn't readDir supposed to silently ignore unsupported directories?

ReadDir is, but ReadDirPicky is't.

ashmeenkaur: Can we add another test with WalkDirPath and validate that all the supported directories are present?

internal/util/util.go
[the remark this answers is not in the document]
gargnitingoogle: obsolete comment.

internal/util/util.go
gargnitingoogle: obsolete comment. code removed.

internal/fs/implicit_dirs_test.go
gargnitingoogle: Done. Added a test for walkDir right after this test.

internal/fs/implicit_dirs_test.go
gargnitingoogle: This change will increase the size of this PR a lot . I'll take that up as a separate PR soon after.

internal/storage/fake/testing/bucket_tests.go
```diff
@@ -4458,3 +4458,106 @@ func (t *cancellationTest) ReadObject() {
...
+			t.bucket,
+			[]string{
+				"//",
+				"a/b",
+				"foo//e",
+				"foo/c/d",
+			}))
+
+	for _, input := range []struct {
+		// inputs
```
gargnitingoogle, line 4477: remove this comment

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +574,144 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+	ExpectTrue(fi.IsDir())
+
+	// Statting the mount-directory/a should return a directory entry named "a".
+	fi, err = os.Stat(path.Join(mntDir, "a"))
+	AssertEq(nil, err)
+
+	ExpectEq("a", fi.Name())
+	ExpectTrue(fi.IsDir())
+
+	// ReadDirPicky on mountdir/foo should fail as the unsupported sub-directories in that can not be read.
```
gargnitingoogle, line 617: fix these comment and the other comment later in this same test.

internal/storage/bucket_handle.go
```diff
@@ -321,11 +322,31 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 		// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/storage.go#L1304
 		// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L370
 		if attrs.Prefix != "" {
-			list.CollapsedRuns = append(list.CollapsedRuns, attrs.Prefix)
+			if util.IsUnsupportedObjectName(attrs.Prefix) {
+				// Do not list unsupported directories such as those including
```
gargnitingoogle, line 326: fix this comment here. Make it shorter e.g. 

```
Ignore GCS directories containing `//` in their names as they have a special meaning in linux file-system and can not be represented by the GCSFuse. 
As an example, GCS can have both the objects a//b and a/b at the same time as two different objects, but to linux FS, both paths are the same. So, GCSFuse will ignore object a//b to avoid causing `input/output error` in linux filesystem.
```

internal/storage/bucket_handle.go
```diff
@@ -321,11 +322,31 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
...
+				// by ignoring the GCS objects which have these specially reserved/unsupported unix names/substrings.
+				logger.Warnf("Ignoring unsupported object-prefix: \"%s\"", attrs.Prefix)
+			} else {
+				list.CollapsedRuns = append(list.CollapsedRuns, attrs.Prefix)
+			}
 		} else {
-			// Converting attrs to *Object type.
-			currObject := storageutil.ObjectAttrsToBucketObject(attrs)
-			list.Objects = append(list.Objects, currObject)
+			// For GCS objects, which have the unsupported substring i.e. //
```
gargnitingoogle, line 339: fix this comment here. Make it shorter e.g. 

```
Ignore GCS objects containing `//` in their names as they have a special meaning in linux file-system and can not be represented by the GCSFuse. 
As an example, GCS can have both the objects a//b and a/b at the same time as two different objects, but to linux FS, both paths are the same. So, GCSFuse will ignore object a//b to avoid causing `input/output error` in linux filesystem.
```

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +574,144 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
 	ExpectThat(ctime, timeutil.TimeNear(mountTime, delta))
 	ExpectThat(mtime, timeutil.TimeNear(mountTime, delta))
 }
+
+// Create objects in implicit directories with
+// unsupported names such as ., .., /, \0 and
```
gargnitingoogle, line 579: Fix comment. Remove reference to ./, ../

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +574,144 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+		path:  path.Join(mntDir, "foo/c"),
+		name:  "c",
+		isDir: true,
+	}, {
+		path: path.Join(mntDir, "foo/c/d"),
+		name: "d",
+	},
+	}
+
+	maxIters := 100
```
gargnitingoogle, line 680: Is this really required ?

internal/storage/fake/testing/bucket_tests.go
```diff
@@ -128,9 +128,9 @@ func interestingNames() (names []string) {
 		// Other tricky URL cases.
 		"foo () bar",
 		"foo [] bar",
-		"foo // bar",
+		//"foo // bar",
```
gargnitingoogle, line 131: add comments here to explain why these are commented:

```
Removing object as this is unsupported ignored by GCSFuse.
```

internal/storage/fake/testing/bucket_tests.go
```diff
@@ -4458,3 +4458,106 @@ func (t *cancellationTest) ReadObject() {
 				HasSubstr("request canceled"))))
 	ExpectLt(time.Since(before), 50*time.Millisecond)
 }
+
+func (t *listTest) TestListForUnsupportedNames() {
+	// Create several objects.
```
gargnitingoogle, line 4463: Remove this comment.

tools/integration_tests/implicit_dir/list_test.go
```diff
@@ -37,9 +37,12 @@ func TestListImplicitObjectsFromBucket(t *testing.T) {
 	// testBucket/dirForImplicitDirTests/explicitFile                                                       -- File
 	// testBucket/dirForImplicitDirTests/explicitDirectory/fileInExplicitDir1                               -- File
 	// testBucket/dirForImplicitDirTests/explicitDirectory/fileInExplicitDir2                               -- File
+	// testBucket/dirForImplicitDirTests/..                                                                 -- Dir
```
gargnitingoogle, line 40: fix the comment and the next line. Replace `..` with `/`

tools/integration_tests/util/setup/implicit_and_explicit_dir_setup/implicit_and_explicit_dir_setup.go
```diff
@@ -83,6 +85,15 @@ func CreateImplicitDirectoryStructure(testDir string) {
 	setup.RunScriptForTestData("../util/setup/implicit_and_explicit_dir_setup/testdata/create_objects.sh", path.Join(setup.TestBucket(), testDir))
 }
 
+func CreateUnsupportedImplicitDirectoryStructure(testDir string) {
+	// Unsupported Implicit Directory Structure
+	// testBucket/testDir/../                                                                -- Dir
```
gargnitingoogle, line 90: fix the comment and the next line. Replace `..` with `/`

internal/util/util.go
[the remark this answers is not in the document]
ashmeenkaur: Checking what we concluded here? Is \0 a valid filename substring? What about \0 filename?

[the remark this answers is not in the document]
gargnitingoogle: I've kept `\0` as out of scope of the current set of changes as it's hard to test. Will revisit this question when we add check for that.

tools/integration_tests/implicit_dir/list_test.go
ashmeenkaur, line 1: Can you separate out integration tests into a separate PR. Also if possible, please use storage client instead of gcloud as gcloud doesn't support parallelism and runs into errors when tests are run in parallel.

docs/semantics.md
```diff
@@ -377,3 +377,21 @@ Not all of the usual file system features are supported. Most prominently:
...
+1. Names beginning with `/`, Or
+1. Names containing `//`.
+
+The above objects cannot be accessed
+through Cloud Storage Fuse, as these names have
+special meanings in linux-based filesystems.
+For example, a file/directory with path `a//b`
+is treated the same as `a/b`.
+
+Any objects having such names are invisible to
```
ashmeenkaur, line 396: Are we ignoring a//b like file names or treating them same as a/b?

gargnitingoogle: done

internal/fs/implicit_dirs_test.go
gargnitingoogle: obsolete comment

gargnitingoogle:
Split the integration tests out int a separate PR in https://github.com/GoogleCloudPlatform/gcsfuse/pull/2011 .

ashmeenkaur (commented):
Some Nit comments, LGTM otherwise.

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +574,179 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+	AssertEq(nil, err)
+	AssertNe(nil, fi)
+	ExpectEq("a", fi.Name())
+	ExpectTrue(fi.IsDir())
+
+	// Statting the mount-directory/a//3 should fail as it should be ignored.
+	_, err = os.Stat(path.Join(mntDir, "a/3"))
+	AssertNe(nil, err)
+
+	// Statting the mount-directory/4 should return a file entry named "a".
```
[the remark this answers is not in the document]
gargnitingoogle: good catch!

internal/storage/bucket_handle_test.go
```diff
@@ -615,12 +599,7 @@ func (testSuite *BucketHandleTest) TestListObjectMethodWithZeroMaxResult() {
 	// Validate that all objects (5) are listed when MaxResults is 0. This has
 	// same behavior as not passing MaxResults in request.
 	assert.Nil(testSuite.T(), err2)
-	assert.Equal(testSuite.T(), 5, len(fiveObjWithZeroMaxResults.Objects))
-	assert.Equal(testSuite.T(), TestObjectRootFolderName, fiveObjWithZeroMaxResults.Objects[0].Name)
-	assert.Equal(testSuite.T(), TestObjectSubRootFolderName, fiveObjWithZeroMaxResults.Objects[1].Name)
-	assert.Equal(testSuite.T(), TestSubObjectName, fiveObjWithZeroMaxResults.Objects[2].Name)
-	assert.Equal(testSuite.T(), TestObjectName, fiveObjWithZeroMaxResults.Objects[3].Name)
-	assert.Equal(testSuite.T(), TestGzipObjectName, fiveObjWithZeroMaxResults.Objects[4].Name)
+	assert.ElementsMatch(testSuite.T(), []string{TestObjectRootFolderName, TestObjectSubRootFolderName, TestSubObjectName, TestGzipObjectName, TestObjectName}, objectsToObjectNames(fiveObjWithZeroMaxResults.Objects))
```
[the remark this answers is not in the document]
gargnitingoogle: We can check exact order, but I don't think there is any value in ensuring that the listing returns the exact same order on each run. 
Also, assert.ElementsMatch keeps the code a lot shorter than assert.Equal, and also allows me to add newer objects in the test (which I was originally planning for this change) with minimal changes in the checks, without having to worry about the new order of the objects. So, I am keeping this change as it is.

internal/storage/fake/testing/bucket_tests.go
```diff
@@ -4458,3 +4458,104 @@ func (t *cancellationTest) ReadObject() {
...
+	AssertEq(
+		nil,
+		createEmpty(
+			t.ctx,
+			t.bucket,
+			[]string{
+				"//",
+				"a/b",
+				"foo//e",
+				"foo/c/d",
```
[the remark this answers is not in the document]
gargnitingoogle: thanks @ashmeenkaur. good idea. I've added some of your suggested entries. I've ignored the ones which don't test anything new like //object/a which is already tested by /object/a.

internal/storage/fake/testing/bucket_tests.go
```diff
@@ -128,9 +128,9 @@ func interestingNames() (names []string) {
 		// Other tricky URL cases.
 		"foo () bar",
 		"foo [] bar",
-		"foo // bar",
+		//"foo // bar", // removed as GCSFuse ignores object names containing // while listing
```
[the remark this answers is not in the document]
gargnitingoogle: I didn't want to completely remove these from here as this is not originally our code, and was borrowed from jacobsa/gcloud. But I guess the change will remain in the SCM history, so it should be okay to remove.

internal/util/util_test.go
```diff
@@ -278,3 +278,39 @@ func (ts *UtilTest) TestIsolateContextFromParentContext() {
...
+			isUnsupported: false,
+		},
+		{
+			name:          "abc//",
+			isUnsupported: true,
+		},
+		{
+			name:          "/",
+			isUnsupported: true,
+		},
```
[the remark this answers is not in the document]
gargnitingoogle: thanks. Added /foo. //foo is the same for testing purposes, so not adding that.

tools/integration_tests/implicit_dir/list_test.go
gargnitingoogle: Moved out the integration tests into https://github.com/GoogleCloudPlatform/gcsfuse/pull/2011

internal/storage/bucket_handle.go
```diff
@@ -321,11 +322,29 @@ func (b *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsRequ
 		// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/storage.go#L1304
 		// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L370
 		if attrs.Prefix != "" {
-			list.CollapsedRuns = append(list.CollapsedRuns, attrs.Prefix)
+			if util.IsUnsupportedObjectName(attrs.Prefix) {
```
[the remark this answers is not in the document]
gargnitingoogle: It is not that simple. attrs.Prefix and attrs.Name are not independent and I don't think IsUnsupportedObjectName checks on them can be combined that way. In other words, it is possible for prefix to be empty (which is unsupported) but Name to be proper and vice-versa. The only code duplication if at all, is in (a) the warning log, which is tailored differently for the prefix and the object Name as is needed, and (b) the comment. I've de-duped the comment as much as I could.

The suggestion to pull this bit of code out as a smaller function, could make sense. But I think that'll lead to reduced readability as it's not an obvious functioal block.

internal/storage/fake/bucket.go
```diff
@@ -465,14 +467,45 @@ func (b *bucket) ListObjects(
 	// Find the range of indexes within the array to scan.
 	indexStart := b.objects.lowerBound(nameStart)
 	prefixLimit := b.objects.prefixUpperBound(req.Prefix)
-	indexLimit := minInt(indexStart+maxResults, prefixLimit)
+	var indexLimit int
+	checkForUnsupportedObjectNames := true
+	if maxResults == 1 {
+		// If maxResults is set to 1 (generally for LookUpInode), do not let
```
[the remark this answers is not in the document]
gargnitingoogle: > there won't be any error with MaxResults 1 even if the object is unsupported

This is obvious from the fact that the customer issues (ones having gs://bucket//a/b/c type objects) related to this have never had any mount issues, only listing issues.

> can we add a test for this case

I agree that we could add tests for this. I'd recommend adding a integration test for this.

> It is impossible from GCSFuse POV to get LookUpInode calls for unsupported file names.

It should be impossible because lookups of files/dirs of type a//b are always be resolved to a/b by linux kernel. Same goes for a/./b, a/../b and other types.

> If the above understanding is correct, I think it is okay to keep MaxResults parameter tests out of scope of this PR , and we can use maxResult 5000 for all tests. ( I think it is okay because otherwise our test implementation is varying significantly from actual implementation).

@ashmeenkaur  Let's discuss this offline as it's an involved topic, plus I'm not sure I fully understand what you're saying.

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +573,191 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+	entries, err = fusetesting.ReadDirPicky(path.Join(mntDir, "a"))
+	AssertEq(nil, err)
+	AssertEq(1, len(entries))
+	AssertNe(nil, entries[0])
+	ExpectEq("2", entries[0].Name())
+	ExpectFalse(entries[0].IsDir())
+}
+
+// Create objects in implicit directories with
+// unsupported names such as those having // in them
```
gargnitingoogle, line 678:
```suggestion
// unsupported names (such as those having // in them)
```

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +573,191 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+// and test that stat and ReadDirPicky on the different directories.
```
gargnitingoogle, line 679:
```suggestion
// and test WalkDir on the mountpath.
```

internal/fs/inode/dir.go
```diff
@@ -344,6 +346,36 @@ func findExplicitInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name
 	}, nil
 }
 
+func removeUnsupportedObjectsFromListing(listing *gcs.Listing) *gcs.Listing {
```
gargnitingoogle, line 349: Add unit tests for this one.

internal/storage/bucket_handle_test.go
```diff
@@ -45,6 +45,16 @@ var ContentDisposition string = "ContentDisposition"
 // Hence, we are not writing tests for these flows.
 // https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/github.com/fsouza/fake-gcs-server/fakestorage/object.go#L515
 
+func objectsToObjectNames(objects []*gcs.Object) (objectNames []string) {
```
gargnitingoogle, line 48: Revert all changes in this file.

gargnitingoogle: Or use exact order using assert.Equal .

gargnitingoogle (commented):
added self-review comments.

internal/storage/fake/bucket.go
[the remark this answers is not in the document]
gargnitingoogle: Obsolete comment. Logic to filter out unsupported objects has been moved up to the dirInode.readObjects() method.

internal/fs/inode/dir.go
gargnitingoogle: done

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +573,191 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+	AssertEq(nil, err)
+	verifyFileInfo(fi, "a", true)
+
+	// Statting the mount-directory/a/2 should be recognized as a file.
+	fi, err = os.Stat(path.Join(mntDir, "a/2"))
+	AssertEq(nil, err)
+	verifyFileInfo(fi, "2", false)
+
+	// Statting the mount-directory/a//3 should fail as it should be ignored.
+	_, err = os.Stat(path.Join(mntDir, "a/3"))
```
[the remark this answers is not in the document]
gargnitingoogle: > Kernel will automatically convert it to a/3 ?

Yes. To kernel, "a//3" and "a/3" are the same thing.

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +573,191 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+	verifyFileInfo(fi, "bar", true)
+
+	// ReadDirPicky on mountdir should not fail as the unsupported sub-directories should be ignored.
+	entries, err = fusetesting.ReadDirPicky(mntDir)
+	AssertEq(nil, err)
+	AssertNe(nil, entries)
+	AssertEq(4, len(entries))
+	verifyFileInfo(entries[0], "4", false)
+	verifyFileInfo(entries[1], "a", true)
+	verifyFileInfo(entries[2], "bar", true)
```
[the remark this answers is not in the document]
gargnitingoogle: > Is this expected? There is no valid object having "bar" directory.

Yes. This is the cost of using MaxResults=1 for lookUpInode. 

If we have objects `bar//1`, and `bar/2` in a bucket, gcs list call during lookUpInode with MaxResults=1 will return `bar//1` . If we reject unsupported object `bar//1` at the time of lookup-inode, then `bar` will never be identified as a directory by GCSFuse, and thus `bar/2` will never be accessible through GCSFuse, which is no better than the current bug. Now, even if there is no object named `bar/2`, we can not know from gcs list call with MaxResults=1, and we end up having to identify `bar` as a directory, albeit empty.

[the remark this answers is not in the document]
gargnitingoogle: > I think this will cause issues in case of RemoveDir operation as the object won't be listed during the operation

I'm not sure I understand what you mean by issues. Are you saying that `bar//1` will not be deleted, or that it'll be deleted without being listed ?

I'll test out RemoveDir manually to see what the behaviour is.

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +573,191 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+			}))
+
+	// Statting the mount directory should return a directory entry.
+	fi, err = os.Stat(mntDir)
+	AssertEq(nil, err)
+	AssertNe(nil, fi)
+	ExpectTrue(fi.IsDir())
+
+	// Statting the mount-directory/foo should return a directory entry named "foo".
+	fi, err = os.Stat(path.Join(mntDir, "foo"))
```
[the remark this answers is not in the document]
gargnitingoogle: I'll merge the two tests into one.

internal/fs/inode/dir.go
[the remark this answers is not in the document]
gargnitingoogle: > Shall we add unit test for readObjects method in dir_test.go

I didn't find any existing tests for readObjects in dir_test.go and adding a fresh test for it for this change seemed too much to me.

internal/storage/bucket_handle_test.go
```diff
@@ -440,13 +450,9 @@ func (testSuite *BucketHandleTest) TestListObjectMethodWithPrefixObjectExist() {
 		})
 
 	assert.Nil(testSuite.T(), err)
-	assert.Equal(testSuite.T(), 4, len(obj.Objects))
-	assert.Equal(testSuite.T(), 1, len(obj.CollapsedRuns))
-	assert.Equal(testSuite.T(), TestObjectRootFolderName, obj.Objects[0].Name)
-	assert.Equal(testSuite.T(), TestObjectSubRootFolderName, obj.Objects[1].Name)
-	assert.Equal(testSuite.T(), TestObjectName, obj.Objects[2].Name)
+	assert.Equal(testSuite.T(), []string{TestObjectRootFolderName, TestObjectSubRootFolderName, TestObjectName, TestGzipObjectName}, objectsToObjectNames(obj.Objects))
```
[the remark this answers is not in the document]
gargnitingoogle: > Does assert takes care of panics in case of length mismatch?

Yes. It doesn't panic and handles length mismatches well by gracefully failing on them.

internal/fs/inode/dir.go
```diff
@@ -580,7 +582,24 @@ func (d *dirInode) ReadDescendants(ctx context.Context, limit int) (map[Name]*Co
 			return descendants, nil
 		}
 	}
+}
+
+func logUnsupportedListings(removedListings *gcs.Listing) {
```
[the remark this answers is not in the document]
gargnitingoogle: > shall we log unsupported objects in the util method itself

I thought of doing that too at first, but that'd require me to create a dependency of util package on logger/config, which is a cyclic dependency. To avoid that, I resorted to this alternative.

internal/fs/inode/dir.go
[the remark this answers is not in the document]
gargnitingoogle: Added test in dir_test.go

internal/util/gcs_util_test.go
```diff
@@ -0,0 +1,113 @@
...
+		Objects:           createObjects([]string{"a", "c/d"}),
+		ContinuationToken: "hfdwefo",
+	}
+	expectedRemovedGcsListing := &gcs.Listing{
+		CollapsedRuns: []string{"/", "b//", "e//f/", "g/h//"},
+		Objects:       createObjects([]string{"/b", "e//f", "g/h//i"}),
+	}
+
+	newGcsListing, removedGcsListing := RemoveUnsupportedObjectsFromListing(origGcsListing)
+	assert.Equal(t.T(), *expectedNewGcsListing, *newGcsListing)
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/fs/implicit_dirs_test.go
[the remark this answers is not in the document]
gargnitingoogle: > I'll test out RemoveDir manually to see what the behaviour is.

I tested this out manually, in the above example, if I try and delete/rename `bar` in the gcsfuse mount, it only deletes/renames `bar/2` and has no impact on `bar//1` . This looks like the correct behaviour to me as we intend to have `bar//1` be invisible to a GCSFuse mount.

gargnitingoogle:
This PR is causing a problem with `rm -r` on a nested directory, so marking it as draft until that issue is resolved.

internal/gcsx/bucket_manager.go
```diff
@@ -234,7 +234,7 @@ func (bm *bucketManager) SetUpBucket(
 	}
 
 	// Periodically garbage collect temporary objects
-	go garbageCollect(bm.gcCtx, bm.config.TmpObjectPrefix, sb)
+	//go garbageCollect(bm.gcCtx, bm.config.TmpObjectPrefix, sb)
```
gargnitingoogle, line 237: Undo this.

internal/fs/fs.go
```diff
@@ -1875,6 +1875,8 @@ func (fs *fileSystem) RmDir(
 		return
 	}
 
+	//logger.Debugf("RmDir %q ... ", childDir.Name().GcsObjectName())
```
gargnitingoogle, line 1878: Remove commented out code.

internal/fs/fs.go
```diff
@@ -1906,6 +1923,8 @@ func (fs *fileSystem) RmDir(
 	// We are done with the child.
 	cleanUpAndUnlockChild()
 
+	//logger.Debugf("... RmDir %q", childDir.Name().GcsObjectName())
```
gargnitingoogle, line 1926: Remove commented out code.

internal/fs/fs.go
```diff
@@ -2260,6 +2279,8 @@ func (fs *fileSystem) Unlink(
 		0,   // Latest generation
 		nil) // No meta-generation precondition
 
+	logger.Debugf("Unlinking %q from %q ...", op.Name, parent.Name().GcsObjectName())
```
gargnitingoogle, line 2282: Remove this debug log.

internal/fs/fs.go

internal/fs/fs.go

internal/fs/fs.go

internal/fs/inode/dir_test.go
```diff
@@ -1579,3 +1635,100 @@ func (t *DirTest) Test_InvalidateKernelListCache() {
...
+
+	AssertEq(nil, err)
+
+	b, err := t.in.HasNoSupportedObjectsInSubtree(t.ctx)
+
+	AssertFalse(b)
+	AssertEq(nil, err)
+}
+
+func (t *DirTest) HasNoSupportedObjectsInSubtree_false_scenario2() {
```
gargnitingoogle, line 1710: Remove this test as this is redundant

gargnitingoogle: this is a different scenario, so keeping this test.

internal/fs/fs.go
```diff
@@ -1893,8 +1893,23 @@ func (fs *fileSystem) RmDir(
 
 		// Are there any entries?
 		if len(entries) != 0 {
-			err = fuse.ENOTEMPTY
-			return
+			childDirObjectName := childDir.Name().GcsObjectName()
```
gargnitingoogle, line 1896: Add a comment here explaining what's going on in this block.

internal/fs/inode/dir.go
```diff
@@ -639,20 +651,25 @@ func (d *dirInode) ReadDescendants(ctx context.Context, limit int) (map[Name]*Co
...
+				continue
+			}
 			name := NewDescendantName(d.Name(), o.Name)
 			descendants[name] = &Core{
 				Bucket:    d.Bucket(),
 				FullName:  name,
 				MinObject: storageutil.ConvertObjToMinObject(o),
 			}
 		}
+		logger.Debugf("Ignored unsupported objects: %v", unsupportedObjects)
```
gargnitingoogle, line 666: Switch all these debug logs to warning logs so that they show up even if log-level is set to info/warning.

internal/fs/fs.go
gargnitingoogle: done

internal/fs/inode/dir.go
gargnitingoogle: done

internal/fs/fs.go
```diff
@@ -1893,8 +1893,30 @@ func (fs *fileSystem) RmDir(
...
+			if err != nil {
+				err = fmt.Errorf("DeleteChildDir (%q): %w", childDirObjectName, err)
+				return
+			}
+			if !hasNoSupportedObjectsInSubtree {
+				logger.Warnf("Failed to delete %q as it still has GCS objects in it.", childDirObjectName)
+				err = fuse.ENOTEMPTY
+				return
+			} else {
+				logger.Warnf("Cannot completely delete %q as this directory has unsupported GCS objects (i.e. objects containing // in their names, which cannot be accessed through gcsfuse) in its sub-tree.", childDirObjectName)
```
ashmeenkaur, line 1915: shall we put this on the semantics doc too? Also what if we don't show up only the directory which has all immediate children unsupported?

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +573,109 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
 	ExpectThat(ctime, timeutil.TimeNear(mountTime, delta))
 	ExpectThat(mtime, timeutil.TimeNear(mountTime, delta))
 }
+
+func verifyInvalidPath(path string) {
+	_, err := os.Stat(path)
```
ashmeenkaur, line 578: lets add t.Helper() here for easy debugging

internal/fs/implicit_dirs_test.go
```diff
@@ -569,3 +573,109 @@ func (t *ImplicitDirsTest) AtimeCtimeAndMtime() {
...
+
+	AssertNe(nil, err, "Failed to get error in stat of %q", path)
+}
+
+// Create objects with unsupported object names and
+// verify the behavior of mount using os.Stat and WalkDir.
+func (t *ImplicitDirsTest) UnsupportedGcsObjectNames() {
```
ashmeenkaur, line 585: Shall we add tests for implicit dir false too?

internal/fs/inode/base_dir.go
```diff
@@ -263,3 +263,9 @@ func (d *baseDirInode) IsUnlinked() bool {
 
 func (d *baseDirInode) Unlink() {
 }
+
+// HasNoSupportedObjectsInSubtree checks and returns true if this directory
```
ashmeenkaur, line 267: This comment can just be in the interface. Here we can just say that baseDirInode can't have unsupported objects

internal/fs/inode/dir.go
```diff
@@ -790,6 +819,75 @@ func (d *dirInode) ReadEntries(
...
+
+	// Do a BFS traversal of the GCS prefix corresponding to this directory,
+	// to find out if it has any supported GCS objects
+	// (i.e. GCS objects not containing // in its name) at all.
+	bucket := d.bucket
+	// Go does not have a queue structure in it, so using a slice in its place.
+	var dirNameBfsQueue []string
+	dirNameBfsQueue = append(dirNameBfsQueue, d.Name().GcsObjectName())
+
+	for len(dirNameBfsQueue) != 0 { // if queue is not empty
```
ashmeenkaur, line 843: Do we really need BFS here? Why can't we just traverse on list response with prefix dirInodeObjectName + "/" and see if any object does not have "//"?

docs/semantics.md
```diff
@@ -385,10 +385,10 @@ Instead, when a conflicting pair of foo and ```foo/``` objects both exist, it ap
 
 ### Unsupported object names
 
-Objects in GCS with double slashes '//' as a name or
-prefix are not supported in GCSfuse. Accessing a directory with such
-named files will cause an 'input/output error', as the Linux
-filesystem does not support files or directories named with a '/'.
+Objects in GCS with double slashes '//' in name or
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/fs/inode/dir.go
```diff
@@ -87,6 +88,16 @@ type DirInode interface {
...
 
+	// HasNoSupportedObjectsInSubtree returns true if the GCS prefix corresponding
+	// to this directory contains only objects of the form `<a>//<b>` where <a> and
+	// <b> are proper path strings not containing a `//`.
+	// If there are no GCS objects for this prefix, then it will return false.
+	// Example: Input directory inode: "a/".
+	// If there is even one GCS object like "a/b" or "a/b/c" etc, then it will return false.
+	// If it contains only "a//b" or "a/b//c" or "a//b/c" etc. then it will return true.
+	// Note: This is a recursive method.
+	HasNoSupportedObjectsInSubtree(ctx context.Context) (hasNoSupportedObjects bool, err error)
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/fs/fs.go
```diff
@@ -1893,8 +1893,30 @@ func (fs *fileSystem) RmDir(
...
+			// these objects are ignored by GCSFuse, and thus can
+			// never be deleted by GCSFuse anyway. On the other hand,
+			// if childDir has even one supported object such as b, or a/b, then
+			// it is not empty wrt GCSFuse
+			// and deletion should fail.
+			childDirObjectName := childDir.Name().GcsObjectName()
+			var hasNoSupportedObjectsInSubtree bool
+			hasNoSupportedObjectsInSubtree, err = childDir.HasNoSupportedObjectsInSubtree(ctx)
+			if err != nil {
+				err = fmt.Errorf("DeleteChildDir (%q): %w", childDirObjectName, err)
```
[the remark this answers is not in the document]

internal/fs/implicit_dirs_test.go
gargnitingoogle: done

internal/fs/inode/base_dir.go
gargnitingoogle: done

internal/fs/fs.go
gargnitingoogle: > shall we put this on the semantics doc too?

This seems too complicated to be explained on the the semantics doc. Do you have a suggestion text ?

> Also what if we don't show up only the directory which has all immediate children unsupported?

Not sure I understand this.

internal/fs/inode/dir.go
gargnitingoogle: let's discuss this offline.

PR #1880, opened 2024-05-02
Fix only dir mounting e2e test

Tulsishah:
### Description
Parallel package execution was causing directory mounting tests to fail due to the deletion of directories with the same name.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/local_file/local_file_test.go
+++ b/tools/integration_tests/local_file/local_file_test.go
@@ -34,7 +34,8 @@ import (
 )
 
 const (
-	testDirName = "LocalFileTest"
+	testDirName                 = "LocalFileTest"
+	testDirOnlyDirForLocalFiles = "testDirOnlyDirForLocalFiles"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


tools/integration_tests/local_file/local_file_test.go
```diff
@@ -34,7 +34,8 @@ import (
 )
 
 const (
-	testDirName = "LocalFileTest"
+	testDirName                 = "LocalFileTest"
+	testDirOnlyDirForLocalFiles = "testDirOnlyDirForLocalFiles"
```
ashmeenkaur, line 38: Nit: how about this name? 
```suggestion
	OnlyDirMounted = "OnlyDirMountLocalFiles"
```

Same for other packages also

tools/integration_tests/util/mounting/only_dir_mounting/only_dir_mounting.go
```diff
@@ -84,10 +84,10 @@ func executeTestsForOnlyDirMounting(flags [][]string, m *testing.M) (successCode
 	return
 }
 
-func RunTests(flags [][]string, m *testing.M) (successCode int) {
+func RunTests(flags [][]string, onlyDir string, m *testing.M) (successCode int) {
```
ashmeenkaur, line 87: I think we can just use dirName here

tools/integration_tests/local_file/local_file_test.go
```diff
@@ -34,7 +34,8 @@ import (
 )
 
 const (
-	testDirName = "LocalFileTest"
+	testDirName    = "LocalFileTest"
+	OnlyDirMounted = "OnlyDirMountLocalFiles"
```
ashmeenkaur, line 38: nit: why is this const exported?

ashmeenkaur (approved):

PR #1885, opened 2024-05-04
Printing flags in test logs

Tulsishah:
### Description
Printing flags in test logs.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
+++ b/tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
@@ -60,7 +60,7 @@ func runTestsOnGivenMountedTestBucket(bucketName string, flags [][]string, rootM
 
 		setup.SetMntDir(mntDirOfTestBucket)
 
-		log.Printf("Running tests with flags: %s", flags[i])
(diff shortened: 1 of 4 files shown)


tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
```diff
@@ -60,6 +60,7 @@ func runTestsOnGivenMountedTestBucket(bucketName string, flags [][]string, rootM
 
 		setup.SetMntDir(mntDirOfTestBucket)
 
+		log.Printf("Running tests with flags: %s", flags[i])
```
ashmeenkaur, line 63: Can we add > Running `dynamic mounting` tests with flag.. and similar at other places

Tulsishah: On the top we are already printing the same (e.g. line#119). Should we keep that also then?

ashmeenkaur: I think it will make it easier to search in case of issues.

ashmeenkaur (approved):

PR #1893, opened 2024-05-08
Implement code generation for config parsing.

kislaykishore:
* Create a new package called "cfg" where the generated code will reside. In future, this code will be moved into the config package to complete the migration.
* Add an import in main.go to force compile the new package. Otherwise, we'll not know of any compile errors in the codegen and generated code.
*  This will also enable developers to add new flags to params.yaml thereby reducing divergence with the current setup.

NA

### Testing details
1. Manual - NA
2. Unit tests - we'll add tests for config parsing in a subsequent PR.
3. Integration tests - NA


main.go
```diff
@@ -440,6 +442,7 @@ func handlePanicWhileMounting() {
 	}
 }
 
+//go:generate go run codegen/main.go --inF=codegen/params.yaml --outF=cfg/config.go --tmplF=codegen/config.tpl
```
ashmeenkaur, line 445: What is this comment for? If this is the command for code generation, shall we move it to the appropriate package? Also let's add a comment to clarify what this is used for.

codegen/main.go
```diff
@@ -0,0 +1,461 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
ashmeenkaur, line 1: Is this also a generated file?

main.go
kislaykishore: This comment is special: when the `go generate` command is triggered from the command line, golang searches through the go files for comments of this type i.e. //go:generate <shell_command> and triggers the shell_command. So, this comment is required for generating the config.go.

codegen/main.go
kislaykishore: Nope. I wrote this.

main.go
ashmeenkaur: Got it, let's add a comment explaining this so that we know in future why this was added

go.mod
```diff
@@ -19,9 +19,12 @@ require (
 	github.com/jacobsa/syncutil v0.0.0-20180201203307-228ac8e5a6c3
 	github.com/jacobsa/timeutil v0.0.0-20170205232429-577e5acbbcf6
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0
+	github.com/spf13/pflag v1.0.5
+	github.com/spf13/viper v1.7.0
 	github.com/stretchr/testify v1.9.0
 	github.com/urfave/cli v1.22.14
 	go.opencensus.io v0.24.0
+	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842
```
ashmeenkaur, line 27: Why did we add this package? Don't see it being used.

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	"fmt"
+	"os"
+	"regexp"
+	"slices"
+	"strings"
+	"text/template"
+	"time"
+	"unicode"
+
+	"golang.org/x/exp/maps"
```
ashmeenkaur, line 29: I think we should avoid using experimental package for stability. Ref warning: https://cs.opensource.google/go/x/exp

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+var (
+	outF  = flag.String("outF", "", "Output file")
```
ashmeenkaur, line 35: NIT: Let's make the variable name more descriptive - outFile, inFile, templateFile since these are exposed flags

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	inF   = flag.String("inF", "", "Input file")
+	tmplF = flag.String("tmplF", "", "Template file")
+
+	paramNameRegexCompiled = regexp.MustCompile(`[a-zA-Z][.a-zA-Z0-9\-]*`)
```
ashmeenkaur, line 39: IIUC this is the regex for flag name, thinking if we should call it flagNameRegex or something similar?

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+)
+
+// Represents the fields Config structs.
+type field struct {
+	Name       string
+	DataType   string
+	ConfigPath string
+}
+type typeTemplateData struct {
```
ashmeenkaur, line 48: Shall we call it configTemplateData as everyone is more familiar with config and flag term?

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+var (
+	outF  = flag.String("outF", "", "Output file")
+	inF   = flag.String("inF", "", "Input file")
+	tmplF = flag.String("tmplF", "", "Template file")
+
+	paramNameRegexCompiled = regexp.MustCompile(`[a-zA-Z][.a-zA-Z0-9\-]*`)
+)
+
+// Represents the fields Config structs.
+type field struct {
```
ashmeenkaur, line 43: Let's call this config or something similar?

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	Name       string
+	DataType   string
+	ConfigPath string
+}
+type typeTemplateData struct {
+	// Name of the type
+	Name string
+	// Fields that are to be included in the
```
ashmeenkaur, line 51: We typically begin comments with a capital letter and end them with a full stop.

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	Fields []field
+}
+
+type flagTemplateData struct {
+	Name               string
+	Shorthand          string
```
ashmeenkaur, line 57: Is shorthand used for categorizing flags? 
I think we should add a comment explaining the usage / use a more descriptive name.

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+type templateData struct {
+	TypeTemplateData []typeTemplateData
+	FlagTemplateData []flagTemplateData
+	// Back-ticks are not supported in templates. So, passing as a parameter.
+	BT string
+}
+
+type Param struct {
```
ashmeenkaur, line 76: Any possibility of merging Param and flagTemplateData struct?

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	return typesFieldPairs, nil
+}
+
+func constructTypeTemplateData(paramsConfig []Param) ([]typeTemplateData, error) {
+	// Use map to deduplicate type names and corresponding field names
+	typeInfoMap := make(map[string]map[string]field)
+
+	// For something like logging.log-rotate.compress, the following entries will
+	// be added to the map:
+	// Config -> Logging -> field{Name: Logging, DataType: LoggingConfig, ConfigPath: "logging"}
```
ashmeenkaur, line 187: For consistency
```suggestion
	// Config -> Logging -> field{Name: Logging, DataType: LoggingConfig, ConfigPath: logging}
```

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	// LoggingConfig -> LogRotate -> field{Name: LogRotate, DataType: LogRotateConfig, ConfigPath: log-rotate}
+	// LogRotateConfig -> Compress -> field{Name: Compress, DataType: bool, ConfigPath: compress}
+	for _, p := range paramsConfig {
+		if p.ConfigPath == "" {
```
ashmeenkaur, line 191: Please add a comment that this is to ignore deprecated flags.

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+type typeInfo struct {
+	typeName   string
+	fieldName  string
+	configPath string
+	dataType   string
+}
+
+func getGoDataType(dt string) (string, error) {
```
ashmeenkaur, line 116: This method can never return error I believe

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+		if err != nil {
+			return nil, err
+		}
+		typesFieldPairs = append(typesFieldPairs, typeInfo{
+			typeName:   typeNameSuffix,
+			fieldName:  fldName,
+			dataType:   dataType,
+			configPath: curCfgPath,
+		})
+		if err != nil {
```
ashmeenkaur, line 172: This check is not required i think

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+			return nil, err
+		}
+		typeNameSuffix = dataType
```
ashmeenkaur, line 175: Shouldn't this always be "Config"

For a config-path of x.y.z for an int param it will yield the following structs
//
//	type Config struct {
//	  X XConfig `yaml:"x"`
//	}
//
//	type XConfig struct {
//	  Y YConfig `yaml:"y"`
//	}
//
//	type ZConfig struct {
//	  Z int `yaml:"z"`
//	}

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	// LogRotateConfig -> Compress -> field{Name: Compress, DataType: bool, ConfigPath: compress}
+	for _, p := range paramsConfig {
+		if p.ConfigPath == "" {
+			continue
+		}
+		typeInfo, err := typesToCreate(p)
+		if err != nil {
+			return nil, err
+		}
+		for _, tfp := range typeInfo {
```
ashmeenkaur, line 198:
```suggestion
		for _, ti (for typeInfo?) := range typeInfo {
```

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+//
+//	type XConfig struct {
+//	  Y int `yaml:"y"`
+//	}
+func typesToCreate(param Param) ([]typeInfo, error) {
+	configPath := param.ConfigPath
+	typeNameSuffix := "Config"
+	typesFieldPairs := make([]typeInfo, 0)
+	for configPath != "" {
+		nestedTypeName := ""
```
ashmeenkaur, line 144: Nit: You can alternatively just declare string variables (by default they will be empty string)

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+		panic(err)
+	}
+
+	td, err := constructTypeTemplateData(paramsConfig)
+	if err != nil {
+		panic(err)
+	}
+
+	fd, err := computeFlagTemplateData(paramsConfig)
+
```
ashmeenkaur, line 440: Nit: remove this line and add line after checking error

codegen/main.go
```diff
@@ -0,0 +1,461 @@
...
+	DeprecationWarning string
+	IsHidden           bool
+	IsShorthandHidden  bool
+}
+
+type templateData struct {
+	TypeTemplateData []typeTemplateData
+	FlagTemplateData []flagTemplateData
+	// Back-ticks are not supported in templates. So, passing as a parameter.
+	BT string
```
ashmeenkaur, line 73:
```suggestion
	Backticks string
```

ashmeenkaur (commented):
I've completed a first pass review of the code. I've added some comments, mainly focused on improving the readability of the code. I apologize if there are quite a few – I hope they're helpful!

I also noticed a potential opportunity to refactor `codegen/main.go`. It might be beneficial to create separate utility files for:
- Parsing the input params.yaml file
- Generating type template data

This could help streamline the main file and make the logic easier to follow.

Overall, I found some parts of the code a little challenging to understand at first. Please see if you can do something to make it more intuitive :)

codegen/main.go
kislaykishore: It's config path regex. I've renamed it as such.

codegen/main.go
kislaykishore: config term is very overloaded: it can mean the config file, or the config struct that captures the entire config. Calling a field config will add to confusion.

codegen/main.go
kislaykishore: Shorthand is a well defined term in context of flags. I've added a comment to explain this and some other fields.

codegen/main.go
kislaykishore: Done. Thanks!

codegen/main.go
kislaykishore: No, a config path of x.y.z might conflict with y.z even though they have different hierarchies. To prevent this, we generate the following structs
type Config struct {
	X XConfig `yaml:"x"`

	Y YConfig `yaml:"y"`
}

type XConfig struct {
	Y YXConfig `yaml:"y"`
}

type YConfig struct {
	Z string `yaml:"z"`
}

type YXConfig struct {
	Z string `yaml:"z"`
}

ashmeenkaur (approved):

cfg/config.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	flagSet.BoolP("debug_mutex", "", false, "Print debug messages when a mutex is held too long.")
+
+	err = viper.BindPFlag("debug.log-mutex", flagSet.Lookup("debug_mutex"))
+	if err != nil {
+		return err
+	}
+
+	flagSet.IntP("file-mode", "", 0, "Permissions bits for files, in octal.")
```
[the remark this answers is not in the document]
kislaykishore: Will do that as a subsequent CL. This is an autogenerated code, so can't add a TODO here.

ashmeenkaur (approved):

ashmeenkaur (approved):

PR #1896, opened 2024-05-09
BucketType method to locally store bucketType in variable

Tulsishah:
### Description
Create BucketType() method to returns the type of bucket, allowing the use of a new API specifically designed for HNS buckets.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added

### DIFF
--- a/internal/storage/gcs/bucket.go
+++ b/internal/storage/gcs/bucket.go
@@ -29,7 +29,6 @@ const (
 	Hierarchical
 )
 
-
 // Bucket represents a GCS bucket, pre-bound with a bucket name and necessary
 // authorization information.
 //
--- a/internal/storage/gcs/bucket.go
+++ b/internal/storage/gcs/bucket.go
@@ -31,7 +31,9 @@ import (
 type Bucket interface {
 	Name() string
 
+	// Return Type of bucket e.g. FLAT or HIERARCHICAL
 	BucketType() string
+
 	// Create a reader for the contents of a particular generation of an object.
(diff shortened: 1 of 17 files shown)


internal/storage/gcs/bucket.go
```diff
@@ -20,6 +20,15 @@ import (
 	"golang.org/x/net/context"
 )
 
+// BucketTypes represents different types of buckets.
```
ankitaluthra1, line 23: should be changed to "BucketType represent type of bucket like Hierarchical nonHierarchical as constants"

internal/storage/gcs/bucket.go
```diff
@@ -20,6 +20,15 @@ import (
...
+type BucketTypes int
```
ankitaluthra1, line 24: please rename to BucketTypes as  at one time it can hold only one value, so should not be plural

ankitaluthra1 (dismissed):
just few nits, please fix those, rest looks good to me

internal/storage/gcs/bucket.go
```diff
@@ -20,6 +20,15 @@ import (
 	"golang.org/x/net/context"
 )
 
+// BucketType represent type of bucket like Hierarchical nonHierarchical as constants
```
raj-prince, line 23: nit: -> "BucketType represents different types of buckets like Hierarchical or NonHierarchical as constants."

raj-prince (dismissed):

raj-prince (approved):

PR #1897, opened 2024-05-09
[PR1] Support to keep ReadDir response in kernel cache

raj-prince:
### Description
- Added a configuration under FileSystem - kernel-dir-cache-ttl-secs
   - if 0 - means no kernel caching.
   - if -1 - means cache for life-time
   - Otherwise, kernel will cache the readdir response and invalidated based on the kernel-dir-cache-ttl-secs value.

- Supported cli flag --kernel-dir-cache-ttl-secs, whose value will take precedence over config.
- With this support list response as part of `ls` command will be saved in the kernel and invalidated based on the kernel-dir-cache-ttl-secs.
- Kernel doesn't keep list response partially, either complete response or none.

NA

### Testing details
1. Manual - Done.
2. Unit tests - Added unit test for newly added method in DirInode interface.
3. Integration tests - NA


internal/fs/fs.go
```diff
@@ -2148,6 +2154,13 @@ func (fs *fileSystem) OpenDir(
 	fs.handles[handleID] = handle.NewDirHandle(in, fs.implicitDirs)
 	op.Handle = handleID
 
+	// Invalidates the kernel dir-cache once the last cached response is out of
+	// kernelDirCacheTTL.
+	op.KeepCache = !in.ShouldInvalidateKernelDirCache(fs.kernelDirCacheTTL)
```
sethiay, line 2159: this operation should be guarded by in.Lock() ?

internal/fs/fs.go
```diff
@@ -2148,6 +2154,13 @@ func (fs *fileSystem) OpenDir(
...
+
+	// Enables kernel dir-cache in case of non-zero kernelDirCacheTTL.
+	op.CacheDir = fs.kernelDirCacheTTL > 0
```
sethiay, line 2162: nit: I would prefer it in if loop for readability.

internal/fs/fs.go
sethiay, line 2159: I think we can avoid this check when cache is disabled ?

internal/fs/fs.go
```diff
@@ -342,8 +343,13 @@ type fileSystem struct {
 	enableNonexistentTypeCache bool
 	inodeAttributeCacheTTL     time.Duration
 	dirTypeCacheTTL            time.Duration
-	renameDirLimit             int64
-	sequentialReadSizeMb       int32
+
+	// kernelDirCacheTTL specifies the duration to keep the dir response cached in kernel
+	// after ttl filesystem forces the kernel to evict old cache entries.
```
sethiay, line 348: nit: . After TTL, gcsfuse (filesystem) on next list (readdir) call from user asks the kernel to evict old cache entries

internal/fs/inode/dir.go
```diff
@@ -670,6 +681,9 @@ func (d *dirInode) ReadEntries(
 		}
 		entries = append(entries, entry)
 	}
+
+	tempTime := time.Now()
```
sethiay, line 685: we should name it something like nowTime instead of tempTime.

sethiay, line 685: should we use cacheClock ? that would allow you to write unit tests as well without adding time.Sleep in the code.

internal/config/mount_config.go
```diff
@@ -90,8 +91,9 @@ type EnableHNS bool
 type CacheDir string
 
 type FileSystemConfig struct {
-	IgnoreInterrupts      bool `yaml:"ignore-interrupts"`
-	DisableParallelDirops bool `yaml:"disable-parallel-dirops"`
+	IgnoreInterrupts           bool  `yaml:"ignore-interrupts"`
+	DisableParallelDirops      bool  `yaml:"disable-parallel-dirops"`
+	KernelDirCacheTtlInSeconds int64 `yaml:"kernel-dir-cache-ttl-secs"`
```
sethiay, line 96: I think we should handle the case where if value less than -1 is passed to this flag then we should return error to user.

internal/util/util.go
```diff
@@ -110,3 +115,17 @@ func IsolateContextFromParentContext(ctx context.Context) (context.Context, cont
...
 }
+
+// SecondsToTimeDuration ensure no int64 overflow while converting secs to
+// time.Duration.
+func SecondsToTimeDuration(secs int64) time.Duration {
+	if secs < 0 {
+		return MaxTimeDuration
+	}
+
+	if secs > (math.MaxInt64 / int64(time.Second)) {
```
sethiay, line 126: i think we should not return MaxTimeDuration here and rather handle this case (throw error) in parsing the config itself.

flags_test.go
```diff
@@ -184,6 +185,7 @@ func (t *FlagsTest) DecimalNumbers() {
 		"--stat-cache-capacity=8192",
 		"--max-idle-conns-per-host=100",
 		"--max-conns-per-host=100",
+		"--kernel-dir-cache-ttl-secs=234",
```
gargnitingoogle, line 188: Can we also tests in this file for other cases of kernel-dir-cache-ttl-secs, e.g. -1, <-1, and `kernel-dir-cache-ttl-secs`>`max supported value of kernel-dir-cache-ttl-secs`

gargnitingoogle (commented):
@raj-prince I've reviewed the full code here and added some comments. 

This also needs some integration tests or new configurations added in the existing integration tests. Please add.

gargnitingoogle (changes requested):
Forgot to mark `Request changes` earlier.

internal/fs/inode/dir_test.go
```diff
@@ -1450,3 +1452,40 @@ func (t *DirTest) LocalFileEntriesWithUnlinkedLocalChildFiles() {
...
+
+	shouldInvalidate := t.in.ShouldInvalidateKernelDirCache(ttl)
+
+	AssertEq(true, shouldInvalidate)
+}
+
+func (t *DirTest) Test_ShouldInvalidateKernelDirCache_True() {
+	_, err := t.readAllEntries()
+	AssertEq(nil, err)
+	ttl := time.Millisecond * 10
```
gargnitingoogle, line 1469: nit: Technically this is not right that you're testing with ttl in milliseconds, which cannot even be set with out interface. Is it possible to set this in seconds at least and then speed up time using a simulated-clock?

internal/fs/inode/dir_test.go
```diff
@@ -1450,3 +1452,40 @@ func (t *DirTest) LocalFileEntriesWithUnlinkedLocalChildFiles() {
...
+
+	shouldInvalidate := t.in.ShouldInvalidateKernelDirCache(ttl)
+
+	AssertEq(true, shouldInvalidate)
+}
+
+func (t *DirTest) Test_ShouldInvalidateKernelDirCache_False() {
+	_, err := t.readAllEntries()
+	AssertEq(nil, err)
+	ttl := time.Second
```
gargnitingoogle, line 1482: Please also add a testcase with ttl = 0 .

internal/util/util.go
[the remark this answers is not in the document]
gargnitingoogle: My bad, just refer to https://github.com/GoogleCloudPlatform/gcsfuse/pull/1897/files#r1611718682 .

internal/fs/inode/dir_test.go
gargnitingoogle: How about with ttl = -1, any possibility of a test for that ?

internal/fs/inode/dir.go
gargnitingoogle: +1

internal/fs/inode/dir_test.go
```diff
@@ -1450,3 +1452,40 @@ func (t *DirTest) LocalFileEntriesWithUnlinkedLocalChildFiles() {
...
+	_, err := t.readAllEntries()
+	AssertEq(nil, err)
+	ttl := time.Millisecond * 10
+
+	// Wait for more than ttl.
+	time.Sleep(2 * ttl)
+
+	shouldInvalidate := t.in.ShouldInvalidateKernelDirCache(ttl)
+
+	AssertEq(true, shouldInvalidate)
```
gargnitingoogle, line 1476: Should we have a composite scenario test to show that even though the directory was invalidated past the ttl, but would be cached again (i.e. not invalidated) immediately after another readAllEntries call ?

internal/fs/inode/dir.go
```diff
@@ -852,3 +866,12 @@ func (d *dirInode) LocalFileEntries(localFileInodes map[Name]Inode) (localEntrie
...
 	return
 }
+
+func (d *dirInode) ShouldInvalidateKernelDirCache(ttl time.Duration) bool {
+	// Listing not happened yet, so invalidate.
+	if d.lastDirListingTimeStamp == nil {
+		return true
+	}
+
+	return time.Since(*d.lastDirListingTimeStamp) > ttl
```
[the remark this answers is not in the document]
gargnitingoogle: -1 . I think this is pretty readable.

internal/fs/inode/dir.go
```diff
@@ -186,6 +190,12 @@ type dirInode struct {
 	//
 	// GUARDED_BY(mu)
 	cache metadata.TypeCache
+
+	// lastDirListingTimeStamp is the last time when the kernel asked the directory
+	// listing from the filesystem.
+	// Specially used when kernelDirCacheTTL > 0 that means kernel dir-cache is
+	// enabled.
+	lastDirListingTimeStamp *time.Time
```
gargnitingoogle, line 198: How about `prevDirEntryListingTimestamp` for less ambiguity ?

internal/fs/fs.go
gargnitingoogle: +1

gargnitingoogle: Looks like it or it might be affected by a concurrent ReadDir call. An RLock() should suffice, if we have it. If we don't have it in the DirInode interface, then we can do it within the ShouldInvalidateKernelDirCache method itself.

internal/fs/inode/dir.go
```diff
@@ -237,6 +247,7 @@ func NewDirInode(
 		name:                        name,
 		attrs:                       attrs,
 		cache:                       metadata.NewTypeCache(typeCacheMaxSizeMB, typeCacheTTL),
+		lastDirListingTimeStamp:     nil,
```
gargnitingoogle, line 250: Why is this needed? Isn't nil the default value for a pointer?

gargnitingoogle (changes requested):
Added some more changes

internal/fs/fs.go
```diff
@@ -2148,6 +2154,13 @@ func (fs *fileSystem) OpenDir(
 	fs.handles[handleID] = handle.NewDirHandle(in, fs.implicitDirs)
 	op.Handle = handleID
```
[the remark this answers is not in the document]
raj-prince: I don't think, this is very common or easy to reproduce, considering gcsfuse takes almost 99% time in the first ReadDir() call (7s for 25K files) (since fetches everything from GCS) and next each ReadDir() call takes around 40us to complete. That's why it's hard to give manual interrupt for ReadDir other than the 1st one.

Although I simulated this programmatically and found kernel doesn't cache partial response. Please let me know, this is what you are looking for. Accordingly, I'll update the PR description and internal doc.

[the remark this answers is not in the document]
sethiay: I meant if ls <dir A> is interrupted after 7th ReadDir call (meaning the 8th and beyond ReadDir calls didn't come). In this case, would the first 7 ReadDir calls output (7 * 2K = 24K) be cached in the kernel's page cache ? If yes, then in next ls <dir A>, are the first 7 ReadDir calls served from page cache and 8th call goes to gcsfuse ? (or there's something else that happens?)

flags.go
```diff
@@ -146,6 +146,13 @@ func newApp() (app *cli.App) {
 					"This prevents those signals from immediately terminating gcsfuse inflight operations.",
 			},
 
+			cli.Int64Flag{
+				Name:  config.KernelDirCacheTtlFlagName,
+				Value: 0,
+				Usage: "How long the directory listing (output of ls <dir>) should be cached in the kernel page cache. Cache which is kept " +
```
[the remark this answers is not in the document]
raj-prince: Slight change:

by gcsfuse on next opendir (just before next readdir) call.

internal/fs/inode/dir.go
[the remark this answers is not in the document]
raj-prince: Keeping the same then.

flags.go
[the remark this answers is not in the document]
raj-prince: Not per entry but per directory.

internal/config/config_util.go
```diff
@@ -60,6 +61,15 @@ func OverrideWithAnonymousAccessFlag(c cliContext, mountConfig *MountConfig, ano
 	}
 }
 
+// OverrideKernelDirCacheTtlFlag overwrites the kernel-dir-cache-ttl-secs config
+// with the kernel-dir-cache-ttl-secs flag value if the flag is set.
+func OverrideKernelDirCacheTtlFlag(c cliContext, mountConfig *MountConfig, ttl int64) {
```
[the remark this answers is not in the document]
raj-prince: Yeah, I was also thinking the same. Thinking to take this as tech-debt. I'll raise another PR for this.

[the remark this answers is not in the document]
raj-prince: Instead - `OverrideKernelListCacheTtlConfigFlag`

internal/fs/fs.go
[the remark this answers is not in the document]
raj-prince: Discussed offline:
Conclusion: kernel doesn't cache partial response of a directory.

internal/config/config_util_test.go
```diff
@@ -185,3 +187,72 @@ func TestOverrideWithAnonymousAccessFlag(t *testing.T) {
...
+	}
+}
+
+func Test_ValidListCacheTtlSecsToDuration(t *testing.T) {
+	var testCases = []struct {
+		testName         string
+		ttlInSecs        int64
+		expectedDuration time.Duration
+	}{
+		{"-1", -1, time.Duration(MaxSupportedTtlInSeconds * int64(time.Second))},
```
gargnitingoogle, line 247: nit: Replace `time.Duration(MaxSupportedTtlInSeconds * int64(time.Second))` here with constant MaxSupportedTTL .

flags_test.go
```diff
@@ -75,6 +75,7 @@ func (t *FlagsTest) Defaults() {
 	assert.Equal(t.T(), -1, f.Gid)
 	assert.False(t.T(), f.ImplicitDirs)
 	assert.False(t.T(), f.IgnoreInterrupts)
+	assert.Equal(t.T(), 0, f.KernelListCacheTtlSeconds)
```
gargnitingoogle, line 78: Use `config.DefaultKernelListCacheTtlSeconds` instead of 0.

internal/config/yaml_parser_test.go
```diff
@@ -263,3 +263,31 @@ func (t *YamlParserTest) TestReadConfigFile_FileSystemConfig_UnsetDisableParalle
 	assert.NotNil(t.T(), mountConfig)
 	assert.False(t.T(), mountConfig.FileSystemConfig.DisableParallelDirops)
 }
+
+func (t *YamlParserTest) TestReadConfigFile_FileSystemConfig_InvalidKernelListCacheTtl() {
+	_, err := ParseConfigFile("testdata/file_system_config/invalid_kernel_list_cache_ttl.yaml")
+
+	assert.ErrorContains(t.T(), err, "invalid kernelListCacheTtlSecs: the value of ttl-secs can't be less than -1")
```
gargnitingoogle, line 270: nit: Replace string with config.TtlInSecsInvalidValueError .

internal/config/yaml_parser_test.go
```diff
@@ -263,3 +263,31 @@ func (t *YamlParserTest) TestReadConfigFile_FileSystemConfig_UnsetDisableParalle
...
+}
+
+func (t *YamlParserTest) TestReadConfigFile_FileSystemConfig_UnsupportedLargeKernelListCacheTtl() {
+	_, err := ParseConfigFile("testdata/file_system_config/unsupported_large_kernel_list_cache_ttl.yaml")
+
+	assert.ErrorContains(t.T(), err, "invalid kernelListCacheTtlSecs: the value of ttl-secs is too high to be supported")
```
gargnitingoogle, line 276: nit: Replace string with config.TtlInSecsTooHighError .

internal/fs/inode/dir.go
```diff
@@ -139,6 +139,13 @@ type DirInode interface {
 
 	// UnlockForChildLookup unlocks the lock taken with LockForChildLookup.
 	UnlockForChildLookup()
+
+	// ShouldInvalidateKernelListCache tells the filesystem whether kernel list-cache
+	// should be invalidated or not.
+	ShouldInvalidateKernelListCache(ttl time.Duration) bool
+
+	RLock()
```
gargnitingoogle, line 147: nit: Add comments

```
// Readonly lock 
// Readonly unlock
```

gargnitingoogle (approved):
Minor comments. I don't mind handling them in a future PR to help this PR along.

gargnitingoogle (approved):
LGTM

charith87 (approved):

sethiay (approved):

PR #1898, opened 2024-05-09
Implement method to return bucket type from server

Tulsishah:
### Description
Implemented FetchAndSetBucketType() method to get bucket type from `GetStorageLayout`(e.g. Hierarchical or NonHierarchical) and set it in bucketType.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added(Create mock control client)
3. Integration tests - Added


internal/storage/storage_handle_test.go
```diff
@@ -298,3 +303,10 @@ func (testSuite *StorageHandleTest) TestCreateClientOptionForGRPCClient() {
 	assert.Nil(testSuite.T(), err)
 	assert.NotNil(testSuite.T(), clientOption)
 }
+
+// EnableHNS flag is false so conrtolClient will be nil.
```
raj-prince, line 307: nit: controlClient.

internal/gcsx/bucket_manager.go
```diff
@@ -222,6 +222,9 @@ func (bm *bucketManager) SetUpBucket(
 		bm.config.TmpObjectPrefix,
 		b)
 
+	// Fetch and set bucket type
+	b.FetchAndSetBucketType()
```
raj-prince, line 226: This will require additional permission to the user while mounting, right?

Do we need to update any documentation or playbook?

raj-prince (approved):

internal/storage/bucket_handle.go
```diff
@@ -430,6 +435,35 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
...
+	}
+
+	var callOptions []gax.CallOption
+	stoargeLayout, err := b.controlClient.GetStorageLayout(context.Background(), &controlpb.GetStorageLayoutRequest{
+		Name:      "projects/_/buckets/" + b.bucketName + "/storageLayout",
+		Prefix:    "",
+		RequestId: "",
+	}, callOptions...)
+
+	// In case bucket does not exist, set type unknown instead of panic.
```
raj-prince, line 452: What happens if error comes because of some permission issue?

internal/storage/bucket_handle.go
```diff
@@ -430,6 +435,35 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
...
+	if err != nil {
+		b.bucketType = gcs.Unknown
+		log.Printf("GetStorageLayout: %v", err)
+		return
+	}
+
+	if stoargeLayout.HierarchicalNamespace != nil && stoargeLayout.GetHierarchicalNamespace().Enabled {
```
raj-prince, line 459: It's better to get HierarchicalNamespace via method instead of accessing directly.

```
hierarchicalNamespace := stoargeLayout.GetHierarchicalNamespace()
if hierarchicalNamespace != nil && hierarchicalNamespace.Enabled {
...
}
```

internal/storage/gcs/bucket.go
```diff
@@ -44,6 +45,9 @@ type Bucket interface {
 	// Return Type of bucket e.g. Hierarchical or NonHierarchical
 	BucketType() BucketType
 
+	// Fetch bucket type from GetStorageLayout and set it. e.g. Hierarchical or NonHierarchical
```
raj-prince, line 48: Is it necessary to add a new method in the bucket interface?

Can't we set at the first BucketType() method call, or during initialization of bucket handle?

raj-prince (commented):
Requires discussion.

internal/storage/bucket_handle.go
Tulsishah: `storage.objects.list` permission is require for this API call. This permission is included in `Storage Object Viewer ` role. Which is I think minimum role require to mount the bucket.

If user does not have `Storage Object Viewer `  role then anyway it will fail on very first List call.

internal/storage/gcs/bucket.go
Tulsishah: All the clients and control client method are in bucket_handle and get called by  bucket_handle interface only. So to keep consistency we decided that.

internal/gcsx/bucket_manager.go
```diff
@@ -222,6 +222,9 @@ func (bm *bucketManager) SetUpBucket(
 		bm.config.TmpObjectPrefix,
 		b)
 
+	// Fetch and set bucket type
```
ankitaluthra1, line 225: update to "Fetch bucket type from storage layout api and set bucket type"

internal/storage/bucket_handle.go
```diff
@@ -430,6 +435,36 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
...
+	stoargeLayout, err := b.controlClient.GetStorageLayout(context.Background(), &controlpb.GetStorageLayoutRequest{
+		Name:      "projects/_/buckets/" + b.bucketName + "/storageLayout",
+		Prefix:    "",
+		RequestId: "",
+	}, callOptions...)
+
+	// In case bucket does not exist, set type unknown instead of panic.
+	if err != nil {
+		b.bucketType = gcs.Unknown
+		log.Printf("GetStorageLayout: %v", err)
```
ankitaluthra1, line 455: log should be of error type and update statement to "Error returned from GetStorageLayout: %v", err"

internal/storage/bucket_handle.go
```diff
@@ -430,6 +435,36 @@ func (b *bucketHandle) ComposeObjects(ctx context.Context, req *gcs.ComposeObjec
...
+		return
+	}
+
+	hierarchicalNamespace := stoargeLayout.GetHierarchicalNamespace()
+	if hierarchicalNamespace != nil && hierarchicalNamespace.Enabled {
+		b.bucketType = gcs.Hierarchical
+		return
+	}
+
+	b.bucketType = gcs.Unknown
```
ankitaluthra1, line 465: in case hierarchicalNamespace is nil or hierarchicalNamespace is not Enabled, bucket type should be NonHierarchical not unknown

internal/storage/bucket_handle_test.go
```diff
@@ -1187,3 +1192,26 @@ func (testSuite *BucketHandleTest) TestComposeObjectMethodWithOneSrcObjectIsDstO
...
+func (testSuite *BucketHandleTest) TestFetchBucketTypeWithError() {
+	var x *controlpb.StorageLayout
+	// Test when the client returns an error.
+	testSuite.mockClient.On("GetStorageLayout", mock.Anything, mock.Anything, mock.Anything).
+		Return(x, errors.New("mocked error"))
+
+	bucketType := mockFetchAndSetBucketType(testSuite.mockClient)
+
+	assert.Equal(testSuite.T(), gcs.Unknown, bucketType, "Expected Unknown when there's an error")
+}
```
ankitaluthra1, line 1217: These two tests are testing mockFetchAndSetBucketType, actual FetchAndSetBucketType method which will be used in production code is not being tested here. To validate FetchAndSetBucketType try to mock only client and create new bucket hadnle with embedded mock client and in the test actual FetchAndSetBucketType method should be called. 

Also, two more test cases needs to be added here, one when control client is nil and second when storagelayout does not give any error but hierarchial block is not present in success response.

internal/storage/bucket_handle.go
ankitaluthra1: @raj-prince in that scenario also, bucket type should be set as unknown, and it will be treated as flat bucket

ankitaluthra1: @Tulsishah "If user does not have Storage Object Viewer  role then anyway it will fail on very first List call" --> as I can see in this PR, set type getting called before first list call inside setup bucket ? are we doing list call before setup bucket as well ?

internal/storage/gcs/bucket.go
ankitaluthra1: can we call FetchAndSetBucketType [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/f86b35dd7d21b90d0b41fda56e736a310304e881/internal/gcsx/bucket_manager.go#L167), before b is converted to corresponding bucket, that way only bucket handle needs to have method, thoughts ? @Tulsishah @raj-prince

internal/storage/bucket_handle_test.go
Tulsishah: Done please check.

internal/storage/gcs/bucket.go
Tulsishah: No we can't do that because b type is gcs.Bucket. So it have to part of bucket interface to call it from setUpBucket.

ankitaluthra1: got it, lets keep it in all bucket types then, regarding two alternatives:
set at first buckettype --> this will slow down the first corresponding operation which will trigger bucket type, which is not expected
initialization of bucket handle--> bucket handle is being initialised in storage handle, storage handle is not the correct domain to call storage layout api, setup bucket method should only know how to fetch type and set for the bucket, this behaviour should be abstracted from storage handl, keeping it in storage handle breaks abstraction principle.

ankitaluthra1 (dismissed):
lgtm

internal/storage/gcs/bucket.go
Tulsishah: Had offline discussion to merge method in BucketType() only.

internal/storage/bucket_handle.go
```diff
@@ -27,24 +27,55 @@ import (
...
+	bucketName    string
+	bucketType    gcs.BucketType
+	controlClient StorageControlClientInterface
 }
 
 func (bh *bucketHandle) Name() string {
 	return bh.bucketName
 }
 
 func (bh *bucketHandle) BucketType() gcs.BucketType {
```
raj-prince, line 53: Can we please write a comment here, similar to below:
```During the first BucketType() call, type is inferred by making a GCS call and cached in a local variable. So, that the first call will be a little slower, and further call will be faster since no GCS call.

Also, this is thread safe, but parallelly calling by mutiple threads the first time might result in the more than one GCS call. So, it's good to call it while mounting first to avoid this case.
```




(

internal/storage/fake/bucket.go
```diff
@@ -440,6 +440,10 @@ func (b *bucket) BucketType() gcs.BucketType {
 	return b.bucketType
 }
 
+func (b *bucket) FetchAndSetBucketType() {
```
raj-prince, line 443: We need to revert this change.

internal/storage/control_client_wrapper.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"context"
+
+	control "cloud.google.com/go/storage/control/apiv2"
+	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googleapis/gax-go/v2"
+)
+
+// Define the interface
```
raj-prince, line 25: I don't see this comment has a lot of value here.

Can we be specific about this interface, if not clearly visible by the name?

internal/storage/control_client_wrapper.go
```diff
@@ -0,0 +1,38 @@
...
+type StorageControlClient interface {
+	GetStorageLayout(ctx context.Context,
+		req *controlpb.GetStorageLayoutRequest,
+		opts ...gax.CallOption) (*controlpb.StorageLayout, error)
+}
+
+type StorageControlClientWrapper struct {
```
raj-prince, line 32: I don't see any use-case of this wrapper class. If not used, can we revert this?

internal/storage/control_client_wrapper.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+
+	control "cloud.google.com/go/storage/control/apiv2"
+	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googleapis/gax-go/v2"
+)
+
+// Define the interface
+type StorageControlClient interface {
```
raj-prince, line 26: Also, could you please check with go-sdk team? Ideally, it would be better to keep this interface go-sdk side.

internal/storage/mock_control_client.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+
+	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googleapis/gax-go/v2"
+	"github.com/stretchr/testify/mock"
+)
+
+// MockStorageControlClient creates a mock version of the StorageControlClient.
+type MockStorageControlClient struct {
```
raj-prince, line 26: should we keep `StorageControlClient` inside this struct? Just to make sure, this mock is implementing all the methods of `StorageControlClient`.

internal/storage/control_client_wrapper.go
Tulsishah: Sure will check with them

internal/storage/control_client_wrapper.go
Tulsishah: Done

raj-prince (commented):
Went through with the interface level change. Added a couple of comments.

internal/storage/gcs/bucket.go
```diff
@@ -26,8 +26,10 @@ type BucketType int
 
 // BucketType enum values.
 const (
-	NonHierarchical BucketType = iota
+	Nil BucketType = iota
```
ankitaluthra1, line 29: please add a comment here Default value Nil represents that bucket type is not set

ankitaluthra1 (dismissed):
lgtm

raj-prince (approved):
Relying on Ankita's review for unit test changes.

PR #1906, opened 2024-05-13
[PR2] Composite tests for Parallel dirops test.

sethiay:
### Description
Composite tests for Parallel dirops test.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - [Run1](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/095b660f-5941-4778-9985-b1e9b23eaab0)

### DIFF
--- a/internal/fs/parallel_dirops_test.go
+++ b/internal/fs/parallel_dirops_test.go
@@ -1,0 +1,687 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// A collection of tests for a file system where parallel dirops are allowed.
+// Dirops refers to readdir and lookup operations.
+
+package fs_test
+
+import (
+	"os"
+	"path"
+	"sync"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type ParallelDiropsTest struct {
+	suite.Suite
+	fsTest
+}
+
+type ParallelDiropsWithoutCachesTest struct {
+	ParallelDiropsTest
+}
+
+// createFilesAndDirStructureInBucket creates the following files and directory
+// structure.
+// bucket
+//
+//			file1.txt
+//			file2.txt
+//			explicitDir1
+//					file1.txt
+//					file2.txt
+//	    implicitDir1
+//					file1.txt
+func (t *ParallelDiropsTest) createFilesAndDirStructureInBucket() {
+	assert.Nil(
+		t.T(),
+		t.createObjects(
+			map[string]string{
+				"file1.txt":              "abcdef",
+				"file2.txt":              "xyz",
+				"explicitDir1/":          "",
+				"explicitDir1/file1.txt": "12345",
+				"explicitDir1/file2.txt": "6789101112",
+				"implicitDir1/file1.txt": "-1234556789",
+			}))
+}
+
+func (t *ParallelDiropsTest) SetupSuite() {
+	t.serverCfg.ImplicitDirectories = true
+	t.serverCfg.MountConfig = &config.MountConfig{
+		FileSystemConfig: config.FileSystemConfig{
+			DisableParallelDirops: false,
+		}}
+	t.serverCfg.RenameDirLimit = 10
+	t.fsTest.SetUpTestSuite()
+}
+
+func (t *ParallelDiropsWithoutCachesTest) SetupSuite() {
+	t.serverCfg.ImplicitDirectories = true
+	t.serverCfg.MountConfig = &config.MountConfig{
+		FileSystemConfig: config.FileSystemConfig{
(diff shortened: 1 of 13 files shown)


internal/fs/parallel_dirops_test.go
```diff
@@ -0,0 +1,687 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
raj-prince, line 1: nit: 2024

internal/fs/parallel_dirops_test.go
```diff
@@ -0,0 +1,687 @@
...
+}
+
+type ParallelDiropsWithoutCachesTest struct {
+	ParallelDiropsTest
+}
+
+// createFilesAndDirStructureInBucket creates the following files and directory
+// structure.
+// bucket
+//
```
raj-prince, line 43: Flat structure seems easy to understand.

internal/fs/parallel_dirops_test.go
```diff
@@ -0,0 +1,687 @@
...
+	go func() {
+		stat1, err1 = lookUpFunc(&wg, filePath)
+	}()
+	go func() {
+		stat2, err2 = lookUpFunc(&wg, filePath)
+	}()
+	wg.Wait()
+
+	// Assert both stats passed and give correct information
+	assert.NoError(t.T(), err1)
```
raj-prince, line 121: Should we create a local function for all these checks and that will be used below?

internal/fs/parallel_dirops_test.go
```diff
@@ -0,0 +1,687 @@
...
+	lookUpFunc := func(wg *sync.WaitGroup, dirPath string) (os.FileInfo, error) {
+		defer wg.Done()
+		fileInfo, err := os.Stat(dirPath)
+		return fileInfo, err
+	}
+	var stat1, stat2 os.FileInfo
+	var err1, err2 error
+
+	// Parallel lookups of explicit dir under mount.
+	dirPath := path.Join(mntDir, "explicitDir1")
```
raj-prince, line 158: Similarly seems replication for explicitDir1, we can write a common method which will be used by both implicitdir and explicit dir.

internal/fs/parallel_dirops_test.go
sethiay: Trying to follow DAMP in unit tests: https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests. 

I know it's a not a good idea to always use DAMP but I think we can use DAMP here without making too much repetitions. LMK if you think otherwise and I will make changes.

Same reason for the second similar comment.

raj-prince: Got it. 

Make sense to use DAMP here, since not much repetitions, and readable test.

internal/fs/parallel_dirops_test.go
raj-prince: Resolving. see the reasoning: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1906#discussion_r1606294076

internal/fs/parallel_dirops_test.go
```diff
@@ -0,0 +1,687 @@
...
+	wg.Wait()
+
+	// Assert both readDirs passed and give correct information
+	assert.NoError(t.T(), err1)
+	assert.NoError(t.T(), err2)
+	assert.Contains(t.T(), dirEntries1[0].Name(), "file1.txt")
+	assert.Contains(t.T(), dirEntries2[0].Name(), "file1.txt")
+
+	// Parallel readDirs of a directory and its parent directory.
+	dirPath = path.Join(mntDir, "explicitDir1")
```
raj-prince, line 245: This test is contradicting the name of the test. Either we can change the name or move the test to another test.

internal/fs/parallel_dirops_test.go
```diff
@@ -0,0 +1,687 @@
...
+		defer wg.Done()
+		file, err := os.Create(filePath)
+		return file, err
+	}
+	var fileInfo os.FileInfo
+	var lookUpErr, createErr error
+	var file *os.File
+
+	// Parallel lookup and creation of a file.
+	filePath := path.Join(mntDir, "explicitDir1", "file3.txt")
```
raj-prince, line 346: Any reason to not add similar test for implicit-dir?

internal/fs/parallel_dirops_test.go
```diff
@@ -0,0 +1,687 @@
...
+	assert.Contains(t.T(), dirEntries1[0].Name(), "file1.txt")
+	assert.Contains(t.T(), dirEntries1[1].Name(), "file2.txt")
+	assert.Contains(t.T(), dirEntries2[0].Name(), "file1.txt")
+}
+
+func TestParallelDiropsTestSuite(t *testing.T) {
+	suite.Run(t, new(ParallelDiropsTest))
+}
+
+func TestParallelDiropsWithoutCachesTestSuite(t *testing.T) {
```
raj-prince, line 685: Can we keep all the constructs of `WithoutCachesTest` together? I mean, Test struct, SetupTestSuites etc.

raj-prince (commented):
LGTM, a could of minor comments.

internal/fs/parallel_dirops_test.go
sethiay: IIUC, you mean why not add test for creation and lookup of a file inside implicitDir ? If yes, then I think there is no reason, I just wanted to add a test for creation and lookup of a file inside directory and just chose that directory to be explicit.

Do you see any value of adding test with implicitDir ? if yes then lmk and we can add.

raj-prince: Okay.
Flagged since, considered implicit-dir too at other place e.g. `TestParallelLookUpsForSameDir`.

But I agree, this will not add much value and explicit-dir test should be enough. So, we can skip it.

raj-prince (approved):

raj-prince (approved):

PR #1910, opened 2024-05-14
upgrade storage sdk to v1.41.0

ashmeenkaur:
### Description
upgrade storage SDK to v1.41.0

NA

### Testing details
1. Manual - Manually validated that retries are happening for connection reset via storage test bench server.
2. Unit tests - Updated
4. Perf tests - 
5W85kHwxwfqmEiX


internal/storage/storageutil/custom_retry_test.go
```diff
@@ -48,37 +41,60 @@ func (t customRetryTest) ShouldRetryReturnsTrueWithGoogleApiError() {
...
+			expectedResult: true,
+		},
+		{
+			name:           "URL Error - Connection Reset",
+			err:            &url.Error{Err: errors.New("connection reset")},
+			expectedResult: true,
+		},
+		{
+			name:           "Op Error - Connection Refused",
+			err:            &net.OpError{Err: errors.New("connection refused")},
```
sethiay, line 84: let's write a case with error message in capitals. e.g. "Connection Refused"

ashmeenkaur: The tests would fail in that case. In general, error strings are not capitalised as per the golang style guide: [Ref](https://go.dev/wiki/CodeReviewComments#:~:text=Error%20strings%20should%20not%20be,usually%20printed%20following%20other%20context.)

sethiay: in that let's expand the error message by say "connection refused by peer"

sethiay (approved):

PR #1918, opened 2024-05-14
Moving auth and config packages to stretchr/testify

sethiay:
### Description
Moving auth and config packages to stretchr/testify

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/config/yaml_parser_test.go
+++ b/internal/config/yaml_parser_test.go
@@ -45,7 +45,7 @@ func validateDefaultConfig(t *testing.T, mountConfig *MountConfig) {
 	assert.False(t, mountConfig.FileCacheConfig.CacheFileForRangeRead)
 	assert.Equal(t, 1, mountConfig.GrpcClientConfig.ConnPoolSize)
 	assert.False(t, mountConfig.AuthConfig.AnonymousAccess)
-	assert.Equal(t, false, bool(mountConfig.EnableHNS))
+	assert.False(t, bool(mountConfig.EnableHNS))
 	assert.False(t, mountConfig.FileSystemConfig.IgnoreInterrupts)
 }
 
@@ -122,7 +122,7 @@ func (t *YamlParserTest) TestReadConfigFile_ValidConfig() {
 	assert.True(t.T(), mountConfig.AuthConfig.AnonymousAccess)
 
 	// enable-hns
-	assert.Equal(t.T(), true, bool(mountConfig.EnableHNS))
+	assert.True(t.T(), bool(mountConfig.EnableHNS))
 
 	// file-system config
 	assert.True(t.T(), mountConfig.FileSystemConfig.IgnoreInterrupts)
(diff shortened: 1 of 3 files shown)


internal/config/yaml_parser_test.go
```diff
@@ -16,253 +16,234 @@ package config
...
+	assert.ErrorContains(t.T(), err, "error parsing config file: yaml: unmarshal errors:\n  line 2: cannot unmarshal !!str `abc` into bool")
 }
 
 func (t *YamlParserTest) TestReadConfigFile_FileSystemConfig_UnsetIgnoreInterruptsValue() {
 	mountConfig, err := ParseConfigFile("testdata/file_system_config/unset_ignore_interrupts.yaml")
 
-	AssertEq(nil, err)
-	AssertNe(nil, mountConfig)
-	AssertEq(false, mountConfig.FileSystemConfig.IgnoreInterrupts)
+	assert.NoError(t.T(), err)
```
Tulsishah, line 232: Not sure but other places I used assert.Nil. Should we maintain consistency ?

sethiay: I think both are correct only but assert.NotError is better because with assert.Nil our intention is to check it's not an error. So, you can keep the assert.Nil but going forward you can use assert.NotError.

internal/config/yaml_parser_test.go
```diff
@@ -16,253 +16,234 @@ package config
...
+	assert.ErrorContains(t.T(), err, fmt.Sprintf(parseConfigFileErrMsgFormat, "log severity should be one of [trace, debug, info, warning, error, off]"))
 }
 
 func (t *YamlParserTest) TestReadConfigFile_InvalidLogRotateConfig1() {
 	_, err := ParseConfigFile("testdata/invalid_log_rotate_config_1.yaml")
 
-	AssertNe(nil, err)
-	AssertTrue(strings.Contains(err.Error(),
-		fmt.Sprintf(parseConfigFileErrMsgFormat, "max-file-size-mb should be atleast 1")))
+	assert.ErrorContains(t.T(), err, fmt.Sprintf(parseConfigFileErrMsgFormat, "max-file-size-mb should be atleast 1"))
```
Tulsishah, line 140: Nit: You have corrected above (t, expected, actual) order. If so can you do it the same here also?

sethiay: The above order is correct for ErrorContains: https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains

internal/config/yaml_parser_test.go
```diff
@@ -16,253 +16,234 @@ package config
...
 
 func (t *YamlParserTest) TestReadConfigFile_FileSystemConfig_UnsetAnonymousAccessValue() {
 	mountConfig, err := ParseConfigFile("testdata/auth_config/unset_anonymous_access.yaml")
 
-	AssertEq(nil, err)
-	AssertNe(nil, mountConfig)
-	AssertEq(false, mountConfig.AuthConfig.AnonymousAccess)
+	assert.NoError(t.T(), err)
+	assert.NotNil(t.T(), mountConfig)
+	assert.Equal(t.T(), false, mountConfig.AuthConfig.AnonymousAccess)
```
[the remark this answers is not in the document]
sethiay: done.

Tulsishah (approved):

raj-prince (approved):

PR #1919, opened 2024-05-14
Moving logger, flag & perms package tests to stretchr testify

sethiay:
### Description
Moving logger, flag & perms package tests to stretchr testify

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/logger/logger_test.go
+++ b/internal/logger/logger_test.go
@@ -22,7 +22,8 @@ import (
 	"testing"
 
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
-	. "github.com/jacobsa/ogletest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/logger/logger_test.go
```diff
@@ -125,7 +127,7 @@ func (t *LoggerTest) TestTextFormatLogs_LogLevelOFF() {
 	}
 
 	// Assert that nothing is logged when log level is OFF.
-	validateLogOutputAtSpecifiedFormatAndSeverity("json", config.OFF, expected)
+	validateLogOutputAtSpecifiedFormatAndSeverity(t.T(), "json", config.OFF, expected)
```
Tulsishah, line 130: Shouldn't we call assertion directly instead of function call?

ref: https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests

sethiay: This PR is to migrate from using jacobsa test libraries for the given packages. Yes, we will be working towards going DAMP in unit tests in future.

Tulsishah (approved):

raj-prince (approved):
Relying on Tulsi's review.

PR #1921, opened 2024-05-14
Implement code coverage checks

kislaykishore:
This change integrates codecov.io into gcsfuse which would keep track of coverage stats and fail presubmit checks if the code-coverage drops beyond a certain target. For code in master, we've set a minimum value of 60% and for patches, the threshold is 80%.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,30 @@
+name: Code Coverage
+on: [push, pull_request]
+jobs:
+  codecov:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/codecov.yml
```diff
@@ -0,0 +1,30 @@
+name: Code Coverage
+on: [push, pull_request]
+jobs:
+  codecov:
```
raj-prince, line 4: Should we add time-out?

raj-prince:
I can see, build are passing still few test packages have coverage in single digit. Is it because of threshold = 0?

What does threshold and target signify?

raj-prince (approved):

PR #1924, opened 2024-05-15
Fix golangci-lint issues and delete unused code.

kislaykishore:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/foreign_modifications_test.go
+++ b/internal/fs/foreign_modifications_test.go
@@ -22,7 +22,6 @@ import (
 	"bytes"
 	"context"
 	"io"
-	"io/ioutil"
 	"math/rand"
 	"os"
 	"path"
@@ -500,7 +499,7 @@ func (t *ForeignModsTest) ReadFromFile_Small() {
 	defer func() { AssertEq(nil, f.Close()) }()
 
 	// Read its entire contents.
-	slice, err := ioutil.ReadAll(f)
+	slice, err := io.ReadAll(f)
 	AssertEq(nil, err)
 	ExpectEq("tacoburritoenchilada", string(slice))
 
@@ -602,7 +601,7 @@ func (t *ForeignModsTest) ReadBeyondEndOfFile() {
 	AssertEq(contents[contentLen-1], buf[0])
 
 	if err == nil {
-		n, err = f.Read(buf)
+		n, _ = f.Read(buf)
 		AssertEq(0, n)
 	}
 }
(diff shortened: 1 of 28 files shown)


internal/fs/foreign_modifications_test.go
```diff
@@ -602,7 +601,7 @@ func (t *ForeignModsTest) ReadBeyondEndOfFile() {
 	AssertEq(contents[contentLen-1], buf[0])
 
 	if err == nil {
-		n, err = f.Read(buf)
+		n, _ = f.Read(buf)
```
Tulsishah, line 604: Instead of this should we put Assert(nil, err) ?

internal/gcsx/append_object_creator_test.go
```diff
@@ -240,7 +240,7 @@ func (t *AppendObjectCreatorTest) CallsComposeObjectsWithObjectProperties() {
 		WillOnce(Return(nil))
 
 	// Call
-	t.call()
+	_, _ = t.call()
```
Tulsishah, line 243: Should we capture err and assert?

internal/gcsx/temp_file.go
```diff
@@ -332,15 +332,15 @@ const (
 func (tf *tempFile) ensure(limit int64) error {
 	switch tf.state {
 	case fileIncomplete:
-		size, err := tf.f.Seek(0, 2)
+		size, _ := tf.f.Seek(0, 2)
```
Tulsishah, line 335: Should we return err instead of ignoring?

Tulsishah (commented):
Can we capture err and handle it accordingly wherever we ignore it through _ ?

internal/fs/foreign_modifications_test.go
kislaykishore: I tried doing that. Seems like these errors have been deliberately ignored. Adding such asserts causes the tests to fail.

internal/gcsx/append_object_creator_test.go
kislaykishore: Asserting for error not being nil causes tests to fail.

internal/gcsx/temp_file.go
kislaykishore: In certain io methods, an EOF is returned as error. So, an error might not always imply a problem. Therefore I'm a bit scared changing this logic as it might lead to behavior changes. WDYT?

internal/gcsx/append_object_creator_test.go
Tulsishah: Is err not nil?

kislaykishore: No, it isn't

PR #1927, opened 2024-05-17
DontReview Call ls -R from within mount

gargnitingoogle:
This call runs `ls -R` on mount-directory,
right after linux mount, and blocks
returning after mount completion to the user
until the `ls -R` completes.
The `ls -R` is simulated with filepath.WalkDir
which recursively visits each sub-directory
and takes count of all files/directories.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,10 +21,12 @@ package main
 
 import (
 	"fmt"
+	"io/fs"
 	"log"
 	"os"
 	"os/signal"
 	"path"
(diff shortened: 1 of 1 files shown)


gargnitingoogle:
https://github.com/GoogleCloudPlatform/gcsfuse/pull/1930

PR #1928, opened 2024-05-17
DontReview Call ls -R from within mount

gargnitingoogle:
This call runs `ls -R` on mount-directory,
right after returning successful mount to user.
This initiates the `ls -R` in a separate
goroutine. On completion of this goroutine,
if there is an error, then it is logged
in gcsfuse debug log-file.
The `ls -R` is simulated with filepath.WalkDir
which recursively visits each sub-directory
and takes count of all files/directories.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,10 +21,12 @@ package main
 
 import (
 	"fmt"
+	"io/fs"
 	"log"
 	"os"
 	"os/signal"
 	"path"
+	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


gargnitingoogle:
Closed in favour of https://github.com/GoogleCloudPlatform/gcsfuse/pull/1930 .

PR #1930, opened 2024-05-20
Metadata-prefetch (aka Recursive listing or ls -R during mount)

gargnitingoogle:
### Description
Metadata-prefetch (aka Recursive listing or ls -R during mount)

Metadata-prefetch is equivalent to running `ls -R`
on the mount-directory to prime/prefill/prefetch the
metadata caches (both type and stat).

This change adds support for mountconfig-flag `metadata-prefetch-on-mount`
in config-file. The type of value passed is a string.
The supported values are: (a) disabled, (b) sync, (c)
async.
If not set, the value `disabled` is taken by default.

If value `disabled` is set, then no difference from the existing behaviour. 

If value `sync` is set, then mount is blocked
until the metadata prefetch is complete. In this mode, if
the metadata-prefetch fails, the mount itself fails.

If value is set to `async`, then mount is not
blocked, but metadata prefetch is
initiated in a go-routine and will complete sometime
later. In this case, if the metadata-prefetch fails,
then an error is logged in the GCSFuse log-file.

NA

### Testing details
1. Manual - Done 
   - Logs with `metadata-prefetch-on-mount: disabled` or with `metadata-prefetch-on-mount` not set:
      ```shell
       message="Mounting file system \"gargnitin-fuse-test-bucket1\"..."
       message="File system has been successfully mounted."
      ```
   - Logs with `metadata-prefetch-on-mount: sync`:
      ```shell
       message="Mounting file system \"gargnitin-fuse-test-bucket1\"..."
       message="Started recursive metadata-prefetch of directory:  \"blah/blah/blah\" ..."
       message="... Completed recursive metadata-prefetch of directory: \"blah/blah/blah\". Number of items discovered: 61"
       message="File system has been successfully mounted."
      ```
   - Logs with `metadata-prefetch-on-mount: async`:
      ```shell
       message="Mounting file system \"gargnitin-fuse-test-bucket1\"..."
       message="File system has been successfully mounted."
       message="Started recursive metadata-prefetch of directory:  \"blah/blah/blah\" ..."
      ```
     Here, control comes back to the user. Later log, after some time,
     ```shell
       message="... Completed recursive metadata-prefetch of directory: \"blah/blah/blah\". Number of items discovered: 61"
      ```
3. Unit tests - Ran locally and all passed.
4. Integration tests - Ran through presubmit.


tools/integration_tests/operations/operations_test.go
```diff
@@ -119,6 +119,28 @@ func createMountConfigsAndEquivalentFlags() (flags [][]string) {
...
+	// Set up config file testing metadata-prefetch-mode: "synchronous".
+	mountConfig4 := config.MountConfig{
+		LogConfig: config.LogConfig{
+			Severity:        config.TRACE,
+			LogRotateConfig: config.DefaultLogRotateConfig(),
+		},
+		MetadataPrefetchMode: config.MetadataPrefetchModeSynchronous,
+	}
+	filePath4 := setup.YAMLConfigFile(mountConfig4, "config4.yaml")
+	flags = append(flags, []string{"--config-file=" + filePath4})
```
gargnitingoogle, line 142: One of these two configurations should be removed. One is enough.

gargnitingoogle:
Integration tests are currently running. I'll add result for that once they finish.

gargnitingoogle:
> Integration tests are currently running. I'll add result for that once they finish.

Tests have passed. [logs](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/2bc2f177-cb4b-4570-af51-d4e5905cb2a7/summary)

internal/config/mount_config.go
```diff
@@ -19,6 +19,10 @@ import (
 	"time"
 )
 
+// Special type declared to stop any random string
+// from being set to MetadataPrefetchMode.
+type metadataPrefetchModeValue string
```
kislaykishore, line 24: How about using string as the type instead of creating a new type and add validation to return error if it's not in the specified list?

internal/config/mount_config.go
```diff
@@ -50,6 +54,15 @@ const (
 	DefaultGrpcConnPoolSize                       = 1
 	DefaultAnonymousAccess                        = false
 	DefaultEnableHNS                              = false
+
+	// MetadataPrefetchModeDisabled is the mode without metadata-prefetch.
+	MetadataPrefetchModeDisabled metadataPrefetchModeValue = "disabled"
```
kislaykishore, line 59: Should this be a child element if metadata-cache? IIUC, the effort is to pre-populate the metadata cache.

internal/config/mount_config.go
```diff
@@ -130,6 +143,13 @@ type MountConfig struct {
 	AuthConfig          `yaml:"auth-config"`
 	EnableHNS           `yaml:"enable-hns"`
 	FileSystemConfig    `yaml:"file-system"`
+
+	// MetadataPrefetchMode indicates whether or not to prefetch the metadata of the mounted bucket at the time of mounting the bucket.
+	// Supported values: MetadataPrefetchModeDisabled, MetadataPrefetchModeSynchronous, and MetadataPrefetchModeAsynchronous.
+	// Any other values will return error on mounting.
+	// This is applicable only to single-bucket mount-points, and not to dynamic-mount points. This is because dynamic-mounts don't mount the bucket(s) at the time of
+	// gcsfuse command itself, which flag is targeted at.
+	MetadataPrefetchMode metadataPrefetchModeValue `yaml:"metadata-prefetch-mode"`
```
kislaykishore, line 152: Should this field be a member of MetadataCacheConfig type since this is about prepopulating the metadata-cache?

main.go
```diff
@@ -207,6 +209,28 @@ func populateArgs(c *cli.Context) (
 	return
 }
 
+func callListRecursive(mountPoint string) (err error) {
+	logger.Debugf("Started recursive metadata-prefetch of directory: \"%s\" ...", mountPoint)
+	numItems := 0
+	err = filepath.WalkDir(mountPoint, func(path string, d fs.DirEntry, err error) error {
+		//logger.Infof("Walked path:%s, d=%s, isDir=%v", path, d.Name(), d.IsDir())
```
kislaykishore, line 216: Uncomment this and log at debug?

main.go
```diff
@@ -207,6 +209,28 @@ func populateArgs(c *cli.Context) (
...
+			}
+		}
+
+		numItems++
+		return err
+	})
+
+	logger.Debugf("... Completed recursive metadata-prefetch of directory: \"%s\". Number of items discovered: %v", mountPoint, numItems)
+
+	return
```
kislaykishore, line 231: Naked returns harm readability. Please fix.

main.go
```diff
@@ -207,6 +209,28 @@ func populateArgs(c *cli.Context) (
...
+		if err != nil {
+			if d == nil {
+				return fmt.Errorf("got error walking: path=\"%s\" does not exist, error = %w", path, err)
+			} else {
+				return fmt.Errorf("got error walking: path=\"%s\", dentry=\"%s\", isDir=%v, error = %w", path, d.Name(), d.IsDir(), err)
+			}
+		}
+
+		numItems++
+		return err
```
kislaykishore, line 226: Should we return nil here?

main.go
```diff
@@ -364,7 +390,27 @@ func runCLIApp(c *cli.Context) (err error) {
...
-			logger.Infof(SuccessfulMountMessage)
+			if isDynamicMount {
+				logger.Infof(SuccessfulMountMessage)
+			} else {
+				switch mountConfig.MetadataPrefetchMode {
+				case config.MetadataPrefetchModeDisabled:
+					logger.Infof(SuccessfulMountMessage)
+				case config.MetadataPrefetchModeSynchronous:
+					if err = callListRecursive(mountPoint); err != nil {
+						err = fmt.Errorf("daemonize.Run: metadata-prefetch failed: %w", err)
```
kislaykishore, line 401: Why specifically daemonize.Run - won't this error trigger in foreground mode?

internal/config/mount_config.go
gargnitingoogle: I am not in favour of this change, as this field does not configure metadata-cache, nor is its effect limited to only metadata-cache (it also creates all the inodes)

Let's discuss this offline.

main.go
gargnitingoogle: It will be very chatty, even for debug logs. A bucket with million of objects will print out millions of log lines.

gargnitingoogle: Let's discuss this offline if needed.

gargnitingoogle: @kislaykishore ^

main.go
gargnitingoogle: I have seen this as norm throughout the GCSFuse code. But I'll change it here to make it more readable.

main.go
```diff
@@ -207,6 +209,28 @@ func populateArgs(c *cli.Context) (
 	return
 }
 
+func callListRecursive(mountPoint string) (err error) {
+	logger.Debugf("Started recursive metadata-prefetch of directory: \"%s\" ...", mountPoint)
+	numItems := 0
+	err = filepath.WalkDir(mountPoint, func(path string, d fs.DirEntry, err error) error {
+		//logger.Infof("Walked path:%s, d=%s, isDir=%v", path, d.Name(), d.IsDir())
+		if err != nil {
```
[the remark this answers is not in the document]
gargnitingoogle: > How should the error be handled? IIUC, the behavior implemented here is to stop the walking entirely if an error is detected. Is that what we want?

Yes, the intent is to fail this function of walking fails even on one directory/file.

> If so, then we should log that the walking failed

It is logged by the caller, which receives and logs the error, dependending on the caller's situation.

> If we are okay to skip it and continue walking other directories, fs.SkipDir should be returned.

Skipping failed walk and silently ignoring it will defeat the purpose of the walk. In case of asynchronous mode, the failed walk generates an error log, and prints it into the logfile. In case of synchronous mode, the failed walk causes the mount itself to fail.

internal/config/mount_config.go
gargnitingoogle: I originally was planning to use it for string type-checking, but have now switched it to string.

main.go
gargnitingoogle: Removed 'daemonize.Run from log'.

tools/integration_tests/operations/operations_test.go
gargnitingoogle: Removed the synchronous config.

main.go
gargnitingoogle: > In that case, let's delete this comment: go/comments#comment_out

Yes. I will.

internal/config/mount_config.go
gargnitingoogle: This is now a moot point as there is no such config-file parameter, and it's been replaced now with a cli flag, which don't have nesting.

main.go
```diff
@@ -381,18 +430,57 @@ func runCLIApp(c *cli.Context) (err error) {
 	{
 		mfs, err = mountWithArgs(bucketName, mountPoint, flags, mountConfig)
 
-		if err == nil {
+		// This utility is to absorb the error
+		// returned by daemonize.SignalOutcome calls by simply
+		// logging them as error logs.
+		callDaemonizeSignalOutcome := func(err error) {
+			if err2 := daemonize.SignalOutcome(err); err2 != nil {
+				logger.Errorf("Failed to signal to daemon: %v", err2)
```
kislaykishore, line 438: Should we use %w here?

main.go
```diff
@@ -381,18 +430,57 @@ func runCLIApp(c *cli.Context) (err error) {
...
+			} else {
+				switch flags.MetadataPrefetchMode {
+				case config.MetadataPrefetchModeDisabled:
+					markSuccessfullMount()
+				case config.MetadataPrefetchModeSynchronous:
+					if err = callListRecursive(mountPoint); err != nil {
+						// Printing via mountStatus will have duplicate logs on the console while
+						// mounting gcsfuse in foreground mode. But this is important to avoid
+						// losing error logs when run in the background mode.
+						logger.Errorf("%s: %v\n", UnsuccessfulMountMessagePrefix, err)
```
kislaykishore, line 460: Use %w for error?

main_test.go
```diff
@@ -205,9 +208,58 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInFlagStorageAsMarshal
...
+	assert.Equal(t.T(), expected, actual)
+}
+
+func (t *MainTest) TestCallListRecursiveOnPopulatedDirectory() {
+	// Set up a mini file-system to test on.
+	rootdir, err := os.MkdirTemp("/tmp", "TestCallRecursive-*")
+	assert.False(t.T(), len(rootdir) == 0 || err != nil, "failed to create temp-directory for test. err = %v", err)
```
kislaykishore, line 220: As per our coding guidelens, the tests should have three segments: Arrange, Act and Assert. Let's follow that and remove asserts that are not really the behavior that we want to test.

main_test.go
```diff
@@ -205,9 +208,58 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInFlagStorageAsMarshal
...
+	defer os.RemoveAll(rootdir)
+	for i := 0; i < 10000; i++ {
```
kislaykishore, line 222: Tests should have minimal logic and be more explicit. Refer this: go/unit-testing-practices#logic
Instead of doing this, we should create a simple directory structure that could help test the feature. For instance, if following directory structure might is good enough to verify the behavior let's use that instead of creating the directory structure in a loop.
A /
   B
      / c.txt
   d.txt

The creation code should be simpler and explicit if possible.

main_test.go
```diff
@@ -205,9 +208,58 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInFlagStorageAsMarshal
...
+		assert.False(t.T(), file1 == nil || err != nil, "failed to create file in \"%s\" for test. err = %v", rootdir, err)
+	}
+	dirPath := rootdir
+	for i := 0; i < 100; i++ {
+		dirPath = path.Join(dirPath, "dir")
+		err = os.Mkdir(dirPath, 0755)
+		assert.False(t.T(), err != nil, "failed to create sub-directory \"%s\" in \"%s\". err = %v", dirPath, rootdir, err)
+
+		file, err := os.CreateTemp(dirPath, "*****")
+		assert.False(t.T(), file == nil || err != nil, "failed to create file in \"%s\" for test. err = %v", dirPath, err)
```
kislaykishore, line 233: Assertions should be done explicitly instead of doing it in a loop.

main_test.go
```diff
@@ -205,9 +208,58 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInFlagStorageAsMarshal
...
+	err = callListRecursive(rootdir)
+
+	// Test that it does not return error.
+	assert.Nil(t.T(), err)
+}
+
+func (t *MainTest) TestCallListRecursiveOnUnpopulatedDirectory() {
+	// Set up a mini file-system to test on.
+	rootdir, err := os.MkdirTemp("/tmp", "TestCallRecursive-*")
+	assert.False(t.T(), len(rootdir) == 0 || err != nil, "failed to create temp-directory for test. err = %v", err)
```
kislaykishore, line 246: Please remove this assert.

flags.go
```diff
@@ -583,9 +622,15 @@ func validateFlags(flags *flagStorage) (err error) {
 
 	if !flags.ClientProtocol.IsValid() {
 		err = fmt.Errorf("client protocol: %s is not valid", flags.ClientProtocol)
+		return
```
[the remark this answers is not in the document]
gargnitingoogle: @kislaykishore Once again, I've just followed the conventions in the code. Technically there is nothing wrong with such returns. I'm making changes now for making forward progress, but it needs to be a part of coding guidelines discussion, rather than in a one-off code review.

flags.go
```diff
@@ -583,9 +622,15 @@ func validateFlags(flags *flagStorage) (err error) {
...
 	}
 
-	return
+	if err = validateMetadataPrefetchMode(flags.MetadataPrefetchMode); err != nil {
+		err = fmt.Errorf("%s: is not valid; error = %w", MetadataPrefetchModeFlag, err)
+		return
```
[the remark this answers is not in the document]
gargnitingoogle: same as above.

main.go
gargnitingoogle: %w only works for fmt.Errorf for wrapping errors to creating more complex errors, not for actually print statements.

main.go
gargnitingoogle: same as above. %w is not supported for actual print statements, only for fmt.Errorf for creating errors wrapping other errors.

main.go
```diff
@@ -208,6 +210,31 @@ func populateArgs(c *cli.Context) (
...
+
+		numItems++
+		return err
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed in recursive metadata-prefetch of directory: \"%s\"; error = %w", mountPoint, err)
+	}
+
+	logger.Debugf("... Completed recursive metadata-prefetch of directory: \"%s\". Number of items discovered: %v", mountPoint, numItems)
```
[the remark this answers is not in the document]
gargnitingoogle: For easier to recognize in the logs, that it completes, what was started above with the ending ellipses.

main.go
```diff
@@ -381,18 +430,57 @@ func runCLIApp(c *cli.Context) (err error) {
...
+		// This utility is to absorb the error
+		// returned by daemonize.SignalOutcome calls by simply
+		// logging them as error logs.
+		callDaemonizeSignalOutcome := func(err error) {
+			if err2 := daemonize.SignalOutcome(err); err2 != nil {
+				logger.Errorf("Failed to signal to daemon: %v", err2)
+			}
+		}
+
+		markSuccessfullMount := func() {
```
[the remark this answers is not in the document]
gargnitingoogle: Good catch! I had a feeling something was off about this.

main_test.go
gargnitingoogle: True, it's not our code, so the error is not expected. But if I ignore this error, then linter will cause smells about the ignored error.

main_test.go
gargnitingoogle: Can you elaborate what you're suggesting I do here?

main_test.go
gargnitingoogle: Removed.

main_test.go
gargnitingoogle: Obsolete comment.

main_test.go
gargnitingoogle: If we don't check the error on the `os.MkdirTemp` or the other functions, which we may have not written, it doesn't exempt us from check their error.

I removed the assert here and got the following error as expected:
`Error return value of `os.CreateTemp` is not checked (errcheck)`.

So, I have to put it back to let the linter shut up about it.

gargnitingoogle: @kislaykishore ^

main.go
kislaykishore, line 438: Error strings should not be capitalized: go/go-style/decisions#error-strings

main.go
```diff
@@ -381,18 +430,57 @@ func runCLIApp(c *cli.Context) (err error) {
...
+						return
+					}
+
+					markSuccessfulMount()
+				case config.MetadataPrefetchModeAsynchronous:
+					markSuccessfulMount()
+
+					go func() {
+						if err := callListRecursive(mountPoint); err != nil {
+							logger.Errorf("Metadata-prefetch failed: %v", err)
```
kislaykishore, line 472: Error strings should not be capitalized: go/go-style/decisions#error-strings

main.go
```diff
@@ -381,18 +430,57 @@ func runCLIApp(c *cli.Context) (err error) {
...
+					markSuccessfulMount()
+				case config.MetadataPrefetchModeSynchronous:
+					if err = callListRecursive(mountPoint); err != nil {
+						// Printing via mountStatus will have duplicate logs on the console while
+						// mounting gcsfuse in foreground mode. But this is important to avoid
+						// losing error logs when run in the background mode.
+						logger.Errorf("%s: %v\n", UnsuccessfulMountMessagePrefix, err)
+						err = fmt.Errorf("%s: mountWithArgs: %w", UnsuccessfulMountMessagePrefix, err)
+						callDaemonizeSignalOutcome(err)
+						return
```
kislaykishore, line 463: Please refrain from using naked returns: go/go-style/decisions#named-result-parameters

main.go
```diff
@@ -365,7 +394,27 @@ func runCLIApp(c *cli.Context) (err error) {
 			err = fmt.Errorf("daemonize.Run: %w", err)
 			return
 		} else {
-			logger.Infof(SuccessfulMountMessage)
+			if isDynamicMount {
+				logger.Infof(SuccessfulMountMessage)
+			} else {
+				switch flags.MetadataPrefetchMode {
+				case config.MetadataPrefetchModeDisabled:
+					logger.Infof(SuccessfulMountMessage)
```
kislaykishore, line 402: The successfulMountMessage is being logged in all cases. Better to log it at the end which should save you three lines of code.

main.go
```diff
@@ -381,18 +430,57 @@ func runCLIApp(c *cli.Context) (err error) {
...
+					markSuccessfulMount()
+
+					go func() {
+						if err := callListRecursive(mountPoint); err != nil {
+							logger.Errorf("Metadata-prefetch failed: %v", err)
+						}
+					}()
+				}
+			}
 		} else {
```
kislaykishore, line 477: We can handle the err != nil case first since it's mainly error handling and a small block of code and return. We can then handle the err == nil case and reduce which is a much more complex block and reducing its indentation will have greater readability benefits - [go/tott/651](https://goto.google.com/tott/651).

main.go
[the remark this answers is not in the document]
kislaykishore: I doubt if that's the right way to deal with this. While not a blocker for this PR, we can look deeper into how to do it in a more standard way.

main_test.go
kislaykishore: Instead of an assert you can instead use t.Fatal to ensure that while an assert in the ACT block is not distracting, the program fails if an error is returned.

main.go
gargnitingoogle: That's only for `fmt.Errorf` . The guideline says that the printed log message can be as needed.

main.go
```diff
@@ -365,7 +394,27 @@ func runCLIApp(c *cli.Context) (err error) {
 			err = fmt.Errorf("daemonize.Run: %w", err)
 			return
 		} else {
-			logger.Infof(SuccessfulMountMessage)
+			if isDynamicMount {
```
kislaykishore, line 397: This code is a bit complex. Should we use an interface with foreground and background handling as two different implementations?

kislaykishore (changes requested):

main.go
gargnitingoogle: It'll be an overkill I think given the code size.

main.go
gargnitingoogle: I am making changes right now in my code, but this is really a coding guidelines discussion.

main.go
gargnitingoogle: done

main.go
gargnitingoogle: > The successfulMountMessage is being logged in all cases. 

Yes, but always in the same order. In async case, the log is before running callListRecursive, but it is after the callListRecursive in sync case. So, I don't see any easy way around this.

> Better to log it at the end which should save you three lines of code.

That'll change output to in case of async.

main.go
gargnitingoogle: That's only for fmt.Errorf(...) . For logs/prints, there is no such constraints.

main.go
```diff
@@ -365,7 +394,27 @@ func runCLIApp(c *cli.Context) (err error) {
...
+				logger.Infof(SuccessfulMountMessage)
```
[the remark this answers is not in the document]
gargnitingoogle: good idea.

main.go
```diff
@@ -365,7 +394,27 @@ func runCLIApp(c *cli.Context) (err error) {
 			err = fmt.Errorf("daemonize.Run: %w", err)
 			return
 		} else {
```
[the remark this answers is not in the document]
gargnitingoogle: done

main_test.go
```diff
@@ -205,9 +207,33 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInFlagStorageAsMarshal
...
+
+	// Test that it does not return error.
+	assert.Nil(t.T(), err)
+}
+
+func (t *MainTest) TestCallListRecursiveOnNonExistingDirectory() {
+	// Set up a mini file-system to test on, which must fail.
+	rootdir := "/path/to/non/existing/directory"
+
+	// Call the target utility.
```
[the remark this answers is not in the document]
gargnitingoogle: removed

main_test.go
```diff
@@ -205,9 +207,33 @@ func (t *MainTest) TestStringifyShouldReturnAllFlagsPassedInFlagStorageAsMarshal
...
+	assert.Equal(t.T(), expected, actual)
+}
+
+func (t *MainTest) TestCallListRecursiveOnExistingDirectory() {
+	// Set up a mini file-system to test on.
+	rootdir, _ := os.MkdirTemp("/tmp", "TestCallListRecursive-*")
+	_, err := os.CreateTemp(rootdir, "abc-*.txt")
+	assert.Nil(t.T(), err)
```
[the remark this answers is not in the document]
gargnitingoogle: done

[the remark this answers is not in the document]
gargnitingoogle: done

main_test.go
gargnitingoogle: SG

main.go
```diff
@@ -381,19 +426,54 @@ func runCLIApp(c *cli.Context) (err error) {
...
+		}
+
+		if err != nil {
+			markFailure(err)
+			return err
+		}
+		if isDynamicMount {
+			markSuccessfulMount()
+		} else {
+			switch flags.MetadataPrefetchOnMount {
```
kislaykishore, line 460: We can log the SuccessfulMountMessage once at the end and save three lines of code.
The goroutine shouldn't really matter since anyway we are not waiting for the goroutine to finish to log the success message.

main.go
```diff
@@ -381,19 +426,54 @@ func runCLIApp(c *cli.Context) (err error) {
...
+			callDaemonizeSignalOutcome(err)
+		}
+
+		if err != nil {
+			markFailure(err)
+			return err
+		}
+		if isDynamicMount {
+			markSuccessfulMount()
+		} else {
```
kislaykishore, line 459: This else can be removed as well if the code returns right after invoking markSuccessfulMount in the isDynamicMount block.

kislaykishore (changes requested):

main.go
gargnitingoogle: No, we are not returning here in case of success, so this is not applicable here.

main.go
```diff
@@ -208,6 +210,31 @@ func populateArgs(c *cli.Context) (
 	return
 }
 
+func callListRecursive(mountPoint string) (err error) {
+	logger.Debugf("Started recursive metadata-prefetch of directory: \"%s\" ...", mountPoint)
+	numItems := 0
+	err = filepath.WalkDir(mountPoint, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
```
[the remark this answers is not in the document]
gargnitingoogle: done

flags.go
```diff
@@ -575,17 +601,33 @@ func populateFlags(c *cli.Context) (flags *flagStorage, err error) {
 	return
 }
 
+func validateMetadataPrefetchOnMount(mode string) error {
```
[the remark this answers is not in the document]
gargnitingoogle: The current code involving switch is more readable to me than the list, so I'll leave it as it is.

main.go
```diff
@@ -362,13 +391,29 @@ func runCLIApp(c *cli.Context) (err error) {
...
+		}
+		if isDynamicMount {
 			logger.Infof(SuccessfulMountMessage)
+			return err
 		}
-
-		return
+		switch flags.MetadataPrefetchOnMount {
+		case config.MetadataPrefetchOnMountDisabled:
+			logger.Infof(SuccessfulMountMessage)
```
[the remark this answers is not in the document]
gargnitingoogle: ok. handled.

kislaykishore (approved):

main.go
```diff
@@ -208,6 +210,29 @@ func populateArgs(c *cli.Context) (
 	return
 }
 
+func callListRecursive(mountPoint string) (err error) {
+	logger.Debugf("Started recursive metadata-prefetch of directory: \"%s\" ...", mountPoint)
+	numItems := 0
+	err = filepath.WalkDir(mountPoint, func(path string, d fs.DirEntry, err error) error {
```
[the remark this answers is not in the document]
gargnitingoogle: > Does that memory contributes to gcsfuse process? Do we see any spike?

Yes, and that is expected.

> Do we have gcsfuse memory usage while running the experiment on 1 million files.

I typically see around 700-800 MB for 1 million files.

main.go
```diff
@@ -254,6 +279,8 @@ func runCLIApp(c *cli.Context) (err error) {
 		return
 	}
 
+	isDynamicMount := bucketName == ""
```
raj-prince, line 282: bucketName == "_" also represents dynamic mount.

gargnitingoogle: good catch @raj-prince . Added fix. PTAL.

main.go
```diff
@@ -381,20 +421,50 @@ func runCLIApp(c *cli.Context) (err error) {
 	{
 		mfs, err = mountWithArgs(bucketName, mountPoint, flags, mountConfig)
 
-		if err == nil {
+		// This utility is to absorb the error
+		// returned by daemonize.SignalOutcome calls by simply
+		// logging them as error logs.
+		callDaemonizeSignalOutcome := func(err error) {
+			if err2 := daemonize.SignalOutcome(err); err2 != nil {
+				logger.Errorf("Failed to signal to daemon: %v", err2)
```
raj-prince, line 429: The correct message should be: "Failed to signal via daemon to parent process" 
`daemonize.SignalOutcome(err)` is called via daemon process.

main.go
```diff
@@ -381,20 +421,50 @@ func runCLIApp(c *cli.Context) (err error) {
...
+			}
+		}
+
+		markSuccessfulMount := func() {
```
raj-prince, line 433: Should we merge `markSuccessfulMount` and `markMountFailure` based on `err` parameter?

gargnitingoogle: I think it'll just make the caller code harder to read without reducing code much.

main.go
gargnitingoogle: fixed.

main.go
```diff
@@ -381,20 +421,50 @@ func runCLIApp(c *cli.Context) (err error) {
...
-			daemonize.SignalOutcome(err)
-			return
+			callDaemonizeSignalOutcome(err)
+		}
+
+		if err != nil {
+			markMountFailure(err)
+			return err
+		}
+		if !isDynamicMount(bucketName) {
```
raj-prince, line 452: Why are we calling the same thing in both background and foreground mode?
Only calling the recursive listing here should be suffice, right?

raj-prince (commented):
Added a couple of comments.

main.go
gargnitingoogle: fixed.

main.go
gargnitingoogle: > Only calling the recursive listing here should be suffice, right?

Right. No need to call it in the daemon's parent process. So, I've removed that bit  of code.

gargnitingoogle:
> Added a couple of comments.

@raj-prince I've addressed your comments. PTAL.

main.go
```diff
@@ -208,6 +210,33 @@ func populateArgs(c *cli.Context) (
...
+	if err != nil {
+		return fmt.Errorf("failed in recursive metadata-prefetch of directory: \"%s\"; error = %w", mountPoint, err)
+	}
+
+	logger.Debugf("... Completed recursive metadata-prefetch of directory: \"%s\". Number of items discovered: %v", mountPoint, numItems)
+
+	return nil
+}
+
+func isDynamicMount(bucketName string) bool {
```
raj-prince, line 236: We do similar check in mount.go too. Can we export it and use it at both the places.

gargnitingoogle: done

raj-prince (approved):

PR #1938, opened 2024-05-21
fix flaky TestRangeReadsWithCacheHit test by reducing the file size

ashmeenkaur:
### Description
TestRangeReadsWithCacheHit was sometimes failing with the following error:
```
2024/05/20 04:36:31 Running tests with flags: [--implicit-dirs=true --config-file=/tmp/gcsfuse_readwrite_test_2956740480/config --o=ro]
=== RUN   TestCacheFileForRangeReadTrueTest/TestRangeReadsWithCacheHit
    helpers_test.go:118: Incorrect cached file size. Expected 52428800, Got: 20044413
    gcs_helper.go:164: CRC32 mismatch. Expected 583443541, Got 3098506051
```
Fix: Reduced the file size to 8 MiB to ensure that complete file gets cached within time limit.

NA

### Testing details
1. Manual - Ran on louhi - the test did not fail in any of the runs
2. Unit tests - NA
3. Integration tests - via KOKORO (The test did not fail in any of the KOKORO runs)


Tulsishah (approved):
These changes looks good to me.

Alternatively, we could remove the CRC check for this test, as caching large files is expected to take a long time. Validating cache hits through log checks should be sufficient.

Just input if it makes sense :)

raj-prince (approved):
Relying on Tulsi's review.

PR #1943, opened 2024-05-23
Test tpc e2e tests

Tulsishah:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/e2e-tests-tpc.cfg
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http:#www.apache.org/licenses/LICENSE-2.0
```
kislaykishore, line 7: URL looks weird.

perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/tpc_build.sh
kislaykishore, line 1: Why is this file named tpc_build.sh when it's job is to run tests? Should run_e2e_tpc_tests.sh or ilk be a better name?

tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
```diff
@@ -99,13 +104,22 @@ func executeTestsForDynamicMounting(flags [][]string, m *testing.M) (successCode
 }
 
 func CreateTestBucketForDynamicMounting() (bucketName string) {
-	project_id, err := metadata.ProjectID()
-	if err != nil {
-		log.Printf("Error in fetching project id: %v", err)
+	var project_id, location string
+	var err error
```
kislaykishore, line 108: nit: declaring err here is not needed; it can directly be initialized as:
project_id, err := metadata.ProjectID()

kislaykishore (changes requested):

perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/tpc_build.sh
Tulsishah: This is similar to [build.sh](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/build.sh) file that's why I kept tpc_build.sh

perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/tpc_build.sh
```diff
@@ -0,0 +1,49 @@
...
+gcloud config set universe_domain apis-tpczero.goog
+gcloud config set api_endpoint_overrides/compute https://compute.apis-tpczero.goog/compute/v1/
+gcloud auth activate-service-account --key-file=/tmp/sa.key.json
+gcloud config set project $PROJECT_ID
+
+# $1 argument is refering to value of testInstalledPackage
+./tools/integration_tests/run_e2e_tests.sh $RUN_E2E_TESTS_ON_INSTALLED_PACKAGE $SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE $RUN_TEST_ON_TPC_ENDPOINT
+
+# Activate default environment after testing.
+gcloud config configurations activate default
```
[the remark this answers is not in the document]
Tulsishah: This I kept for local run. If locally some one run this script from GCE VM, than it should set default after testing.  On kokoro automation testing we don't need this.

Although fixed it by setting set +e and then -e.

tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
kislaykishore: Right, I was suggesting to just remove err declaration and keep project_id and location declarations as is. Is that also giving errors?

tools/integration_tests/util/client/storage_client.go
```diff
@@ -19,25 +19,56 @@ import (
...
-func CreateStorageClient(ctx context.Context) (*storage.Client, error) {
+func CreateStorageClient(ctx context.Context) (client *storage.Client, err error) {
 	// Create new storage client.
-	client, err := storage.NewClient(ctx)
+	if setup.TestOnTPCEndPoint() {
+		var ts oauth2.TokenSource
+		// Set up the TPC endpoint and provide a token source for authentication.
+		ts, err = getTokenSrc("/tmp/sa.key.json")
+		if err != nil {
+			return nil, fmt.Errorf(fmt.Sprintf("Unable to fetch tokenSrc for TPC: %v", err))
```
kislaykishore, line 44: nit: can we replace fmt.Errorf with
fmt.Errorf("unable to fetch token-source for TPC: %w", err)

because:
* it's concise: no need to call fmt.Sprintf
* errors are supposed to start with lowercase letters
* token-source is more readable than token-source
* Replace %v with %w [reasoning here](https://go.dev/wiki/ErrorValueFAQ#i-am-already-using-fmterrorf-with-v-or-s-to-provide-context-for-an-error-when-should-i-switch-to-w)

tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
```diff
@@ -99,13 +104,22 @@ func executeTestsForDynamicMounting(flags [][]string, m *testing.M) (successCode
...
+
+	if setup.TestOnTPCEndPoint() {
+		project_id = "tpczero-system:gcsfuse-test-project"
+		location = "u-us-prp1"
+	} else {
+		project_id, err = metadata.ProjectID()
+		if err != nil {
+			log.Printf("Error in fetching project id: %v", err)
+		}
+		location = "us-west1"
```
kislaykishore, line 118: Can we add a comment explaining why we are hardcoding the location?

perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/tpc_build.sh
kislaykishore: I'd still recommend changing the name of this and maybe the build.sh as well since it's confusing.

kislaykishore (changes requested):

tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go
Tulsishah:
```
func main() {
  var x int = 5
  
  if true {
      x, y := 2,4
      fmt.Println(x, y)
  }
}
```

Yes it is like
```
func main() {
  var x int = 5
  
  if true {
      x, y := 2,4
      fmt.Println(x, y)
  }
}
```

Yes it is like
```
func main() {
  var x int = 5
  
  if true {
      x, y := 2,4
      fmt.Println(x, y)
  }
}
```

Yes, it will give an error. If we initialize a variable and then reassign it within an if block, it will trigger an "unused variable" error.

perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/tpc_build.sh
Tulsishah: It's a build script for Kokoro. Kokoro triggers this script upon activation. We've used the name "build.sh" for all Kokoro triggers. Also it is under specific directory like e2e tests or perf test or presubmit test. Wouldn't that be sufficient for giving clarity on for what this build script is?

kislaykishore: To me, building is more about creation of the binary, that's why I had a concern about the name but then it could also be interpreted as building the binary + running the tests. So, I guess it's fine to keep it the way it is.

kislaykishore (changes requested):

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -273,6 +276,26 @@ function main(){
 
   set +e
 
+  # TPC tests will run in sequentially due to running gcloud commands in parallel is crashing frequently.
+  if [ $RUN_TEST_ON_TPC_ENDPOINT == true ]; then
+       # Clean bucket after testing.
```
kislaykishore, line 281: Clean bucket "before" testing?

kislaykishore (dismissed):

kislaykishore (approved):

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -274,6 +277,27 @@ function main(){
 
   set +e
 
+  # Run test in TPC environment
```
raj-prince, line 280: Should we create a method here, to keep the main() clean?

tools/integration_tests/operations/operations_test.go
```diff
@@ -178,6 +178,12 @@ func TestMain(m *testing.M) {
 	mountConfigFlags := createMountConfigsAndEquivalentFlags()
 	flagsSet = append(flagsSet, mountConfigFlags...)
 
+	// gRPC tests will not run in TPC environment
+	if setup.TestOnTPCEndPoint() && testing.Short() {
```
raj-prince, line 182: Later, this will cause issue when grpc test will be executed for testing.Short().

I think, we should keep TPC execution independent of Short(). That means, we need to guard in grpc flag in operation package with setup.TestonTPCEndpoint() check.

raj-prince (commented):
LGTM, a couple of minor comments.

tools/integration_tests/operations/operations_test.go
Tulsishah: Is this what you were asking for?
I added gRPC flags in case of TPC flag false only.

tools/integration_tests/operations/operations_test.go
```diff
@@ -181,13 +184,19 @@ func TestMain(m *testing.M) {
...
 
-	if !testing.Short() {
+	if !testing.Short() && !setup.TestOnTPCEndPoint() {
 		flagsSet = append(flagsSet, []string{"--client-protocol=grpc", "--implicit-dirs=true"})
 	}
 
 	mountConfigFlags := createMountConfigsAndEquivalentFlags()
 	flagsSet = append(flagsSet, mountConfigFlags...)
 
+	// gRPC tests will not run in TPC environment
```
raj-prince, line 194: Move this comment to the place where you are actually guarding grpc flag.

Please correct the comment - // Only running static_mounting test for TPC.

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -271,6 +294,11 @@ function main(){
...
```
raj-prince, line 297: Please correct the comment: `Run tpc test and exist in case RUN_TEST_ON_TPC_ENDPOINT is true.`
Anyhow if we can write exit within this block, that will improve readability.

raj-prince (commented):
Minor comments.

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -294,7 +294,7 @@ function main(){
 
   set +e
 
-  # Run test in TPC environment
+  # Run tpc test and exist in case RUN_TEST_ON_TPC_ENDPOINT is true.
```
raj-prince, line 297: nit: exist -> exit

raj-prince (dismissed):

raj-prince (approved):

PR #1945, opened 2024-05-24
Fix semantic doc for implicit dirs

Tulsishah:
### Description
The documentation is misleading because it fails to mention that Pantheon's "create folder" action also creates a 0-byte object, making it visible even without the implicit-dir flag.

Our documentation currently states that using the mkdir command in the filesystem is necessary, which is misleading.

Corrected the documentation to clarify it clearly.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -202,9 +202,7 @@ The above example was based on greenfield deployments which assumes starting fre
 
 However, if a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory, with the same name, using mkdir on the local instance (or by creating the folder in the WebUI). This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.  These backing objects for directories are special 0-byte objects, which are placeholders for directories and are not are not created by the Cloud Storage CLI tools such as gcloud or gsutil.
 
-So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using
-
-`gsutil cp -r`
+So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using `gsutil cp -r` command)
 - A/1.txt
 - A/B/2.txt
 - C/3.txt
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -202,7 +202,8 @@ The above example was based on greenfield deployments which assumes starting fre
 
 However, if a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory, with the same name, using mkdir on the local instance (or by creating the folder in the WebUI). This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.  These backing objects for directories are special 0-byte objects, which are placeholders for directories and are not created by the Cloud Storage CLI tools such as gcloud or gsutil.
 
-If a user has the following objects in their Cloud Storage buckets, for example created by uploading a local directory using `gsutil cp -r` command. - A/1.txt
+If a user has the following objects in their Cloud Storage buckets, for example created by uploading a local directory using `gsutil cp -r` command.
+- A/1.txt
 - A/B/2.txt
 - C/3.txt
 
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -207,7 +207,7 @@ If a user has the following objects in their Cloud Storage buckets, for example 
 - C/3.txt
 
 then mounting the bucket and running ```ls``` to see its content will not show any files until placeholder directory objects A/, A/B/, and C/ are created in the GCS bucket. These placeholder directory objects can be created by running `mkdir ./A`, `mkdir ./A/B` and `mkdir ./C` on GCSFuse mounted directory.
-At which point they will correctly see:
+At which point the user will correctly see:
 - A/
 - A/1.txt
 - A/B/
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -1,5 +1,3 @@
-
-
 # Read/Writes
 
 **Reads**
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -200,19 +200,23 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object 
 
 The above example was based on greenfield deployments which assumes starting fresh, where the directories are created from Cloud Storage FUSE. If a user unmounts this Cloud Storage FUSE bucket, and then re-mounts it to a different path, the user will see the directory structure correctly in the filesystem because it was originally created by Cloud Storage FUSE.
 
-However, If a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the Cloud Storage bucket using Cloud Storage FUSE, then those directories (and the sub-directories and objects under them) will not be visible. This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching Cloud Storage object ending in a slash exists.
+However, if a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory, with the same name, using mkdir on the local instance (or by creating the folder in the WebUI). This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.  These backing objects for directories are special 0-byte objects, which are placeholders for directories and are not are not created by the Cloud Storage CLI tools such as gcloud or gsutil.
 
-**Note**: These backing objects for directories are special 0-byte objects, which are placeholders for directories, and can only be created either with Cloud Storage Fuse using mkdir command or other Cloud clients e.g. Google Cloud Console. These backing objects are not created by the Cloud Storage CLI tools such as `gcloud` and `gsutil`.
+So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using
 
-Lets take an example. The user creates the following objects by uploading a local directory using `gcloud cp -r` command (learn more [here](https://cloud.google.com/storage/docs/objects#command-line)). Since gcloud creates [simulated folders](https://cloud.google.com/storage/docs/objects#simulated-folders) the following hierarchy may get created:
-- A/ (no backing object on Cloud Storage bucket)
+`gsutil cp -r`
 - A/1.txt
-- A/B/ (no backing object on Cloud Storage bucket)
 - A/B/2.txt
-- C/ (no backing object on Cloud Storage bucket)
 - C/3.txt
 
-Then, mounting the bucket and running ```ls``` to see its content will not show any files until either the directories A/, A/B/, and C/ get their backing Cloud Storage objects or `--implicit-dirs` flag is set in Cloud Storage Fuse mount command.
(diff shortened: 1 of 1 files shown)


docs/semantics.md
```diff
@@ -199,21 +200,20 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
+However, If a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the Cloud Storage bucket using Cloud Storage FUSE, then those directories (and the sub-directories and objects under them) will not be visible, which do not have backing Cloud Storage objects. This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching Cloud Storage object ending in a slash exists.
 
+**Note**: These backing objects for directories are special 0-byte objects, which are placeholders for directories, and can only be created either with Cloud Storage Fuse using mkdir command or other Cloud clients e.g. Cloud Web UI. These backing objects are not created by the Cloud Storage CLI tools such as `gcloud` and `gsutil`.
```
gargnitingoogle, line 205: > Cloud Web UI

I really meant Pantheon here. But I am not aware of what Pantheon is called in outside world. Is it `Cloud Console` ?

docs/semantics.md
```diff
@@ -228,6 +228,8 @@ However, implicit directories does have drawbacks:
 
 Alternatively, users can create a script which lists the buckets and creates the appropriate objects for the directories so that the ```--implicit-dirs``` flag is not used.
```
gargnitingoogle, line 229: This line is a duplicate of what I've added in the note right after it. Let's merge this into the Node to avoid redundancy.

docs/semantics.md
```diff
@@ -199,21 +200,20 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
 
-However, If a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory using mkdir on the local instance. This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.
+However, If a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the Cloud Storage bucket using Cloud Storage FUSE, then those directories (and the sub-directories and objects under them) will not be visible, which do not have backing Cloud Storage objects. This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching Cloud Storage object ending in a slash exists.
```
mustvicky, line 203: ", which do not have backing Cloud Storage objects" seems redundant as the prior sentence says "If a user already has objects with prefixes to simulate a directory"

docs/semantics.md
mustvicky: yes, say "Google Cloud Console"

docs/semantics.md
```diff
@@ -228,6 +228,8 @@ However, implicit directories does have drawbacks:
 
 Alternatively, users can create a script which lists the buckets and creates the appropriate objects for the directories so that the ```--implicit-dirs``` flag is not used. 
 
+**Note:** `--implicit-dirs` flag is detrimental to performance, so it should be avoided unless necessary. To avoid it, you can mount using `--implicit-dirs` flag once, then create the directories on the mounted filesystem using the ```mkdir``` command, which will create the Cloud  Storage objects backing the directories, and then mount without --implicit-dirs  flag from the next time onwards.
```
mustvicky, line 231: I am not sure if I want to make a strong statement like "**Note:** `--implicit-dirs` flag is detrimental to performance". There are simple ways to avoid the List calls
1. sufficient Metadata
2. script to convert.

docs/semantics.md
```diff
@@ -199,21 +200,20 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
 
+**Note**: These backing objects for directories are special 0-byte objects, which are placeholders for directories, and can only be created either with Cloud Storage Fuse using mkdir command or other Cloud clients e.g. Cloud Web UI. These backing objects are not created by the Cloud Storage CLI tools such as `gcloud` and `gsutil`.
 
-So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using ``` gsutil cp -r``` command)
+Let us assume the user has the following objects in their Cloud Storage bucket (for example created by uploading a local directory using ``` gsutil cp -r``` command),
```
mustvicky, line 207: I think this is the main reason for confusion and why we are even doing this work. 

Can we rewrite it like the following

Lets take an example. The user creates the following objects by uploading a local directory using ``` gcloud cp -r ``` command (learn more [here](https://cloud.google.com/storage/docs/objects#command-line)). Since gcloud creates [simulated folders](https://cloud.google.com/storage/docs/objects#simulated-folders) the following hierarchy may get created:

mustvicky (commented):
Overall, I think we are not leveraging https://cloud.google.com/storage/docs/objects#simulated-folders kinda links to connect the dots to GCS. 

Some comments to consider.

docs/semantics.md
```diff
@@ -199,22 +200,22 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
-
+However, if a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory, with the same name, using mkdir on the local instance (or by creating the folder in the WebUI). This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.  These backing objects for directories are special 0-byte objects, which are placeholders for directories and are not are not created by the Cloud Storage CLI tools such as gcloud or gsutil.
```
ankitaluthra1, line 204: typo in line - which are placeholders for directories and are not are not created by the Cloud Storage CLI tools such as gcloud or gsutil., are not is added twice.

docs/semantics.md
```diff
@@ -199,22 +200,22 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
 
-So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using ``` gsutil cp -r``` command)
+So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using `gsutil cp -r` command).
+- A/1.txt
+- A/B/2.txt
+- C/3.txt
 
+then mounting the bucket and running ```ls``` to see its content will not show any files until the directories A/, A/B/, and C/ are created on the local filesystem using the `mkdir` command (or create these from the WebUI). A user would have to run `mkdir ./A`, `mkdir ./A/B` and `mkdir ./C`.
```
ankitaluthra1, line 210: are created on the local filesystem using the `mkdir` command (or create these from the WebUI), can remove are (or create these from the WebUI) from this text as webUI will show A, B and C as folders, it might confuse the reader.

docs/semantics.md
```diff
@@ -199,22 +200,22 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
-
+However, if a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory, with the same name, using mkdir on the local instance (or by creating the folder in the WebUI). This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.  These backing objects for directories are special 0-byte objects, which are placeholders for directories and are not are not created by the Cloud Storage CLI tools such as gcloud or gsutil.
 
-So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using ``` gsutil cp -r``` command)
+So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using `gsutil cp -r` command).
```
ankitaluthra1, line 205: So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using `gsutil cp -r` command). --->
 if a user has the following objects in their Cloud Storage buckets,  for example created by uploading a local directory using `gsutil cp -r` command.
 Source of creating is not needed here as whatever will be the source, if this is the list of objects on the bucket, fuse will not show anything on ls.

ankitaluthra1 (changes requested):

ankitaluthra1:
@Tulsishah please get Marco's review done as well

docs/semantics.md
Tulsishah: Wasn't the motive of updating this document to highlight that there are methods other than `mkdir` for creating directories?

marcoa6 (dismissed):
LGTM

docs/semantics.md
```diff
@@ -199,22 +200,22 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
-
+However, if a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory, with the same name, using mkdir on the local instance (or by creating the folder in the WebUI). This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.  These backing objects for directories are special 0-byte objects, which are placeholders for directories and are not created by the Cloud Storage CLI tools such as gcloud or gsutil.
```
ashmeenkaur, line 203: Apart from GCSFuse and cloud console WebUI, simulated folders can also be created via storage client. Shall we mention that as well or make it generic by adding "For eg" before mentioning the sources of simulated folders? cc: @marcoa6

ashmeenkaur: I also think " and are not created by the Cloud Storage CLI tools such as gcloud or gsutil." is unecessary

docs/semantics.md
```diff
@@ -199,22 +200,22 @@ Even though A/, A/B/, and C/ are directories in the filesystem, a 0-byte object
...
 
-So if a user has the following objects in their Cloud Storage buckets, that were not originally created via Cloud Storage FUSE (for example created by uploading a local directory using ``` gsutil cp -r``` command)
+If a user has the following objects in their Cloud Storage buckets, for example created by uploading a local directory using `gsutil cp -r` command.
+- A/1.txt
+- A/B/2.txt
+- C/3.txt
 
+then mounting the bucket and running ```ls``` to see its content will not show any files until the directories A/, A/B/, and C/ are created on the local filesystem using the `mkdir` command (or create these from the WebUI). A user would have to run `mkdir ./A`, `mkdir ./A/B` and `mkdir ./C`.
+At which point they will correctly see:
```
ashmeenkaur, line 211:
```suggestion
At which point the user will correctly see:
```

docs/semantics.md
Tulsishah: You removed `or create these from the WebUI`. Is this intended?

ashmeenkaur: Yes, I agree with @ankitaluthra1 here that this will cause confusion as implicit directories will already be visible on WebUI. Also, since we have already mentioned how simulated folders can be created in the previous paragraph, it is redundant. What do you think? CC: @marcoa6

docs/semantics.md
marcoa6: Good feedback

However, if a user already has objects with prefixes to simulate a directory structure in their buckets that did not originate from Cloud Storage FUSE, and mounts the bucket using Cloud Storage FUSE, the directories and objects under the directories will not be visible until a user manually creates the directory, with the same name, using mkdir on the local instance. This is because with Cloud Storage FUSE, directories are by default not implicitly defined; they exist only if a matching object ending in a slash exists.  These backing objects for directories are special 0-byte objects which are placeholders for directories. Note that these can also be created via the WebUI and Cloud Storage SDKs, but not by the Cloud Storage CLI tools such as gcloud or gsutil.

docs/semantics.md
Tulsishah: Umm, agree that we already convey how simulated folders are created.

ashmeenkaur (dismissed):

ashmeenkaur (approved):

ankitaluthra1 (approved):

PR #1949, opened 2024-05-27
Move code in the main package to cmd package

kislaykishore:
* Moved the content in the main package to cmd package. The erstwhile main.go becomes legacy_main.go.
* This change is aimed to ease the move to Cobra + Viper for config management.
* A Cobra app has usually a very lightweight main function and delegates to cmd for the logic. We plan to follow the same template.
* This refactoring will help share common logic between the legacy_main.go and root.go more easily since now they are in the same package.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


main.go
```diff
@@ -20,503 +20,16 @@
...
-		logger.Fatal("Panic: %v", a)
-	}
-}
-
 func main() {
 	if strings.ToLower(os.Getenv("ENABLE_GCSFUSE_VIPER_CONFIG")) == "true" {
 		cmd.Execute()
 		return
 	}
-	defer handlePanicWhileMounting()
```
ashmeenkaur, line 507: Nit: Shall we make this first line of main and remove it from ExecuteLegacyMain for readability?

ashmeenkaur (dismissed):

main.go
kislaykishore: In Cobra, the main is usually empty and the logic is implemented in the corresponding commands in cmd/ package. There is a good reason for it. Consider that GCSFuse is an application that can potentially perform several functions:
* Mounting
* Displaying the version
* Generating a config (hypothetical)

setting up profiling is something that is probably associated with mounting and would not be needed by the other features. So, it ought to be implemented in the Cobra command for mount and not in main. WDYT?

ashmeenkaur: I see, that makes sense. Guess we need to refactor legacy main in future in that case (we can create a backlog ticket for this).
However, I still think `defer handlePanic()` should still be part of main. WDYT?

kislaykishore: I'm still not sure if we should move handlePanicWhileMounting() to main since it seems useful only for mount command. Do you think, a command such as `gcsfuse version` requires to recover?

kislaykishore: Nvm, thanks for the explanation. I see that handlePanicWhileMounting() just logs the panic. I've renamed the method to indicate that.

ashmeenkaur (approved):

cmd/version.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package cmd
```
ashmeenkaur, line 15: ldflag usage in this file needs to change as it is no longer in main

cmd/version.go
```diff
@@ -12,15 +12,15 @@
...
 
 import (
 	"fmt"
 	"runtime"
 )
 
-// Set with `-ldflags -X main.gcsfuseVersion=1.2.3` by tools/build_gcsfuse. If
-// not defined, we use "unknown" in getVersion.
+// Set with `-ldflags -X github.com/googlecloudplatform/gcsfuse/v2/cmd.gcsfuseVersion=1.2.3`
```
sethiay, line 22: I would rather suggest to keep this file unchanged and at the same path as it was (outside of cmd directory). Would it be okay if we keep this file as it was ?

main.go
```diff
@@ -20,478 +20,15 @@
...
-		return
-	}
-
-	err = appErr
-	return
-}
-
-// Handle panic if the crash occurs during mounting.
-func handlePanicWhileMounting() {
+func logPanic() {
```
sethiay, line 31: nit: logPanicWhileMounting

cmd/version.go
kislaykishore: Not really - as discussed offline, a package cannot refer to the main package since it introduces cyclic dependency.
Another approach could be to pass the version from the main command to cmd but it seems to be too much of an effort.
So, we will keep it as is.

main.go
kislaykishore: This is a catch-all and not limited to mounting. We would like to log even for non-mounting cases such as showing the version or generating the config file from the given set of flags. That's why I think we should rename it.

sethiay (approved):

PR #1957, opened 2024-05-28
Mark  "metadata-prefetch-on-mount" as experimental

gargnitingoogle:
### Description
Add "This is an experimental feature" to the `gcsfuse --help` description of "metadata-prefetch-on-mount".
This is because "metadata-prefetch-on-mount" is not a completely safe flag to use as of now. It doesn't terminate the mount process on Ctrl+C.

NA

### Testing details
1. Manual - Output of `gcsfuse --help` after the change
```
$ gcsfuse --help
... 
--experimental-metadata-prefetch-on-mount value       Experimental: This indicates whether or not to prefetch the metadata (prefilling of metadata caches and creation of inodes) of the mounted bucket at the time of mounting the bucket. Supported values: "disabled", "sync" and "async". Any other values will return error on mounting. This is applicable only to single-bucket mount-points, and not to dynamic mount-points. (default: "disabled")
...
```
3. Unit tests - NA
4. Integration tests - NA


flags.go
```diff
@@ -381,7 +381,7 @@ func newApp() (app *cli.App) {
 			cli.StringFlag{
 				Name:  MetadataPrefetchOnMountFlag,
 				Value: config.DefaultMetadataPrefetchOnMount,
-				Usage: "This indicates whether or not to prefetch the metadata (prefilling of metadata caches and creation of inodes) of the mounted bucket at the time of mounting the bucket. Supported values: \"disabled\", \"sync\" and \"async\". Any other values will return error on mounting. This is applicable only to single-bucket mount-points, and not to dynamic-mount points.",
+				Usage: "This is an experimental feature. This indicates whether or not to prefetch the metadata (prefilling of metadata caches and creation of inodes) of the mounted bucket at the time of mounting the bucket. Supported values: \"disabled\", \"sync\" and \"async\". Any other values will return error on mounting. This is applicable only to single-bucket mount-points, and not to dynamic-mount points.",
```
kislaykishore, line 384: "not to dynamic-mount points" - Should we move the hyphen between "mount" and "points" i.e. "not to dynamic mount-points"?

Tulsishah, line 384: Not sure but should we make flag name also have prefix of experimental?

flags.go
```diff
@@ -381,7 +381,7 @@ func newApp() (app *cli.App) {
 			cli.StringFlag{
 				Name:  MetadataPrefetchOnMountFlag,
```
kislaykishore, line 382: Do we want to change the flag name to experimentalMetadataPrefetchOnMountFlag? This is what has been done for other flags to mark them as experimental.

flags.go
gargnitingoogle: Added.

gargnitingoogle: done

flags.go
gargnitingoogle: This is just an internal variable name, which makes no difference to the end-user. So, I don't think it's worth making changing at many places to rename this variable.

Tulsishah (dismissed):

raj-prince (approved):

flags.go
```diff
@@ -37,7 +37,7 @@ const (
 
 	// MetadataPrefetchOnMountFlag is the name of the commandline flag for enabling
 	// metadata-prefetch mode aka 'ls -R' during mount.
-	MetadataPrefetchOnMountFlag = "metadata-prefetch-on-mount"
+	MetadataPrefetchOnMountFlag = "experimental-metadata-prefetch-on-mount"
```
sethiay, line 40: I would suggest using ExperimentalMetadataPrefetchOnMountFlag as name to maintain consistency.

flags.go
```diff
@@ -381,7 +381,7 @@ func newApp() (app *cli.App) {
 			cli.StringFlag{
 				Name:  MetadataPrefetchOnMountFlag,
 				Value: config.DefaultMetadataPrefetchOnMount,
-				Usage: "This indicates whether or not to prefetch the metadata (prefilling of metadata caches and creation of inodes) of the mounted bucket at the time of mounting the bucket. Supported values: \"disabled\", \"sync\" and \"async\". Any other values will return error on mounting. This is applicable only to single-bucket mount-points, and not to dynamic-mount points.",
+				Usage: "Experimental: This indicates whether or not to prefetch the metadata (prefilling of metadata caches and creation of inodes) of the mounted bucket at the time of mounting the bucket. Supported values: \"disabled\", \"sync\" and \"async\". Any other values will return error on mounting. This is applicable only to single-bucket mount-points, and not to dynamic mount-points.",
```
sethiay, line 384: nit: single-bucket mount-points, and not to dynamic mount-points. -> static mounting and not on dynamic mounting.

gargnitingoogle: done

flags.go
gargnitingoogle: changed everywhere.

sethiay (approved):
Please merge post running e2e tests.

PR #1960, opened 2024-05-29
Allow passing bucket location flag to run_e2e_tests.sh script

sethiay:
### Description
Allow passing bucket location flag to run_e2e_tests.sh script

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - [NA](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/f153ebd0-0f6f-406e-922b-06d1d64191fb/log)

### DIFF
--- a/tools/integration_tests/run_e2e_tests.sh
+++ b/tools/integration_tests/run_e2e_tests.sh
@@ -22,6 +22,9 @@ RUN_E2E_TESTS_ON_PACKAGE=$1
 # By default, this script runs all the integration tests.
 SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE=$2
 
+# The location (region) where bucket should be created.
+BUCKET_LOCATION=$3
+
 INTEGRATION_TEST_TIMEOUT=60m
 
 if [ "$SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE" == true ]; then
@@ -31,7 +34,6 @@ if [ "$SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE" == true ]; then
   echo "Changing the integration test timeout to: $INTEGRATION_TEST_TIMEOUT"
 fi
 
-readonly BUCKET_LOCATION="us-west1"
 readonly RANDOM_STRING_LENGTH=5
 # Test directory arrays
 TEST_DIR_PARALLEL=(
--- a/tools/integration_tests/run_e2e_tests.sh
+++ b/tools/integration_tests/run_e2e_tests.sh
@@ -27,6 +27,12 @@ BUCKET_LOCATION=$3
 
 INTEGRATION_TEST_TIMEOUT=60m
 
+if [ "$#" -ne 3 ]
+then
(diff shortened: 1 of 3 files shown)


gargnitingoogle (dismissed):
LGTM. But I have a couple of follow-ups to keep an eye on.
1. From the code the change looks correct, but can't be 100% sure. So, we need to watch the next runs to see that the change works correctly.
2. We also need to ensure that the previous runs' output data is preserved and is accessible even after this change. As we are creating new buckets each run, that too with a random name string, it should not cause a problem, but good to just keep an eye on for first few runs.

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -22,6 +22,9 @@ RUN_E2E_TESTS_ON_PACKAGE=$1
 # By default, this script runs all the integration tests.
 SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE=$2
 
+# The location (region) where bucket should be created.
+BUCKET_LOCATION=$3
```
Tulsishah, line 26: Should we keep default value something if user does not pass any? It will be helpful when someone run this script locally.

sethiay: I would rather enforce this because for them there are chances that the person executing this script will also forget to pass location and their VM location may be different from default bucket location, increasing the time to run.

Tulsishah: Hmm...okay then just suggestion, should we put check that $3 is not passed then echo "Pass bucket location same as your VM location". Currently it will break at the time of creating bucket.

Tulsishah (approved):

tools/integration_tests/run_e2e_tests.sh
sethiay: I have added a check to ensure all the three flags are passed for now because we need to pass all three flags (we can't simply pass bucket location without passing others). If required in future, we will move to take arguments like https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/perfmetrics/scripts/read_cache/run_read_cache_fio_workload.sh

gargnitingoogle: probable better to put an error log if `$#` < 3.

ashmeenkaur (approved):

Tulsishah (commented):
Not sure if you want to change this also
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/mounting/dynamic_mounting/testdata/create_bucket.sh

PTAL

sethiay:
> Not sure if you want to change this also https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/mounting/dynamic_mounting/testdata/create_bucket.sh
> 
> PTAL

Let's merge this first and see if this is also required to further reduce time.

PR #1962, opened 2024-05-30
Parallel download read cache

sethiay:
### Description
Assumptions: If there is a file of size S bytes already present then it is safe for multiple threads to write into disjoint sets of ranges into the same file where each thread has its own file handle.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/cache/file/downloader/parallel_downloader.go
```diff
@@ -0,0 +1,196 @@
...
+
+	for {
+		select {
+		case <-job.cancelCtx.Done():
+			return
+		default:
+			if start < end {
+				// Parallely download different ranges not more than ReadChunkSize and
+				// not using go routines more than AsyncDownloadParallelism
+				asyncDownloaderErrors := make([]error, job.downloadParallelism)
```
raj-prince, line 135: Instead, we should use ErrGroup: https://medium.com/@yardenlaif/go-sync-or-go-home-errgroup-f91a0ee72d3f
(a) No need to maintain error of all the go-routines.
(b) Abort if any of the go routine returns error and returns the first error (if multiple go routines returns error).

PR #1963, opened 2024-05-31
Exclude tools/ and benchmarks/ from code-coverage computation

kislaykishore:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,3 +1,6 @@
+ignore:
+  - "tools"
+  - "benchmarks"
 coverage:
   round: down
   precision: 2


Tulsishah (approved):
Will it also cause issue if we change scripts or docs? If so should we also exclude that?

kislaykishore:
> Will it also cause issue if we change scripts or docs? If so should we also exclude that?

Docs and scripts aren't included in the coverage so they shouldn't cause problems since they are not code. I've also excluded benchmarks/ which was earlier getting included in coverage computation.

Tulsishah (approved):
Please add description.

ashmeenkaur (approved):

PR #1964, opened 2024-05-31
Close the reader of async job in file cache to have the closing debug logs

sethiay:
### Description
Close the reader of async job in file cache to have the closing logs

NA

### Testing details
1. Manual - Checked with and without changes. With changes the closing logs appear.
2. Unit tests - NA
3. Integration tests -[link](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/6e27566c-42a6-4de7-ad43-01c1db7877f5/log)

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -362,6 +362,10 @@ func (job *Job) downloadObjectAsync() {
 				}
 				start += maxRead
 				if start == newReaderLimit {
+					err = newReader.Close()
+					if err != nil {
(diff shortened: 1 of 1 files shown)


raj-prince (approved):
LGTM.

Just thinking, should we log all the file-cache path logs as warning, as those are not the fatal error for customer?

sethiay:
> LGTM.
> 
> Just thinking, should we log all the file-cache path logs as warning, as those are not the fatal error for customer?

yes, I think we can for majority of logs. Let's use another PR for that. Thanks.

PR #1966, opened 2024-05-31
Add handling for custom types in Config object.

kislaykishore:
We have added support to declare and decode the following types:
* LogSeverity
* ResolvedPath
* Octal
* Protocol

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/decode_hook.go
+++ b/cfg/decode_hook.go
@@ -15,18 +15,13 @@
 package cfg
 
 import (
-	"fmt"
 	"net/url"
 	"reflect"
-	"slices"
-	"strconv"
-	"strings"
 
-	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
 	"github.com/mitchellh/mapstructure"
 )
 
-func customTypeDecodeHook() mapstructure.DecodeHookFuncType {
+func decodeURLHook() mapstructure.DecodeHookFuncType {
 	return func(
 		f reflect.Type,
 		t reflect.Type,
@@ -35,33 +30,15 @@ func customTypeDecodeHook() mapstructure.DecodeHookFuncType {
 		if f.Kind() != reflect.String {
 			return data, nil
 		}
-		s := data.(string)
-		switch t {
-		case reflect.TypeOf(Octal(0)):
-			return strconv.ParseInt(s, 8, 32)
-		case reflect.TypeOf(url.URL{}):
-			u, err := url.Parse(s)
-			if err != nil {
-				return nil, err
-			}
-			return *u, nil
-		case reflect.TypeOf(LogSeverity("")):
-			level := strings.ToUpper(s)
-			if !slices.Contains([]string{"TRACE", "DEBUG", "INFO", "WARNING", "ERROR", "OFF"}, level) {
-				return nil, fmt.Errorf("invalid logseverity: %s", s)
-			}
-			return level, nil
-		case reflect.TypeOf(Protocol("")):
-			protocol := strings.ToLower(s)
-			if !slices.Contains([]string{"http1", "http2", "grpc"}, protocol) {
-				return nil, fmt.Errorf("invalid protocol: %s", s)
-			}
-			return protocol, nil
-		case reflect.TypeOf(ResolvedPath("")):
-			return util.GetResolvedPath(s)
-		default:
+		if t != reflect.TypeOf(url.URL{}) {
 			return data, nil
 		}
+		s := data.(string)
+		u, err := url.Parse(s)
+		if err != nil {
+			return nil, err
+		}
+		return *u, nil
 	}
 }
 
@@ -69,7 +46,7 @@ func customTypeDecodeHook() mapstructure.DecodeHookFuncType {
 func DecodeHook() mapstructure.DecodeHookFunc {
 	return mapstructure.ComposeDecodeHookFunc(
 		mapstructure.TextUnmarshallerHookFunc(),
-		customTypeDecodeHook(),
+		decodeURLHook(),
 		mapstructure.StringToTimeDurationHookFunc(), // default hook
 		mapstructure.StringToSliceHookFunc(","),     // default hook
 	)
--- a/cfg/decode_hook.go
+++ b/cfg/decode_hook.go
@@ -21,7 +21,7 @@ import (
 	"github.com/mitchellh/mapstructure"
 )
 
-func decodeURLHook() mapstructure.DecodeHookFuncType {
+func stringToURLHookFunc() mapstructure.DecodeHookFuncType {
 	return func(
 		f reflect.Type,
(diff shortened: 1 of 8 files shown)


tools/config-gen/params.yaml
```diff
@@ -37,3 +38,24 @@
   config-path: "debug.log-mutex"
   type: "bool"
   usage: "Print debug messages when a mutex is held too long."
+
+- flag-name: "log-file"
+  config-path: "logging.file-path"
+  type: "resolvedPath"
+  usage: >-
```
ashmeenkaur, line 45: Why is >- needed?

tools/config-gen/params.yaml
```diff
@@ -37,3 +38,24 @@
...
+    The file for storing logs that can be parsed by fluentd. When not provided,
+    plain text logs are printed to stdout.
+
+- flag-name: "log-severity"
+  config-path: "logging.severity"
+  type: "logSeverity"
+  default: "info"
+  usage: "Specifies the logging severity"
```
ashmeenkaur, line 53: Assuming this will show up in GCSFuse help, I think we should write following instead
```suggestion
  usage: "Specifies the logging severity expressed as one of [trace, debug, info, warning, error]"
```

tools/config-gen/params.yaml
kislaykishore: ">" is one of the ways to write multiline strings in YAML. The other is "|". While "|" preserves newline, ">" doesn't. However, it adds a newline character at the end of the string. 

So,
a: >
  Hello
  World

is semantically equivalent to
a := "Hello World\n"

We don't need to have an additional newline character in our usage doc. So, I added the "-" sign which does nothing except removing the trailing newline.

cfg/decode_hook.go
```diff
@@ -0,0 +1,76 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
+	"github.com/mitchellh/mapstructure"
+)
+
+func customTypeDecodeHook() mapstructure.DecodeHookFuncType {
+	return func(
+		f reflect.Type,
```
ashmeenkaur, line 31: can we call these reflect types src and target?

tools/config-gen/params.yaml
```diff
@@ -37,3 +38,26 @@
...
+    The file for storing logs that can be parsed by fluentd. When not provided,
+    plain text logs are printed to stdout when Cloud Storage FUSE is run 
+    in the foreground, or to syslog when Cloud Storage FUSE is run in the 
+    background.
+
+- flag-name: "log-severity"
+  config-path: "logging.severity"
+  type: "logSeverity"
+  default: "info"
+  usage: "Specifies the logging severity expressed as one of [trace, debug, info, warning, error]"
```
ashmeenkaur, line 55: Sorry for the back and forth here but I forgot to add "off"

cfg/decode_hook.go
```diff
@@ -0,0 +1,76 @@
...
+		case reflect.TypeOf(url.URL{}):
+			u, err := url.Parse(s)
+			if err != nil {
+				return nil, err
+			}
+			return *u, nil
+		case reflect.TypeOf(LogSeverity("")):
+			level := strings.ToUpper(s)
+			if !slices.Contains([]string{"TRACE", "DEBUG", "INFO", "WARNING", "ERROR", "OFF"}, level) {
+				return nil, fmt.Errorf("invalid logseverity: %s", s)
```
ashmeenkaur, line 51: Let's make the error message more descriptive for the expected values

cfg/decode_hook.go
```diff
@@ -0,0 +1,76 @@
...
+		case reflect.TypeOf(Protocol("")):
+			protocol := strings.ToLower(s)
+			if !slices.Contains([]string{"http1", "http2", "grpc"}, protocol) {
+				return nil, fmt.Errorf("invalid protocol: %s", s)
+			}
+			return protocol, nil
+		case reflect.TypeOf(ResolvedPath("")):
+			return util.GetResolvedPath(s)
+		default:
+			return data, nil
```
ashmeenkaur, line 63: Shall we throw an error / log a warning in this case? I believe this should never happen in ideal scenarios?

We can probably change the order of decode hook functions to following:
```
		customTypeDecodeHook(),
```

cfg/decode_hook_test.go
```diff
@@ -0,0 +1,307 @@
...
+			args: []string{"--pathParam=~/test.txt"},
+			testFn: func(t *testing.T, c TestConfig) {
+				h, err := os.UserHomeDir()
+				if assert.Nil(t, err) {
+					assert.Equal(t, path.Join(h, "test.txt"), string(c.PathParam))
+				}
+			},
+		},
+		{
+			name: "ResolvedPath2",
```
ashmeenkaur, line 206: Nit: add test where full path is given

cfg/decode_hook.go
kislaykishore: I've restructured the code to use the encoding.TextUnmarshaler interface. Let me know if it doesn't make sense.
Secondly, if the decode fails, error is anyway thrown.  The test cases in TestParsingError test exactly that.

cfg/decode_hook.go
kislaykishore: This has been the convention for decode hooks. I can still change it if you want.

cfg/decode_hook_test.go
```diff
@@ -0,0 +1,307 @@
...
+		{
+			name: "Duration",
+			args: []string{"--durationParam=30s"},
+			testFn: func(t *testing.T, c TestConfig) {
+				assert.Equal(t, 30*time.Second, c.DurationParam)
+			},
+		},
+		{
+			name: "StringSlice1",
+			args: []string{"--stringSliceParam=a,b"},
```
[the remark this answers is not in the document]
kislaykishore: -o uses the string slice. I don't think there are any use-cases for int slice but still it seems something worth supporting.

cfg/types.go
```diff
@@ -18,8 +18,21 @@ import (
...
+type Protocol string
+
+// LogSeverity represents the logging severity and can accept the following values
+// "TRACE", "DEBUG", "INFO", "WARNING", "ERROR", "OFF"
+type LogSeverity string
+
+// ResolvedPath represents a file-path which could be the absolute or relative
+// path or could be resolved based on the value of GCSFUSE_PARENT_PROCESS_DIR
+// env var.
+type ResolvedPath string
```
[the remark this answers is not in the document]
kislaykishore: I've updated the documentation. I am planning to resolve the paths upfront.

ashmeenkaur (approved):

PR #1984, opened 2024-06-05
[POC] parallel download using transfer manager

raj-prince:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -20,11 +20,14 @@
 package storage
 
 import (
+	"cloud.google.com/go/storage/transfermanager"
 	"context"
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
+	"sync"
 
 	"cloud.google.com/go/storage"
 	control "cloud.google.com/go/storage/control/apiv2"
@@ -44,6 +47,7 @@ type bucketHandle struct {
 	bucketName    string
 	bucketType    gcs.BucketType
 	controlClient StorageControlClient
+	downloader    *transfermanager.Downloader
 }
 
 func (bh *bucketHandle) Name() string {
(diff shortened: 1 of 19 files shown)


internal/storage/bucket_handle.go
```diff
@@ -481,3 +485,59 @@ func (b *bucketHandle) getStorageLayout() (*controlpb.StorageLayout, error) {
...
+	if req.PartSize == 0 {
+		return fmt.Errorf("invalid part-size")
+	}
+
+	size := req.Range.Limit - req.Range.Start
+	numParts := (size-1)/req.PartSize + 1
+
+	wg := sync.WaitGroup{}
+
+	callback := func(out *transfermanager.DownloadOutput) {
```
raj-prince, line 504: @tritone What should be the right way to return the error on the first encounter and cancel the other worker?

internal/storage/bucket_handle.go
```diff
@@ -481,3 +485,59 @@ func (b *bucketHandle) getStorageLayout() (*controlpb.StorageLayout, error) {
...
+		}
+
+		// Add download. This is non-blocking so the download will start in the
+		// background.
+		wg.Add(1)
+		err := b.downloader.DownloadObject(ctx, in)
+		if err != nil {
+			log.Fatalf("DownloadObject: %v", err)
+		}
+	}
```
tritone, line 540: Need to add a `wg.Wait()` here to ensure that all parts actually complete. Also, the file handle should be closed (here or elsewhere).

internal/storage/bucket_handle.go
tritone: I think it would make sense to launch a sentinel goroutine that can monitor for errors via the callback, then cancel the context for the downloads if something errors. I will add this to my paste example.

tritone: I updated the paste with this, PTAL

PR #1988, opened 2024-06-07
Added flag to enable crc check 

vadlakondaswetha:
### Description
Just the flag is added. not used any where. The flag is true by default.

NA

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -86,6 +86,7 @@ type Job struct {
 
 	mu locker.Locker
 
+	// specifies whether the crc check needs to be done after the file is downloaded to cache.
 	enableCrcCheck bool
 }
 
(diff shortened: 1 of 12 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -85,6 +85,8 @@ type Job struct {
 	removeJobCallback func()
 
 	mu locker.Locker
+
+	enableCrcCheck bool
```
sethiay, line 89: nit: add comment "Specifies whether Crc check needs to be done at the end of downloading the object into cache"

sethiay (dismissed):

ashmeenkaur (dismissed):

ashmeenkaur (approved):

PR #1989, opened 2024-06-07
Added code changes to add crc check for sequential download behind flag

vadlakondaswetha:
### Description
CRC check will be done after the entire file is downloaded into cache based on the enableCRCCheck flag.

NA

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - No new integration tests

### DIFF
--- a/internal/cache/file/downloader/downloader_test.go
+++ b/internal/cache/file/downloader/downloader_test.go
@@ -20,6 +20,7 @@ import (
 	"path"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/locker"
@@ -80,6 +81,7 @@ func (dt *downloaderTest) waitForCrcCheckToBeCompleted() {
 			break
 		}
 		dt.job.mu.Unlock()
+		time.Sleep(10 * time.Millisecond)
 	}
 }
 
--- a/internal/cache/file/downloader/downloader_test.go
+++ b/internal/cache/file/downloader/downloader_test.go
@@ -74,9 +74,12 @@ func (dt *downloaderTest) waitForCrcCheckToBeCompleted() {
 	// Last notification is sent after the entire file is downloaded and before the CRC check is done.
 	// Hence, explicitly waiting till the CRC check is done.
 	for {
+		dt.job.mu.Lock()
 		if dt.job.status.Name == Completed || dt.job.status.Name == Failed {
+			dt.job.mu.Unlock()
 			break
 		}
+		dt.job.mu.Unlock()
 	}
 }
 
--- a/internal/cache/file/downloader/downloader_test.go
+++ b/internal/cache/file/downloader/downloader_test.go
@@ -16,7 +16,6 @@ package downloader
 
 import (
 	"context"
-	testutil "github.com/googlecloudplatform/gcsfuse/v2/internal/util"
 	"os"
 	"path"
 	"sync"
@@ -30,6 +29,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/util"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	testutil "github.com/googlecloudplatform/gcsfuse/v2/internal/util"
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/operations"
 	. "github.com/jacobsa/ogletest"
 )
--- a/internal/cache/file/downloader/downloader_test.go
+++ b/internal/cache/file/downloader/downloader_test.go
@@ -15,6 +15,8 @@
 package downloader
 
 import (
+	"context"
+	testutil "github.com/googlecloudplatform/gcsfuse/v2/internal/util"
 	"os"
 	"path"
 	"sync"
@@ -28,10 +30,8 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/util"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
-	testutil "github.com/googlecloudplatform/gcsfuse/v2/internal/util"
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/operations"
 	. "github.com/jacobsa/ogletest"
-	"golang.org/x/net/context"
 )
 
 var cacheDir = path.Join(os.Getenv("HOME"), "cache/dir")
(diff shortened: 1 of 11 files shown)


internal/storage/storageutil/object_attrs.go
```diff
@@ -147,6 +147,7 @@ func ConvertObjToMinObject(o *gcs.Object) *gcs.MinObject {
 		Updated:         o.Updated,
 		Metadata:        o.Metadata,
 		ContentEncoding: o.ContentEncoding,
+		CRC32C:          o.CRC32C,
```
sethiay, line 150: @gargnitingoogle Please let us know if this requires any change in logic to compute size of minObject.

charith87: Just curious, why is CRC needed to be stored in Stat Cache? Is there a way we can do by not storing in Stat Cache. ( like not storing all properties ).

vadlakondaswetha: If its not in stat cache, we end up making a stat call every time a file is downloaded.

charith87: Thanks, Got it. 
To store GCS response crc in memory for async download comparison.

internal/cache/file/downloader/job.go
```diff
@@ -463,3 +469,41 @@ func (job *Job) GetStatus() JobStatus {
...
+	if err != nil {
+		return
+	}
+
+	// If checksum matches, simply return.
+	if *job.object.CRC32C == crc32Val {
+		return nil
+	}
+
+	// If the checksum doesn't match there is an error in downloading the object contents.
```
sethiay, line 488: Couple of points here:
1. We have made responsibility of CacheHandler to insert or evict into the file cache: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/cache/file/cache_handler.go#L80. So, we can do the same here i.e. we just fail (plus update the offset in fileInfoCache to 0) the job and let the CacheHandler eventually evict this from cache from next read of same file or when the cache is full. However, I checked and it should be fine to evict (not insert) from job itself.
2. Before evicting the from file cache and after CRC check fail, we should change the state of job to FAILED to prevent others from reusing this file in cache.

internal/cache/file/downloader/job.go
```diff
@@ -463,3 +469,41 @@ func (job *Job) GetStatus() JobStatus {
 	defer job.mu.Unlock()
 	return job.status
 }
+
+func (job *Job) validateChecksum() (err error) {
```
sethiay, line 473: nit: please add func doc.

internal/cache/file/downloader/job_test.go
```diff
@@ -437,6 +437,9 @@ func (dt *downloaderTest) Test_Download_WhenAlreadyCompleted() {
 	expectedJobStatus := JobStatus{Downloading, nil, int64(objectSize)}
 	AssertTrue(reflect.DeepEqual(expectedJobStatus, jobStatus))
```
sethiay, line 439: nit: remove this line space to follow AAA

internal/cache/file/downloader/job_test.go
```diff
@@ -571,9 +574,7 @@ func (dt *downloaderTest) Test_Download_Concurrent() {
 	objectName := "path/in/gcs/foo.txt"
 	objectSize := 25 * util.MiB
 	objectContent := testutil.GenerateRandomBytes(objectSize)
-	var callbackExecutionCount atomic.Int32
-	removeCallback := func() { callbackExecutionCount.Add(1) }
-	dt.initJobTest(objectName, objectContent, DefaultSequentialReadSizeMb, uint64(objectSize*2), removeCallback)
+	dt.initJobTest(objectName, objectContent, DefaultSequentialReadSizeMb, uint64(objectSize*2), nil)
```
sethiay, line 577: I think we shouldn't remove the `removeCallback` test.  Any particular reason for removing this ?

internal/cache/file/downloader/downloader_test.go
```diff
@@ -68,6 +68,19 @@ func (dt *downloaderTest) TearDown() {
...
+	// Last notification is sent after the entire file is downloaded and before the CRC check is done.
+	// Hence, explicitly waiting till the CRC check is done.
+	for {
+		dt.job.mu.Lock()
+		if dt.job.status.Name == Completed || dt.job.status.Name == Failed {
+			dt.job.mu.Unlock()
+			break
+		}
+		dt.job.mu.Unlock()
+	}
```
sethiay, line 81: should we keep a time gap of say a millisecond between consecutive checks ?

internal/cache/file/downloader/job_test.go
```diff
@@ -780,3 +780,63 @@ func (dt *downloaderTest) Test_Invalidate_Download_Concurrent() {
...
+	// Verify fileInfoCache update
+	dt.verifyFileInfoEntry(uint64(jobStatus.Offset))
+	// Tamper the file
+	err = os.WriteFile(dt.fileSpec.Path, []byte("test"), 0644)
+	AssertEq(nil, err)
+
+	err = dt.job.validateChecksum()
+
+	AssertNe(nil, err)
+	AssertTrue(strings.Contains(err.Error(), "checksum mismatch detected"))
```
sethiay, line 809: We should add a check that the job.Status is FAILED, file in cache is deleted and the entry for file in file info cache is erased.

sethiay (commented):
LGTM with some minor comments.

internal/cache/util/util_test.go
```diff
@@ -246,6 +246,20 @@ func (ut *utilTest) Test_IsCacheHandleValid_False() {
 	}
 }
 
+func (ut *utilTest) Test_CalculateFileCRC32_ShouldReturnCrcForValidFile() {
```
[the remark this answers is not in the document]
vadlakondaswetha: done

internal/cache/file/downloader/job_test.go
vadlakondaswetha: This test is a bit different. The job downloads the file and then we are tampering the file and calling validate checksum to validate the CRC which fails. Setting job status to failed is being done by the caller if error is returned by validate checksum.

internal/cache/file/downloader/job_test.go
vadlakondaswetha: There is no way to wait for the downloadObjectAsync to be executed completed. We are either waiting for the CRC check to be completed or the notification. These 2 doesn't mean the entire function is executed. So the defer is not getting called before our Assert statement. 

We should add waits to do that validation. I dont prefer adding waits since the time keep on changing everytime logic is changed and increases test time execution.

internal/cache/file/downloader/job.go
vadlakondaswetha: 2. The caller is changing the job status to failed. anyways we are serving the file before the CRC check is completed. so whats wrong in serving the file till the entry is removed?

internal/cache/file/downloader/downloader_test.go
vadlakondaswetha: Added.

internal/cache/file/downloader/job.go
```diff
@@ -463,3 +469,41 @@ func (job *Job) GetStatus() JobStatus {
...
+	if !job.enableCrcCheck {
```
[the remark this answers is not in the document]
vadlakondaswetha: Since this will be called from multiple places -> sequential & parallel download i would prefer putting this check inside this method.

Rename id done.

internal/cache/file/downloader/job_test.go
vadlakondaswetha: done.

internal/storage/storageutil/object_attrs.go
vadlakondaswetha: Memory computation changes will be done in a seperate PR. Resolving this comment.

sethiay (dismissed):

sethiay (approved):

PR #1990, opened 2024-06-10
Parallel downloads config

sethiay:
### Description
Adding new configuration flags for parallel downloads (the flag names are not yet finalized and we will rename them with new PRs if need be)

NA

### Testing details
1. Manual - Tried passing 0 to download-parallelism-per-file, max-download-parallellism and get error.
2. Unit tests - NA
3. Integration tests - NA


internal/config/mount_config.go
```diff
@@ -106,9 +110,13 @@ type FileSystemConfig struct {
 }
 
 type FileCacheConfig struct {
-	MaxSizeMB             int64 `yaml:"max-size-mb"`
-	CacheFileForRangeRead bool  `yaml:"cache-file-for-range-read"`
-	EnableCrcCheck        bool  `yaml:"enable-crc-check"`
+	MaxSizeMB                  int64 `yaml:"max-size-mb"`
+	CacheFileForRangeRead      bool  `yaml:"cache-file-for-range-read"`
+	EnableParallelDownloads    bool  `yaml:"enable-parallel-downloads"`
+	DownloadParallelismPerFile uint  `yaml:"download-parallelism-per-file"`
```
kislaykishore, line 116: Use unsigned integers sparingly: refer: https://yaqs.corp.google.com/eng/q/8035320036228857856. Same for ReadRequestSizeMB.

internal/cache/file/downloader/job.go
```diff
@@ -58,6 +58,9 @@ type Job struct {
 	fileInfoCache        *lru.Cache
 	sequentialReadSizeMb int32
 	fileSpec             data.FileSpec
+	// Specifies whether the crc check needs to be done after the file is
```
kislaykishore, line 61: crc -> CRC
This helps indicate that it's an abbreviation.

internal/cache/file/downloader/downloader.go
```diff
@@ -42,8 +42,20 @@ type JobManager struct {
...
+	// downloadParallelismPerFile specifies the maximum number of goroutines that
+	// are spawned for downloading object from GCS per job.
+	downloadParallelismPerFile uint
+	// maxDownloadParallelism specifies the maximum number of goroutines that can be
+	// spawned for downloading objects from GCS across jobs.
+	maxDownloadParallelism int
+	// parallelReadRequestSizeMb specifies the size of read request that each of
+	// the goroutines make to GCS.
+	parallelReadRequestSizeMb uint
+	// Specifies whether Crc check needs to be done.
```
kislaykishore, line 57: Crc -> CRC

internal/config/testdata/file_cache_config/invalid_max_download_parallelism.yaml
```diff
@@ -0,0 +1,4 @@
+file-cache:
+  cache-file-for-range-read: true
+  download-parallelism-per-file: 100
+  max-download-parallelism: -10
```
kislaykishore, line 4: Let's add a newline at the end: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

internal/config/yaml_parser.go
```diff
@@ -72,7 +74,10 @@ func IsValidLogRotateConfig(config LogRotateConfig) error {
 
 func (fileCacheConfig *FileCacheConfig) validate() error {
 	if fileCacheConfig.MaxSizeMB < -1 {
-		return fmt.Errorf("the value of max-size-mb for file-cache can't be less than -1")
+		return fmt.Errorf(FileCacheMaxSizeMBInvalidValueError)
+	}
+	if fileCacheConfig.MaxDownloadParallelism < -1 {
```
kislaykishore, line 79: Should we also validate download-parallelism-per-file param?

internal/fs/fs.go
```diff
@@ -219,6 +219,16 @@ func NewFileSystem(
 }
 
 func createFileCacheHandler(cfg *ServerConfig) (fileCacheHandler *file.CacheHandler, err error) {
+	// If parallel downloads are enabled then download-parallelism-per-file or
+	// max-download-parallelism or read-request-size-mb can't be 0.
+	if cfg.MountConfig.FileCacheConfig.EnableParallelDownloads {
```
kislaykishore, line 224: How about doing this validation as part of config-parsing i.e. in func (fileCacheConfig *FileCacheConfig) validate() error

kislaykishore, line 224: This and the enclosed "if" conditions can be combined. It'll reduce the indentation as well.

kislaykishore (changes requested):

internal/fs/fs.go
sethiay: I kept it that way because i want to keep yaml_parser to just have simple checks on the input values and here it required little business logic.
Now, i have made this check simple inside yaml_parser only and removed this if condition.

kislaykishore (approved):

internal/cache/file/downloader/downloader.go
```diff
@@ -42,8 +42,20 @@ type JobManager struct {
...
+	// are spawned for downloading object from GCS per job.
+	downloadParallelismPerFile int
+	// maxDownloadParallelism specifies the maximum number of goroutines that can be
+	// spawned for downloading objects from GCS across jobs.
+	maxDownloadParallelism int
+	// parallelReadRequestSizeMb specifies the size of read request that each of
+	// the goroutines make to GCS.
+	parallelReadRequestSizeMb int
+	// Specifies whether CRC check needs to be done.
+	enableCrcCheck bool
```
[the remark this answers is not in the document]
sethiay: To be handled in next PR.

internal/config/yaml_parser.go
```diff
@@ -72,7 +76,16 @@ func IsValidLogRotateConfig(config LogRotateConfig) error {
...
+		return fmt.Errorf(MaxDownloadParallelismInvalidValueError)
+	}
+	if fileCacheConfig.EnableParallelDownloads && fileCacheConfig.DownloadParallelismPerFile < 1 {
+		return fmt.Errorf(DownloadParallelismPerFileInvalidValueError)
+	}
+	if fileCacheConfig.EnableParallelDownloads && fileCacheConfig.ReadRequestSizeMB < 1 {
```
raj-prince, line 87: As discussed offline, if we can remove the fileCacheConfig.EnableParallelDownloads from here.

internal/fs/fs.go
kislaykishore: I think we should resolve configuration related validations at one place unless it's impossible: for instance, if we keep simple configuration in the YAML, and more complicated logic here, there is fragmentation of validation logic and it's not straightforward to understand what the validation associated with a field is. And it would also impact readability. Maybe I'm missing something; something that we can discuss offline.

sethiay: Sure, we can discuss offline and fix this separately if need be.

raj-prince (approved):
LGTM.

Have question on `maxDownloadParallelism`?
if -1 is for infinite parallelism what does 0 signify. Anyway not blocker for the this PR, we can change later after the final decision.

internal/config/yaml_parser.go
```diff
@@ -72,7 +76,16 @@ func IsValidLogRotateConfig(config LogRotateConfig) error {
 
 func (fileCacheConfig *FileCacheConfig) validate() error {
 	if fileCacheConfig.MaxSizeMB < -1 {
-		return fmt.Errorf("the value of max-size-mb for file-cache can't be less than -1")
```
kislaykishore, line 75: I don't see the merit of using constants for error messages. Let's discuss this offline as well. But it's not a blocker for submission.

kislaykishore (approved):

PR #1993, opened 2024-06-10
Support fstab mount with hyphens

ashmeenkaur:
### Description
GCSFuse persistent mounting does not allow inherent GCSFuse flags with hyphens '-'. This PR adds support for hyphens '-' in GCSFuse persistent mount with the backward compatibility of supporting underscores "_".  
This PR also adds missing flags to the persistent mounting binary.

NA

### Testing details
1. Manual - Manually verified that the persistent mounting works with underscore and hyphen using the following mount commands:
- By using `mount -t` command.
- By adding an entry to `/etc/fstab` file and then running `mount -a`.
- By building `mount_gcsfuse` binary and using it for mounting.
2. Unit tests - Added

### DIFF
--- a/tools/config-gen/config.tpl
+++ b/tools/config-gen/config.tpl
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// GENERATED CODE - DO NOT EDIT config.go FILE MANUALLY.
+
 package cfg
 
 import (
--- a/tools/config-gen/config.tpl
+++ b/tools/config-gen/config.tpl
@@ -30,9 +30,6 @@ type {{ .TypeName}} struct {
 
 func BindFlags(flagSet *pflag.FlagSet) error {
   var err error
-  flagSet.SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
-  	return pflag.NormalizedName(strings.ReplaceAll(name, "-", "_"))
-  })
   {{range .FlagTemplateData}}
   flagSet.{{ .Fn}}("{{ .FlagName}}", "{{ .Shorthand}}", {{ .DefaultValue}}, {{ .Usage}})
   {{if .IsDeprecated}}
--- a/tools/config-gen/config.tpl
+++ b/tools/config-gen/config.tpl
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// GENERATED CODE - DO NOT EDIT MANUALLY.
-
 package cfg
 
 import (
@@ -32,6 +30,9 @@ type {{ .TypeName}} struct {
 
 func BindFlags(flagSet *pflag.FlagSet) error {
   var err error
+  flagSet.SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
+  	return pflag.NormalizedName(strings.ReplaceAll(name, "-", "_"))
+  })
   {{range .FlagTemplateData}}
   flagSet.{{ .Fn}}("{{ .FlagName}}", "{{ .Shorthand}}", {{ .DefaultValue}}, {{ .Usage}})
   {{if .IsDeprecated}}
--- a/tools/config-gen/config.tpl
+++ b/tools/config-gen/config.tpl
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// GENERATED CODE - DO NOT EDIT config.go FILE MANUALLY.
+// GENERATED CODE - DO NOT EDIT MANUALLY.
 
 package cfg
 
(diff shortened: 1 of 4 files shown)


tools/config-gen/config.tpl
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// GENERATED CODE - DO NOT EDIT MANUALLY.
```
kislaykishore, line 15: Was this removed mistakenly?

tools/mount_gcsfuse/main.go
```diff
@@ -123,7 +127,16 @@ func makeGcsfuseArgs(
...
 			"config_file",
 			"experimental_metadata_prefetch_on_mount",
-			"kernel_list_cache_ttl_secs":
+			"kernel_list_cache_ttl_secs",
+			"stat_cache_max_size_mb",
+			"type_cache_max_size_mb",
+			"metadata_cache_ttl_secs",
+			"log_severity",
+			"log_rotate_max_file_size_mb",
+			"log_rotate_backup_file-count",
```
kislaykishore, line 136: This should be "_count"

tools/mount_gcsfuse/main_test.go
```diff
@@ -0,0 +1,163 @@
...
+			expected: []string{"--implicit-dirs=true", "--file-mode", "0644", "--debug_fuse", "-o", "allow_other", "gcsfuse", "/mnt/gcs"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			args, err := makeGcsfuseArgs(tc.device, tc.mountPoint, tc.opts)
+			if tc.expectErr && err == nil {
+				t.Errorf("Expected error, but got none")
+			} else if !tc.expectErr && err != nil {
```
kislaykishore, line 153: nit: "else" is unnecessary here.

kislaykishore (changes requested):

tools/config-gen/config.tpl
ashmeenkaur: Updated to mention the file name (to avoid confusion with config.tpl file)

tools/mount_gcsfuse/main.go
ashmeenkaur: Good catch, thanks!

tools/config-gen/config.tpl
ashmeenkaur: makes sense.. reverted the changes!

tools/mount_gcsfuse/main.go
```diff
@@ -123,7 +127,16 @@ func makeGcsfuseArgs(
...
+			"stat_cache_max_size_mb",
+			"type_cache_max_size_mb",
+			"metadata_cache_ttl_secs",
+			"log_severity",
+			"log_rotate_max_file_size_mb",
+			"log_rotate_backup_file_count",
+			"file_cache_max_size_mb",
+			"grpc_conn_pool_size":
+
 			args = append(args, "--"+strings.Replace(name, "_", "-", -1), value)
```
kislaykishore, line 140: Use strings.ReplaceAll

kislaykishore (changes requested):
LGTM. Just some nits.

tools/mount_gcsfuse/main_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			args, err := makeGcsfuseArgs(tc.device, tc.mountPoint, tc.opts)
+			assert.Nil(t, err)
+
+			// Assert that all flags are present (no matter the order).
+			assert.ElementsMatch(t, args[:len(args)-2], tc.expected[:len(tc.expected)-2])
+			// Assert that device and mount-point are present at correct position.
```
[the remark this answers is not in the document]
kislaykishore: A more succinct version would be:
 assert.Equal(t, args[len(args) - 2:], []string{"gcsfuse", "/mnt/gcs"})

kislaykishore (approved):

tools/mount_gcsfuse/main.go
```diff
@@ -123,8 +127,17 @@ func makeGcsfuseArgs(
...
-			"kernel_list_cache_ttl_secs":
-			args = append(args, "--"+strings.Replace(name, "_", "-", -1), value)
+			"kernel_list_cache_ttl_secs",
+			"stat_cache_max_size_mb",
+			"type_cache_max_size_mb",
+			"metadata_cache_ttl_secs",
+			"log_severity",
+			"log_rotate_max_file_size_mb",
+			"log_rotate_backup_file_count",
+			"file_cache_max_size_mb",
```
sethiay, line 137: do we support flags like `file_cache_max_size`, `type-cache-max-size-mb` ?

sethiay (approved):

tools/mount_gcsfuse/main.go
ashmeenkaur: The flags are behind an environment variable currently, will be supported soon. This change won't make any difference for the end user as they will see an error in both cases (with and without this change). I am adding the flags here as it will make manual testing with persistent mounting easier for us during development.

PR #1995, opened 2024-06-10
Passing spread sheet id as param

Tulsishah:
### Description
Add an argument to the "write results in gsheet" function that allows dynamically changing the spreadsheet ID for different experiments. For example, if we run additional tests on an n2-standard-96 machine, this argument would enable saving the results to a different Google Sheet document.

NA

### Testing details
1. Manual - Tested locally by running script
2. Unit tests - NA
3. Integration tests - NA


charith87 (approved):
Seems fine as it seems to be adding a spreadsheet id to the performance scripts. If it changes something, it will be obvious in the next runs. Approving.

PR #1999, opened 2024-06-11
Add all config-params to params.yaml

kislaykishore:
### Description
* All all currently used params to params.yaml.
* Generate int flags with type as int64 as opposed to platform-dependent int.
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -73,7 +73,17 @@ type DebugConfig struct {
 type FileCacheConfig struct {
 	CacheFileForRangeRead bool `yaml:"cache-file-for-range-read"`
 
+	DownloadParallelismPerFile int64 `yaml:"download-parallelism-per-file"`
+
+	EnableCrcCheck bool `yaml:"enable-crc-check"`
+
+	EnableParallelDownloads bool `yaml:"enable-parallel-downloads"`
+
+	MaxDownloadParallelism int64 `yaml:"max-download-parallelism"`
+
 	MaxSizeMb int64 `yaml:"max-size-mb"`
+
+	ReadRequestSizeMb int64 `yaml:"read-request-size-mb"`
 }
 
 type FileSystemConfig struct {
@@ -313,6 +323,20 @@ func BindFlags(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	flagSet.IntP("download-parallelism-per-file", "", 10, "Number of concurrent download requests per file.")
+
+	err = viper.BindPFlag("file-cache.download-parallelism-per-file", flagSet.Lookup("download-parallelism-per-file"))
+	if err != nil {
+		return err
+	}
+
+	flagSet.BoolP("enable-crc-check", "", true, "Performs CRC check to ensure that file is correctly downloaded into cache.")
+
+	err = viper.BindPFlag("file-cache.enable-crc-check", flagSet.Lookup("enable-crc-check"))
+	if err != nil {
+		return err
+	}
+
 	flagSet.BoolP("enable-empty-managed-folders", "", false, "This handles the corner case in listing managed folders. There are two corner cases (a) empty managed folder (b) nested managed folder which doesn't contain any descendent as object. This flag always works in conjunction with --implicit-dirs flag. (a) If only ImplicitDirectories is true, all managed folders are listed other than above two mentioned cases. (b) If both ImplicitDirectories and EnableEmptyManagedFolders are true, then all the managed folders are listed including the above-mentioned corner case. (c) If ImplicitDirectories is false then no managed folders are listed irrespective of enable-empty-managed-folders flag.")
 
 	err = viper.BindPFlag("list.enable-empty-managed-folders", flagSet.Lookup("enable-empty-managed-folders"))
@@ -330,6 +354,13 @@ func BindFlags(flagSet *pflag.FlagSet) error {
 	flagSet.BoolP("enable-nonexistent-type-cache", "", false, "Once set, if an inode is not found in GCS, a type cache entry with type NonexistentType will be created. This also means new file/dir created might not be seen. For example, if this flag is set, and metadata-cache-ttl-secs is set, then if we create the same file/node in the meantime using the same mount, since we are not refreshing the cache, it will still return nil.")
 
 	err = viper.BindPFlag("metadata-cache.enable-nonexistent-type-cache", flagSet.Lookup("enable-nonexistent-type-cache"))
+	if err != nil {
+		return err
+	}
+
+	flagSet.BoolP("enable-parallel-downloads", "", false, "Enable parallel downloads.")
+
+	err = viper.BindPFlag("file-cache.enable-parallel-downloads", flagSet.Lookup("enable-parallel-downloads"))
 	if err != nil {
 		return err
 	}
@@ -508,6 +539,13 @@ func BindFlags(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	flagSet.IntP("max-download-parallelism", "", -1, "Sets an uber limit of number of concurrent file download requests that are made across all files.")
+
+	err = viper.BindPFlag("file-cache.max-download-parallelism", flagSet.Lookup("max-download-parallelism"))
+	if err != nil {
+		return err
+	}
+
 	flagSet.IntP("max-idle-conns-per-host", "", 100, "The number of maximum idle connections allowed per server.")
 
 	err = viper.BindPFlag("gcs-connection.max-idle-conns-per-host", flagSet.Lookup("max-idle-conns-per-host"))
@@ -546,6 +584,13 @@ func BindFlags(flagSet *pflag.FlagSet) error {
 	flagSet.StringP("only-dir", "", "", "Mount only a specific directory within the bucket. See docs/mounting for more information")
 
 	err = viper.BindPFlag("only-dir", flagSet.Lookup("only-dir"))
+	if err != nil {
+		return err
+	}
+
+	flagSet.IntP("read-request-size-mb", "", 0, "Size of chunks in MiB that each concurrent request downloads.")
+
+	err = viper.BindPFlag("file-cache.read-request-size-mb", flagSet.Lookup("read-request-size-mb"))
 	if err != nil {
 		return err
 	}
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -285,7 +285,7 @@ func BindFlags(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.BoolP("debug_gcs", "", false, "This flag is unused. Debug fuse logs are now controlled by log-severity flag, please use log-severity trace to view the logs.")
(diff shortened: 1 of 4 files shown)


tools/config-gen/params.yaml
```diff
@@ -15,16 +31,308 @@
...
+    exceeds this limit, the retry continues with this specified maximum value.
+  default: "30s"
+
+- flag-name: "stat-cache-max-size-mb"
+  config-path: "metadata-cache.stat-cache-max-size-mb"
+  type: "int"
+  usage: >-
+    The maximum size of stat-cache in MiBs. It can also be set to -1 for
+    no-size-limit, 0 for no cache. Values below -1 are not supported.
+  default: "math.MinInt64"
```
ashmeenkaur, line 135: shouldn't this be 32?

tools/config-gen/params.yaml
```diff
@@ -15,16 +31,308 @@
...
+  default: "4"
+
+- flag-name: "metadata-cache-ttl"
+  config-path: "metadata-cache.ttl-secs"
+  type: "int"
+  usage: >-
+    The ttl value in seconds to be used for expiring items in metadata-cache. It
+    can be set to -1 for no-ttl, 0 for no cache and > 0 for ttl-controlled
+    metadata-cache. Any value set below -1 will throw an error."
+  default: "math.MinInt64"
```
ashmeenkaur, line 150: shouldn't this be 60s and type duration?

tools/config-gen/params.yaml
```diff
@@ -15,16 +31,308 @@
...
+    now, the minimum of stat-cache-ttl and type-cache-ttl values, rounded up to
+    the next higher multiple of a second is used as ttl for both stat-cache and
+    type-cache, when metadata-cache-ttl-secs is not set.
+  default: "60s"
+  deprecated: true
+  deprecation-warning: >-
+    This flag has been deprecated (starting v2.0) in favor of
+    metadata-cache-ttl-secs.
+
+- flag-name: "kernel-list-cache-ttl-secs"
```
ashmeenkaur, line 199: Should the flag be `kernel-list-cache-ttl` with type duration?

tools/config-gen/params.yaml
ashmeenkaur, line 1: Nit: For ease of reading, can we add default values for cases where zero value and default value are same? (eg default value of bool flag false, int flag 0)

tools/config-gen/params.yaml
```diff
@@ -15,16 +31,308 @@
...
+  config-path: "gcs-connection.http-client-timeout"
+  type: "duration"
+  usage: >-
+    The time duration that http client will wait to get response from the
+    server. The default value 0 indicates no timeout.
+
+- flag-name: "max-retry-duration"
+  type: "duration"
+  usage: "This is currently unused."
+  deprecated: true
```
ashmeenkaur, line 220: @Tulsishah just confirming that we are going to deprecate this flag (in favor of adding max-retry-count flag)?

@kislaykishore we can probably update the warning message accordingly.

tools/config-gen/params.yaml
```diff
@@ -15,16 +31,308 @@
...
+
+- flag-name: "stackdriver-export-interval"
+  config-path: "metrics.stackdriver-export-interval"
+  type: "duration"
+  usage: >-
+    Export metrics to stackdriver with this interval. The default value 0
+    indicates no exporting.
+  default: "0s"
+
+- flag-name: "experimental-opentelemetry-collector-address"
```
ashmeenkaur, line 277: Shall we skip this from config file as it is deprecated?
same for other flags like `experimental-enable-json-read`

tools/config-gen/params.yaml
```diff
@@ -15,16 +31,308 @@
...
+  default: "true"
   deprecated: true
   deprecation-warning: "This flag is currently unused."
 
 - flag-name: "debug_fuse"
   type: "bool"
-  default: "true"
+  usage: "This flag is currently unused."
+  deprecated: true
+  deprecation-warning: "This flag is currently unused."
```
ashmeenkaur, line 319: Can we instead write "debug fuse logs are now controlled by log-severity flag, please use log-severity trace to view the logs." or something similar?

Same for debug_gcs flag

tools/config-gen/params.yaml
```diff
@@ -39,25 +347,117 @@
...
+    (a) If only ImplicitDirectories is true, all managed folders are listed other than above two mentioned cases.
+    (b) If both ImplicitDirectories and EnableEmptyManagedFolders are true, then all the managed folders are listed including the above-mentioned corner case.
+    (c) If ImplicitDirectories is false then no managed folders are listed irrespective of enable-empty-managed-folders flag.
+
+- flag-name: "grpc-conn-pool-size"
+  config-path: "gcs-connection.grpc-conn-pool-size"
+  type: "int"
+  usage: "The number of gRPC channel in grpc client."
+  deprecated: true
+  deprecation-warning: "Experimental flag: can be removed in a minor release."
```
ashmeenkaur, line 437: name of the flag doesn't say experimental?

tools/config-gen/params.yaml
```diff
@@ -39,25 +347,117 @@
...
+  type: "bool"
+  usage: "Enables support for HNS buckets"
+
+- flag-name: "log-rotate-max-log-file-size-mb"
+  config-path: "logging.log-rotate.max-file-size-mb"
+  type: "int"
+  default: "512"
+  usage: "the maximum size in megabytes that a log file can reach before it is rotated."
+
+- flag-name: "log-rotate-backup-file-count"
```
ashmeenkaur, line 450: NIT: can we keep all related flags together?

tools/config-gen/params.yaml
```diff
@@ -39,25 +347,117 @@
...
+  usage: >-
+    The maximum number of backup log files to retain after they have been
+    rotated. The default value is 10. When value is set to 0, all backup files are
+    retained.
+
+- flag-name: "log-rotate-compress"
+  config-path: "logging.log-rotate.compress"
+  type: "bool"
+  default: "true"
+  usage: "Whether the rotated log files should be gzipped."
```
ashmeenkaur, line 463: NIT:
```suggestion
  usage: "Controls whether the rotated log files should be compressed using GZIP."
```

tools/config-gen/params.yaml
```diff
@@ -39,25 +347,117 @@
...
+- flag-name: "enable-hns"
+  config-path: "enable-hns"
+  type: "bool"
+  usage: "Enables support for HNS buckets"
+
+- flag-name: "log-rotate-max-log-file-size-mb"
+  config-path: "logging.log-rotate.max-file-size-mb"
+  type: "int"
+  default: "512"
+  usage: "the maximum size in megabytes that a log file can reach before it is rotated."
```
ashmeenkaur, line 448: NIT:
```suggestion
  usage: "The maximum size in megabytes that a log file can reach before it is rotated."
```

ashmeenkaur (commented):
Have not taken a look at generated file

tools/config-gen/params.yaml
kislaykishore: It's currently an int flag: https://github.com/GoogleCloudPlatform/gcsfuse/blob/9cfb931f63dcc5559cee29aa2a31fb295ac620af/internal/config/mount_config.go#L89C2-L89C27

tools/config-gen/params.yaml
kislaykishore: It's of type int: https://github.com/GoogleCloudPlatform/gcsfuse/blob/9cfb931f63dcc5559cee29aa2a31fb295ac620af/internal/config/mount_config.go#L128

tools/config-gen/params.yaml
kislaykishore: We still need to have a config-path for it in order to fetch the value, right?

tools/config-gen/params.yaml
kislaykishore: We can do that at a later stage. For now, I've tried to keep the order like it was in flags.go so that I don't miss any entries.

tools/config-gen/params.yaml
ashmeenkaur: my bad.. I am guessing we kept it this way to support `-1`. Resolving.

tools/config-gen/params.yaml
ashmeenkaur: Noted, thanks. Shall we change the default value though?

tools/config-gen/params.yaml
kislaykishore: Sure, I've set it to 32. In the current code, it's currently set to math.MinInt64 initially by default and later on it's value is resolved based on this and some other flags. We will have to change that logic before we start using this new config. Same for metadata-cache-ttl.

tools/config-gen/params.yaml
Tulsishah: Yes. We are

kislaykishore: I'd like to keep the warning as is. We can update the warning once the new flag is added.

cfg/config.go
```diff
@@ -17,64 +17,261 @@
...
+
+	Gid int64 `yaml:"gid"`
+
+	IgnoreInterrupts bool `yaml:"ignore-interrupts"`
+
+	RenameDirLimit int64 `yaml:"rename-dir-limit"`
+
+	TempDir string `yaml:"temp-dir"`
+
+	Uid int64 `yaml:"uid"`
```
ashmeenkaur, line 104: disable-parallel-dirops config missing

tools/config-gen/params.yaml
```diff
@@ -15,49 +33,460 @@
...
+    This handles the corner case in listing managed folders.
+    There are two corner cases (a) empty managed folder (b) nested managed folder which doesn't contain any descendent as object.
+    This flag always works in conjunction with --implicit-dirs flag.
+    (a) If only ImplicitDirectories is true, all managed folders are listed other than above two mentioned cases.
+    (b) If both ImplicitDirectories and EnableEmptyManagedFolders are true, then all the managed folders are listed including the above-mentioned corner case.
+    (c) If ImplicitDirectories is false then no managed folders are listed irrespective of enable-empty-managed-folders flag.
+  default: false
+
+- flag-name: "experimental-grpc-conn-pool-size"
+  config-path: "gcs-connection.grpc-conn-pool-size"
```
ashmeenkaur, line 460: should this also be experimental

kislaykishore: I don't think so. This config currently exists (and is not a new config). So, renaming it is out of scope for this PR.

cfg/config.go
```diff
@@ -86,6 +285,25 @@ func BindFlags(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	flagSet.BoolP("debug_gcs", "", false, "This flag is unused. Debug fuse logs are now controlled by log-severity flag, please use log-severity trace to view the logs.")
```
ashmeenkaur, line 288:
```suggestion
	flagSet.BoolP("debug_gcs", "", false, "This flag is unused. Debug GCS logs are now controlled by log-severity flag, please use log-severity trace to view the logs.")
```

ashmeenkaur (approved):

PR #2002, opened 2024-06-11
updating memory computation because a new field is added in minObject.go

vadlakondaswetha:
NA

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/cache/metadata/stat_cache_test.go
+++ b/internal/cache/metadata/stat_cache_test.go
@@ -158,12 +158,12 @@ func (t *StatCacheTest) FillUpToCapacity() {
 	m1 := &gcs.MinObject{Name: "taco"}
 	m2 := &gcs.MinObject{Name: "quesadilla"}
 
-	t.cache.Insert(m0, expiration)                    // size = 1394 bytes
-	t.cache.Insert(m1, expiration)                    // size = 1382 bytes (cumulative = 2776 bytes)
-	t.cache.AddNegativeEntry("enchilada", expiration) // size = 178 bytes (cumulative = 2954 bytes)
-	t.cache.Insert(m2, expiration)                    // size = 1406 bytes (cumulative = 4360 bytes)
-	t.cache.AddNegativeEntry("fajita", expiration)    // size = 172 bytes (cumulative = 4532 bytes)
-	t.cache.AddNegativeEntry("salsa", expiration)     // size = 170 bytes (cumulative = 4702 bytes)
+	t.cache.Insert(m0, expiration)                    // size = 1410 bytes
+	t.cache.Insert(m1, expiration)                    // size = 1398 bytes (cumulative = 2808 bytes)
+	t.cache.AddNegativeEntry("enchilada", expiration) // size = 178 bytes (cumulative = 2986 bytes)
+	t.cache.Insert(m2, expiration)                    // size = 1422 bytes (cumulative = 4408 bytes)
+	t.cache.AddNegativeEntry("fajita", expiration)    // size = 172 bytes (cumulative = 4580 bytes)
+	t.cache.AddNegativeEntry("salsa", expiration)     // size = 170 bytes (cumulative = 4750 bytes)
 
 	// Before expiration
 	justBefore := expiration.Add(-time.Nanosecond)
@@ -199,15 +199,15 @@ func (t *StatCacheTest) ExpiresLeastRecentlyUsed() {
 	o1 := &gcs.MinObject{Name: "taco"}
 	o2 := &gcs.MinObject{Name: "quesadilla"}
 
-	t.cache.Insert(o0, expiration)                         // size = 1394 bytes
-	t.cache.Insert(o1, expiration)                         // Least recent, size = 1382 bytes (cumulative = 2776 bytes)
-	t.cache.AddNegativeEntry("enchilada", expiration)      // Third most recent, size = 178 bytes (cumulative = 2954 bytes)
-	t.cache.Insert(o2, expiration)                         // Second most recent, size = 1406 bytes (cumulative = 4360 bytes)
+	t.cache.Insert(o0, expiration)                         // size = 1410 bytes
+	t.cache.Insert(o1, expiration)                         // Least recent, size = 1398 bytes (cumulative = 2808 bytes)
+	t.cache.AddNegativeEntry("enchilada", expiration)      // Third most recent, size = 178 bytes (cumulative = 2986 bytes)
+	t.cache.Insert(o2, expiration)                         // Second most recent, size = 1422 bytes (cumulative = 4408 bytes)
 	AssertEq(o0, t.cache.LookUpOrNil("burrito", someTime)) // Most recent
 
 	// Insert another.
 	o3 := &gcs.MinObject{Name: "queso"}
-	t.cache.Insert(o3, expiration) // size = 1386 bytes (cumulative = 5746 bytes)
+	t.cache.Insert(o3, expiration) // size = 1402 bytes (cumulative = 5810 bytes)
 	// This would evict the least recent entry i.e o1/"taco".
 
 	// See what's left.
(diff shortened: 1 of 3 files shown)


gargnitingoogle:
@vadlakondaswetha @sethiay  This is more an FYI than a review comment. Currently the declared value of `AverageSizeOfPositiveStatCacheEntry ` is 1400 bytes (@ https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/mount/flag.go#L48) . With this new crc32 field being added, each entry-size goes up by exactly 24 bytes  (2x for RSS (8bytes for pointer + 4 bytes for crc32 value)) . I don't think this change (1.7%) is big enough for us to change the value of AverageSizeOfPositiveStatCacheEntry. But still just wanted to highlight this.

sethiay:
> @vadlakondaswetha @sethiay This is more an FYI than a review comment. Currently the declared value of `AverageSizeOfPositiveStatCacheEntry ` is 1400 bytes (@
> 
> https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/mount/flag.go#L48
> 
> ) . With this new crc32 field being added, each entry-size goes up by exactly 24 bytes (2x for RSS (8bytes for pointer + 4 bytes for crc32 value)) . I don't think this change (1.7%) is big enough for us to change the value of AverageSizeOfPositiveStatCacheEntry. But still just wanted to highlight this.

I think that should be fine but should we verify as we did last time ?

internal/util/sizeof_test.go
```diff
@@ -278,13 +291,14 @@ func (t *SizeofTest) TestNestedSizeOfGcsMinObject() {
...
 	var metaGeneration int64 = 858734899
+	var crc32 uint32 = 1234
 	updated := time.Now()
 	customMetadaField1 := "google-symlink"
 	customMetadaValue1 := "true"
 	customMetadaField2 := "google-xyz-field"
 	customMetadaValue2 := "google-symlink"
 	customMetadataFields := map[string]string{customMetadaField1: customMetadaValue1, customMetadaField2: customMetadaValue2}
-	customMetadataFieldsContentSize := emptyStringSize + contentSizeOfString(&customMetadaField1) + emptyStringSize + contentSizeOfString(&customMetadaValue1) + emptyStringSize + contentSizeOfString(&customMetadaField2) + emptyStringSize + contentSizeOfString(&customMetadaValue2)
+	customMetadataFieldsContentSize := +emptyStringSize + contentSizeOfString(&customMetadaField1) + emptyStringSize + contentSizeOfString(&customMetadaValue1) + emptyStringSize + contentSizeOfString(&customMetadaField2) + emptyStringSize + contentSizeOfString(&customMetadaValue2)
```
gargnitingoogle, line 301: stray `+`; looks like result of rebase. Please remove.

internal/util/sizeof_test.go
```diff
@@ -294,10 +308,11 @@ func (t *SizeofTest) TestNestedSizeOfGcsMinObject() {
 		Generation:      generation,
 		MetaGeneration:  metaGeneration,
 		Updated:         updated,
+		CRC32C:          &crc32,
```
gargnitingoogle, line 311: Didn't notice it earlier. It's probably outside the scope of this review. But just curious. What does the trailing `C` in CRC32**C** stand for?

gargnitingoogle: Also, the local variable on the RHS should also get a trailing `c` then.

gargnitingoogle (changes requested):

gargnitingoogle:
> > @vadlakondaswetha @sethiay This is more an FYI than a review comment. Currently the declared value of `AverageSizeOfPositiveStatCacheEntry ` is 1400 bytes (@
> > https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/mount/flag.go#L48
> > 
> > ) . With this new crc32 field being added, each entry-size goes up by exactly 24 bytes (2x for RSS (8bytes for pointer + 4 bytes for crc32 value)) . I don't think this change (1.7%) is big enough for us to change the value of AverageSizeOfPositiveStatCacheEntry. But still just wanted to highlight this.
> 
> I think that should be fine but should we verify as we did last time ?

@sethiay  If by "verify as we did last time" you mean run the linear regression for the size-estimation again, then I'd say no, as we've only added a fixed-size delta (24 bytes) to the cache-entry size here, unless we think the typical input set of stat-cache entries itself has changed since then, which I don't think has.

I had suggested @vadlakondaswetha earlier during our discussion that this change might break existing unit tests which are on the boundary (i.e. actual size of inserted objects ~= configured size of cache), but those must be passing as part of the `ci / linux-tests ` check.
We could update the comments about the size calculations in the stat-cache unit tests such as https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/cache/metadata/stat_cache_test.go#L161, which are more for readability and maintainability, but have no functional value really.

Another concern could be impact on runtime of the size calculation, but that'll be negligible with this change.

internal/util/sizeof_test.go
vadlakondaswetha: crc32c stands for Castagnoli CRC32, a different polynomial used for computing the checksum
I believe it should be fine to not have c in crc32c at every place since gcs only uses crc32c

gargnitingoogle: > I believe it should be fine to not have c in crc32c at every place since gcs only uses crc32c

sounds good.

vadlakondaswetha:
> > > @vadlakondaswetha @sethiay This is more an FYI than a review comment. Currently the declared value of `AverageSizeOfPositiveStatCacheEntry ` is 1400 bytes (@
> > > https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/mount/flag.go#L48
> > > 
> > > ) . With this new crc32 field being added, each entry-size goes up by exactly 24 bytes (2x for RSS (8bytes for pointer + 4 bytes for crc32 value)) . I don't think this change (1.7%) is big enough for us to change the value of AverageSizeOfPositiveStatCacheEntry. But still just wanted to highlight this.
> > 
> > 
> > I think that should be fine but should we verify as we did last time ?
> 
> @sethiay If by "verify as we did last time" you mean run the linear regression for the size-estimation again, then I'd say no, as we've only added a fixed-size delta (24 bytes) to the cache-entry size here, unless we think the typical input set of stat-cache entries itself has changed since then, which I don't think has.
> 
> I had suggested @vadlakondaswetha earlier during our discussion that this change might break existing unit tests which are on the boundary (i.e. actual size of inserted objects ~= configured size of cache), but those must be passing as part of the `ci / linux-tests ` check. We could update the comments about the size calculations in the stat-cache unit tests such as
> 
> https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/cache/metadata/stat_cache_test.go#L161
> 
> , which are more for readability and maintainability, but have no functional value really.
> Another concern could be impact on runtime of the size calculation, but that'll be negligible with this change.

@gargnitingoogle -> Its 4 + 4 -> 8 bytes not 12 bytes.

vadlakondaswetha:
> > > > @vadlakondaswetha @sethiay This is more an FYI than a review comment. Currently the declared value of `AverageSizeOfPositiveStatCacheEntry ` is 1400 bytes (@
> > > > https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/mount/flag.go#L48
> > > > 
> > > > ) . With this new crc32 field being added, each entry-size goes up by exactly 24 bytes (2x for RSS (8bytes for pointer + 4 bytes for crc32 value)) . I don't think this change (1.7%) is big enough for us to change the value of AverageSizeOfPositiveStatCacheEntry. But still just wanted to highlight this.
> > > 
> > > 
> > > I think that should be fine but should we verify as we did last time ?
> > 
> > 
> > @sethiay If by "verify as we did last time" you mean run the linear regression for the size-estimation again, then I'd say no, as we've only added a fixed-size delta (24 bytes) to the cache-entry size here, unless we think the typical input set of stat-cache entries itself has changed since then, which I don't think has.
> > I had suggested @vadlakondaswetha earlier during our discussion that this change might break existing unit tests which are on the boundary (i.e. actual size of inserted objects ~= configured size of cache), but those must be passing as part of the `ci / linux-tests ` check. We could update the comments about the size calculations in the stat-cache unit tests such as
> > https://github.com/GoogleCloudPlatform/gcsfuse/blob/66135bbf496b43bab6852d92a4356f9de74f0638/internal/cache/metadata/stat_cache_test.go#L161
> > 
> > , which are more for readability and maintainability, but have no functional value really.
> > Another concern could be impact on runtime of the size calculation, but that'll be negligible with this change.
> 
> @gargnitingoogle -> Its 4 + 4 -> 8 bytes not 12 bytes.

@gargnitingoogle  -> sorry its 8+4 -> 12 bytes sorry for the confusion.

internal/cache/metadata/stat_cache_test.go
```diff
@@ -158,12 +158,12 @@ func (t *StatCacheTest) FillUpToCapacity() {
 	m1 := &gcs.MinObject{Name: "taco"}
 	m2 := &gcs.MinObject{Name: "quesadilla"}
 
-	t.cache.Insert(m0, expiration)                    // size = 1394 bytes
-	t.cache.Insert(m1, expiration)                    // size = 1382 bytes (cumulative = 2776 bytes)
-	t.cache.AddNegativeEntry("enchilada", expiration) // size = 178 bytes (cumulative = 2954 bytes)
-	t.cache.Insert(m2, expiration)                    // size = 1406 bytes (cumulative = 4360 bytes)
-	t.cache.AddNegativeEntry("fajita", expiration)    // size = 172 bytes (cumulative = 4532 bytes)
-	t.cache.AddNegativeEntry("salsa", expiration)     // size = 170 bytes (cumulative = 4702 bytes)
+	t.cache.Insert(m0, expiration)                    // size = 1410 bytes
```
gargnitingoogle, line 161: are you sure that this is 1410 and not 1418 as should be with additional 24 bytes added per positive entry?

vadlakondaswetha: min object has only name, so CRC32 pointer is nil.

gargnitingoogle: ok.

gargnitingoogle (approved):
LGTM

sethiay (approved):
.

PR #2003, opened 2024-06-12
Use 1.22 as the min Go version

kislaykishore:
The "go" line in go.mod specifies the minimum Go version that's required by the current module. While building, go automatically picks up the installed go version if it's same or more recent than the specified go version in go.mod. Since GCSFuse is compatible with go v1.22, we can declare the same in go.mod and let Go decide the right toolchain to build.

b/346695794

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/googlecloudplatform/gcsfuse/v2
 
-go 1.22.4
+go 1.22
 
 require (
 	cloud.google.com/go/compute/metadata v0.3.0


kislaykishore:
Performance benchmarks:
image

kislaykishore:
One can verify the go version used to build the binary using to `go version -v <path_to_binary>` command. For me it came to:
```
% go version -v gcsfuse 
gcsfuse: go1.23-20240419-RC02 cl/626470163 +7f76c00fc5 X:fieldtrack,boringcrypto
```

raj-prince (approved):
LGTM.

Since we install the latest go version (latest patch) while building the package, hence we are good to go.

PR #2004, opened 2024-06-12
passes cache config to job

ankitaluthra1:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -59,7 +59,7 @@ type Job struct {
 	fileInfoCache        *lru.Cache
 	sequentialReadSizeMb int32
 	fileSpec             data.FileSpec
-	// Specifies whether the CRC check needs to be done after the file is
+	fileCacheConfig      *config.FileCacheConfig
 	// downloaded to cache.
 
 	/////////////////////////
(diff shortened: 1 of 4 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -60,7 +61,6 @@ type Job struct {
 	fileSpec             data.FileSpec
 	// Specifies whether the CRC check needs to be done after the file is
```
sethiay, line 62: remove the comment of field being removed.

internal/cache/file/downloader/job.go
```diff
@@ -87,7 +87,8 @@ type Job struct {
 	// is responsibility of JobManager to pass this function.
 	removeJobCallback func()
 
-	mu locker.Locker
+	mu              locker.Locker
+	fileCacheConfig *config.FileCacheConfig
```
sethiay, line 91: move the fileCacheConfig just below fileSpec and in place of enableCrcCheck

internal/cache/file/downloader/job.go
ankitaluthra1: done

internal/cache/file/downloader/job.go
ankitaluthra1: done

sethiay (approved):

PR #2005, opened 2024-06-12
Add parallel downloads job

sethiay:
### Description
Add parallel downloads job

Note: I will add more unit/composite tests in the next PR. Also, I will enable the configs to run e2e tests with parallel downloads in a separate PR.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -242,6 +242,9 @@ func (job *Job) updateStatusAndNotifySubscribers(statusName jobStatusName, statu
 	job.mu.Unlock()
 }
 
+// updateStatusOffset updates the offset in job's status and in file info cache
+// with the given offset. If the the update is successful, this function also
+// notify the subscribers.
 func (job *Job) updateStatusOffset(downloadedOffset int64) (err error) {
 	fileInfoKey := data.FileInfoKey{
 		BucketName: job.bucket.Name(),
@@ -376,7 +379,7 @@ func (job *Job) downloadObjectAsync() {
 		}
 	}()
 
-	// Both parallel or non-parallel download functions supports cancellation in
+	// Both parallel and non-parallel download functions supports cancellation in
 	// case of job's cancellation.
 	if job.fileCacheConfig.EnableParallelDownloads {
 		err = job.parallelDownloadObjectToFile(cacheFile)
@@ -400,7 +403,6 @@ func (job *Job) downloadObjectAsync() {
 		err = job.validateCRC()
 		if err != nil {
 			job.updateStatusAndNotifySubscribers(Failed, err)
-			return
 		} else {
 			job.updateStatusAndNotifySubscribers(Completed, err)
 		}
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -383,7 +383,7 @@ func (job *Job) downloadObjectAsync() {
 		}
 	}()
 
-	// Both parallel and non-parallel download functions supports cancellation in
+	// Both parallel and non-parallel download functions support cancellation in
 	// case of job's cancellation.
 	if job.fileCacheConfig.EnableParallelDownloads {
 		err = job.parallelDownloadObjectToFile(cacheFile)
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -225,45 +225,110 @@ func (job *Job) notifySubscribers() {
 	}
 }
 
-// failWhileDownloading changes the status of job to failed and notifies
-// subscribers about the download error.
+// updateStatusAndNotifySubscribers is a helper function which updates the
+// status's name and err to given name and error and notifies subscribers.
 //
-// Acquires and releases LOCK(job.mu)
-func (job *Job) failWhileDownloading(downloadErr error) {
-	logger.Errorf("Job:%p (%s:/%s) failed with: %v", job, job.bucket.Name(), job.object.Name, downloadErr)
+// Acquires and releases Lock(job.mu)
+func (job *Job) updateStatusAndNotifySubscribers(statusName jobStatusName, statusErr error) {
+	if statusName == Failed {
+		logger.Errorf("Job:%p (%s:/%s) Failed with error: %v", job, job.bucket.Name(), job.object.Name, statusErr)
+	} else {
+		logger.Tracef("Job:%p (%s:/%s) status changed to %v with error: %v", job, job.bucket.Name(), job.object.Name, statusName, statusErr)
+	}
 	job.mu.Lock()
-	job.status.Err = downloadErr
-	job.status.Name = Failed
+	job.status.Err = statusErr
+	job.status.Name = statusName
 	job.notifySubscribers()
 	job.mu.Unlock()
 }
 
-// updateFileInfoCache updates the file info cache with latest offset downloaded
-// by job. Returns error in case of failure.
-//
-// Not concurrency safe and requires LOCK(job.mu)
-func (job *Job) updateFileInfoCache() (err error) {
+func (job *Job) updateStatusOffset(downloadedOffset int64) (err error) {
 	fileInfoKey := data.FileInfoKey{
 		BucketName: job.bucket.Name(),
 		ObjectName: job.object.Name,
 	}
 	fileInfoKeyName, err := fileInfoKey.Key()
 	if err != nil {
-		err = fmt.Errorf("updateFileInfoCache: error while creating fileInfoKeyName for bucket %s and object %s %w",
+		err = fmt.Errorf("updateStatusOffset: error while creating fileInfoKeyName for bucket %s and object %s %w",
 			fileInfoKey.BucketName, fileInfoKey.ObjectName, err)
 		return
 	}
 
 	updatedFileInfo := data.FileInfo{
 		Key: fileInfoKey, ObjectGeneration: job.object.Generation,
-		FileSize: job.object.Size, Offset: uint64(job.status.Offset),
+		FileSize: job.object.Size, Offset: uint64(downloadedOffset),
 	}
-
-	logger.Tracef("Job:%p (%s:/%s) downloaded till %v offset.", job, job.bucket.Name(), job.object.Name, job.status.Offset)
+	job.mu.Lock()
+	defer job.mu.Unlock()
 	err = job.fileInfoCache.UpdateWithoutChangingOrder(fileInfoKeyName, updatedFileInfo)
-	if err != nil {
-		err = fmt.Errorf("updateFileInfoCache: error while inserting into fileInfoCache %s: %w", updatedFileInfo.Key, err)
-		return
+	if err == nil {
+		job.status.Offset = downloadedOffset
+		// Notify subscribers if file cache is updated.
+		logger.Tracef("Job:%p (%s:/%s) downloaded till %v offset.", job, job.bucket.Name(), job.object.Name, job.status.Offset)
+		job.notifySubscribers()
+	} else {
+		err = fmt.Errorf("updateStatusOffset: error while updating offset: %v in fileInfoCache %s: %w", downloadedOffset, updatedFileInfo.Key, err)
+	}
+	return
+}
+
+// downloadObjectToFile downloads the backing object from GCS into the given
+// file and updates the file info cache. It uses gcs.Bucket's NewReader method
+// to download the object.
+func (job *Job) downloadObjectToFile(cacheFile *os.File) (err error) {
+	var newReader io.ReadCloser
+	var start, end, sequentialReadSize, newReaderLimit int64
+	end = int64(job.object.Size)
+	sequentialReadSize = int64(job.sequentialReadSizeMb) * cacheutil.MiB
+
+	// Each iteration of this for loop, reads ReadChunkSize size of range of the
+	// backing object from reader into the file handle and updates the file info
+	// cache. In case, reader is not present for reading, it creates a
+	// gcs.Bucket's NewReader with size min(sequentialReadSize, object.Size).
+	for start < end {
+		if newReader == nil {
+			newReaderLimit = min(start+sequentialReadSize, end)
+			newReader, err = job.bucket.NewReader(
+				job.cancelCtx,
+				&gcs.ReadObjectRequest{
+					Name:       job.object.Name,
+					Generation: job.object.Generation,
+					Range: &gcs.ByteRange{
+						Start: uint64(start),
+						Limit: uint64(newReaderLimit),
+					},
+					ReadCompressed: job.object.HasContentEncodingGzip(),
+				})
+			if err != nil {
+				err = fmt.Errorf("downloadObjectToFile: error in creating NewReader with start %d and limit %d: %w", start, newReaderLimit, err)
+				return err
+			}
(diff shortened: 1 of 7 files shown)


internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -0,0 +1,182 @@
...
+		select {
+		case <-job.cancelCtx.Done():
+			return
+		default:
+			if start < end {
+				// Parallely download different ranges not more than maxSingleStepReadSize
+				// and not using go routines more than job.downloadParallelism
+				downloadErrGroup, downloadErrGroupCtx := errgroup.WithContext(job.cancelCtx)
+
+				var singleStepReadSize int64 = 0
```
raj-prince, line 120: Why singleStepReadSize and maxSingleStepReadSize variables are required? Anyway we are checking the end case using (start < end), right?

internal/cache/file/downloader/parallel_downloads_job_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2023 Google Inc. All Rights Reserved.
```
raj-prince, line 1: nit: 2024

internal/cache/file/downloader/job.go
```diff
@@ -269,26 +268,33 @@ func (job *Job) updateFileInfoCache() (err error) {
 	return
 }
 
+// cleanUpDownloadAsyncJob is a helper function which performs clean up tasks
+// for the async job and this should be called at the end of async job.
+//
+// Acquires and releases LOCK(job.mu)
+func (job *Job) cleanUpDownloadAsyncJob() {
```
raj-prince, line 275: Should we write unit test for this?

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
+	"golang.org/x/sync/errgroup"
+)
+
+// downloadRangeToFile is a helper function to download a given range of object
+// from GCS into given file handle.
+//
+// This function doesn't take locks and can be executed parallely.
+func (job *Job) downloadRangeToFile(ctx context.Context, cacheFile *os.File, start, end int64) error {
```
raj-prince, line 38: Should we pass the offsetWriter here?

raj-prince (commented):
A couple of minor comments.

internal/cache/file/downloader/parallel_downloads_job.go
raj-prince, line 38: Also, assuming unit test for this method will be in the next separate PR.

sethiay: Trying to offload some work from `parallelDownloadObjectAsync` as that's already long and allow unit testing that separately.

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: So, this check is to ensure that in one step (includes writing to file info cache), we only write chunk size * download-parallelism-per-file content only whereas start and end are start and end of file

internal/cache/file/downloader/job.go
sethiay: done

internal/cache/file/downloader/parallel_downloads_job.go
vadlakondaswetha: Isn;t this simple loop sufficient:


for (ind = 0; ind < dowloadParallelism && start < end; ind ++) {
   range_start = start 
   range_end = min(end, range_start+chunk_size) // chunk_size tell how much each go routine should download
   start = range_end
}

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -0,0 +1,182 @@
...
+						return job.downloadRangeToFile(downloadErrGroupCtx, cacheFile, rangeStart, rangeEnd)
+					})
+
+					singleStepReadSize = singleStepReadSize + rangeEnd - rangeStart
+					start = rangeEnd
+				}
+				// If any of the go routine failed, consider the async job failed.
+				err = downloadErrGroup.Wait()
+				if err != nil {
+					if errors.Is(err, context.Canceled) {
```
vadlakondaswetha, line 136: why is context canceled means invalid job?

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -0,0 +1,182 @@
...
+					return
+				}
+				job.mu.Unlock()
+				// Change status of job in case of error while updating file cache.
+				if err != nil {
+					job.failWhileDownloading(err)
+					return
+				}
+			} else {
+				err = job.validateCRC()
```
vadlakondaswetha, line 168: why are we duplicating the code so much?

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: As discussed, I will make this change

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -0,0 +1,182 @@
...
+func (job *Job) parallelDownloadObjectAsync() {
+	// Cleanup the async job in all cases - completion/failure/invalidation.
+	defer job.cleanUpDownloadAsyncJob()
+
+	// Create, and open cache file for writing object into it.
+	cacheFile, err := cacheutil.CreateFile(job.fileSpec, os.O_TRUNC|os.O_RDWR)
+	if err != nil {
+		err = fmt.Errorf("parallelDownloadObjectAsync: error in creating cache file: %w", err)
+		job.failWhileDownloading(err)
+		return
```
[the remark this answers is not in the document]
sethiay: I think both the logic are already small to be created a separate function for. If you think creating separate ones improve readability then I will.

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: the context passed is the context associated with the job and in the current job flows, context is canceled only when the job is Invalidated (during eviction).

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: I tried to but it's not straightforward to break down the function into logically independent blocks. 
Trying one more thing, hopefully that should (but would require additional unit testing)

internal/cache/file/downloader/parallel_downloads_job.go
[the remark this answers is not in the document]
vadlakondaswetha: we need to create seperate functions.

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -0,0 +1,182 @@
...
+	var start, end int64
+	end = int64(job.object.Size)
+	var parallelReadRequestSize = int64(job.fileCacheConfig.ReadRequestSizeMB) * cacheutil.MiB
+	var maxSingleStepReadSize = parallelReadRequestSize * int64(job.fileCacheConfig.DownloadParallelismPerFile)
+
+	for {
+		select {
+		case <-job.cancelCtx.Done():
+			return
+		default:
```
vadlakondaswetha, line 114: Looking at the code more i see only the default clause is different for sequential download vs parallel download, why can;t we put a if -else based on flag and invoke seperate functions?

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: Done. PTAL

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -0,0 +1,182 @@
...
+		err = fmt.Errorf("downloadRangeToFile: error at the time of copying content to cache file %w", err)
+	}
+	return err
+}
+
+// parallelDownloadObjectAsync does parallel download of the backing GCS object
+// into a file as part of file cache using multiple NewReader method of
+// gcs.Bucket running in parallel
+//
+// Note: There can only be one async parallel or non-parallel download running
```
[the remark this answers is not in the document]
sethiay: Removed.

internal/cache/file/downloader/parallel_downloads_job.go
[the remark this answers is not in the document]
sethiay: refactored

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: refactored. Will add unit tests in next PR for refactored methods.

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: done. PTAL

internal/cache/file/downloader/job.go
```diff
@@ -226,196 +225,186 @@ func (job *Job) notifySubscribers() {
...
 	}
-
-	logger.Tracef("Job:%p (%s:/%s) downloaded till %v offset.", job, job.bucket.Name(), job.object.Name, job.status.Offset)
+	job.mu.Lock()
+	defer job.mu.Unlock()
 	err = job.fileInfoCache.UpdateWithoutChangingOrder(fileInfoKeyName, updatedFileInfo)
-	if err != nil {
-		err = fmt.Errorf("updateFileInfoCache: error while inserting into fileInfoCache %s: %w", updatedFileInfo.Key, err)
-		return
+	if err == nil {
```
vadlakondaswetha, line 267: lets use the same format everywhere
if err != nil {
  // do something
return
}

continue with other logic

internal/cache/file/downloader/job.go
```diff
@@ -226,196 +225,186 @@ func (job *Job) notifySubscribers() {
...
+		start += maxRead
+		if start == newReaderLimit {
+			err = newReader.Close()
+			if err != nil {
+				logger.Errorf("Job:%p (%s:/%s) error while closing reader: %v", job, job.bucket.Name(), job.object.Name, err)
+			}
+			newReader = nil
+		}
+
+		err = job.updateStatusOffset(start)
```
vadlakondaswetha, line 331: Please add a comment why are we not returning from the above err.

internal/cache/file/downloader/job.go
```diff
@@ -226,196 +225,186 @@ func (job *Job) notifySubscribers() {
...
+					Range: &gcs.ByteRange{
+						Start: uint64(start),
+						Limit: uint64(newReaderLimit),
+					},
+					ReadCompressed: job.object.HasContentEncodingGzip(),
+				})
+			if err != nil {
+				err = fmt.Errorf("downloadObjectToFile: error in creating NewReader with start %d and limit %d: %w", start, newReaderLimit, err)
+				return err
+			}
```
vadlakondaswetha, line 308: can you extract this into seperate function so it can be reused in parallel_downloads_job also

internal/cache/file/downloader/job.go
```diff
@@ -226,196 +225,186 @@ func (job *Job) notifySubscribers() {
...
 	}
 
-	var newReader io.ReadCloser
-	var start, end, sequentialReadSize, newReaderLimit int64
-	end = int64(job.object.Size)
-	sequentialReadSize = int64(job.sequentialReadSizeMb) * cacheutil.MiB
-
-	for {
-		select {
-		case <-job.cancelCtx.Done():
```
vadlakondaswetha, line 322: where are we checking if context is cancelled?

sethiay: It's not required because both parallelDownloadObjectToFile & downloadObjectToFile returns err (context.Canceled type) if context is canceled. This is because the `newReader` that we get as output of NewReader call is derived from job.cancelCtx and hence `_, err = io.CopyN(offsetWriter, newReader, maxRead)` fails in that case with context canceled error. This is equivalent to what was there before.

internal/cache/file/downloader/job.go
sethiay: Could you elaborate what code to put in separate function ?

Assuming you are asking to put {NewReader + copy} logic into a function and then use that in both parallel and non parallel downloads. If yes, then I would avoid that because:
1. NewReader and Copy functions are already abstracted enough and are only called at both the places with their returned errors handled.
2. The read + copy logic is not common in parallel and non parallel downloads. In former case whatever is read is copied to file whereas in latter, a read request of 200MiB is made and data is copied in chunks of 8MiB with updateOffset being called.
3. Even if we go ahead with common function, it's signature will be something like downloadRange(ctx context.Context, dstWriter io.Writer, start int64, end int64, readRequestSize int64, copyChunkSize int64, debugReadType string). This will remove slight duplicacy at the cost of many arguments. 
So, I would prefer keeping it as it is but lmk.

In case you are asking to just abstract out NewReader + err handling code, then also I would prefer keeping it as it is.

internal/cache/file/downloader/job.go
sethiay: Added. PTAL

internal/cache/file/downloader/job.go
```diff
@@ -275,130 +365,52 @@ func (job *Job) updateFileInfoCache() (err error) {
...
-			job.failWhileDownloading(err)
+			job.handleError(err)
 		}
 	}()
 
-	var newReader io.ReadCloser
-	var start, end, sequentialReadSize, newReaderLimit int64
-	end = int64(job.object.Size)
-	sequentialReadSize = int64(job.sequentialReadSizeMb) * cacheutil.MiB
+	// Both parallel and non-parallel download functions supports cancellation in
```
raj-prince, line 386: nit: download functions `support`

internal/cache/file/downloader/job_test.go
```diff
@@ -89,6 +89,11 @@ func (dt *downloaderTest) initJobTest(objectName string, objectContent []byte, s
 	AssertEq(nil, err)
 }
 
+func (dt *downloaderTest) verifyInvalidError(err error) {
```
raj-prince, line 92: Optional: but I would prefer to duplicate for more readability.

internal/cache/file/downloader/job.go
```diff
@@ -226,47 +225,138 @@ func (job *Job) notifySubscribers() {
...
-	return
+
+	err = fmt.Errorf("updateStatusOffset: error while updating offset: %v in fileInfoCache %s: %w", downloadedOffset, updatedFileInfo.Key, err)
+	return err
+}
+
+// downloadObjectToFile downloads the backing object from GCS into the given
+// file and updates the file info cache. It uses gcs.Bucket's NewReader method
+// to download the object.
+func (job *Job) downloadObjectToFile(cacheFile *os.File) (err error) {
```
raj-prince, line 282: Are we writing unit test for this?

raj-prince (commented):
Few nits.

LGTM, assuming we are adding unit tests for newly added method as separate PR.

internal/cache/file/downloader/job.go
sethiay: Yes, will be there in next PR. Not a blocker for this PR since we already have unit tests for Download function and didn't want to make the PR longer.

vadlakondaswetha (approved):

PR #2008, opened 2024-06-12
adds change for not waiting for download in case of sequential read

ankitaluthra1:
### Description
adds code change for not waiting for download in case of sequential read when parallel download is true, this PR includes code changes, e2e/integration tests will be added once code is merged

b/345645002

### Testing details
1. Manual - NA
2. Unit tests - done
3. Integration tests - NA


internal/cache/file/cache_handle.go
```diff
@@ -150,10 +152,12 @@ func (fch *CacheHandle) Read(ctx context.Context, bucket gcs.Bucket, object *gcs
...
 	isSequentialRead := fch.IsSequential(offset)
-	waitForDownload := true
-	if !isSequentialRead {
-		fch.isSequential = false
-		waitForDownload = false
+	fch.isSequential = isSequentialRead
+
+	waitForDownload := false
+
+	if isSequentialRead && !fch.fileDownloadJob.IsParallelDownloadEnabled() {
```
sethiay, line 159: it's not safe to assume that fileDownloadJob will never be nil. Please add the check for parallel download inside the block which is there in the later part of code. (there will be no changed in this existing if condition)
`
if fch.fileDownloadJob != nil {
`

internal/cache/file/cache_handle.go
```diff
@@ -84,10 +83,13 @@ func (fch *CacheHandle) shouldReadFromCache(jobStatus *downloader.JobStatus, req
 		jobStatus.Name == downloader.Failed {
 		err := fmt.Errorf("%s: jobStatus: %s jobError: %w", util.InvalidFileDownloadJobErrMsg, jobStatus.Name, jobStatus.Err)
 		return err
-	} else if jobStatus.Offset < requiredOffset {
+	}
+
+	if jobStatus.Offset < requiredOffset {
```
sethiay, line 88: any advantage of moving this to `if` from `else if` ?

internal/cache/file/downloader/job.go
```diff
@@ -139,6 +131,14 @@ func (job *Job) checkInvariants() {
 	}
 }
 
+func (job *Job) IsParallelDownloadEnabled() bool {
```
sethiay, line 134: nit: ParallelDownloads

internal/cache/file/downloader/job_test.go
```diff
@@ -848,3 +848,26 @@ func (dt *downloaderTest) Test_validateCRC_ForTamperedFileWhenEnableCrcCheckIsFa
 	// Verify fileInfoCache update
 	dt.verifyFileInfoEntry(uint64(jobStatus.Offset))
 }
+
+func (dt *downloaderTest) Test_When_Parallel_Download_Is_Not_Enabled() {
+	objectName := "path/in/gcs/file1.txt"
+	objectSize := 8 * util.MiB
+	objectContent := testutil.GenerateRandomBytes(objectSize)
```
sethiay, line 855: you don't need to do dt.initiJobTest again as it was already done as part of test setup before every test.
(Also, you can explicitly set dt.job.fileCacheConfig.EnableParallelDownloads = false as you are doing in case of enabled)

internal/cache/file/cache_handle.go
ankitaluthra1: fch.fileDownloadJob.IsParallelDownloadEnabled() is not assuming file handle will never be null, there is a null check inside (ref)[	if job != nil && job.fileCacheConfig != nil && job.fileCacheConfig.EnableParallelDownloads {], IsParallelDownloadEnabled is working as expected with fileDownloadJob as Nil, based on below existing  tests passing with this change 
Test_Read_WithNilFileDownloadJobAndCacheMiss
Test_Read_WithNilFileDownloadJobAndCacheHit

internal/cache/file/cache_handle.go
ankitaluthra1: having if with return is more readable than if else ladder, else is redundant here, as there is an if inside

internal/cache/file/cache_handle.go
sethiay: fch.fileDownloadJob can itself be nil (it's a pointer) and in that case this call will translate to nil. IsParallelDownloadEnabled which will give runtime error.
Also, the job != nil inside IsParallelDownloadEnabled is redundant, I don't think it's possible to call method of a struct with struct being nil.

sethiay: The tests that you mentioned passed because the first mentioned test is not sequential reads and hence the if condition where you are calling isParallelDownloadEnabled got true without reaching the call: https://github.com/GoogleCloudPlatform/gcsfuse/blob/ad99e175b1e56ed71ec3d1001782066904d653b3/internal/cache/file/cache_handle_test.go#L536
And the second test passed because isSequential is set to false https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/cache/file/cache_handle_test.go#L558

PR #2013, opened 2024-06-13
Making compute crc method context cancellable

vadlakondaswetha:
### Description
Making compute crc method context cancellable, so it can be cancelled as part of job cancellation

NA

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/job.go
+++ b/internal/cache/file/downloader/job.go
@@ -305,13 +305,6 @@ func (job *Job) downloadObjectAsync() {
 		}
 	}()
 
-	notifyInvalid := func() {
-		job.mu.Lock()
-		job.status.Name = Invalid
-		job.notifySubscribers()
-		job.mu.Unlock()
-	}
-
 	var newReader io.ReadCloser
 	var start, end, sequentialReadSize, newReaderLimit int64
 	end = int64(job.object.Size)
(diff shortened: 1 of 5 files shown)


internal/cache/file/downloader/job_test.go
```diff
@@ -848,3 +850,23 @@ func (dt *downloaderTest) Test_validateCRC_ForTamperedFileWhenEnableCrcCheckIsFa
 	// Verify fileInfoCache update
 	dt.verifyFileInfoEntry(uint64(jobStatus.Offset))
 }
+
+func (dt *downloaderTest) Test_validateCRC_WheContextIsCancelled() {
+	objectName := "path/in/gcs/file2.txt"
+	objectSize := 10 * util.MiB
```
sethiay, line 856: use 20MiB for safer side and request 1MiB in Download function as it's possible that job completes by the time execution reaches to Assert block of this unit test.

internal/cache/file/downloader/job.go
```diff
@@ -483,7 +483,7 @@ func (job *Job) validateCRC() (err error) {
 		return
 	}
 
-	crc32Val, err := cacheutil.CalculateFileCRC32(job.fileSpec.Path)
+	crc32Val, err := cacheutil.CalculateFileCRC32(job.cancelCtx, job.fileSpec.Path)
 	if err != nil {
 		return
```
sethiay, line 488: cancellation happens during invalidation of job (and cache), so add additional check similar to this https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/cache/file/downloader/job.go#L342 here as well: https://github.com/GoogleCloudPlatform/gcsfuse/blob/ad99e175b1e56ed71ec3d1001782066904d653b3/internal/cache/file/downloader/job.go#L409

sethiay: also, modify this to check Invalid status for completion: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/cache/file/downloader/downloader_test.go#L79

internal/cache/file/downloader/job_test.go
vadlakondaswetha: 10MB works fine. If i request for 1MB there is chance that context is cancelled before entire 20MB is dowloaded, hence crc check is not invoked.

sethiay (approved):

PR #2022, opened 2024-06-17
Run all except cache tests concurrently

kislaykishore:
### Description
* Cache-tests cannot be run concurrently. So, except for cache-tests run other tests concurrently.
* Hardcoding the port in fakestorage leads to bind conflicts  if another fakestorage instance is running on the same port. This happens when the tests are run concurrently. Remove the hardcoding to let fakestorage find the port by itself and avoid the port conflict.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


ashmeenkaur (approved):
LGTM logically. @raj-prince can you check why running cache tests in parallel even with dynamic port runs into errors/conflicts and if we can do something to fix it?

PR #2027, opened 2024-06-18
Make Wait for Download false for all read scenarios when parallel downloads is true

ankitaluthra1:
### Description
Make Wait for Download false for all read scenarios when parallel downloads is true

b/345645002

### Testing details
1. Manual - NA
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/cache_handle_test.go
+++ b/internal/cache/file/cache_handle_test.go
@@ -820,3 +820,31 @@ func (cht *cacheHandleTest) Test_MultipleReads_CacheHitShouldBeFalseThenTrue() {
 	assert.True(cht.T(), cacheHit)
 	assert.Nil(cht.T(), err)
 }
+
+func (cht *cacheHandleTest) Test_Read_Sequential_Parallel_Download_True() {
+	dst := make([]byte, ReadContentSize)
+	offset := int64(cht.object.Size - ReadContentSize)
+	cht.cacheHandle.isSequential = true
+	cht.cacheHandle.cacheFileForRangeRead = true
+
+	fileDownloadJob := downloader.NewJob(
+		cht.object,
+		cht.bucket,
+		cht.cache,
+		DefaultSequentialReadSizeMb,
+		cht.fileSpec,
+		func() {},
+		&config.FileCacheConfig{EnableCrcCheck: true, EnableParallelDownloads: true},
+	)
+
+	cht.cacheHandle.fileDownloadJob = fileDownloadJob
+
+	n, cacheHit, err := cht.cacheHandle.Read(context.Background(), cht.bucket, cht.object, offset, dst)
+
+	jobStatus := cht.cacheHandle.fileDownloadJob.GetStatus()
+	assert.Equal(cht.T(), downloader.Downloading, jobStatus.Name)
+	assert.False(cht.T(), cacheHit)
+	assert.NotNil(cht.T(), err)
+	assert.Equal(cht.T(), 0, n)
+	assert.True(cht.T(), strings.Contains(err.Error(), util.FallbackToGCSErrMsg))
+}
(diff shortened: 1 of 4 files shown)


internal/cache/file/cache_handle_test.go
```diff
@@ -820,3 +820,31 @@ func (cht *cacheHandleTest) Test_MultipleReads_CacheHitShouldBeFalseThenTrue() {
 	assert.True(cht.T(), cacheHit)
 	assert.Nil(cht.T(), err)
 }
+
+func (cht *cacheHandleTest) Test_Read_Sequential_Parallel_Download_True() {
+	dst := make([]byte, ReadContentSize)
+	offset := int64(cht.object.Size - ReadContentSize)
+	cht.cacheHandle.isSequential = true
+	cht.cacheHandle.cacheFileForRangeRead = true
+
```
sethiay, line 829: nit: remove this line (to signify Arrange block)

internal/cache/file/cache_handle_test.go
```diff
@@ -820,3 +820,31 @@ func (cht *cacheHandleTest) Test_MultipleReads_CacheHitShouldBeFalseThenTrue() {
...
+	fileDownloadJob := downloader.NewJob(
+		cht.object,
+		cht.bucket,
+		cht.cache,
+		DefaultSequentialReadSizeMb,
+		cht.fileSpec,
+		func() {},
+		&config.FileCacheConfig{EnableCrcCheck: true, EnableParallelDownloads: true},
```
sethiay, line 837: pass some values to `DownloadParallelismPerFile` and `ReadRequestSizeMb`

internal/cache/file/cache_handle_test.go
```diff
@@ -820,3 +820,31 @@ func (cht *cacheHandleTest) Test_MultipleReads_CacheHitShouldBeFalseThenTrue() {
...
+	)
+
+	cht.cacheHandle.fileDownloadJob = fileDownloadJob
+
+	n, cacheHit, err := cht.cacheHandle.Read(context.Background(), cht.bucket, cht.object, offset, dst)
+
+	jobStatus := cht.cacheHandle.fileDownloadJob.GetStatus()
+	assert.Equal(cht.T(), downloader.Downloading, jobStatus.Name)
+	assert.False(cht.T(), cacheHit)
+	assert.NotNil(cht.T(), err)
```
sethiay, line 847: you can use [assert.ErrorContains](https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.ErrorContains) and avoid two separate NotNil and assert.True check.

internal/cache/file/cache_handle_test.go
```diff
@@ -820,3 +820,31 @@ func (cht *cacheHandleTest) Test_MultipleReads_CacheHitShouldBeFalseThenTrue() {
 	assert.True(cht.T(), cacheHit)
 	assert.Nil(cht.T(), err)
 }
+
+func (cht *cacheHandleTest) Test_Read_Sequential_Parallel_Download_True() {
```
sethiay, line 824: also, add Read_Random test for parallel downloads

internal/cache/file/cache_handle_test.go
```diff
@@ -820,3 +820,31 @@ func (cht *cacheHandleTest) Test_MultipleReads_CacheHitShouldBeFalseThenTrue() {
...
+	dst := make([]byte, ReadContentSize)
+	offset := int64(cht.object.Size - ReadContentSize)
+	cht.cacheHandle.isSequential = true
+	cht.cacheHandle.cacheFileForRangeRead = true
```
sethiay, line 828: let's keep to cacheFileForRangeRead = false in this case. This will ensure that downloads are triggered in case of Sequential reads and Parallel downloads.

internal/cache/file/cache_handle_test.go
ankitaluthra1: done

internal/cache/file/cache_handle_test.go
ankitaluthra1: Added

internal/cache/file/cache_handle_test.go
ankitaluthra1: Yes, it should be false , nice catch !!!

internal/cache/file/cache_handle_test.go
ankitaluthra1: Done

sethiay (approved):

PR #2028, opened 2024-06-18
Set crc32 when converting minobject to object

vadlakondaswetha:
### Description
Set crc32 when converting minobject to object

NA

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/storage/storageutil/object_attrs.go
+++ b/internal/storage/storageutil/object_attrs.go
@@ -217,5 +217,6 @@ func ConvertMinObjectToObject(m *gcs.MinObject) *gcs.Object {
 		Updated:         m.Updated,
 		Metadata:        m.Metadata,
 		ContentEncoding: m.ContentEncoding,
+		CRC32C:          m.CRC32C,
 	}
 }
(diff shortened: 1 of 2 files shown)


ashmeenkaur (approved):

internal/storage/storageutil/object_attrs.go
```diff
@@ -217,5 +217,6 @@ func ConvertMinObjectToObject(m *gcs.MinObject) *gcs.Object {
 		Updated:         m.Updated,
 		Metadata:        m.Metadata,
 		ContentEncoding: m.ContentEncoding,
+		CRC32C:          m.CRC32C,
```
sethiay, line 220: Do we know the reason of why CRC32 attribute is a pointer and not the value itself like Generation ?

sethiay (approved):

internal/storage/storageutil/object_attrs.go
vadlakondaswetha: Not sure. Its defined jacobsa/fuse. Never looked into it.

sethiay: I think you meant it was defined in jacobsa/gcloud and we took the code from there. Not a blocker for this PR but I think we can understand and change in future.

PR #2032, opened 2024-06-18
Implement limit on max-concurrency.

kislaykishore:
### Description
Limit max-concurrency of parallel downloads across jobs using a semaphore.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/downloader.go
+++ b/internal/cache/file/downloader/downloader.go
@@ -15,6 +15,7 @@
 package downloader
 
 import (
+	"math"
 	"os"
 
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/data"
@@ -23,6 +24,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/locker"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"golang.org/x/sync/semaphore"
 )
 
 // JobManager is responsible for maintaining, getting and removing file download
@@ -55,12 +57,17 @@ type JobManager struct {
 	// concatenation of bucket name, "/", and object name. e.g. object path for an
 	// object named "a/b/foo.txt" in bucket named "test_bucket" would be
 	// "test_bucket/a/b/foo.txt"
-	jobs map[string]*Job
-	mu   locker.Locker
+	jobs               map[string]*Job
+	mu                 locker.Locker
+	uberConcurrencySem *semaphore.Weighted
 }
 
 func NewJobManager(fileInfoCache *lru.Cache, filePerm os.FileMode, dirPerm os.FileMode,
 	cacheDir string, sequentialReadSizeMb int32, c *config.FileCacheConfig) (jm *JobManager) {
+	maxDownloadParallelism := int64(math.MaxInt64)
+	if c.MaxDownloadParallelism > 0 {
+		maxDownloadParallelism = int64(c.MaxDownloadParallelism)
+	}
 	jm = &JobManager{
 		fileInfoCache:        fileInfoCache,
 		filePerm:             filePerm,
@@ -68,6 +75,7 @@ func NewJobManager(fileInfoCache *lru.Cache, filePerm os.FileMode, dirPerm os.Fi
 		cacheDir:             cacheDir,
 		sequentialReadSizeMb: sequentialReadSizeMb,
 		fileCacheConfig:      c,
+		uberConcurrencySem:   semaphore.NewWeighted(maxDownloadParallelism),
 	}
 	jm.mu = locker.New("JobManager", func() {})
 	jm.jobs = make(map[string]*Job)
@@ -105,7 +113,7 @@ func (jm *JobManager) CreateJobIfNotExists(object *gcs.MinObject, bucket gcs.Buc
 	removeJobCallback := func() {
 		jm.removeJob(object.Name, bucket.Name())
 	}
-	job = NewJob(object, bucket, jm.fileInfoCache, jm.sequentialReadSizeMb, fileSpec, removeJobCallback, jm.fileCacheConfig)
+	job = NewJob(object, bucket, jm.fileInfoCache, jm.sequentialReadSizeMb, fileSpec, removeJobCallback, jm.fileCacheConfig, jm.uberConcurrencySem)
 	jm.jobs[objectPath] = job
 	return job
 }
--- a/internal/cache/file/downloader/downloader.go
+++ b/internal/cache/file/downloader/downloader.go
@@ -57,9 +57,9 @@ type JobManager struct {
 	// concatenation of bucket name, "/", and object name. e.g. object path for an
 	// object named "a/b/foo.txt" in bucket named "test_bucket" would be
 	// "test_bucket/a/b/foo.txt"
-	jobs               map[string]*Job
-	mu                 locker.Locker
-	uberConcurrencySem *semaphore.Weighted
+	jobs              map[string]*Job
+	mu                locker.Locker
+	maxParallelismSem *semaphore.Weighted
 }
 
 func NewJobManager(fileInfoCache *lru.Cache, filePerm os.FileMode, dirPerm os.FileMode,
@@ -75,7 +75,8 @@ func NewJobManager(fileInfoCache *lru.Cache, filePerm os.FileMode, dirPerm os.Fi
 		cacheDir:             cacheDir,
 		sequentialReadSizeMb: sequentialReadSizeMb,
 		fileCacheConfig:      c,
-		uberConcurrencySem:   semaphore.NewWeighted(maxDownloadParallelism),
+		// Shared between jobs - Limits the overall concurrency of downloads.
+		maxParallelismSem: semaphore.NewWeighted(maxDownloadParallelism),
 	}
 	jm.mu = locker.New("JobManager", func() {})
 	jm.jobs = make(map[string]*Job)
@@ -113,7 +114,7 @@ func (jm *JobManager) CreateJobIfNotExists(object *gcs.MinObject, bucket gcs.Buc
 	removeJobCallback := func() {
 		jm.removeJob(object.Name, bucket.Name())
 	}
-	job = NewJob(object, bucket, jm.fileInfoCache, jm.sequentialReadSizeMb, fileSpec, removeJobCallback, jm.fileCacheConfig, jm.uberConcurrencySem)
+	job = NewJob(object, bucket, jm.fileInfoCache, jm.sequentialReadSizeMb, fileSpec, removeJobCallback, jm.fileCacheConfig, jm.maxParallelismSem)
 	jm.jobs[objectPath] = job
 	return job
 }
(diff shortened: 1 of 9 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -86,7 +87,8 @@ type Job struct {
 	// is responsibility of JobManager to pass this function.
 	removeJobCallback func()
 
-	mu locker.Locker
+	mu                 locker.Locker
+	uberConcurrencySem *semaphore.Weighted
```
sethiay, line 91: Please add a comment here that this semaphore is shared between jobs.

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -85,7 +85,17 @@ func (job *Job) parallelDownloadObjectToFile(cacheFile *os.File) (err error) {
 			rangeStart := start
 			rangeEnd := min(rangeStart+parallelReadRequestSize, end)
 
+			if goRoutineIdx == 0 {
+				if err = job.uberConcurrencySem.Acquire(downloadErrGroupCtx, 1); err != nil {
+					logger.Tracef("Error while acquiring semaphore resource: %v", err)
```
sethiay, line 90: we can avoid having this log statement as the err returned from this function will be logged in job.go

internal/cache/file/downloader/parallel_downloads_job_test.go
```diff
@@ -33,10 +33,12 @@ type parallelDownloaderTest struct {
...
 
-func init() { RegisterTestSuite(&parallelDownloaderTest{}) }
+func init() {
+	RegisterTestSuite(&parallelDownloaderTest{})
+}
 
 func (dt *parallelDownloaderTest) SetUp(*TestInfo) {
 	dt.defaultFileCacheConfig = &config.FileCacheConfig{EnableParallelDownloads: true,
-		DownloadParallelismPerFile: 5, ReadRequestSizeMB: 2, EnableCrcCheck: true}
+		DownloadParallelismPerFile: 5, ReadRequestSizeMB: 2, EnableCrcCheck: true, MaxDownloadParallelism: 6}
```
sethiay, line 42: I think MaxDownloadParallelism doesn't matter here since we are anyways passing semaphore.NewWeighted(math.MaxInt64) in job_test.go. so, would recommend removing this field from here to avoid confusion in the past.

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/lru"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/util"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func getMinObject(t *testing.T, objectName string, bucket gcs.Bucket) gcs.MinObject {
```
sethiay, line 35: There is a similar function in job_test.go. Could we reuse that across the tests under downloader/ directory ? Similar for other functions. 

It's fine to have some duplication in unit tests but please see if we can put the common setup code together.

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
```diff
@@ -0,0 +1,199 @@
...
+	job2 := jm.CreateJobIfNotExists(&minObj2, bucket)
+	s1 := job1.subscribe(10 * util.MiB)
+	s2 := job2.subscribe(5 * util.MiB)
+	ctx := context.Background()
+
+	_, err1 := job1.Download(ctx, 10*util.MiB, false)
+	_, err2 := job2.Download(ctx, 5*util.MiB, false)
+
+	notif1, notif2 := false, false
+	timeout := time.After(1000 * time.Second)
```
sethiay, line 184: i think the timeout is too high. I think we should set it to something like 5-10s

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
```diff
@@ -0,0 +1,199 @@
...
+		select {
+		case <-s1:
+			notif1 = true
+		case <-s2:
+			notif2 = true
+		case <-timeout:
+			assert.Fail(t, "Test timed out")
+			return
+		}
+		if assert.Nil(t, err1) && assert.Nil(t, err2) && notif1 && notif2 {
```
sethiay, line 195: could we also test that the file being populated till the requested offset is correct (verify the portion of data being downloaded is correct) ? similar to how it is done in job.go

Same for the other test.

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
```diff
@@ -0,0 +1,199 @@
...
+		t.Fatalf("Error occurred while retrieving fileInfoKey: %v", err)
+	}
+	_, err = cache.Insert(fileInfoKeyName, fileInfo)
+	if err != nil {
+		t.Fatalf("Error occurred while inserting fileinfo into cache: %v", err)
+	}
+	return minObj
+}
+
+func TestParallelDownloads(t *testing.T) {
```
sethiay, line 105: would prefer to move these tests in downloader_test.go (just that the long name job_manager_parallel_downloads_test looks big)

sethiay (commented):
Logic looks good, some comments on the tests.

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
kislaykishore: I've moved that method into test_util and reused that everywhere.

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
kislaykishore: Sorry, it was unintended. I was using that to debug tests and missed to revert it before raising the PR.

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
kislaykishore: downloader_test uses some common setup and teardown which go against testing practices. I personally face readability challenges when I tried going through it. It's also counterintuitive that job_test.go is triggered by downloader_test.go. And downloader_test uses ogletest which we are moving away from.
So, I think a separate test is the better path here. I've renamed it to jm_parallel_downloads_test.go to make the name short.
In the longer run, we should fix the tests in downloader_test.go to be more readable and use less of common not-so-obvious setup and tear-down logic.

internal/cache/file/downloader/parallel_downloads_job.go
kislaykishore: Makes sense. Thanks!

internal/cache/file/downloader/jm_parallel_downloads_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/lru"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/util"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func createObjectInStore(t *testing.T, objPath string, objSize int64, bucket gcs.Bucket) []byte {
```
sethiay, line 36: I didn't get what is "Store" ? you meant Bucket ?

internal/cache/file/downloader/jm_parallel_downloads_test.go
```diff
@@ -0,0 +1,196 @@
...
+		readReqSize            int
+		maxDownloadParallelism int
+		downloadOffset         int64
+		expectedOffset         int64
+		subscribedOffset       int64
+	}{
+		{
+			name:                   "download in chunks of concurrency * readReqSize",
+			objectSize:             15 * util.MiB,
+			readReqSize:            4,
```
sethiay, line 106: please specify the downloadParallelismPerFile also here, so it's clear what its value is.

internal/cache/file/downloader/job.go
```diff
@@ -111,6 +115,7 @@ func NewJob(
 	fileSpec data.FileSpec,
 	removeJobCallback func(),
 	fileCacheConfig *config.FileCacheConfig,
+	uberConcurrencySem *semaphore.Weighted,
```
sethiay, line 118: nit: uberConcurrencySem -> maxParallelismSem

internal/cache/file/downloader/test_util.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		panic(fmt.Errorf("error occured while statting the object: %w", err))
+	}
+	if minObject != nil {
+		return *minObject
+	}
+	return gcs.MinObject{}
+}
+
+func verifyFileTillOffset(t *testing.T, spec data.FileSpec, offset int64, content []byte) {
```
sethiay, line 41: Shall we replace https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/cache/file/downloader/job_test.go#L97 with this function ?

internal/cache/file/downloader/job_manager_parallel_downloads_test.go
sethiay: Sure, we can work to make downloader_test.go and job_test.go separate and make it more readable. Fine with keeping how you have kept the tests now in this PR.

internal/cache/file/downloader/jm_parallel_downloads_test.go
kislaykishore: I was trying to shorten FakeStorage but bucket works!

internal/cache/file/downloader/test_util.go
kislaykishore: I don't know how to pass the *testing.T variable into this from job_test.go.

sethiay: Okay, then we can keep it like this and modify in next PR.

internal/cache/file/downloader/downloader.go
```diff
@@ -55,19 +57,25 @@ type JobManager struct {
...
-	mu   locker.Locker
+	jobs               map[string]*Job
+	mu                 locker.Locker
+	uberConcurrencySem *semaphore.Weighted
 }
 
 func NewJobManager(fileInfoCache *lru.Cache, filePerm os.FileMode, dirPerm os.FileMode,
 	cacheDir string, sequentialReadSizeMb int32, c *config.FileCacheConfig) (jm *JobManager) {
+	maxDownloadParallelism := int64(math.MaxInt64)
+	if c.MaxDownloadParallelism > 0 {
```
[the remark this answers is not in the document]
kislaykishore: As discussed we'll handle this in the config section. I'll raise a separate PR for this.

sethiay (approved):

PR #2036, opened 2024-06-19
Kernel List Cache: Composite Test

raj-prince:
### Description
- Added basic composite test to make sure kernel-list-cache is working fine.
- I'll add more tests in another PR - like creation or deletion of file inside a directory invalidates the cache.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/kernel_list_cache_inifinite_ttl_test.go
+++ b/internal/fs/kernel_list_cache_inifinite_ttl_test.go
@@ -1,0 +1,87 @@
+// Copyright 2024 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Please refer kernel_list_cache_test.go for the documentation.
+
+package fs_test
+
+import (
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type KernelListCacheTestWithInfiniteTtl struct {
+	suite.Suite
+	fsTest
+	KernelListCacheTestCommon
+}
+
+func (t *KernelListCacheTestWithInfiniteTtl) SetupSuite() {
+	t.serverCfg.ImplicitDirectories = true
+	t.serverCfg.MountConfig = &config.MountConfig{
+		ListConfig: config.ListConfig{
+			KernelListCacheTtlSeconds: -1,
+		},
+		MetadataCacheConfig: config.MetadataCacheConfig{
+			TtlInSeconds: 0,
+		},
+	}
+	t.serverCfg.RenameDirLimit = 10
+	t.fsTest.SetUpTestSuite()
+}
+
+func TestKernelListCacheTestInfiniteTtlSuite(t *testing.T) {
+	suite.Run(t, new(KernelListCacheTestWithInfiniteTtl))
+}
+
+// (a) First ReadDir() will be served from GCSFuse filesystem.
+// (b) Second ReadDir() will also be served from GCSFuse filesystem.
+func (t *KernelListCacheTestWithInfiniteTtl) TestKernelListCache_AlwaysCacheHit() {
+	// First read, kernel will cache the dir response.
+	f, err := os.Open(path.Join(mntDir, "explicitDir"))
+	assert.Nil(t.T(), err)
+	defer func() {
+		assert.Nil(t.T(), f.Close())
+	}()
+	names1, err := f.Readdirnames(-1)
+	assert.Nil(t.T(), err)
+	assert.Equal(t.T(), 2, len(names1))
+	assert.Equal(t.T(), "file1.txt", names1[0])
+	assert.Equal(t.T(), "file2.txt", names1[1])
+	err = f.Close()
+	assert.Nil(t.T(), err)
+	// Adding one object to make sure to change the ReadDir() response.
+	assert.Nil(t.T(), t.createObjects(map[string]string{
+		"explicitDir/file3.txt": "123456",
+	}))
+	defer t.deleteObjectOrFail("explicitDir/file3.txt")
+	// Advancing time by 5 years (157800000 seconds).
+	cacheClock.AdvanceTime(157800000 * time.Second)
+
+	// No invalidation since infinite ttl.
+	f, err = os.Open(path.Join(mntDir, "explicitDir"))
+	assert.Nil(t.T(), err)
+	names2, err := f.Readdirnames(-1)
+
+	assert.Nil(t.T(), err)
+	assert.Equal(t.T(), 2, len(names2))
+	assert.Equal(t.T(), "file1.txt", names2[0])
+	assert.Equal(t.T(), "file2.txt", names2[1])
+}
--- a/internal/fs/kernel_list_cache_inifinite_ttl_test.go
+++ b/internal/fs/kernel_list_cache_inifinite_ttl_test.go
@@ -36,7 +36,7 @@ type KernelListCacheTestWithInfiniteTtl struct {
 func (t *KernelListCacheTestWithInfiniteTtl) SetupSuite() {
 	t.serverCfg.ImplicitDirectories = true
 	t.serverCfg.MountConfig = &config.MountConfig{
-		ListConfig: config.ListConfig{
+		FileSystemConfig: config.FileSystemConfig{
 			KernelListCacheTtlSeconds: -1,
 		},
 		MetadataCacheConfig: config.MetadataCacheConfig{
--- a/internal/fs/kernel_list_cache_inifinite_ttl_test.go
+++ b/internal/fs/kernel_list_cache_inifinite_ttl_test.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
 
@@ -62,7 +63,7 @@ func (t *KernelListCacheTestWithInfiniteTtl) TestKernelListCache_AlwaysCacheHit(
 	}()
 	names1, err := f.Readdirnames(-1)
 	assert.Nil(t.T(), err)
-	assert.Equal(t.T(), 2, len(names1))
+	require.Equal(t.T(), 2, len(names1))
 	assert.Equal(t.T(), "file1.txt", names1[0])
 	assert.Equal(t.T(), "file2.txt", names1[1])
 	err = f.Close()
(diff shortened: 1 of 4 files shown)


internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+		"implicitDir/file1.txt": "-1234556789",
+		"implicitDir/file2.txt": "kdfkdj9",
+	}))
+}
+
+func (t *KernelListCacheTestCommon) deleteObjectOrFail(objectName string) {
+	assert.Nil(t.T(), t.deleteObject(objectName))
+}
+
+type KernelListCacheTestWithPositiveTtl struct {
```
Tulsishah, line 96: Creating two test suites within the same file can reduce readability and make the code more confusing. 

Also @kislaykishore suggested me, for better readability in Go tests, it's preferable to initialize necessary variables and resources directly within each test function, rather than using separate setup and teardown methods. This approach keeps each test more self-contained and easier to understand.

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	assert.Nil(t.T(), err)
+	defer func() {
+		assert.Nil(t.T(), f.Close())
+	}()
+	names1, err := f.Readdirnames(-1)
+	assert.Nil(t.T(), err)
+	assert.Equal(t.T(), 2, len(names1))
+	assert.Equal(t.T(), names1[0], "file1.txt")
+	assert.Equal(t.T(), names1[1], "file2.txt")
+	err = f.Close()
```
Tulsishah, line 130: Why do we need to close again if we are dong it in defer?

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+const (
+	kernelListCacheTtlSeconds = 1000
+)
+
+type KernelListCacheTestCommon struct {
```
ashmeenkaur, line 41: Is KernelListCacheTestCommon created because we want to add more test suites to this file?

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	// Adding one object to make sure to change the ReadDir() response.
+	assert.Nil(t.T(), t.createObjects(map[string]string{
+		"explicitDir/file4.txt": "123456",
+	}))
+	defer t.deleteObjectOrFail("explicitDir/file4.txt")
+	// Advancing the time within ttl.
+	cacheClock.AdvanceTime(kernelListCacheTtlSeconds * time.Second / 2)
+
+	// Within ttl, so will be served from kernel.
+	f, err = os.Open(path.Join(mntDir, "explicitDir"))
```
Tulsishah, line 235: Have we closed this f? same for all the other places

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+const (
+	kernelListCacheTtlSeconds = 1000
+)
+
+type KernelListCacheTestCommon struct {
+	suite.Suite
+	fsTest
+}
+
+func (t *KernelListCacheTestCommon) SetupSuite() {
```
ashmeenkaur, line 46: I think this is unnecessary, WDYT?

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+type KernelListCacheTestWithPositiveTtl struct {
+	suite.Suite
+	fsTest
+	KernelListCacheTestCommon
+}
+
+func (t *KernelListCacheTestWithPositiveTtl) SetupSuite() {
+	t.serverCfg.ImplicitDirectories = true
```
Tulsishah, line 103: Just to get more info: will it work with ImplicitDirectories = false?

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	t.serverCfg.RenameDirLimit = 10
+	t.fsTest.SetUpTestSuite()
+}
+
+func (t *KernelListCacheTestCommon) SetupTest() {
+	t.createFilesAndDirStructureInBucket()
+	cacheClock.SetTime(time.Date(2015, 4, 5, 2, 15, 0, 0, time.Local))
+}
+
+func (t *KernelListCacheTestCommon) TearDownTest() {
```
ashmeenkaur, line 61: I think we should delete the files directories created by SetupTest here.
There are error logs like the following on running tests:
```
{"timestamp":{"seconds":1719562345,"nanos":424749656},"severity":"ERROR","message":"fuse_errors: *fuseops.RmDirOp error: directory not empty"}
{"timestamp":{"seconds":1719562345,"nanos":430420365},"severity":"ERROR","message":"fuse_errors: *fuseops.RmDirOp error: directory not empty"}
```

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	err = f.Close()
+	assert.Nil(t.T(), err)
+	// Adding one object to make sure to change the ReadDir() response.
+	assert.Nil(t.T(), t.createObjects(map[string]string{
+		"explicitDir/file3.txt": "123456",
+	}))
+	defer t.deleteObjectOrFail("explicitDir/file3.txt")
+
+	// Zero ttl, means readdir will always be served from gcsfuse.
+	f, err = os.Open(path.Join(mntDir, "explicitDir"))
```
Tulsishah, line 344: Sorry.. I did not get.. how are we separating two tests, 1. Served from gcsfuse 2. Serve from kernal cache.  here?

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+
+type KernelListCacheTestWithPositiveTtl struct {
+	suite.Suite
+	fsTest
+	KernelListCacheTestCommon
+}
+
+func (t *KernelListCacheTestWithPositiveTtl) SetupSuite() {
+	t.serverCfg.ImplicitDirectories = true
+	t.serverCfg.MountConfig = &config.MountConfig{
```
ashmeenkaur, line 104: does metadata cache ttl effects results? If not, can we explicitly set to 0 here?

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	// First read, kernel will cache the dir response.
+	f, err := os.Open(path.Join(mntDir, "explicitDir"))
+	assert.Nil(t.T(), err)
+	defer func() {
+		assert.Nil(t.T(), f.Close())
+	}()
+	names1, err := f.Readdirnames(-1)
+	assert.Nil(t.T(), err)
+	assert.Equal(t.T(), 2, len(names1))
+	assert.Equal(t.T(), names1[0], "file1.txt")
```
ashmeenkaur, line 128: nit: expected on left side. In other tests too.

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// 2nd read, ReadDir() will be served from page-cache, that means no change in
+	// response.
+	f, err = os.Open(path.Join(mntDir, "explicitDir"))
+	assert.Nil(t.T(), err)
+	names2, err := f.Readdirnames(-1)
+
+	assert.Nil(t.T(), err)
+	assert.Equal(t.T(), 2, len(names2))
+	assert.Equal(t.T(), names1[0], names2[0])
```
ashmeenkaur, line 148: Nit: can we explicitly compare with constants/strings here instead of variables to make the test more readable. same for other tests

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,353 @@
...
+	err = f.Close()
+	assert.Nil(t.T(), err)
+	// Adding one object to make sure to change the ReadDir() response.
+	assert.Nil(t.T(), t.createObjects(map[string]string{
+		"explicitDir/file3.txt": "123456",
+	}))
+	defer t.deleteObjectOrFail("explicitDir/file3.txt")
+	// Advancing the clock within time.
+	cacheClock.AdvanceTime(kernelListCacheTtlSeconds * time.Second / 2)
+
```
ashmeenkaur, line 139: What happens if I stat file3 via GCSFuse? Will it show up in cache?

internal/fs/kernel_list_cache_test.go
ashmeenkaur: +1, in addition to this, can we add some tests for implicit dir cache validation too?

internal/fs/kernel_list_cache_test.go
kislaykishore: Here is the [internal link](https://g3doc.corp.google.com/go/g3doc/style/best-practices.md?cl=head#t-common-setup-scope) regarding the guidance on test setup.

internal/fs/kernel_list_cache_test.go
raj-prince: Yes, please check the defer statement. The last `f` is close as part of defer statement.

internal/fs/kernel_list_cache_test.go
raj-prince: Please check the documentation at the top of this file and let me know, if something is unclear.

internal/fs/kernel_list_cache_test.go
raj-prince: We are closing the last `f` via defer statement.

internal/fs/kernel_list_cache_test.go
raj-prince: Great point. I hadn't tested before. (Not advisable for workload containing read/write from multiple nodes).
Stat doesn't effect the kernel cache and next readdir() within ttl is served from the kernel-cache.
Only creation or deletion of files or folders evicts the kernel cache for their parent directory.

internal/fs/kernel_list_cache_test.go
raj-prince: Created different files for different suites.

Still debatable, 
The above link says, we should use the setup thing only for required test-methods, not for every test-method.
In our case, setup is required by all the test-methods in the suites. Not convinced here, maybe we can discuss offline.

internal/fs/kernel_list_cache_test.go
raj-prince: Yes, this will work for implicitDir = false.

So, whatever GCSFuse will return to kernel as part of readdir() response, that will be cached in the page-cache. Validated this.

Added the cache hit/miss case for implicit-dir.

Tulsishah (dismissed):

internal/fs/kernel_list_cache_inifinite_ttl_test.go
```diff
@@ -0,0 +1,87 @@
...
+func (t *KernelListCacheTestWithInfiniteTtl) TestKernelListCache_AlwaysCacheHit() {
+	// First read, kernel will cache the dir response.
+	f, err := os.Open(path.Join(mntDir, "explicitDir"))
+	assert.Nil(t.T(), err)
+	defer func() {
+		assert.Nil(t.T(), f.Close())
+	}()
+	names1, err := f.Readdirnames(-1)
+	assert.Nil(t.T(), err)
+	assert.Equal(t.T(), 2, len(names1))
```
ashmeenkaur, line 65: nit: require.Equal when you want to stop test execution from advancing further

internal/fs/kernel_list_cache_test.go
```diff
@@ -0,0 +1,325 @@
...
+		"implicitDir/file1.txt": "-1234556789",
+		"implicitDir/file2.txt": "kdfkdj9",
+	}
+}
+func (t *KernelListCacheTestCommon) createFilesAndDirStructureInBucket() {
+	assert.Nil(t.T(), t.createObjects(getFilesAndDirStructureObjects()))
+}
+
+func (t *KernelListCacheTestCommon) deleteFilesAndDirStructureInBucket() {
+	filesOrDirStructure := getFilesAndDirStructureObjects()
```
ashmeenkaur, line 88: Nit:
```suggestion
	filesAndDirStructure := getFilesAndDirStructureObjects()
```

ashmeenkaur (dismissed):

ashmeenkaur (approved):

PR #2041, opened 2024-06-20
adds get folder method in bucket.go

ankitaluthra1:
### Description
integrates with get folder method of control client for all buckets (except fast stat bucket - next PR).
it is not called from anywhere, hence no change in existing functionality.

NA

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/prefix_bucket_test.go
+++ b/internal/gcsx/prefix_bucket_test.go
@@ -396,3 +396,25 @@ func (t *PrefixBucketTest) DeleteObject() {
 	var notFoundErr *gcs.NotFoundError
 	ExpectTrue(errors.As(err, &notFoundErr))
 }
+
+func (t *PrefixBucketTest) GetFolder_Prefix() {
+	var err error
+
+	// Create a few objects.
+	err = storageutil.CreateObjects(
+		t.ctx,
+		t.wrapped,
+		map[string][]byte{
+			"something": []byte(""),
+		})
+
+	AssertEq(nil, err)
+
+	result, err := t.bucket.GetFolder(
+		t.ctx,
+		"something")
+
+	AssertEq(nil, err)
+	AssertEq("projects/_/buckets/some_bucket/folders/foo_something", result.GetName())
+
+}
--- a/internal/gcsx/prefix_bucket_test.go
+++ b/internal/gcsx/prefix_bucket_test.go
@@ -400,7 +400,7 @@ func (t *PrefixBucketTest) DeleteObject() {
 func (t *PrefixBucketTest) GetFolder_Prefix() {
 	var err error
 
-	// Create a few objects.
+	// Replace the use of CreateObject with CreateFolder once the CreateFolder API has been successfully implemented.
 	err = storageutil.CreateObjects(
 		t.ctx,
 		t.wrapped,
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -513,7 +513,7 @@ func (b *bucketHandle) GetFolder(ctx context.Context, folderName string) (*contr
 	var callOptions []gax.CallOption
 
 	folder, err := b.controlClient.GetFolder(ctx, &controlpb.GetFolderRequest{
-		Name: folderName,
+		Name: "projects/_/buckets/" + b.bucketName + "/folders/" + folderName,
 	}, callOptions...)
 
 	if err != nil {
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -518,10 +518,9 @@ func (b *bucketHandle) GetFolder(ctx context.Context, folderName string) (*contr
 
 	if err != nil {
 		err = fmt.Errorf("error getting metadata for folder: %s, %w", folderName, err)
-		return nil, err
 	}
 
-	return folder, nil
+	return folder, err
 }
 
 func isStorageConditionsNotEmpty(conditions storage.Conditions) bool {
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -509,6 +509,21 @@ func (b *bucketHandle) getStorageLayout() (*controlpb.StorageLayout, error) {
 	return stoargeLayout, err
 }
 
+func (b *bucketHandle) GetFolder(ctx context.Context, folderName string) (*controlpb.Folder, error) {
+	var callOptions []gax.CallOption
+
+	folder, err := b.controlClient.GetFolder(ctx, &controlpb.GetFolderRequest{
+		Name: folderName,
+	}, callOptions...)
+
+	if err != nil {
+		err = fmt.Errorf("error getting metadata for folder: %s, %w", folderName, err)
+		return nil, err
+	}
+
+	return folder, nil
+}
+
 func isStorageConditionsNotEmpty(conditions storage.Conditions) bool {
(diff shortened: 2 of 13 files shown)


internal/storage/mock_bucket.go
```diff
@@ -345,3 +346,26 @@ func (m *mockBucket) UpdateObject(p0 context.Context, p1 *gcs.UpdateObjectReques
...
+	_, file, line, _ := runtime.Caller(1)
+
+	// Hand the call off to the controller, which does most of the work.
+	retVals := m.controller.HandleMethodCall(
+		m,
+		"GetFolder",
+		file,
+		line,
+		[]interface{}{})
+	if len(retVals) != 1 {
```
Tulsishah, line 363: Shouldn't it be 2? as we are getting two return values?

internal/storage/mock_bucket.go
```diff
@@ -345,3 +346,26 @@ func (m *mockBucket) UpdateObject(p0 context.Context, p1 *gcs.UpdateObjectReques
...
+		panic(fmt.Sprintf("mockBucket.GetFolder: invalid return values: %v", retVals))
+	}
+	// o0 string
+	if retVals[0] != nil {
+		err = retVals[0].(error)
```
Tulsishah, line 368: There will be two values also we should use o0 and o1 to keep consistency.

internal/storage/mock_control_client.go
```diff
@@ -43,3 +43,15 @@ func (m *MockStorageControlClient) DeleteFolder(ctx context.Context,
 	args := m.Called(ctx, req, opts)
 	return args.Error(0)
 }
+
+func (m *MockStorageControlClient) GetFolder(ctx context.Context,
+	req *controlpb.GetFolderRequest,
+	opts ...gax.CallOption) (*controlpb.Folder, error) {
+	args := m.Called(ctx, req, opts)
+
+	if folder, ok := args.Get(0).(*controlpb.Folder); ok {
```
Tulsishah, line 52: Why are we writing this logic here? Shouldn't we just return whatever we get from the main method? Otherwise, we will not be able to properly check if the error is nil in the case of a successful get object.

Tulsishah (commented):
We should not put internal ticket id externally.
PTAL on that also.

internal/storage/mock_bucket.go
```diff
@@ -345,3 +346,26 @@ func (m *mockBucket) UpdateObject(p0 context.Context, p1 *gcs.UpdateObjectReques
 
 	return
 }
+
+func (m *mockBucket) GetFolder(
+	ctx context.Context,
+	prefix string) (folder *controlpb.Folder, err error) {
```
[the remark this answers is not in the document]
ankitaluthra1: did not find o0, o1 readable as they are not descriptive, any reason chaning it to this variable name

internal/storage/mock_control_client.go
ankitaluthra1: This is needed, as in positive scenario error is not returned , mock method should also return nil and in folder not found scenario, error is returned, but folder is not returned, hence single line cannot assert both

internal/storage/bucket_handle_test.go
```diff
@@ -1291,3 +1293,39 @@ func (testSuite *BucketHandleTest) TestDeleteFolderWhenFolderNotExistForHierarch
...
+	ctx := context.Background()
+	mockClient := new(MockStorageControlClient)
+	folderPath := "projects/_/buckets/" + TestBucketName + "/folders/random"
+
+	mockClient.On("GetFolder", ctx, &controlpb.GetFolderRequest{Name: folderPath}, mock.Anything).
+		Return(nil, status.Error(codes.NotFound, "folder not found"))
+	testSuite.bucketHandle.controlClient = mockClient
+	testSuite.bucketHandle.bucketType = gcs.Hierarchical
+
+	result, err := testSuite.bucketHandle.GetFolder(ctx, "random")
```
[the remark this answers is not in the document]
ankitaluthra1: done

internal/storage/mock_bucket.go
ankitaluthra1: done

internal/storage/mock_bucket.go
ankitaluthra1: updated

internal/storage/mock_bucket.go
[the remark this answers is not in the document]
ankitaluthra1: kept o0, o1 for now, can refactor whole class in next PR to make it consitent

internal/gcsx/prefix_bucket_test.go
```diff
@@ -396,3 +396,25 @@ func (t *PrefixBucketTest) DeleteObject() {
 	var notFoundErr *gcs.NotFoundError
 	ExpectTrue(errors.As(err, &notFoundErr))
 }
+
+func (t *PrefixBucketTest) GetFolder_Prefix() {
+	var err error
+
+	// Create a few objects.
```
Tulsishah, line 403: nit: Write comment
"Replace the use of CreateObject with CreateFolder once the CreateFolder API has been successfully implemented."

internal/storage/bucket_handle.go
```diff
@@ -509,6 +509,21 @@ func (b *bucketHandle) getStorageLayout() (*controlpb.StorageLayout, error) {
...
 }
 
+func (b *bucketHandle) GetFolder(ctx context.Context, folderName string) (*controlpb.Folder, error) {
+	var callOptions []gax.CallOption
+
+	folder, err := b.controlClient.GetFolder(ctx, &controlpb.GetFolderRequest{
+		Name: "projects/_/buckets/" + b.bucketName + "/folders/" + folderName,
+	}, callOptions...)
+
+	if err != nil {
```
Tulsishah, line 519: Should we just return `folder, err` and catch the err where ever we call this method like `GetFolder: error` ?

internal/storage/mock_control_client.go
Tulsishah: May be we can write same comment so in future we have reference why we did?

Tulsishah (changes requested):
LGTM from my side except few minor changes.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -294,3 +295,18 @@ func (b *fastStatBucket) StatObjectFromGcs(ctx context.Context,
...
+	ctx context.Context,
+	prefix string) (folder *controlpb.Folder, err error) {
+	// Fetch the listing.
+	folder, err = b.wrapped.GetFolder(ctx, prefix)
+	if err != nil {
+		return
+	}
+
+	// TODO: add folder metadata in stat cache
+	// b.insertFolder(folder)
```
Tulsishah, line 309: It should happen in this PR only ?

internal/storage/bucket_handle.go
ankitaluthra1: done

internal/storage/caching/fast_stat_bucket.go
ankitaluthra1: trying to make small PRs, since not adding this change is not breaking any existing functionality, will be in next PR where Get folder response is stat cached

Tulsishah (approved):

charith87 (approved):

PR #2042, opened 2024-06-20
[Config-CLI Parity][Fix] Populate config object when config file is not provided

ashmeenkaur:
### Description
Config object should be populated with only CLI flags in cases where config file is not provided.
Also resolves the config file path before usage.

NA

### Testing details
1. Manual - Tested manually and printed the config object to confirm it is working as intended
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -58,14 +58,14 @@ func init() {
 }
 
 func initConfig() {
-	if cfgFile == "" {
-		return
+	if cfgFile != "" {
+		viper.SetConfigFile(cfgFile)
+		viper.SetConfigType("yaml")
+		if err := viper.ReadInConfig(); err != nil {
+			logger.Fatal("error while reading the config: %v", err)
+		}
 	}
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -58,14 +59,18 @@ func init() {
...
 func initConfig() {
-	if cfgFile == "" {
-		return
-	}
-	viper.SetConfigFile(cfgFile)
-	viper.SetConfigType("yaml")
-	if err := viper.ReadInConfig(); err != nil {
-		logger.Fatal("error while reading the config: %v", err)
+	if cfgFile != "" {
+		cfgFile, err := util.GetResolvedPath(cfgFile)
```
kislaykishore, line 63: Currently, the helpdoc for cfgFile says that it accepts only absolute path.  Should we change that?

kislaykishore (approved):

kislaykishore (approved):

raj-prince (approved):

PR #2044, opened 2024-06-21
[Config CLI Parity] add squash to config structure tags to make decoding easier

ashmeenkaur:
### Description
Add "squash" to config structure tags to make decoding easier from legacy flagStorage & mountConfig.
This is a temporary change to make transition easier to new CLI and config flags and will be reverted post transition.
Reason to revert this in future - This change doesn't allow having leaf config with conflicting names.

Reference: https://pkg.go.dev/github.com/mitchellh/mapstructure#hdr-Embedded_Structs_and_Squashing
Open issue on mapstructure library which is blocking the change: https://github.com/mitchellh/mapstructure/issues/356

NA

### Testing details
1. Manual - Manually verified that viper configs and cobra flags are still working as intended
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -29,37 +29,37 @@ type Config struct {
 
 	CacheDir ResolvedPath `yaml:"cache-dir"`
 
-	Debug DebugConfig `yaml:"debug"`
+	Debug DebugConfig `yaml:"debug,squash"`
 
 	EnableHns bool `yaml:"enable-hns"`
 
-	FileCache FileCacheConfig `yaml:"file-cache"`
+	FileCache FileCacheConfig `yaml:"file-cache,squash"`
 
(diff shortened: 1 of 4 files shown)


kislaykishore (approved):
Please verify that the logic works for two levels of nesting. As per [this](https://pkg.go.dev/github.com/mitchellh/mapstructure#hdr-Embedded_Structs_and_Squashing), if a structure is *not* nested, that's when we can include the embedded struct to the tag value. It doesn't explicitly talks about the case where the struct is nested.

ashmeenkaur:
> s _not_ nested, that's when we can include the embedded struct to the tag value. It doesn't explicitly talk about the case where the struct is nested.

Verified that it is working for nested structs too

raj-prince (approved):
Relying on Kislay's review.

PR #2047, opened 2024-06-21
Use default MaxDownloadParallelism based on number of CPU cores

sethiay:
### Description
Use default MaxDownloadParallelism based on number of CPU cores. Also set a minimum value of 16 in case the cores of CPU is less (2 or 4).

Using default 2 * number of cores as with the testing so far with a workload, this value looks good.

And couple of more things to tell upfront :
1. This is based on CPU cores count of machine, so on GKE this will translate to machine's CPU cores and not on the CPU limit on pod. We are fine here because pod limit <= machine CPU cores. So, we are just assigning higher value to max parallel downloads there.
2. It's possible to mount multiple GCSFuse mounts on single GCE VM/GKE node. In that case, one GCSFuse default is not affected by the other.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/config/config_util.go
+++ b/internal/config/config_util.go
@@ -114,6 +114,6 @@ func ListCacheTtlSecsToDuration(secs int64) time.Duration {
 	return time.Duration(secs * int64(time.Second))
 }
 
-func DefaultMaxDownloadParallelism() int {
+func DefaultMaxParallelDownloads() int {
 	return max(16, 2*runtime.NumCPU())
 }
--- a/internal/config/config_util.go
+++ b/internal/config/config_util.go
@@ -17,6 +17,7 @@ package config
 import (
 	"fmt"
 	"math"
+	"runtime"
 	"time"
 )
 
@@ -111,4 +112,8 @@ func ListCacheTtlSecsToDuration(secs int64) time.Duration {
 	}
 
 	return time.Duration(secs * int64(time.Second))
+}
+
+func DefaultMaxDownloadParallelism() int {
+	return max(16, 2*runtime.NumCPU())
 }
--- a/internal/config/config_util_test.go
+++ b/internal/config/config_util_test.go
@@ -267,3 +267,7 @@ func Test_ListCacheTtlSecsToDuration_InvalidCall(t *testing.T) {
 	// Calling with invalid argument to trigger panic.
 	ListCacheTtlSecsToDuration(-3)
 }
+
+func Test_DefaultMaxDownloadParallelism(t *testing.T) {
+	assert.GreaterOrEqual(t, DefaultMaxDownloadParallelism(), 16)
+}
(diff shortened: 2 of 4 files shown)


kislaykishore:
@sethiay could you update the params.yaml as well? You'll need to make the following changes:
* Add "runtime" [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/config-gen/config.tpl#L21)
* Update the default [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/config-gen/params.yaml#L454) to "int64(math.Max(16, 2*runtime.NumCPU()))". Alternatively, you can create a function in cfg/ and reference it here.
* Regenerate the config.go by calling `go generate ./...`

internal/config/config_util.go
```diff
@@ -112,3 +113,7 @@ func ListCacheTtlSecsToDuration(secs int64) time.Duration {
 
 	return time.Duration(secs * int64(time.Second))
 }
+
+func DefaultMaxDownloadParallelism() int {
+	return max(16, 2*runtime.NumCPU())
```
charith87, line 118: Shouldnt the default come after testing. Do we want this to be default now?

sethiay: As discussed, I want to introduce the dependency on CPU cores, we can change to a better value based on our testing.

sethiay:
> @sethiay could you update the params.yaml as well? You'll need to make the following changes:
> 
> * Add "runtime" [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/config-gen/config.tpl#L21)
> * Update the default [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/config-gen/params.yaml#L454) to "int64(math.Max(16, 2*runtime.NumCPU()))". Alternatively, you can create a function in cfg/ and reference it here.
> * Regenerate the config.go by calling `go generate ./...`

As discussed offline, we will handle this in a separate PR.

internal/config/config_util.go
charith87: is this function a well known function to depend on it. is it in a standard package which will work in all VM's?

internal/config/config_util_test.go
```diff
@@ -267,3 +267,7 @@ func Test_ListCacheTtlSecsToDuration_InvalidCall(t *testing.T) {
 	// Calling with invalid argument to trigger panic.
 	ListCacheTtlSecsToDuration(-3)
 }
+
+func Test_DefaultMaxParallelDownloads(t *testing.T) {
```
charith87, line 271: I dont know how to add a test. But this is possibly checking for outliers like 0 or -1 which is also good. Is it possible to get the machine type ( only for the test ) and check if its as expected. ( like get machine name like n2-standard-32 and check 32 * 2 is the number. )

sethiay: I don't think that it's possible. We have to only rely on manual testing.

internal/config/config_util.go
charith87: Resolving for now based on chat comments that its a standard package in go

charith87 (approved):

PR #2048, opened 2024-06-21
Return error if the config file is invalid.

kislaykishore:
* Return error if config file is invalid
* Refactor the root command to make it testable.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


cmd/root.go
```diff
@@ -27,36 +27,34 @@ import (
...
+			return mountFn(configObj)
+		},
 	}
-}
-
-func init() {
 	cobra.OnInitialize(initConfig)
-	rootCmd.PersistentFlags().StringVar(&cfgFile, "config-file", "", "The path to the config file where all gcsfuse related config needs to be specified."+
-		"Refer to 'https://cloud.google.com/storage/docs/gcsfuse-cli#config-file' for possible configurations.")
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config-file", "", "Refer to 'https://cloud.google.com/storage/docs/gcsfuse-cli#config-file' for possible configurations.")
```
ashmeenkaur, line 51: Nit: is this intentional?

cmd/root.go
```diff
@@ -72,12 +70,11 @@ func initConfig() {
...
 
-	err := viper.Unmarshal(&configObj, viper.DecodeHook(cfg.DecodeHook()), func(decoderConfig *mapstructure.DecoderConfig) {
+	cfgErr = viper.Unmarshal(&configObj, viper.DecodeHook(cfg.DecodeHook()), func(decoderConfig *mapstructure.DecoderConfig) {
 		// By default, viper supports mapstructure tags for unmarshalling. Override that to support yaml tag.
 		decoderConfig.TagName = "yaml"
+		// Reject the config file if any of the fields in the YAML don't map to the struct.
+		decoderConfig.ErrorUnused = true
 	},
 	)
-	if err != nil {
```
ashmeenkaur, line 80: Nit: Why not log fatal error in this case too?

main.go
```diff
@@ -63,9 +65,21 @@ func main() {
 	// Make logging output better.
 	log.SetFlags(log.Ldate | log.Ltime | log.Lmicroseconds)
 	if strings.ToLower(os.Getenv("ENABLE_GCSFUSE_VIPER_CONFIG")) == "true" {
-		os.Args = convertToPosixArgs(os.Args)
-		cmd.Execute()
+		// TODO: implement the mount logic instead of simply returning nil.
+		rootCmd, err := cmd.NewRootCmd(func(config cfg.Config) error { return nil })
+		if err != nil {
+			exitOnError(err)
```
ashmeenkaur, line 71: We can use logger.Fatal here too?

kislaykishore: It seems logger.Fatal writes to stdout instead of stderr. I didn't want to change the failure behavior.

ashmeenkaur: When GCSFuse runs in foreground, logs go to stdout, when GCSFuse runs in background, logs go to sys logs and when log file is specified, logs go to log file. I think it is okay to write error log with logger.Fatal (to stdout) in this case as it has the benefit of capturing the log when GCSFuse is run in different mode. WDYT?

ashmeenkaur (approved):

PR #2049, opened 2024-06-24
Raise error if params.yaml contains invalid fields

kislaykishore:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/config-gen/parser.go
+++ b/tools/config-gen/parser.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"slices"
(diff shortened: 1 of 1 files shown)


tools/config-gen/parser.go
```diff
@@ -38,17 +39,17 @@ type Param struct {
...
 func parseParamsConfig() ([]Param, error) {
-	yamlFile, err := os.ReadFile(*paramsFile)
+	buf, err := os.ReadFile(*paramsFile)
 	if err != nil {
 		return nil, err
 	}
 	var paramsConfig []Param
-	err = yaml.Unmarshal(yamlFile, &paramsConfig)
-	if err != nil {
+	dec := yaml.NewDecoder(bytes.NewReader(buf))
```
ashmeenkaur, line 47: yaml.Unmarshal documentation says:
```
The type of the decoded values should be compatible with the respective values in out. If one or more values cannot be decoded due to a type mismatches, decoding continues partially until the end of the YAML content, and a *yaml.TypeError is returned with details for all missed values.
```
Ref: https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshal

Are we solving something different here?

kislaykishore: Maybe it handles type mismatches. But if you add a new field in the YAML, there is no error.

ashmeenkaur (approved):

PR #2061, opened 2024-06-24
Create new config object from legacy flags and config

ashmeenkaur:
### Description
Create new config object from legacy flags and legacy mount config.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA


cmd/legacy_param_converter_test.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	for _, tt := range populateConfigFromLegacyFlags {
+		t.T().Run(tt.testName, func(m *testing.T) {
+			testContext := &MockContext{isFlagSet: tt.isFlagSet}
+
+			resolvedConfig, err := PopulateNewConfigFromLegacyFlagsAndConfig(testContext, tt.legacyFlagStorage, tt.legacyMountConfig)
+
+			assert.Nil(t.T(), err)
+			assert.Equal(t.T(), tt.expectedConfig, resolvedConfig)
```
kislaykishore, line 332: Asserts just report errors - they don't immediately cause the control flow to exit from the test function. So, maybe we can assert the equality only if the error is nil.
```suggestion
			if assert.Nil(t.T(), err) {
			    assert.Equal(t.T(), tt.expectedConfig, resolvedConfig)
			}
```

kislaykishore (changes requested):

cmd/legacy_param_converter.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/cfg"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
+	"github.com/mitchellh/mapstructure"
+)
+
+// cliContext is abstraction over the IsSet() method of cli.Context, Specially
```
kislaykishore, line 26:
```suggestion
// cliContext is abstraction over the IsSet() method of cli.Context, specially
```

cmd/legacy_param_converter.go
kislaykishore, line 1: Nit: I think legacy_param_mapper.go is a better name for this file. Such tasks are usually called as mapping. Google "automapper" and I think you'll understand what I mean.

cmd/legacy_param_converter.go
```diff
@@ -0,0 +1,155 @@
...
+				resolvedConfig.Logging.Format = value.(string)
+			case "ignore-interrupts":
+				resolvedConfig.FileSystem.IgnoreInterrupts = value.(bool)
+			case "anonymous-access":
+				resolvedConfig.GcsAuth.AnonymousAccess = value.(bool)
+			case "kernel-list-cache-ttl-secs":
+				resolvedConfig.List.KernelListCacheTtlSecs = value.(int64)
+			}
+		}
+	}
```
kislaykishore, line 152: Refer go/tott/651
```suggestion
		if !c.IsSet(flagName) {
		  continue
		}
	switch flagName {
	case "log-file":
		resolvedConfig.Logging.FilePath = value.(cfg.ResolvedPath)
	case "log-format":
		resolvedConfig.Logging.Format = value.(string)
	case "ignore-interrupts":
		resolvedConfig.FileSystem.IgnoreInterrupts = value.(bool)
	case "anonymous-access":
		resolvedConfig.GcsAuth.AnonymousAccess = value.(bool)
	case "kernel-list-cache-ttl-secs":
		resolvedConfig.List.KernelListCacheTtlSecs = value.(int64)
		}
	}
```

cmd/legacy_param_converter.go
```diff
@@ -0,0 +1,155 @@
...
+		"log-format":                 resolvedConfig.Logging.Format,
+		"ignore-interrupts":          resolvedConfig.FileSystem.IgnoreInterrupts,
+		"anonymous-access":           resolvedConfig.GcsAuth.AnonymousAccess,
+		"kernel-list-cache-ttl-secs": resolvedConfig.List.KernelListCacheTtlSecs,
+	}
+
+	// Decoding config to the same config structure.
+	err = decoder.Decode(legacyConfig)
+	if err != nil {
+		return nil, fmt.Errorf("decoder.Decode(config): %v", err)
```
kislaykishore, line 133: It's recommended to use %w for errors - go/go-style/best-practices#error-percent-w
```suggestion
		return nil, fmt.Errorf("decoder.Decode(config): %w", err)
```

cmd/legacy_param_converter.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, fmt.Errorf("decoder.Decode(structuredFlags): %v", err)
+	}
+
+	// If config file is not present, no need to decode it.
+	if legacyConfig == nil || reflect.DeepEqual(*legacyConfig, config.MountConfig{}) {
+		return resolvedConfig, nil
+	}
+
+	// Save overlapping flags in a map to override the config value later.
+	overlapFlags := map[string]interface{}{
```
kislaykishore, line 122: How about getting rid of the map altogether like this:

if c.IsSet("log-file") {
  resolvedConfig.Logging.FilePath = cfg.ResolvedPath(resolvedConfig.Logging.FilePath)
}

and so on?

cmd/legacy_param_converter_test.go
```diff
@@ -0,0 +1,323 @@
...
+				},
+				GcsConnection: cfg.GcsConnectionConfig{
+					ClientProtocol: cfg.Protocol("http2"),
+				},
+			},
+		},
+	}
+
+	for _, tt := range populateConfigFromLegacyFlags {
+		t.Run(tt.testName, func(m *testing.T) {
```
kislaykishore, line 313: You can shadow the variable here.
```suggestion
		t.Run(tt.testName, func(t *testing.T) {
```

cmd/legacy_param_converter_test.go
```diff
@@ -0,0 +1,323 @@
...
+			testContext := &MockContext{isFlagSet: tt.isFlagSet}
+
+			resolvedConfig, err := PopulateNewConfigFromLegacyFlagsAndConfig(testContext, tt.legacyFlagStorage, tt.legacyMountConfig)
+
+			if assert.Nil(t, err) {
```
kislaykishore, line 318: If you don't want to shadow "t", using "m" is the better choice here since that's the testing.T that's been made available to your sub-test.

kislaykishore (changes requested):

cmd/legacy_param_converter.go
ashmeenkaur: The flag values in resolvedConfig will be over-witten by the  legacy mount config values so we need to store the resolved values from flags somewhere to over-write.

kislaykishore (changes requested):

cmd/legacy_param_mapper.go
```diff
@@ -0,0 +1,155 @@
...
+		case "ignore-interrupts":
+			resolvedConfig.FileSystem.IgnoreInterrupts = value.(bool)
+		case "anonymous-access":
+			resolvedConfig.GcsAuth.AnonymousAccess = value.(bool)
+		case "kernel-list-cache-ttl-secs":
+			resolvedConfig.List.KernelListCacheTtlSecs = value.(int64)
+		}
+	}
+	return resolvedConfig, nil
+}
```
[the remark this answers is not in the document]
ashmeenkaur: > The solution currently does type-casting that can potentially fail at runtime

Approach 1: I initially had the same thought and added appropriate checks but it was not unit testable as we are reading the values in the same method so I removed those checks. We can still do those checks but know that it won't be unit testable:
```
val, ok := value.(cfg.ResolvedPath)
if ok {
//assign value
}
```

Going by the logic of failing at runtime, even this method can fail:
```
func overrideWithFlag[T any](c cliContext, name string, toUpdate *T, updateValue T) {
	if !c.IsSet(name) {
		return
	}
	*toUpdate = updateValue
}
```

Approach 2: Alternatively, we can check for each flag but there will be code repetition:
```
if c.IsSet(name) {
 configPath = value (stored as var without type casting).
}
```

Let me know what you thing would be the best fit?

cmd/legacy_param_mapper.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, fmt.Errorf("mapstructure.NewDecoder: %w", err)
+	}
+	// Decoding flags.
+	err = decoder.Decode(structuredFlags)
+	if err != nil {
+		return nil, fmt.Errorf("decoder.Decode(structuredFlags): %w", err)
+	}
+
+	// If config file is not present, no need to decode it.
+	if legacyConfig == nil || reflect.DeepEqual(*legacyConfig, config.MountConfig{}) {
```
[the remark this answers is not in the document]
ashmeenkaur: Just realized that we will always have some values in the mount config (default values), so removed this code.

cmd/legacy_param_mapper.go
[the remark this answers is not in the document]
kislaykishore: As discussed offline, the generics implementation would verify at compile time that the types are incompatible. So, that's the way to go.

cmd/legacy_param_mapper_test.go
```diff
@@ -0,0 +1,347 @@
...
+			},
+			expectedErr: nil,
+		},
+	}
+
+	for _, tc := range populateConfigFromLegacyFlags {
+		t.Run(tc.testName, func(t *testing.T) {
+			resolvedConfig, err := PopulateNewConfigFromLegacyFlagsAndConfig(tc.mockCLICtx, tc.legacyFlagStorage, tc.legacyMountConfig)
+
+			require.Equal(t, tc.expectedErr, err)
```
kislaykishore, line 343: I think the usual way is to
if assert.Nil(...) {
  assert.Equal(..)
}

Curious why you have adopted this style.

cmd/legacy_param_mapper_test.go
```diff
@@ -0,0 +1,388 @@
...
+			resolvedConfig, err := PopulateNewConfigFromLegacyFlagsAndConfig(tc.mockCLICtx, tc.legacyFlagStorage, tc.legacyMountConfig)
+
+			if assert.Equal(t, tc.expectedErr, err) {
+				assert.Equal(t, tc.expectedConfig, resolvedConfig)
+			}
+		})
+	}
+}
+
+func TestOverrideWithFlag(t *testing.T) {
```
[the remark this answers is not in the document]
ashmeenkaur: will remove in follow up PR

PR #2063, opened 2024-06-25
Remove flakiness in parallel downloads unit tests

sethiay:
### Description
Remove flakiness in parallel downloads unit tests. Will add the tests back in new PR.

NA

### Testing details
1. Manual - Ran go test with count 10.
2. Unit tests - NA
3. Integration tests - NA


internal/cache/file/downloader/job_test.go
```diff
@@ -959,6 +959,7 @@ func (dt *downloaderTest) Test_validateCRC_WheContextIsCancelled() {
 	_, err := dt.job.Download(context.Background(), offset, true)
 	AssertEq(nil, err)
 	AssertEq(Downloading, dt.job.status.Name)
```
ashmeenkaur, line 961: Do we need to remove this?

ashmeenkaur (approved):

PR #2066, opened 2024-06-25
Add constraints on number of args

kislaykishore:
### Description
Limit the number of arguments that the root command can receive to the range [1, 2]
NA

### Testing details
1. Manual - NA
2. Unit tests - Added unit tests.
3. Integration tests - NA

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -40,7 +40,7 @@ func NewRootCmd(mountFn func(config cfg.Config) error) (*cobra.Command, error) {
 and access Cloud Storage buckets as local file systems. For a technical overview
 of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 		Version: getVersion(),
-		Args:    cobra.RangeArgs(1, 2),
+		Args:    cobra.RangeArgs(2, 3),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if cfgErr != nil {
 				return fmt.Errorf("error while parsing config: %w", cfgErr)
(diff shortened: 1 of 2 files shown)


cmd/root_test.go
```diff
@@ -46,3 +46,50 @@ func TestValidConfig(t *testing.T) {
...
+			name:        "Two args is okay",
+			args:        []string{"abc", "pqr"},
+			expectError: false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cmd, err := NewRootCmd(func(config cfg.Config) error { return nil })
+			if err != nil {
```
ashmeenkaur, line 81: Nit: you can alternatively use require.NotNil(err)

cmd/root_test.go
```diff
@@ -46,3 +46,50 @@ func TestValidConfig(t *testing.T) {
...
+				t.Fatalf("Error while creating the root command: %v", err)
+			}
+			cmd.SetArgs(tc.args)
+
+			err = cmd.Execute()
+
+			if tc.expectError {
```
ashmeenkaur, line 88: Can we pass expected errors in test cases:
`expectedErr: fmt.Errorf("accepts between %d and %d arg(s), received %d", 1, 2, 3),`

And do a normal assert:
assert.Equal(t, tc.expectedErr, err)

cmd/root.go
```diff
@@ -40,6 +40,7 @@ func NewRootCmd(mountFn func(config cfg.Config) error) (*cobra.Command, error) {
 and access Cloud Storage buckets as local file systems. For a technical overview
 of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 		Version: getVersion(),
+		Args:    cobra.RangeArgs(1, 2),
```
ashmeenkaur, line 43: Based on manual testing, it is also counting root command as one of the argument. So the range should be 2,3

Can you also make sure that this reflects in unit test?

kislaykishore: Good catch! Thanks!

cmd/root_test.go
kislaykishore: I think we can rely on cobra to return a sensible error and not do that validation ourselves. If a future version of cobra changes the error message, the test shouldn't break. I can still add if you insist.

ashmeenkaur (approved):

PR #2078, opened 2024-06-28
Use a viper instance for deserialization

kislaykishore:
Using a global viper instance could cause problems since the same viper is shared across the entire application and libraries. This can lead to races. So, use instance level viper.

NA

### Testing details
1. Manual - NA
2. Unit tests - Existing tests should suffice.
3. Integration tests - NA

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -61,6 +61,8 @@ of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 }
 
 func initConfig(v *viper.Viper) func() {
+	// Returning a function instead of using a package-level variable since it
+	// allows for the viper instance to be GC'ed.
 	return func() {
 		if cfgFile != "" {
 			cfgFile, err := util.GetResolvedPath(cfgFile)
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -48,35 +48,40 @@ of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
...
-		if err != nil {
-			logger.Fatal("error while resolving config-file path[%s]: %v", cfgFile, err)
-		}
-		viper.SetConfigFile(cfgFile)
-		viper.SetConfigType("yaml")
-		if err := viper.ReadInConfig(); err != nil {
-			logger.Fatal("error while reading the config: %v", err)
+func initConfig(v *viper.Viper) func() {
+	// Returning a function instead of using a package-level variable since it
+	// allows for the viper instance to be GC'ed.
```
ashmeenkaur, line 65: What is GC'ed? Garbage collected?

ashmeenkaur (approved):

cmd/root.go
kislaykishore: Yes

kislaykishore: Raised https://github.com/GoogleCloudPlatform/gcsfuse/pull/2079 for the fix.

PR #2086, opened 2024-07-01
upgrade dependencies

ashmeenkaur:
### Description
This PR is combination of the following dependabot PRs:
#2085
#2053
#2025

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
4. Perf tests - 
6XVKH895UBGQABY


raj-prince (approved):
LGTM.

Do we know the reason behind so much fluctuation in throughput of 1 GB, rand read?

ashmeenkaur:
> Do we know the reason behind so much fluctuation in throughput of 1 GB, rand read?

Based on past runs, the observed throughput for 1 GB random reads is typically in the range of 6XX-7XX MiB/s, so this fluctuation is likely a temporary issue.

PR #2089, opened 2024-07-01
Migrate a few flags to the new config

kislaykishore:
### Description
Migrate the following flags to the new config:
* app-name
* foreground
* rename-dir-limit
* billing-project
* limit-bytes-per-sec
* limit-ops-per-sec
* stackdriver-export-interval
* experimental-opentelemetry-collector-address
* experimental-enable-json-read
* temp-dir
* enable-nonexistent-type-cache
* file-mode
* dir-mode
* uid
* gid

NA

### Testing details
1. Manual - Tested manually to see that the flags are correctly being reflected.
2. Unit tests - Existing unit-tests in legacy_param_mapper.go should already cover it.
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main_test.go
+++ b/cmd/legacy_main_test.go
@@ -43,16 +43,18 @@ type MainTest struct {
 
 func (t *MainTest) TestCreateStorageHandle() {
 	flags := &flagStorage{
-		ClientProtocol:      mountpkg.HTTP1,
 		MaxConnsPerHost:     5,
 		MaxIdleConnsPerHost: 100,
 		HttpClientTimeout:   5,
 		MaxRetrySleep:       7,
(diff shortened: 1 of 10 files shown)


cmd/legacy_main_test.go
```diff
@@ -52,7 +51,9 @@ func (t *MainTest) TestCreateStorageHandle() {
 		AppName:             "app",
```
ashmeenkaur, line 51: AppName should also move to new config. Same for TestCreateStorageHandle_WithClientProtocolAsGRPC Test

kislaykishore: AppName wasn't required. So removed it.

ashmeenkaur (approved):

PR #2090, opened 2024-07-01
Code-gen URL params as pointer

kislaykishore:
Code-gen URL params as pointer types. This is in line with the URL params in the legacy config.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/decode_hook_test.go
+++ b/cfg/decode_hook_test.go
@@ -37,7 +37,7 @@ func bindFlag(t *testing.T, v *viper.Viper, key string, f *flag.Flag) {
 func TestParsingSuccess(t *testing.T) {
 	type TestConfig struct {
 		OctalParam       Octal
-		URLParam         url.URL
+		URLParam         *url.URL
 		BoolParam        bool
 		StringParam      string
 		IntParam         int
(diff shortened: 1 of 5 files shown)


cfg/decode_hook_test.go
```diff
@@ -37,7 +37,7 @@ func bindFlag(t *testing.T, v *viper.Viper, key string, f *flag.Flag) {
 func TestParsingSuccess(t *testing.T) {
 	type TestConfig struct {
 		OctalParam       Octal
-		URLParam         url.URL
+		URLParam         *url.URL
```
ashmeenkaur, line 40: https://github.com/GoogleCloudPlatform/gcsfuse/blob/8637ce1941e0234e142b4d9470e2e7ab4318ca41/cfg/decode_hook.go#L33

 I think we should change this to `if t != reflect.TypeOf(&url.URL{}) {` too.

kislaykishore: This doesn't seem necessary. The tests would have failed otherwise. I also debugged this manually by adding breakpoints to confirm that this predicate was still returning true for URL decoding.

kislaykishore: Discussed with @ashmeenkaur who mentioned that although it works, the current predicate causes an inefficiency and requires multiple attempts to decode the URL: once as a pointer and subsequently as an object. So, fixed the predicate to check for the pointer type.

ashmeenkaur (approved):

PR #2095, opened 2024-07-03
Migrate most flags to the new config

kislaykishore:
### Description
Migrate most flags to use the new config
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cmd/flags_test.go
+++ b/cmd/flags_test.go
@@ -299,18 +299,7 @@ func (t *FlagsTest) TestResolvePathForTheFlagsInContext() {
 	assert.Equal(t.T(), nil, err)
 }
 
-func (t *FlagsTest) TestValidateFlagsForValidSequentialReadSizeAndHTTP1ClientProtocol() {
-	flags := &flagStorage{
-		SequentialReadSizeMb:                10,
-		ExperimentalMetadataPrefetchOnMount: config.DefaultExperimentalMetadataPrefetchOnMount,
-	}
-
-	err := validateFlags(flags)
-
-	assert.Equal(t.T(), nil, err)
-}
-
-func (t *FlagsTest) TestValidateFlagsForZeroSequentialReadSizeAndValidClientProtocol() {
+func (t *FlagsTest) TestValidateFlagsForZeroSequentialReadSize() {
 	flags := &flagStorage{
 		SequentialReadSizeMb:                0,
 		ExperimentalMetadataPrefetchOnMount: config.DefaultExperimentalMetadataPrefetchOnMount,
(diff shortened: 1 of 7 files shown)


cmd/flags_test.go
```diff
@@ -302,7 +302,6 @@ func (t *FlagsTest) TestResolvePathForTheFlagsInContext() {
 func (t *FlagsTest) TestValidateFlagsForValidSequentialReadSizeAndHTTP1ClientProtocol() {
```
ashmeenkaur, line 302: Rename test (same for other tests)

cmd/flags_test.go
```diff
@@ -337,22 +334,9 @@ func (t *FlagsTest) TestValidateFlagsForSequentialReadSizeGreaterThan1024AndVali
...
-		ClientProtocol:                      mountpkg.ClientProtocol("http4"),
-		ExperimentalMetadataPrefetchOnMount: config.DefaultExperimentalMetadataPrefetchOnMount,
-	}
-
-	err := validateFlags(flags)
-
-	assert.Equal(t.T(), "client protocol: http4 is not valid", err.Error())
-}
-
 func (t *FlagsTest) TestValidateFlagsForValidSequentialReadSizeAndHTTP2ClientProtocol() {
```
ashmeenkaur, line 337: This test can be removed as it is now same as TestValidateFlagsForValidSequentialReadSizeAndHTTP1ClientProtocol

I think there might be other tests too which can be removed. Can you check?

cmd/legacy_param_mapper_test.go
```diff
@@ -451,3 +451,25 @@ func TestCustomEndpointResolutionFromFlags(t *testing.T) {
...
+}
+
+func TestInvalidClientProtocol(t *testing.T) {
+	flags := &flagStorage{
+		ClientProtocol: mountpkg.ClientProtocol("http3"),
+	}
+
+	_, err := PopulateNewConfigFromLegacyFlagsAndConfig(&mockCLIContext{}, flags, &config.MountConfig{})
+
+	assert.NotNil(t, err)
```
ashmeenkaur, line 474: Can we assert on actual error in this case. I believe it's a custom error in this case.

kislaykishore: Usually, it's not a good idea to test errors for reasons outlined here - https://goto.google.com/go-style/decisions#test-error-semantics. Let me know if you think otherwise.

ashmeenkaur: Makes sense, let's add appropriate error checks for the API (decode hooks later).

cmd/mount.go
```diff
@@ -111,9 +112,8 @@ be interacting with the file system.`)
 		CacheClock:                 timeutil.RealClock(),
 		BucketManager:              bm,
 		BucketName:                 bucketName,
-		LocalFileCache:             false,
```
ashmeenkaur, line 114: is this intended?

kislaykishore: Well, LocalFileCache is always false and setting this to false is a no-op. So, I was trying to remove it but I guess, it could be done as part of a separate PR. So reverted it.

ashmeenkaur (approved):

PR #2097, opened 2024-07-04
Folder metadata in stat cache

ankitaluthra1:
### Description
Adds insertFolder, AddNegativeEntryForFolder and lookup folder in stat cache for HNS buckets.
This PR only adds new methods with corresponding implementations, currently these are not called from any bucket type.
NA

### Testing details
1. Manual - NA
2. Unit tests - Done
3. Integration tests - NA


internal/cache/metadata/stat_cache.go
```diff
@@ -203,3 +219,46 @@ func (sc *statCacheBucketView) LookUp(
 
 	return
 }
+
+func (sc *statCacheBucketView) LookUpFolder(
+	folderName string,
+	now time.Time) (bool, *controlpb.Folder) {
+	// Look up in the LRU cache.
+	value := sc.sharedCache.LookUp(sc.key(folderName))
```
Tulsishah, line 227: Should we create common lookUp function which will return entry and then on respective functions we will assign folder or minObject?

Same for insert.

internal/cache/metadata/stat_cache.go
```diff
@@ -203,3 +219,46 @@ func (sc *statCacheBucketView) LookUp(
 
 	return
 }
+
```
Tulsishah, line 222: AddNegativeEntry function will require to set f to be nil

Tulsishah, line 222: Size()  function will require to add size of e.f.

Not sure if you want to make changes in next PR for these

ankitaluthra1: negative entry and size are part of next PR, trying to make smallest PR for quicker reviews

Tulsishah: 2 line changes I guess required for rest of the two functions, I guess we can add that in this PR so all the stat cache related changes got merged in single PR.

ankitaluthra1: it wont be just 2 lines, new unit tests are also needed to test negative case

Tulsishah: Oh okay sorry forgot about unit tests.

internal/cache/metadata/stat_cache_test.go
```diff
@@ -408,3 +411,72 @@ func (t *MultiBucketStatCacheTest) Test_ExpiresLeastRecentlyUsed() {
...
+
+	t.statCache.InsertFolder(newEntry, expiration)
+
+	hit, entry := t.statCache.LookUpFolder(name, someTime)
+	assert.True(t.T(), hit)
+	assert.Equal(t.T(), "key1", entry.Name)
+	assert.Equal(t.T(), int64(2), entry.Metageneration)
+}
+
+func (t *StatCacheTest) Test_Lookup_Return_False_If_Expiration_Is_Passed() {
```
Tulsishah, line 450: Should we write unit tests for entry equals to nil?

internal/cache/metadata/stat_cache.go
Tulsishah: e.g.
```
lookUp(return entry)
lookUpObject(set object from entry)
lookUpFolder(set folder from entry)
```
If it takes more effort then we can skip may be.

ankitaluthra1: Done for lookup, Insert is not having much repetitive code for now and so not refactoring insert for now.

internal/cache/metadata/stat_cache_test.go
ankitaluthra1: added

internal/cache/metadata/stat_cache_test.go
```diff
@@ -111,6 +113,7 @@ type MultiBucketStatCacheTest struct {
 func (t *StatCacheTest) SetupTest() {
 	cache := lru.NewCache(uint64((mount.AverageSizeOfPositiveStatCacheEntry + mount.AverageSizeOfNegativeStatCacheEntry) * capacity))
 	t.cache.wrapped = metadata.NewStatCacheBucketView(cache, "") // this demonstrates
+	t.statCache = metadata.NewStatCacheBucketView(cache, "")     // this demonstrates
```
Tulsishah, line 116: can't we use `t.cache` only? why did we need this separate variable?

Tulsishah (commented):
LGTM

internal/cache/metadata/stat_cache_test.go
ankitaluthra1: t.cache is the old wrapper class in tests that calls metadata.StatCache. This approach tests wrapper methods instead of actual functionality, compromising the safety net. If the helper class changes internally, tests won't fail. I've refactored this in my PR to test metadata.StatCache directly, removing unnecessary wrappers for accurate unit testing. Changing every test will increase the scope and actual hns work will be affected, so taking cautious call to just add new test in refactored way in first go, we can update rest of tests slowly later.

Tulsishah: Okay just write comment above it.

Tulsishah (approved):

internal/cache/metadata/stat_cache_test.go
```diff
@@ -408,3 +411,128 @@ func (t *MultiBucketStatCacheTest) Test_ExpiresLeastRecentlyUsed() {
...
+
+	assert.True(t.T(), hit1)
+	assert.Equal(t.T(), "1", entry1.Name)
+	assert.True(t.T(), hit2)
+	assert.Equal(t.T(), "2", entry2.Name)
+
+	t.statCache.InsertFolder(folderEntry, expiration) //adds size of 220 and exceeds capacity
+
+	hit1, entry1 = t.statCache.LookUp("1", someTime)
+	hit3, entry3 := t.statCache.LookUpFolder("3", someTime)
```
Tulsishah, line 531: If I am correct It will cache now objectEntry1  and folderEntry and evict objectEntry2, so size function should return `1428 + 220 = 1648` if so Let's make assert for this.  

Also should we make block of setup, then calls size function(block2) and the assert of size(block3)?

ankitaluthra1: > If I am correct It will cache now objectEntry1 and folderEntry and evict objectEntry2, so size function should return 1428 + 220 = 1648 if so Let's make assert for this.
assertions for entry 1 and 2 are already in test, entry 1 is evicted, assertions are added for the same and not 2, least recently used is entry 1

ankitaluthra1: I have added blocks based on sequence of testing, since this test is testing two scenarios in one test, AAA for first scenario - where two object entries are added and then AAA for second scenario - in which folder entry is added

Tulsishah: Got it thanks

Tulsishah (dismissed):

Tulsishah (dismissed):

internal/cache/metadata/stat_cache.go
```diff
@@ -176,30 +201,86 @@ func (sc *statCacheBucketView) AddNegativeEntry(objectName string, expiration ti
...
+	// Look up in the LRU cache.
+	entry := sc.lruCacheLookup(folderName, now)
+	if entry == nil {
+		return false, nil
+	}
+
+	return true, entry.f
+}
+
+func (sc *statCacheBucketView) lruCacheLookup(key string, now time.Time) *entry {
```
vadlakondaswetha, line 248: can you rename this to cachelookup or sharedcachelookup.
too many names will be confusing - method is called lrucache , inside we are using sharedcache.

vadlakondaswetha, line 248: how about returning a bool here which says entry is found or not.

internal/cache/metadata/stat_cache_test.go
```diff
@@ -408,3 +418,128 @@ func (t *MultiBucketStatCacheTest) Test_ExpiresLeastRecentlyUsed() {
 	assert.Equal(t.T(), cardamom, spices.LookUpOrNil("cardamom", someTime))
 	assert.Equal(t.T(), saffron, spices.LookUpOrNil("saffron", someTime))
 }
+
+func (t *StatCacheTest) Test_Create_Entry_When_No_Entry_Is_Present() {
```
vadlakondaswetha, line 422: naming of test methods

Test_methodNameUnderTest_conditionAndExpectedResult or
methodNameUnderTest_condition_expectedResult

See what is followed in gcsfuse and name appropriately.

internal/cache/metadata/stat_cache_test.go
```diff
@@ -408,3 +418,128 @@ func (t *MultiBucketStatCacheTest) Test_ExpiresLeastRecentlyUsed() {
...
+	assert.Equal(t.T(), int64(1), entry.Metageneration)
+}
+
+func (t *StatCacheTest) Test_Override_Entry_Old_Entry_Is_Already_Present() {
+	const name = "key1"
+	existingEntry := &controlpb.Folder{
+		Name:           name,
+		Metageneration: 1,
+	}
+	t.statCache.InsertFolder(existingEntry, expiration)
```
vadlakondaswetha, line 443: check the metageneration of existing entry.

internal/cache/metadata/stat_cache_test.go
```diff
@@ -408,3 +418,128 @@ func (t *MultiBucketStatCacheTest) Test_ExpiresLeastRecentlyUsed() {
...
+
+	assert.True(t.T(), hit1)
+	assert.Equal(t.T(), "1", entry1.Name)
+	assert.True(t.T(), hit2)
+	assert.Equal(t.T(), "2", entry2.Name)
+
+	t.statCache.InsertFolder(folderEntry, expiration) //adds size of 220 and exceeds capacity
+
+	hit1, entry1 = t.statCache.LookUp("1", someTime)
+	hit3, entry3 := t.statCache.LookUpFolder("3", someTime)
```
vadlakondaswetha, line 538: can we also check if entry2 is still present.

internal/cache/metadata/stat_cache.go
```diff
@@ -92,14 +113,18 @@ type entry struct {
...
+	// First, calculate size on heap (including folder size also in case of hns buckets, in case of non-hns buckets 0 will be added as e.f will be Nil ).
 	// Additional 2*util.UnsafeSizeOf(&e.key) is to account for the copies of string
 	// struct stored in the cache map and in the cache linked-list.
 	size = uint64(util.UnsafeSizeOf(&e) + len(e.key) + 2*util.UnsafeSizeOf(&e.key) + util.NestedSizeOfGcsMinObject(e.m))
 	if e.m != nil {
 		size += 515
 	}
 
+	if e.f != nil {
+		size = size + uint64(util.UnsafeSizeOf(&e.f))
```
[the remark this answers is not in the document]
ankitaluthra1: done

internal/cache/metadata/stat_cache.go
ankitaluthra1: I also had same dilemma to return additional bool or not then I thought its redundant information based entry returned caller can find out if cache is a hit or miss.

internal/cache/metadata/stat_cache_test.go
ankitaluthra1: added Test_methodNameUnderTest_ as per convention

internal/cache/metadata/stat_cache_test.go
ankitaluthra1: done

internal/cache/metadata/stat_cache_test.go
ankitaluthra1: We could do that but it was making test less readable with multipl AAA blocks, addin that assert would be testing that first insert is working as expected which is already tested in another test where no entry is present, so did not add it in this test

internal/cache/metadata/stat_cache.go
ankitaluthra1: renamed to sharedcachelookup.

vadlakondaswetha: I would prefer the other way. It would also make the caller job simpler.

func (sc *statCacheBucketView) LookUp(
	objectName string,
	now time.Time) (hit bool, m *gcs.MinObject) {
hit, entry := sc.sc.lruCacheLookup(folderName, now)
if hit {
m = entry.m
}
return
}

ankitaluthra1: done

internal/cache/metadata/stat_cache.go
```diff
@@ -176,30 +201,86 @@ func (sc *statCacheBucketView) AddNegativeEntry(objectName string, expiration ti
...
-	now time.Time) (hit bool, m *gcs.MinObject) {
+	now time.Time) (bool, *gcs.MinObject) {
 	// Look up in the LRU cache.
-	value := sc.sharedCache.LookUp(sc.key(objectName))
+	hit, entry := sc.sharedcachelookup(objectName, now)
+	if hit {
+		return hit, entry.m
+	}
+
+	return false, nil
```
vadlakondaswetha, line 233: Isn't the hit false here. It will be easy when you use named parameters for return types.
func (sc *statCacheBucketView) LookUp(
	objectName string,
	now time.Time) (hit bool, m *gcs.MinObject) {
hit, entry := sc.sharedCacheLookup(objectName, now)
if hit {
   m = entry.m
}

return

internal/cache/metadata/stat_cache.go
```diff
@@ -176,30 +201,86 @@ func (sc *statCacheBucketView) AddNegativeEntry(objectName string, expiration ti
...
+	// Look up in the LRU cache.
+	hit, entry := sc.sharedcachelookup(folderName, now)
+	if hit {
+		return hit, entry.f
+	}
+
+	return false, nil
+}
+
+func (sc *statCacheBucketView) sharedcachelookup(key string, now time.Time) (bool, *entry) {
```
vadlakondaswetha, line 248: camelCase for method names => sharedCacheLookup

internal/cache/metadata/stat_cache.go
ankitaluthra1: line 230 hit is true and hit (true) is returned and line no 233 it should be false, seems working as expected, could please point which line no is returning incorrect value. Also should we avoid naked returns in all methods to be consistent, please see [ref](https://go.dev/tour/basics/7#:~:text=They%20can%20harm%20readability%20in%20longer%20functions.)

vadlakondaswetha (approved):

PR #2100, opened 2024-07-04
Fix DeadLock in Kernel List Cache

raj-prince:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -774,6 +774,7 @@ func (fs *fileSystem) mintInode(ic inode.Core) (in inode.Inode) {
 // LOCK_FUNCTION(in)
 func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Inode) {
 
+	logger.Info("Entering lookUpOrCreateInodeIfNotStale: ", ic.FullName)
 	// Sanity check.
 	if err := ic.SanityCheck(); err != nil {
 		panic(err.Error())
@@ -836,6 +837,8 @@ func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Ino
 		return
 	}
 
+	logger.Info("Calling lookUpOrCreateInodeIfNotStale: Midway")
+
 	oGen := inode.Generation{
 		Object:   ic.MinObject.Generation,
 		Metadata: ic.MinObject.MetaGeneration,
@@ -852,6 +855,7 @@ func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Ino
 			in = fs.mintInode(ic)
 			fs.generationBackedInodes[in.Name()] = in.(inode.GenerationBackedInode)
 
+			logger.Info("Minting")
 			in.Lock()
 			return
 		}
@@ -861,6 +865,8 @@ func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Ino
 		// acquiring the file system lock, so drop it while acquiring the inode's
 		// lock, then reacquire.
 		fs.mu.Unlock()
+		logger.Info("Sleeping for 5 sec")
+		time.Sleep(time.Second * 5)
 		existingInode.Lock()
 		fs.mu.Lock()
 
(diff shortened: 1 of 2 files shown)


internal/fs/fs.go
```diff
@@ -2154,10 +2154,16 @@ func (fs *fileSystem) OpenDir(
 
 	// Enables kernel list-cache in case of non-zero kernelListCacheTTL.
 	if fs.kernelListCacheTTL > 0 {
+		// Following lock ordering rules: first taking dirInode.RLock() and then fs.lock().
+		fs.mu.Unlock()
 
 		// Taking RLock() since ShouldInvalidateKernelListCache only reads the DirInode
 		// properties, no modification.
 		in.RLock()
```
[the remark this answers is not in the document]
vadlakondaswetha: couple of questions:
1. prevDirListingTimestamp is getting updated in RmDir flow also. Is that correct?
2. Assuming that same directory list call will not be made if the data is already cached, do you need this locking here?
3. What happens if you read inconsistent data. Just trying to avoid the locking.

internal/fs/fs.go
```diff
@@ -2154,10 +2154,16 @@ func (fs *fileSystem) OpenDir(
 
 	// Enables kernel list-cache in case of non-zero kernelListCacheTTL.
 	if fs.kernelListCacheTTL > 0 {
+		// Following lock ordering rules: first taking dirInode.RLock() and then fs.lock().
+		fs.mu.Unlock()
```
ashmeenkaur, line 2158: we can unlock fs after adding directory handle (line 2152) and remove the defer lock (line 2141)

internal/fs/fs.go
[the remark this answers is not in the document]
raj-prince: 1. Yes.
2. Once data is already in cache, then we only get OpenDir() and ReleaseDirHandle() call. No ReadDir() call. 
Although It makes sense to add the lock in OpenDir() but it doesn't help in reducing the no. of GCS List call. Since, OpenDir() will get completed before other's ReadDir(). So, taking a lock doesn't help here.
3. I don't see any harm in reading the inconsistent data. The same reasoning, OpenDir() gets completed before others ReadDir() call, so we would not be able to save the GCS List call by reading the consistent data. I am align with removing the lock from here. Making the change.

CC: @sethiay please let me know, if you want to discuss.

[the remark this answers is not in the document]
sethiay: One case to check is: What happens to the performance (duration of all lists to complete) if there are parallel lists on same directory with large number of files (say 1M flat directory) with and without this read lock ? IIUC, without this lock listing from GCS will happen twice and increase the duration ?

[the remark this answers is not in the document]
sethiay: Confirmed offline: lock will not save the GCS list call in this case because the readdir calls come for 2K entries at once. So, say there is a dir handle doing 1st readdir call and there is second call waiting on opendir read's lock. As soon as the first dir handle's first readdir's call complete, the opendir call will proceed and see that the prev update time is updated, however the kernel would not have cached the listing yet. After that the dir handle's second readdir call will come until the last last readdir and the opendir call of second reader will proceed to readdir making GCS calls.

internal/fs/fs.go
charith87, line 1: Is there any unit tests along with the code changes of this file particularly.

raj-prince: We write composite test (mixture of e2e test and unit test) for the changes in fs.go. For these changes, written in `kernel_list_cache_test.go`.

charith87: Thanks Prince. Marking it Resolved.

internal/fs/fs.go
```diff
@@ -2138,7 +2138,6 @@ func (fs *fileSystem) OpenDir(
 	ctx context.Context,
 	op *fuseops.OpenDirOp) (err error) {
 	fs.mu.Lock()
-	defer fs.mu.Unlock()
```
vadlakondaswetha, line 2141: what happens if line no: 2146 fails, will the fs.mu gets unlocked?

ashmeenkaur (approved):

internal/fs/fs.go
ashmeenkaur: Based on the documentation: https://github.com/golang/go/blob/87ec2c959c73e62bfae230ef7efca11ec2a90804/src/builtin/builtin.go#L265, the defer won't execute in this case but IIUC I think it should be okay because we are panicking in that case and as soon as the go-routine finishes, GCSFuse will terminate.

raj-prince: +1
Any failure in line no: 2146 leads to panic, which we handle in https://github.com/GoogleCloudPlatform/gcsfuse/blob/cecf4ac1902791d21bafcb87fceae0002db3c57f/internal/fs/wrappers/error_mapping.go#L92 and terminate the gcsfuse process.

vadlakondaswetha (approved):

PR #2112, opened 2024-07-08
Don't change "-h" shorthand to "--h"

kislaykishore:
urfave/cli uses the golang flags library which treats the "-flagName" and "--flagName" as equivalent, however Cobra, which uses the pFlags (POSIX flag standard), treats single-hyphen-prefixed flags as shorthands while doubly-hyphen-prefixed as flags. In order to be backward compatible, we prefix the flags that have been specified with a single hyphen with another hyphen. This however causes problems for certain shorthand flags such as "-v" and "-h" which have special meaning to Cobra. Prefixing them with another hyphen causes errors since to Cobra, Cobra only knows about "v" and "h" shorthands not flags.
As part of this PR we address the following:
* Don't modify "-h" to "--h"
* Convert "--h" and "--v" to "-h" and "-v" respectively.

NA

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -47,11 +47,6 @@ func TestPosixFlagsConversion(t *testing.T) {
 			expected: []string{"abc", "--flag=\"-test\""},
 		},
 		{
-			name:     "Version shorthand stays unchanged.",
-			input:    []string{"abc", "-v"},
-			expected: []string{"abc", "-v"},
-		},
-		{
 			name:     "Help shorthand stays unchanged.",
 			input:    []string{"abc", "-h"},
 			expected: []string{"abc", "-h"},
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -51,6 +51,26 @@ func TestPosixFlagsConversion(t *testing.T) {
 			input:    []string{"abc", "-v"},
 			expected: []string{"abc", "-v"},
 		},
+		{
+			name:     "Help shorthand stays unchanged.",
+			input:    []string{"abc", "-h"},
+			expected: []string{"abc", "-h"},
+		},
+		{
+			name:     "Version shorthand stays unchanged.",
```
ashmeenkaur, line 60: repeated test case (line 50)

ashmeenkaur (approved):

PR #2114, opened 2024-07-08
Move logging to the new config

kislaykishore:
### Description
* Localized the log-severity resolution logic in the `Rationalize` method.
* Move logging to use the new cfg.Config type.

NA

### Testing details
1. Manual - Added unit tests.
2. Unit tests - NA
3. Integration tests - NA


cmd/legacy_param_mapper.go
```diff
@@ -141,6 +141,7 @@ func PopulateNewConfigFromLegacyFlagsAndConfig(c cliContext, flags *flagStorage,
 	overrideWithFlag(c, "anonymous-access", &resolvedConfig.GcsAuth.AnonymousAccess, anonymousAccess)
 	overrideWithFlag(c, "kernel-list-cache-ttl-secs", &resolvedConfig.FileSystem.KernelListCacheTtlSecs, kernelListCacheTTLSecs)
 
+	cfg.OverrideWithLoggingFlags(resolvedConfig, resolvedConfig.Debug.Fuse, resolvedConfig.Debug.Gcs, resolvedConfig.Debug.LogMutex)
```
ashmeenkaur, line 144: IIUC, this should be in a separate method as we won't be using PopulateNewConfigFromLegacyFlagsAndConfig function post migration but this is needed on new configs regardless for backward compatibility.

kislaykishore: I've moved it to Rationalize.

internal/logger/logger_test.go
```diff
@@ -273,38 +273,36 @@ func (t *LoggerTest) TestInitLogFile() {
...
+	assert.Equal(t.T(), fileSize, defaultLoggerFactory.logRotate.MaxFileSizeMb)
+	assert.Equal(t.T(), backupFileCount, defaultLoggerFactory.logRotate.BackupFileCount)
+	assert.True(t.T(), defaultLoggerFactory.logRotate.Compress)
 }
 
 func (t *LoggerTest) TestSetLogFormatToText() {
 	defaultLoggerFactory = &loggerFactory{
-		file:            nil,
-		level:           config.INFO, // setting log level to INFO by default
-		logRotateConfig: config.DefaultLogRotateConfig(),
```
ashmeenkaur, line 307: Shall we clean up config.DefaultLogRotateConfig()

cfg/rationalize.go
```diff
@@ -38,5 +38,9 @@ func Rationalize(c *Config) error {
 		return err
 	}
 
+	if c.Debug.Fuse || c.Debug.Gcs || c.Debug.LogMutex {
```
ashmeenkaur, line 41: Same method is also there in config_util.go (OverrideWithLoggingFlags).
Let's refactor/cleanup the code to use the common logic.

cmd/legacy_param_mapper_test.go
```diff
@@ -505,6 +505,80 @@ func TestInvalidClientProtocol(t *testing.T) {
 	assert.NotNil(t, err)
 }
 
+func TestLogSeverityRationalization(t *testing.T) {
```
ashmeenkaur, line 508: Nit: I think this unit test is not required. WDYT?

internal/logger/logger.go
```diff
@@ -103,10 +103,10 @@ func InitLogFile(legacyLogConfig config.LogConfig, newLogConfig cfg.LoggingConfi
 // init initializes the logger factory to use stdout and stderr.
 func init() {
 	defaultLoggerFactory = &loggerFactory{
-		file:            nil,
-		format:          defaultFormat,
-		level:           config.INFO, // setting log level to INFO by default
-		logRotateConfig: config.DefaultLogRotateConfig(),
+		file:      nil,
+		format:    defaultFormat,
+		level:     config.INFO, // setting log level to INFO by default
```
ashmeenkaur, line 108: Shall we move these constants to new config package and use from there?

ashmeenkaur (commented):
Shall we try and remove LogConfig in mount_config.go and use new config there?

internal/logger/logger.go
kislaykishore: Currently, newConfig has dependency on old config. To avoid cyclic dependency, I'll wait for https://github.com/GoogleCloudPlatform/gcsfuse/pull/2258 to be submitted where you remove that dependency.

cmd/legacy_param_mapper_test.go
kislaykishore: It helps validate that Rationalize is being called in the mapper, although it doesn't need to be this exhaustive. In any case, no harm no foul.

cfg/defaults.go
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cfg
+
+// GetDefaultLoggingConfig returns the default configuration that is to be used
+// during the application startup - when the provided configuration hasn't been
+// parsed yet.
+func GetDefaultLoggingConfig() LoggingConfig {
```
ashmeenkaur, line 20: Nit: we are never accessing Severity from this field. Shall we just create a function for DefaultLogRotateConfig.
Also we should remove Get prefix: https://google.github.io/styleguide/go/best-practices.html#naming-conventions

internal/config/mount_config.go
```diff
@@ -174,21 +174,18 @@ type LogRotateConfig struct {
 	Compress        bool `yaml:"compress"`
 }
 
-func DefaultLogRotateConfig() LogRotateConfig {
-	return LogRotateConfig{
-		MaxFileSizeMB:   defaultMaxFileSizeMB,
```
ashmeenkaur, line 179: Nit: shall we revert the changes made to this and yaml_parser file?
If not, we should remove the unused constants like "defaultMaxFileSizeMB"

cfg/defaults.go
kislaykishore: I've changed the code to use severity from the DefaultLogRotateConfig; had missed out earlier to do that.

ashmeenkaur: Shall we remove the Get prefix too?

kislaykishore: Done!

internal/logger/logger.go
```diff
@@ -101,11 +99,12 @@ func InitLogFile(newLogConfig cfg.LoggingConfig) error {
 
 // init initializes the logger factory to use stdout and stderr.
 func init() {
+	logConfig := cfg.DefaultLoggingConfig()
 	defaultLoggerFactory = &loggerFactory{
 		file:      nil,
-		format:    defaultFormat,
-		level:     cfg.INFO, // setting log level to INFO by default
-		logRotate: cfg.GetDefaultLoggingConfig().LogRotate,
+		format:    logConfig.Format,
```
ashmeenkaur, line 105: logConfig.format needs to be set in the Default function

ashmeenkaur (approved):

PR #2115, opened 2024-07-08
Pass the bucket and mountpoint into the mount fn

kislaykishore:
Pass the bucket and mountpoint into the mount fn

### Description
* Refactor the populateArgs function to take a slice of args as input instead of the cliContext.
* Invoke populateArgs in the root command so that the new config gets access to the bucket and mountpoint and can invoke the mount function with the config, bucket and mountpoint.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added the unit tests.
3. Integration tests - NA

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -45,11 +45,11 @@ of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 			if cfgErr != nil {
 				return fmt.Errorf("error while parsing config: %w", cfgErr)
 			}
-			if bucket, mountpoint, err := populateArgs(args[1:]); err != nil {
-				return fmt.Errorf("error occurred while extracting the bucket and mountpoint: %w", err)
-			} else {
-				return mountFn(&configObj, bucket, mountpoint)
+			bucket, mountPoint, err := populateArgs(args[1:])
+			if err != nil {
+				return fmt.Errorf("error occurred while extracting the bucket and mountPoint: %w", err)
 			}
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -45,7 +45,11 @@ of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 			if cfgErr != nil {
 				return fmt.Errorf("error while parsing config: %w", cfgErr)
 			}
-			return mountFn(configObj)
+			if bucket, mountpoint, err := populateArgs(args[1:]); err != nil {
+				return fmt.Errorf("error occurred while extracting the bucket and mountpoint: %w", err)
+			} else {
```
ashmeenkaur, line 50: Nit: remove indented else block (https://google.github.io/styleguide/go/decisions.html#indent-error-flow) and rename mountpoint -> mountPoint

cmd/root_test.go
```diff
@@ -106,3 +108,63 @@ func TestCobraArgsNumInRange(t *testing.T) {
...
+				return nil
+			})
+			require.Nil(t, err)
+			cmd.SetArgs(tc.args)
+
+			err = cmd.Execute()
+
+			if assert.Nil(t, err) {
+				assert.Equal(t, tc.expectedBucket, bucketName)
+				if assert.Nil(t, err) {
```
ashmeenkaur, line 164: Is this if block required considering we are checking error on line 162?

ashmeenkaur (approved):

PR #2117, opened 2024-07-08
Decode empty string as nil for URL

kislaykishore:
Instead of unmarshalling/decoding empty string as a pointer pointing to an empty url.URL{} object, decode it as nil.

### Description
Moving stringToURLFunc() to the last in the list of decode hooks since the other decoders like `StringToTimeDurationHookFunc()` and `mapstructure.StringToSliceHookFunc`  are unable to handle nils.

NA

### Testing details
1. Manual - NA
3. Integration tests - NA

### DIFF
--- a/cfg/decode_hook.go
+++ b/cfg/decode_hook.go
@@ -34,6 +34,9 @@ func stringToURLHookFunc() mapstructure.DecodeHookFuncType {
 			return data, nil
 		}
 		s := data.(string)
+		if s == "" {
+			return nil, nil
+		}
 		u, err := url.Parse(s)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


cfg/decode_hook.go
```diff
@@ -46,8 +49,8 @@ func stringToURLHookFunc() mapstructure.DecodeHookFuncType {
 func DecodeHook() mapstructure.DecodeHookFunc {
 	return mapstructure.ComposeDecodeHookFunc(
 		mapstructure.TextUnmarshallerHookFunc(),
-		stringToURLHookFunc(),
 		mapstructure.StringToTimeDurationHookFunc(), // default hook
 		mapstructure.StringToSliceHookFunc(","),     // default hook
+		stringToURLHookFunc(),
```
charith87, line 54: Seems like a hack tbh. If this always needs to be in the end how about adding a comment that to add a b/XXXXX and short desc that it should be in the end. So nobody else changes it in the future.

kislaykishore:
We decided to change the URL params to string flags instead.

PR #2118, opened 2024-07-08
 Add parallel download configuration to existing read cache e2e tests

ashmeenkaur:
### Description

This PR adds parallel download configuration to existing read cache e2e tests.
Changes involved:
- Parallel download config added to the existing tests.
- It is possible that cache file is not fully downloaded after the read is served, so retry for 5 seconds until the file is cached before matching CRC with GCS.
- Some miscellaneous changes to return error / use stretchr testify to make the test more debuggable.
- Reduce file size for tools/integration_tests/read_cache/range_read_test.go file so that the test executes faster. Parallel download tests are not added for this file as it is not predictable that the file will be cached for a chunk in this case. 

NA

### Testing details
1. Manual - Ran multiple times manually to confirm that the test is not flaky.
2. Unit tests - NA
3. Integration tests - Ran via KOKORO.

### DIFF
--- a/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
+++ b/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
@@ -144,12 +144,9 @@ func TestCacheFileForRangeReadFalseTest(t *testing.T) {
 
 	// Define flag set to run the tests.
 	flagsSet := [][]string{
-		{"--implicit-dirs=true"},
+		{"--implicit-dirs", "--config-file=" + createConfigFile(cacheCapacityForRangeReadTestInMiB, false, configFileName, false)},
+		{"--config-file=" + createConfigFile(cacheCapacityForRangeReadTestInMiB, false, configFileNameForParallelDownloadTests, true)},
 	}
-	setup.AppendFlagsToAllFlagsInTheFlagsSet(&flagsSet,
-		"--config-file="+createConfigFile(cacheCapacityForRangeReadTestInMiB, false, configFileName, false),
-		"--config-file="+createConfigFile(cacheCapacityForRangeReadTestInMiB, false, configFileNameForParallelDownloadTests, true),
-	)
 
 	// Run tests.
 	for _, flags := range flagsSet {
--- a/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
+++ b/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
@@ -147,9 +147,9 @@ func TestCacheFileForRangeReadFalseTest(t *testing.T) {
 		{"--implicit-dirs=true"},
 	}
 	setup.AppendFlagsToAllFlagsInTheFlagsSet(&flagsSet,
-		"--config-file="+createConfigFile(cacheCapacityForRangeReadTestInMiB, false, configFileName+"ForReadCache", false))
-	flagsSet = append(flagsSet, []string{"--implicit-dirs", "--config-file=" + createConfigFile(cacheCapacityForRangeReadTestInMiB,
-		false, configFileNameForParallelDownloadTests, true)})
+		"--config-file="+createConfigFile(cacheCapacityForRangeReadTestInMiB, false, configFileName, false),
+		"--config-file="+createConfigFile(cacheCapacityForRangeReadTestInMiB, false, configFileNameForParallelDownloadTests, true),
+	)
 
 	// Run tests.
 	for _, flags := range flagsSet {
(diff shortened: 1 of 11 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -107,27 +110,40 @@ func getCachedFilePath(fileName string) string {
...
-	if (*fileInfo).Size() != filesize {
-		t.Errorf("Incorrect cached file size. Expected %d, Got: %d", filesize, (*fileInfo).Size())
+	var err error
+
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		// Validate that the file is present in cache location.
+		var fileInfo *fs.FileInfo
+		fileInfo, err = operations.StatFile(expectedPathOfCachedFile)
+		if fileInfo == nil {
+			err = fmt.Errorf("received nil FileInfo %s", expectedPathOfCachedFile)
```
Tulsishah, line 123: It should be t.Fatalf. Otherwise it will panic in below statement while accessing `(*fileInfo).Size()`.

tools/integration_tests/util/client/gcs_helper.go
```diff
@@ -155,12 +156,13 @@ func CreateNFilesInDir(ctx context.Context, storageClient *storage.Client, numFi
...
 
-func ValidateCRCWithGCS(gotCRC32Value uint32, objectPath string, ctx context.Context, storageClient *storage.Client, t *testing.T) {
+func ValidateCRCWithGCS(gotCRC32Value uint32, objectPath string, ctx context.Context, storageClient *storage.Client) error {
 	attr, err := StatObject(ctx, storageClient, objectPath)
 	if err != nil {
-		t.Errorf("Failed to fetch object attributes: %v", err)
+		return fmt.Errorf("Failed to fetch object attributes: %v", err)
 	}
 	if attr.CRC32C != gotCRC32Value {
-		t.Errorf("CRC32 mismatch. Expected %d, Got %d", attr.CRC32C, gotCRC32Value)
```
Tulsishah, line 164: Should we put first check if `attrs!= nil`. Sometime it's crash due to that.

tools/integration_tests/read_cache/helpers_test.go
ashmeenkaur: err == nil before checking `(*fileInfo).Size()` will take care of this right? I don't want the test execution to stop until retries are done so didn't use t.Fatalf .

Tulsishah: Okay got it. Thanks

Tulsishah (approved):

Tulsishah (approved):

sethiay (commented):
> Increased timeout as the tests are taking more time to execute.
Do we know why ? Also, any numbers to compare from your runs ?

> Reduce file size for tools/integration_tests/read_cache/range_read_test.go file so that the test executes faster. Parallel download tests are not added for this file as it is not predictable that the file will be cached for a chunk in this case.
Is it possible to add the wait logic in this case also ?

tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -107,27 +110,40 @@ func getCachedFilePath(fileName string) string {
...
+		}
+		// Validate file size in cache directory matches actual file size.
+		if err == nil && filesize != (*fileInfo).Size() {
```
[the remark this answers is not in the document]
ashmeenkaur: Really sorry for this, thanks for catching!

ashmeenkaur:
> Do we know why ? Also, any numbers to compare from your runs ?

Numbers were comparable with and without parallel downloads. I think this was happening because of the retry loop not breaking. I've reverted the change.

tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
```diff
@@ -82,6 +82,10 @@ func (s *cacheFileForRangeReadFalseTest) TestRangeReadsWithCacheMiss(t *testing.
 }
 
 func (s *cacheFileForRangeReadFalseTest) TestConcurrentReads_ReadIsTreatedNonSequentialAfterFileIsRemovedFromCache(t *testing.T) {
+	if isParallelDownloadsEnabled(s.flags) {
+		// This test is not valid when Parallel Downloads are enabled.
```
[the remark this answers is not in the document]
ashmeenkaur: It was because we couldn't concretely say whether a read would be cache hit. I have instead enabled this test now and added the condition to cache Hit check

sethiay (approved):

PR #2121, opened 2024-07-08
ListFolders implementation - 1

Tulsishah:
### Description
This PR introduces the List folder API for the GCS bucket interface, along with implementations in relevant structures.
implementations will be addressed in a subsequent PR.

1. Bucket handle
2. Fast stat bucket
3. Fake bucket

Method is not getting called from anywhere in the code.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/gcs/bucket.go
+++ b/internal/storage/gcs/bucket.go
@@ -148,4 +148,10 @@ type Bucket interface {
 
 	// Atomically rename folder for Hierarchical bucket.
 	RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (*control.RenameFolderOperation, error)
+
+	// List the folders in the Hierarchical bucket that meet the criteria defined by the
+	// request, returning a result folders that contains the results and
+	// potentially a cursor for retrieving the next portion of the larger set of
+	// results.
+	ListFolders(ctx context.Context, req *controlpb.ListFoldersRequest) (*ListingFolders, error)
 }
--- a/internal/storage/gcs/bucket.go
+++ b/internal/storage/gcs/bucket.go
@@ -153,5 +153,5 @@ type Bucket interface {
 	// request, returning a result folders that contains the results and
 	// potentially a cursor for retrieving the next portion of the larger set of
 	// results.
-	ListFolders(ctx context.Context, req *controlpb.ListFoldersRequest) (*ListingFolders, error)
+	ListFolders(ctx context.Context, req *ListFolderRequest) (*ListingFolders, error)
 }
--- a/internal/storage/gcs/bucket.go
+++ b/internal/storage/gcs/bucket.go
@@ -153,5 +153,5 @@ type Bucket interface {
 	// request, returning a result folders that contains the results and
 	// potentially a cursor for retrieving the next portion of the larger set of
 	// results.
-	ListFolders(ctx context.Context, req *ListFolderRequest) (*ListingFolders, error)
+	ListFolders(ctx context.Context, req *ListFoldersRequest) (*ListingFolders, error)
 }
--- a/internal/storage/gcs/bucket.go
+++ b/internal/storage/gcs/bucket.go
@@ -153,5 +153,5 @@ type Bucket interface {
 	// request, returning a result folders that contains the results and
 	// potentially a cursor for retrieving the next portion of the larger set of
 	// results.
-	ListFolders(ctx context.Context, req *ListFoldersRequest) (*ListingFolders, error)
+	ListFolders(ctx context.Context, req *ListFoldersRequest) (*ListFoldersResponse, error)
 }
--- a/internal/storage/gcs/request.go
+++ b/internal/storage/gcs/request.go
@@ -325,7 +325,7 @@ type Listing struct {
 // ListingFolders contains a set of folders by a call to ListFolders.
 type ListingFolders struct {
 	// Records for folders matching the listing criteria.
-	folders []*controlpb.Folder
+	Folders []*controlpb.Folder
 
 	// A continuation token, for fetching more results.
 	//
--- a/internal/storage/gcs/request.go
+++ b/internal/storage/gcs/request.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 
+	"cloud.google.com/go/storage/control/apiv2/controlpb"
 	storagev1 "google.golang.org/api/storage/v1"
 )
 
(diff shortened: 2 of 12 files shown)


internal/storage/gcs/request.go
```diff
@@ -321,6 +322,23 @@ type Listing struct {
 	ContinuationToken string
 }
 
+// ListingFolders contains a set of folders by a call to ListFolders.
+type ListingFolders struct {
+	// Records for folders matching the listing criteria.
+	Folders []*controlpb.Folder
```
ankitaluthra1, line 328: should we use our own folder struct here ?

internal/storage/gcs/bucket.go
```diff
@@ -149,4 +149,10 @@ type Bucket interface {
 
 	// Atomically rename folder for Hierarchical bucket.
 	RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (*control.RenameFolderOperation, error)
+
+	// List the folders in the Hierarchical bucket that meet the criteria defined by the
+	// request, returning a result folders that contains the results and
+	// potentially a cursor for retrieving the next portion of the larger set of
+	// results.
+	ListFolders(ctx context.Context, req *controlpb.ListFoldersRequest) (*ListingFolders, error)
```
ankitaluthra1, line 157: Can we create *controlpb.ListFoldersRequest inside bucket handle and pass only requiredvalues which are needed to make the request, only bucket handle should not know about controlpb package.. that way we could decouple external dependency code from most of our classes. For eg: if in future suppose ListFoldersRequest moves to another package or some other change comes in sdk for this strcut all our classes will have to change, if we limit it to one class and have our won structs be used in method signatures, that way we can decouple the sdk code.
its on similar lines why we have created our own folder struct., adding you @vadlakondaswetha also in case you have other thoughts around this..

ankitaluthra1 (changes requested):

internal/storage/gcs/bucket.go
vadlakondaswetha: Agree with Ankita. We should not add dependencies of SDK anywhere except the bucket_handle.go code.

internal/storage/gcs/request.go
Tulsishah: Done

internal/storage/gcs/bucket.go
Tulsishah: Done

internal/storage/gcs/request.go
```diff
@@ -321,6 +321,23 @@ type Listing struct {
...
+	// A continuation token, for fetching more results.
+	//
+	// If non-empty, this listing does not represent the full set of matching
+	// folders in the bucket. Call ListFolders again with the request's
+	// ContinuationToken field set to this value to continue where you left off.
+	//
+	// Note that there is no guarantee of atomicity of listings. Folders written
+	// and deleted concurrently with a single or multiple listing requests may or
+	// may not be returned.
+	ContinuationToken string
```
ankitaluthra1, line 338: who passes this continuation token ? is it kernel ?

internal/storage/gcs/request.go
```diff
@@ -376,3 +393,22 @@ type DeleteObjectRequest struct {
...
+	// set of results to view.
+	PageToken string
+	// Optional. Filter results to folders whose names begin with this prefix.
+	// If set, the value must either be an empty string or end with a '/'.
+	Prefix string
+	// Optional. If set, returns results in a directory-like mode. The results
+	// will only include folders that either exactly match the above prefix, or
+	// are one level below the prefix. The only supported value is '/'.
+	Delimiter string
+}
```
ankitaluthra1, line 414: we can pass fields directly instead of wrapper and if we have more than 2 fields then an a struct to hold arguments ahsould be created anyways exact same wrapper is not required here imo, Delimiter can be added from inside bucket handle, can we add other fields also from inside bucket handle (parent, page size and page token) ot all of them are passed via kernel only ?

internal/storage/gcs/request.go
Tulsishah: In the API results, we received a token, which we assign from the response (e.g., `folderList.ContinuationToken = folderIterator.PageInfo().Token`). This is the same process as with ListObjects.

internal/storage/gcs/request.go
Tulsishah: I feel we should align more closely with the existing [ListObject struct](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/dir.go#L591) that we're passing from dir.go. We can't add the Token field, and PageSize from bucket_handle as it is defined in dir.go.

Tulsishah: Prefix, PageSize, and PageToken are required. I kept Delimiter to align with the ListObject struct.

Tulsishah: As discussed offline. We are keeping this format only.

internal/storage/gcs/request.go
```diff
@@ -376,3 +393,19 @@ type DeleteObjectRequest struct {
...
+	// Optional. A previously-returned page token representing part of the larger
+	// set of results to view.
+	PageToken string
+	// Optional. Filter results to folders whose names begin with this prefix.
+	// If set, the value must either be an empty string or end with a '/'.
+	Prefix string
+	// Optional. If set, returns results in a directory-like mode. The results
+	// will only include folders that either exactly match the above prefix, or
+	// are one level below the prefix. The only supported value is '/'.
+	Delimiter string
```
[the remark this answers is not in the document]
Tulsishah: Added

internal/storage/gcs/request.go
```diff
@@ -321,6 +321,23 @@ type Listing struct {
 	ContinuationToken string
 }
 
+// ListingFolders contains a set of folders by a call to ListFolders.
+type ListingFolders struct {
+	// Records for folders matching the listing criteria.
```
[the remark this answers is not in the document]
Tulsishah: Done

internal/storage/gcs/request.go
```diff
@@ -321,6 +321,23 @@ type Listing struct {
 	ContinuationToken string
 }
 
+// ListingFolders contains a set of folders by a call to ListFolders.
+type ListingFolders struct {
```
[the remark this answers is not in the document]
Tulsishah: Done

PR #2122, opened 2024-07-08
Remove stringify tests

kislaykishore:
### Description
These tests attempt to test the behavior of stringify method but in reality don't serve much purpose for the following reasons:
* This test doesn't really test whether the util.Stringify was called.
* They test the behavior of util.Stringify but there are tests in util package that do that [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/2fb8bba983cd81270357bfc590ccb0d5b4e6fb76/internal/util/util_test.go#L151).
* They are usually a nuisance while adding flags as the tests break on flag addition.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


ankitaluthra1 (dismissed):
"stringify" should not test actual flag being passed, that not part of stringify behaviour. that should be checked in integration or composite tests @kislaykishore please also add reference to any of the  integration or composite test testing that behaviour in PR description for reference.

ashmeenkaur (dismissed):

kislaykishore:
> "stringify" should not test actual flag being passed, that not part of stringify behaviour. that should be checked in integration or composite tests @kislaykishore please also add reference to any of the integration or composite test testing that behaviour in PR description for reference.

Created https://github.com/GoogleCloudPlatform/gcsfuse/issues/2134 to create such an integration test.

ashmeenkaur (approved):

PR #2128, opened 2024-07-08
Fast stat bucket changes

ankitaluthra1:
### Description

- Adds new struct folder
- Updates stat cache to use new struct folder
- Updates Get folder to use new struct folder

currently these methods are not being used in any flow

NA

### Testing details
1. Manual - NA
2. Unit tests - Done
3. Integration tests - Done

### DIFF
--- a/internal/cache/metadata/stat_cache.go
+++ b/internal/cache/metadata/stat_cache.go
@@ -18,7 +18,6 @@ import (
 	"math"
 	"time"
 
-	"cloud.google.com/go/storage/control/apiv2/controlpb"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/cache/lru"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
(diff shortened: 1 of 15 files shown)


Tulsishah (approved):

internal/cache/metadata/stat_cache.go
```diff
@@ -18,7 +18,6 @@ import (
 	"math"
 	"time"
 
-	"cloud.google.com/go/storage/control/apiv2/controlpb"
```
charith87, line 21: Is the old one not used anymore?

charith87: What about non HNS buckets? Would this not affect there.

Tulsishah: This is for folders only. We have created separate folder structure [here](https://github.com/GoogleCloudPlatform/gcsfuse/tree/master/internal/storage/gcs) for controlpb.folder to enable the addition of custom data in the future and to decouple folder struct in gcsfuse code from control package struct.

ankitaluthra1: yes this is just a small refactoring to have a our own struct for decoupling our methods from sdk folder

Issue #1476, opened 2023-10-27
MaxRetryDuration flag is not implemented


Tulsishah:
[max-retry-duration](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/flags.go#L225) flag is  not implemented. 

The flag is for the all GCS calls get retried upto that max time.

PR #2131, opened 2024-07-08
Integrating max retry attempts with client and changing the default retry attempts to 0

Tulsishah:
### Description
- Integrating max-retry-attempts in storage client to stop infinite retries on retry able errors.
- Changing default value to 0 to be backward compatible. setting max retry attempts to 0 implies infinite retries.

https://github.com/GoogleCloudPlatform/gcsfuse/issues/1476

### Testing details
1. Manual - 
    Tested flag with storage test bench server. 
      1. Cmd line (max-retry-attempts = 2) and config file flag not set.
          two retries occurred.
      3. Cmd line (max-retry-attempts = 2 )and config file (max-retry-attempts = 5)
          two retries occurred.
      5. Cmd line (max-retry-attempts  not set) and config file (max-retry-attempts = 5)
          five retries occurred.
      4. Cmd line (max-retry-attempts  not set) and config file (max-retry-attempts not set) = default 0 which means infinite
          Infinite retries occurred.
7. Unit tests - NA
8. Integration tests - NA


internal/storage/storage_handle_test.go
```diff
@@ -298,3 +298,13 @@ func (testSuite *StorageHandleTest) TestCreateClientOptionForGRPCClient() {
...
 }
+
+func (testSuite *StorageHandleTest) TestNewStorageHandleWithMaxRetryAttemptsNotZero() {
+	sc := storageutil.GetDefaultStorageClientConfig()
+	sc.MaxRetryAttempts = 100
+
+	handleCreated, err := NewStorageHandle(context.Background(), sc)
+
+	assert.Nil(testSuite.T(), err)
+	assert.NotNil(testSuite.T(), handleCreated)
```
kislaykishore, line 309: You should only check whether handleCreated is not nil if there is no error.
```suggestion
	if assert.Nil(testSuite.T(), err) {
	    assert.NotNil(testSuite.T(), handleCreated)
	}
```

kislaykishore (changes requested):
The title appears misleading; should it be something like: "Changing the default retry attempts to 0". Also, in description we can add that "setting max retry attempts to 0 implies infinite retries".

internal/storage/storage_handle_test.go
Tulsishah: Done

Tulsishah:
> The title appears misleading; should it be something like: "Changing the default retry attempts to 0". Also, in description we can add that "setting max retry attempts to 0 implies infinite retries".

Done

kislaykishore (approved):

kislaykishore (approved):

raj-prince (approved):

PR #2139, opened 2024-07-10
Function to convert controlpb.folder attributes to gcs.Folder

Tulsishah:
### Description
- Function to convert controlpb.folder attribute to gcs.Folder as it will required in many places.
- In HNS, folder paths returned by control client APIs are in the form of: `projects/_/buckets/{bucket}/folders/{folder}`
Introduced method extracts the folder name from such a string.
  e.g.
  control client GetFolder(`folder1/folder2`) returns folder Name as `projects/_/buckets/{bucket}/folders/folder1/folder2`
  We will fetch `folder1/folder2` from response and set it in folder attribute.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added


internal/storage/storageutil/folder_attrs.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2024 Google Inc. All Rights Reserved.
```
ankitaluthra1, line 1: I think these are part of folder domain only, we can keep single file for folder domain, can we move these methods to folder.go file itself ? any concerns with that ?

Tulsishah: Isn't it more of helper functions only which should not be part of actual folder file? Also with that we will have to add controlpb dependency over there. Which we did not want to do I guess.

ankitaluthra1: As per Domain Driven Design , If helper functions change domain state, it should be part of domain [ref](https://medium.com/@chikim79/enriching-entity-classes-with-business-logic-with-domain-driven-design-principles-0a937983f529#:~:text=Benefits%20of%20Implementing%20Business%20Logic%20within%20the%20Entities)

ankitaluthra1: the purpose of restricting controlpb is not leak into aother business logics like other bucket wrappers and dir.go, folder domain is like an ACL to achieve that and is tighly coupled with control pb folder hence adding that dependency should not be an issue.

Tulsishah: got it done.

internal/storage/bucket_handle.go
```diff
@@ -512,13 +512,8 @@ func (b *bucketHandle) GetFolder(ctx context.Context, folderName string) (*gcs.F
...
 	}
 
-	folder := gcs.Folder{
-		Name:           clientFolder.Name,
-		Metageneration: clientFolder.Metageneration,
-		UpdateTime:     clientFolder.GetUpdateTime().AsTime(),
-	}
-
-	return &folder, err
+	folder := gcs.ControlFolderAttrsToGCSFolder(b.bucketName, clientFolder)
```
ankitaluthra1, line 515: ControlFolderAttrsToGCSFolder can be renamed to ConvertToGCSFolderFromGetFolderResponse, method name makes more sense as this method main purpose is conversion and also should avoid short forms and acronyms in method names like Attrs.

Tulsishah: Done.

ankitaluthra1 (approved):

internal/storage/gcs/folder.go
```diff
@@ -15,11 +15,34 @@
...
+	"cloud.google.com/go/storage/control/apiv2/controlpb"
 )
 
 type Folder struct {
 	Name           string
 	Metageneration int64
 	UpdateTime     time.Time
 }
+
+func ConvertToGCSFolderFromGetFolderResponse(bucketName string, attrs *controlpb.Folder) *Folder {
```
charith87, line 30: nit Could we call this getGCSFolder() instead of a elaborateName.

internal/storage/gcs/folder.go
```diff
@@ -15,11 +15,34 @@
...
+		Metageneration: attrs.Metageneration,
+		UpdateTime:     attrs.GetUpdateTime().AsTime(),
+	}
+}
+
+// In HNS, folder paths returned by control client APIs are in the form of:
+//
+// projects/_/buckets/{bucket}/folders/{folder}
+// This method extracts the folder name from such a string.
+func findFolderName(bucketName, object string) string {
```
charith87, line 43: nit getFolderName for consistency. find is a difficult word to search.

charith87, line 43: also "object string" could be "object fullPath" or something signifying what actually needs to be passed.

internal/storage/bucket_handle.go
```diff
@@ -512,13 +512,8 @@ func (b *bucketHandle) GetFolder(ctx context.Context, folderName string) (*gcs.F
...
 	}
 
-	folder := gcs.Folder{
-		Name:           clientFolder.Name,
-		Metageneration: clientFolder.Metageneration,
-		UpdateTime:     clientFolder.GetUpdateTime().AsTime(),
-	}
-
-	return &folder, err
+	folder := gcs.ConvertToGCSFolderFromGetFolderResponse(b.bucketName, clientFolder)
```
charith87, line 515: nit = Can clientFolder be renamed to folderResponse or something to signify its a network response. There are too many similar names which hampers readability.

internal/storage/gcs/folder.go
Tulsishah: It is converting the type, that's why I feel convert make sense here.

charith87 (approved):
I've only added nit comments. So approving the review.

internal/storage/gcs/folder.go
Tulsishah: It will unnecessary increase the type conversion. I can change object with fullPath name.

internal/storage/gcs/folder.go
charith87: as per go/go-style/best-practices#function-name-conventions it says even noun names should be good. i.e. gcsFolder(bucketName, attrs *controlpb.Folder) should be fine.
But if this is how the naming code is already present in other files in our codebase, its fine to do it later. Will let you take the call for it.

Tulsishah: Ohh okay.. then let's do it.

Tulsishah: Done

charith87 (dismissed):

charith87 (approved):

PR #2154, opened 2024-07-11
Functional test infinite  kernel cache ttl - 2

Tulsishah:
### Description
1. addition of new file.
  (a) First ReadDir() will be served from GCSFuse filesystem.
  (b) Second ReadDir() will also be served from GCSFuse filesystem
2. deletion of new file.
   (a) First ReadDir() will be served from GCSFuse filesystem.
   (b) Second ReadDir() will also be served from GCSFuse filesystem
3. file rename.
   (a) First ReadDir() will be served from GCSFuse filesystem.
   (b) Second ReadDir() will also be served from GCSFuse filesystem
NA

### Testing details
1. Manual - NA
5. Unit tests - NA
6. Integration tests - Automated

### DIFF
--- a/tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
+++ b/tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
@@ -172,7 +172,7 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_CacheMissOnDeletionOfF
 	// Waiting for 5 seconds to see if the kernel cache expires.
 	time.Sleep(5 * time.Second)
 
-	// Ideally no invalidation since infinite ttl, but creation of a new file inside
+	// Ideally no invalidation since infinite ttl, but deletion of file inside
 	// directory evicts the list cache for that directory.
 	err = os.Remove(path.Join(targetDir, "file2.txt"))
 	require.NoError(t, err)
--- a/tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
+++ b/tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
@@ -218,7 +218,7 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_CacheMissOnFileRename(
 	// Waiting for 5 seconds to see if the kernel cache expires.
 	time.Sleep(5 * time.Second)
 
-	// Ideally no invalidation since infinite ttl, but creation of a new file inside
+	// Ideally no invalidation since infinite ttl, but rename of a file inside
 	// directory evicts the list cache for that directory.
 	err = os.Rename(path.Join(targetDir, "file2.txt"), path.Join(targetDir, "renamed_file2.txt"))
 	require.NoError(t, err)
--- a/tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
+++ b/tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
@@ -92,52 +92,146 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_AlwaysCacheHit(t *test
 // (b) Second ReadDir() will also be served from GCSFuse filesystem, because of
 // addition of new file.
 func (s *infiniteKernelListCacheTest) TestKernelListCache_CacheMissOnAdditionOfFile(t *testing.T) {
-	operations.CreateDirectory(path.Join(testDirPath, "explicit_dir"), t)
+	targetDir := path.Join(testDirPath, "explicit_dir")
+	operations.CreateDirectory(targetDir, t)
 	// Create test data
-	f1 := operations.CreateFile(path.Join(testDirPath, "explicit_dir", "file1.txt"), setup.FilePermission_0600, t)
+	f1 := operations.CreateFile(path.Join(targetDir, "file1.txt"), setup.FilePermission_0600, t)
 	operations.CloseFile(f1)
-	f2 := operations.CreateFile(path.Join(testDirPath, "explicit_dir", "file2.txt"), setup.FilePermission_0600, t)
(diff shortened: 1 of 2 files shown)


tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
```diff
@@ -88,6 +88,152 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_AlwaysCacheHit(t *test
 	assert.Equal(t, "file2.txt", names2[1])
 }
 
+// (a) First ReadDir() will be served from GCSFuse filesystem.
+// (b) Second ReadDir() will also be served from GCSFuse filesystem, because of
+// addition of new file.
+func (s *infiniteKernelListCacheTest) TestKernelListCache_CacheMissOnAdditionOfFile(t *testing.T) {
+	targetDir := path.Join(testDirPath, "explicit_dir")
```
ashmeenkaur, line 95: If this is something we have to do for every test, shall we add it to setup function?

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
```diff
@@ -88,6 +88,152 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_AlwaysCacheHit(t *test
...
+	require.Equal(t, "file2.txt", names1[1])
+	err = f.Close()
+	assert.NoError(t, err)
+
+	// Adding one object to make sure to change the ReadDir() response.
+	client.CreateObjectInGCSTestDir(ctx, storageClient, testDirName, path.Join("explicit_dir", "file3.txt"), "", t)
+	// Waiting for 5 seconds to see if the kernel cache expires.
+	time.Sleep(5 * time.Second)
+
+	// Ideally no invalidation since infinite ttl, but creation of a new file inside
```
ashmeenkaur, line 175:
```suggestion
	// Ideally no invalidation since infinite ttl, but deletion of file inside
```

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
```diff
@@ -88,6 +88,152 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_AlwaysCacheHit(t *test
...
+	require.NoError(t, err)
+
+	f, err = os.Open(targetDir)
+	assert.NoError(t, err)
+	names2, err := f.Readdirnames(-1)
+
+	assert.NoError(t, err)
+	require.Equal(t, 2, len(names2))
+	assert.Equal(t, "file1.txt", names2[0])
+	assert.Equal(t, "file3.txt", names2[1]) // file2.txt deleted, hence file3.txt
```
ashmeenkaur, line 187:
```suggestion
	assert.Equal(t, "file3.txt", names2[1])
```

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
```diff
@@ -88,6 +88,152 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_AlwaysCacheHit(t *test
...
+	require.Equal(t, 2, len(names1))
+	require.Equal(t, "file1.txt", names1[0])
+	require.Equal(t, "file2.txt", names1[1])
+	err = f.Close()
+	assert.NoError(t, err)
+
+	// Adding one object to make sure to change the ReadDir() response.
+	client.CreateObjectInGCSTestDir(ctx, storageClient, testDirName, path.Join("explicit_dir", "file3.txt"), "", t)
+	// Waiting for 5 seconds to see if the kernel cache expires.
+	time.Sleep(5 * time.Second)
```
ashmeenkaur, line 219: This sleep is not adding any value considering these tests are for infinite cache ttl. Shall we remove this?
For other tests too.

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
```diff
@@ -88,6 +88,152 @@ func (s *infiniteKernelListCacheTest) TestKernelListCache_AlwaysCacheHit(t *test
...
+
+	// Ideally no invalidation since infinite ttl, but creation of a new file inside
```
ashmeenkaur, line 221:
```suggestion
	// Ideally no invalidation since infinite ttl, but rename of a file inside
```

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
Tulsishah: Isn't it in guideline of writing tests that we should not sacrifice readability over  repetition?

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
Tulsishah: Umm...wdyt @raj-prince ?

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
ashmeenkaur: I don't think moving the directory setup to setup function will reduce test readability since it is part of the setup but if you think otherwise, you can skip :)

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
raj-prince: We have added this in one test, we can remove from other tests.

tools/integration_tests/kernel-list-cache/infinite_kernel_list_cache_test.go
Tulsishah: then targetDir and all we have to keep it as global variable. Prince was not align with it. So let's keep it like this only.

ashmeenkaur (approved):

PR #2155, opened 2024-07-11
Adding e2e test for kernel-list-cache concurrency test 

raj-prince:
### Description
**Test case added**
(a) Added test for parellel OpenDir and LookUp concurrently.
(b) Added test for parallel ReadDir and LookUp concurrently.
(c) Added test for parallel ReadDir by 10 go routines.
(d) Added test for concurrent ReadDir and delete/create/rename on file underneath.
(e) Added test for concurrent ReadDir and delete/create/rename folder underneath.
(f) Added test for concurrent ReadDir and FileEdit underneath.
(g) Added a kitchen task to run multiple operations concurrently.

**Parallelize the test within test-suite**
Test time got  reduced from 300s to 170s (on n2-std-96 vm). Ref: https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently

NA

### Testing details
1. Manual - Tested on n2-std-96 vm.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/concurrent_operations/concurrent_listing_test.go
+++ b/tools/integration_tests/concurrent_operations/concurrent_listing_test.go
@@ -15,10 +15,18 @@
 package concurrent_operations
 
 import (
+	"fmt"
+	"os"
+	"path"
+	"strings"
+	"sync"
 	"testing"
+	"time"
 
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/operations"
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/setup"
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/test_setup"
+	"github.com/stretchr/testify/assert"
 )
 
 ////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -33,20 +44,486 @@ func (s *concurrentListingTest) Setup(t *testing.T) {
 
 func (s *concurrentListingTest) Teardown(t *testing.T) {}
 
+// createDirectoryStructureForTestCase creates initial directory structure in the
+// given testCaseDir.
+// bucket
+//
+//		file1.txt
```
ashmeenkaur, line 51: file1.txt and file2.txt are not created

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -33,20 +44,486 @@ func (s *concurrentListingTest) Setup(t *testing.T) {
...
+
+			err = f.Close()
+			assert.Nil(t, err)
+		}
+	}()
+
+	// Goroutine 2: Repeatedly stats
+	go func() {
+		defer wg.Done()
+		for i := 0; i < iterationsPerGoroutine; i++ {
```
ashmeenkaur, line 148: 40 stat calls will take significantly less time than ReadDir. Do you want to do unequal number of iterations in this case? Can we check this for other test cases too where operations in go-routines are different.

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -33,20 +44,486 @@ func (s *concurrentListingTest) Setup(t *testing.T) {
...
+	// Create multiple go routines to listing concurrently.
+	for i := 0; i < goroutineCount; i++ {
+		go func() {
+			defer wg.Done()
+
+			for j := 0; j < iterationsPerGoroutine; j++ {
+				f, err := os.Open(targetDir)
+				assert.Nil(t, err)
+
+				_, err = f.Readdirnames(-1) // Read all directory entries
```
ashmeenkaur, line 191: at some places we have used 0 is Readdirnames too. Let's use either -1 or 0 at all places

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -33,20 +44,486 @@ func (s *concurrentListingTest) Setup(t *testing.T) {
...
+		assert.FailNow(t, "Possible deadlock or race condition detected during Readdir and directory operations")
+	}
+}
+
+// Test_MultipleConcurrentOperations tests for potential deadlocks or race conditions when
+// listing, file or folder operations, stat, opendir, file modifications happening concurrently.
+func (s *concurrentListingTest) Test_MultipleConcurrentOperations(t *testing.T) {
+	t.Parallel() // Mark the test parallelizable.
+
+	testCaseDir := "Test_MultipleConcurrentOperations"
```
ashmeenkaur, line 417: Do you want to delete these directories after test completes?

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
raj-prince: Yes, thanks!

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -486,7 +483,7 @@ func (s *concurrentListingTest) Test_MultipleConcurrentOperations(t *testing.T)
 	// Goroutine 4: Repeatedly stats
 	go func() {
 		defer wg.Done()
-		for i := 0; i < iterationsPerGoroutine; i++ {
+		for i := 0; i < iterationsForMediumOperations; i++ {
```
ashmeenkaur, line 486: Stat should be a light operation?

ashmeenkaur (approved):

PR #2163, opened 2024-07-12
[bug fix] Kernel list cache remove directory operation fails when files are created on GCS 

ashmeenkaur:
### Description
Scenario:
1. Read directory (listing) with kernel list cache enabled
2. List response is cached in kernel.
3. Create a file on the same GCS bucket inside the same directory outside of the GCSFuse mount via another tool like pantheon, gsutil.
4. call os.RemoveAll on the same directory in GCS Fuse.
5. Kernel will use the list cache response to list & delete the files and the newly created file (outside GCSFuse mount) won't be deleted.
6. os.RemoveAll will fail with directory not empty error when trying to delete the parent directory.

Fix: 
1. As part of the RmDir call, we will invalidate the cache by setting the previous directory listing timestamp to 0. 
2. This will make the open dir call to invalidate the cache so os.RemoveAll will start working.

Note: even after this fix, rm -r from terminal continues to fail as there in that case, we don't get first RmDir call (which invalidates the cache). Ref: https://github.com/GoogleCloudPlatform/gcsfuse/blob/8d72ea9205d148739f9bf2d20dda7dda7f72eb47/internal/fs/fs.go#L1752

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - Added

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1832,9 +1832,12 @@ func (fs *fileSystem) RmDir(
 			err = fmt.Errorf("ReadEntries: %w", err)
 			return err
 		}
-		//Clear kernel list cache after removing a directory. This ensures remote
-		//GCS files are included in future directory listings for unlinking.
-		childDir.InvalidateKernelListCache()
+
+		if fs.kernelListCacheTTL > 0 {
+			// Clear kernel list cache after removing a directory. This ensures remote
+			// GCS files are included in future directory listings for unlinking.
+			childDir.InvalidateKernelListCache()
+		}
 
 		// Are there any entries?
 		if len(entries) != 0 {
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1832,6 +1832,9 @@ func (fs *fileSystem) RmDir(
 			err = fmt.Errorf("ReadEntries: %w", err)
 			return err
 		}
+		//Clear kernel list cache after removing a directory. This ensures remote
+		//GCS files are included in future directory listings for unlinking.
+		childDir.InvalidateKernelListCache()
 
 		// Are there any entries?
 		if len(entries) != 0 {
--- a/internal/fs/inode/base_dir.go
+++ b/internal/fs/inode/base_dir.go
@@ -243,6 +243,11 @@ func (d *baseDirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
 	return true
 }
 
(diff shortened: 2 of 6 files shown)


internal/fs/inode/dir.go
```diff
@@ -147,6 +147,10 @@ type DirInode interface {
 	// should be invalidated or not.
 	ShouldInvalidateKernelListCache(ttl time.Duration) bool
 
+	//InvalidateKernelListCache sets the prevDirListingTimeStamp to zero so that
```
raj-prince, line 150: Let's not use `prevDirListingTimeStamp` as these are internal details. We can say something like - 
`InvalidateKernelListCache guarantees that the subsequent list call will be served from GCSFuse.`

internal/fs/fs.go
```diff
@@ -1832,6 +1832,9 @@ func (fs *fileSystem) RmDir(
 			err = fmt.Errorf("ReadEntries: %w", err)
 			return err
 		}
+		//Clear kernel list cache after removing a directory. This ensures remote
+		//GCS files are included in future directory listings for unlinking.
+		childDir.InvalidateKernelListCache()
```
raj-prince, line 1837: Should we guard it inside kernel-list-cache-ttl flag?

internal/fs/inode/base_dir.go
```diff
@@ -243,6 +243,9 @@ func (d *baseDirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
 	return true
 }
 
+// Kernel list cache is not supported for baseDirInode.
```
raj-prince, line 246: Instead, we should say list operation is not supported for baseDirInode.

internal/fs/inode/dir_test.go
```diff
@@ -1608,3 +1608,12 @@ func (t *DirTest) TestRenameFolderWithNonExistentSourceFolder() {
 	_, err = t.bucket.GetFolder(t.ctx, renameFolderName)
 	ExpectTrue(errors.As(err, &notFoundErr))
 }
+
+func (t *DirTest) Test_InvalidateKernelListCache() {
+	d := t.in.(*dirInode)
+	d.prevDirListingTimeStamp = d.cacheClock.Now()
```
raj-prince, line 1614: nit: good to validate before prevDirListingTimeStamp is not zero.

raj-prince (commented):
LGTM, few minor comments.

raj-prince (approved):

PR #2173, opened 2024-07-16
Change rename folder signature 

Tulsishah:
### Description
- The signature of the RenameFolder function has been modified.
- We previously returned control.RenameFolderOperation, but to keep the control client integration within the bucket handle, we now return gcs.Folder.
- Unit testing a successful rename is not possible because the control client's rename folder operation returns a control.RenameFolderOperation response containing private data that we cannot access externally so we can not mock it in our unit tests.
- We've added a wait call directly within the bucket handle to block until the long-running operation completes successfully.(Previously we thought to add this call from where we call RenameFolder API)

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -477,14 +477,20 @@ func (b *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (err
 	return err
 }
 
-func (b *bucketHandle) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (resp *control.RenameFolderOperation, err error) {
+func (b *bucketHandle) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (folder *gcs.Folder, err error) {
 	req := &controlpb.RenameFolderRequest{
 		Name:                "projects/_/buckets/" + b.bucketName + "/folders/" + folderName,
 		DestinationFolderId: destinationFolderId,
 	}
-	resp, err = b.controlClient.RenameFolder(ctx, req)
+	resp, err := b.controlClient.RenameFolder(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("longrunning operation: %w", err)
+	}
 
-	return resp, err
+	controlFolder, err := resp.Wait(ctx)
+	folder = gcs.GCSFolder(b.bucketName, controlFolder)
+
+	return folder, err
 }
 
 // TODO: Consider adding this method to the bucket interface if additional
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -471,7 +471,7 @@ func (b *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (err
 	var callOptions []gax.CallOption
 
 	err = b.controlClient.DeleteFolder(ctx, &controlpb.DeleteFolderRequest{
-		Name: "projects/_/buckets/" + b.bucketName + "/folders/" + folderName,
+		Name: fmt.Sprintf("projects/_/buckets/%s/folders/%s", b.bucketName, folderName),
 	}, callOptions...)
 
 	return err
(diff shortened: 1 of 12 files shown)


internal/storage/bucket_handle.go
```diff
@@ -477,14 +477,23 @@ func (b *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (err
...
+		return nil, fmt.Errorf("longrunning operation: %w", err)
+	}
+
+	// Wait blocks until the long-running operation is completed,
+	// returning the response and any errors encountered.
+	controlFolder, err = resp.Wait(ctx)
+	folder = gcs.GCSFolder(b.bucketName, controlFolder)
 
-	return resp, err
+	return folder, err
```
vadlakondaswetha, line 496: what is returned here, the destination foldername? if yes, can't we write a unit test?

Tulsishah: It is crashing at the point of resp.Wait(ctx), as we are mocking resp as nil.
Reason: control client's rename folder operation returns a [control.RenameFolderOperation](https://github.com/googleapis/google-cloud-go/blob/main/storage/control/apiv2/auxiliary.go#L30) response containing private data that we cannot access externally so we can not mock it in our unit tests.

internal/storage/bucket_handle.go
```diff
@@ -477,14 +477,23 @@ func (b *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (err
 	return err
 }
 
-func (b *bucketHandle) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (resp *control.RenameFolderOperation, err error) {
+func (b *bucketHandle) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (folder *gcs.Folder, err error) {
```
vadlakondaswetha, line 480: line no: 473
I would suggest using fmt.sprintf . 

go/go-style/best-practices#string-concat-fmt

internal/storage/bucket_handle.go
```diff
@@ -477,14 +477,23 @@ func (b *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (err
...
+	var controlFolder *controlpb.Folder
 	req := &controlpb.RenameFolderRequest{
 		Name:                "projects/_/buckets/" + b.bucketName + "/folders/" + folderName,
 		DestinationFolderId: destinationFolderId,
 	}
-	resp, err = b.controlClient.RenameFolder(ctx, req)
+	resp, err := b.controlClient.RenameFolder(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("longrunning operation: %w", err)
```
vadlakondaswetha, line 488: nit: what does appennding longrunning operation mean here?

Tulsishah: There were two error handling scenarios: one for when RenameFolder received a long-running operation and another after fetching the full response. I added code to differentiate between these cases. 
However, since the Go SDK handles returning the appropriate errors, I removed the custom error handling.

Done

internal/storage/bucket_handle.go
Tulsishah: Done

internal/storage/bucket_handle.go
```diff
@@ -471,28 +471,37 @@ func (b *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (err
 	var callOptions []gax.CallOption
 
 	err = b.controlClient.DeleteFolder(ctx, &controlpb.DeleteFolderRequest{
-		Name: "projects/_/buckets/" + b.bucketName + "/folders/" + folderName,
+		Name: fmt.Sprintf("projects/_/buckets/%s/folders/%s", b.bucketName, folderName),
```
vadlakondaswetha, line 474: This is used at two places. please create a constant.

vadlakondaswetha (approved):

internal/storage/bucket_handle.go
vadlakondaswetha: you need to create a constant for projects/_/buckets/%s/folders/%s and then use it in fmt.sprintf

vadlakondaswetha (approved):

PR #2176, opened 2024-07-16
Move to stretchr testify for cache_handler_test.go

sethiay:
### Description
Move to stretchr testify for cache_handler_test.go

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/cache_handler_test.go
+++ b/internal/cache/file/cache_handler_test.go
@@ -37,15 +37,20 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/operations"
-	. "github.com/jacobsa/ogletest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
 )
 
 const HandlerCacheMaxSize = TestObjectSize + ObjectSizeToCauseEviction
 const ObjectSizeToCauseEviction = 20
 
-func TestCacheHandler(t *testing.T) { RunTests(t) }
+func TestCacheHandler(t *testing.T) {
+	suite.Run(t, new(cacheHandlerTest))
+}
 
 type cacheHandlerTest struct {
+	suite.Suite
 	jobManager      *downloader.JobManager
 	bucket          gcs.Bucket
 	fakeStorage     storage.FakeStorage
@@ -57,9 +62,7 @@ type cacheHandlerTest struct {
 	cacheDir        string
 }
 
-func init() { RegisterTestSuite(&cacheHandlerTest{}) }
-
-func (chrT *cacheHandlerTest) SetUp(*TestInfo) {
+func (chrT *cacheHandlerTest) SetupTest() {
 	locker.EnableInvariantsCheck()
 	chrT.cacheDir = path.Join(os.Getenv("HOME"), "cache/dir")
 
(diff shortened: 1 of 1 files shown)


internal/cache/file/cache_handler_test.go
```diff
@@ -71,7 +73,7 @@ func (chrT *cacheHandlerTest) SetUp(*TestInfo) {
 	// Create test object in the bucket.
 	testObjectContent := make([]byte, TestObjectSize)
 	_, err := rand.Read(testObjectContent)
-	AssertEq(nil, err)
+	assert.NoError(chrT.T(), err)
```
kislaykishore, line 76: Should you use require.NoError to immediately abort test execution if there is an error? Otherwise, despite the error, the next set of statements would continue getting executed.  Please review other usages as well.

require.* methods specify precondition - something that must be satisfied unless there is something wrong with the tests. However, for tests that don't follow the Arrange-Act-Assert pattern e.g. have multiple assert blocks, it's not clear what's the precondition and what should be asserted. So, I guess, we can leave such cases as is and fix them once they are refactored into smaller tests.

kislaykishore (changes requested):

internal/cache/file/cache_handler_test.go
sethiay: I was using require.* in the discarded PR that I created. It will require me to go to each assert again and make that change. 

I have been making that change in the next PR (because that is easier for me using the discarded PR but anyways I will include here only, will take some time to do that)

internal/cache/file/cache_handler_test.go
```diff
@@ -166,71 +169,70 @@ func doesFileExist(filePath string) bool {
...
+	require.Equal(chrT.T(), jobStatusBefore.Name, downloader.Downloading)
 
 	err = chrT.cacheHandler.cleanUpEvictedFile(&fileInfoData)
 
-	ExpectEq(nil, err)
+	assert.NoError(chrT.T(), err)
 	jobStatusAfter := fileDownloadJob.GetStatus()
-	ExpectEq(jobStatusAfter.Name, downloader.Invalid)
-	ExpectFalse(doesFileExist(chrT.downloadPath))
+	assert.Equal(chrT.T(), jobStatusAfter.Name, downloader.Invalid)
```
kislaykishore, line 198: The second argument is what's expected and the third argument is what's actual. This helps understand the error messages better: "failed because expected $this but found $that". However, here and at a few other places the order seems to have been reversed.

internal/cache/file/cache_handler_test.go
```diff
@@ -479,8 +480,8 @@ func (chrT *cacheHandlerTest) Test_GetCacheHandle_ConcurrentSameFile() {
 		var err error
 		cacheHandle, err := chrT.cacheHandler.GetCacheHandle(minObj, chrT.bucket, false, 0)
 
-		AssertEq(nil, err)
-		AssertEq(nil, cacheHandle.validateCacheHandle())
+		assert.NoError(chrT.T(), err)
+		assert.Nil(chrT.T(), cacheHandle.validateCacheHandle())
```
kislaykishore, line 484: nit: If we receive an error, perhaps we should not validate the cacheHandle?
```suggestion
		if assert.NoError(chrT.T(), err) {
		  assert.Nil(chrT.T(), cacheHandle.validateCacheHandle())
		}
```

kislaykishore (changes requested):

internal/cache/file/cache_handler_test.go
sethiay: This is not reversed in this PR and was in reversed format before as well. I can correct this and maybe some more but I can't guarantee to correct all in scope of this PR.

kislaykishore: Thanks!

internal/cache/file/cache_handler_test.go
sethiay: Yeah, I also believe it's not required. However, I don't want to remove this as part of this PR. Will take it as separate task from pd_comp_test_* PRs.

kislaykishore (approved):

ashmeenkaur (approved):
Relying on @kislaykishore's review

PR #2184, opened 2024-07-17
DeleteFolder fast stat bucket fix.

Tulsishah:
### Description
-  Added unit test of DeleteFolder in fast stat bucket.
-  We have implemented the DeleteFolder function similarly to [DeleteObject](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/caching/fast_stat_bucket.go#L273), which is incorrect because we are removing the object from the cache regardless of whether the call is successful or not. (We can address this in a separate pull request.) Although it will not raise any functional issue, but we can fix for new API.
- Currently I have addressed the issue for DeleteFolder method.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA


internal/storage/caching/fast_stat_bucket_test.go
```diff
@@ -783,3 +784,25 @@ func (t *StatObjectTest) TestRenameFolder() {
 	AssertEq(nil, err)
 	ExpectEq(result, folder)
 }
+
+func (t *StatObjectTest) TestShouldDeleteFolder() {
```
raj-prince, line 788: nit: you can renamed the test like this - Test_DeleteFolder_Success, Test_DeleteFolder_Failure

raj-prince (changes requested):
Please incorporate.

internal/storage/caching/fast_stat_bucket_test.go
Tulsishah: Done

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -279,8 +279,12 @@ func (b *fastStatBucket) DeleteObject(
 }
 
 func (b *fastStatBucket) DeleteFolder(ctx context.Context, folderName string) error {
-	b.invalidate(folderName)
 	err := b.wrapped.DeleteFolder(ctx, folderName)
+	if err != nil {
+		return err
+	}
+	b.invalidate(folderName)
```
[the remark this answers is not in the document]
Tulsishah: Done

Tulsishah:
> Please incorporate.

Done

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -279,8 +288,14 @@ func (b *fastStatBucket) DeleteObject(
...
```
raj-prince, line 295: I feel, you don't have to invalidate since you are overwriting the entry.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -97,6 +97,15 @@ func (b *fastStatBucket) addNegativeEntry(name string) {
 	b.cache.AddNegativeEntry(name, expiration)
 }
 
+// LOCKS_EXCLUDED(b.mu)
+func (b *fastStatBucket) addNegativeEntryForFolder(name string) {
```
raj-prince, line 101: should we write unit test for this?

internal/storage/caching/fast_stat_bucket.go
Tulsishah: Makes sense. Removed it.

internal/storage/caching/fast_stat_bucket.go
Tulsishah: I don't think so, it is just calling cache `AddNegativeEntryForFolder` only with expiration which have unit test in stat cache file. There are similar method in the file also.

raj-prince: Okay, atleast we are testing as part of public bucket method.

raj-prince (approved):
LGTM, but we need to create a task to remove locks from here. As this is used to guard shared lock which comes with their own lock.

PR #2187, opened 2024-07-18
[HNS] Add method to parallely call stat and get folder object

ankitaluthra1:
### Description

- This PR is part of implicit dirs changes
- add findExplicitFolder in dir.go, method to call new get folder api in hns case , no usages yet, will be used in subsequent changes.
- makes metageneration in folder.go camel case
- refactor create folder in fake bucket to return created folder
- fix TODOs in unit test to use create folder

NA

### Testing details
1. Manual - NA
2. Unit tests - Done
3. Integration tests - NA


charith87 (approved):

internal/storage/gcs/folder.go
```diff
@@ -23,15 +23,15 @@ import (
...
 }
 
 func GCSFolder(bucketName string, attrs *controlpb.Folder) *Folder {
 	// Setting the parameters in Folder and doing conversions as necessary.
 	return &Folder{
 		Name:           getFolderName(bucketName, attrs.Name),
-		Metageneration: attrs.Metageneration,
+		MetaGeneration: attrs.Metageneration,
 		UpdateTime:     attrs.GetUpdateTime().AsTime(),
 	}
```
Tulsishah, line 36: Why are we changing `Metageneration` to `MetaGeneration` if control folder is returning in `Metageneration` format then shouldn't we keep the same?

ankitaluthra1: no Metageneration is not correct convention, if there is incorrect naming convention in library method, we should not leak it to our code as well, also MinObject also has MetaGeneration to maintain consistency.

internal/storage/gcs/folder.go
```diff
@@ -46,3 +46,12 @@ func getFolderName(bucketName string, fullPath string) string {
 
 	return folderName
 }
+
+func (f *Folder) ConvertFolderToMinObject() *MinObject {
```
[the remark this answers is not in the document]
ankitaluthra1: yes agreed, will update with subsequent changes.

PR #2191, opened 2024-07-19
Implement rename folder function in dir Inode interface

Tulsishah:
### Description
- RenameFolder function in DirInode interface. We will call this function further from fs.go whenever renameDir call will make for HNS. Currently this function is not being called from anywhere.
- Unit test

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -943,7 +943,6 @@ func (d *dirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
 }
 
 func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (folder *gcs.Folder, err error) {
-
 	folder, err = d.bucket.RenameFolder(ctx, folderName, destinationFolderId)
 	if err != nil {
 		return nil, err
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -64,6 +64,8 @@ type DirInode interface {
 	// true.
 	LookUpChild(ctx context.Context, name string) (*Core, error)
 
+	RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (*gcs.Folder, error)
+
 	// Read the children objects of this dir, recursively. The result count
 	// is capped at the given limit. Internal caches are not refreshed from this
 	// call.
@@ -938,4 +940,16 @@ func (d *dirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
 
 	cachedDuration := d.cacheClock.Now().Sub(d.prevDirListingTimeStamp)
 	return cachedDuration >= ttl
+}
+
+func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (folder *gcs.Folder, err error) {
+
+	folder, err = d.bucket.RenameFolder(ctx, folderName, destinationFolderId)
+	if err != nil {
+		return nil, err
+	}
+
+	d.cache.Erase(folderName)
+
+	return
 }
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -943,16 +943,17 @@ func (d *dirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
 	return cachedDuration >= ttl
 }
 
-func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (folder *gcs.Folder, err error) {
-	folder, err = d.bucket.RenameFolder(ctx, folderName, destinationFolderId)
+func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinationFolderName string) (*gcs.Folder, error) {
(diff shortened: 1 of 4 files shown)


internal/fs/inode/dir.go
```diff
@@ -939,3 +942,17 @@ func (d *dirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
...
+func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (folder *gcs.Folder, err error) {
+	folder, err = d.bucket.RenameFolder(ctx, folderName, destinationFolderId)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove old entry from type cache.
+	d.cache.Erase(folderName)
+	// Add new renamed folder in type cache.
+	d.cache.Insert(d.cacheClock.Now(), destinationFolderId, metadata.ExplicitDirType)
```
ankitaluthra1, line 955: @vadlakondaswetha should we add a TODO here, cache update will not needed once type cache usage is removed from hns

internal/fs/inode/dir_test.go
```diff
@@ -1565,3 +1565,46 @@ func (t *DirTest) TestShouldReturnNilWhenGCSFolderNotFound() {
...
+	_, err = t.bucket.GetFolder(t.ctx, folderName)
+	var notFoundErr *gcs.NotFoundError
+	ExpectTrue(errors.As(err, &notFoundErr))
+	// Verify the renamed folder exists.
+	_, err = t.bucket.GetFolder(t.ctx, renameFolderName)
+	AssertEq(nil, err)
+	AssertEq(renameFolderName, f.Name)
+}
+
+func (t *DirTest) TestShouldRenameFolderWithSrcFolderDoesNotExist() {
```
ankitaluthra1, line 1594: test name should indicate that exception will occur

internal/fs/inode/dir.go
```diff
@@ -64,6 +64,9 @@ type DirInode interface {
 	// true.
 	LookUpChild(ctx context.Context, name string) (*Core, error)
 
+	// Rename the directiory/folder.
+	RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (*gcs.Folder, error)
```
ankitaluthra1, line 68: can we make them consistent .. folderName string, destinationFolderId string, either folderName, destinationFolderName or folderId, destinationFolderId

internal/fs/inode/dir.go
```diff
@@ -939,3 +942,17 @@ func (d *dirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
...
+		return nil, err
+	}
+
+	// Remove old entry from type cache.
+	d.cache.Erase(folderName)
+	// Add new renamed folder in type cache.
+	d.cache.Insert(d.cacheClock.Now(), destinationFolderId, metadata.ExplicitDirType)
+
+	return
+}
```
ankitaluthra1, line 958: we should prefer named returns instead of naked returns to improve readability and avoid unexpected returns, [ref](https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#named-result-parameters:~:text=Once%20it%27s%20a%20medium%2Dsized%20function%2C%20be%20explicit%20with%20your%20returned%20values.)

internal/fs/inode/dir.go
```diff
@@ -939,3 +942,17 @@ func (d *dirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
...
 	return cachedDuration >= ttl
 }
+
+func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (folder *gcs.Folder, err error) {
+	folder, err = d.bucket.RenameFolder(ctx, folderName, destinationFolderId)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove old entry from type cache.
```
ankitaluthra1, line 952: No need to update stat cache here ?

ankitaluthra1 (changes requested):

internal/fs/inode/dir.go
Tulsishah: Stat cache will update when we call RenameFolder method as a part of fast stat bucket wrapper.

internal/fs/inode/dir_test.go
Tulsishah: Can you please check again?

internal/fs/inode/dir.go
Tulsishah: In the RenameFolder API, the control client currently requires Name and DestinationFolderId. I thought to keep the same structure but as it is part of file system logic we can change it.

Done.

internal/fs/inode/dir.go
Tulsishah: Done

internal/fs/inode/dir.go
Tulsishah: Done

internal/fs/inode/dir_test.go
ankitaluthra1: lgtm

ankitaluthra1 (approved):

PR #2193, opened 2024-07-21
Migrate goo.gl links to TinyURL.

jacobsa:
Google is [shutting down](https://developers.googleblog.com/en/google-url-shortener-links-will-no-longer-be-available/) `goo.gl` short link redirects (after [previously announcing](https://developers.googleblog.com/en/transitioning-google-url-shortener-to-firebase-dynamic-links/) that they'll continue to work).

Google also previously announced that you could export your links from "the `goo.gl` console", but I can't find that anywhere (the site is now just a redirect to their stupid blog post), so I had to do this by hand. There may be some errors.


ashmeenkaur (approved):
Thanks for creating this PR!

PR #2207, opened 2024-07-23
Lookup for hns 1

ankitaluthra1:
### Description

- This PR is the first PR for implementing implicit-dirs changes as part of HNS support.
- This contains changes in lookupInode implementation to use get folder api instead of ListObjects in case of HNS and corresponding tests for it.
- This also covers fix for an edge case in stat cache in which same name objects and folder can co-exist in HNS bucket.

NA

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Done


internal/cache/metadata/stat_cache.go
```diff
@@ -247,8 +247,19 @@ func (sc *statCacheBucketView) LookUpFolder(
...
 	hit, entry := sc.sharedCacheLookup(folderName, now)
+
 	if hit {
-		return hit, entry.f
+		// Adds negative entry scenario
+		if entry.f == nil && entry.m == nil {
+			return true, nil
+		}
+
+		// Adds uninitiated object scenario
```
charith87, line 257: What is uninitiated object scenario? Could we add clarity here.

internal/cache/metadata/stat_cache.go
```diff
@@ -247,8 +247,19 @@ func (sc *statCacheBucketView) LookUpFolder(
 	now time.Time) (bool, *gcs.Folder) {
 	// Look up in the LRU cache.
 	hit, entry := sc.sharedCacheLookup(folderName, now)
+
 	if hit {
-		return hit, entry.f
+		// Adds negative entry scenario
+		if entry.f == nil && entry.m == nil {
```
charith87, line 253: Are there older tests for stat cache negative scenario and are they passing?

internal/fs/inode/dir.go
```diff
@@ -366,10 +365,13 @@ func findExplicitFolder(ctx context.Context, bucket *gcsx.SyncerBucket, name Nam
 		return nil, fmt.Errorf("error in get folder for lookup : %w", folderErr)
 	}
 
+	folderObject := folderResult.ConvertFolderToMinObject()
+	folderObject.Name = name.objectName
```
charith87, line 369: When would the folderName and the requestedName be different and would it be possible to move it inside the convertFolderToMinObject() if its needed in more places than here.

ankitaluthra1: This is the test representing TestCopyDirectoryInNonExistingDirectory case when folderName and the requestedName are different.

internal/cache/metadata/stat_cache.go
ankitaluthra1: yes *StatCacheTest* has multiple stat cache negative scenarios for both object and folders which is passing for this PR.

internal/cache/metadata/stat_cache.go
ankitaluthra1: Updated comment to be more descriptive, also corresponding test is also updated test name (Test_ShouldReturnHitFalseWhenOnlyObjectAlreadyHasEntry) for better documentation

internal/fs/inode/dir.go
ankitaluthra1: Moved name inside and also added test case for same.

charith87 (approved):

PR #2209, opened 2024-07-23
Upgrade gem lock dependency

Tulsishah:
### Description
Dependabot dependency
https://github.com/GoogleCloudPlatform/gcsfuse/pull/2205/commits/0cb462f80ec56855a6d7e57eeb4e4f5b3b9dc341

NA

### Testing details
1. Manual - Package building successfully
2. Unit tests - NA
3. Integration tests - NA
4. Perf tests - B8dees2utG5HZvL

### DIFF
--- a/tools/gem_dependency/Gemfile.lock
+++ b/tools/gem_dependency/Gemfile.lock
@@ -23,8 +23,8 @@ GEM
       insist
       mustache (= 0.99.8)
       stud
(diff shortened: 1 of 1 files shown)


ashmeenkaur (approved):
This is only being used during package building so LGTM if package is building fine after the dependency upgrade

PR #2215, opened 2024-07-24
Add kernel-cache-ttl=0 in e2e tests

Tulsishah:
### Description
Add kernel cache ttl=0 in e2e tests

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Autoamated

### DIFF
--- a/tools/integration_tests/concurrent_operations/setup_test.go
+++ b/tools/integration_tests/concurrent_operations/setup_test.go
@@ -62,7 +62,7 @@ func TestMain(m *testing.M) {
 	setup.SetUpTestDirForTestBucketFlag()
 
 	flagsSet := [][]string{
-		{"--kernel-list-cache-ttl-secs=-1"},
+		{"--kernel-list-cache-ttl-secs=-1", "--kernel-list-cache-ttl-secs=0"},
 	}
 	successCode := static_mounting.RunTests(flagsSet, m)
 


tools/integration_tests/concurrent_operations/setup_test.go
```diff
@@ -62,7 +62,7 @@ func TestMain(m *testing.M) {
 	setup.SetUpTestDirForTestBucketFlag()
 
 	flagsSet := [][]string{
-		{"--kernel-list-cache-ttl-secs=-1"},
+		{"--kernel-list-cache-ttl-secs=-1", "--kernel-list-cache-ttl-secs=0"},
```
gargnitingoogle, line 65: @Tulsishah is there any expected (significant) increase in runtime of e2e tests with this change? 

If yes, then we should increase the timeout for e2e tests.

gargnitingoogle (approved):
LGTM

tools/integration_tests/concurrent_operations/setup_test.go
```diff
@@ -62,7 +62,7 @@ func TestMain(m *testing.M) {
...
 	}
```
charith87, line 66: This is only for concurrent tests right. Not for other tests right, because there are tests which may fail due to list caching.

charith87 (approved):
Approving.

tools/integration_tests/concurrent_operations/setup_test.go
Tulsishah: Yes only for concurrent operations.

PR #2218, opened 2024-07-25
Set EnableEmptyManagedFolders if EnableHns is set

kislaykishore:
### Description
Add a function to post-process config-params.
Currently it only caters to List.EnableEmptyManagedFolders but it will be expanded to handle MetadataCache attributes in the future.
I haven't removed the similar vetting config in mount_config since EnableEmptyManagedFolders has not yet been migrated to the new config.

NA

### Testing details
1. Manual - NA
3. Integration tests - NA


cfg/vet.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cfg
+
+func VetConfig(c *Config) {
```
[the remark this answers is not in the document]
gargnitingoogle: Plus, the word `vet` doesn't sound right here. Vetting means examining, checking, but this actually changing the value of the config. How about `Rationalize` ?

cfg/vet.go
```diff
@@ -0,0 +1,22 @@
...
+	// The EnableEmptyManagedFolders flag must be set to true to enforce folder prefixes for Hierarchical buckets.
+	if c.EnableHns {
+		c.List.EnableEmptyManagedFolders = true
```
gargnitingoogle, line 20: How about the following instead of the complete if-block ?

```suggestion
		c.List.EnableEmptyManagedFolders |= c.EnableHns
```

cmd/config_vetting_test.go
```diff
@@ -0,0 +1,59 @@
...
+			args:     []string{"--enable-hns", "--enable-empty-managed-folders=false"},
+			expected: true,
+		},
+		{
+			name:     "enable-hns not true but enable-empty-managed-folders set to true",
+			args:     []string{"--enable-hns=false", "--enable-empty-managed-folders=true"},
+			expected: true,
+		},
+		{
+			name:     "both enable-hns and enable-empty-managed-folders not true",
```
gargnitingoogle, line 45: For better readability

```suggestion
			name:     "both enable-hns and enable-empty-managed-folders set to false",
```

cmd/legacy_param_mapper_test.go
```diff
@@ -504,3 +504,53 @@ func TestInvalidClientProtocol(t *testing.T) {
...
+		expectedEnableEmptyManagedFolders bool
+	}{
+		{
+			name:                              "both enable-hns and enable-empty-managed-folders set to true",
+			enableHns:                         true,
+			enableEmptyManagedFolders:         true,
+			expectedEnableEmptyManagedFolders: true,
+		},
+		{
+			name:                              "enable-hns set to true",
```
gargnitingoogle, line 522:
```suggestion
			name:                              "enable-hns set to true, but enableEmptyManagedFolders set to false",
```

cmd/legacy_param_mapper_test.go
```diff
@@ -504,3 +504,53 @@ func TestInvalidClientProtocol(t *testing.T) {
...
+			expectedEnableEmptyManagedFolders: true,
+		},
+		{
+			name:                              "enable-hns not true but enable-empty-managed-folders set to true",
+			enableHns:                         false,
+			enableEmptyManagedFolders:         true,
+			expectedEnableEmptyManagedFolders: true,
+		},
+		{
+			name:                              "both enable-hns and enable-empty-managed-folders not true",
```
gargnitingoogle, line 534:
```suggestion
			name:                              "both enable-hns and enable-empty-managed-folders set to false",
```

cmd/legacy_param_mapper_test.go
```diff
@@ -504,3 +504,53 @@ func TestInvalidClientProtocol(t *testing.T) {
...
+			flags := &flagStorage{
+				ClientProtocol: mountpkg.ClientProtocol("http1"),
+			}
+			c := config.NewMountConfig()
+			c.EnableHNS = tc.enableHns
+			c.ListConfig.EnableEmptyManagedFolders = tc.enableEmptyManagedFolders
+
+			resolvedConfig, err := PopulateNewConfigFromLegacyFlagsAndConfig(&mockCLIContext{}, flags, c)
+
+			if assert.NoError(t, err) {
```
gargnitingoogle, line 551: When would `assert.NoError(t, err)` be false ?

cmd/legacy_param_mapper_test.go
```diff
@@ -288,7 +288,7 @@ func TestPopulateConfigFromLegacyFlags(t *testing.T) {
 				IgnoreInterrupts:          false,
 				AnonymousAccess:           false,
 				KernelListCacheTtlSeconds: -1,
-				MaxRetryAttempts:          100,
+				MaxRetryAttempts:          15,
```
gargnitingoogle, line 291: What is this change for ?

gargnitingoogle: the same goes for the next change.

gargnitingoogle (changes requested):
Add some review comments.

cmd/legacy_param_mapper_test.go
gargnitingoogle: > When error is not nil.

No, I understood that, but where will the error in it come from? If it's from the test case, then we should have had a `break` on the previous test-case failure rather than calling the function `PopulateNewConfigFromLegacyFlagsAndConfig` in this iteration, but not using/testing its value. This seems redundant.

cmd/config_vetting_test.go
```diff
@@ -0,0 +1,59 @@
...
+			args:     []string{"--enable-hns"},
+			expected: true,
+		},
+		{
+			name:     "enable-hns set to true but enable-empty-managed-folders set to false",
+			args:     []string{"--enable-hns", "--enable-empty-managed-folders=false"},
+			expected: true,
+		},
+		{
+			name:     "enable-hns not true but enable-empty-managed-folders set to true",
```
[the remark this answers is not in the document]
gargnitingoogle: Oh, my bad. I got misled by the original comment `enable-hns not true but enable-empty-managed-folders set to true` .

cfg/vet.go
gargnitingoogle: yeah, fair point.

cmd/legacy_param_mapper_test.go
gargnitingoogle: I get it. My point is that we're skipping the test for `resolvedConfig.List.EnableEmptyManagedFolders` if `PopulateNewConfigFromLegacyFlagsAndConfig(&mockCLIContext{}, flags, c)` failed. This sort of reduces the effectiveness of this test.

Is it possible for `PopulateNewConfigFromLegacyFlagsAndConfig(&mockCLIContext{}, flags, c)` in this scenario to fail?

kislaykishore: If `PopulateNewConfigFromLegacyFlagsAndConfig` returns an error, the test fails. Also, if error is returned, the value of resolvedConfig could be anything - there are no expectations on what its value is. The only expectation is that it should not be used. So, there is no point asserting on the value.

> Is it possible for PopulateNewConfigFromLegacyFlagsAndConfig(&mockCLIContext{}, flags, c) in this scenario to fail?

I guess I'm not getting your concern. If the call fails, the test fails (assert.NoError(...) also causes the test to fail) and if it doesn't we assert on the value. So, I guess we are covered regardless of whether the call fails.

cfg/rationalize.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cfg
+
+func Rationalize(c *Config) {
```
ashmeenkaur, line 17: Let's add a comment on top for documentation as this is an exported functions. Also shall we rename this to: UpdateDependentConfigs ?

cmd/legacy_param_mapper.go
```diff
@@ -150,6 +150,7 @@ func PopulateNewConfigFromLegacyFlagsAndConfig(c cliContext, flags *flagStorage,
 	if err := cfg.ValidateConfig(resolvedConfig); err != nil {
 		return nil, err
 	}
+	cfg.Rationalize(resolvedConfig)
```
ashmeenkaur, line 153: Shall we do this before validation?
Same in root.go

cmd/legacy_param_mapper_test.go
```diff
@@ -504,3 +504,53 @@ func TestInvalidClientProtocol(t *testing.T) {
 
 	assert.NotNil(t, err)
 }
+
+func TestEnableEmptyManagedFoldersVetting(t *testing.T) {
```
ashmeenkaur, line 508: I think we should add this test to cfg package. We can add a happy path test to the existing test to validate Rationalize is called.

Saying this because eventually we are going to get rid of PopulateNewConfigFromLegacyFlagsAndConfig and there will be more migration work at that time.

kislaykishore: Done. I've added unit tests for Rationalize().

cmd/legacy_param_mapper.go
kislaykishore: `ValidateConfig` ensures that the user-specified config is valid and `Rationalize` augments the config. 
* If `ValidateConfig` fails then user specified a wrong config; something that we don't support and err out.
* If `ValidateConfig` passes then `Rationalize` is just business logic to augment the config.

If we `ValidateConfig` after `Rationalize` then we are combining validation for both user-specified configuration as well as the augmented config which seems confusing.

ashmeenkaur (approved):

gargnitingoogle (approved):
LGTM

PR #2221, opened 2024-07-26
Pass directories instead of files to codegen

kislaykishore:
* Specify directories for template and output files. As we add more templates and have more output files, specifying individual paths will become cumbersome.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


tools/config-gen/main.go
```diff
@@ -40,26 +41,27 @@ type templateData struct {
...
-		return fmt.Errorf("template filename cannot be empty")
+	if *templateDir == "" {
+		return fmt.Errorf("template directory cannot be empty")
 	}
 	return nil
 }
 
-func write(data templateData) error {
-	outputFile, err := os.Create(*outFile)
+func write(data any, outF, tmplF string) error {
```
vadlakondaswetha, line 55: Can you put more descriptive names to these parameters. 
here data is of any type, it will not be easy to understand what data is getting passed and why is it of any type?

kislaykishore: Changed the names to be more descriptive.

tools/config-gen/main.go
```diff
@@ -104,7 +108,7 @@ func main() {
 		FlagTemplateData: fd,
 		TypeTemplateData: td,
 		Backticks:        "`",
-	})
+	}, path.Join(*outDir, "config.go"), path.Join(*templateDir, "config.tpl"))
```
gargnitingoogle, line 111: nit: the formatting seems a bit odd here. These two (or at least one of these two) should have gone to the next row I think. Could you check?

gargnitingoogle (changes requested):

tools/config-gen/main.go
```diff
@@ -40,30 +41,33 @@ type templateData struct {
...
 	return nil
 }
 
-func write(data templateData) error {
-	outputFile, err := os.Create(*outFile)
+// write applies the dataObj which could be an object of any type, to the
+// templateFile and writes the generated text into the outputFile.
+func write(dataObj any, outputFile, templateFile string) error {
+	o, err := os.Create(outputFile)
 	if err != nil {
```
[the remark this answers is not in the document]
kislaykishore: Can you check now?

tools/config-gen/main.go
```diff
@@ -40,34 +41,37 @@ type templateData struct {
...
+		closeErr := outF.Close()
+		if err == nil {
+			err = closeErr
+		}
+	}()
+
+	file := path.Base(templateFile)
+	if tmpl, err := template.New(file).ParseFiles(templateFile); err == nil {
+		return tmpl.Execute(outF, dataObj)
+	} else {
```
gargnitingoogle, line 72: Should just move the `return err` out of the else-block and remove the else-block (as it's unnecessary as the if-block always returns anyway).

gargnitingoogle (changes requested):
one minor comment.

tools/config-gen/main.go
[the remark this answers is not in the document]
gargnitingoogle: LGTM now

tools/config-gen/main.go
kislaykishore: I think the "err" within the "if" block shadows the "err" in the outer scope. So, modifications to the "err" within the if-block will probably have no effect on the "err" in the outer scope. That said, my understanding of Go-scopes could be wrong here. So, let me know if you think differently.

vadlakondaswetha (approved):

gargnitingoogle (dismissed):
one minor nitpick comment. Not blocking for merge.

gargnitingoogle (approved):
LGTM

vadlakondaswetha (approved):

PR #2226, opened 2024-07-29
Decode empty string as nil for URL

kislaykishore:
### Description
Instead of unmarshalling/decoding empty string as a pointer pointnig to an empty url.URL{} object, decode it as nil.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/decode_hook.go
+++ b/cfg/decode_hook.go
@@ -34,6 +34,9 @@ func stringToURLHookFunc() mapstructure.DecodeHookFuncType {
 			return data, nil
 		}
 		s := data.(string)
+		if s == "" {
+			return nil, nil
+		}
 		u, err := url.Parse(s)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


cfg/decode_hook.go
```diff
@@ -46,8 +49,11 @@ func stringToURLHookFunc() mapstructure.DecodeHookFuncType {
 func DecodeHook() mapstructure.DecodeHookFunc {
 	return mapstructure.ComposeDecodeHookFunc(
 		mapstructure.TextUnmarshallerHookFunc(),
-		stringToURLHookFunc(),
 		mapstructure.StringToTimeDurationHookFunc(), // default hook
 		mapstructure.StringToSliceHookFunc(","),     // default hook
+		// TODO: stringToURLHookFunc() needs to be in the end in order to avoid
+		// error being thrown.
```
ashmeenkaur, line 55: Maybe I understood this incorrectly but I don't think ordering will effect the error being thrown? Can you confirm?

PR #2235, opened 2024-07-29
[CLI config parity] migrate url type to string type

ashmeenkaur:
### Description
- removes url type from supported new config types. This was being used by custom end point parameter.
- Refactors StorageClientConfig's CustomEndPoint to string type (from url.Url earlier) for simplification.

NA

### Testing details
1. Manual - NA
2. Unit tests - Updated

### DIFF
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -17,11 +17,12 @@ package cfg
 import "net/url"
 
 func decodeURL(u string) (string, error) {
-	decodedUrl, err := url.Parse(u)
+	// TODO: check if we can replace url.Parse with url.ParseRequestURI.
+	decodedURL, err := url.Parse(u)
 	if err != nil {
 		return "", err
 	}
-	return decodedUrl.String(), nil
+	return decodedURL.String(), nil
 }
 
 // Rationalize updates the config fields based on the values of other fields.
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -14,10 +14,28 @@
 
 package cfg
 
+import "net/url"
+
+func decodeURL(u string) (string, error) {
+	decodedUrl, err := url.Parse(u)
+	if err != nil {
+		return "", err
+	}
+	return decodedUrl.String(), nil
+}
+
 // Rationalize updates the config fields based on the values of other fields.
-func Rationalize(c *Config) {
+func Rationalize(c *Config) error {
 	// The EnableEmptyManagedFolders flag must be set to true to enforce folder prefixes for Hierarchical buckets.
 	if c.EnableHns {
 		c.List.EnableEmptyManagedFolders = true
 	}
+
+	var err error
+	c.GcsConnection.CustomEndpoint, err = decodeURL(c.GcsConnection.CustomEndpoint)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/cfg/rationalize_test.go
+++ b/cfg/rationalize_test.go
@@ -75,7 +75,7 @@ func TestRationalizeCustomEndpointSuccessful(t *testing.T) {
 		expectedCustomEndpoint string
 	}{
 		{
-			name: "Valid Config where input and expected custom endpoint matches.",
+			name: "Valid Config where input and expected custom endpoint match.",
 			config: &Config{
 				GcsConnection: GcsConnectionConfig{
 					CustomEndpoint: "https://bing.com/search?q=dotnet",
--- a/cfg/rationalize_test.go
+++ b/cfg/rationalize_test.go
@@ -68,42 +68,29 @@ func TestRationalizeEnableEmptyManagedFolders(t *testing.T) {
 	}
 }
 
-func TestRationalizeCustomEndpoint(t *testing.T) {
+func TestRationalizeCustomEndpointSuccessful(t *testing.T) {
 	testCases := []struct {
 		name                   string
 		config                 *Config
 		expectedCustomEndpoint string
-		wantErr                bool
 	}{
 		{
-			name: "Valid Config 1",
+			name: "Valid Config where input and expected custom endpoint matches.",
 			config: &Config{
 				GcsConnection: GcsConnectionConfig{
 					CustomEndpoint: "https://bing.com/search?q=dotnet",
 				},
 			},
 			expectedCustomEndpoint: "https://bing.com/search?q=dotnet",
-			wantErr:                false,
 		},
 		{
-			name: "Valid Config 2",
+			name: "Valid Config where input and expected custom endpoint differ.",
 			config: &Config{
 				GcsConnection: GcsConnectionConfig{
 					CustomEndpoint: "https://j@ne:password@google.com",
 				},
 			},
 			expectedCustomEndpoint: "https://j%40ne:password@google.com",
-			wantErr:                false,
-		},
-		{
-			name: "Invalid Config",
-			config: &Config{
-				GcsConnection: GcsConnectionConfig{
-					CustomEndpoint: "a_b://abc",
-				},
-			},
-			expectedCustomEndpoint: "",
-			wantErr:                true,
 		},
 	}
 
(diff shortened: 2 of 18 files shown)


cfg/validate_test.go
```diff
@@ -0,0 +1,86 @@
...
+		Compress:        false,
+		MaxFileSizeMb:   1,
+	}
+}
+
+func TestValidateConfig(t *testing.T) {
+	testCases := []struct {
+		name                   string
+		config                 *Config
+		expectedCustomEndpoint string
```
kislaykishore, line 35: I think it's good enough to check whether Validate returns an error. This expectedCustomEndpoint doesn't add any value and can be removed.

cfg/validate.go
```diff
@@ -28,10 +29,25 @@ func isValidLogRotateConfig(config *LogRotateLoggingConfig) error {
...
 // ValidateConfig returns a non-nil error if the config is invalid.
 func ValidateConfig(config *Config) error {
-	if err := isValidLogRotateConfig(&config.Logging.LogRotate); err != nil {
+	var err error
+
+	if err = isValidLogRotateConfig(&config.Logging.LogRotate); err != nil {
 		return fmt.Errorf("error parsing log-rotate config: %w", err)
 	}
+
+	if config.GcsConnection.CustomEndpoint, err = isValidUrl(config.GcsConnection.CustomEndpoint); err != nil {
```
kislaykishore, line 48: Let's not change the parsed config in ValidateConfig. It's mandate is to just return an error if the config is invalid.

ashmeenkaur: Valid concern, but I am not sure what will be the best fix in this case. 

We can (1)move it / (2)duplicate the check to rationalize? 
(1) throwing an error during rationalize doesn't look right
(2) I am not in favor of duplicating

kislaykishore: I am okay with (2) i.e. duplicate the url.Parse call with a TODO to check whether there is any benefit of parsing the rawURL first and then changing the URL back to raw URL.

cfg/validate.go
```diff
@@ -28,10 +29,25 @@ func isValidLogRotateConfig(config *LogRotateLoggingConfig) error {
 	return nil
 }
 
+func isValidUrl(u string) (string, error) {
```
kislaykishore, line 32: As per Go convention it should be named as isValidURL.

cfg/validate_test.go
ashmeenkaur: Added a test case where this is being used. PTAL.

cfg/validate.go
ashmeenkaur: As discussed offline, rawURL parsing is required (added a unit test for it). We will continue with the current approach of doing the check and assignment in the validation method.

cfg/validate.go
```diff
@@ -28,10 +29,25 @@ func isValidLogRotateConfig(config *LogRotateLoggingConfig) error {
 	return nil
 }
 
+func isValidURL(u string) (string, error) {
```
gargnitingoogle, line 32: The function name is misleading. It sounds like it is only validating the given URI, but it's also trying to parse it and return the successfully parsed url or error. How about 

```suggestion
func decodedURL(u string) (string, error) {
```

or 

```suggestion
func parsedURL(u string) (string, error) {
```

gargnitingoogle (changes requested):
one minor readability comment. Please address.

cfg/validate.go
ashmeenkaur: @kislaykishore and I had a discussion and we decided to duplicate this in the Rationalise() function keeping the responsibility of validate and rationalise separate. I've taken these suggestions in the rationalise part.

cfg/rationalize.go
```diff
@@ -14,10 +14,28 @@
...
 package cfg
 
+import "net/url"
+
+func decodeURL(u string) (string, error) {
+	decodedUrl, err := url.Parse(u)
+	if err != nil {
+		return "", err
+	}
+	return decodedUrl.String(), nil
```
kislaykishore, line 24:
```suggestion
	decodedURL, err := url.Parse(u)
	if err != nil {
		return "", err
	}
	return decodedURL.String(), nil
```

cfg/validate.go
```diff
@@ -28,10 +29,25 @@ func isValidLogRotateConfig(config *LogRotateLoggingConfig) error {
...
 // ValidateConfig returns a non-nil error if the config is invalid.
 func ValidateConfig(config *Config) error {
-	if err := isValidLogRotateConfig(&config.Logging.LogRotate); err != nil {
+	var err error
+
+	if err = isValidLogRotateConfig(&config.Logging.LogRotate); err != nil {
 		return fmt.Errorf("error parsing log-rotate config: %w", err)
 	}
+
+	if err = isValidURL(config.GcsConnection.CustomEndpoint); err != nil {
```
kislaykishore, line 48: Can we reuse the `decodeURL` method here.

cmd/legacy_param_mapper.go
```diff
@@ -35,6 +35,14 @@ func PopulateNewConfigFromLegacyFlagsAndConfig(c cliContext, flags *flagStorage,
 		return nil, fmt.Errorf("PopulateNewConfigFromLegacyFlagsAndConfig: unexpected nil flags or mount config")
 	}
 
+	// resolve custom endpoint url type to string.
+	var customEndPoint string
+	if flags.CustomEndpoint == nil {
+		customEndPoint = ""
+	} else {
+		customEndPoint = flags.CustomEndpoint.String()
+	}
```
kislaykishore, line 44: This is more succinct.
```suggestion
	var customEndPoint string
	if flags.CustomEndpoint != nil {
		customEndPoint = flags.CustomEndpoint.String()
	} 
```

cfg/rationalize.go
```diff
@@ -14,10 +14,28 @@
 
 package cfg
 
+import "net/url"
+
+func decodeURL(u string) (string, error) {
+	decodedUrl, err := url.Parse(u)
```
kislaykishore, line 20: nit: Feel free to ignore - Can we add a TODO to replace `Parse` call with [ParseRequestURI](https://pkg.go.dev/net/url#ParseRequestURI).

kislaykishore (changes requested):

kislaykishore (approved):

cfg/validate_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}{
+		{
+			name: "Valid Config 1",
+			config: &Config{
+				Logging: LoggingConfig{LogRotate: validLogRotateConfig()},
+				GcsConnection: GcsConnectionConfig{
+					CustomEndpoint: "https://bing.com/search?q=dotnet",
+				},
+			},
+			wantErr: false,
```
gargnitingoogle, line 45: nit: We don't need to set the `false` value explicitly as it's default and is obvious.

gargnitingoogle (commented):
Added some readability comments/questions.

cfg/validate_test.go
ashmeenkaur: I think this is okay for test as it enhances the test readability

cfg/validate.go
```diff
@@ -28,10 +28,22 @@ func isValidLogRotateConfig(config *LogRotateLoggingConfig) error {
 	return nil
 }
 
+func isValidURL(u string) (err error) {
+	_, err = decodeURL(u)
+	return
+}
+
 // ValidateConfig returns a non-nil error if the config is invalid.
 func ValidateConfig(config *Config) error {
```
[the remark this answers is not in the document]
ashmeenkaur: Will do this in a separate PR to keep the scope of changes limited to PR description.

cfg/rationalize_test.go
```diff
@@ -59,9 +59,64 @@ func TestRationalize(t *testing.T) {
...
+			expectedCustomEndpoint: "",
+			wantErr:                true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actualErr := Rationalize(tc.config)
+
+			assert.Equal(t, tc.config.GcsConnection.CustomEndpoint, tc.expectedCustomEndpoint)
```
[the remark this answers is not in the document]
ashmeenkaur: I have separated out success and failure case test based on https://google.github.io/styleguide/go/decisions.html#table-driven-tests

cfg/rationalize_test.go
```diff
@@ -68,55 +68,63 @@ func TestRationalizeEnableEmptyManagedFolders(t *testing.T) {
...
+func TestRationalizeCustomEndpointSuccessful(t *testing.T) {
 	testCases := []struct {
 		name                   string
 		config                 *Config
 		expectedCustomEndpoint string
-		wantErr                bool
 	}{
 		{
-			name: "Valid Config 1",
+			name: "Valid Config where input and expected custom endpoint matches.",
```
gargnitingoogle, line 78: s/matches/match/

cfg/validate_test.go
```diff
@@ -28,53 +28,61 @@ func validLogRotateConfig() LogRotateLoggingConfig {
...
 	testCases := []struct {
-		name    string
-		config  *Config
-		wantErr bool
+		name   string
+		config *Config
 	}{
 		{
-			name: "Valid Config 1",
+			name: "Valid Config where input and expected custom endpoint matches.",
```
gargnitingoogle, line 37: s/matches/match/

gargnitingoogle (commented):
comment about minor typo

gargnitingoogle (approved):
LGTM

kislaykishore (approved):

charith87 (approved):

Issue #2220, opened 2024-07-25
ReadDir: input/output error, readAllEntries: fixConflictingNames: weird dirent type pair for name "<filename>": 8, 8 when reading and writing at the same time


timeyr:
When writing at least two files and reading the same directory "simultaneously", sometimes the following error occurs:

```
readAllEntries: fixConflictingNames: weird dirent type pair for name "<filename>": 8, 8
```

We are running gcsfuse version `2.0.1-gke.0` as a sidecar container in GKE.

To reproduce it, I have created a small Nodejs script running as a deployment in GKE, and it can reproduce the issue with about 10 % probability when writing 2 files and issuing 10 list directory requests at the same time.

Here is the debug log: [debug-gcsfuse-logs.json](https://github.com/user-attachments/files/16379992/debug-gcsfuse-logs.json)

I can also share the code and configuration used to create the test environment I've used to reproduce this (requires a GCP account/project of course).

Deployment annotations:
```
gke-gcsfuse/cpu-limit: "0"
gke-gcsfuse/cpu-request: 100m
gke-gcsfuse/memory-limit: 512M
gke-gcsfuse/memory-request: 256M
gke-gcsfuse/volumes: "true"
```

Mount options (w/o debugging):
```
- uid=-1
- gid=-1
- dir-mode=777
- file-mode=666
```

Tulsishah:
Hi @timeyr ,

This is a known issue. We will let you know once it's fixed. In the meantime, the workaround is to pass create-empty-file as true. [ref](https://cloud.google.com/storage/docs/gcsfuse-config-file)

timeyr:
Hi @Tulsishah !
This is great news! I will try out the workaround as soon as I have time. Cheers!

koi8:
A bit struggling to apply workaround in GKE cluster:
1. How to specify custom config file, if I don't know the content of current one? I want to change only one option and not accidentally break something else. 
2.  How to mount the configmap, if sidecar container is created by annotations following this guide - https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/cloud-storage-fuse-csi-driver .
There is no mentioning of configmaps.

raj-prince:
@koi8 sorry for the inconvenience!

1. You can try appending **write:create-empty-file:false** in the mountOptions.  E.g. mountOptions: "uid=-1,gid=-1,dir-mode=777,file-mode=666,**write:create-empty-file:false**"
[...]

PR #2237, opened 2024-07-29
Fixing error in listing for duplicate files

vadlakondaswetha:
When there is local file synced to GCS but not deleted from localFileInodes, listing fails with input/output error. This is because it detects duplicate entries with the same name.
Fix: Remove the duplicate entry when GCS entry matches with local file inode.

Issue link: https://github.com/GoogleCloudPlatform/gcsfuse/issues/2220

### Testing details
1. Manual - Done
2. Unit tests - Added as required
3. Integration tests - Existing tests are sufficient

### DIFF
--- a/internal/fs/handle/dir_handle.go
+++ b/internal/fs/handle/dir_handle.go
@@ -137,9 +137,9 @@ func fixConflictingNames(entries []fuseutil.Dirent, localEntries map[string]fuse
 
 		if eIsDir == prevIsDir {
 			if _, ok := localEntries[e.Name]; ok && !eIsDir {
-				// if we find same entry in GCS vs local file entries, i.e, the entry is
-				// uploaded to GCS but not yet deleted from local entries. Hence,
-				// don't return it as part of list response.
+				// We have found same entry in GCS and local file entries, i.e, the
+				// entry is uploaded to GCS but not yet deleted from local entries.
+				// Do not return the duplicate entry as part of list response.
 				continue
 			} else {
 				err = fmt.Errorf(
--- a/internal/fs/handle/dir_handle.go
+++ b/internal/fs/handle/dir_handle.go
@@ -106,7 +106,7 @@ func (dh *DirHandle) checkInvariants() {
 // GCS object names, to conflicting file names.
 //
 // Input must be sorted by name.
-func fixConflictingNames(entries []fuseutil.Dirent) (err error) {
+func fixConflictingNames(entries []fuseutil.Dirent, localEntries map[string]fuseutil.Dirent) (output []fuseutil.Dirent, err error) {
 	// Sanity check.
 	if !sort.IsSorted(sortedDirents(entries)) {
 		err = fmt.Errorf("expected sorted input")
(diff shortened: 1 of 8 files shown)


internal/fs/handle/dir_handle.go
```diff
@@ -185,7 +195,9 @@ func readAllEntries(
 	}
 
 	// Append local file entries (not synced to GCS).
-	entries = append(entries, localEntries...)
+	for _, localEntry := range localEntries {
```
raj-prince, line 198: Shouldn't we handle the duplicate entries here (while merging)? If we handle here, then there will be minimal code changes as we don't need to touch the fixConflictingNames() method also the purpose of fixConflictingNames() is different.

tools/integration_tests/local_file/read_dir_test.go
```diff
@@ -168,8 +167,8 @@ func TestReadDirWithSameNameLocalAndGCSFile(t *testing.T) {
 
 	// Attempt to list testDir.
 	_, err := os.ReadDir(testDirPath)
-	if err == nil || !strings.Contains(err.Error(), "input/output error") {
-		t.Fatalf("Expected error: %s, Got error: %v", "input/output error", err)
+	if err != nil {
```
raj-prince, line 170: We also need to change here - https://github.com/GoogleCloudPlatform/gcsfuse/blob/f6ad2829c5e1504c2414042a53daff82e6531fcb/tools/integration_tests/concurrent_operations/concurrent_listing_test.go#L244

raj-prince: We can remove this too - https://github.com/GoogleCloudPlatform/gcsfuse/blob/f6ad2829c5e1504c2414042a53daff82e6531fcb/tools/integration_tests/concurrent_operations/concurrent_listing_test.go#L364

As pre-creation of files don't need once this issue is fixed.

internal/fs/handle/dir_handle.go
vadlakondaswetha: FixConflictingNames logic  - checking if there are any duplicate entries and fixing the names.

handling duplicates requires the same logic as fixconflictingNames, sorting and comparing each entry. or other way would be to create a map of entire gcs response and check while adding local entries.  I would want to avoid another iteration of the GCS data which is huge.

raj-prince (commented):
LGTM.

tools/integration_tests/local_file/read_dir_test.go
vadlakondaswetha: Fixed.

raj-prince (approved):

internal/fs/handle/dir_handle.go
```diff
@@ -194,16 +206,9 @@ func readAllEntries(
...
-	// Option 1: [Selected]
-	// Throw an error while fixing conflicting names. The error will be fixed in
-	// subsequent ls calls assuming that entry will be removed from localFileInodes.
-	// Option 2: [Not Selected]
-	// Restrict fixConflictingNames to only GCS entries and show duplicate
-	// entries when ReadDir is called. In this case, a local file can have
-	// same name as directory and LookUpInode call will fetch directory details
-	// for both of them.
-	err = fixConflictingNames(entries)
+	// To handle this scenario, we are removing the duplicate entry before
```
ashmeenkaur, line 209: Nit: Shall we move this comment to fixConflictingNames() method

ashmeenkaur (approved):

internal/fs/handle/dir_handle.go
vadlakondaswetha: Its there at both places. Kept it on purpose - gives a reason about why the local entries are passed to fixConflictingNames.

PR #2250, opened 2024-07-31
Folder core creation poc

Tulsishah:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -373,12 +373,10 @@ func findExplicitFolder(ctx context.Context, bucket *gcsx.SyncerBucket, name Nam
 		return nil, fmt.Errorf("error in get folder for lookup : %w", folderErr)
 	}
 
-	folderObject := folderResult.ConvertFolderToMinObject()
-
 	return &Core{
 		Bucket:    bucket,
 		FullName:  name,
-		MinObject: folderObject,
+		Folder: folderResult,
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


internal/fs/inode/dir.go
```diff
@@ -678,10 +676,13 @@ func (d *dirInode) readObjects(
 		// the value of records["foo"].
 		if strings.HasSuffix(o.Name, "/") {
 			dirName := NewDirName(d.Name(), nameBase)
-			explicitDir := &Core{
-				Bucket:    d.Bucket(),
-				FullName:  dirName,
-				MinObject: storageutil.ConvertObjToMinObject(o),
+			var explicitDir *Core
+			if !d.isBucketHierarchical() {
```
ankitaluthra1, line 680: As discussed offline this can be merged with check on line no 679.

internal/fs/inode/dir.go
```diff
@@ -706,16 +707,25 @@ func (d *dirInode) readObjects(
 	for _, p := range listing.CollapsedRuns {
 		pathBase := path.Base(p)
 		dirName := NewDirName(d.Name(), pathBase)
-		if c, ok := cores[dirName]; ok && c.Type() == metadata.ExplicitDirType {
-			continue
-		}
+		if !d.isBucketHierarchical() {
```
ankitaluthra1, line 710: imo hns vs non hns check should be moved outside collapsed run, otherwise for every element in collpsed this is being checked

ankitaluthra1 (commented):
Overall approach lgtm.

PR #2258, opened 2024-08-01
[CLI Config Parity] Migrate write config to new config

ashmeenkaur:
### Description
This PR is part of migration to cobra/viper configs and migrates legacy config write config to new config. Some refactoring is done to remove circular dependency.

NA

### Testing details
1. Manual - Manually verified that write config is working.
2. Unit tests - Updated/Added.

### DIFF
--- a/cmd/testdata/valid_config.yaml
+++ b/cmd/testdata/valid_config.yaml
@@ -1,1 +1,3 @@
 app-name: hello
+write:
+  create-empty-file: true
--- a/tools/integration_tests/operations/operations_test.go
+++ b/tools/integration_tests/operations/operations_test.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/v2/cfg"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/config"
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/client"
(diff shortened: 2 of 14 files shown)


charith87 (approved):

tools/integration_tests/operations/operations_test.go
```diff
@@ -98,7 +99,7 @@ func createMountConfigsAndEquivalentFlags() (flags [][]string) {
 
 	// Set up config file with create-empty-file: true.
 	mountConfig1 := config.MountConfig{
-		WriteConfig: config.WriteConfig{
+		WriteConfig: cfg.WriteConfig{
```
kislaykishore, line 102: I'd rather we don't change the integration tests as much as possible. I'd like to run integration tests without changes after the migration is complete and ensure that the changes are backward compatible.

cmd/testdata/valid_config.yaml
```diff
@@ -1 +1,3 @@
 app-name: hello
+write:
+  create-empty-file: true
```
kislaykishore, line 3: Files should have a blank line at the end. Please add.

tools/integration_tests/operations/operations_test.go
ashmeenkaur: Makes sense but this is the easiest way to migrate the old config to new config. I'll raise a follow up PR for all the integration tests so that they don't rely on config object to create yaml file.

cmd/testdata/valid_config.yaml
ashmeenkaur: Will do this in the follow up PR with file cache changes.

kislaykishore (approved):

PR #2266, opened 2024-08-02
Reusing mount/unmount functions from ls script

anushka567:
### Description
Reusing mount/unmount functions from listing benchmark scripts.Moved the existing methods and unit tests under utils.
NA

### Testing details
1. Manual - Ran the listing benchmark script which makes calls to the refactored method.
2. Unit tests - [Tests](https://github.com/GoogleCloudPlatform/gcsfuse/blob/77f397107eafb3d73772880e38b154bea47c5779/perfmetrics/scripts/utils/mount_unmount_util_test.py#L19)
3. Integration tests - NA


Tulsishah (approved):
Approved as only refactoring happened in this PR.

ashmeenkaur (approved):

PR #2275, opened 2024-08-04
Adding troubleshooting guide for ls hang issue

raj-prince:
### Description
- Adding fix for `ls` hang issue in the troubleshooting doc.
- Preview: 
image

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


docs/troubleshooting.md
```diff
@@ -24,3 +24,4 @@ discusses potential solutions to the same.
...
...Experiencing hang while executing `ls` on a directory containing large number of files/directories.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   | Happens when ls triggers `stat` for each list entries other than listing. Fix (a) Use `ls --color=never` or `\ls` to remove the stat part of `ls`. Or (b) Increase the metadata cache ttl/capacity to serve the stat faster from cache (from the first execution as cache is populate as part of listing only). This is a generic solution and work for listing where stat is required. E.g. `ls -lah`                                                                                                                                                                                                                 |
```
vadlakondaswetha, line 27: do you want to add breaks to these options, so its rendered properly.

vadlakondaswetha, line 27: other than listing -> should it be inaddition to listing?

vadlakondaswetha, line 27: be explicit, that stat will be served from cache if ls is made on the same directory multiple times.

charith87, line 27: Can we start the sentence like this - "The default "ls" unix command behaviour does a listing and additionally, does a stat for each file entry which slows down the entire listing operation and may look like hang, but in reality is iterating one by one and slow. " 
For the next sentence, i would not call it a fix, since its not a bug and intended behaviour from the OS/Kernel. Instead to tell to ignore the stat calls or guidance is to the other command.

raj-prince: Added a note -`ls` will be faster from the first execution.

docs/troubleshooting.md
```diff
@@ -24,3 +24,4 @@ discusses potential solutions to the same.
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | By default `ls` does listing but sometimes additionaly does `stat` for each list entry. Additional stat on each entry slows down the entire `ls` operation and may look like hang, but in reality is iterating one by one and slow. There are two ways to overcome this: <ul><li>**Get rid of `stat`:** Execute without coloring `ls --color=never` or `\ls` to remove the stat part of `ls`.</li> <li>**Improve the `stat` lookup time:** increase the metadata cache ttl/capacity which is populated as part of listing and makes the `stat` lookup faster. Important to note here, `ls` will be faster from the first execution itself as cache is populated in listing phase, leading to quicker stat lookup for each entry.</li></ul>    |
```
charith87, line 27: Pls add a link to the metadata cache configuring documentation section

charith87 (dismissed):
Overall seems fine. Approving. go/lgtm-with-comments

charith87 (approved):

vadlakondaswetha (approved):

PR #2277, opened 2024-08-05
Rename fs level integration 2

Tulsishah:
### Description
- The renameFolder method should now utilize atomic rename operations.
- Implement a conditional check: If the bucket is hierarchical, then proceed to call the renameFolder method.
- The typed cache addition for new directories has been removed. This is because we cannot add a new directory type cache from the Inode of the old directory.

**Test on Rename when an empty destination directory exists -** `os.Rename` fails even if the destination directory (dstDirectory) is empty. It does not even test our rename functionality it gets fails in LookUp only, kernel does not make rename call. 
      
- This behavior is different from the 'mv' command, which works even with an empty destination directory. In the `mv` command, `mv A B` results in `B/A`, which effectively creates a scenario where the destination directory doesn't exist initially. That's why we haven't write unit/composite tests for the same. Although we have this test in our e2e tests so it will run as a part of e2e tests.

NA

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Automated(RenameFolder tests for HNS will be enabled in next PR)


internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+	defer fs.releaseInodes(&pendingInodes)
+
+	oldDir, err := fs.getBucketDirInode(ctx, oldParent, oldName)
+	if err != nil {
+		return err
+	}
+	pendingInodes = append(pendingInodes, oldDir)
+
+	if err = fs.ensureNoOpenFilesInDirectory(oldDir, oldName); err != nil {
+		return err
```
gargnitingoogle, line 2081: This implies that if the directory being moved/renamed has any of the files inside it open at the time, then the move/rename returns error. This is I think an acceptable limitation as it exists in most file systems. Just wanted to point out. But we probably should document it as it is a known limitation.

gargnitingoogle, line 2081: The error should contain the error string to indicate why it failed, and not just the base error.

```suggestion
		return fmt.Errorf("the directory being renamed/moved has open files: %w", err)
```

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+	}
+
+	newDir, err := fs.getBucketDirInode(ctx, newParent, newName)
+	if err == nil {
+		// If the directory exists, then check if it is empty or not.
+		if err = fs.checkDirNotEmpty(newDir, newName); err != nil {
+			return err
```
gargnitingoogle, line 2088: same as above

```suggestion
			return fmt.Errorf("the destination directory already exists and is not empty : %w", err)
```

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+		return err
+	}
+	pendingInodes = append(pendingInodes, oldDir)
+
+	if err = fs.ensureNoOpenFilesInDirectory(oldDir, oldName); err != nil {
+		return err
+	}
+
+	newDir, err := fs.getBucketDirInode(ctx, newParent, newName)
+	if err == nil {
```
gargnitingoogle, line 2085: What if the err is not nil? Does that indicate that everything is right for proceeding?

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+		defer oldParent.Unlock()
+	} else {
+		oldParent.Lock()
+		defer oldParent.Unlock()
+
+		newParent.Lock()
+		defer newParent.Unlock()
+	}
+
+	// Rename old directory to the new directory, keeping both parent directory locked.
```
gargnitingoogle, line 2107:
```suggestion
	// Rename old directory to the new directory, keeping both parent directories locked.
```

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+// exists and is non-empty, return ENOTEMPTY.
+//
+// LOCKS_EXCLUDED(fs.mu)
+// LOCKS_EXCLUDED(oldParent)
+// LOCKS_EXCLUDED(newParent)
+func (fs *fileSystem) renameFolder(ctx context.Context, oldParent inode.DirInode, oldName string, newParent inode.DirInode, newName string) (err error) {
+	var pendingInodes []inode.DirInode
+	defer fs.releaseInodes(&pendingInodes)
+
+	oldDir, err := fs.getBucketDirInode(ctx, oldParent, oldName)
```
gargnitingoogle, line 2074: It is difficult to understand what the difference between oldParent and oldDir is. The same goes for the difference between newParent and newDir. Could you add a line or two of comments to explain this please? Otherwise, it's confusing for the reader.

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+		// If the directory exists, then check if it is empty or not.
+		if err = fs.checkDirNotEmpty(newDir, newName); err != nil {
+			return err
+		}
+		pendingInodes = append(pendingInodes, newDir)
+	}
+
+	oldDirName := inode.NewDirName(oldParent.Name(), oldName)
+	newDirName := inode.NewDirName(newParent.Name(), newName)
+	if oldParent == newParent {
```
gargnitingoogle, line 2095: nit: A shorter alternative:

```suggestion
    oldParent.Lock()
    defer oldParent.Unlock()
    
	if newParent != oldParent {
	     newParent.Lock()
      	 defer newParent.Unlock()
	}
```

gargnitingoogle, line 2095: It is also not clear what oldParent and newParent being same means and what is needed to be done in such a case. Please clarify.

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+		defer newParent.Unlock()
+	}
+
+	// Rename old directory to the new directory, keeping both parent directory locked.
+	_, err = oldParent.RenameFolder(ctx, oldDirName.GcsObjectName(), newDirName.GcsObjectName())
+	if err != nil {
+		return err
+	}
+
+	fs.releaseInodes(&pendingInodes)
```
gargnitingoogle, line 2113: This is already there in the beginning of this function as a deferred statement. I am guessing you want to keep that one, and want to remove this one.

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+	fs.checkInvariantsForImplicitDirs()
```
gargnitingoogle, line 2114: I am guessing that this should also be a deferred statement.

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
 	return nil
 }
 
+// Rename an old folder to a new folder in hierarchical bucket. If the new folder already
+// exists and is non-empty, return ENOTEMPTY.
```
gargnitingoogle, line 2065: What specific error (if any) do we expect if the old directory does not exist, or has a file open?

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
 	assert.NotNil(t.T(), err)
 }
+
+func (t *HNSBucketTests) TestRenameFolderWithSrcDirectoryDoesNotExist() {
+	oldDirPath := path.Join(mntDir, "foo_not_exist")
+	newDirPath := path.Join(mntDir, "foo_rename")
+
+	err = os.Rename(oldDirPath, newDirPath)
+
+	assert.NotNil(t.T(), err)
```
gargnitingoogle, line 110: We should also expect a specific error or a specific error string to make sure about the cause of the error.

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+	_, err = os.Stat(newDirPath)
+	assert.NotNil(t.T(), err)
```
gargnitingoogle, line 112: We should also expect a specific error or a specific error string to make sure about the cause of the error.

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+}
+
+func (t *HNSBucketTests) TestRenameFolderWithDstDirectoryIsNotEmpty() {
```
gargnitingoogle, line 115:
```suggestion
func (t *HNSBucketTests) TestRenameFolderWithDstDirectoryNotEmpty() {
```

or 

```suggestion
func (t *HNSBucketTests) TestRenameFolderWhenDstDirectoryIsNotEmpty() {
```

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+	oldDirPath := path.Join(mntDir, "foo")
+	_, err = os.Stat(oldDirPath)
+	assert.NoError(t.T(), err)
+	newDirPath := path.Join(mntDir, "bar")
+	_, err = os.Stat(newDirPath)
+	assert.NoError(t.T(), err)
+
+	err = os.Rename(oldDirPath, newDirPath)
```
gargnitingoogle, line 123: We haven't really put any file in the dst directory, so either the test name is wrong, or you missed to add a file in the dst directory.

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+
+	assert.NotNil(t.T(), err)
```
gargnitingoogle, line 125: nit: following is more suited.

```suggestion
	assert.Error(t.T(), err)
```

internal/fs/fs.go
```diff
@@ -2058,6 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+		defer oldParent.Unlock()
+
+		newParent.Lock()
+		defer newParent.Unlock()
+	}
+
+	// Rename old directory to the new directory, keeping both parent directory locked.
+	_, err = oldParent.RenameFolder(ctx, oldDirName.GcsObjectName(), newDirName.GcsObjectName())
+	if err != nil {
+		return err
```
gargnitingoogle, line 2110: Same as above, 

```suggestion
		return fmt.Errorf("failed to rename folder: %w", err)
```

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+	oldDirPath := path.Join(mntDir, "foo")
+	_, err = os.Stat(oldDirPath)
+	assert.NoError(t.T(), err)
+	newDirPath := path.Join(mntDir, "foo_rename")
+
+	err = os.Rename(oldDirPath, newDirPath)
+
+	assert.NoError(t.T(), err)
+	_, err = os.Stat(oldDirPath)
+	assert.NotNil(t.T(), err)
```
gargnitingoogle, line 138: nit:

```suggestion
	assert.Error(t.T(), err)
```

gargnitingoogle: Also check for specific error code/string

gargnitingoogle (changes requested):
still going thru test code

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+			assert.False(t.T(), dirEntries[i].IsDir())
+		case "file2.txt":
+			assert.Equal(t.T(), "file2.txt", dirEntries[i].Name())
+			assert.False(t.T(), dirEntries[i].IsDir())
+		case "implicit_dir":
+			assert.Equal(t.T(), "implicit_dir", dirEntries[i].Name())
+			assert.True(t.T(), dirEntries[i].IsDir())
+		}
+	}
+}
```
gargnitingoogle, line 200: We also need tests for the following scenarios I think (please correct me if we have those have covered already).

1. When source directory has files opened. Here, rename should fail.
2. When destination directory exists and is emoty. Here rename should pass.
3. When destination directory path is a parent/ancestor of the given source directory path. And the reverse. Both these cases, rename should fail.
4. When both source and destination directory are the same. Here rename should pass without doing anything.

gargnitingoogle (changes requested):
finished reviewing the complete PR

internal/fs/fs.go
Tulsishah: We are already returning respective error from the function. I think we should not repeat the things.

internal/fs/fs.go
Tulsishah: We are already returning respective error from the function. I think we should not repeat the things.

internal/fs/fs.go
Tulsishah: If err is not nil means we don't get Inode, that refers directory does not exist. So we can go ahead with renaming

internal/fs/hns_bucket_test.go
Tulsishah: The test is

A/a.txt
B/b.txt

We are renaming A to B and it should throw an issue.

here we are renaming `foo` to `bar`.  and  `bar` is not empty 

> We haven't really put any file in the dst directory, so either the test name is wrong, or you missed to add a file in the dst directory.

We have done it in SetUp.

internal/fs/inode/dir.go
```diff
@@ -1000,8 +1000,6 @@ func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinat
 	// TODO: Cache updates won't be necessary once type cache usage is removed from HNS.
 	// Remove old entry from type cache.
 	d.cache.Erase(folderName)
-	// Add new renamed folder in type cache.
```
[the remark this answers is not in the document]
Tulsishah: d refers to oldDirInode, we can't add type cache of newDir if newDir parent is not same as oldDir.

internal/fs/fs.go
Tulsishah: Yeah.. In existing renameDir function also we do the same.

internal/fs/fs.go
```diff
@@ -1972,6 +1972,9 @@ func (fs *fileSystem) Rename(
 	}
 
 	if child.FullName.IsDir() {
+		if child.Bucket.BucketType() == gcs.Hierarchical {
```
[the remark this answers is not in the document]
Tulsishah: I don't think we should add suffixes like ManagedFolder or HNSFolder here because those terms aren't relevant at the file system level. renameFolder makes more sense in this context. I considered merging renameDir and renameFolder, but it was getting too complicated. So, I believe renameFolder is the right choice.

internal/fs/fs.go
Tulsishah: Actually I thought we can remove this `fs.releaseInodes(&pendingInodes)` as we already doing it in defer above, but wanted to make sure with @vadlakondaswetha about importance of putting it twice.

internal/fs/fs.go
Tulsishah: The "old parent" refers to the parent directory of the original directory, while the "new parent" refers to the parent directory of the renamed directory. We are renaming "oldDir" to "newDir". 

If both directories share the same parent directory, we only need to acquire the lock once. This information was also mentioned in the comments.

internal/fs/fs.go
Tulsishah: We are returning manual errors and not fuse errors directly on those cases. renameDir() function also mentioned the same info.

internal/fs/hns_bucket_test.go
```diff
@@ -100,3 +100,101 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+	assert.NotNil(t.T(), err)
+	_, err = os.Stat(newDirPath)
+	assert.NoError(t.T(), err)
+	dirEntries, err := os.ReadDir(newDirPath)
+	assert.NoError(t.T(), err)
+	assert.Equal(t.T(), 5, len(dirEntries))
+	for i := 0; i < len(dirEntries); i++ {
+		switch dirEntries[i].Name() {
+		case "test":
+			assert.Equal(t.T(), "test", dirEntries[i].Name())
```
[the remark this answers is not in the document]
Tulsishah: I found in other scenarios order does not remains same every time, that's why I preferred switch/case approach.

internal/fs/hns_bucket_test.go
Tulsishah: 2. I thought to write, but `os.Rename` fails even if on empty dstDirectory. It works with `mv` command only
3 & 4 Is not even possible from file system, it will not even come to rename layer. Also 4 will be similar to 2 only sort of to as os.Rename does not support if dst directory even exist.

Tulsishah: For 1 I have added the test. Thanks.

gargnitingoogle (dismissed):
LGTM

internal/fs/fs.go
```diff
@@ -2058,7 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+	var pendingInodes []inode.DirInode
+	defer fs.releaseInodes(&pendingInodes)
+
+	oldDirInode, err := fs.getBucketDirInode(ctx, oldParent, oldName)
+	if err != nil {
+		return err
+	}
+	pendingInodes = append(pendingInodes, oldDirInode)
+
+	if err = fs.ensureNoOpenFilesInDirectory(oldDirInode, oldName); err != nil {
```
vadlakondaswetha, line 2084: dont call this method ensureNoOpenFiles. Rename it to ensureNoLocalFilesInDirectory. OpenFiles and localfiles are different. We can have a gcs file which is open as well. you are only checking for local files, so mention that in the method name.

internal/fs/hns_bucket_test.go
```diff
@@ -98,5 +101,105 @@ func (t *HNSBucketTests) TestDeleteFolder() {
...
+	_, err = os.Stat(newDirPath)
+	assert.NoError(t.T(), err)
+
+	err = os.Rename(oldDirPath, newDirPath)
+
+	assert.Error(t.T(), err)
+	assert.True(t.T(), strings.Contains(err.Error(), "file exists"))
+}
+
+func (t *HNSBucketTests) TestRenameFolderWithSourceDirectoryHaveOpenFiles() {
```
vadlakondaswetha, line 135: haveLocalFiles

internal/fs/hns_bucket_test.go
```diff
@@ -98,5 +101,105 @@ func (t *HNSBucketTests) TestDeleteFolder() {
 
 	assert.NoError(t.T(), err)
 	_, err = os.Stat(dirPath)
-	assert.NotNil(t.T(), err)
+	assert.Error(t.T(), err)
+	assert.True(t.T(), strings.Contains(err.Error(), "no such file or directory"))
+}
+
+func (t *HNSBucketTests) TestRenameFolderWithSrcDirectoryDoesNotExist() {
```
vadlakondaswetha, line 108: What about Test for DestDirectoryNotExist?

Tulsishah: DestDirectoryNotExist --> This test will be successful.

It's covered in TestRenameFolderWithSameParent() and TestRenameFolderWithDifferentParents(), where the destination folder does not exist.

internal/fs/fs.go
Tulsishah: Makes sense. Thanks. Done

internal/fs/fs.go
```diff
@@ -2058,7 +2061,60 @@ func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name strin
...
+	oldParent.Lock()
+	defer oldParent.Unlock()
+
+	if newParent != oldParent {
+		newParent.Lock()
+		defer newParent.Unlock()
+	}
+
+	// Rename old directory to the new directory, keeping both parent directories locked.
+	_, err = oldParent.RenameFolder(ctx, oldDirName.GcsObjectName(), newDirName.GcsObjectName())
```
ankitaluthra1, line 2109: please add a comment renameDirLimit is not used in rename folder as there is no user defined limit on new rename

ankitaluthra1 (changes requested):

internal/fs/fs.go
ankitaluthra1: just to confirm here, if we have gcs files open during rename, thats not a problem in rename right? do we have any integration test case for both open files and local files scenarios ? I think having those would be useful.

internal/fs/fs.go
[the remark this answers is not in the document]
ankitaluthra1: @vadlakondaswetha would appreciate your opinion here.

internal/fs/hns_bucket_test.go
```diff
@@ -31,6 +32,19 @@ type HNSBucketTests struct {
 	fsTest
 }
 
+type dirEntry struct {
```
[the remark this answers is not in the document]
Tulsishah: We have a test with the following structure:
```
foo/
foo/file1.txt
foo/file2.txt
foo/test2
foo/test
foo/test/file3.txt
foo/implicit_dir/file3.txt
```
We will rename "foo" to "foo_rename" and then perform the following checks:

Verify that "foo" no longer exists.
Confirm that "foo_rename" exists.
Ensure the structure of "foo_rename" is identical to the original "foo."
I believe this covers most scenarios comprehensively. Please check the TestRenameFolderWithSameParent() and TestRenameFolderWithDifferentParents() tests.

Additionally, creating a new directory and validating its emptiness requires [this pull request](https://github.com/GoogleCloudPlatform/gcsfuse/pull/2346) to be merged. I will add those checks to the same tests once the pull request is submitted.

internal/fs/fs.go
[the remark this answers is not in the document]
Tulsishah: Done

internal/fs/fs.go
Tulsishah: Done

internal/fs/hns_bucket_test.go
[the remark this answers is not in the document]
ankitaluthra1: these are composite tests, lets add an IT as well for this case but not blocking this PR on it as in subsequent PR we are having other ITs as well, please make sure to have this case there.

internal/fs/fs.go
Tulsishah: > just to confirm here, if we have gcs files open during rename, thats not a problem in rename right?

Yes. Added unit test. will add e2e test in subsequent PR. 

> do we have any integration test case for both open files and local files scenarios ? I think having those would be useful.

With local files yes. with open GCS file no. will add.

internal/fs/fs.go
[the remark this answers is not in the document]
vadlakondaswetha: Agreed. I also got confused with renameDir and renameFolder. Irrespective of filesystem conventions, it was confusing. I suggest renaming it to renameHierarchialDir and renameNonHierarchialDir/renameFlatDir

internal/fs/hns_bucket_test.go
vadlakondaswetha: can you add a comment about this.

internal/fs/hns_bucket_test.go
vadlakondaswetha: What about test case with dst directory present and empty.

Also what about testcase with src directory empty?

Tulsishah: > What about test case with dst directory present and empty.

Added in description also and you can check this https://github.com/GoogleCloudPlatform/gcsfuse/pull/2277/files#r1710876346

> Also what about testcase with src directory empty?

Added

vadlakondaswetha (approved):

internal/fs/hns_bucket_test.go
Tulsishah: Generally it recommend to avoid comparing error strings.
https://bitfieldconsulting.com/posts/comparing-errors#:~:text=Comparing%20the%20string%20form%20of,should%20try%20to%20avoid%20it.&text=This%20is%20Part%202%20of,Comparing%20Go%20error%20values

ankitaluthra1 (approved):

Issue #2273, opened 2024-08-03
Size reported is wrong for symlinks


nbdd0121:
**Describe the issue**

[Man page of stat](https://linux.die.net/man/2/stat) states:
> The st_size field gives the size of the file (if it is a regular file or a symbolic link) in bytes. The size of a symbolic link is the length of the pathname it contains, without a terminating null byte. 

But gcsfuse reports symlink as having size 0.

**Reproduction**

Mount gcsfuse, create a symlink within. Then call `stat symlink`, it reports size of 0.

**System (please complete the following information):**
 - OS: NixOS 24.05
 - Platform: Bare-metal
 - Version: master branch

**Additional context**

**SLO:**
We strive to respond to all bug reports within 24 business hours.

vyncint:
Took a look at this since it was still open. The report is accurate, and the root cause is a single unset return value.

`SymlinkInode.Attributes` in `internal/fs/inode/symlink.go` never assigns `size`, so it returns the zero value:

```go
func (s *SymlinkInode) Attributes(
	ctx context.Context, clobberedCheck bool) (size uint64, mtime time.Time, nlink uint32, err error) {
	mtime = s.mtime
	nlink = 1
	return
}
```

`fs.getAttributes` is the only caller and assigns that straight to `attr.Size`, so `stat` reports 0.

Two things suggest this is an oversight rather than a deliberate choice. First, `stat(2)` is explicit that `st_size` for a symlink is the length of the pathname it contains. Second, the intended semantics are already written down a few lines above, in `UpdateSize`:

```go
// The size of a symlink is its target's length, not the backing object's size.
```

So the contract is stated in the file; `Attributes` just doesn't implement it. For a legacy symlink the backing object is empty, which is why the reported size is 0 rather than something merely wrong.

The fix is `size = uint64(len(s.target))`. I checked the blast radius: `Attributes` has exactly one caller, and the rename path uses `Source()` (the backing object's size) rather than `Attributes`, so it is unaffected.

One thing worth a maintainer decision before this goes anywhere, and the reason I am commenting rather than just sending a patch: the current behaviour is pinned by four existing assertions, so fixing it is a deliberate behaviour change, not just a gap being filled.

- `internal/fs/inode/symlink_test.go` - `TestAttributes` asserts `uint64(0)`
- `internal/fs/local_modifications_test.go` - `SymlinkTest.CreateLink` asserts `0` twice, once via `Lstat` and once via the readdir entry
[...]

PR #2285, opened 2024-08-05
Fix size of symlinks and enable kernel symlink cache

nbdd0121:
### Description
Symlink inodes are expected to have size equal to the length of its target. This is defined in POSIX spec.

FUSE's symlink cache also expects this, so fixing this allows the enabling of the symlink cache. Note that since symlink targets are not mutable, the cache is always valid while the inode is valid, so there is no need for any TTL or invalidation logic.

This is expected to bring significant time saving for workloads that make use of symlinks (especially for a directory symlink, since all ops inside the symlink directory would require reading the symlink at least once).

#2273 

### Testing details
1. Manual:
    - Mount a bucket on two mount points (with separate GCSFuse processes)
    - Create a symlink in mount1
    - Stat and see its size is correct and point to the correct target in mount2.
    - Remove and recreate the symlink in mount1 to a different target
    - Stat and see its size is correct and point to the correct target in mount2.
2. Unit tests - NA
3. Integration tests - I didn't add a test because I'm unfamiliar with this project (and go in general)'s test setup and it seems that the test infra is being reworked on.

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -174,6 +174,9 @@ func getFuseMountConfig(fsName string, newConfig *cfg.Config, mountConfig *confi
 		// access two files under same directory parallely, then the lookups also
 		// happen parallely.
 		EnableParallelDirOps: !(mountConfig.FileSystemConfig.DisableParallelDirops),
+		// Symlink target is not mutable (removing and re-creating would cause a different
+		// inode to be created), so it's safe to enable symlink caching.
+		EnableSymlinkCaching: true,
 	}
 
 	mountCfg.ErrorLogger = logger.NewLegacyLogger(logger.LevelError, "fuse: ")
--- a/internal/fs/inode/symlink.go
+++ b/internal/fs/inode/symlink.go
@@ -84,6 +84,10 @@ func NewSymlinkInode(
 			Atime: m.Updated,
 			Ctime: m.Updated,
 			Mtime: m.Updated,
+			// POSIX spec requires that for symbolic links, size attribute
+			// reflects the length in bytes of the pathname contained in the
+			// symbolic link. This is also expected by symlink cache in FUSE.
+			Size: uint64(len(m.Metadata[SymlinkMetadataKey])),
 		},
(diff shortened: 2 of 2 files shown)


nbdd0121:
This improves perf by 10% in our workload on top of #2269.

raj-prince:
Thanks @nbdd0121 for the contribution!!

Changes looks good to me, but requires a couple of changes to make it mergeable.

(a) We need to guard the symlink caching under a flag like, --experimental-enable-symlink-cache (by default disabled). Keeping this experimental as we haven't performed all the testing.
(b) Could you please add unit test corresponding to size changes - [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/symlink_test.go)?
(c) If possible, could you please write the simple composite test for both true and false value of `--experimental-enable-symlink-cache`?

Note: Adding an experimental flag would be easy if you do this after 1 week. We are making the changes to simplify those path.

raj-prince:
@nbdd0121 Here, is the steps to add a new flag - https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/dev_guide.md#adding-a-new-param-in-gcsfuse

nbdd0121:
Tried to write a test but adding the test to symlink_test.go doesn't seem trivial to me. The test that exists there don't use `NewSymlinkInode`. I tried to just use `inode.NewSymLinkInode` but then I can't call `inode.Attribute` because I don't have a ctx?

(My experience is mostly with C and Rust and I have no experience with Go at all.)

ashmeenkaur:
Hi @nbdd0121,

Sorry for the delay here! We were busy putting together a formal development guide, and it's finally ready: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/dev_guide.md#dev-guide

Please give it a look and follow the instructions – looking forward to get these changes merged.

Thanks!

nbdd0121:
Unit test for size change is now added.

nbdd0121:
@raj-prince  can you take another look?

PR #2287, opened 2024-08-06
Made the parallel downloads go routines running without synchronization

vadlakondaswetha:
### Description
Incase of parallel downloads, we trigger n number of goroutines each downloading xMB and then we wait for x*nMB to be downloaded and then make the goroutines fetch another x*nMB.
Instead of that, we are making each goroutine continuously download xMB and then next xMB instead of worrying about other goroutines. Which offset to fetch will be passed via channel,.

NA

### Testing details
1. Manual - Done
2. Unit tests - Existing tests are sufficient
3. Integration tests - Fixed as required.

### DIFF
--- a/internal/cache/data/object_range.go
+++ b/internal/cache/data/object_range.go
@@ -1,0 +1,6 @@
+package data
+
+type ObjectRange struct {
+	Start uint64
+	End   uint64
+}
--- a/internal/cache/data/object_range.go
+++ b/internal/cache/data/object_range.go
@@ -1,6 +1,6 @@
 package data
 
 type ObjectRange struct {
-	Start uint64
-	End   uint64
+	Start int64
+	End   int64
 }
--- a/internal/cache/data/object_range.go
+++ b/internal/cache/data/object_range.go
@@ -1,5 +1,6 @@
 package data
 
+// ObjectRange specifies the range within the gcs object.
 type ObjectRange struct {
 	Start int64
 	End   int64
--- a/internal/cache/file/downloader/jm_parallel_downloads_test.go
+++ b/internal/cache/file/downloader/jm_parallel_downloads_test.go
@@ -100,7 +100,7 @@ func TestParallelDownloads(t *testing.T) {
 		subscribedOffset         int64
 	}{
 		{
-			name:                     "download in chunks of concurrency * readReqSize",
+			name:                     "download the entire object when object size > no of goroutines * readReqSize",
 			objectSize:               15 * util.MiB,
 			readReqSize:              3,
 			parallelDownloadsPerFile: 100,
--- a/internal/cache/file/downloader/jm_parallel_downloads_test.go
+++ b/internal/cache/file/downloader/jm_parallel_downloads_test.go
@@ -144,7 +144,7 @@ func TestParallelDownloads(t *testing.T) {
 				select {
 				case jobStatus := <-subscriberC:
 					if assert.Nil(t, err) {
-						assert.LessOrEqual(t, tc.expectedOffset, jobStatus.Offset)
+						assert.Equal(t, tc.expectedOffset, jobStatus.Offset)
 						verifyFileTillOffset(t,
 							data.FileSpec{Path: util.GetDownloadPath(path.Join(cacheDir, storage.TestBucketName), "path/in/gcs/foo.txt"), FilePerm: util.DefaultFilePerm, DirPerm: util.DefaultDirPerm}, tc.expectedOffset,
 							content)
--- a/internal/cache/file/downloader/jm_parallel_downloads_test.go
+++ b/internal/cache/file/downloader/jm_parallel_downloads_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func createObjectInBucket(t *testing.T, objPath string, objSize int64, bucket gcs.Bucket) []byte {
@@ -138,7 +139,7 @@ func TestParallelDownloads(t *testing.T) {
 				select {
 				case jobStatus := <-subscriberC:
 					if assert.Nil(t, err) {
-						assert.GreaterOrEqual(t, tc.objectSize, jobStatus.Offset)
+						require.GreaterOrEqual(t, tc.objectSize, jobStatus.Offset)
 						verifyFileTillOffset(t,
 							data.FileSpec{Path: util.GetDownloadPath(path.Join(cacheDir, storage.TestBucketName), "path/in/gcs/foo.txt"), FilePerm: util.DefaultFilePerm, DirPerm: util.DefaultDirPerm}, jobStatus.Offset,
 							content)
--- a/internal/cache/file/downloader/jm_parallel_downloads_test.go
+++ b/internal/cache/file/downloader/jm_parallel_downloads_test.go
@@ -97,7 +97,6 @@ func TestParallelDownloads(t *testing.T) {
 		parallelDownloadsPerFile int64
 		maxParallelDownloads     int64
 		downloadOffset           int64
-		expectedOffset           int64
 		subscribedOffset         int64
 	}{
 		{
@@ -108,10 +107,6 @@ func TestParallelDownloads(t *testing.T) {
 			maxParallelDownloads:     3,
 			subscribedOffset:         7,
 			downloadOffset:           10,
-			// Concurrency can go to (maxParallelDownloads + 1) in case
-			// parallelDownloadsPerFile > maxParallelDownloads because we always
-			// spawn a minimum of 1 go routine per async job.
-			expectedOffset: 12 * util.MiB,
 		},
 		{
 			name:                     "download only upto the object size",
@@ -121,7 +116,6 @@ func TestParallelDownloads(t *testing.T) {
 			maxParallelDownloads:     3,
 			subscribedOffset:         7,
 			downloadOffset:           10,
-			expectedOffset:           10 * util.MiB,
 		},
 	}
 	for _, tc := range tbl {
@@ -144,9 +138,9 @@ func TestParallelDownloads(t *testing.T) {
 				select {
 				case jobStatus := <-subscriberC:
 					if assert.Nil(t, err) {
-						assert.Equal(t, tc.expectedOffset, jobStatus.Offset)
+						assert.GreaterOrEqual(t, tc.objectSize, jobStatus.Offset)
 						verifyFileTillOffset(t,
-							data.FileSpec{Path: util.GetDownloadPath(path.Join(cacheDir, storage.TestBucketName), "path/in/gcs/foo.txt"), FilePerm: util.DefaultFilePerm, DirPerm: util.DefaultDirPerm}, tc.expectedOffset,
+							data.FileSpec{Path: util.GetDownloadPath(path.Join(cacheDir, storage.TestBucketName), "path/in/gcs/foo.txt"), FilePerm: util.DefaultFilePerm, DirPerm: util.DefaultDirPerm}, jobStatus.Offset,
 							content)
 					}
 					return
--- a/internal/cache/file/downloader/jm_parallel_downloads_test.go
+++ b/internal/cache/file/downloader/jm_parallel_downloads_test.go
@@ -17,7 +17,6 @@ package downloader
 import (
 	"context"
 	"crypto/rand"
-	"math"
 	"os"
 	"path"
 	"testing"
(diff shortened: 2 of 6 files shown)


internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -58,61 +59,148 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
...
+// return the appropriate error.
+func (job *Job) handleJobCompletion(ctx context.Context, offset uint64, group *errgroup.Group) error {
+	// Close the channel since we are ending the job.
+	close(job.offsetChan)
+
+	// First check if the context has reported any error.
+	err := ctx.Err()
+	if err != nil {
+		// Also wait for all the goroutines to finish to ensure job is stopped.
+		waitErr := group.Wait()
```
sethiay, line 195: why can't we simply directly wait for `err = group.Wait` and then return that err instead of checking the ctx.Err separately ?

tools/integration_tests/read_cache/job_chunk_test.go
```diff
@@ -83,7 +81,7 @@ func createConfigFileForJobChunkTest(cacheSize int64, cacheFileForRangeRead bool
 
 func (s *jobChunkTest) TestJobChunkSizeForSingleFileReads(t *testing.T) {
 	var fileSize int64 = 24 * util.MiB
-	chunkCount := math.Ceil(float64(fileSize) / float64(s.chunkSize))
+	//chunkCount := math.Ceil(float64(fileSize) / float64(s.chunkSize))
```
sethiay, line 84: nit: remove this line

tools/integration_tests/read_cache/job_chunk_test.go
```diff
@@ -92,16 +90,21 @@ func (s *jobChunkTest) TestJobChunkSizeForSingleFileReads(t *testing.T) {
...
-	for i := 0; int64(i) < int64(chunkCount); i++ {
-		offset := min(s.chunkSize*int64(i+1), fileSize)
-		assert.Equal(t, offset, structuredJobLogs[0].JobEntries[i].Offset)
+
+	// We need to check that downloadedOffset is always greater than the previous downloadedOffset
+	// and is in multiples of chunkSize.
+	for i := 1; i < len(structuredJobLogs[0].JobEntries); i++ {
+		offsetDiff := structuredJobLogs[0].JobEntries[i].Offset - structuredJobLogs[0].JobEntries[i-1].Offset
+		assert.Greater(t, offsetDiff, int64(0))
+		assert.Equal(t, int64(0), offsetDiff%s.chunkSize)
```
sethiay, line 99: L99 will not be true if file size is not multiple of chunkSize, either we can handle it or just add a comment.

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -58,61 +59,148 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
...
 
-	_, err = io.CopyN(dstWriter, newReader, end-start)
+	_, err = io.CopyN(dstWriter, newReader, int64(end-start))
 	if err != nil {
 		err = fmt.Errorf("downloadRange: error at the time of copying content to cache file %w", err)
 	}
 	return err
 }
 
+func (job *Job) updateRangeMap(rangeMap map[uint64]uint64, offsetStart uint64, offsetEnd uint64) error {
```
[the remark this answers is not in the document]
sethiay: I think we should also add unit tests for this func ?

internal/cache/file/downloader/jm_parallel_downloads_test.go
```diff
@@ -98,31 +97,25 @@ func TestParallelDownloads(t *testing.T) {
...
 		downloadOffset           int64
-		expectedOffset           int64
 		subscribedOffset         int64
 	}{
 		{
 			name:                     "download in chunks of concurrency * readReqSize",
 			objectSize:               15 * util.MiB,
 			readReqSize:              3,
-			parallelDownloadsPerFile: math.MaxInt,
+			parallelDownloadsPerFile: 100,
```
sethiay, line 106: nit: change the name "download in chunks of concurrency * readReqSize" of this unit test?

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -58,61 +59,148 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
...
-			downloadErrGroup.Go(func() error {
-				if currGoRoutineIdx > 0 {
-					defer job.maxParallelismSem.Release(1)
-				}
-				// Copy the contents from NewReader to cache file at appropriate offset.
-				offsetWriter := io.NewOffsetWriter(cacheFile, rangeStart)
-				return job.downloadRange(downloadErrGroupCtx, offsetWriter, rangeStart, rangeEnd)
-			})
+	rangeMap := make(map[uint64]uint64)
+	job.offsetChan = make(chan offset, 2*job.fileCacheConfig.ParallelDownloadsPerFile)
```
sethiay, line 143: nit: add a comment why the length of chan is 2X of ParallelDownloadsPerFile

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -32,15 +33,15 @@ import (
 // GCS into given destination writer.
 //
 // This function doesn't take locks and can be executed parallely.
-func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, end int64) error {
+func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, end uint64) error {
```
kislaykishore, line 36: Use unsigned ints sparingly: https://yaqs.corp.google.com/eng/q/8035320036228857856

tools/integration_tests/read_cache/job_chunk_test.go
vadlakondaswetha: added comment.

internal/cache/file/downloader/parallel_downloads_job.go
vadlakondaswetha: updated the comment. In a scenario where context is cancelled before any goroutines are started or waiting for input. group.Wait doesn't return any error.

Ideally group.Wait is not required, but added to ensure we dont left any goroutines running.

internal/cache/file/downloader/parallel_downloads_job.go
vadlakondaswetha: minObject.Size is uint64. Using uint64 to keep it consistent with that.

kislaykishore: It seems like a step in the reverse direction - the size in the MinObj is uint64 which is assigned from ObjectAttrs.Size, an int64 here.
For this PR, I think we should use continue using int64 and add a TODO to change MinObj.Size to int64 later.

internal/cache/data/object_range.go
```diff
@@ -0,0 +1,6 @@
+package data
+
+type ObjectRange struct {
+	Start uint64
```
kislaykishore, line 4: Please use int64.

internal/cache/file/downloader/job.go
```diff
@@ -91,6 +91,10 @@ type Job struct {
 	// This semaphore is shared across all jobs spawned by the job manager and is
 	// used to limit the download concurrency.
 	maxParallelismSem *semaphore.Weighted
+
+	// Channel which is used by goroutines to know which offsets need to be
+	// downloaded when parallel download is enabled.
+	offsetChan chan data.ObjectRange
```
sethiay, line 97: nit: shall we call it rangeChan ?

internal/cache/file/downloader/jm_parallel_downloads_test.go
```diff
@@ -145,9 +138,9 @@ func TestParallelDownloads(t *testing.T) {
 				select {
 				case jobStatus := <-subscriberC:
 					if assert.Nil(t, err) {
-						assert.Equal(t, tc.expectedOffset, jobStatus.Offset)
+						assert.GreaterOrEqual(t, tc.objectSize, jobStatus.Offset)
 						verifyFileTillOffset(t,
-							data.FileSpec{Path: util.GetDownloadPath(path.Join(cacheDir, storage.TestBucketName), "path/in/gcs/foo.txt"), FilePerm: util.DefaultFilePerm, DirPerm: util.DefaultDirPerm}, tc.expectedOffset,
+							data.FileSpec{Path: util.GetDownloadPath(path.Join(cacheDir, storage.TestBucketName), "path/in/gcs/foo.txt"), FilePerm: util.DefaultFilePerm, DirPerm: util.DefaultDirPerm}, jobStatus.Offset,
```
kislaykishore, line 143: We should probably verifyFileTillOffset iff assert.GreaterOrEqual has returned true i.e.

```
if assert.Nil(t, err) && assert.GreaterOrEqual(t, tc.objectSize, jobStatus.Offset) {
    verifyFileTillOffset(...)
}
```

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -58,61 +60,158 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
...
 
-	_, err = io.CopyN(dstWriter, newReader, end-start)
+	_, err = io.CopyN(dstWriter, newReader, int64(end-start))
 	if err != nil {
 		err = fmt.Errorf("downloadRange: error at the time of copying content to cache file %w", err)
 	}
 	return err
 }
 
+// RangeMap maintains the ranges downloaded by the different goroutines. This
```
sethiay, line 72: nit: Add a line - "If the ranges on merging completes a range starting 0, then update the offset in job's status"

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -58,61 +60,158 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
...
-			rangeEnd := min(rangeStart+parallelReadRequestSize, end)
-			currGoRoutineIdx := goRoutineIdx
-
-			// Respect max download parallelism only beyond first go routine.
-			if currGoRoutineIdx > 0 && !job.maxParallelismSem.TryAcquire(1) {
-				break
-			}
+	rangeMap := make(map[uint64]uint64)
+	// Trying to keep the channel size greater than PrallelDownloadsPerFile to ensure
+	// that there is no goroutine waiting for data.
```
sethiay, line 151: nit: Trying to keep -> Keeping.
no goroutine waiting for data -> no goroutine waiting for next range to download.

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -58,61 +60,158 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
...
+		finalEnd = rightEnd
+		delete(rangeMap, offsetEnd)
+		delete(rangeMap, rightEnd)
+	}
+
+	rangeMap[finalStart] = finalEnd
+	rangeMap[finalEnd] = finalStart
+
+	if finalStart == 0 {
+		updateErr := job.updateStatusOffset(finalEnd)
```
kislaykishore, line 107: Nit:
```suggestion
		if updateErr := job.updateStatusOffset(finalEnd); updateErr != nil {
		    return updateErr
		}
```

sethiay (approved):

kislaykishore (commented):
Haven't reviewed the logic. Just looked at it from a readability angle.

internal/cache/file/downloader/parallel_downloads_job.go
vadlakondaswetha: 1. Trying is already verb. I dont we can use trying to keeping.
2. Updated

internal/cache/file/downloader/parallel_downloads_job.go
vadlakondaswetha: is there any go guideline on this. I prefer keeping in seperate lines to improve readability.

internal/cache/file/downloader/parallel_downloads_job.go
vadlakondaswetha: Thanks . Missed it. updated now.

internal/cache/file/downloader/jm_parallel_downloads_test.go
vadlakondaswetha: When we have multiple asserts, 2nd assert gets executed only if 1st assert succeeds. so its the same.

sethiay (dismissed):

internal/cache/file/downloader/parallel_downloads_job.go
kislaykishore: It's Idiomatic - go/go-style/guide#concision

sethiay (dismissed):

internal/cache/data/object_range.go
```diff
@@ -0,0 +1,6 @@
+package data
+
+type ObjectRange struct {
```
kislaykishore, line 3: Exported type. Please add a doc: https://google.github.io/styleguide/go/decisions#doc-comments

kislaykishore (dismissed):

kislaykishore (approved):

sethiay (approved):

PR #2296, opened 2024-08-06
fake bucket implementation fix for HNS APIs

Tulsishah:
### Description
- **CreateFolder**: In the hierarchical bucket, control client API creates folders  as well as prefixes for backward compatibility. Therefore, a prefix object entry needs to be created.
- **DeleteFolder**: In the hierarchical bucket, control client API deletes folders as well as prefixes for backward compatibility. Therefore, a prefix object entry needs to be deleted.
- **ListObjects**: To include created folder only as prefix in the list, added check.
- **RenameFolder**: Change all the entries with given source prefix to destination prefix.

NA

### Testing details
1. Manual - NA
2. Unit tests - Automated
3. Integration tests - NA


internal/storage/fake/bucket.go
```diff
@@ -973,6 +1004,25 @@ func (b *bucket) CreateFolder(ctx context.Context, folderName string) (*gcs.Fold
...
+	// Find any existing record for this name.
+	existingObjectPrefixIndex := b.objects.find(folderName)
+
+	// Create a prefix object record from the given attributes.
+	var prefixObject fakeObject
+	prefixObject.metadata = gcs.Object{Name: folderName}
+
+	// Replace an entry in or add an entry to our list of objects.
+	if existingObjectPrefixIndex < len(b.objects) {
+		b.objects[existingIndex] = prefixObject
```
vadlakondaswetha, line 1020: are you supposed to use existingObjectPrefixIndex here?

internal/storage/fake/bucket.go
```diff
@@ -992,29 +1042,33 @@ func (b *bucket) RenameFolder(ctx context.Context, folderName string, destinatio
 		return nil, err
 	}
 
-	dst := b.folders[srcIndex]
-	dst.Name = destinationFolderId
+	// Find all folders starting with the given prefix and update their names.
+	for i := range b.folders {
+		if strings.HasPrefix(b.folders[i].Name, folderName) {
```
vadlakondaswetha, line 1047: Rename folder is for a specific folder right. why are you doing this prefix check?

Tulsishah: Let's say we have the following structure:
```
A/B
A/a.txt
A/C/d.txt
```
If we rename "A" to "A_new", we need to check for the prefix "A/" in both the folder and object arrays to rename "A/B" to "A_new/B" and "A/a.txt" to "A_new/a.txt".

internal/storage/fake/bucket.go
Tulsishah: Done thanks.

internal/storage/fake/bucket.go
vadlakondaswetha: Got it. It is to update the child directories.

vadlakondaswetha (approved):

PR #2303, opened 2024-08-07
Add new file to type cache

kislaykishore:
### Description
Add new file to type cache

This avoids an additional list call in case of implicit-dirs.
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1657,9 +1657,12 @@ func (fs *fileSystem) createLocalFile(
 	// Find the parent.
 	fs.mu.Lock()
 	parent := fs.dirInodeOrDie(parentID)
-	fs.mu.Unlock()
 
 	defer func() {
+		if err != nil {
+			// fs.mu lock is already taken
+			delete(fs.localFileInodes, child.Name())
+		}
 		// We need to release the filesystem lock before acquiring the inode lock.
 		fs.mu.Unlock()
 
@@ -1670,33 +1673,37 @@ func (fs *fileSystem) createLocalFile(
 		}
 	}()
 
-	fs.mu.Lock()
-
 	fullName := inode.NewFileName(parent.Name(), name)
-	child, ok := fs.localFileInodes[fullName]
+	chld, ok := fs.localFileInodes[fullName]
+	child = chld
 
 	if ok && !child.(*inode.FileInode).IsUnlinked() {
 		return
 	}
 
 	// Create a new inode when a file is created first time, or when a local file is unlinked and then recreated with the same name.
-	var result *inode.Core
-	result, err = parent.CreateLocalChildFile(name)
-	if err != nil {
-		return
-	}
-
-	child = fs.mintInode(*result)
+	core := parent.LocalChildFileCore(name)
+	child = fs.mintInode(core)
 	fs.localFileInodes[child.Name()] = child
 
 	// Empty file is created to be able to set attributes on the file.
 	fileInode := child.(*inode.FileInode)
-	err = fileInode.CreateEmptyTempFile()
-	if err != nil {
-		return
+	if err := fileInode.CreateEmptyTempFile(); err != nil {
+		return nil, err
 	}
-
-	return
+	fs.mu.Unlock()
+	parent.Lock()
+	defer func() {
+		if err != nil {
+			parent.EraseFromTypeCache(name)
+		}
+		parent.Unlock()
+	}()
+	parent.InsertFileIntoTypeCache(name)
+	// Even though there is no action here that requires locking, adding locking
+	// so that the defer call that unlocks the mutex doesn't fail.
+	fs.mu.Lock()
+	return child, nil
 }
 
 // LOCKS_EXCLUDED(fs.mu)
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1695,9 +1695,7 @@ func (fs *fileSystem) createLocalFile(
 	}
 	fs.mu.Unlock()
 	parent.Lock()
-	defer func() {
-		parent.Unlock()
-	}()
+	defer parent.Unlock()
 	parent.InsertFileIntoTypeCache(name)
 	// Even though there is no action here that requires locking, adding locking
 	// so that the defer call that unlocks the mutex doesn't fail.
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1650,7 +1650,7 @@ func (fs *fileSystem) createFile(
 // Creates localFileInode with the given name under the parent inode.
 // LOCKS_EXCLUDED(fs.mu)
 // UNLOCK_FUNCTION(fs.mu)
-// LOCK_FUNCTION(in)
+// LOCK_FUNCTION(child)
 func (fs *fileSystem) createLocalFile(
 	parentID fuseops.InodeID,
 	name string) (child inode.Inode, err error) {
@@ -1674,15 +1674,17 @@ func (fs *fileSystem) createLocalFile(
 	}()
 
 	fullName := inode.NewFileName(parent.Name(), name)
-	chld, ok := fs.localFileInodes[fullName]
-	child = chld
+	child, ok := fs.localFileInodes[fullName]
 
 	if ok && !child.(*inode.FileInode).IsUnlinked() {
 		return
 	}
 
 	// Create a new inode when a file is created first time, or when a local file is unlinked and then recreated with the same name.
-	core := parent.LocalChildFileCore(name)
+	core, err := parent.CreateLocalChildFileCore(name)
+	if err != nil {
+		return
+	}
 	child = fs.mintInode(core)
 	fs.localFileInodes[child.Name()] = child
 
@@ -1694,9 +1696,6 @@ func (fs *fileSystem) createLocalFile(
 	fs.mu.Unlock()
 	parent.Lock()
 	defer func() {
-		if err != nil {
-			parent.EraseFromTypeCache(name)
-		}
 		parent.Unlock()
 	}()
 	parent.InsertFileIntoTypeCache(name)
--- a/internal/fs/inode/base_dir.go
+++ b/internal/fs/inode/base_dir.go
@@ -199,8 +199,12 @@ func (d *baseDirInode) CreateChildFile(ctx context.Context, name string) (*Core,
 	return nil, fuse.ENOSYS
 }
 
-func (d *baseDirInode) CreateLocalChildFile(name string) (*Core, error) {
-	return nil, fuse.ENOSYS
+func (d *baseDirInode) InsertFileIntoTypeCache(_ string) {}
+
+func (d *baseDirInode) EraseFromTypeCache(_ string) {}
+
+func (d *baseDirInode) LocalChildFileCore(_ string) Core {
+	return Core{}
 }
 
 func (d *baseDirInode) CloneToChildFile(ctx context.Context, name string, src *gcs.MinObject) (*Core, error) {
--- a/internal/fs/inode/base_dir.go
+++ b/internal/fs/inode/base_dir.go
@@ -203,8 +203,8 @@ func (d *baseDirInode) InsertFileIntoTypeCache(_ string) {}
 
 func (d *baseDirInode) EraseFromTypeCache(_ string) {}
 
-func (d *baseDirInode) LocalChildFileCore(_ string) Core {
-	return Core{}
+func (d *baseDirInode) CreateLocalChildFileCore(_ string) (Core, error) {
+	return Core{}, fuse.ENOSYS
 }
 
 func (d *baseDirInode) CloneToChildFile(ctx context.Context, name string, src *gcs.MinObject) (*Core, error) {
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -92,9 +92,14 @@ type DirInode interface {
 	// Return the full name of the child and the GCS object it backs up.
 	CreateChildFile(ctx context.Context, name string) (*Core, error)
 
-	// Create an empty local child file with the supplied (relative) name. Local
-	// file means the object is not yet created in GCS.
-	CreateLocalChildFile(name string) (*Core, error)
+	// LocalChildFileCore returns an empty local child file core.
+	LocalChildFileCore(name string) Core
+
+	// InsertFileIntoTypeCache adds the given file-name to type-cache
+	InsertFileIntoTypeCache(name string)
+
+	// EraseFromTypeCache removes the given file-name from type-cache
+	EraseFromTypeCache(name string)
 
 	// Like CreateChildFile, except clone the supplied source object instead of
 	// creating an empty object.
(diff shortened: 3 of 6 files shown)


internal/fs/inode/dir.go
```diff
@@ -771,15 +771,19 @@ func (d *dirInode) CreateChildFile(ctx context.Context, name string) (*Core, err
 	}, nil
 }
 
-func (d *dirInode) CreateLocalChildFile(name string) (*Core, error) {
+func (d *dirInode) CreateLocalChildFile(name string, updateInodeFn func(core Core) error) error {
```
raj-prince, line 774: Since we are modifying the d.cache, that means this method requires LOCKS_REQUIRED(d).

internal/fs/inode/dir.go
```diff
@@ -94,7 +94,7 @@ type DirInode interface {
 
 	// Create an empty local child file with the supplied (relative) name. Local
 	// file means the object is not yet created in GCS.
-	CreateLocalChildFile(name string) (*Core, error)
+	CreateLocalChildFile(name string, updateInodeFn func(core Core) error) error
```
raj-prince, line 97: I understand this callback is to make sure, cache insertion happens after creation of local file inode. But I would avoid passing callback, as it will complicate the lock ordering.

Also, I don't see any problem in inserting into the type-cache before inode creation. We were doing similar before only.

@vadlakondaswetha any thought here?

raj-prince (commented):
(a) (Out of scope of this PR) Currently type cache is guarded twice, one with dirInode lock another one with lru lock. We need to remove one.
(b) Can we write a concurrent test, stat and fileCreation.

internal/fs/inode/dir.go
vadlakondaswetha, line 774: Did you consider the scenarios of updating it here vs updating the type-cache when the sync to GCS is done.
for eg: create file (not synced to GCS), delete, create dir with same name?

internal/fs/inode/dir.go
```diff
@@ -771,15 +771,19 @@ func (d *dirInode) CreateChildFile(ctx context.Context, name string) (*Core, err
...
-
-	return &Core{
+	core := Core{
 		Bucket:    d.Bucket(),
 		FullName:  fullName,
 		MinObject: nil,
 		Local:     true,
-	}, nil
+	}
+	if err := updateInodeFn(core); err != nil {
```
vadlakondaswetha, line 782: Also can you add some details in description why all these changes are required.

internal/fs/inode/dir.go
raj-prince: We don't get parent Inode in sync/flush op.

internal/fs/fs.go
```diff
@@ -1619,26 +1618,19 @@ func (fs *fileSystem) createLocalFile(
 		}
 	}()
 
-	fs.mu.Lock()
-
 	fullName := inode.NewFileName(parent.Name(), name)
 	child, ok := fs.localFileInodes[fullName]
```
raj-prince, line 1622: This requires fs lock. Ref: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1735

Seems we don't have concurrent test for this scenario. This would have caught with concurrent create and delete files test.

internal/fs/fs.go
```diff
@@ -1619,26 +1618,19 @@ func (fs *fileSystem) createLocalFile(
...
-		child = fs.mintInode(*result)
-		fs.localFileInodes[child.Name()] = child
-
-		// Empty file is created to be able to set attributes on the file.
-		fileInode := child.(*inode.FileInode)
-		err = fileInode.CreateEmptyTempFile()
-		if err != nil {
-			return
-		}
+		parent.Lock()
```
raj-prince, line 1624: This doesn't follow the lock ordering. parent.Lock() should be acquired before fs.Lock()

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -623,6 +623,106 @@ func (s *concurrentListingTest) Test_ListWithMoveDir(t *testing.T) {
 	}
 }
 
+// Test_ListWithNewFileWrite tests for potential deadlocks or race conditions when
+// listing and creating a new file happen concurrently.
+func (s *concurrentListingTest) Test_ListWithNewFileWrite(t *testing.T) {
```
raj-prince, line 628: We already have this test - Test_Parallel_ReadDirAndFileEdit

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -623,6 +623,106 @@ func (s *concurrentListingTest) Test_ListWithMoveDir(t *testing.T) {
...
+
+	select {
+	case <-done:
+		// Success: Both operations finished before timeout
+	case <-time.After(timeout):
+		assert.FailNow(t, "Possible deadlock or race condition detected")
+	}
+}
+
+// Test_ListWithNewFileWrite tests for potential deadlocks or race conditions when
```
raj-prince, line 678: nit: Comment not matching with the test.

internal/fs/fs.go
kislaykishore: I've already added a test for it. Also fixed the locking.

internal/fs/inode/dir.go
kislaykishore: Refactored it to make the code more obvious.

internal/fs/inode/dir.go
kislaykishore: As discussed with @vadlakondaswetha , we can update the type-cache when the temp-file gets created. If the file gets deleted, DeleteFile will be called where we remove it from the cache.

internal/fs/fs.go
```diff
@@ -1670,33 +1673,37 @@ func (fs *fileSystem) createLocalFile(
...
-		return
+	if err := fileInode.CreateEmptyTempFile(); err != nil {
+		return nil, err
 	}
-
-	return
+	fs.mu.Unlock()
+	parent.Lock()
+	defer func() {
+		if err != nil {
```
vadlakondaswetha, line 1697: I didnt follow this. If err is not nil, you are returning from line 1692. What error will come here?

internal/fs/fs.go
```diff
@@ -1670,33 +1673,37 @@ func (fs *fileSystem) createLocalFile(
...
+			parent.EraseFromTypeCache(name)
+		}
+		parent.Unlock()
+	}()
+	parent.InsertFileIntoTypeCache(name)
```
vadlakondaswetha, line 1702: Also shouldn't these methods worry about locking inside them?

internal/fs/inode/base_dir.go
```diff
@@ -199,8 +199,12 @@ func (d *baseDirInode) CreateChildFile(ctx context.Context, name string) (*Core,
...
 }
 
-func (d *baseDirInode) CreateLocalChildFile(name string) (*Core, error) {
-	return nil, fuse.ENOSYS
+func (d *baseDirInode) InsertFileIntoTypeCache(_ string) {}
+
+func (d *baseDirInode) EraseFromTypeCache(_ string) {}
+
+func (d *baseDirInode) LocalChildFileCore(_ string) Core {
+	return Core{}
```
vadlakondaswetha, line 207: Returning empty core is not good. All the parameters will have default values and becomes difficult to debug if something goes wrong.

tools/integration_tests/concurrent_operations/setup_test.go
```diff
@@ -47,7 +47,7 @@ func TestMain(m *testing.M) {
 
 	// Create common storage client to be used in test.
 	ctx = context.Background()
-	closeStorageClient := client.CreateStorageClientWithTimeOut(&ctx, &storageClient, time.Minute*15)
+	closeStorageClient := client.CreateStorageClientWithTimeOut(&ctx, &storageClient, time.Minute*30)
```
vadlakondaswetha, line 50: why is this increased?

internal/fs/inode/dir_test.go
```diff
@@ -1436,21 +1436,31 @@ func (t *DirTest) DeleteChildDir_ImplicitDirTrue() {
...
+	d := t.in.(*dirInode)
 
-	// Object shouldn't get created in GCS.
-	result, err := t.in.LookUpChild(t.ctx, name)
-	AssertEq(nil, err)
-	AssertEq(nil, result)
-	ExpectEq(metadata.UnknownType, t.getTypeFromCache(name))
+	tp := t.tc.Get(d.cacheClock.Now(), "abc")
+
+	AssertEq(2, tp)
```
vadlakondaswetha, line 1453: There shouldn't be new lines between line 1449 and 1453.
One new line after method under test invocation.

internal/fs/inode/dir_test.go
```diff
@@ -1436,21 +1436,31 @@ func (t *DirTest) DeleteChildDir_ImplicitDirTrue() {
...
-	// Create the local file inode.
-	core, err := t.in.CreateLocalChildFile(name)
+	AssertEq(t.bucket.Name(), core.Bucket.Name())
+	AssertEq("foo/bar/qux", core.FullName.objectName)
+	AssertTrue(core.Local)
+}
 
-	AssertEq(nil, err)
-	AssertEq(true, core.Local)
-	AssertEq(nil, core.MinObject)
```
vadlakondaswetha, line 1447: We should also check if minObject is nil.

internal/fs/inode/dir_test.go
```diff
@@ -1436,21 +1436,31 @@ func (t *DirTest) DeleteChildDir_ImplicitDirTrue() {
...
+func (t *DirTest) InsertIntoTypeCache() {
+	t.in.InsertFileIntoTypeCache("abc")
+	d := t.in.(*dirInode)
 
-	// Object shouldn't get created in GCS.
-	result, err := t.in.LookUpChild(t.ctx, name)
-	AssertEq(nil, err)
```
vadlakondaswetha, line 1451: why are we not checking this?

kislaykishore: Added them back. Initially, it seemed like we are not really doing anything while creating the core except mapping it to a Core object. So, wasn't sure if we should test all these flows, but now I think it makes sense.

tools/integration_tests/concurrent_operations/setup_test.go
raj-prince: FYI, we want to get rid of this timeout: https://github.com/GoogleCloudPlatform/gcsfuse/pull/2385

internal/fs/inode/dir.go
```diff
@@ -92,9 +92,14 @@ type DirInode interface {
...
-	// Create an empty local child file with the supplied (relative) name. Local
-	// file means the object is not yet created in GCS.
-	CreateLocalChildFile(name string) (*Core, error)
+	// LocalChildFileCore returns an empty local child file core.
+	LocalChildFileCore(name string) Core
+
+	// InsertFileIntoTypeCache adds the given file-name to type-cache
+	InsertFileIntoTypeCache(name string)
+
+	// EraseFromTypeCache removes the given file-name from type-cache
```
raj-prince, line 101: nit: Here, name can file or directory can we change the comment file-name -> child name

internal/fs/fs.go
```diff
@@ -1670,33 +1673,37 @@ func (fs *fileSystem) createLocalFile(
...
 	fileInode := child.(*inode.FileInode)
-	err = fileInode.CreateEmptyTempFile()
-	if err != nil {
-		return
+	if err := fileInode.CreateEmptyTempFile(); err != nil {
+		return nil, err
 	}
-
-	return
+	fs.mu.Unlock()
```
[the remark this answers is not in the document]
raj-prince: nit: Although out of scope of this PR but very minor  - can we change the comment `// LOCK_FUNCTION(in)` -> `// LOCK_FUNCTION(child)`

internal/fs/local_file_test.go
```diff
@@ -857,3 +857,44 @@ func (t *LocalFileTest) TestStatLocalFileAfterRecreatingItWithSameName() {
 	ExpectEq("test.txt", f.Name())
 	ExpectFalse(f.IsDir())
 }
+
+func (t *LocalFileTest) TestStatSucceedsOnNewFile() {
+	t.serverCfg.NewConfig = &cfg.Config{
```
vadlakondaswetha, line 862: Is this test method same as statOnLocalFile() ?

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -623,6 +623,54 @@ func (s *concurrentListingTest) Test_ListWithMoveDir(t *testing.T) {
 	}
 }
 
+// Test_StatWithNewFileWrite tests for potential deadlocks or race conditions when
+// listing and creating a new file happen concurrently.
```
vadlakondaswetha, line 627: does os.stat do listing on a directory?

internal/fs/fs.go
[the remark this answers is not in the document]
kislaykishore: The defer call expects that fs.mu should be locked. If we move the fs.mu.Unlock() before CreateEmptyTempFile() and for some reason CreateEmptyTempFile() fails and returns an error, the defer call would fail because fs.mu is not locked but it will still try to unlock it.

internal/fs/fs.go
```diff
@@ -1670,33 +1673,37 @@ func (fs *fileSystem) createLocalFile(
 		}
 	}()
 
-	fs.mu.Lock()
-
 	fullName := inode.NewFileName(parent.Name(), name)
-	child, ok := fs.localFileInodes[fullName]
+	chld, ok := fs.localFileInodes[fullName]
```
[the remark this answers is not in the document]
kislaykishore: Wasn't sure about the redeclaration rules so, it seemed confusing whether the same child (as that of the named-return) was being assigned or a new shadow child object was being created. I read a bit further on this and found that it's the former. So, yeah, it's safe to assign directly.

internal/fs/local_file_test.go
kislaykishore: Yes, removed it now. Thanks!

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
kislaykishore: No, it's stating. Fixed the doc.

raj-prince (approved):

tools/integration_tests/concurrent_operations/concurrent_listing_test.go
```diff
@@ -623,6 +623,54 @@ func (s *concurrentListingTest) Test_ListWithMoveDir(t *testing.T) {
...
+	targetDir := path.Join(testDirPath, testCaseDir, "explicitDir")
+	var wg sync.WaitGroup
+	wg.Add(2)
+	timeout := 400 * time.Second // Adjust timeout as needed
+
+	// Goroutine 1: Repeatedly calls Stat
+	go func() {
+		defer wg.Done()
+		for i := 0; i < iterationsForLightOperations; i++ {
+			_, err := os.Stat(targetDir)
```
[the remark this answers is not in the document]
raj-prince: image


Ohh, this was to ensure stability when two locks - fs.lock() and dirInode.lock() - come into the picture together.

vadlakondaswetha (approved):

raj-prince (approved):

Issue #2239, opened 2024-07-30
Trying to understand dir and file mode


IchordeDionysos:
**Describe the issue**
While trying to get gcsfuse to run on Cloud Run now that https://github.com/GoogleCloudPlatform/gcsfuse/issues/1016 is fixed, I'm a bit confused on the file/dir modes that are logged by gcsfuse.

First of all, they don't seem to log what is passed by `--dir-mode` and `--file-mode`.

However, the logged file mode doesn't make sense, `438`. How can the other permissions be 8 (isn't the maximum possible number 7?).

How can if go from the `DirMode` and `FileMode` mound command flags go back to what has been used for `--dir-mode` and `--file-mode`?

Thanks for any insights :)

**To Collect more Debug logs**

```
Cloud Storage FUSE will be mounted to an internal path not accessible from customer containers

severity=INFO message="Start gcsfuse/2.1.0 (Go version go1.22.3) for app \"serverless\" using mount point: /var/lib/volumes/gcs/atlantis_volume\n"

[...]

PR #2309, opened 2024-08-07
Log new configuration

kislaykishore:
### Description
Here is how the logs appear on Logs explorer:
**JSON Format**
image

**Text Format**
image

[2239](https://github.com/GoogleCloudPlatform/gcsfuse/issues/2239)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/types.go
+++ b/cfg/types.go
@@ -35,7 +35,7 @@ func (o *Octal) UnmarshalText(text []byte) error {
 	return nil
 }
 
-func (o Octal) MarshalText() (text []byte, _ error) {
+func (o Octal) MarshalText() ([]byte, error) {
 	return []byte(strconv.FormatInt(int64(o), 8)), nil
 }
 
--- a/cfg/types.go
+++ b/cfg/types.go
@@ -35,7 +35,7 @@ func (o *Octal) UnmarshalText(text []byte) error {
 	return nil
 }
 
-func (o Octal) MarshalText() (text []byte, err error) {
+func (o Octal) MarshalText() (text []byte, _ error) {
 	return []byte(strconv.FormatInt(int64(o), 8)), nil
 }
 
--- a/cfg/types.go
+++ b/cfg/types.go
@@ -35,8 +35,8 @@ func (o *Octal) UnmarshalText(text []byte) error {
 	return nil
 }
 
-func (o *Octal) String() string {
-	return fmt.Sprintf("%o", *o)
+func (o Octal) MarshalText() (text []byte, err error) {
+	return []byte(strconv.FormatInt(int64(o), 8)), nil
 }
 
 // Protocol is the datatype that specifies the type of connection: http1/http2/grpc.
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -292,7 +292,7 @@ func runCLIApp(c *cli.Context) (err error) {
 	// if these are already being logged into a log-file, otherwise
 	// there will be duplicate logs for these in both places (stdout and log-file).
 	if newConfig.Foreground || newConfig.Logging.FilePath == "" {
-		logger.Logger().Info("GCSFuse config", "config", newConfig)
+		logger.SLogger().Info("GCSFuse config", "config", newConfig)
 	}
 
 	// The following will not warn if the user explicitly passed the default value for StatCacheCapacity.
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -90,13 +90,9 @@ func getUserAgent(appName string, config string) string {
 }
 
 func logNewConfiguration(newConfig *cfg.Config) {
-	cfgStr, err := util.YAMLStringify(newConfig)
-	if err != nil {
-		logger.Warnf("failed to stringify configuration: %v", err)
-		return
-	}
-	logger.Infof("GCSFuse mount config: %s", cfgStr)
+	logger.Logger().Info("GCSFuse config", "config", newConfig)
 }
+
 func logLegacyConfiguration(flags *flagStorage, mountConfig *config.MountConfig) {
 	flagsStringified, err := util.JSONStringify(*flags)
 	if err != nil {
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -89,26 +89,6 @@ func getUserAgent(appName string, config string) string {
 	}
 }
 
-func logNewConfiguration(newConfig *cfg.Config) {
-	logger.Logger().Info("GCSFuse config", "config", newConfig)
-}
-
-func logLegacyConfiguration(flags *flagStorage, mountConfig *config.MountConfig) {
-	flagsStringified, err := util.JSONStringify(*flags)
-	if err != nil {
-		logger.Warnf("failed to stringify cli flags: %v", err)
-	} else {
-		logger.Infof("GCSFuse mount command flags: %s", flagsStringified)
-	}
-
-	mountConfigStringified, err := util.JSONStringify(*mountConfig)
-	if err != nil {
-		logger.Warnf("failed to stringify config-file: %v", err)
-	} else {
-		logger.Infof("GCSFuse mount config flags: %s", mountConfigStringified)
-	}
-}
-
 func getConfigForUserAgent(mountConfig *cfg.Config) string {
 	// Minimum configuration details created in a bitset fashion. Right now, its restricted only to File Cache Settings.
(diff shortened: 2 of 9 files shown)


internal/util/util_test.go
```diff
@@ -148,6 +148,43 @@ func (ts *UtilTest) ResolveWhenParentProcDirEnvSetAndAbsoluteFilePath() {
 	assert.Equal(ts.T(), "/var/dir/test.txt", resolvedPath)
 }
 
+func TestYAMLStringify(t *testing.T) {
```
ashmeenkaur, line 151: IMO, this is less readable than stringified JSON config. Shall we stick to JSON?

cmd/legacy_main.go
```diff
@@ -297,18 +321,10 @@ func runCLIApp(c *cli.Context) (err error) {
...
-			logger.Warnf("failed to stringify cli flags: %v", err)
-		} else {
-			logger.Infof("GCSFuse mount command flags: %s", flagsStringified)
-		}
-
-		mountConfigStringified, err := util.Stringify(*mountConfig)
-		if err != nil {
-			logger.Warnf("failed to stringify config-file: %v", err)
+		if cfg.IsNewConfigEnabled() {
+			logNewConfiguration(newConfig)
```
ashmeenkaur, line 325: Shall we log new config object regardless as that is the final one being used in the code (eventually)

kislaykishore: It does make sense, but decided against it because with the recent change, we are logging a json object instead of a plain string. I'd like to see how it works with Pantheon logging before turning it on by default. Another minor concern was that it'd add an extra log :)

kislaykishore (commented):
.

cmd/legacy_main.go
ashmeenkaur: We can check how it works with pantheon logging using https://cloud.google.com/logging/docs/agent/logging/installation.

> I didn't want to add more logs to log configuration. We are already logging the config and flag and this will add more duplicate information.

I was thinking about just logging the new config.

internal/logger/logger.go
```diff
@@ -42,6 +42,14 @@ var (
 	defaultLogger        *slog.Logger
 )
 
+// Logger exports the slog logger.
+// This allows one to use the full breadth of features provided by slog.
+//
+// TODO: export defaultLogger instead of using this function.
+func Logger() *slog.Logger {
```
gargnitingoogle, line 49: I think we should rename this function to 'SLogger' or something more explicit to make the caller code clearer.

cmd/legacy_main.go
```diff
@@ -297,19 +297,7 @@ func runCLIApp(c *cli.Context) (err error) {
...
-			logger.Infof("GCSFuse mount command flags: %s", flagsStringified)
-		}
-
-		mountConfigStringified, err := util.Stringify(*mountConfig)
-		if err != nil {
-			logger.Warnf("failed to stringify config-file: %v", err)
-		} else {
-			logger.Infof("GCSFuse mount config flags: %s", mountConfigStringified)
-		}
+		logger.Logger().Info("GCSFuse config", "config", newConfig)
```
gargnitingoogle, line 300: Can you confirm the log printed by this is the exact same as it was generated by util.Stringify ?
The util.Stringify supported returning error earlier, but I see that `logger.Logger().Info("GCSFuse config", "config", newConfig)` doesn't . What happens to the error cases now ?

kislaykishore: Error gets printed in the log. It'll look something like:
{"timestamp":{"seconds":1723196744,"nanos":102771135},"severity":"INFO","message":"GCSFuse config","config":"<error_msg>"}

cfg/types.go
```diff
@@ -35,8 +35,8 @@ func (o *Octal) UnmarshalText(text []byte) error {
 	return nil
 }
 
-func (o *Octal) String() string {
-	return fmt.Sprintf("%o", *o)
+func (o Octal) MarshalText() (text []byte, err error) {
+	return []byte(strconv.FormatInt(int64(o), 8)), nil
```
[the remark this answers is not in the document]
kislaykishore: Yes, it needs this to conform to the interface.

cmd/legacy_main.go
gargnitingoogle: > It won't be exactly the same.

Could you share the older log and newer both so we could see the difference ? I am concerned that if we change the format enough, it might affect those who are dependent on our existing log format.

> Error gets printed in the log. It'll look something like:

I see problems with that change. It changes what used to be a `[WARNING]` log earlier, and would now be an info/debug log, which is easier to miss than warning log. I don't know exactly where the stringify was supposed to fail, but if was not a harmless failure, then I think we should keep it as warning/error log.

gargnitingoogle: @kislaykishore ^

kislaykishore: @gargnitingoogle Who is dependent on our existing log format?

kislaykishore: Regarding the error scenario: we are using mostly basic types and the custom types implement `MarshalText()` which don't return error. So, no errors will be thrown while logging and if in the very remote case they are thrown, they will still get logged as INFO. The way that we are logging now changes the format and makes it a lot cleaner. So, I think it's worth doing it considering the trade-off involved.

internal/logger/logger.go
```diff
@@ -42,6 +42,14 @@ var (
 	defaultLogger        *slog.Logger
 )
 
+// SLogger exports the slog logger.
+// This allows one to use the full breadth of features provided by slog.
+//
+// TODO: export defaultLogger instead of using this function.
+func SLogger() *slog.Logger {
```
ashmeenkaur, line 49: Could we not use logger.Info() ?

kislaykishore: [logger.Info()](https://github.com/GoogleCloudPlatform/gcsfuse/blob/340fa01712e8e8231d4dcc71e7b5bcf12fda96ba/internal/logger/logger.go#L145) takes the arguments and converts them to a single string. So, the structure is lost. Let me know if I am missing something.

ashmeenkaur (dismissed):

internal/logger/logger.go
kislaykishore: Changed the implementation of `logger.Info(message string, args ...any)` to be a dumb wrapper around `slog.Info(message, args ...any)`

cmd/legacy_main.go
gargnitingoogle: > Who is dependent on our existing log format?

GCSFuse logs are a publicly feature, available to all our users and are especially used by internal users and are also used by us for debugging, so changing the log format should not be silently done.

I suggest that we call out the format change in our documentation (not sure what the best place is).

> we are using mostly basic types and the custom types implement MarshalText() which don't return error. So, no errors will be thrown while logging

According to https://pkg.go.dev/encoding/json#Marshal, the error could have occurred for one of the following: floating point value NaN, Channel, complex, and function values, cyclic data structures, so theoretically it's possible to have an error if we're missing a check on the values of any floating point in the config, which is hard to guarantee in the future.

> if in the very remote case they are thrown, they will still get logged as INFO

Is it okay to change it to the following ?

```go
if _, err := json.Marshal(config); err != nil {
    logger.Logger().Warning("GCSFuse config", "config", newConfig)
} else {
    logger.Logger().Info("GCSFuse config", "config", newConfig)
}
```

or some variation of it.

This will keep logging at INFO if there is no issue, and will log a warning if there is an error.

kislaykishore: Since it's used for debugging, I'll apprise the team regarding the log format change.

We are keeping things simple. So, such cases don't really apply here and I don't want to add further complexity if it's not needed and that too when we are not losing on the error.

The code you shared ties the logging format with json which is probably not a good idea.

kislaykishore: @gargnitingoogle Could you review this again? I wouldn't want to increase code complexity due to cases that are not gonna happen. Seems like a straightforward application of [YAGNI](https://martinfowler.com/bliki/Yagni.html) and [KISS](https://pattersonjunior.medium.com/the-kiss-principle-cb480c9adcf8) principles.

gargnitingoogle: Sure.

gargnitingoogle (approved):
LGTM from readability perspective. 

I am not fully aware of the functional changes in the PR, so not commenting on that.

ashmeenkaur (approved):

PR #2317, opened 2024-08-08
Changes to use O_DIRECT in parallel downloads

sethiay:
### Description
Changes to use O_DIRECT in parallel downloads while writing to file-cache. Writing with O_DIRECT requires memory aligned buffer and this PR takes care of creating memory aligned buffer.
This PR also exposes a flag (hidden set to true) which allows changing size of memory aligned buffers but always keep that in multiple of 4096.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added 
3. Integration tests - [Run 1](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/71b9543b-c584-450c-af24-bd67cd037685/summary), [Run2](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/d0c6639b-27a7-4691-9dba-61284eb3b14a). [Run 3](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/6b7cb35d-a210-4421-bda3-c4e418aaddc9)


internal/cache/file/downloader/jm_parallel_downloads_test.go
```diff
@@ -134,7 +134,9 @@ func TestParallelDownloads(t *testing.T) {
 			bucket := storageHandle.BucketHandle(storage.TestBucketName, "")
 			minObj, content := createObjectInStoreAndInitCache(t, cache, bucket, "path/in/gcs/foo.txt", tc.objectSize)
 			jm := NewJobManager(cache, util.DefaultFilePerm, util.DefaultDirPerm, cacheDir, 2, &cfg.FileCacheConfig{EnableParallelDownloads: true,
-				ParallelDownloadsPerFile: tc.parallelDownloadsPerFile, DownloadChunkSizeMb: tc.readReqSize, EnableCrc: true, MaxParallelDownloads: tc.maxParallelDownloads})
+				ParallelDownloadsPerFile: tc.parallelDownloadsPerFile,
```
vadlakondaswetha, line 137: nit: Either put all parameters in one line or one per line.

internal/cache/file/downloader/job.go
```diff
@@ -410,6 +424,14 @@ func (job *Job) downloadObjectAsync() {
 		return
 	}
 
+	// Truncate as the parallel downloads can create file of size little higher.
+	err = cacheFile.Truncate(int64(job.object.Size))
```
vadlakondaswetha, line 428: because of the explicit buffer size and buffer is not completely filled? If yes, can you add that info to the comment here.

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -60,7 +61,14 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
 
 	monitor.CaptureGCSReadMetrics(ctx, util.Parallel, end-start)
 
-	_, err = io.CopyN(dstWriter, newReader, end-start)
+	_, err = cacheutil.CopyUsingMemoryAlignedBuffer(ctx, newReader, dstWriter, end-start,
+		job.fileCacheConfig.WriteBufferSize)
+	// If context is canceled while reading/writing in CopyUsingMemoryAlignedBuffer
+	// then it returns invalid argument error, and we need to report that error as
+	// context cancelled.
+	if errors.Is(ctx.Err(), context.Canceled) {
```
vadlakondaswetha, line 69: I didnt follow this. You mentioned in the comment that it will return invalid argument error and checking if it is context.Cancelled error?

internal/cache/util/util.go
```diff
@@ -180,3 +183,97 @@ func TruncateAndRemoveFile(filePath string) error {
...
+		if l != 0 {
+			buffer = nil
+			err = fmt.Errorf("failed to align buffer")
+		}
+	}
+
+	// Though we haven't seen any error while aligning buffer but still it is safer
+	// to attempt few times in case alignment fails.
+	for try := 0; try < 3; try++ {
+		createAndAlignBuffer()
```
vadlakondaswetha, line 215: make this function return buffer and err. Its confusing you are calling a nested function which is setting your return params

internal/cache/util/util_test.go
```diff
@@ -388,3 +393,230 @@ func Test_CreateCacheDirectoryIfNotPresentAt_ShouldReturnErrorWhenDirectoryDoesN
...
+			alignSize:           13,
+			expectedBufferSize:  7,
+			bufferPtrMultipleOf: 13,
+		},
+		{
+			name:                "buffer size smaller than align size",
+			bufferSize:          200,
+			alignSize:           4096,
+			expectedBufferSize:  200,
+			bufferPtrMultipleOf: 4096,
```
vadlakondaswetha, line 452: isn;t this always same as align size. Can we use the same param to avoid confusion

cfg/config.go
```diff
@@ -85,6 +85,8 @@ type FileCacheConfig struct {
 	MaxSizeMb int64 `yaml:"max-size-mb,omitempty"`
 
 	ParallelDownloadsPerFile int64 `yaml:"parallel-downloads-per-file,omitempty"`
+
+	WriteBufferSize int64 `yaml:"write-buffer-size,omitempty"`
```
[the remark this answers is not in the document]
sethiay: this will not be documented anywhere and is just a safety net incase something goes wrong. Yes, this can be configured anything in multiple of 4096 (4KiB)

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: Will update the comment to be more clear but the intention is that if the context is canceled then we can return context canceled error. (we are doing this because the copy function can return other error - invalid argument)

internal/cache/util/util_test.go
sethiay: it is different for the case of align size 0 (case: buffer size non 0 and align size 0)

vadlakondaswetha (dismissed):

cfg/config.go
```diff
@@ -739,5 +741,17 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	flagSet.IntP("write-buffer-size", "", 4194304, "Size of in-memory buffer that is used per goroutine in parallel downloads while writing to file-cache.")
```
raj-prince, line 744: nit: should we create a const for default buffer-size?

sethiay: this is auto-generated code, we are specifying the default here (and I followed the convention which is used currently i.e. directly specifying the default, which I think is fine) : 
image

I will add a comment though that it is 4MiB.

internal/cache/util/util.go
```diff
@@ -180,3 +183,97 @@ func TruncateAndRemoveFile(filePath string) error {
...
+		return 0, fmt.Errorf("buffer size (%v) should be a multiple of %v", bufferSize, alignSize)
+	}
+
+	calculateReqBufferSize := func(remainingContentSize int64) int64 {
+		numOfAlignSizeBlocks := math.Ceil(float64(min(bufferSize, remainingContentSize)) / float64(alignSize))
+		return int64(numOfAlignSizeBlocks) * alignSize
+	}
+
+	reqBufferSize := calculateReqBufferSize(contentSize)
+	buffer, err := GetMemoryAlignedBuffer(uint64(reqBufferSize), uint64(alignSize))
```
raj-prince, line 242: Why are we not handling the error here?

internal/cache/util/util.go
```diff
@@ -180,3 +183,97 @@ func TruncateAndRemoveFile(filePath string) error {
...
+// If contentSize is lesser than MinimumAlignSizeForWriting then extra null data
+// is written at the last.
+func CopyUsingMemoryAlignedBuffer(ctx context.Context, src io.Reader, dst io.Writer, contentSize, bufferSize int64) (n int64, err error) {
+	var alignSize int64 = MinimumAlignSizeForWriting
+	if bufferSize < alignSize || ((bufferSize % alignSize) != 0) {
+		return 0, fmt.Errorf("buffer size (%v) should be a multiple of %v", bufferSize, alignSize)
+	}
+
+	calculateReqBufferSize := func(remainingContentSize int64) int64 {
+		numOfAlignSizeBlocks := math.Ceil(float64(min(bufferSize, remainingContentSize)) / float64(alignSize))
```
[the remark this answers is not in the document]
raj-prince: Also these calculations will required only when remainingContentSize > bufferSize. Otherwise bufferSize will be the answer.

internal/cache/util/util.go
sethiay: good catch, handled.

raj-prince (commented):
LGTM overall, couple of minor comments.

internal/cache/util/util.go
```diff
@@ -38,15 +39,16 @@ const (
...
-	DefaultFilePerm = os.FileMode(0600)
-	DefaultDirPerm  = os.FileMode(0700)
-	FileCache       = "gcsfuse-file-cache"
+	MiB                        = 1024 * 1024
+	KiB                        = 1024
+	DefaultFilePerm            = os.FileMode(0600)
+	DefaultDirPerm             = os.FileMode(0700)
+	FileCache                  = "gcsfuse-file-cache"
+	BufferSizeForCRC           = 65536
+	MinimumAlignSizeForWriting = 4096
```
raj-prince, line 51: should we get the MinimumAlignSize using - https://pkg.go.dev/golang.org/x/sys/unix#Getpagesize?

sethiay: I am not sure if there's a link between page size and the align size. If you know could you please point to that ? If not, shall we unblock this PR and continue finding ?

raj-prince: Discussed offline, there is no link b/w page size and align size. It's more related to device block-size, as we are keeping minimum to work with all the block-devices.

raj-prince (approved):

PR #2320, opened 2024-08-08
Specify coverpkg to include every package

kislaykishore:
Without coverpkg being set, the coverage computed for a given test is limited to its own package alone and doesn't include the code from other packages that were involved. Setting coverpkg to ./... includes that coverage as well.

This is useful in cases like the cfg/ package where the `bindFlags` method is invoked by the tests in cmd/root_test.go package.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -20,10 +20,10 @@ jobs:
         run: go mod download
 
       - name: Coverage for all except cache tests parallely.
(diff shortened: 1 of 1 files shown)


gargnitingoogle (approved):
LGTM

vadlakondaswetha (approved):

PR #2326, opened 2024-08-09
Update troubleshoot for kernel-list-cache ttl behavior with empty-dir.

raj-prince:
Updating troubleshoot doc for kernel-list-cache ttl behavior with empty-dir.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


docs/troubleshooting.md
```diff
@@ -3,28 +3,29 @@
...
...                                                                                                                                                                                                                                                                                                                              | In case a new file is added to the empty directory remotely, outside of the mount, the client will not be able to see the new file in listing even if ttl is expired.<br></br>You need to create a new file/directory from the mount to see the newly added remote files.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
ashmeenkaur, line 31: Shall we add that kernel list cache should only be used for read only workloads?

PR #2327, opened 2024-08-09
[CLI-Config Parity Cleanup] Remove duplicate code

ashmeenkaur:
### Description
This PR cleans up separate override methods for configs as this is already done in legacy_param_mapper.go:
https://github.com/GoogleCloudPlatform/gcsfuse/blob/2051a7db5a3cf7f253d47ae4ffa601255b48beaf/cmd/legacy_param_mapper.go#L150
And tested here:
https://github.com/GoogleCloudPlatform/gcsfuse/blob/2051a7db5a3cf7f253d47ae4ffa601255b48beaf/cmd/legacy_param_mapper_test.go#L318

NA

### Testing details
1. Manual - NA
2. Unit tests - NA


kislaykishore:
As discussed offline, this is a breaking change. We need to migrate those three params before we can perform the cleanup here.

kislaykishore (changes requested):
As discussed offline, this is a breaking change. We need to migrate those three params before we can perform the cleanup here.

PR #2346, opened 2024-08-13
Create a new inode for a folder if it is deleted/rename and then recreated

Tulsishah:
### Description
When creating inodes for new objects, we verify if an inode with the same name already exists. If it does, the system compares the generation and metageneration values. If these values don't match, a new inode is created instead of reusing an existing one.

However, for folders, which are identified solely by their name, the same inode was being reused even when a new folder was created after deleting the old one. This led to inconsistencies in the inode sequence.

We encountered this scenario while running end-to-end tests on the local-file package. One test was failing due to a previous test, as both tests were using the same directory for their operations.

Scenario:
```
Create a folder named "dir"
Create a local file named "dir/file1.txt"
Remove the "dir" directory.
Create the "dir" directory again.
Unable to create files or directories within "dir" because the system reuses the same inode instead of creating a new one, leading to an inconsistent kernel sequence.
```

NA

### Testing details
1. Manual - Tested
2. Unit tests - Done
3. Integration tests - Done


internal/fs/inode/dir.go
```diff
@@ -210,6 +215,10 @@ type dirInode struct {
 	// enabled.
 	prevDirListingTimeStamp time.Time
 	isHNSEnabled            bool
+
+	// Represents if folder has been unlinked in hierarchical bucket. This is not getting used in
+	// non-hierarchical bucket.
+	unlink bool
```
vadlakondaswetha, line 221: unlink is a verb. you need a variable to specify the state. Rename it to unlinked. Look at file.go.

internal/fs/hns_bucket_test.go
```diff
@@ -48,8 +48,13 @@ var expectedFooDirEntries = []dirEntry{
 func TestHNSBucketTests(t *testing.T) { suite.Run(t, new(HNSBucketTests)) }
 
 func (t *HNSBucketTests) SetupSuite() {
-	t.serverCfg.NewConfig = &cfg.Config{EnableHns: true}
 	t.serverCfg.ImplicitDirectories = false
+	t.serverCfg.NewConfig = &cfg.Config{
+		Write: cfg.WriteConfig{
+			CreateEmptyFile: false,
```
vadlakondaswetha, line 54: isn;t this false by default?

internal/fs/hns_bucket_test.go
```diff
@@ -253,3 +269,58 @@ func (t *HNSBucketTests) TestRenameFolderWithOpenGCSFile() {
...
 	assert.False(t.T(), dirEntries[0].IsDir())
 }
+
+func (t *HNSBucketTests) TestCreateDirectoryWithSameNameAfterRename() {
+	oldDirPath := path.Join(mntDir, "foo")
+	_, err = os.Stat(oldDirPath)
+	require.NoError(t.T(), err)
+	newDirPath := path.Join(mntDir, "foo_rename")
+	err = os.Rename(oldDirPath, newDirPath)
+	require.NoError(t.T(), err)
```
vadlakondaswetha, line 279: Can you add some comments in between. Its difficult what operations are being done.

internal/fs/hns_bucket_test.go
```diff
@@ -253,3 +269,58 @@ func (t *HNSBucketTests) TestRenameFolderWithOpenGCSFile() {
...
+	assert.NoError(t.T(), err)
+	assert.Equal(t.T(), 0, len(dirEntries))
+}
+
+func (t *HNSBucketTests) TestCreateLocalFileInSamePathAfterDeletingParentDirectory() {
+	dirPath := path.Join(mntDir, "foo", "test2")
+	filePath := path.Join(dirPath, "test.txt")
+	f1, err := os.Create(filePath)
+	defer require.NoError(t.T(), f1.Close())
+	require.NoError(t.T(), err)
```
vadlakondaswetha, line 312: same here.

internal/fs/inode/base_dir.go
```diff
@@ -250,3 +251,10 @@ func (d *baseDirInode) RenameFolder(ctx context.Context, folderName string, dest
 	err = fuse.ENOSYS
 	return
 }
+
+func (d *baseDirInode) IsUnlink() bool {
```
vadlakondaswetha, line 255: can you add comments here that these are not supported on base_dir

internal/fs/inode/dir_test.go
```diff
@@ -1378,7 +1378,7 @@ func (t *DirTest) DeleteChildFile_TypeCaching() {
 func (t *DirTest) DeleteChildDir_DoesntExist() {
 	const name = "qux"
 
-	err := t.in.DeleteChildDir(t.ctx, name, false)
+	err := t.in.DeleteChildDir(t.ctx, name, false, t.in)
```
vadlakondaswetha, line 1381: is the last parameter correct?. t.in is the pareent and you need to pass child to DeleteChildDir()

Tulsishah: Thanks for catch. Fixed

internal/fs/hns_bucket_test.go
Tulsishah: I thought showing it explicitly makes code more readable.

vadlakondaswetha (approved):

PR #2349, opened 2024-08-14
Local file bug fix - Recreation of local file had issue in stating

Tulsishah:
### Description
I found an issue with the behavior of local files where recreating a local file with the same name causes a problem. The situation is as follows.
**Behavior:**
1. Directory creation: Successful.
2. Local file creation within the directory: Successful.
3. Stat operation on the local file: Successful.
4. Delete local file: Successful
5. Recreation of the local file within the directory with the same name: Successful.
6. Stat operation on the recreated local file: Failed (due to the reuse of the same inode).

e.g.
```
create dir
create local file dir/file.txt
stat dir/file.txt
rm dir/file.txt
create local file dir/file.txt
stat dir/file.txt -- Fail
```

**Fix:**
The issue was resolved by ensuring the creation of a new inode whenever a file with the same name created which unlinked.

NA

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - Added

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1675,23 +1675,26 @@ func (fs *fileSystem) createLocalFile(
 
 	fullName := inode.NewFileName(parent.Name(), name)
 	child, ok := fs.localFileInodes[fullName]
+
+	if ok && !child.(*inode.FileInode).IsUnlinked() {
+		return
+	}
+
 	// Create a new inode if a local file is unlinked and then recreated with the same name.
-	if !ok || (child != nil && child.(*inode.FileInode).IsUnlinked()) {
-		var result *inode.Core
-		result, err = parent.CreateLocalChildFile(name)
-		if err != nil {
-			return
-		}
+	var result *inode.Core
+	result, err = parent.CreateLocalChildFile(name)
+	if err != nil {
+		return
+	}
 
-		child = fs.mintInode(*result)
-		fs.localFileInodes[child.Name()] = child
+	child = fs.mintInode(*result)
+	fs.localFileInodes[child.Name()] = child
 
-		// Empty file is created to be able to set attributes on the file.
-		fileInode := child.(*inode.FileInode)
-		err = fileInode.CreateEmptyTempFile()
-		if err != nil {
-			return
-		}
+	// Empty file is created to be able to set attributes on the file.
+	fileInode := child.(*inode.FileInode)
+	err = fileInode.CreateEmptyTempFile()
+	if err != nil {
+		return
 	}
 
 	return
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1680,7 +1680,7 @@ func (fs *fileSystem) createLocalFile(
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


internal/fs/fs.go
```diff
@@ -1675,7 +1675,12 @@ func (fs *fileSystem) createLocalFile(
 
 	fullName := inode.NewFileName(parent.Name(), name)
 	child, ok := fs.localFileInodes[fullName]
-	if !ok {
+	// Create a new inode if a local file is unlinked and then recreated with the same name.
+	var fileInode *inode.FileInode
+	if child != nil {
+		fileInode = fs.fileInodeOrDie(child.ID())
+	}
+	if !ok || (fileInode != nil && fileInode.IsUnlinked()) {
```
vadlakondaswetha, line 1683: Can you simply do this: 
if !ok || child.(*inode.FileInode).IsUnlinked()

Tulsishah: What if child is nil? Shouldn't we keep `if !ok || (child!=nil && child.(*inode.FileInode).IsUnlinked())` ?

internal/fs/fs.go
```diff
@@ -1675,7 +1675,8 @@ func (fs *fileSystem) createLocalFile(
 
 	fullName := inode.NewFileName(parent.Name(), name)
 	child, ok := fs.localFileInodes[fullName]
-	if !ok {
+	// Create a new inode if a local file is unlinked and then recreated with the same name.
+	if !ok || (child != nil && child.(*inode.FileInode).IsUnlinked()) {
```
kislaykishore, line 1679: We can reduce nesting by returning quickly: https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

if !(!ok || (child != nil && !child.(*inode.FileInode).IsUnlinked())) {
  return
}

The predicate can probably be simplified further by applying the [De Morgan's law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)

vadlakondaswetha: +1 Agree with short circuits part - you should always write shorter blocks first. 
but multiple negations in if clause are even complicated to understand.

child != nil condition is not required since we are already checking ok .
So you can simplify as 
if ok && !child.(*inode.FileInode).isUnlinked() {
  return
}

internal/fs/fs.go
```diff
@@ -1675,22 +1675,26 @@ func (fs *fileSystem) createLocalFile(
...
+		return
+	}
 
-		// Empty file is created to be able to set attributes on the file.
-		fileInode := child.(*inode.FileInode)
-		err = fileInode.CreateEmptyTempFile()
-		if err != nil {
-			return
-		}
+	// Create a new inode if a local file is unlinked and then recreated with the same name.
```
vadlakondaswetha, line 1683: update the comment. it is created for the first time as well.

Tulsishah: I thought it's an obvious/default scenario that's why I haven't added. Added comment for the change only. 

But feel it is confusing so anyway I added the comment.

vadlakondaswetha (approved):

PR #2357, opened 2024-08-19
Handle empty gsheet credentials when running load tests

anushka567:
### Description
Includes fixes to the build scripts from continuous tests , which ensures that the perf tests are executed for both hns and flat buckets and the results are uploaded to respective gsheet whenever the perf experiments are triggered. 

NA

### Testing details
1. Manual - Manually tested perf tests script which runs load test for fio benchmarking and performs listing benchmark.Successful run and upload to gsheet.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
+++ b/perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
@@ -40,46 +40,47 @@ then
 fi
 
 run_load_test_and_fetch_metrics(){
-  gcsfuse_fio_flags=$1
-  upload_flags=$2
-  bucket_name=$3
-  spreadsheet_id=$4
-  gsheet_credentials=$5
-  test_type=$6
-  gcsfuse_ls_flags=$7
-  ls_config_file=$8
+  fio_flags=$1
+  gcsfuse_flags="$COMMON_MOUNT_FLAGS $fio_flags"
+  bucket_name=$2
+  spreadsheet_id=$3
 
   # Executing perf tests
-  ./run_load_test_and_fetch_metrics.sh "$gcsfuse_fio_flags" "$upload_flags" "$bucket_name" "$spreadsheet_id" "$gsheet_credentials"
+  ./run_load_test_and_fetch_metrics.sh "$gcsfuse_flags" "$UPLOAD_FLAGS" "$bucket_name" "$spreadsheet_id"
+}
 
-  # ls_metrics test. This test does gcsfuse mount with the passed flags first and then does the testing.
-  log_file_list_tests=${KOKORO_ARTIFACTS_DIR}/gcsfuse-list-logs-$test_type.txt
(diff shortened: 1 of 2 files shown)


perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
```diff
@@ -40,23 +40,23 @@ then
...
-  SPREADSHEET_ID=$4
-  GSHEET_CREDENTIALS=$5
-  TEST_TYPE=$6
-  GCSFUSE_LS_FLAGS=$7
-  LS_CONFIG_FILE=$8
+  gcsfuse_fio_flags=$1
+  upload_flags=$2
+  bucket_name=$3
+  spreadsheet_id=$4
+  gsheet_credentials=$5
```
Tulsishah, line 47: Can we avoid `gsheet_credentials` variable as I feel credential will be same for both the cases so we can capture it from any one bucket?

perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
```diff
@@ -40,23 +40,23 @@ then
...
 run_load_test_and_fetch_metrics(){
-  GCSFUSE_FIO_FLAGS=$1
-  UPLOAD_FLAGS=$2
-  BUCKET_NAME=$3
-  SPREADSHEET_ID=$4
-  GSHEET_CREDENTIALS=$5
-  TEST_TYPE=$6
-  GCSFUSE_LS_FLAGS=$7
-  LS_CONFIG_FILE=$8
+  gcsfuse_fio_flags=$1
```
Tulsishah, line 43: Ideally we should catch gcsfuse_flags only and then create fio_flag and ls_flags in the function.

perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
```diff
@@ -40,23 +40,23 @@ then
...
+  upload_flags=$2
+  bucket_name=$3
+  spreadsheet_id=$4
```
Tulsishah, line 46: Do we really needs these arguments which are not changing for any cases like "upload_flags" or "gcsfuse_flags". can't we access it without passing it to the function as it is global.

perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
anushka567: Updated.

perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
anushka567: Updated with common global flags.

perfmetrics/scripts/continuous_test/gcp_ubuntu/build.sh
anushka567: Reverted to always use creds.json present in periodic perf test bucket

Tulsishah (approved):

PR #2366, opened 2024-08-19
[CLI Config Parity] Migrate metadata cache config/flags

ashmeenkaur:
### Description
Migrate metadata cache config/flags. This PR also changes the logic of resolving deprecated and new metadata cache configs by using viper's IsSet method.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


cmd/mount.go
```diff
@@ -85,11 +83,11 @@ be interacting with the file system.`)
 		gid = uint32(newConfig.FileSystem.Gid)
 	}
 
-	metadataCacheTTL := mount.ResolveMetadataCacheTTL(newConfig.MetadataCache.DeprecatedStatCacheTtl, newConfig.MetadataCache.DeprecatedTypeCacheTtl, mountConfig.MetadataCacheConfig.TtlInSeconds)
-	statCacheMaxSizeMB, err := mount.ResolveStatCacheMaxSizeMB(mountConfig.StatCacheMaxSizeMB, int(newConfig.MetadataCache.DeprecatedStatCacheCapacity))
+	metadataCacheTTL := cfg.ResolveMetadataCacheTTL(newConfig.MetadataCache.DeprecatedStatCacheTtl, newConfig.MetadataCache.DeprecatedTypeCacheTtl, newConfig.MetadataCache.TtlSecs)
+	statCacheMaxSizeMB, err := cfg.ResolveStatCacheMaxSizeMB(newConfig.MetadataCache.StatCacheMaxSizeMb, int(newConfig.MetadataCache.DeprecatedStatCacheCapacity))
```
kislaykishore, line 87: Can we resolve the ttl and statCacheMaxSizeMB upfront in rationalize.go and use the resolved values here? It seems to make more sense there than here. If you agree, we can also keep the Resolve* methods unexported.

cfg/config_util.go
```diff
@@ -41,3 +44,43 @@ func ListCacheTTLSecsToDuration(secs int64) time.Duration {
 
 	return time.Duration(secs * int64(time.Second))
 }
+
+// ResolveMetadataCacheTTL returns the ttl to be used for stat/type cache based on the user flags/configs.
+func ResolveMetadataCacheTTL(statCacheTTL, typeCacheTTL time.Duration, ttlInSeconds int64) (metadataCacheTTL time.Duration) {
```
kislaykishore, line 49: Should these functions be in rationalize.go?

cfg/constants.go
```diff
@@ -45,7 +47,39 @@ const (
...
+	AverageSizeOfPositiveStatCacheEntry uint64 = 1400
+	// AverageSizeOfNegativeStatCacheEntry is the assumed size of each negative stat-cache-entry,
+	// meant for two purposes.
+	// 1. for conversion from stat-cache-capacity to stat-cache-max-size-mb.
+	// 2. internal testing.
+	AverageSizeOfNegativeStatCacheEntry uint64 = 240
+	// defaultStatCacheMaxSizeMB is the default for stat-cache-max-size-mb
+	// and is to be used when neither stat-cache-max-size-mb nor
+	// stat-cache-capacity is set.
+	defaultStatCacheMaxSizeMB = 32
```
kislaykishore, line 84: Is this constant being used anywhere? If not, let's delete it.

kislaykishore (changes requested):

cfg/config_util.go
```diff
@@ -41,3 +44,43 @@ func ListCacheTTLSecsToDuration(secs int64) time.Duration {
...
+		if ttlInSeconds == -1 {
+			metadataCacheTTL = time.Duration(math.MaxInt64)
+		} else {
+			metadataCacheTTL = time.Second * time.Duration(ttlInSeconds)
+		}
+	} else {
+		metadataCacheTTL = time.Second * time.Duration(uint64(math.Ceil(math.Min(statCacheTTL.Seconds(), typeCacheTTL.Seconds()))))
+	}
+	return
+}
```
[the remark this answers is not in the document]
ashmeenkaur: Done, took a slightly different approach of using 2 separate implementation of the isSet interface for old (manually implemented) and new config(uses viper). Please take another look and see if it looks good now.

cfg/validate.go
```diff
@@ -107,6 +107,40 @@ func isValidKernelListCacheTTL(TTLSecs int64) error {
 	return nil
 }
 
+func isValidMetadataCache(c MetadataCacheConfig) error {
+	// Validate ttl-secs.
+	if c.TtlSecs != TtlInSecsUnsetSentinel {
```
kislaykishore, line 112: What if someone sets the ttl-secs to TtlInSecsUnsetSentinel via Viper. Validation won't fail here. And it will skip the rationalization as well since the concerned Viper key is set.
So, the value of ttl-secs will be set to TtlInSecsUnsetSentinel itself. While this case is improbable, I was wondering if we can handle it better.

cmd/legacy_param_mapper.go
```diff
@@ -156,7 +156,7 @@ func PopulateNewConfigFromLegacyFlagsAndConfig(c cliContext, flags *flagStorage,
 	if err := cfg.ValidateConfig(resolvedConfig); err != nil {
 		return nil, fmt.Errorf("cfg.ValidateConfig: %w", err)
 	}
-	if err := cfg.Rationalize(resolvedConfig); err != nil {
+	if err := cfg.Rationalize(&isSet{resolvedConfig}, resolvedConfig); err != nil {
```
kislaykishore, line 159: Nice :)

internal/fs/type_cache_test.go
```diff
@@ -77,20 +76,24 @@ var (
...
 	}
 
 	// Fill server-cfg from mount-config.
-	func(mountConfig *config.MountConfig, serverCfg *gcsfusefs.ServerConfig) {
-		serverCfg.DirTypeCacheTTL = mount.ResolveMetadataCacheTTL(mount.DefaultStatOrTypeCacheTTL, mount.DefaultStatOrTypeCacheTTL,
-			mountConfig.TtlInSeconds)
+	func(newConfig *cfg.Config, serverCfg *gcsfusefs.ServerConfig) {
+		if newConfig.MetadataCache.TtlSecs == -1 {
+			maxSupportedTTLInSecs := math.MaxInt64 / int64(time.Second)
+			newConfig.MetadataCache.TtlSecs = maxSupportedTTLInSecs
```
kislaykishore, line 90:
```suggestion
			newConfig.MetadataCache.TtlSecs = math.MaxInt64 / int64(time.Second)
```

cfg/validate.go
ashmeenkaur: Made the changes to use IsSet implementation here as well instead of relying on UnsetSentinel values.

cfg/rationalize.go
```diff
@@ -25,8 +37,36 @@ func decodeURL(u string) (string, error) {
...
+	// Else, use deprecated stat/type cache ttl to resolve metadataCacheTTL.
+	c.TtlSecs = int64(math.Ceil(math.Min(c.DeprecatedStatCacheTtl.Seconds(), c.DeprecatedTypeCacheTtl.Seconds())))
+}
+
+// resolveStatCacheMaxSizeMB returns the stat-cache size in MiBs based on the
+// user old and new flags/configs.
+func resolveStatCacheMaxSizeMB(v isSet, c *MetadataCacheConfig) {
+	// If metadata-cache:stat-cache-size-mb has been set, then it overrides
+	// stat-cache-capacity.
+	if v.IsSet("metadata-cache.stat-cache-max-size-mb") {
```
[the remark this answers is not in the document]
ashmeenkaur: Good catch! Thanks!

cfg/rationalize.go
```diff
@@ -25,8 +37,39 @@ func decodeURL(u string) (string, error) {
 	return decodedURL.String(), nil
 }
 
+// resolveMetadataCacheTTL returns the ttl to be used for stat/type cache based
+// on the user flags/configs.
+func resolveMetadataCacheTTL(v isSet, c *MetadataCacheConfig) {
+	// If metadata-cache:ttl-secs has been set, then it overrides both
+	// stat-cache-ttl and type-cache-tll.
+	if v.IsSet("metadata-cache.ttl-secs") {
```
kislaykishore, line 45: The key "metadata-cache.ttl-secs" and "metadata-cache.stat-cache-max-size-mb" are being repeated at multiple places. Shall we create constants for them?

kislaykishore (approved):

PR #2368, opened 2024-08-20
add test with cache pointing to ram based dir

sethiay:
### Description
add test with cache pointing to ram based dir

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/disabled_cache_ttl_test.go
+++ b/tools/integration_tests/read_cache/disabled_cache_ttl_test.go
@@ -96,8 +96,8 @@ func TestDisabledCacheTTLTest(t *testing.T) {
 
 	// Define flag set to run the tests.
 	flagsSet := [][]string{
-		{"--implicit-dirs", "--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileName, false)},
-		{"--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileNameForParallelDownloadTests, true)},
+		{"--implicit-dirs", "--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileName, false, getDefaultCacheDirPathForTests())},
+		{"--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileNameForParallelDownloadTests, true, getDefaultCacheDirPathForTests())},
 	}
 
 	// Run tests.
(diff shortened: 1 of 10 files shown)


tools/integration_tests/read_cache/setup_test.go
```diff
@@ -99,8 +99,12 @@ func mountGCSFuseAndSetupTestDir(flags []string, ctx context.Context, storageCli
 	testDirPath = client.SetupTestDirectory(ctx, storageClient, testDirName)
 }
 
-func createConfigFile(cacheSize int64, cacheFileForRangeRead bool, fileName string, enableParallelDownloads bool) string {
-	cacheDirPath = path.Join(setup.TestDir(), cacheDirName)
+func getDefaultCacheDirPathForTests() string {
+	return path.Join(setup.TestDir(), cacheDirName)
+}
+
+func createConfigFile(cacheSize int64, cacheFileForRangeRead bool, fileName string, enableParallelDownloads bool, customCacheDirPath string) string {
```
ashmeenkaur, line 106: nit: can we directly pass parameter as cacheDirPath and use it (instead of using local cacheDirPath variable)?

tools/integration_tests/read_cache/disabled_cache_ttl_test.go
```diff
@@ -96,8 +96,8 @@ func TestDisabledCacheTTLTest(t *testing.T) {
 
 	// Define flag set to run the tests.
 	flagsSet := [][]string{
-		{"--implicit-dirs", "--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileName, false)},
-		{"--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileNameForParallelDownloadTests, true)},
+		{"--implicit-dirs", "--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileName, false, getDefaultCacheDirPathForTests())},
+		{"--stat-cache-ttl=0s", "--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileNameForParallelDownloadTests, true, getDefaultCacheDirPathForTests())},
```
ashmeenkaur, line 100: are we explicitly not using ramCache here and in other tests?

tools/integration_tests/read_cache/setup_test.go
sethiay: `cacheDirPath` is not a local variable to this function and is used at other places also. that's why I had to use like this. 

The other option is to set cacheDirPath outside of this function and then pass it the way you are mentioning but that would require changes in all test files.

tools/integration_tests/read_cache/disabled_cache_ttl_test.go
sethiay: Yes, the intention of this test is just to ensure that file-cache with and w/o parallel downloads work with ram based storage system, so we don't need to add in all the tests.

ashmeenkaur (approved):

PR #2370, opened 2024-08-20
Collecting rename metrics for nested folders 

anushka567:
### Description
Made changes to existing functions to collected and accumulate metrics for nested folders as well.Previously, we had code which collected metrics related to rename operation for single level folders i.e.
```
folder/
            file1.txt
            file2.txt
            ...
            file100.txt
```

Thus, metrics reported for these folders were in the form of average/min/max etc. time taken for each folder level rename operation.

Since, hns atomic rename is especially crucial in nested folder rename scenario , hence we do perf tests on simple nested scenario  i.e. 
```
top_level_folder/
                            folder1/
                                        file1.txt
                                        file2.txt
                                        ...
                                        file100.txt
                            folder2/
                                        file1.txt
                                        file2.txt
                                        ...
                                        file100.txt
```

Here we report the time of rename operation for renaming top_level_folder to top_level_folder_renamed so that we can observe performance comparison in cases where implicit-dirs are involved.

NA

### Testing details
1. Manual - Tested the script to check for correct metric aggregation and upload to gsheet
2. Unit tests - Added nested folder scenario for all existing unit tests.Also, added unit test for the common functions related to gsheet upload.
3. Integration tests - NA


perfmetrics/scripts/hns_rename_folders_metrics/renaming_benchmark.py
```diff
@@ -66,6 +66,40 @@ def _upload_to_gsheet(worksheet, data, spreadsheet_id) -> (int):
   return exit_code
 
 
+def _calculate_num_files(folder_structure):
```
Tulsishah, line 69: Add comment about function and arguments

perfmetrics/scripts/hns_rename_folders_metrics/renaming_benchmark_test.py
```diff
@@ -17,6 +17,30 @@
 
 class TestRenamingBenchmark(unittest.TestCase):
 
+  def test_calculate_num_files(self):
```
Tulsishah, line 20: Add test case for nested folders.

anushka567: This function is only relevant for nested case and  takes in the folder structure JSON.So, I have updated the existing test.

Tulsishah (approved):

PR #2371, opened 2024-08-20
Add tracing to fs ops

kislaykishore:
### Description
* Add tracing to fs ops
* Increase signal to noise ratio by refactoring code.
* Annotating spans with error info will be handled in a subsequent PR.

image

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/fs/wrappers/monitoring.go
```diff
@@ -29,12 +29,18 @@ import (
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/codes"
 )
 
+const name = "github.com/googlecloudplatform/gcsfuse/fs"
```
charith87, line 36: We can change this to google's host. It shouldnt be github domain.

charith87 (commented):
Pls check if it causes a memory leak or higher cpu as it is initializing always.

charith87 (approved):
Overall Approving to unblock. 
go/lgtm-with-comments

charith87 (dismissed):

charith87 (approved):

internal/fs/wrappers/monitoring_test.go
```diff
@@ -117,3 +124,148 @@ func TestFsErrStrAndCategory(t *testing.T) {
...
+func (d dummyFS) Destroy() {}
+
+func TestSpan(t *testing.T) {
+	ex := newInMemoryExporter(t)
+	m := monitoring{
+		wrapped: dummyFS{},
+	}
+
+	err := m.StatFS(context.Background(), nil)
+	require.NoError(t, err)
```
gargnitingoogle, line 265: how is this different from assert.NoError ? If no difference, let's use assert.NoError for consistency.

internal/fs/wrappers/monitoring_test.go
```diff
@@ -117,3 +124,148 @@ func TestFsErrStrAndCategory(t *testing.T) {
...
+	return nil
+}
+
+func (d dummyFS) Fallocate(_ context.Context, _ *fuseops.FallocateOp) error {
+	return nil
+}
+
+func (d dummyFS) Destroy() {}
+
+func TestSpan(t *testing.T) {
```
gargnitingoogle, line 258: The name is a bit unclear to me. This sounds like it's testing the span functionality, which is not a gcsfuse feature.. How about TestInvocationSpan ?

internal/fs/wrappers/monitoring_test.go
```diff
@@ -117,3 +124,148 @@ func TestFsErrStrAndCategory(t *testing.T) {
...
+
+func (d dummyFS) SetXattr(_ context.Context, _ *fuseops.SetXattrOp) error {
+	return nil
+}
+
+func (d dummyFS) Fallocate(_ context.Context, _ *fuseops.FallocateOp) error {
+	return nil
+}
+
+func (d dummyFS) Destroy() {}
```
gargnitingoogle, line 256: Given that Destroy is the only function of FS which is different from the rest in signature, and destroy is also interesting for deinitialization point of view, I suggest we add a test with it, or update the existing test with a call to `m.Destroy()` to see that it get recorded by monitoring.

gargnitingoogle:
My bad. I guess this PR was open in  a tab in my browser for a few hours before I got to it, and didn't realize that it had been approved and merged already.

PR #2374, opened 2024-08-20
Add a build workflow to detect flakiness

kislaykishore:
This workflow runs the same test 5 times to figure out if there is any flakiness in the tests. It's run as a post-submit so that it doesn't impede dev-velocity.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


.github/workflows/flake-detector.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  # Enable triggering the workflow manually
+  workflow_dispatch:
+  push:
+    branches: [ "master" ]
+
+jobs:
+  codecov:
+    runs-on: ubuntu-latest
+    timeout-minutes: 75
```
Tulsishah, line 11: Will it run on the each PR? If yes isn't it more time to detect failure. Our development cycle may get affect due to that.

vadlakondaswetha: +1. 
Can you add more details to the PR. It runs 5 times and does it report even though the test failed once?

kislaykishore: It'll be run as a postsubmit i.e. after the code gets pushed to master, so won't block submissions. 
And it reports an error for each time a particular test failed.

vadlakondaswetha (dismissed):

Tulsishah (dismissed):

vadlakondaswetha (approved):

PR #2376, opened 2024-08-21
Add support to ignore flaky tests.

kislaykishore:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/flaky_tests.lst
+++ b/flaky_tests.lst
@@ -1,0 +1,2 @@
+# b/361259658
+TestAppendObjectCreator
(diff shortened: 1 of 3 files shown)


flaky_tests.lst
```diff
@@ -0,0 +1,8 @@
+# Each flaky test should be represented here in the following format:
```
Tulsishah, line 1: Do we require to add copy writes in these files?

charith87 (approved):

PR #2377, opened 2024-08-21
Make default value of enable-hns flag true

Tulsishah:
### Description
- Make default value of `enable-hns` flag true.
- Remove enableEmptyManagedFolder flag true with enableHNS flag true as it is changing flag value for flat bucket also.
- Rename enableEmptyManagedFolder in dir.go with IncludeFoldersAsPrefixes.
- Removed/Updated all the unit tests for changes.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA


cfg/config.go
kislaykishore, line 1: Please don't edit this file manually. The params.yaml should be changed to reflect the correct default value. Post that, run `make build` to generate this configuration.

kislaykishore (changes requested):

cfg/config.go
Tulsishah: Thanks.

kislaykishore (approved):
Approving for config changes. Please get the business logic change reviewed from someone in the know.

ankitaluthra1 (approved):

cfg/rationalize.go
```diff
@@ -27,11 +27,6 @@ func decodeURL(u string) (string, error) {
 
 // Rationalize updates the config fields based on the values of other fields.
 func Rationalize(c *Config) error {
-	// The EnableEmptyManagedFolders flag must be set to true to enforce folder prefixes for Hierarchical buckets.
-	if c.EnableHns {
-		c.List.EnableEmptyManagedFolders = true
```
ashmeenkaur, line 32: Default value of EnableEmptyManagedFolders should also have been changed to true? @Tulsishah

Tulsishah: As discussed offline it is intended.

PR #2382, opened 2024-08-22
Remove fs-error label

kislaykishore:
fs-error has a high cardinality and will not be supported by GCM. So, either we can drop it in gcsfuse or gcs-fuse-csi-driver will have to make changes to drop the label. fs-error is likely not needed since we already are classifying the most frequent errors into a new label called fs-err-category and the other errors are being classified into appropriate buckets. Moreover, we'd continue to have logs if we need to get more details about an (infrequent) error. So, we can remove fs-error label.

### Description
image

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/fs/wrappers/monitoring.go
```diff
@@ -75,24 +75,17 @@ func init() {
...
+func categorize(err error) string {
 	if err == nil {
-		return "", ""
+		return ""
 	}
 	var errno syscall.Errno
 	if !errors.As(err, &errno) {
 		errno = DefaultFSError
 	}
-	return errno.Error(), errCategory(errno)
```
charith87, line 89: Not really related to this CL though, I think we should cleanup this categories and have it standardized in another file or enums instead of being free form strings as a followup. We should have it defined as strings DEVICE_ERROR , DIRECTORY_NOT_EMPTY , FILE_EXISTS that type of error codes which is the standard way for error codes to be maintained. That way the other services need not worry about it being a sentence, it looks like a error code.

internal/fs/wrappers/monitoring.go
```diff
@@ -75,24 +75,17 @@ func init() {
 	}
 }
 
-// errStrAndCategory maps an error to an error string and an error category.
-// Uncommon errors are bucketed into categories to reduce the cardinality of the
-// error so that the metric is not rejected by Cloud Monarch.
-func errStrAndCategory(err error) (str string, category string) {
+// categorize maps an error to an error-category.
+// This helps reduce the cardinality of the labels to less than 30.
```
charith87, line 79: Do we still log the particular error subtype and category in Error Log?

sethiay (approved):

internal/fs/wrappers/monitoring.go
kislaykishore: It doesn't affect logging. We don't log the category in the error logs though and I guess, logging it might not be necessary as well since we usually have the errors themselves in the logs.

internal/fs/wrappers/monitoring.go
kislaykishore: Yeah, I will make this change in a follow-up PR.

charith87 (approved):

PR #2390, opened 2024-08-23
Change category name to look more like codes

kislaykishore:
Currently, the category names are phrases/sentences. Keeping it as codes makes them more succinct and easily searchable.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/wrappers/monitoring.go
+++ b/internal/fs/wrappers/monitoring.go
@@ -48,7 +48,7 @@ var (
 
 // Error categories
 const (
-	deviceErr         = "DEVICE_ERRORS"
+	deviceErr         = "DEVICE_ERROR"
 	dirNotEmptyErr    = "DIR_NOT_EMPTY"
 	fileExistsErr     = "FILE_EXISTS"
 	fileDirErr        = "FILE_DIR_ERROR"
--- a/internal/fs/wrappers/monitoring.go
+++ b/internal/fs/wrappers/monitoring.go
@@ -48,22 +48,22 @@ var (
 
 // Error categories
 const (
-	deviceErr         = "DEVICE_ERROR"
-	dirNotEmptyErr    = "DIR_NOT_EMPTY"
-	fileExistsErr     = "FILE_EXISTS"
-	fileDirErr        = "FILE_DIR_ERROR"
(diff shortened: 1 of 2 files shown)


internal/fs/wrappers/monitoring.go
```diff
@@ -46,6 +46,26 @@ var (
 	tracer = otel.Tracer(name)
 )
 
+// Error categories
+const (
+	deviceErr         = "DEVICE_ERRORS"
```
charith87, line 51: Could we directly use "DEVICE_ERRORS" or "FILE_EXISTS" instead of deviceErr, dirNotEmptyErr

kislaykishore: It goes against the styleguide: https://google.github.io/styleguide/go/decisions#constant-names

charith87: Ohh Thanks Kislay for this. 
I was thinking how to make this more readable. Could we prefix this with err instead brings a bit of consistency to make it more readable.
Could we then keep it errDevice, errFileNotFound, errUnimplemented, etc.

charith87: this is more of a nit at this point. Either way upto you if you want to take it or keep it same.

internal/fs/wrappers/monitoring.go
```diff
@@ -46,6 +46,26 @@ var (
...
 
+// Error categories
+const (
+	deviceErr         = "DEVICE_ERROR"
+	dirNotEmptyErr    = "DIR_NOT_EMPTY"
+	fileExistsErr     = "FILE_EXISTS"
+	fileDirErr        = "FILE_DIR_ERROR"
+	notImplementedErr = "NOT_IMPLEMENTED"
+	ioErr             = "IO_ERROR"
+	intrErr           = "INTERRUPT_ERROR"
```
charith87, line 57: we can keep the variable names normal instead of shortened -> intrErr = interruptError or nwErr as networkError

charith87 (approved):
Approving.

hime (approved):

charith87 (approved):

PR #2392, opened 2024-08-23
Fix a bug in how read cache e2e tests are run for different configs.

sethiay:
### Description
Fix a bug in how read cache e2e tests are run for different configs - Without this fix, if same file name is put at the time of creation of config file and there are multiple places in the same test where we are creating config file, then only the last config is run (because the creation function overwrites the config file).

In this fix, we create config file just before running test for that config, so even if there are overwrites to config file, there is no problem.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/read_only_test.go
+++ b/tools/integration_tests/read_cache/read_only_test.go
@@ -163,15 +163,48 @@ func TestReadOnlyTest(t *testing.T) {
 	}
 
 	// Define flag set to run the tests.
-	flagsSet := [][]string{
-		{"--implicit-dirs", "--config-file=" + createConfigFile(cacheCapacityInMB, true, configFileName, false, getDefaultCacheDirPathForTests())},
-		{"--config-file=" + createConfigFile(cacheCapacityInMB, false, configFileNameForParallelDownloadTests, true, getDefaultCacheDirPathForTests())},
+	flagsSet := []gcsfuseTestFlags{
+		{
+			cliFlags:                []string{"--implicit-dirs"},
+			cacheSize:               cacheCapacityInMB,
+			cacheFileForRangeRead:   true,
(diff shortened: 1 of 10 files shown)


tools/integration_tests/read_cache/read_only_test.go
```diff
@@ -163,15 +163,48 @@ func TestReadOnlyTest(t *testing.T) {
...
+			cacheSize:               cacheCapacityInMB,
+			cacheFileForRangeRead:   true,
+			fileName:                configFileName,
+			enableParallelDownloads: false,
+			cacheDirPath:            getDefaultCacheDirPathForTests(),
+		},
+		{
+			cliFlags:                nil,
+			cacheSize:               cacheCapacityInMB,
+			cacheFileForRangeRead:   true,
```
ashmeenkaur, line 178: This should be false, right?

tools/integration_tests/read_cache/read_only_test.go
```diff
@@ -163,15 +163,48 @@ func TestReadOnlyTest(t *testing.T) {
...
+			fileName:                configFileName,
+			enableParallelDownloads: false,
+			cacheDirPath:            getDefaultCacheDirPathForTests(),
+		},
+		{
+			cliFlags:                []string{"--o=ro"},
+			cacheSize:               cacheCapacityInMB,
+			cacheFileForRangeRead:   true,
```
ashmeenkaur, line 194: same here

tools/integration_tests/read_cache/read_only_test.go
sethiay: yes, good catch. fixed.

ashmeenkaur (approved):

PR #2414, opened 2024-08-28
Wire the new configuration into mount pipeline.

kislaykishore:
* Trigger the mount pipeline when Viper-based new configuration is enabled.
* Create a new function type to represent the type of the mount function.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


main.go
```diff
@@ -70,7 +69,7 @@ func main() {
 	log.SetFlags(log.Ldate | log.Ltime | log.Lmicroseconds)
 	if strings.ToLower(os.Getenv("ENABLE_GCSFUSE_VIPER_CONFIG")) == "true" {
 		// TODO: implement the mount logic instead of simply returning nil.
-		rootCmd, err := cmd.NewRootCmd(func(*cfg.Config, string, string) error { return nil })
+		rootCmd, err := cmd.NewRootCmd(cmd.Mount)
```
ashmeenkaur, line 72: Should we move profiling handlers to this main now: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/cmd/legacy_main.go#L481-L483

kislaykishore: Good catch!

ashmeenkaur (approved):

PR #2419, opened 2024-08-29
Map apiError code with Fuse error

Tulsishah:
### Description
Previously, we were converting errors by performing string comparisons, which is not ideal. String representations can change, potentially leading to unnecessary code failures. Instead, we should check the apiError code directly and convert it into the corresponding FUSE error.

Below is the mapping of all the API error-codes to FUSE errors in GCSFuse:

```
OK --> ok
Canceled --> interruptible system calls
Unknown --> input-output error
InvalidArgument --> input-output error
DeadlineExceeded --> input-output error
NotFound --> file/dir not found
AlreadyExists --> input-output error
PermissionDenied --> permission denied error
ResourceExhausted --> input-output error
FailedPrecondition --> input-output error
Aborted --> input-output error
OutOfRange --> input-output error
Unimplemented --> input-output error
Internal --> input-output error
Unavailable --> input-output error
DataLoss --> input-output error
Unauthenticated --> permission denied error
```

NA

### Testing details
1. Manual - NA
2. Unit tests - Added


internal/fs/wrappers/error_mapping.go
```diff
@@ -59,16 +60,28 @@ func errno(err error) error {
...
-	// The control client API returns an RPC error code instead of googleapi code.
-	if strings.Contains(err.Error(), "oauth2: cannot fetch token") || strings.Contains(err.Error(), "PermissionDenied") {
+	if strings.Contains(err.Error(), "oauth2: cannot fetch token") {
 		return syscall.EACCES
 	}
 
-	// Translate API errors into an em errno
-	var apiErr *googleapi.Error
+	// The control client API returns an RPC error code instead of googleapi code.
+	var apiErr *apierror.APIError
```
vadlakondaswetha, line 68: apiError has a method FromError(). can you use that to write unit tests?

internal/fs/wrappers/error_mapping.go
```diff
@@ -59,16 +60,28 @@ func errno(err error) error {
...
 	if errors.As(err, &apiErr) {
-		switch apiErr.Code {
-		case http.StatusForbidden:
+		switch apiErr.GRPCStatus().Code() {
```
vadlakondaswetha, line 70: can you add a comment why only grpc statuc code is checked here. There is also googleapi error in the apierr definition.

internal/fs/wrappers/error_mapping.go
Tulsishah: FromError parses a Status error or a googleapi.Error. RPC errors require a Status error. However, Status has private attributes that cannot be assigned from outside its package. This prevents us from creating a Status error, and thus we cannot use FromError to create an apiError.

internal/fs/wrappers/error_mapping.go
Tulsishah: Done

internal/fs/wrappers/error_mapping.go
Tulsishah: Done

internal/fs/wrappers/error_mapping.go
```diff
@@ -59,16 +60,29 @@ func errno(err error) error {
...
-	// The control client API returns an RPC error code instead of googleapi code.
-	if strings.Contains(err.Error(), "oauth2: cannot fetch token") || strings.Contains(err.Error(), "PermissionDenied") {
+	if strings.Contains(err.Error(), "oauth2: cannot fetch token") {
 		return syscall.EACCES
 	}
 
-	// Translate API errors into an em errno
-	var apiErr *googleapi.Error
+	// The control client API returns an RPC error code instead of googleapi code.
+	// Currently, we only have gRPC control client APIs, so we are checking the gRPC status code.
```
vadlakondaswetha, line 68: control client apis are only for hierarchial buckets? for regular buckets we dont have control client apis or are those http apis?

vadlakondaswetha (approved):

internal/fs/wrappers/error_mapping.go
Tulsishah: control client apis are only for hierarchial buckets? --> Yes 
For regular buckets we dont have control client apis or are those http apis? --> Http/Grpc storage client APIs.

PR #2421, opened 2024-08-29
Increase timeout of test runs

kislaykishore:
go tests have a timeout of 10 minutes by default. Since the tests are run several times in this workflow, we need to increase the timeout accordingly. Otherwise, the tests will time-out after 10 minutes.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/workflows/flake-detector.yml
+++ b/.github/workflows/flake-detector.yml
@@ -32,10 +32,10 @@ jobs:
         run: go mod download
 
       - name: Test all except caching parallely
-        run: CGO_ENABLED=0 go test  -count 20 -skip `cat flaky_tests.lst | go run tools/scripts/skip_tests/main.go` `go list ./... | grep -v internal/cache/...`
+        run: CGO_ENABLED=0 go test -timeout 75m -count 20 -skip `cat flaky_tests.lst | go run tools/scripts/skip_tests/main.go` `go list ./... | grep -v internal/cache/...`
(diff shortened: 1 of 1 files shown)


.github/workflows/flake-detector.yml
```diff
@@ -32,10 +32,10 @@ jobs:
         run: go mod download
 
       - name: Test all except caching parallely
-        run: CGO_ENABLED=0 go test  -count 20 -skip `cat flaky_tests.lst | go run tools/scripts/skip_tests/main.go` `go list ./... | grep -v internal/cache/...`
+        run: CGO_ENABLED=0 go test -timeout 75m -count 20 -skip `cat flaky_tests.lst | go run tools/scripts/skip_tests/main.go` `go list ./... | grep -v internal/cache/...`
```
sethiay, line 35: Does it really take 75m ? 
(Ideally, it should be %20 or similar higher than the actual run time, so that developers try to parallelize/optimize when writing new tests to fit into that run time and be conscious of increasing it)

kislaykishore: I'd like to increase the number of iterations as much as possibledv cv v cxvcxvxczvz
++  3 that It doesn't really. But here, we are already enforcing a top-level timeout.

sethiay (approved):

PR #2439, opened 2024-09-03
Adding flag to ls perf script to make 1m files listing optional

anushka567:
### Description
HNS and flat bucket in use for the listing benchmarks script contains test folder with 1m files. Making that a one time manual run only since Kokoro periodic test job times out due to this case.

To run list perf test on folder with 1m files, run the listing benchmark script with run_1m_test flag.
NA

### Testing details
1. Manual - Ran the script with/without the added flag , working as intended.
2. Unit tests - Added unit tests to check both scenarios when run_1m_test is set to true and false.
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/ls_metrics/listing_benchmark_test.py
+++ b/perfmetrics/scripts/ls_metrics/listing_benchmark_test.py
@@ -128,6 +128,28 @@ DIRECTORY_STRUCTURE3 = {
     ]
 }
 
+DIRECTORY_STRUCTURE4 = {
+    'name': 'fake_bucket',
+    'num_folders': 2,
+    'num_files': 1,
+    'file_size': '1kb',
+    'file_name_prefix': 'file',
+    'folders': [
+        {
+            'name': '2KB_3files_0subdir',
+            'num_files': 3,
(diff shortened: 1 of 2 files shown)


perfmetrics/scripts/ls_metrics/listing_benchmark_test.py
```diff
@@ -573,6 +597,39 @@ def test_compare_directory_structure_false_file_folder_multi_level_dir(
...
+
+    self.assertEqual(gcs_bucket_results, persistent_disk_results)
+    mock_record_time_of_operation.assert_called_once_with('ls -R','fake_bucket/2KB_3files_0subdir',2)
+    self.assertEqual(gcs_bucket_results, {
+        '2KB_3files_0subdir': [1, 1],
+    })
+
+  @patch('listing_benchmark.RUN_1M_TEST',True)
+  @patch('listing_benchmark._record_time_of_operation', return_value=[1])
+  def test_skip_1m_test_true(self, mock_record_time_of_operation):
```
Tulsishah, line 616: Change test name. Currently both are same

Tulsishah (approved):

PR #2442, opened 2024-09-04
replacing gcloud command with storage client to UploadGcsObject

vipnydav:
### Description
Replaced gcloud command with storage client to UploadGcsObject as gcloud commands can't be run concurrently.

NA

### Testing details
1. Manual - Done
2. Unit tests - NA


tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +223,59 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+		}
+	} else {
+		// Copy the file contents directly to the object writer (no compression)
+		if _, err := io.Copy(w, f); err != nil {
+			return fmt.Errorf("failed to copy file to object: %v", err)
+		}
+	}
+
+	// Close the writer to finalize the upload
+	if err := w.Close(); err != nil {
```
gargnitingoogle, line 275: Instead of having this close statement here, I'd suggest moving this block as a deferred block right after the `w := obj.NewWriter(ctx)` above.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +223,59 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
 	}
 	return attrs, nil
 }
+
+func UploadGcsObject(localPath, bucketName, objectName string, uploadGzipEncoded bool) error {
+	ctx := context.Background()
+
+	client, err := storage.NewClient(ctx)
```
gargnitingoogle, line 230: Instead of creating a new storage client here each time, pass the context and the client in the function just like in all the other utilities above.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +223,59 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
 	}
 	return attrs, nil
 }
+
+func UploadGcsObject(localPath, bucketName, objectName string, uploadGzipEncoded bool) error {
+	ctx := context.Background()
```
gargnitingoogle, line 228: Pass context into this function for better control, rather than using context.Background here.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +223,59 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create storage client: %v", err)
+	}
+	defer client.Close()
+
+	// Open the local file for reading
+	f, err := os.Open(localPath)
```
gargnitingoogle, line 237: Use https://github.com/GoogleCloudPlatform/gcsfuse/blob/8aea5961f83c74fdbf5d4326d585c96d10951dad/tools/integration_tests/util/operations/file_operations.go#L327 here (you'd need to also make that method public in the file_operations.go for reuse here).

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +223,59 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+
+	// Create a writer to upload the object
+	obj := client.Bucket(bucketName).Object(objectName)
+	w := obj.NewWriter(ctx)
+
+	// Set content encoding if gzip compression is needed
+	if uploadGzipEncoded {
+
+		// Create a gzip writer on top of the object writer
+		gw := gzip.NewWriter(w)
```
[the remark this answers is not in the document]
gargnitingoogle: See if you can directly reuse that function here (or if needed, move that utility from there to util/file_operations.go for reuse here) rather than duplicating the logic here, which is always reason for double the maintenance.

gargnitingoogle (changes requested):

tools/integration_tests/util/client/storage_client.go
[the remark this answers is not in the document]
gargnitingoogle: > I am not sure that gzip.NewWriter would work for an object-writer created from a GCS client

If it does work, then you can leave it as it is. Still see if you can merge the two utilities int o one or at least reduce the logic duplication.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +223,59 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+		return fmt.Errorf("failed to create storage client: %v", err)
+	}
+	defer client.Close()
+
+	// Open the local file for reading
+	f, err := os.Open(localPath)
+	if err != nil {
+		return fmt.Errorf("failed to open local file: %v", err)
+	}
+	defer f.Close()
```
gargnitingoogle, line 241: check the output error of this f.close .

gargnitingoogle (changes requested):

tools/integration_tests/gzip/gzip_test.go
```diff
@@ -132,22 +134,28 @@ func setup_testdata(m *testing.M) error {
 	}
 
 	for _, fmd := range fmds {
-		var localFilePath string
-		localFilePath, err := helpers.CreateLocalTempFile(fmd.filesize, fmd.enableGzipEncodedContent)
+		content, err := helpers.CreateDataOfSize(fmd.filesize)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to create data: %w", err)
```
gargnitingoogle, line 139:
```suggestion
			return fmt.Errorf("failed to create content for testing: %w", err)
```

tools/integration_tests/gzip/gzip_test.go
```diff
@@ -132,22 +134,28 @@ func setup_testdata(m *testing.M) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to create local file: %w", err)
 		}
 
 		defer os.Remove(localFilePath)
 
 		// upload to the test-bucket for testing
-		gcsObjectPath := path.Join(setup.TestBucket(), TestBucketPrefixPath, fmd.filename)
+		objectPrefixPath := path.Join(TestBucketPrefixPath, fmd.filename)
+
```
gargnitingoogle, line 150: remove the unnecessary blank line

tools/integration_tests/gzip/gzip_test.go
```diff
@@ -132,22 +134,28 @@ func setup_testdata(m *testing.M) error {
...
 
 	for _, fmd := range fmds {
-		var localFilePath string
-		localFilePath, err := helpers.CreateLocalTempFile(fmd.filesize, fmd.enableGzipEncodedContent)
+		content, err := helpers.CreateDataOfSize(fmd.filesize)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to create data: %w", err)
+		}
+		localFilePath, err := operations.CreateLocalTempFile(content, false)
```
gargnitingoogle, line 141: Why is the second argument hard-coded to false ?

gargnitingoogle: Shouldn't it be fmd.enableGzipEncodedContent ?

tools/integration_tests/gzip/helpers/helpers.go
```diff
@@ -131,7 +82,11 @@ func DownloadGzipGcsObjectAsCompressed(bucketName, objPathInBucket string) (stri
 		return "", fmt.Errorf("failed to get size of gcs object %s: %w", gcsObjectPath, err)
 	}
 
-	tempfile, err := CreateLocalTempFile(1, false)
+	content, err := CreateDataOfSize(1)
+	if err != nil {
+		return "", fmt.Errorf("failed to create data: %w", err)
+	}
+	tempfile, err := operations.CreateLocalTempFile(content, false)
```
gargnitingoogle, line 89: I see a disk-leak issue here. We create and write this file and return if the creation and writing succeed, but we don't delete this file if creation succeeds but the writing fails. Agree that this is a pre-existing issue, but we should fix it now that we have found it.

tools/integration_tests/gzip/write_gzip_test.go
```diff
@@ -50,12 +50,17 @@ func verifyFullFileOverwrite(t *testing.T, filename string) {
 			mountedFilePath, (*fi).Size(), gcsObjectPath, gcsObjectSize)
 	}
 
+	content, err := helpers.CreateDataOfSize(OverwittenFileSize)
```
gargnitingoogle, line 53: Pre-existing typo. OverwittenFileSize . Please fix.

```suggestion
	content, err := helpers.CreateDataOfSize(OverwrittenFileSize)
```

gargnitingoogle (changes requested):
added some more comments

tools/integration_tests/gzip/helpers/helpers.go
```diff
@@ -31,20 +30,9 @@ import (
...
-// contentSize is the size of the uncompressed content. In case gzipCompress is true, the actual output file size will be
-// different from contentSize (typically gzip-compressed file size < contentSize).
-// Caller is responsible for deleting the created file when done using it.
-// Failure cases:
-// 1. contentSize <= 0
-// 2. os.CreateTemp() returned error or nil handle
-// 3. gzip.NewWriter() returned nil handle
-// 4. Failed to write the content to the created temp file
-func CreateLocalTempFile(contentSize int, gzipCompress bool) (string, error) {
+func CreateDataOfSize(contentSize int) (string, error) {
```
gargnitingoogle, line 35: This function is only used within this package (gzip), so it doesn't need to be public, so it should start with small letter. Yes, it was wrong even earlier I think.

```suggestion
func createDataOfSize(contentSize int) (string, error) {
```

tools/integration_tests/gzip/helpers/helpers.go
```diff
@@ -124,55 +75,75 @@ func CreateLocalTempFile(contentSize int, gzipCompress bool) (string, error) {
 // Uses go storage client library to download object. Use of gsutil/gcloud is not
 // possible as they both always read back objects with content-encoding: gzip as
 // uncompressed/decompressed irrespective of any argument passed.
-func DownloadGzipGcsObjectAsCompressed(bucketName, objPathInBucket string) (string, error) {
+func DownloadGzipGcsObjectAsCompressed(bucketName, objPathInBucket string) (tempfile string, err error) {
```
gargnitingoogle, line 78: Same as above.

```suggestion
func downloadGzipGcsObjectAsCompressed(bucketName, objPathInBucket string) (tempfile string, err error) {
```

tools/integration_tests/gzip/helpers/helpers.go
gargnitingoogle: If it needs to be public, then it must have heaader.

```suggestion
// CreateDataOfSize ...
func CreateDataOfSize(contentSize int) (string, error) {
```

tools/integration_tests/gzip/helpers/helpers.go
gargnitingoogle: Or

```suggestion
// DownloadGzipGcsObjectAsCompressed ...
func DownloadGzipGcsObjectAsCompressed(bucketName, objPathInBucket string) (tempfile string, err error) {
```

tools/integration_tests/gzip/helpers/helpers.go
```diff
@@ -124,55 +75,75 @@ func CreateLocalTempFile(contentSize int, gzipCompress bool) (string, error) {
...
+	}
+	tempfile, err = operations.CreateLocalTempFile(content, false)
 	if err != nil {
-		return "", fmt.Errorf("failed to create tempfile for downloading gcs object: %w", err)
+		err = fmt.Errorf("failed to create tempfile for downloading gcs object: %w", err)
+		return
 	}
+	defer func() {
+		if err != nil {
+			os.Remove(tempfile)
```
gargnitingoogle, line 99: os.Remove itself can fail (https://pkg.go.dev/os#Remove), and must be checked.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +222,53 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
 	}
 	return attrs, nil
 }
+
+// UploadGcsObject uploads a local file to a specified GCS bucket and object.
+// Handles gzip compression if requested.
+func UploadGcsObject(ctx context.Context, localPath, bucketName, objectName string, uploadGzipEncoded bool) error {
+	client, err := storage.NewClient(ctx)
+
```
gargnitingoogle, line 230: Remove the unnecessary blank line

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +222,53 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+	if err != nil {
+		return fmt.Errorf("failed to create storage client: %v", err)
+	}
+	defer client.Close()
+
+	// Create a writer to upload the object.
+	obj := client.Bucket(bucketName).Object(objectName)
+	w := obj.NewWriter(ctx)
+	defer func() {
+		err := w.Close()
```
gargnitingoogle, line 240: This function absorbs any existing non-nil err value at this point, thus hiding error, so this shouldn't even bother doing anything if `err != nil` already.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +222,53 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+		if err != nil {
+			return err
+		}
+
+		content := string(data)
+		filePathToUpload, err = operations.CreateLocalTempFile(content, true)
+		if err != nil {
+			return fmt.Errorf("failed to create local gzip file from %s for upload to bucket: %w", localPath, err)
+		}
+		defer os.Remove(filePathToUpload)
```
gargnitingoogle, line 259: Again, as earlier os.Remove can fail and should be handled.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,60 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+	filenameTemplate := "testfile-*.txt"
+	if gzipCompress {
+		filenameTemplate += ".gz"
+	}
+
+	// create a temp file
+	f, err := os.CreateTemp(TmpDirectory, filenameTemplate)
+	if err != nil {
+		return "", fmt.Errorf("failed to create tempfile for template %s: %w", filenameTemplate, err)
+	} else if f == nil {
```
gargnitingoogle, line 656: `else` is not needed

```suggestion
	} if f == nil {
```

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,60 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+	defer CloseFile(f)
+	filepath := f.Name()
+
+	if gzipCompress {
+		w := gzip.NewWriter(f)
+		if w == nil {
+			return "", fmt.Errorf("failed to create gzip writer for file %s", filepath)
+		}
+		defer func() {
+			if err := w.Close(); err != nil {
```
gargnitingoogle, line 668: It absorbs an existing non-nil value of err.

gargnitingoogle (changes requested):
Added some comments. Needs some fixes.

tools/integration_tests/gzip/helpers/helpers.go
vipnydav: This is used in both packages helpers and gzip_test.

vipnydav: Got it will move both methods CreateDataOfSize and DownloadGzipGcsObjectAsCompressed to gzip_test

tools/integration_tests/util/client/storage_client.go
vipnydav: a new variable named err is being created here as we are using err:=w.close(), hence it should not absorb any existing non-nil err.

tools/integration_tests/util/operations/file_operations.go
vipnydav: here a new variable err is created in scope of if statement. hence it won't absorb existing value of err.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,61 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+				fmt.Printf("failed to close gzip writer for file %s: %v", filepath, err)
+			}
+		}()
+
+		// write the content created above as gzip
+		n, err := w.Write([]byte(content))
+		if err != nil {
+			return "", fmt.Errorf("failed to write content to %s using gzip-writer: %w", filepath, err)
+		} else if n != contentSize {
```
gargnitingoogle, line 678: nit: my bad, I missed it earlier. `else` not required.

```suggestion
		} 
		if n != contentSize {
```

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,61 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+			return "", fmt.Errorf("failed to write to gzip file %s. Content-size: %d bytes, wrote = %d bytes", filepath, contentSize, n)
+		}
+	} else {
+		// write the content created above as text
+		n, err := f.WriteString(content)
+		if err != nil {
+			return "", err
+		} else if n != contentSize {
```
gargnitingoogle, line 686: Same here.

```suggestion
		}
		if n != contentSize {
```

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +222,57 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+	}
+	defer client.Close()
+
+	// Create a writer to upload the object.
+	obj := client.Bucket(bucketName).Object(objectName)
+	w := obj.NewWriter(ctx)
+	defer func() {
+		err := w.Close()
+		if err != nil {
+			fmt.Printf("Failed to close GCS object gs://%s/%s: %v", bucketName, objectName, err)
```
gargnitingoogle, line 241: nit: ideally we should return this error and not just print it.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,61 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+	filepath := f.Name()
+
+	if gzipCompress {
+		w := gzip.NewWriter(f)
+		if w == nil {
+			return "", fmt.Errorf("failed to create gzip writer for file %s", filepath)
+		}
+		defer func() {
+			if err := w.Close(); err != nil {
+				fmt.Printf("failed to close gzip writer for file %s: %v", filepath, err)
```
gargnitingoogle, line 670: nit: ideally we should return this error and not just print it.

tools/integration_tests/gzip/gzip_test.go
```diff
@@ -172,6 +180,46 @@ func destroy_testdata(m *testing.M) error {
 	return nil
 }
 
+// createDataOfSize generates a string of the specified content size in bytes.
+// Failure cases:
+// 1. contentSize <= 0
+func createDataOfSize(contentSize int) (string, error) {
```
gargnitingoogle, line 186: nit: for consistency

```suggestion
func createContentOfSize(contentSize int) (string, error) {
```

gargnitingoogle (dismissed):
Added some nitpick comments, but otherwise LGTM

tools/integration_tests/util/client/storage_client.go
vipnydav: We can't return anything in defer function and also returning a non nil error will result in test failure but not closing writer should not result in test failure.

tools/integration_tests/util/operations/file_operations.go
vipnydav: We can't return anything in defer function and also returning a non nil error will result in test failure but not closing writer should not result in test failure.

tools/integration_tests/util/client/storage_client.go
gargnitingoogle: > We can't return anything in defer function

The way to return a value in defer function is to set a named returned value in it.

> also returning a non nil error will result in test failure but not closing writer should not result in test failure

Why is that ?

tools/integration_tests/util/operations/file_operations.go
gargnitingoogle: see my point at https://github.com/GoogleCloudPlatform/gcsfuse/pull/2442#discussion_r1760472483

gargnitingoogle (approved):

tools/integration_tests/util/operations/file_operations.go
vipnydav: Resolving this as per discussion.

tools/integration_tests/util/client/storage_client.go
vipnydav: Resolving this as per discussion.

tools/integration_tests/gzip/gzip_test.go
```diff
@@ -172,9 +183,62 @@ func destroy_testdata(m *testing.M) error {
...
+	// fail if contentSize <= 0
+	if contentSize <= 0 {
+		return "", fmt.Errorf("unsupported fileSize: %d", contentSize)
+	}
+
+	// Create text-content of given size.
+	// strings.builder is used as opposed to string appends
+	// as this is much more efficient when multiple concatenations
+	// are required.
+	var contentBuilder strings.Builder
```
kislaykishore, line 199: nit: you know the size of the string that will be stored in contentBuilder. So, you can call contentBuilder.Grow(contentSize) to pre-allocate the necessary space.

tools/integration_tests/gzip/read_gzip_test.go
```diff
@@ -108,6 +112,79 @@ func verifyRangedRead(t *testing.T, filename string) {
...
+	}
+
+	content, err := createContentOfSize(1)
+	if err != nil {
+		return "", fmt.Errorf("failed to create data: %w", err)
+	}
+	tempfile, err = operations.CreateLocalTempFile(content, false)
+	if err != nil {
+		return "", fmt.Errorf("failed to create tempfile for downloading gcs object: %w", err)
+	}
```
kislaykishore, line 136: Nit: Use short-if
```suggestion
	if tempfile, err = operations.CreateLocalTempFile(content, false); err != nil {
		return "", fmt.Errorf("failed to create tempfile for downloading gcs object: %w", err)
	}
```

tools/integration_tests/gzip/read_gzip_test.go
```diff
@@ -108,6 +112,79 @@ func verifyRangedRead(t *testing.T, filename string) {
...
+	defer func() {
+		if err != nil {
+			removeErr := os.Remove(tempfile)
+			if removeErr != nil {
+				t.Logf("Error removing temporary file %s: %v", tempfile, removeErr)
+			}
+		}
+	}()
```
kislaykishore, line 144: This defer call will get executed when the method `downloadGzipGcsObjectAsCompressed` returns. At that time, if the value of err is not nil, then this code will not attempt to delete the file. I don't think that's intended. 
Do we really need to check if err != nil? If err != nil, then the defer statement would be unreachable. So, how about simply removing the file without checking the value of err?

tools/integration_tests/gzip/read_gzip_test.go
```diff
@@ -108,6 +112,79 @@ func verifyRangedRead(t *testing.T, filename string) {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create data: %w", err)
+	}
+	tempfile, err = operations.CreateLocalTempFile(content, false)
+	if err != nil {
+		return "", fmt.Errorf("failed to create tempfile for downloading gcs object: %w", err)
+	}
+	defer func() {
+		if err != nil {
+			removeErr := os.Remove(tempfile)
```
kislaykishore, line 139: I don't get the point of removing the tempfile after creating it here. How is it being used since even though the tempfile is returned, it'd get removed by the time the method returns.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +222,51 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+// Handles gzip compression if requested.
+func UploadGcsObject(ctx context.Context, client *storage.Client, localPath, bucketName, objectName string, uploadGzipEncoded bool) error {
+	// Create a writer to upload the object.
+	obj := client.Bucket(bucketName).Object(objectName)
+	w := obj.NewWriter(ctx)
+	defer func() {
+		err := w.Close()
+		if err != nil {
+			log.Printf("Failed to close GCS object gs://%s/%s: %v", bucketName, objectName, err)
+		}
```
kislaykishore, line 236: Nit
```suggestion
		if err := w.Close(); err != nil {
			log.Printf("Failed to close GCS object gs://%s/%s: %v", bucketName, objectName, err)
		}
```

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +222,51 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+		content := string(data)
+		filePathToUpload, err = operations.CreateLocalTempFile(content, true)
+		if err != nil {
+			return fmt.Errorf("failed to create local gzip file from %s for upload to bucket: %w", localPath, err)
+		}
+		defer func() {
+			removeErr := os.Remove(filePathToUpload)
+			if removeErr != nil {
+				log.Printf("Error removing temporary gzip file %s: %v", filePathToUpload, removeErr)
+			}
```
kislaykishore, line 256: Nit
```suggestion
			if removeErr := os.Remove(filePathToUpload); removeErr != nil {
				log.Printf("Error removing temporary gzip file %s: %v", filePathToUpload, removeErr)
			}
```

tools/integration_tests/util/client/storage_client.go
```diff
@@ -222,3 +222,51 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
...
+		data, err := os.ReadFile(localPath)
+		if err != nil {
+			return err
+		}
+
+		content := string(data)
+		filePathToUpload, err = operations.CreateLocalTempFile(content, true)
+		if err != nil {
+			return fmt.Errorf("failed to create local gzip file from %s for upload to bucket: %w", localPath, err)
+		}
```
kislaykishore, line 251: Nit:
```suggestion
		if filePathToUpload, err = operations.CreateLocalTempFile(content, true); err != nil {
			return fmt.Errorf("failed to create local gzip file from %s for upload to bucket: %w", localPath, err)
		}
```

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,63 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+		if err != nil {
+			return "", err
+		}
+		if n != contentSize {
+			return "", fmt.Errorf("failed to write to text file %s. Content-size: %d bytes, wrote = %d bytes", filepath, contentSize, n)
+		}
+	}
+
+	return filepath, nil
+}
```
kislaykishore, line 694: Reduce nesting: https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

kislaykishore (changes requested):

tools/integration_tests/gzip/read_gzip_test.go
vipnydav: Tempfile will be returned only when returned err==nil and tempfile will be removed only when returned err!=nil. We are doing this to avoid disk leak issue. You can refer to this comment https://github.com/GoogleCloudPlatform/gcsfuse/pull/2442#discussion_r1750041834.

tools/integration_tests/gzip/read_gzip_test.go
vipnydav: Tempfile will be returned only when returned err==nil and tempfile will be removed only when returned err!=nil. We are doing this to avoid disk leak issue. You can refer to this comment https://github.com/GoogleCloudPlatform/gcsfuse/pull/2442#discussion_r1750041834.

tools/integration_tests/util/operations/file_operations.go
vipnydav: Would it be better to implement methods writeGzipToFile and writeTextToFile
 and make code something like this?

 if gzipCompress {
		return writeGzipToFile(f, filepath, content, contentSize)
	}

	return writeTextToFile(f, filepath, content, contentSize)
	
or it is good without these methods too?

kislaykishore: That'd be great!

tools/integration_tests/gzip/gzip_test.go
```diff
@@ -172,9 +179,62 @@ func destroy_testdata(m *testing.M) error {
...
+
+	if contentSize > 0 {
+		err := writeString(&contentBuilder, tempStr[0:contentSize])
+		if err != nil {
+			return "", err
+		}
+	}
+
+	return contentBuilder.String(), nil
+}
```
kislaykishore, line 222: This seems a lot more concise and readable.
```suggestion
// createContentOfSize generates a string of the specified content size in bytes.
func createContentOfSize(contentSize int) (string, error) {
	if contentSize <= 0 {
		return "", fmt.Errorf("unsupported fileSize: %d", contentSize)
	}
        const tempStr = "This is a test file\n"
	iter := (contentSize+len(tempStr) - 1)/len(tempStr)
	str := strings.Repeat(tempStr, iter)
	return str[:contentSize], nil
}	
```

tools/integration_tests/gzip/read_gzip_test.go
kislaykishore: Makes sense.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,70 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+	}
+	defer CloseFile(f)
+	filepath := f.Name()
+
+	contentSize := len(content)
+	if gzipCompress {
+		return writeGzipToFile(f, filepath, content, contentSize)
+	}
+
+	return writeTextToFile(f, filepath, content, contentSize)
```
kislaykishore, line 700: More concise and less cognitive load.
```suggestion
        if gzipCompress {
		return writeGzipToFile(f, f.Name(), content, len(content))
	}

	return writeTextToFile(f, f.Name(), content, len(content))
```

kislaykishore (changes requested):

tools/integration_tests/util/operations/file_operations.go
kislaykishore, line 1: Please fix lint issues.

vipnydav: Fixed

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -642,3 +632,67 @@ func SizeOfFile(filepath string) (size int64, err error) {
...
+// 2. gzip.NewWriter() returned nil handle
+// 3. Failed to write the content to the created temp file
+func CreateLocalTempFile(content string, gzipCompress bool) (string, error) {
+	// create appropriate name template for temp file
+	filenameTemplate := "testfile-*.txt"
+	if gzipCompress {
+		filenameTemplate += ".gz"
+	}
+
+	// create a temp file
```
kislaykishore, line 684: [Avoid comments that just repeat what the code does](https://testing.googleblog.com/2017/07/code-health-to-comment-or-not-to-comment.html)
```suggestion
```

kislaykishore (dismissed):

kislaykishore (approved):

PR #2443, opened 2024-09-04
Fetching VM metrics relevant to rename operations

anushka567:
### Description
Currently, the vm_metrics script fetches metrics which include cpu metrics, received bytes metrics, ops error count, etc. For rename operations, we also need memory usage metrics, sent bytes and average load in terms of os threads in flight.

The functions from the updated script will further be used by renaming benchmark to get the necessary metrics.

NA

### Testing details
1. Manual - As a part of rename benchmark script run, the script was successfully retrieving correct metrics.
2. Unit tests - Added unit tests for the new metrics added. [Peak/mean memory usage, peak/mean sent bytes count, average load]
3. Integration tests - NA


ankitaluthra1 (dismissed):

perfmetrics/scripts/vm_metrics/vm_metrics.py
```diff
@@ -300,16 +365,20 @@ def fetch_metrics(self, start_time_sec, end_time_sec, instance, period, test_typ
...
 
     Returns:
+      list[[period end time, interval end time,CPU_UTI_PEAK, CPU_UTI_MEAN,
+      REC_BYTES_PEAK, REC_BYTES_MEAN,SENT_BYTES_PEAK,SENT_BYTES_MEAN,OPS_ERROR_COUNT,
+      MEMORY_USAGE_PEAK,MEMORY_USAGE_MEAN,LOAD_AVG_OS_THREADS_MEAN]] in case of rename
+
       list[[period end time, interval end time, CPU_UTI_PEAK, CPU_UTI_MEAN,
       REC_BYTES_PEAK, REC_BYTES_MEAN, READ_BYTES_COUNT, OPS_ERROR_COUNT,
-      OPS_MEAN_LATENCY]]
+      OPS_MEAN_LATENCY]] otherwise
```
Tulsishah, line 378: otherwise?

anushka567: When test type is rename, only then we are returning the metrics mentioned in the RENAME_METRICS_LIST, but all any other cases ,by default , we are returning this metrics mentioned in METRICS_LIST

Tulsishah (approved):

PR #2444, opened 2024-09-04
[CLI Config Parity] fix error to handle negative numbers in gcsfuse arguments

ashmeenkaur:
### Description
To maintain backward compatibility, we're adding a "-" prefix to all GCSFuse arguments, aligning with cobra/viper's use of POSIX-compliant flags (which require a "--" prefix). However, this led to a bug where negative numbers were also receiving the "-" prefix, as we hadn't explicitly accounted for this scenario. The fix involves implementing a check to bypass negative numbers when adding prefixes.

NA

### Testing details
1. Manual - Manually verified '--kernel-list-cache-ttl-secs -1` flag is working now.
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/cfg/config_util_test.go
+++ b/cfg/config_util_test.go
@@ -105,3 +105,25 @@ func Test_ListCacheTtlSecsToDuration_InvalidCall(t *testing.T) {
 	// Calling with invalid argument to trigger panic.
 	ListCacheTTLSecsToDuration(-3)
 }
+
+func TestIsNegativeNumber(t *testing.T) {
+	testCases := []struct {
+		name     string
+		input    string
+		expected bool
+	}{
+		{"negative_integer", "-123", true},
+		{"positive_integer", "456", false},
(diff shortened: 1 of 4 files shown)


ankitaluthra1 (approved):

cfg/config_util_test.go
```diff
@@ -105,3 +105,25 @@ func Test_ListCacheTtlSecsToDuration_InvalidCall(t *testing.T) {
 	// Calling with invalid argument to trigger panic.
 	ListCacheTTLSecsToDuration(-3)
 }
+
+func TestIsNegativeNumber(t *testing.T) {
```
sethiay, line 109: could you please also write test case for --flag=postitive & --flag=negative number ? Basically try with = instead of " ".

ashmeenkaur: Added to main_test.go. Please take another look.

sethiay (approved):

PR #2446, opened 2024-09-04
Add lock for erasing/inserting entries for folder

Tulsishah:
### Description
While inserting and removing folder entries from the cache, we did not implement a lock. Consequently, during concurrent operations tests, it was failing with the error `Concurrent operation on map iterating and erasing`.

To resolve this issue, implement a lock before performing any operation on the cache.

NA

### Testing details
1. Manual - Tested by running concurrent_operation package many times locally.
2. Unit tests - NA

### DIFF
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -85,6 +85,7 @@ func (b *fastStatBucket) insertMultiple(objs []*gcs.Object) {
 	}
 }
 
+// LOCKS_EXCLUDED(b.mu)
 func (b *fastStatBucket) eraseEntriesWithGivenPrefix(folderName string) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -34,10 +34,10 @@ import (
 // bucket. Records are invalidated when modifications are made through this
 // bucket, and after the supplied TTL.
 func NewFastStatBucket(
-		ttl time.Duration,
-		cache metadata.StatCache,
-		clock timeutil.Clock,
-		wrapped gcs.Bucket) (b gcs.Bucket) {
+	ttl time.Duration,
+	cache metadata.StatCache,
+	clock timeutil.Clock,
+	wrapped gcs.Bucket) (b gcs.Bucket) {
 	fsb := &fastStatBucket{
 		cache:   cache,
 		clock:   clock,
@@ -85,7 +85,7 @@ func (b *fastStatBucket) insertMultiple(objs []*gcs.Object) {
 	}
 }
 
-func (b *fastStatBucket) eraseEntriesWithGivenPrefixes(folderName string) {
+func (b *fastStatBucket) eraseEntriesWithGivenPrefix(folderName string) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -85,6 +85,14 @@ func (b *fastStatBucket) insertMultiple(objs []*gcs.Object) {
 	}
 }
 
+// LOCKS_EXCLUDED(b.mu)
+func (b *fastStatBucket) eraseEntriesWithGivenPrefix(folderName string) {
+	b.mu.Lock()
```
ankitaluthra1, line 90: is there any thread unsafe logic or race condition which we are avoiding by adding lock at this level? if its only for manipulation of lru cache then adding lock at lru level [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/1816e83c65021755b12870286bd6e047a7a31f80/internal/cache/lru/lru.go#L272) makes more sense. Adding a lock at higher level reduces performance and forces unnecessary synchronisation. This lock should be redundant if we add at lru level.

@raj-prince thoughts on this ?

ankitaluthra1 (changes requested):

internal/storage/caching/fast_stat_bucket.go
Tulsishah: I believe we need the lock at the bucket level instead of the LRU cache level for this. For instance, with dynamic bucket mounting and different bucket operations, we're okay to update the map.

Tulsishah: Also for erasing each entries we are already acquiring the lru lock.

ankitaluthra1: For instance, with dynamic bucket mounting and different bucket operations, we're okay to update the map. >> Are we maintaining different lru cache instancefor every bucket ?

ankitaluthra1: Also for erasing each entries we are already acquiring the lru lock. >> IIUC lock is needed per loop and not per entry, in erasing each entries we are acquiring the lock on cache per entry deletion [ref](https://github.com/GoogleCloudPlatform/gcsfuse/blob/47b45fe21553eb936fd2a5a96d78986449bfebfe/internal/cache/lru/lru.go#L188)

ankitaluthra1: @vadlakondaswetha Can you please suggest here, are we missing some use case here which requires bucket level locking ? or lru level lock should be good enough for erasing the entries from cache ?

vadlakondaswetha: if there is no bucket level locking, we need to think about how the concurrent operations work. For example, you are  running a loop to erase few entries and at the same time one entry matching the prefix is being read  or added by other flow. Is it even possible to happen something like this since we take locks at the directory level? can you confirm.

One more thing to check what happens when a new element is added to c.index while you are looping through the c.index elements. Atleast in java it fails. 

If something like that can't happen because we are taking dirInode locks before this it should fine to add a comment and remove the lock

ankitaluthra1: @Tulsishah discussed with @vadlakondaswetha offline, we need to add lock at lru level only with a common erase logic which does not include locking mechanism.

ankitaluthra1: Closing Comments: :ock on lru level is more optimised so we should add at that level, however with current fix all tests are working as expected, so can we avoid blocking this PR for now, as due to this issue HNS tests on CD pipeline are also blocked, locally on all VMs tests are passing with this fix.

The fix will need all this testing again as ntermittent failures were encountered earlier when we didn't take a bucket-level lock. @Tulsishah let's do this optimisation in next PR then.

ankitaluthra1 (approved):

PR #2470, opened 2024-09-11
Fix error return in tokenSrc 

Tulsishah:
### Description
- Made it more readable
- Fixed return  error
```
jwtConfig, err := google.JWTConfigFromJSON(contents, scope)
if err != nil {
	err = fmt.Errorf("JWTConfigFromJSON: %w", err)
	return nil, err
}
```
NA

### Testing details
1. Manual - Tested on TPC environment
2. Unit tests - NA

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -58,13 +58,14 @@ func newTokenSourceFromPath(ctx context.Context, path string, scope string) (oau
 		err = fmt.Errorf("JWTConfigFromJSON: %w", err)
 		return nil, err
 	}
-	// Create the token source.
-	ts := jwtConfig.TokenSource(ctx)
 
 	domain, err := getUniverseDomain(ctx, contents, scope)
 	if err != nil {
 		return nil, err
 	}
+
+	// By default, a standard OAuth 2.0 token source is created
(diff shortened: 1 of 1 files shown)


raj-prince (approved):

internal/auth/auth.go
```diff
@@ -43,30 +43,27 @@ func getUniverseDomain(ctx context.Context, contents []byte, scope string) (stri
...
 	// By default, a standard OAuth 2.0 token source is created
 	// Create a config struct based on its contents.
 	jwtConfig, err := google.JWTConfigFromJSON(contents, scope)
 	if err != nil {
 		err = fmt.Errorf("JWTConfigFromJSON: %w", err)
+		return nil, err
 	}
 	// Create the token source.
-	ts = jwtConfig.TokenSource(ctx)
+	ts := jwtConfig.TokenSource(ctx)
```
raj-prince, line 62: If we are not using `ts`, why are we creating here?

Tulsishah: We are using right. If it will not go in the if condition then it will return ts and err at the end.

raj-prince (approved):

PR #2478, opened 2024-09-13
Do away with the requirement of adding every flag in mount_gcsfuse

kislaykishore:
* Get the list of flags from the code-generated list and use that to automatically convert the persistent mount flags to appropriate format.

https://b.corp.google.com/issues/368503898

### Testing details
1. Manual - NA
2. Unit tests - Tested with existing unit tests.
3. Integration tests - NA

### DIFF
--- a/tools/mount_gcsfuse/main.go
+++ b/tools/mount_gcsfuse/main.go
@@ -106,6 +106,11 @@ func makeGcsfuseArgs(
 	}
 
 	boolFlags, nonBoolFlags := flagTypes(&flagSet)
+	// Handle "o" by not treating it as a flag for persistent mounting.
+	nonBoolFlags = slices.DeleteFunc(nonBoolFlags, func(s string) bool {
+		return s == "o"
+	})
+
 	nonBoolFlags = append(nonBoolFlags, cfg.ConfigFileFlagName)
 	// TODO: Clean this up after we gain enough confidence on CLI-Config Parity changes.
 	boolFlags = append(boolFlags, "disable-viper-config")
--- a/tools/mount_gcsfuse/main.go
+++ b/tools/mount_gcsfuse/main.go
@@ -106,6 +106,7 @@ func makeGcsfuseArgs(
 	}
 
 	boolFlags, nonBoolFlags := flagTypes(&flagSet)
+	nonBoolFlags = append(nonBoolFlags, "config-file")
 	noopOptions := []string{"user", "nouser", "auto", "noauto", "_netdev", "no_netdev"}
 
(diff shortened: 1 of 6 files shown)


tools/mount_gcsfuse/main.go
```diff
@@ -72,89 +100,35 @@ func makeGcsfuseArgs(
...
+
 	// Deal with options.
 	for name, value := range opts {
-		switch strings.ReplaceAll(name, "-", "_") {
-		// Don't pass through options that are relevant to mount(8) but not to
-		// gcsfuse, and that fusermount chokes on with "Invalid argument" on Linux.
-		case "user", "nouser", "auto", "noauto", "_netdev", "no_netdev":
-
-		// Special case: support mount-like formatting for gcsfuse bool flags.
-		case "implicit_dirs",
```
ashmeenkaur, line 83: Did you find any discrepancy between the flags supported in persistent mounting and the normal flags? Just want to make sure that there are no flags which should not be supported here.

kislaykishore: I don't think there is any value in doing that. For instance, "foreground" is something that doesn't really seem to make any sense here (correct me if I'm wrong), but it's already there.

ashmeenkaur (approved):

tools/mount_gcsfuse/main.go
kislaykishore: I've added "o" flag as an exception to ensure that its behavior is the same as before.

tools/mount_gcsfuse/main.go
```diff
@@ -72,89 +100,35 @@ func makeGcsfuseArgs(
 	device string,
 	mountPoint string,
 	opts map[string]string) (args []string, err error) {
+	var flagSet pflag.FlagSet
+	if err := cfg.BuildFlagSet(&flagSet); err != nil {
+		return nil, err
+	}
+
```
ashmeenkaur, line 107: Nit: remove empty line

tools/mount_gcsfuse/main.go
```diff
@@ -72,89 +100,40 @@ func makeGcsfuseArgs(
...
+	boolFlags, nonBoolFlags := flagTypes(&flagSet)
+	// Handle "o" by not treating it as a flag for persistent mounting.
+	nonBoolFlags = slices.DeleteFunc(nonBoolFlags, func(s string) bool {
+		return s == "o"
+	})
+
```
ashmeenkaur: Nit: remove empty line

ashmeenkaur (approved):

PR #2479, opened 2024-09-16
Rename flag

kislaykishore:
Rename log-rotate-max-log-file-size-mb to log-rotate-max-file-size-mb to make it more consistent with the corresponding name of the param in the config file.
The flag is not yet published/released. So, it shouldn't impact any users.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -529,9 +529,9 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.IntP("log-rotate-max-log-file-size-mb", "", 512, "The maximum size in megabytes that a log file can reach before it is rotated.")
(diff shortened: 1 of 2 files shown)


ankitaluthra1:
Since this is a breaking change, Can you please add confirmation in description no tests in GKE which might break due to change in value and also please check no backward compatability issue should occur due to hard coding in clients like GKE or vertex AI

ankitaluthra1 (approved):

kislaykishore:
> Since this is a breaking change, Can you please add confirmation in description no tests in GKE which might break due to change in value and also please add confirmation no backward compatability issue should occur due to hard coding in clients like GKE or vertex AI

This is a new flag. So, no breaking changes expected.

PR #2503, opened 2024-09-17
Catch crash logs in stderr log file for background process

Tulsishah:
### Description
**Background**
In a foreground process crash, we get both an error message and a crash stack trace. But for a background process, while the stack trace is lost, we can still retrieve the error message using a recovery function. This we are doing as well in our code.

During some testing, I encountered the "fatal error: concurrent map iteration and map write". The recovery function in Go is designed to handle panics triggered by panic() calls, not fatal errors like runtime crashes. The error I faced is a runtime error, not a panic. That's why I did not get this fatal error message in with background process. 

I've added a stderr file parameter to the daemonize.Run method to print error/crash logs to a file. [See the pull request for more details](https://github.com/jacobsa/daemonize/pull/2).

To utilize this, I'm updating the daemonize dependency to include the latest changes. Afterward, I have created a dump file if the user provides the log-file parameter(logFile.stderr) when running daemonize.Run.

User will able to check stderr dump on `logfile.stderr`.

NA

### Testing details
1. Manual - Manually checked `concurrent map update` crash and OOM crash. Checked log rotation scenario working fine.
2. Unit tests - NA


cmd/legacy_main.go
```diff
@@ -375,8 +375,15 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
 		// programme is running as daemon process.
 		env = append(env, fmt.Sprintf("%s=true", logger.GCSFuseInBackgroundMode))
 
+		var logFile *os.File
+		if newConfig.Logging.FilePath != "" {
```
kislaykishore, line 379: What happens if the log-file gets rotated by log-rotate? Will the file-handle continue to work?

cmd/legacy_main.go
```diff
@@ -375,8 +375,15 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
...
+			logFile, err = os.OpenFile(string(newConfig.Logging.FilePath), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
+			if err != nil {
+				return err
+			}
```
kislaykishore, line 383: nit: Prefer short-if
```suggestion
			if logFile, err = os.OpenFile(string(newConfig.Logging.FilePath), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644); err != nil {
				return err
			}
```

cmd/legacy_main.go
```diff
@@ -375,8 +375,15 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
...
+		}
 		// Run.
-		err = daemonize.Run(path, args, env, os.Stdout)
+		err = daemonize.Run(path, args, env, os.Stdout, logFile)
```
kislaykishore, line 386: Should we redirect it to os.Stderr instead of creating a new file handle?

Tulsishah: We can't redirect it to os.Stderr because if the user closes the terminal after mounting, the error messages will be lost.

cmd/legacy_main.go
```diff
@@ -375,8 +375,17 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
 		// programme is running as daemon process.
 		env = append(env, fmt.Sprintf("%s=true", logger.GCSFuseInBackgroundMode))
 
+		// logfile.stderr will capture the standard error (stderr) output of the gcsfuse background process.
+		// Default it will go on stdout if logfile not passed.
```
kislaykishore, line 379: Please add the rationale why this is directed to stdout instead of stderr.

kislaykishore (approved):

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -35,8 +35,9 @@ const (
 	logDirName         = "gcsfuse_integration_test_logs"
 	maxFileSizeMB      = 2
 	activeLogFileCount = 1
+	stderrLogFileCount = 1
 	backupLogFileCount = 2
-	logFileCount       = activeLogFileCount + backupLogFileCount
+	logFileCount       = activeLogFileCount + backupLogFileCount + stderrLogFileCount // Adding 1 forstderr logs file
```
kislaykishore, line 40:
```suggestion
	logFileCount       = activeLogFileCount + backupLogFileCount + stderrLogFileCount // Adding 1 for stderr logs file
```

kislaykishore (approved):

PR #2510, opened 2024-09-18
Adds a section for HNS to semantics doc

ankitaluthra1:
### Description

- This PR adds a section describing the support and advantages of HNS in gcsfuse

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -248,6 +248,16 @@ However, implicit directories does have drawbacks:
 
 Alternatively, users can create a script which lists the buckets and creates the appropriate objects for the directories so that the ```--implicit-dirs``` flag is not used.
 
+**Using Cloud Storage Buckets with Hierarchical Namespaces Enabled :**
+
+Cloud Storage Fuse also offers seamless support for buckets with hierarchical namespaces enabled. Mounting an HNS-enabled bucket using gcsfuse works exactly like mounting a standard bucket, with no additional configuration required.
+
+HNS-enabled buckets offer several advantages over standard buckets when used with cloud storage fuse:
+
+- HNS-enabled buckets eliminates the need for --implicit-dirs flag. HNS buckets inherently understand directories, so gcsfuse no longer needs this flag to simulate directory structures. Users will see consistent directory listings with or without the flag.
+- In HNS buckets, renaming a folder and its child folders is an atomic operation, meaning all associated resources—including objects and managed folders—are renamed in a single step. This ensures data consistency and significantly improves operation performance.
+- HNS buckets treat folders as first-class entities, closely aligning with traditional file system semantics. Commands like mkdir now directly create folder resources within the bucket, unlike with traditional buckets where directories were simulated using prefixes and 0-byte objects.
+- The [Objects.list](https://cloud.google.com/storage/docs/json_api/v1/objects/list) api is replaced with more efficient [Folder:get](https://cloud.google.com/storage/docs/json_api/v1/folders/getfoldermetadata) api during lookup operations.
 # Generations
 
 With each record in Cloud Storage is stored object and metadata [generation numbers](https://cloud.google.com/storage/docs/generations-preconditions). These provide a total order on requests to modify an object's contents and metadata, compatible with causality. So if insert operation A happens before insert operation B, then the generation number resulting from A will be less than that resulting from B.
@@ -402,8 +412,8 @@ Transient errors can occur in distributed systems like Cloud Storage, such as ne
 ## Missing features
 
 Not all of the usual file system features are supported. Most prominently:
-- Renaming directories is by default not supported. A directory rename cannot be performed atomically in Cloud Storage and would therefore be arbitrarily expensive in terms of Cloud Storage operations, and for large directories would have high probability of failure, leaving the two directories in an inconsistent state.
-- However, if your application can tolerate the risks, you may enable renaming directories in a non-atomic way, by setting ```--rename-dir-limit```. If a directory contains fewer files than this limit and no subdirectory, it can be renamed.
+- Renaming directories is by default not supported in non-hierarchical buckets. A directory rename cannot be performed atomically in these buckets and would therefore be arbitrarily expensive in terms of Cloud Storage operations, and for large directories would have high probability of failure, leaving the two directories in an inconsistent state. This can be resolved using cloud storage buckets with hierarchical namespace enabled. HNS enabled buckets support atomic folder renames and are fully compatible with Cloud Storage FUSE, offering seamless integration.
+- However, if your application is using non-hierarchical buckets and can tolerate the risks, you may enable renaming directories in a non-atomic way, by setting ```--rename-dir-limit```. If a directory contains fewer files than this limit and no subdirectory, it can be renamed.
 - File and directory permissions and ownership cannot be changed. See the permissions section above.
 - Modification times are not tracked for any inodes except for files.
 - No other times besides modification time are tracked. For example, ctime and atime are not tracked (but will be set to something reasonable). Requests to change them will appear to succeed, but the results are unspecified.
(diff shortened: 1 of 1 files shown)


docs/semantics.md
```diff
@@ -248,6 +248,16 @@ However, implicit directories does have drawbacks:
 
 Alternatively, users can create a script which lists the buckets and creates the appropriate objects for the directories so that the ```--implicit-dirs``` flag is not used.
 
+**Using Cloud Storage Buckets with Hierarchical Namespaces Enabled :**
+
+Cloud Storage Fuse also offers seamless support for buckets with hierarchical namespaces enabled. Mounting an HNS-enabled bucket using gcsfuse works exactly like mounting a standard bucket, with no additional configuration required.
+
+HNS-enabled buckets offer several advantages over standard buckets when used with cloud storage fuse:
+
+- HNS-enabled buckets eliminates the need for --implicit-dirs flag. HNS buckets inherently understand directories, so gcsfuse no longer needs this flag to simulate directory structures. Users will see consistent directory listings with or without the flag.
```
charith87, line 257: we could add gcsfuse does not need to simulate directories using placeholder objects ( 0-byte objects ending with '/' )

docs/semantics.md
```diff
@@ -248,6 +248,16 @@ However, implicit directories does have drawbacks:
...
+- In HNS buckets, renaming a folder and its child folders is an atomic operation, meaning all associated resources—including objects and managed folders—are renamed in a single step. This ensures data consistency and significantly improves operation performance.
+- HNS buckets treat folders as first-class entities, closely aligning with traditional file system semantics. Commands like mkdir now directly create folder resources within the bucket, unlike with traditional buckets where directories were simulated using prefixes and 0-byte objects.
+- The [Objects.list](https://cloud.google.com/storage/docs/json_api/v1/objects/list) api is replaced with more efficient [Folder:get](https://cloud.google.com/storage/docs/json_api/v1/folders/getfoldermetadata) api during lookup operations.
```
charith87, line 260: I think we can leave out list Objects. Its a little bit of internal detail.
We can just mention we use GetFolderMetadata which is more efficient and inline to fetch metadata for lookup operations.

docs/semantics.md
```diff
@@ -402,8 +412,8 @@ Transient errors can occur in distributed systems like Cloud Storage, such as ne
...
-- Renaming directories is by default not supported. A directory rename cannot be performed atomically in Cloud Storage and would therefore be arbitrarily expensive in terms of Cloud Storage operations, and for large directories would have high probability of failure, leaving the two directories in an inconsistent state.
-- However, if your application can tolerate the risks, you may enable renaming directories in a non-atomic way, by setting ```--rename-dir-limit```. If a directory contains fewer files than this limit and no subdirectory, it can be renamed.
+- Renaming directories is by default not supported in non-hierarchical buckets. A directory rename cannot be performed atomically in these buckets and would therefore be arbitrarily expensive in terms of Cloud Storage operations, and for large directories would have high probability of failure, leaving the two directories in an inconsistent state. This can be resolved using cloud storage buckets with hierarchical namespace enabled. HNS enabled buckets support atomic folder renames and are fully compatible with Cloud Storage FUSE, offering seamless integration.
```
charith87, line 415: Can we use same terminology here regarding non-heirarchical buckets. Its mentioned as Heierarchical namespace buckets above. We can use similar here - "Non Heirarchical Namespace Buckets"  with a link to HNS feature

docs/semantics.md
```diff
@@ -402,8 +412,8 @@ Transient errors can occur in distributed systems like Cloud Storage, such as ne
...
+- However, if your application is using non-hierarchical buckets and can tolerate the risks, you may enable renaming directories in a non-atomic way, by setting ```--rename-dir-limit```. If a directory contains fewer files than this limit and no subdirectory, it can be renamed.
```
charith87, line 416: This comment is not related to this PR but it was be good to add to complete this guidance. 
"by setting rename-dir-limit to number of maximum files you expect to be under a directory" and/or link to the public documentation where the rename dir limit is documented.

docs/semantics.md
marcoa6: Lets just use what we did in cloud docs:
[List object](https://pkg.go.dev/cloud.google.com/go/storage#hdr-Listing_objects) calls (BucketHandle.Objects), are converted to [get folder](https://pkg.go.dev/cloud.google.com/go/storage@v1.43.0/control/apiv2#StorageControlClient.GetFolder) calls, resulting in quicker response times and fewer overall list calls for every operation.

docs/semantics.md
marcoa6: I suggest rewording:


Renaming directories is only supported in Hierarchical Namespace Buckets, where they are fast and atomic. Renaming directories in flat namespace buckets is by default not supposed. A directory rename cannot be performed atomically in these flat buckets and would therefore be arbitrarily expensive in terms of Cloud Storage operations, and for large directories would have high probability of failure, leaving the two directories in an inconsistent state.

docs/semantics.md
marcoa6: Change non-hierarchical to flat

docs/semantics.md
ankitaluthra1: done

docs/semantics.md
ankitaluthra1: Adding same as cloud with minor tweakes, please see @marcoa6 , I felt converted is a bit ambiguous so used replaced and specified every operation as every lookup operation to be precise.

[List object](https://pkg.go.dev/cloud.google.com/go/storage#hdr-Listing_objects) calls (BucketHandle.Objects), are converted to [get folder](https://pkg.go.dev/cloud.google.com/go/storage@v1.43.0/control/apiv2#StorageControlClient.GetFolder) calls, resulting in quicker response times and fewer overall list calls for every lookup operation.

docs/semantics.md
```diff
@@ -248,6 +248,16 @@ However, implicit directories does have drawbacks:
 
 Alternatively, users can create a script which lists the buckets and creates the appropriate objects for the directories so that the ```--implicit-dirs``` flag is not used.
 
+**Using Cloud Storage Buckets with Hierarchical Namespaces Enabled :**
+
+Cloud Storage Fuse also offers seamless support for buckets with hierarchical namespaces enabled. Mounting an HNS-enabled bucket using gcsfuse works exactly like mounting a standard bucket, with no additional configuration required.
```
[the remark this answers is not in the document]
ankitaluthra1: added

docs/semantics.md
ankitaluthra1: Done

docs/semantics.md
ankitaluthra1: Done

docs/semantics.md
```diff
@@ -402,8 +413,8 @@ Transient errors can occur in distributed systems like Cloud Storage, such as ne
 ## Missing features
 
 Not all of the usual file system features are supported. Most prominently:
-- Renaming directories is by default not supported. A directory rename cannot be performed atomically in Cloud Storage and would therefore be arbitrarily expensive in terms of Cloud Storage operations, and for large directories would have high probability of failure, leaving the two directories in an inconsistent state.
-- However, if your application can tolerate the risks, you may enable renaming directories in a non-atomic way, by setting ```--rename-dir-limit```. If a directory contains fewer files than this limit and no subdirectory, it can be renamed.
+- Renaming directories is only supported in Hierarchical Namespace Buckets, where they are fast and atomic. Renaming directories in flat namespace buckets is by default not supposed. A directory rename cannot be performed atomically in these flat buckets and would therefore be arbitrarily expensive in terms of Cloud Storage operations, and for large directories would have high probability of failure, leaving the two directories in an inconsistent state.
```
charith87, line 416: *supposed = supported.

charith87 (approved):

docs/semantics.md
ankitaluthra1: done

PR #2530, opened 2024-09-24
Add negative cache for folders

Tulsishah:
### Description
We're adding a negative stat cache for folders to reduce GCS calls for negative entries. Although we initially implemented functions for this, we encountered corner cases that needed addressing, which didn't fit within the release timeframe.  Therefore, @ankitaluthra1 and I decided to postpone this feature as a post-launch item, as it's not a release blocker.

Scenarios:
BqdmEy5DnjUtiP5
We can say that StatObject(A) will cache a negative folder entry for "A" because "A" can only exist as an object, not a folder. Similarly, GetFolder(A/) will cache a negative object entry for "A/" as it can only exist as a folder, not an object, in HNS with GCSFuse.

**Other changes**
- Replaced the negative cache entry with an invalidation step in the DeleteFolder function. This change aligns DeleteFolder with the existing DeleteObject function, which also uses invalidation. (Because adding negative cache was changing behaviour compare to object deletion)
- Added logic to erase cache entries for renamed folders as part of the stale data cleanup process. Added stat call to fill negative cache in composite test.
- Implicit_dir test changes - The test were failing due to all the tests were performing operations on single folder.  One test were filling negative cache and second one try to access it. So I created separate folder for each test to avoid failure.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -386,16 +381,24 @@ func (b *fastStatBucket) GetFolder(
...
-	folder, error := b.wrapped.GetFolder(ctx, prefix)
+	// Fetch the Folder from GCS
+	return b.getFolderFromGCS(ctx, prefix)
+}
 
-	if error != nil {
-		return nil, error
+func (b *fastStatBucket) getFolderFromGCS(ctx context.Context, prefix string) (*gcs.Folder, error) {
+	f, err := b.wrapped.GetFolder(ctx, prefix)
+	if err != nil {
```
vadlakondaswetha, line 390: to avoid nested if, please swap the if-else

if err == nil {
b.insertFolder(f)
return f, nil
}

internal/cache/metadata/stat_cache_test.go
```diff
@@ -511,7 +511,8 @@ func (t *StatCacheTest) Test_ShouldReturnHitFalseWhenOnlyObjectAlreadyHasEntry()
 
 	hit, entry := t.statCache.LookUpFolder(name, someTime)
 
-	assert.False(t.T(), hit)
+	// If "key1" object exist then corresponding folder entry will be nil, but hit will be true as key have entry in the cache for object.
+	assert.True(t.T(), hit)
 	assert.Nil(t.T(), entry)
```
vadlakondaswetha, line 516: if folder is nil, isn't that a problem: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/dir.go#L387

internal/gcsx/prefix_bucket.go
```diff
@@ -245,6 +245,12 @@ func (b *prefixBucket) CreateFolder(ctx context.Context, folderName string) (*gc
 func (b *prefixBucket) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (*gcs.Folder, error) {
 	mFolderName := b.wrappedName(folderName)
 	mDestinationFolderId := b.wrappedName(destinationFolderId)
-	o, err := b.wrapped.RenameFolder(ctx, mFolderName, mDestinationFolderId)
-	return o, err
+	f, err := b.wrapped.RenameFolder(ctx, mFolderName, mDestinationFolderId)
+
+	// Modify the returned folder.
+	if f != nil {
```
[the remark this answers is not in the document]
Tulsishah: The returned folder in [fs.go]((https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/fs.go#L2116)) is not being utilized. Currently, we only rename the folder without creating a corresponding inode. The assumption is that an inode will be generated when an operation is performed on the renamed folder.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -344,9 +344,7 @@ func (b *fastStatBucket) DeleteFolder(ctx context.Context, folderName string) er
 	if err != nil {
 		return err
 	}
-	// Add negative entry in the cache.
-	b.addNegativeEntryForFolder(folderName)
-
+	b.invalidate(folderName)
```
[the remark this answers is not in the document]
Tulsishah: The reason for this removal was to address a test failure originating from the kernel_list_cache package.

Scenario:
```
A folder is created.
The folder is deleted using a client outside of GCSFuse.
A new folder with the same name is created within GCSFuse.
Attempting to stat the folder results in failure.
````
To ensure consistent behavior between objects and folders in the tests, it was determined that folder deletion should mirror the existing object deletion behavior. 

We can do this as a part of that bug. It will be similar to what we will do for objects.

File: internal/storage/caching/fast_stat_bucket.go [hunk 5]
```diff
@@ -413,11 +416,16 @@ func (b *fastStatBucket) CreateFolder(ctx context.Context, folderName string) (f
 	return
 }
 
-func (b *fastStatBucket) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (o *gcs.Folder, err error) {
-	o, err = b.wrapped.RenameFolder(ctx, folderName, destinationFolderId)
+func (b *fastStatBucket) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (*gcs.Folder, error) {
```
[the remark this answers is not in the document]
Tulsishah: It is not possible. We are returning error in that case. [ref](https://github.com/GoogleCloudPlatform/gcsfuse/blob/9c0ff1dff36a1de9c0b9b11dfaa47a495de84359/internal/fs/fs.go#L2096)

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -413,11 +416,16 @@ func (b *fastStatBucket) CreateFolder(ctx context.Context, folderName string) (f
...
+	f, err := b.wrapped.RenameFolder(ctx, folderName, destinationFolderId)
+	if err != nil {
+		return nil, err
+	}
 
 	// Invalidate cache for old directory.
 	b.eraseEntriesWithGivenPrefix(folderName)
+	// Clean all the stale entries related to renamed folder.
+	b.eraseEntriesWithGivenPrefix(destinationFolderId)
```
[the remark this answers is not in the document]
vadlakondaswetha: eraseEntries is a costly operation. you are taking a lock on entire cache and looping through all the elements. I would prefer avoiding it if its being about cautious.

Also regarding stat B throw a "not exist" error -> why are you not inserting destination folder into cache after rename?

internal/cache/metadata/stat_cache_test.go
Tulsishah: We are returning NotFound error in that case. [ref](https://github.com/GoogleCloudPlatform/gcsfuse/blob/9c0ff1dff36a1de9c0b9b11dfaa47a495de84359/internal/storage/caching/fast_stat_bucket.go#L378). Which means it should hit [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/v2.5.0/internal/fs/inode/dir.go#L382) case and return nil,nil.

internal/storage/caching/fast_stat_bucket.go
Tulsishah: Done

internal/storage/caching/fast_stat_bucket.go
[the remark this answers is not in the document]
Tulsishah: Discussed offline, Adding insert destination folder for now. but renaming folder into existing folder is not supported by RenameFolder api, which we have to communicate to folder team.

internal/cache/metadata/stat_cache_test.go
```diff
@@ -511,7 +511,8 @@ func (t *StatCacheTest) Test_ShouldReturnHitFalseWhenOnlyObjectAlreadyHasEntry()
 
 	hit, entry := t.statCache.LookUpFolder(name, someTime)
 
-	assert.False(t.T(), hit)
+	// If "key1" object exist then corresponding folder entry will be nil, but hit will be true as key have entry in the cache for object.
```
vadlakondaswetha, line 514: I didnt get this comment. aren't the keys different for objects and folders. I mean for folders they end with "/"

internal/storage/caching/fast_stat_bucket.go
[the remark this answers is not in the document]
vadlakondaswetha: Can you add a comment about this here. that caching negative entires for both object and folder will be done together because of test failures.

internal/cache/metadata/stat_cache_test.go
vadlakondaswetha: got it. Ideally this scenario should never get triggered right.

vadlakondaswetha (approved):

internal/cache/metadata/stat_cache_test.go
Tulsishah: Yes.. But in test we have written like this so I wrote comment for better understanding

PR #2534, opened 2024-09-25
Instrument fs ops only when tracing enabled.

kislaykishore:
### Description
Tracing is disabled by default. While the calls to tracing APIs are cheap (0.2s per 1M calls), we can reduce that impact too if we conditionally enable tracing depending upon whether it's enabled.
b/369493319

### Testing details
1. Manual - Verified that the instrumentation is enabled only when the tracing mode is set and not otherwise.
2. Unit tests - Added
3. Integration tests - NA


kislaykishore:
Closing against https://github.com/GoogleCloudPlatform/gcsfuse/pull/2542

PR #2539, opened 2024-09-26
Fixes retry for control client

ankitaluthra1:
### Description
Separates retry for control client from data client | increses timeout for retry | adds more error code for retry in control client

NA

### Testing details
1. Manual - NA
2. Unit tests - done
3. Integration tests - NA

### DIFF
--- a/internal/storage/storageutil/control_client.go
+++ b/internal/storage/storageutil/control_client.go
@@ -18,20 +18,31 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	control "cloud.google.com/go/storage/control/apiv2"
 	"github.com/googleapis/gax-go/v2"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
 	"google.golang.org/api/option"
+	"google.golang.org/grpc/codes"
 )
 
 func storageControlClientRetryOptions(clientConfig *StorageClientConfig) []gax.CallOption {
-	return []gax.CallOption{gax.WithRetry(func() gax.Retryer {
-		return gax.OnErrorFunc(gax.Backoff{
-			Max:        clientConfig.MaxRetrySleep,
-			Multiplier: clientConfig.RetryMultiplier,
-		}, ShouldRetry)
-	})}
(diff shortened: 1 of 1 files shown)


internal/storage/storageutil/control_client.go
```diff
@@ -18,20 +18,31 @@ import (
...
 )
 
 func storageControlClientRetryOptions(clientConfig *StorageClientConfig) []gax.CallOption {
-	return []gax.CallOption{gax.WithRetry(func() gax.Retryer {
-		return gax.OnErrorFunc(gax.Backoff{
-			Max:        clientConfig.MaxRetrySleep,
-			Multiplier: clientConfig.RetryMultiplier,
-		}, ShouldRetry)
-	})}
+	return []gax.CallOption{
```
Tulsishah, line 31: In go sdk for control client, CreateFolder and DeleteFolder is not retrybe https://github.com/googleapis/google-cloud-go/blob/main/storage/control/apiv2/storage_control_client.go#L76

Do we want to keep the same?

ankitaluthra1: yes I also had same doubt, these methods mght not be idempotent so should not be retried, I didnt change it to keep the previous behaviour same in case we added retry after seeing any retriable error

Tulsishah (approved):

internal/storage/storageutil/control_client.go
```diff
@@ -18,25 +18,34 @@ import (
...
+				Multiplier: clientConfig.RetryMultiplier,
+			})
+		}),
+	}
 }
 
 func setRetryConfigForFolderAPIs(sc *control.StorageControlClient, clientConfig *StorageClientConfig) {
-	sc.CallOptions.CreateFolder = storageControlClientRetryOptions(clientConfig)
-	sc.CallOptions.DeleteFolder = storageControlClientRetryOptions(clientConfig)
 	sc.CallOptions.RenameFolder = storageControlClientRetryOptions(clientConfig)
```
charith87, line 49: Isnt RenameFolder also not idempotent?

charith87 (approved):

PR #2542, opened 2024-09-27
Instrument fs ops only when tracing enabled.

kislaykishore:
While the calls to tracing APIs are cheap,
we can reduce the performance impact by
conditionally instrumenting code only if tracing is enabled.

b/369493319

### Testing details
1. Manual - Verified that the instrumentation is enabled only when the tracing mode is set and not otherwise.
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/fs/wrappers/tracing.go
+++ b/internal/fs/wrappers/tracing.go
@@ -1,0 +1,175 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


internal/fs/wrappers/tracing.go
```diff
@@ -0,0 +1,172 @@
...
+		tracer:  otel.Tracer(name),
+	}
+}
+
+func (fs *tracing) Destroy() {
+	fs.wrapped.Destroy()
+}
+
+func (fs *tracing) invokeWrapped(ctx context.Context, opName string, w wrappedCall) error {
+	// Span's SpanKid is set to trace.SpanKindServer since GCSFuse is like a server for the requests that the Kernel sends.
```
ashmeenkaur, line 45:
```suggestion
	// Span's SpanKind is set to trace.SpanKindServer since GCSFuse is like a server for the requests that the Kernel sends.
```

ashmeenkaur (dismissed):

ashmeenkaur (approved):

PR #2543, opened 2024-09-27
[Buffered Writes] add configs for buffered writes feature

ashmeenkaur:
### Description

The following hidden flags/configs and their associated unit/composite tests are added for buffered writes:
```
- config-path: "write.experimental-enable-streaming-writes"
  flag-name: "experimental-enable-streaming-writes"
  type: "bool"
  usage: "To enable streaming uploads during write file operation."
  default: false

- config-path: "write.global-max-blocks"
  flag-name: "write-global-max-blocks"
  type: "int"
  usage: >-
    To configure the maximum number of blocks to be used by all files for
    streaming writes. Configured value should be >= 2 or -1 (for infinite blocks).
  default: -1 

- config-path: "write.max-blocks-per-file"
  flag-name: "write-max-blocks-per-file"
  type: "int"
  usage: >-
    To configure the maximum number of blocks to be used by a single file for 
    streaming writes. Configured value should be >= 2 or -1 (for infinite blocks).
  default: -1

- config-path: "write.block-size-mb"
  flag-name: "write-block-size-mb"
  type: "int"
  usage: >-
    Specifies the block size for streaming writes. The value should be more 
    than 0.
  default: 64 #TODO: revisit default value after perf testing.
  hide-flag: true
 ```

The default values will be revisited later.

NA

### Testing details
1. Manual - validated that flags are hidden.
2. Unit tests - Added
3. Integration tests - NA

### CODE REVIEW

File: cfg/params.yaml
Comment (kislaykishore) [inline cfg/params.yaml:1]: Should we add these flags as experimental and once implemented remove the experimental tag?

Comment (vadlakondaswetha) [reply cfg/params.yaml:1]: Agreed. We should always start with experimental and then move to non-experimental. To keep it simple we can make only the enable-streaming-writes as experimental since we don't consider other flags if this is false.

Comment (ashmeenkaur) [reply cfg/params.yaml:1]: Changed `enable-streaming-writes` to `experimental-enable-streaming-writes`. Please take another look!

File: cfg/validate_test.go [hunk 1]
```diff
@@ -303,3 +303,122 @@ func Test_IsTtlInSecsValid_ValidScenarios(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.testName, func(t *testing.T) {
+			assert.Error(t, isValidWriteStreamingConfigs(&tc.writeConfig))
+		})
+	}
+}
+
+func Test_isValidWriteConfig_ValidScenarios(t *testing.T) {
```
kislaykishore, line 377:
```suggestion
func Test_isValidWriteConfig_validScenarios(t *testing.T) {
```

ashmeenkaur: Could not find the style guide for this so renamed it as per the example given in [go tip](https://g3doc.corp.google.com/company/teams/go-community/gotip/episodes/50.md?cl=head).

kislaykishore (approved):

PR #2566, opened 2024-10-09
Add SetRetry in CreateStorageClient

vipnydav:
### Description
Added a SetRetry feature in CreateStorageClient to handle transient errors. Since our end-to-end tests use separate buckets for different VMs and each test operates on distinct resources, we can safely bypass the idempotency check.

NA

### Testing details
1. Manual - Not tested as required effort was more than scope of project.
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -52,9 +52,10 @@ func CreateStorageClient(ctx context.Context) (client *storage.Client, err error
 	if err != nil {
 		return nil, fmt.Errorf("storage.NewClient: %w", err)
 	}
-	//RetryAlways causes all operations to be retried when the service returns transient error,
-	//regardless of idempotency considerations. Since the concurrent execution of our CI/CD tests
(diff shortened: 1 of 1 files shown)


tools/integration_tests/util/client/storage_client.go
```diff
@@ -49,6 +52,14 @@ func CreateStorageClient(ctx context.Context) (client *storage.Client, err error
 	if err != nil {
 		return nil, fmt.Errorf("storage.NewClient: %w", err)
 	}
+	client.SetRetry(
```
raj-prince, line 55: Can we write a comment here?

Can we write a comment here?

// RetryAlways causes all operations to be retried when the service returns transient error, regardless of idempotency considerations. Since the concurrent execution of our CI/CD tests (VMs, threads) doesn't share any cloud-storage resources, hence it's safe to disregard idempotency.

raj-prince (dismissed):

tools/integration_tests/util/client/storage_client.go
vipnydav: Added

raj-prince (approved):

PR #2574, opened 2024-10-10
Add a loop to read 5 times from kernel cache and take minimum listing time

vipnydav:
### Description
Added a loop to read 5 times from kernel cache and take minimum listing time in order to ensure that listing from GCS takes twice or more time than listing from kernel cache .

NA

### Testing details
1. Manual - Done
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
+++ b/tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
@@ -15,6 +15,7 @@
 package list_large_dir_test
 
 import (
+	"math"
 	"os"
 	"path"
 	"strconv"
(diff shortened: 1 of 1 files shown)


tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
```diff
@@ -164,21 +165,31 @@ func TestListDirectoryWithTwelveThousandFiles(t *testing.T) {
...
-	startTime = time.Now()
-	objs, err = os.ReadDir(dirPath)
-	if err != nil {
-		t.Errorf("Error in listing directory: %v", err)
+	minSecondListTime := time.Duration(math.MaxInt64)
+	for i := 0; i < 5; i++ {
+		startTime = time.Now()
+		objs, err = os.ReadDir(dirPath)
+		if err != nil {
+			t.Errorf("Error in listing directory: %v", err)
```
Tulsishah, line 175: nit: t.Fatalf to avoid further execution of test.

tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
```diff
@@ -201,21 +213,31 @@ func TestListDirectoryWithTwelveThousandFilesAndHundredExplicitDir(t *testing.T)
...
 	firstListTime := endTime.Sub(startTime)
+
 	// Listing the directory a second time should retrieve the response from the kernel cache.
-	startTime = time.Now()
-	objs, err = os.ReadDir(dirPath)
-	if err != nil {
-		t.Errorf("Error in listing directory: %v", err)
+	minSecondListTime := time.Duration(math.MaxInt64)
+	for i := 0; i < 5; i++ {
+		startTime = time.Now()
```
Tulsishah, line 220: Think if we can write common function for this loop as we are using thrice the same code.

vipnydav: But we are using different methods to validate directory inside each loop.

tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
vipnydav: Done

tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
vipnydav: Done

Tulsishah (approved):

PR #2575, opened 2024-10-10
Adding config for retry read-stalled-req

raj-prince:
### Description
Adding  hidden flags for read-stall-req.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


cfg/params.yaml
```diff
@@ -346,6 +346,49 @@
   usage: Param for exponential backoff algorithm, which is used to increase waiting time b/w two consecutive retries.
   default: 2
 
+- config-path: "gcs-retries.read-stall.enable"
+  flag-name: "enable-read-stall-retry"
+  type: "bool"
+  usage: >-
+    Enable/disable retry for stalled read-request based on dynamic timeout.
```
kislaykishore, line 353: ">" is used for breaking a long string into multiple lines and is not needed here. The "-" at the end removes the newline in multiline strings that gets added automatically. They are not needed here since it's a short string.
```suggestion
  usage: Enables retry for stalled read-request based on dynamic timeout.
```

cfg/params.yaml
```diff
@@ -346,6 +346,49 @@
...
+  default: false
+  hide-flag: true
+
+- config-path: "gcs-retries.read-stall.initial-req-timeout"
+  flag-name: "initial-read-stall-req-timeout"
```
kislaykishore, line 358: following the naming convention that's been used throughout this file.
```suggestion
  flag-name: "read-stall-initial-req-timeout"
```

cfg/params.yaml
```diff
@@ -346,6 +346,49 @@
...
+  type: "duration"
+  usage: Initial value of the read-request dynamic timeout.
+  default: 20s
+  hide-flag: true
+
+- config-path: "gcs-retries.read-stall.max-req-timeout"
+  flag-name: "max-read-stall-req-timeout"
```
kislaykishore, line 365: following the naming convention that's been used throughout this file.
```suggestion
  flag-name: "read-stall-max-req-timeout"
```

cfg/params.yaml
```diff
@@ -346,6 +346,49 @@
...
+  type: "duration"
+  usage: Upper bound of the read-request dynamic timeout.
+  default: 20m
+  hide-flag: true
+
+- config-path: "gcs-retries.read-stall.min-req-timeout"
+  flag-name: "min-read-stall-req-timeout"
```
kislaykishore, line 372: following the naming convention that's been used throughout this file.
```suggestion
  flag-name: "read-stall-min-req-timeout"
```

cfg/params.yaml
```diff
@@ -346,6 +346,49 @@
...
+  type: "duration"
+  usage: Lower bound of the read request dynamic timeout.
+  default: 500ms
+  hide-flag: true
+
+- config-path: "gcs-retries.read-stall.req-increase-rate"
+  flag-name: "read-stall-req-increase-rate"
+  type: "float64"
+  usage: Rate to change the dynamic timeout if required.
```
kislaykishore, line 381: I don't understand how the dynamic timeout will be changed. So, maybe make this text more elaborate, or if it's very complicated to explain, then add a TODO to link to the cloud docs when the page is up.

cfg/validate.go
```diff
@@ -158,6 +158,19 @@ func isValidWriteStreamingConfig(wc *WriteConfig) error {
...
 
+func isValidReadStallGcsRetriesConfig(rsc *ReadStallGcsRetriesConfig) error {
+	if rsc == nil {
+		return nil
+	}
+	if rsc.ReqIncreaseRate <= 0 {
+		return fmt.Errorf("invalid value of increaseRate: %f, can't be 0 or negative", rsc.ReqIncreaseRate)
+	}
+	if rsc.ReqTargetPercentile <= 0 || rsc.ReqTargetPercentile >= 1 {
+		return fmt.Errorf("invalid value of targetPercentile: %f, should be greater than 0 and less than 1", rsc.ReqTargetPercentile)
```
kislaykishore, line 169:
```suggestion
		return fmt.Errorf("invalid value of targetPercentile: %f, should be in the range (0, 1)", rsc.ReqTargetPercentile)
```

kislaykishore (changes requested):

cfg/params.yaml
raj-prince: Added more detailed description, which requires newline.

cfg/params.yaml
raj-prince: This is just for internal testing, we will never expose this to user. Rephrased it.

kislaykishore (approved):

PR #2578, opened 2024-10-11
upgrade to golang v1.23.2

ashmeenkaur:
### Description
upgrade to golang v1.23.2

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
4. Perf tests - 
B5W77Y96Etk5XNG

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/googlecloudplatform/gcsfuse/v2
 
-go 1.23.0
+go 1.23.2
 
 require (
 	cloud.google.com/go/compute/metadata v0.5.2
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@
 # Mount the gcsfuse to /mnt/gcs:
 #  > docker run --privileged --device /fuse -v /mnt/gcs:/gcs:rw,rshared gcsfuse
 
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/googlecloudplatform/gcsfuse/v2
 
-go 1.23.0
+go 1.23.2
```
kislaykishore, line 3: Why do we need to change this? The source should still be compatible with golang 1.23.0, right?

ashmeenkaur: This ensures the `package_gcsfuse` Dockerfile/end user/developer always compiles GCSFuse with Go 1.23.2, which includes important security fixes. Since end users are not expected to compile GCSFuse themselves, this restriction should not cause any issues IMO. WDYT?

kislaykishore: I think it's not optimal to update the source version and making users download/install a newer version of GoLang everytime. Development and release environments are also not guaranteed to be exactly the same even if we upgrade the go.mod version since if the development environment has a more recent version of Golang installed, that'll be used instead of what's specified in the go.mod.
TBH, it's not a big deal for us since GCSFuse is a binary and not a library, so I'm okay approving it.

kislaykishore (approved):

PR #2579, opened 2024-10-11
Fix to not use O_DIRECT by default

sethiay:
### Description
Fix to not use O_DIRECT by default.

Tests performed manually:
machine spec:
n2-standard-96
16LSSDs

fio
```
; -- use nrfiles and rw to CLI args to control readtype and number of files --
[global]
ioengine=libaio
direct=0
fadvise_hint=0
iodepth=64
invalidate=1
nrfiles=16
thread=1
openfiles=1
group_reporting=1
create_serialize=0
allrandrepeat=0
file_service_type=random
numjobs=1
filename_format=$jobname.$jobnum/$filenum

[Workload]
directory=${DIR}
bs=128K
filesize=10G
rw=read
```

pytorch data loader
```
Loaded Llama 70B model
```

With `enable-o-direct=true` 
Fio throughput: 733MiB/s (221s)
Llama 70B model load time: 275s

Without `enable-o-direct=false` (default with this fix)
Fio throughput: 812MiB/s (201s)
Llama 70B model load time: 267s

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/file/downloader/parallel_downloads_job.go
+++ b/internal/cache/file/downloader/parallel_downloads_job.go
@@ -62,10 +62,8 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
 
 	monitor.CaptureGCSReadMetrics(ctx, util.Parallel, end-start)
 
-	// Use of memory aligned buffer is not required if use of O_DIRECT is disabled.
+	// Use memory aligned buffer if O_DIRECT is enabled.
 	if job.fileCacheConfig.EnableODirect {
-		_, err = io.CopyN(dstWriter, newReader, end-start)
-	} else {
 		_, err = cacheutil.CopyUsingMemoryAlignedBuffer(ctx, newReader, dstWriter, end-start,
 			job.fileCacheConfig.WriteBufferSize)
 		// If context is canceled while reading/writing in CopyUsingMemoryAlignedBuffer
@@ -74,6 +72,8 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
 		if !errors.Is(err, context.Canceled) && errors.Is(ctx.Err(), context.Canceled) {
 			err = errors.Join(err, ctx.Err())
 		}
+	} else {
+		_, err = io.CopyN(dstWriter, newReader, end-start)
 	}
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -62,10 +62,8 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
 
 	monitor.CaptureGCSReadMetrics(ctx, util.Parallel, end-start)
 
-	// Use of memory aligned buffer is not required if use of O_DIRECT is disabled.
+	// Use memory aligned buffer if O_DIRECT is enabled.
 	if job.fileCacheConfig.EnableODirect {
```
vadlakondaswetha, line 66: It is recommended to put smaller code in if-block whenever there is else case. so its easier to understand the else part also.

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -62,8 +62,9 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
 
 	monitor.CaptureGCSReadMetrics(ctx, util.Parallel, end-start)
 
-	// Use of memory aligned buffer is not required if use of O_DIRECT is disabled.
-	if job.fileCacheConfig.EnableODirect {
+	// Use standard copy function if O_DIRECT is disabled and memory aligned
+	// buffer otherwise.
+	if !job.fileCacheConfig.EnableODirect {
```
kislaykishore, line 67: Can we get rid of the negation and reverse the if-else clauses so that it is easier to understand the flow. Also, please reduce nesting since the else-block is likely not needed here - https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

kislaykishore (approved):

internal/cache/file/downloader/parallel_downloads_job.go
sethiay: Discussed offline to keep it like because there is a statement(formatting error) below the `if-else` which is common to both `if` and `else` blocks.

vadlakondaswetha (approved):

PR #2582, opened 2024-10-14
[Error fix] GCSFuse mount command results in error if directory with bucket name exists in the working directory

ashmeenkaur:
### Description
Kernel mount command converts device/bucket name to path if directory with same name as bucket exists in the working directory. This results into error and failure of GCSFuse persistent mounts.
In order to fix this issue, we will convert the device name path received from kernel to it's base file name which will be bucket name.

NA

### Testing details
1. Manual - Manually verified that after this fix, user is able to mount GCSFuse even if bucket name directory exists in current working directory. The fix is working for both 
- `mount -t` command and 
- `/etc/fstab` entry.
2. Unit tests - Added

### DIFF
--- a/tools/mount_gcsfuse/main.go
+++ b/tools/mount_gcsfuse/main.go
@@ -187,10 +187,19 @@ func parseArgs(
 		// Is this the device?
 		case positionalCount == 0:
 			device = s
-			if strings.Contains(device, "\\") {
+			fmt.Println("device before: ", device)
+			// Kernel might be converting bucket name to path if a directory with same
+			// name as bucket exists in the root folder from which mount command is
+			// executed.
+			// As of 10th Oct, 2024, bucket names don't support "/", so it is safe to
+			// convert received bucket name with "/" to it's base file name.
+			if strings.Contains(device, "/") {
+				fmt.Println("Changing device..")
 				// Get the last part of the path (bucket name)
 				device = filepath.Base(device)
 			}
+			fmt.Println("device after: ", device)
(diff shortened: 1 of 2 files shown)


tools/mount_gcsfuse/main.go
```diff
@@ -186,6 +187,16 @@ func parseArgs(
 		// Is this the device?
 		case positionalCount == 0:
 			device = s
+			// The kernel might be converting the bucket name to a path if a directory with the
+			// same name as the bucket exists in the folder where the mount command is executed.
+			//
+			// As of October 10th, 2024, bucket names don't support "/", so it is safe to
```
Tulsishah, line 193: Should we add link as reference?

tools/mount_gcsfuse/main_test.go
```diff
@@ -148,3 +148,41 @@ func TestMakeGcsfuseArgs(t *testing.T) {
...
+			expectedDevice: "fake_bucket",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			gotDevice, _, _, err := parseArgs([]string{"/path_to_executable", tc.device, tc.mountPoint})
+
+			if assert.Nil(t, err) {
+				assert.Equal(t, tc.expectedDevice, gotDevice)
```
Tulsishah, line 184: nit:assert.NoError

Tulsishah (approved):
two small comments :) LGTM

Tulsishah (approved):

raj-prince (approved):

PR #2596, opened 2024-10-15
Replaced gcloud command with control client code in Create/DeleteManagedFoldersInBucket in managed_folders test package 

anushka567:
### Description
Replaced gcloud command in helper function in  managed folders test package, using control client to :
1.  DeleteManagedFoldersinBucket
2. CreateManagedFoldersinBucket
Motivation:  gcloud commands can't be run concurrently as part of parallelizing CD tests.Thus, needs to be replaced.
NA

### Testing details
1. Manual - Ran integration tests on GCE VM
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/tools/integration_tests/util/client/control_client.go
+++ b/tools/integration_tests/util/client/control_client.go
@@ -1,3 +1,17 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


tools/integration_tests/util/client/control_client.go
```diff
@@ -0,0 +1,64 @@
...
+func CreateManagedFoldersInBucket(ctx context.Context, client *control.StorageControlClient, managedFolderPath, bucket string) error {
+	mf := &controlpb.ManagedFolder{}
+	req := &controlpb.CreateManagedFolderRequest{
+		Parent:          fmt.Sprintf("projects/_/buckets/%v", bucket),
+		ManagedFolder:   mf,
+		ManagedFolderId: managedFolderPath,
+	}
+	if _, err := client.CreateManagedFolder(ctx, req); err != nil && !strings.Contains(err.Error(), "The specified managed folder already exists") {
+		log.Fatalf(fmt.Sprintf("Error while creating managed folder: %v", err))
+		return err
```
Tulsishah, line 61: If you do log.Fatalf then then return err will not get executed

anushka567: Updated.

Tulsishah (dismissed):

Tulsishah (approved):

PR #2597, opened 2024-10-15
[Buffered Writes] bucket interface changes for buffered writes part 1

ashmeenkaur:
### Description
This PR includes changes to bucket.go interface and all it's associated implementations to include the following methods:
- CreateObjectChunkWriter()
- FinalizeUpload()
fake bucket, prefix bucket and and content type bucket implementations are added in a subsequent PR #2606

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA


internal/storage/bucket_handle.go
```diff
@@ -237,6 +237,44 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 	o = storageutil.ObjectAttrsToBucketObject(attrs)
 	return
 }
+func (bh *bucketHandle) CreateObjectChunkWriter(ctx context.Context, req *gcs.CreateObjectRequest, chunkSize int, callBack func(bytesUploadedSoFar int64)) (gcs.Writer, error) {
+	// For phase 1 of buffered writes, we are doing chunk uploads only for new
+	// file uploads.
+	preconditions := storage.Conditions{}
```
vadlakondaswetha, line 243: why do you want to restrict for not exist scenario. why can't we just use the generation from the request

internal/storage/bucket_handle_test.go
```diff
@@ -417,6 +419,157 @@ func (testSuite *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObje
...
+		},
+		100,
+		func(_ int64) {})
+	require.NoError(testSuite.T(), err)
+	assert.Equal(testSuite.T(), objName, wr.ObjectName())
+
+	o, err := testSuite.bucketHandle.FinalizeUpload(context.Background(), wr)
+
+	assert.Error(testSuite.T(), err)
+	assert.Nil(testSuite.T(), o)
```
vadlakondaswetha, line 570: Should we validate the error type?

internal/storage/bucket_handle_test.go
```diff
@@ -417,6 +419,157 @@ func (testSuite *BucketHandleTest) TestCreateObjectMethodWhenGivenGenerationObje
...
+			o, err := testSuite.bucketHandle.FinalizeUpload(context.Background(), wr)
+
+			assert.NoError(t, err)
+			assert.NotNil(testSuite.T(), o)
+		})
+	}
+}
+
+func (testSuite *BucketHandleTest) createObject(objName string) {
+	testSuite.T().Helper()
```
vadlakondaswetha, line 534: I see this as definition for helper:
 Helper marks the calling function as a test helper function.
// When printing file and line information, that function will be skipped.
// Helper may be called simultaneously from multiple goroutines.

Don't we want to print file and line info when something goes wrong?

internal/storage/caching/fast_stat_bucket_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"testing"
 	"time"
 
+	strg "cloud.google.com/go/storage"
```
vadlakondaswetha, line 23: nit: call it gostorage

internal/storage/caching/fast_stat_bucket_test.go
```diff
@@ -136,6 +137,145 @@ func (t *CreateObjectTest) WrappedSucceeds() {
...
+// CreateObject
+////////////////////////////////////////////////////////////////////////
+
+type CreateObjectChunkWriterTest struct {
+	fastStatBucketTest
+}
+
+func init() { RegisterTestSuite(&CreateObjectChunkWriterTest{}) }
+
+func (t *CreateObjectChunkWriterTest) CallsWrapped() {
```
vadlakondaswetha, line 150: are you verifying the request parameters here. can you name the method accordingly

internal/ratelimit/throttled_bucket.go
```diff
@@ -94,6 +94,30 @@ func (b *throttledBucket) CreateObject(
...
+
+	// Call through.
+	wc, err = b.wrapped.CreateObjectChunkWriter(ctx, req, chunkSize, callBack)
+
+	return
+}
+
+func (b *throttledBucket) FinalizeUpload(ctx context.Context, w gcs.Writer) (*gcs.Object, error) {
+	// Wait for permission to call through.
+	err := b.opThrottle.Wait(ctx, 1)
```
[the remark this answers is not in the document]
ashmeenkaur: Yes, I think it is okay to just throttle for create writer operation as they will anyway share the context. We are in a way double counting here too. Thanks for catching this!

internal/storage/bucket_handle.go
ashmeenkaur: I was restricting to phase 1 of the project but makes sense to do this in one go as well. I'll update this.

internal/storage/bucket_handle_test.go
ashmeenkaur: This will still print the error but not with the file name and line number of test helper but with the calling Test function so it is easier to debug from where the error is coming from.

Eg: With Test Helper:
```
=== RUN   TestBucketHandleTestSuite/TestFinalizeUploadWithGenerationAsZeroWhenObjectAlreadyExists
    bucket_handle_test.go:554: 
        	Error Trace:	/usr/local/google/home/ashmeen/gcsfuse/internal/storage/bucket_handle_test.go:543
        	            				/usr/local/google/home/ashmeen/gcsfuse/internal/storage/bucket_handle_test.go:554
        	Error:      	An error is expected but got nil.
        	Test:       	TestBucketHandleTestSuite/TestFinalizeUploadWithGenerationAsZeroWhenObjectAlreadyExists
```

Without test helper:

```
=== RUN   TestBucketHandleTestSuite/TestFinalizeUploadWithGenerationAsZeroWhenObjectAlreadyExists
    bucket_handle_test.go:543: 
        	Error Trace:	/usr/local/google/home/ashmeen/gcsfuse/internal/storage/bucket_handle_test.go:543
        	            				/usr/local/google/home/ashmeen/gcsfuse/internal/storage/bucket_handle_test.go:554
        	Error:      	An error is expected but got nil.
        	Test:       	TestBucketHandleTestSuite/TestFinalizeUploadWithGenerationAsZeroWhenObjectAlreadyExists
--- FAIL: TestBucketHandleTestSuite (0.01s)
```

internal/storage/gcs/bucket.go
```diff
@@ -43,6 +44,12 @@ const (
 	ReqIdField string = "GcsReqId"
 )
 
+type Writer interface {
```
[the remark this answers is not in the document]
ashmeenkaur: We need different implementations of writer for testing - particularly for fake bucket that we use for composite tests. That is why I added this interface. Fake bucket writer implementation is in the subsequent PR.

internal/storage/storageutil/object_attrs.go
```diff
@@ -62,6 +62,9 @@ func convertACLRuleToObjectAccessControl(element storage.ACLRule) *storagev1.Obj
 }
 
 func ObjectAttrsToBucketObject(attrs *storage.ObjectAttrs) *gcs.Object {
+	if attrs == nil {
```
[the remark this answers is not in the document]
ashmeenkaur: No not a real scenario, this is added just to make unit tests simpler.

internal/storage/caching/fast_stat_bucket_test.go
ashmeenkaur: This test is validating that the wrapped bucket is called with the expected parameters. Let me know if you still think we should rename the test.

vadlakondaswetha (approved):

vadlakondaswetha (approved):

PR #2599, opened 2024-10-16
Fix e2e test

Tulsishah:
### Description
We are getting this error exec: "python": executable file not found in $PATH while running e2e tests in periodically due to newly added test which uses python library. Fixing it by using python3 instead of python while running tests.
- Small fixes to stat directory after renaming not exist.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/run_e2e_tests.sh
+++ b/tools/integration_tests/run_e2e_tests.sh
@@ -139,9 +139,7 @@ function install_packages() {
   sudo rm -rf /usr/local/go && tar -xzf go_tar.tar.gz && sudo mv go /usr/local
   export PATH=$PATH:/usr/local/go/bin
   sudo apt-get install -y python3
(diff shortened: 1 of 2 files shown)


tools/integration_tests/run_e2e_tests.sh
```diff
@@ -139,6 +139,7 @@ function install_packages() {
   sudo rm -rf /usr/local/go && tar -xzf go_tar.tar.gz && sudo mv go /usr/local
   export PATH=$PATH:/usr/local/go/bin
   sudo apt-get install -y python3
+  export PATH=$PATH:/usr/bin/python3
```
kislaykishore, line 142: Is this statement really necessary? I'd expect that after installation, python3 should automatically be available in the path.

ashmeenkaur (approved):

PR #2600, opened 2024-10-16
Added block interface and memory block implementations.

vadlakondaswetha:
### Description
Added block interface and memory block implementation

NA

### Testing details
1. Manual - N/A since this is standlone class for now
2. Unit tests - Added
3. Integration tests - N/A

### DIFF
--- a/internal/block/block.go
+++ b/internal/block/block.go
@@ -1,0 +1,64 @@
+package block
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+)
+
+// Block represents the buffer which holds the data.
+type Block interface {
+	// Reuse resets the blocks for reuse.
+	Reuse()
+
+	// Size provides the current data size of the block. The capacity of the block
+	// can be >= data_size.
+	Size() int64
+
+	// Write writes the given data to block.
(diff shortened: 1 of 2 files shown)


internal/block/block.go
```diff
@@ -0,0 +1,78 @@
...
+	start, end int64
+}
+
+type memoryBlock struct {
+	Block
+	buffer []byte
+	offset offset
+}
+
+func (m *memoryBlock) Reuse() {
```
ashmeenkaur, line 52: Eventually we would need Reuse function to take a `Block chan` as parameter and freed up `memoryBlock` will be put in that Block chan for reuse. Shall we do this in subsequent PR when required?

ashmeenkaur (approved):

internal/block/block.go
vadlakondaswetha: I am planning to make it the responsibility of the caller to put the block after reuse to the channel (basically what to do after reuse). The block struct shoudn't worry about it

ashmeenkaur: SG, thanks for clarifying!

PR #2601, opened 2024-10-17
Make control client retry able in e2e tests

Tulsishah:
### Description
Make control client retry able in e2e tests to avoid failures in case of server errors.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/util/client/control_client.go
+++ b/tools/integration_tests/util/client/control_client.go
@@ -18,14 +18,46 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"reflect"
 	"strings"
+	"time"
 
 	control "cloud.google.com/go/storage/control/apiv2"
 	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googleapis/gax-go/v2"
+	"google.golang.org/grpc/codes"
 )
+
+func storageControlClientRetryOptions(clientConfig *control.StorageControlClient) []gax.CallOption {
+	return []gax.CallOption{
+		gax.WithTimeout(300000 * time.Millisecond),
+		gax.WithRetry(func() gax.Retryer {
+			return gax.OnCodes([]codes.Code{
+				codes.ResourceExhausted,
+				codes.Unavailable,
+				codes.DeadlineExceeded,
+				codes.Internal,
+				codes.Unknown,
+			}, gax.Backoff{
+				Max:        30 * time.Second,
+				Multiplier: 2,
+			})
+		}),
+	}
+}
 
 func CreateControlClient(ctx context.Context) (client *control.StorageControlClient, err error) {
 	client, err = control.NewStorageControlClient(ctx)
+
+	opts := &control.StorageControlCallOptions{}
+
+	// Loop over all fields in the struct and assign retry options
+	v := reflect.ValueOf(opts).Elem()
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		field.Set(reflect.ValueOf(storageControlClientRetryOptions(client)))
+	}
+
 	if err != nil {
 		return nil, fmt.Errorf("control.NewStorageControlClient: #{err}")
 	}
--- a/tools/integration_tests/util/client/control_client.go
+++ b/tools/integration_tests/util/client/control_client.go
@@ -48,7 +48,6 @@ func storageControlClientRetryOptions(clientConfig *control.StorageControlClient
 
 func CreateControlClient(ctx context.Context) (client *control.StorageControlClient, err error) {
 	client, err = control.NewStorageControlClient(ctx)
-
 	opts := &control.StorageControlCallOptions{}
 
 	// Loop over all fields in the struct and assign retry options
--- a/tools/integration_tests/util/client/control_client.go
+++ b/tools/integration_tests/util/client/control_client.go
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This code initializes a control client solely for the purpose of setting up test data for
+// end-to-end tests.
+// This client is not used in the application logic itself.
+
 package client
 
 import (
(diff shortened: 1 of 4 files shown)


tools/integration_tests/util/client/control_client.go
```diff
@@ -18,14 +18,45 @@ import (
...
+	"time"
 
 	control "cloud.google.com/go/storage/control/apiv2"
 	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googleapis/gax-go/v2"
+	"google.golang.org/grpc/codes"
 )
 
+func storageControlClientRetryOptions(clientConfig *control.StorageControlClient) []gax.CallOption {
+	return []gax.CallOption{
```
charith87, line 32: Shouldn't we try to reuse this https://github.com/GoogleCloudPlatform/gcsfuse/blob/21633e8ffc6f4d2c1c72e79e08996a98543c95de/internal/storage/storageutil/control_client.go#L30 rather than defining it here.
It will give rise to maintenance issues when this is changed.

tools/integration_tests/util/client/control_client.go
```diff
@@ -19,13 +19,27 @@ import (
...
 func CreateControlClient(ctx context.Context) (client *control.StorageControlClient, err error) {
 	client, err = control.NewStorageControlClient(ctx)
+	clientConfig := &storageutil.StorageClientConfig{MaxRetrySleep: 30 * time.Second, RetryMultiplier: 2}
+
+	client.CallOptions.RenameFolder = storageutil.StorageControlClientRetryOptions(clientConfig)
+	client.CallOptions.GetFolder = storageutil.StorageControlClientRetryOptions(clientConfig)
+	client.CallOptions.GetStorageLayout = storageutil.StorageControlClientRetryOptions(clientConfig)
+	client.CallOptions.CreateFolder = storageutil.StorageControlClientRetryOptions(clientConfig)
+	client.CallOptions.DeleteFolder = storageutil.StorageControlClientRetryOptions(clientConfig)
+	client.CallOptions.CreateManagedFolder = storageutil.StorageControlClientRetryOptions(clientConfig)
```
charith87, line 38: The managed folder retry options also should come from the application logic itself.

tools/integration_tests/util/client/control_client.go
```diff
@@ -19,13 +19,27 @@ import (
...
+	"time"
 
 	control "cloud.google.com/go/storage/control/apiv2"
 	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
 )
 
 func CreateControlClient(ctx context.Context) (client *control.StorageControlClient, err error) {
 	client, err = control.NewStorageControlClient(ctx)
+	clientConfig := &storageutil.StorageClientConfig{MaxRetrySleep: 30 * time.Second, RetryMultiplier: 2}
```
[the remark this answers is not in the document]
Tulsishah: As discussed offline this code is  part of util, part of helper functions to do setup of objects/folders in bucket to perform e2e tests on gcsfuse. So we can explicitly define value to this.

tools/integration_tests/util/client/control_client.go
Tulsishah: This is not part of application logic and only used in creating test data on bucket to perform e2e tests.

tools/integration_tests/util/client/control_client.go
```diff
@@ -19,13 +19,45 @@ import (
...
 func CreateControlClient(ctx context.Context) (client *control.StorageControlClient, err error) {
 	client, err = control.NewStorageControlClient(ctx)
+
+	client.CallOptions.RenameFolder = storageControlClientRetryOptions()
+	client.CallOptions.GetFolder = storageControlClientRetryOptions()
+	client.CallOptions.GetStorageLayout = storageControlClientRetryOptions()
+	client.CallOptions.CreateFolder = storageControlClientRetryOptions()
+	client.CallOptions.DeleteFolder = storageControlClientRetryOptions()
+	client.CallOptions.CreateManagedFolder = storageControlClientRetryOptions()
+	client.CallOptions.DeleteManagedFolder = storageControlClientRetryOptions()
```
charith87, line 57: Lets keep the scope to only Create and Delete Managed folders which are needed here. The others we can remove. That way we keep this only for cases where the workflow cannot be invoked by gcsfuse.

Tulsishah: Okay.. as and when we add other methods like Create, DeleteFolder we can make it retryable at that time. Done.

tools/integration_tests/util/client/control_client.go
```diff
@@ -19,13 +19,45 @@ import (
...
 	"strings"
+	"time"
 
 	control "cloud.google.com/go/storage/control/apiv2"
 	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googleapis/gax-go/v2"
+	"google.golang.org/grpc/codes"
 )
 
+func storageControlClientRetryOptions() []gax.CallOption {
```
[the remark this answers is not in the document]
Tulsishah: Done. Added on top.

tools/integration_tests/util/client/control_client.go
charith87: Ideally create, delete folders should not come via this route. They are valid workflows in gcsfuse so its better they are invoked that way so they are tested as well.

charith87 (approved):

tools/integration_tests/util/client/control_client.go
Tulsishah: Yeah, but sometimes we write test cases, where we create/delete data from outside GCSFuse(e.g. Storage client, gcloud) and validate the result with GCSFuse. At that point we need this.

charith87: Sure, yea we can take it case by case basis.

PR #2602, opened 2024-10-17
Added blockpool struct

vadlakondaswetha:
### Description
Added blockpool struct which is responsible for creating and maintaining blocks as per the user configuration.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/block/block.go
+++ b/internal/block/block.go
@@ -94,3 +94,18 @@ func (m *memoryBlock) Deallocate() error {
 
 	return nil
 }
+
+// createBlock creates a new block.
+func createBlock(blockSize int64) (Block, error) {
+	prot, flags := syscall.PROT_READ|syscall.PROT_WRITE, syscall.MAP_ANON|syscall.MAP_PRIVATE
+	addr, err := syscall.Mmap(-1, 0, int(blockSize), prot, flags)
+	if err != nil {
+		return nil, fmt.Errorf("mmap error: %v", err)
+	}
+
+	mb := memoryBlock{
+		buffer: addr,
+		offset: offset{0, 0},
+	}
+	return &mb, nil
+}
--- a/internal/block/block.go
+++ b/internal/block/block.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"syscall"
 )
 
 // Block represents the buffer which holds the data.
@@ -35,6 +36,8 @@ type Block interface {
 	// Reader interface helps in copying the data directly to storage.writer
 	// while uploading to GCS.
 	Reader() io.Reader
+
+	DeAllocate() error
 }
 
 // TODO: check if we need offset or just storing end is sufficient. We might need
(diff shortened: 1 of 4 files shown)


internal/block/block_pool_test.go
```diff
@@ -0,0 +1,138 @@
...
+	require.Nil(t.T(), err)
+
+	block, err := bp.Get()
+
+	require.Nil(t.T(), err)
+	require.NotNil(t.T(), block)
+	assert.Equal(t.T(), int64(0), block.Size())
+}
+
+func (t *BlockPoolTest) TestCreateBlock() {
```
ashmeenkaur, line 111: This test is same as TestGetWhenTotalBlocksIsLessThanThanMaxBlocks
Shall we remove one?

internal/block/block_pool_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+func TestBlockPoolTestSuite(t *testing.T) {
+	suite.Run(t, new(BlockPoolTest))
+}
+
+func (t *BlockPoolTest) TestInitBlockPool() {
+	bp, err := InitBlockPool(1024, 10)
+
+	require.Nil(t.T(), err)
+	assert.NotNil(t.T(), bp)
```
ashmeenkaur, line 41: Nit:
```suggestion
	require.NotNil(t.T(), bp)
```

internal/block/block_pool.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"fmt"
+	"syscall"
+)
+
+// BlockPool handles the creation of blocks as per the user configuration.
+type BlockPool struct {
+	// Channel holding free blocks.
+	blocksCh chan Block
```
charith87, line 25: *nit = Can we rename to blocksChannel or similar like freeBlocksChannel instead to signify its usage and helps in readability.

internal/block/block_pool.go
```diff
@@ -0,0 +1,97 @@
...
+				return b, nil
+			}
+		}
+	}
+}
+
+// createBlock creates a new block.
+func (ib *BlockPool) createBlock() (Block, error) {
+	prot, flags := syscall.PROT_READ|syscall.PROT_WRITE, syscall.MAP_ANON|syscall.MAP_PRIVATE
+	addr, err := syscall.Mmap(-1, 0, int(ib.blockSize), prot, flags)
```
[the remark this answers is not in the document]
charith87: +1, can we add freeBlock() function as well so its abstracted here with totalcount and other free related things.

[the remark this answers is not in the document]
vadlakondaswetha: Added DeAllocate method. 
syscall.Map will throw error when virtual address space is full. Avoids crashing in few scenarios.

internal/block/block_pool.go
vadlakondaswetha: Done.

internal/block/block_pool_test.go
vadlakondaswetha: Done.

internal/block/block_pool_test.go
vadlakondaswetha: yep. Removed.

internal/block/block.go
```diff
@@ -76,3 +79,18 @@ func (m *memoryBlock) Write(bytes []byte) error {
 func (m *memoryBlock) Reader() io.Reader {
 	return bytes.NewReader(m.buffer[0:m.offset.end])
 }
+
+func (m *memoryBlock) Deallocate() error {
```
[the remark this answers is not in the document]
vadlakondaswetha: createBlock is actually creating block and returning so other methods are called using that instance. So we can't put createBlock as a method on Block interface. 

Other option is to just put the function in the block.go file and call it from blockpool.go.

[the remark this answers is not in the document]
ashmeenkaur: > put the function in the block.go file and call it from blockpool.go.

I think this will be better. WDYT?

ashmeenkaur (approved):

PR #2604, opened 2024-10-18
[Refactoring] change all receivers to bh in bucket_handle.go

ashmeenkaur:
### Description
IDE suggested fix: change all receivers to bh in bucket_handle.go

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


BrennaEpp (commented):
Is there a reason for this change other than it being an IDE suggested fix? Both `b` and `bh` should be fine as long as it's consistent, and there seems like little reason to change what's already in use.

ashmeenkaur:
> Is there a reason for this change other than it being an IDE suggested fix? Both `b` and `bh` should be fine as long as it's consistent, and there seems like little reason to change what's already in use.

No reason apart from consistency and it being an IDE suggested fix. We are using `b` and `bh` both in this file. After this change, the receiver will be `bh` throughout the file consistently.

raj-prince (approved):

PR #2612, opened 2024-10-21
feat: support to retry for stalled read request

raj-prince:
### Description
- Integrating the go-sdk's read-stall-retry changes: https://pkg.go.dev/cloud.google.com/go/storage@v1.45.0/experimental#WithReadStallTimeout
- Disabled by default.

NA

### Testing details
1. Manual - NA
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -39,6 +39,7 @@ import (
 
 const (
 	// Used to modify the hidden options in go-sdk for read stall retry.
+	// Ref: http://shortn/_DIoIm9dDT3
 	dynamicReadReqIncreaseRateEnv   = "DYNAMIC_READ_REQ_INCREASE_RATE"
 	dynamicReadReqInitialTimeoutEnv = "DYNAMIC_READ_REQ_INITIAL_TIMEOUT"
 )
@@ -169,7 +170,6 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 			TargetPercentile: clientConfig.ReqTargetPercentile,
 		}))
 	}
-
 	return storage.NewClient(ctx, clientOpts...)
 }
 
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -39,7 +39,7 @@ import (
 
 const (
 	// Used to modify the hidden options in go-sdk for read stall retry.
-	// Ref: http://shortn/_DIoIm9dDT3
+	// Ref: https://github.com/googleapis/google-cloud-go/blob/main/storage/option.go#L30
 	dynamicReadReqIncreaseRateEnv   = "DYNAMIC_READ_REQ_INCREASE_RATE"
 	dynamicReadReqInitialTimeoutEnv = "DYNAMIC_READ_REQ_INITIAL_TIMEOUT"
 )
@@ -155,7 +155,7 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 
 	if clientConfig.ReadStallRetryConfig.Enable {
 		// Hidden way to modify the increase rate for dynamic delay algorithm in go-sdk.
-		// Ref: http://shortn/_417eTbNbKK
+		// Ref: https://github.com/googleapis/google-cloud-go/blob/main/storage/option.go#L47
 		// Temporarily we kept an option to change the increase-rate, will be removed
 		// once we get a good default.
 		err = os.Setenv(dynamicReadReqIncreaseRateEnv, strconv.FormatFloat(clientConfig.ReadStallRetryConfig.ReqIncreaseRate, 'f', -1, 64))
@@ -164,7 +164,7 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		}
 
 		// Hidden way to modify the initial-timeout of the dynamic delay algorithm in go-sdk.
-		// Ref: http://shortn/_xArUKvvGQZ
+		// Ref: https://github.com/googleapis/google-cloud-go/blob/main/storage/option.go#L62
 		// Temporarily we kept an option to change the initial-timeout, will be removed
 		// once we get a good default.
 		err = os.Setenv(dynamicReadReqInitialTimeoutEnv, clientConfig.ReadStallRetryConfig.InitialReqTimeout.String())
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -18,9 +18,11 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 
 	"cloud.google.com/go/storage"
 	control "cloud.google.com/go/storage/control/apiv2"
+	"cloud.google.com/go/storage/experimental"
 	"github.com/googleapis/gax-go/v2"
 	"github.com/googlecloudplatform/gcsfuse/v2/cfg"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
@@ -33,6 +35,12 @@ import (
 	// Side effect to run grpc client with direct-path on gcp machine.
 	_ "google.golang.org/grpc/balancer/rls"
 	_ "google.golang.org/grpc/xds/googledirectpath"
+)
+
+const (
+	// Used to modify the hidden options in go-sdk for read stall retry.
+	dynamicReadReqIncreaseRateEnv   = "DYNAMIC_READ_REQ_INCREASE_RATE"
+	dynamicReadReqInitialTimeoutEnv = "DYNAMIC_READ_REQ_INITIAL_TIMEOUT"
 )
 
 type StorageHandle interface {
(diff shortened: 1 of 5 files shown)


internal/storage/storage_handle.go
```diff
@@ -144,6 +153,23 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		clientOpts = append(clientOpts, option.WithEndpoint(clientConfig.CustomEndpoint))
 	}
 
+	if clientConfig.EnableReadStallRetry {
+		// Hidden way to modify the increase rate for dynamic delay algorithm in go-sdk.
+		// Ref: http://shortn/_417eTbNbKK
+		// Temporarily we kept an option to change the increase-rate, will be removed
+		// once we get a good default.
+		os.Setenv(dynamicReadReqIncreaseRateEnv, strconv.FormatFloat(clientConfig.ReqIncreaseRate, 'f', -1, 64))
```
kislaykishore, line 161: This method returns an error. How should it be handled?

raj-prince: Added a warning log, as these attributes are for testing purpose to find the good default. This will never be exposed to the customer and will be removed once we get the good default.

internal/storage/storage_handle.go
```diff
@@ -35,6 +37,13 @@ import (
 	_ "google.golang.org/grpc/xds/googledirectpath"
 )
 
+const (
+	// Used to modify the hidden options in go-sdk for read stall retry.
+	// Ref: http://shortn/_DIoIm9dDT3
```
gargnitingoogle, line 42: shortn links aren't available to readers outside google. This anyway is a link to a public github codebase, we should use the full github link here.

internal/storage/storage_handle.go
```diff
@@ -144,6 +153,29 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		clientOpts = append(clientOpts, option.WithEndpoint(clientConfig.CustomEndpoint))
 	}
 
+	if clientConfig.ReadStallRetryConfig.Enable {
+		// Hidden way to modify the increase rate for dynamic delay algorithm in go-sdk.
+		// Ref: http://shortn/_417eTbNbKK
```
gargnitingoogle, line 158: Same here. Switch this to full link.

internal/storage/storage_handle.go
```diff
@@ -144,6 +153,29 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
+		// Temporarily we kept an option to change the increase-rate, will be removed
+		// once we get a good default.
+		err = os.Setenv(dynamicReadReqIncreaseRateEnv, strconv.FormatFloat(clientConfig.ReadStallRetryConfig.ReqIncreaseRate, 'f', -1, 64))
+		if err != nil {
+			logger.Warnf("Error while setting the env %s: %v", dynamicReadReqIncreaseRateEnv, err)
+		}
+
+		// Hidden way to modify the initial-timeout of the dynamic delay algorithm in go-sdk.
+		// Ref: http://shortn/_xArUKvvGQZ
```
gargnitingoogle, line 167: Same here. Switch this to full link.

internal/storage/storage_handle.go
```diff
@@ -144,6 +153,29 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		clientOpts = append(clientOpts, option.WithEndpoint(clientConfig.CustomEndpoint))
 	}
 
+	if clientConfig.ReadStallRetryConfig.Enable {
+		// Hidden way to modify the increase rate for dynamic delay algorithm in go-sdk.
+		// Ref: http://shortn/_417eTbNbKK
+		// Temporarily we kept an option to change the increase-rate, will be removed
+		// once we get a good default.
+		err = os.Setenv(dynamicReadReqIncreaseRateEnv, strconv.FormatFloat(clientConfig.ReadStallRetryConfig.ReqIncreaseRate, 'f', -1, 64))
```
gargnitingoogle, line 161: Setting environment variables in the middle of the lifetime of a daemon seems like a brittle design. I am assuming that this env var is read by the go sdk. Is it not possible to pass this down to go sdk (or whoever the consumer is) some other way e.g. a function call ?

internal/storage/storage_handle.go
```diff
@@ -144,6 +153,29 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
+		if err != nil {
+			logger.Warnf("Error while setting the env %s: %v", dynamicReadReqIncreaseRateEnv, err)
+		}
+
+		// Hidden way to modify the initial-timeout of the dynamic delay algorithm in go-sdk.
+		// Ref: http://shortn/_xArUKvvGQZ
+		// Temporarily we kept an option to change the initial-timeout, will be removed
+		// once we get a good default.
+		err = os.Setenv(dynamicReadReqInitialTimeoutEnv, clientConfig.ReadStallRetryConfig.InitialReqTimeout.String())
```
gargnitingoogle, line 170: Same as above. Please switch setting env vars to some other way i possible.

internal/storage/storage_handle_test.go
```diff
@@ -253,6 +254,186 @@ func (testSuite *StorageHandleTest) TestCreateHTTPClientHandle_WithGRPCClientPro
 	assert.Contains(testSuite.T(), err.Error(), fmt.Sprintf("client-protocol requested is not HTTP1 or HTTP2: %s", cfg.GRPC))
 }
 
+func (testSuite *StorageHandleTest) TestCreateHTTPClientHandle_WithReadStallRetry() {
+	testCases := []struct {
+		name                 string
```
gargnitingoogle, line 259: nit:
```suggestion
		tcname                 string
```

internal/storage/storage_handle_test.go
```diff
@@ -253,6 +254,186 @@ func (testSuite *StorageHandleTest) TestCreateHTTPClientHandle_WithGRPCClientPro
...
+			enableReadStallRetry: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		testSuite.Run(tc.name, func() {
+			sc := storageutil.GetDefaultStorageClientConfig()
+			sc.ReadStallRetryConfig.Enable = tc.enableReadStallRetry
+
+			storageClient, err := createHTTPClientHandle(context.Background(), &sc)
```
gargnitingoogle, line 277: Is there any way to test that the returned storageClient has read-stall-retry enabled/disabled as expected ?

internal/storage/storage_handle_test.go
```diff
@@ -253,6 +254,186 @@ func (testSuite *StorageHandleTest) TestCreateHTTPClientHandle_WithGRPCClientPro
...
+		},
+	}
+
+	for _, tc := range testCases {
+		testSuite.Run(tc.name, func() {
+			sc := storageutil.GetDefaultStorageClientConfig()
+			sc.ReadStallRetryConfig.Enable = true
+			sc.ReadStallRetryConfig.InitialReqTimeout = tc.initialReqTimeout
+
+			storageClient, err := createHTTPClientHandle(context.Background(), &sc)
```
gargnitingoogle, line 306: Same as above.. Is there any way to test that the returned storageClient has requested timeout as expected ?

internal/storage/storage_handle_test.go
```diff
@@ -253,6 +254,186 @@ func (testSuite *StorageHandleTest) TestCreateHTTPClientHandle_WithGRPCClientPro
...
+	for _, tc := range testCases {
+		testSuite.Run(tc.name, func() {
+			sc := storageutil.GetDefaultStorageClientConfig()
+			sc.ReadStallRetryConfig.Enable = true
+			sc.ReadStallRetryConfig.ReqIncreaseRate = tc.reqIncreaseRate
+
+			storageClient, err := createHTTPClientHandle(context.Background(), &sc)
+
+			if tc.expectErr {
+				assert.NotNil(testSuite.T(), err)
```
gargnitingoogle, line 375: Do we know what the expected error type or string is? If yes, we should use that for more accurate checking,.

internal/storage/storageutil/test_util.go
```diff
@@ -42,5 +42,13 @@ func GetDefaultStorageClientConfig() (clientConfig StorageClientConfig) {
 		ExperimentalEnableJsonRead: false,
 		AnonymousAccess:            true,
 		EnableHNS:                  false,
+		ReadStallRetryConfig: cfg.ReadStallGcsRetriesConfig{
+			Enable:              false,
```
gargnitingoogle, line 46: nit: Put these constant default values in a config somewhere as constants rather than here where they are harder to find.

gargnitingoogle (changes requested):

internal/storage/storage_handle.go
raj-prince: As said in the comment, these are temporary change only for testing and will be removed after finding a good default.

internal/storage/storage_handle.go
raj-prince: Not possible.

internal/storage/storage_handle_test.go
raj-prince: Yes, there is a way to do this, but that requires integration with storage-test-bench server (similar to retry error).

internal/storage/storage_handle_test.go
raj-prince: Doesn't testCase itself tell the name is for test-case?

gargnitingoogle (approved):

PR #2624, opened 2024-10-23
Added bufferedwriteHandler for handling buffered writes flow

vadlakondaswetha:
### Description
Added all required method defintion to bufferedWriteHandler struct. Implementation of sync/flush will be added after uploadHandler changes are done.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -1,0 +1,127 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bufferedwrites
+
+import (
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/block"
+	"golang.org/x/sync/semaphore"
+)
+
+// Note: All the write operations take inode lock in fs.go, hence we don't need any locks here
+// as we will get write operations serially.
+
+// BufferedWriteHandler is responsible for filling up the buffers with the data
+// as it receives and handing over to uploadHandler which uploads to GCS.
+type BufferedWriteHandler struct {
+	current   block.Block
+	blockPool *block.BlockPool
+	// Total size of data buffered so far. Some part of buffered data might have
+	// been uploaded to GCS as well.
+	totalSize int64
+	// Stores the mtime value updated by kernel as part of setInodeAttributes call.
+	mtime time.Time
+}
+
+// WriteFileInfo is used as part of serving fileInode attributes (GetInodeAttributes call).
+type WriteFileInfo struct {
+	TotalSize int64
+	Mtime     time.Time
+}
+
+// InitBWHandler initialises the bufferedWriteHandler struct.
+func InitBWHandler(blockSize int64, maxBlocks int32, globalMaxBlocksSem *semaphore.Weighted) (bwh *BufferedWriteHandler, err error) {
+	bp, err := block.InitBlockPool(blockSize, maxBlocks, globalMaxBlocksSem)
+	if err != nil {
+		return
+	}
+
+	bwh = &BufferedWriteHandler{
+		current:   nil,
+		blockPool: bp,
+		totalSize: 0,
+		mtime:     time.Now(),
+	}
+	return
+}
+
+// Write writes the given data to the buffer. It writes to an existing buffer if
+// the capacity is available otherwise writes to a new buffer.
+func (wh *BufferedWriteHandler) Write(data []byte, offset int64) (err error) {
+	if offset > wh.totalSize {
+		// TODO: Will be handled as part of ordered writes.
+		return fmt.Errorf("non sequential writes")
+	}
+
+	dataWritten := 0
+	for dataWritten < len(data) {
+		if wh.current == nil {
+			wh.current, err = wh.blockPool.Get()
+			if err != nil {
+				return
+			}
+		}
+
+		remainingBlockSize := float64(wh.blockPool.BlockSize()) - float64(wh.current.Size())
+		pendingDataForWrite := float64(len(data)) - float64(dataWritten)
+		bytesToCopy := int(math.Min(remainingBlockSize, pendingDataForWrite))
+		err := wh.current.Write(data[dataWritten : dataWritten+bytesToCopy])
+		if err != nil {
+			return err
+		}
+
+		dataWritten += bytesToCopy
+
+		if wh.current.Size() == wh.blockPool.BlockSize() {
+			// TODO: err = trigger upload
+			if err != nil {
+				return err
+			}
+			wh.current = nil
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -0,0 +1,127 @@
...
+	for dataWritten < len(data) {
+		if wh.current == nil {
+			wh.current, err = wh.blockPool.Get()
+			if err != nil {
+				return
+			}
+		}
+
+		remainingBlockSize := float64(wh.blockPool.BlockSize()) - float64(wh.current.Size())
+		pendingDataForWrite := float64(len(data)) - float64(dataWritten)
```
gargnitingoogle, line 81: Can `remainingBlockSize` be 0 here? If yes, then it's needed to be handled before the next 2-3 lines.

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -0,0 +1,127 @@
...
+		// TODO: Will be handled as part of ordered writes.
+		return fmt.Errorf("non sequential writes")
+	}
+
+	dataWritten := 0
+	for dataWritten < len(data) {
+		if wh.current == nil {
+			wh.current, err = wh.blockPool.Get()
+			if err != nil {
+				return
```
gargnitingoogle, line 76: The error should be a bit more descriptive to clarify what went wrong, rather than just forwarding the downstream error. E.g.

```suggestion
				return fmt.Errorf("failed to get current block to write to: %w", err)
```

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -0,0 +1,127 @@
...
+		pendingDataForWrite := float64(len(data)) - float64(dataWritten)
+		bytesToCopy := int(math.Min(remainingBlockSize, pendingDataForWrite))
+		err := wh.current.Write(data[dataWritten : dataWritten+bytesToCopy])
+		if err != nil {
+			return err
+		}
+
+		dataWritten += bytesToCopy
+
+		if wh.current.Size() == wh.blockPool.BlockSize() {
```
gargnitingoogle, line 90: nit: It would be good to define a capacity variable in `block.Block`. Then you could support something like `wh.current.IsFull()` or `wh.current.RemainingSize()` rather than using `wh.blockPool.BlockSize()` everywhere. 
Secondary: Then you could even have blocks of different sizes, though not sure if that's supported by the HLD here.

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	wh.totalSize += int64(dataWritten)
+	return
+}
+
+// Sync uploads all the pending full buffers to GCS.
+func (wh *BufferedWriteHandler) Sync() (err error) {
+	// TODO: Will be added after uploadHandler changes are done.
+	return nil
```
gargnitingoogle, line 106: nit:

```suggestion
	return fmt.Errorf("not implemented")
```

And in other unimplemented function below.

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Flush finalizes the upload.
+func (wh *BufferedWriteHandler) Flush() (err error) {
+	// TODO: Will be added after uploadHandler changes are done.
+	return nil
+}
+
+// SetMtime stores the mtime with the bufferedWriteHandler.
+func (wh *BufferedWriteHandler) SetMtime(mtime time.Time) {
+	wh.mtime = mtime
```
gargnitingoogle, line 117: Is it possible to have `mtime` < `wh.mtime` say due to multiple parallel writes to the same object  ? If yes, then this isn't a good update .

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// WriteFileInfo return the file info i.e, how much data is buffered so far
```
gargnitingoogle, line 120:
```suggestion
// WriteFileInfo returns the file info i.e, how much data has been buffered so far
```

internal/bufferedwrites/buffered_write_handler_test.go
```diff
@@ -0,0 +1,80 @@
...
+	require.Nil(testSuite.T(), err)
+	fileInfo := testSuite.bwh.WriteFileInfo()
+	assert.Equal(testSuite.T(), testSuite.bwh.mtime, fileInfo.Mtime)
+	assert.Equal(testSuite.T(), int64(size), fileInfo.TotalSize)
+}
+
+func (testSuite *BufferedWriteTest) TestWriteDataSizeGreaterThanBlockSize() {
+	size := 2000
+	data := strings.Repeat("A", size)
+	err := testSuite.bwh.Write([]byte(data), 0)
```
gargnitingoogle, line 74: Could you add a test where you pass an empty byte-slice to `testSuite.bwh.Write(...)` ?

internal/bufferedwrites/buffered_write_handler.go
vadlakondaswetha: Since i can't paste the lld link here, adding the conents here. Mtime is valid only till the file is closed and not synced to GCS for buffered writes. hence we need not worry about parallel writes in this case. And also application doesn't set mtime, its the kernel which does it. So irrespective of number of application threads, mtime calls will be serialized to GCSFuse.

> SetInodeAttributes has two operations -> setting mtime and truncating the file.
> Mtime
> Mtime is stored as part of a metadata map in GCS object properties. If mtime is not present in the metadata map we will use objects lastUpdated timestamp as mtime when serving inode attributes to Kernel.
> Current Flow
> When writing to large files, the kernel calls setInodeAttributes with mtime multiple times to update the last modified time of the file. During write, mtime is stored locally and updated to GCS during flush/sync calls.
> Proposed Flow
> Incase of resumable upload api, we can only set the object attributes during upload start and can’t update them during the rest of the call (i.e., during further chunk uploads and finalization). Hence we can’t update the mtime sent by the kernel during writeflow to GCS. One option is to keep track of mtime and then invoke updateObject() call to GCS after the resumable upload is done. But serving objects lastUpdated time as mtime is also not a problem and we can avoid extra calls to GCS during file write. 
> 
> We will keep track of mtime in bufferedWriteHandler and serve in getInodeAttributes before the file is synced to GCS. After the file is synced, we will use the object's lastUpdatedAt timestamp.  Changes are required to store mtime field in bufferedWritesHandler if buffered writes are enabled.
>

internal/bufferedwrites/buffered_write_handler.go
vadlakondaswetha: Blocks of different sizes are not going to be supported. At this point i dont want to introduce new functions for just for one usecase. Anyways this is going to get changed for ordered writes. Will keep it as is for now.

gargnitingoogle: ok, your call.

internal/bufferedwrites/buffered_write_handler.go
gargnitingoogle: > we have that check below

If by this you meant `if wh.current.Size() == wh.blockPool.BlockSize()`, I noticed. I am wondering though if `wh.current.Write(...)` works with an empty slice which is the case of when bytesToCopy is 0. 

Also, size==blockSize  checks for the size being full after the write, so it doesn't avoid the unnecessary `wh.current.Write` in that case. Though bytesToCopy being 0 at this point is most likely rare.

gargnitingoogle: Based on the newly added test in internal/bufferedwrites/buffered_write_handler_test.go, I guess wh.current.Write() works with empty-slice. So, marking this resolved.

ashmeenkaur (approved):

PR #2625, opened 2024-10-24
Fix flaky managed folder test

Tulsishah:
### Description
- Created a file in the temporary directory with a random suffix because tests were intermittently failing due to permission overrides during parallelization.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/managed_folders/test_helper.go
+++ b/tools/integration_tests/managed_folders/test_helper.go
@@ -78,7 +78,7 @@ func providePermissionToManagedFolder(bucket, managedFolderPath, serviceAccount,
 
 	f, err := os.CreateTemp(os.TempDir(), "iam-policy-*.json")
 	if err != nil {
-		t.Fatalf(fmt.Sprintf("Error in creating file for iam-policy : %v", err))
+		t.Fatalf("error in creating file for iam-policy : %v", err)
 	}
 	fileName := path.Join(os.TempDir(), f.Name())
 	defer operations.RemoveFile(fileName)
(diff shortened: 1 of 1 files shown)


kislaykishore (dismissed):

tools/integration_tests/managed_folders/test_helper.go
```diff
@@ -76,7 +76,7 @@ func providePermissionToManagedFolder(bucket, managedFolderPath, serviceAccount,
 		t.Fatalf(fmt.Sprintf("Error in marshal the data into JSON format: %v", err))
 	}
 
-	localIAMPolicyFilePath := path.Join(os.Getenv("HOME"), "iam_policy.json")
+	localIAMPolicyFilePath := path.Join(os.TempDir(), "iam_policy-"+setup.GenerateRandomString(5)+".json")
```
gargnitingoogle, line 79: Please use the golang-alternative, which is much cleanear and better-managed.

```suggestion
	localIAMPolicyFilePath := os.CreateTemp(os.TempDir(), "iam_policy-*.json")
```

Also, are we deleting this file somewhere ? If not, we should, for cleanup.

gargnitingoogle: Ref https://pkg.go.dev/os#CreateTemp for more details.

gargnitingoogle (changes requested):

ashmeenkaur (approved):

PR #2629, opened 2024-10-25
adding gcs call log for storage layout api

anushka567:
### Description
[Updated]
Since the logging information was incorrect in the previous implementation described below, now we only add a log (at info level) for actual getStorageLayout API call and report the time taken to serve the request.

[Previous implementation detail]
As part of HNS , storage layout API is called, logs for this does not come at TRACE severity unlike other GCS calls. Hence, added the logic for the same so that when log severity is set to TRACE, then storage layout api calls are also logged as gcs calls. 

Any error faced during fetching the bucket type is already logged at error level, and will show up in the logs separately . However, as for the GCS call log, the storage layout api returns with status OK even when error is faced and logged immediately afterwards due to the nature of implementation.

For e.g:
```
{"timestamp":{"seconds":1729840243,"nanos":34938039},"severity":"TRACE","message":"gcs: Req              0x0: <- GetStorageLayout(\"non-existent-bucket\")"}
{"timestamp":{"seconds":1729840243,"nanos":653348262},"severity":"ERROR","message":"Error returned from GetStorageLayout: rpc error: code = NotFound desc = The specified bucket does not exist."}
{"timestamp":{"seconds":1729840243,"nanos":653398322},"severity":"TRACE","message":"gcs: Req              0x0: -> GetStorageLayout(\"non-existent-bucket\") (618.464263ms): OK"}
```

NA

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - Yes

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
 	"cloud.google.com/go/storage"
 	control "cloud.google.com/go/storage/control/apiv2"
@@ -64,15 +65,17 @@ func (bh *bucketHandle) BucketType() gcs.BucketType {
 			bh.bucketType = gcs.NonHierarchical
 			return bh.bucketType
 		}
-
+		startTime := time.Now()
+		logger.Infof("GetStorageLayout <- (%s)", bh.bucketName)
 		storageLayout, err := bh.getStorageLayout()
-
+		duration := time.Since(startTime)
 		// In case bucket does not exist, set type unknown instead of panic.
 		if err != nil {
 			bh.bucketType = gcs.Unknown
 			logger.Errorf("Error returned from GetStorageLayout: %v", err)
 			return bh.bucketType
 		}
+		logger.Infof("GetStorageLayout -> (%s) %v msec", bh.bucketName, duration.Milliseconds())
 
 		hierarchicalNamespace := storageLayout.GetHierarchicalNamespace()
 		if hierarchicalNamespace != nil && hierarchicalNamespace.Enabled {
--- a/internal/storage/debug_bucket.go
+++ b/internal/storage/debug_bucket.go
@@ -126,16 +126,6 @@ func (b *debugBucket) Name() string {
 }
 
 func (b *debugBucket) BucketType() gcs.BucketType {
-	var err error
-
-	// BucketType internally calls  GetStorageLayout API from control client , hence
-	// logging this as a GetStorageLayout call.
-	id, desc, start := b.startRequest("GetStorageLayout(%q)", b.Name())
-	// In case getStorageLayout call returns error , the bucket type is set to unknown and
-	// underlying method does not panic.Since no error is returned here, gcs call is
-	// logged with error set to nil in this case.
-	defer b.finishRequest(id, desc, start, &err)
-
 	return b.wrapped.BucketType()
 }
 
--- a/internal/storage/debug_bucket.go
+++ b/internal/storage/debug_bucket.go
@@ -126,6 +126,12 @@ func (b *debugBucket) Name() string {
 }
 
 func (b *debugBucket) BucketType() gcs.BucketType {
+	id, desc, start := b.startRequest("GetStorageLayout(%q)", b.Name())
+	var err error
+	// In case getStorageLayout call returns error , the bucket type is set to unknown and
+	// underlying method does not panic.Since no error is returned here, gcs call is
+	// logged with error set to nil in this case.
+	defer b.finishRequest(id, desc, start, &err)
 	return b.wrapped.BucketType()
 }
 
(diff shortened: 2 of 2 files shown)


internal/storage/debug_bucket.go
```diff
@@ -126,6 +126,12 @@ func (b *debugBucket) Name() string {
 }
 
 func (b *debugBucket) BucketType() gcs.BucketType {
+	id, desc, start := b.startRequest("GetStorageLayout(%q)", b.Name())
+	var err error
+	// In case getStorageLayout call returns error , the bucket type is set to unknown and
```
Tulsishah, line 131: nit:
```
var err error

id, desc, start := b.startRequest("GetStorageLayout(%q)", b.Name())
defer b.finishRequest(id, desc, start, &err)

return b.wrapped.BucketType()
```

Also add comment that bucketType internally calls GetStorageLayout from control client to avoild confusion why we kept `GetStorageLayout` instead of BucketType in log.

anushka567: Done.

Tulsishah (approved):

internal/storage/debug_bucket.go
```diff
@@ -126,6 +126,16 @@ func (b *debugBucket) Name() string {
 }
 
 func (b *debugBucket) BucketType() gcs.BucketType {
+	var err error
+
+	// BucketType internally calls  GetStorageLayout API from control client , hence
+	// logging this as a GetStorageLayout call.
```
ankitaluthra1, line 132: Will the call be logged even if GetStorageLayout GCS call has not happened ? BucketType() only calls the api once and uses the type throughout the lifecycle of mount

anushka567: Yes. Even when the actual network call has not happened, this is logged as a GCS call (time of request completion is considerably less).
Another scenario I am unsure about is when the first call comes in for a non hierarchical bucket with nil control client, 
```
if bh.bucketType == gcs.Nil {
		if bh.controlClient == nilControlClient {
			bh.bucketType = gcs.NonHierarchical
			return bh.bucketType
		}
```
then GetStorageLayout api is not called, but this is also logged with minimal time taken for serving request.

ankitaluthra1: then the logging information is not correct, the log should only come when gcs call happens

internal/storage/bucket_handle.go
```diff
@@ -64,15 +65,17 @@ func (bh *bucketHandle) BucketType() gcs.BucketType {
 			bh.bucketType = gcs.NonHierarchical
 			return bh.bucketType
 		}
-
+		startTime := time.Now()
+		logger.Infof("\"GetStorageLayout <- (%s)\"", bh.bucketName)
```
gargnitingoogle, line 69: Why do we need the quotes around the log message ? Isn't that automatically added by the logger ?

internal/storage/bucket_handle.go
```diff
@@ -64,15 +65,17 @@ func (bh *bucketHandle) BucketType() gcs.BucketType {
...
 		storageLayout, err := bh.getStorageLayout()
-
+		duration := time.Since(startTime)
 		// In case bucket does not exist, set type unknown instead of panic.
 		if err != nil {
 			bh.bucketType = gcs.Unknown
 			logger.Errorf("Error returned from GetStorageLayout: %v", err)
 			return bh.bucketType
 		}
+		logger.Infof("\"GetStorageLayout -> (%s)\" %v msec", bh.bucketName, duration.Milliseconds())
```
gargnitingoogle, line 78: Do we need to explicitly calculate and print duration, when every logger.Infof call anyway prints out time ? Moreover these calls are consecutive, so it should be easy to just manually spot the duration if needed.

gargnitingoogle (changes requested):

internal/storage/bucket_handle.go
anushka567: This log was added here as a part of investigating why mount time for gcsfuse suddenly spiked to 10 sec  from a few msec . While it might be easier to manually spot the duration here, explicitly stating will help in debugging faster when required (Also maintaining consistency with other GCS call logs). Since we discussed [here ](https://github.com/GoogleCloudPlatform/gcsfuse/pull/2629#discussion_r1828007082) that it is not logically correct to log this GCS call at debug_bucket.go , hence the log is added here at info level.

internal/storage/bucket_handle.go
anushka567: Removed.

internal/storage/debug_bucket.go
anushka567: Resolving since as per offline discussion, logging logic is now in bucket_handle.go , so , this issue is taken care of.

internal/storage/bucket_handle.go
```diff
@@ -64,15 +65,17 @@ func (bh *bucketHandle) BucketType() gcs.BucketType {
 			bh.bucketType = gcs.NonHierarchical
 			return bh.bucketType
 		}
-
+		startTime := time.Now()
```
Tulsishah, line 68: I think Line#68 and 69 should be swapped

Tulsishah (approved):

ankitaluthra1 (approved):

gargnitingoogle (approved):
LGTM

PR #2641, opened 2024-10-29
Handle SIGTERM to gracefully unmount

kislaykishore:
SIGTERM is a termination signal that's sent by system (like SIGINT is sent by user). For instance, `docker stop` command uses it to give the container a chance to gracefully terminate.
Handle it to gracefully unmount GCSFuse.

b/376174285

### Testing details
1. Manual - Tested with and without flag
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -28,6 +28,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/googlecloudplatform/gcsfuse/v2/cfg"
 	"github.com/googlecloudplatform/gcsfuse/v2/common"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/canned"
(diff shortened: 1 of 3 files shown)


cmd/legacy_main.go
```diff
@@ -53,22 +55,32 @@ const (
 // Helpers
 ////////////////////////////////////////////////////////////////////////
 
-func registerSIGINTHandler(mountPoint string) {
+func registerTerminatingSignalHandler(mountPoint string, c *cfg.Config) {
 	// Register for SIGINT.
 	signalChan := make(chan os.Signal, 1)
-	signal.Notify(signalChan, os.Interrupt)
+	signal.Notify(signalChan, unix.SIGINT)
```
raj-prince, line 61: shouldn't we use syscall.SIGTERM and syscall.SIGINT instead of unix.SIGINT?

raj-prince (commented):
LGTM overall, just one comment.

cmd/legacy_main.go
kislaykishore: I changed back unix.SIGINT to os.Interrupt since it helps make the code platform independent. Regarding syscall.SIGTERM, the documentation of syscall recommends using [golang.org/x/sys](https://pkg.go.dev/golang.org/x/sys) wherever possible. So, I'd prefer using unix.SIGTERM.

raj-prince: Got it, thanks!
Pasting for future reference, why didn't we use syscall instead? - https://go.googlesource.com/proposal/+/refs/heads/master/design/freeze-syscall.md

raj-prince (approved):

PR #2646, opened 2024-11-03
chore: upgrade go-stroage from 1.45.0 to 1.46.0

raj-prince:
### Description
Upgrading go-sdk version from 1.45.0 to 1.46.0

Why? - Last go-sdk version 1.45.0 has a regression in gRPC because of that gRPC directpath stopped working in GCSFuse. Go-storage 1.46.0 contains the fix for that.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,9 +4,9 @@ go 1.23.2
 
 require (
 	cloud.google.com/go/compute/metadata v0.5.2
-	cloud.google.com/go/secretmanager v1.14.2
 	cloud.google.com/go/iam v1.2.2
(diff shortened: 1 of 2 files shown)


raj-prince:
image

ashmeenkaur (approved):
Is it possible to add tests for this use case to catch this early in future?

raj-prince:
> Is it possible to add tests for this use case to catch this early in future?

Yes, but a little hard to validate and ongoing discussion happening for this. We have a plan to add e2e test for this.

PR #2647, opened 2024-11-04
Add cloud-metrics-export-interval-secs flag

kislaykishore:
* This flag controls the export interval of metrics to GCP cloud monitoring.
    * Deprecate stackdriver-export-interval flag in favor of cloud-metrics-export-interval-secs.

b/377176812

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -86,6 +86,12 @@ func resolveStreamingWriteConfig(w *WriteConfig) {
 	}
 }
 
+func resolveCloudMetricsUploadIntervalSecs(m *MetricsConfig) {
+	if m.CloudMetricsExportIntervalSecs == 0 {
+		m.CloudMetricsExportIntervalSecs = int64(m.StackdriverExportInterval.Seconds())
+	}
+}
+
 // Rationalize updates the config fields based on the values of other fields.
 func Rationalize(v isSet, c *Config) error {
 	var err error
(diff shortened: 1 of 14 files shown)


cfg/rationalize_test.go
```diff
@@ -371,3 +371,54 @@ func TestRationalize_WriteConfig(t *testing.T) {
...
+
+func TestRationalizeMetricsConfig(t *testing.T) {
+	t.Parallel()
+	testCases := []struct {
+		name     string
+		config   *Config
+		expected int64
+	}{
+		{
+			name: "both_0",
```
charith87, line 383: Can we add one for both unset

cfg/rationalize.go
```diff
@@ -86,6 +86,12 @@ func resolveStreamingWriteConfig(w *WriteConfig) {
 	}
 }
 
+func resolveCloudMetricsUploadIntervalSecs(m *MetricsConfig) {
+	if m.CloudMetricsExportIntervalSecs == 0 {
+		m.CloudMetricsExportIntervalSecs = int64(m.StackdriverExportInterval.Seconds())
```
charith87, line 91: what happens if stackDriverExportInterval is not set

kislaykishore: It'll be 0 by default.

cfg/rationalize_test.go
kislaykishore: Both will be set to 0. So it's already covered.

charith87 (approved):

PR #2656, opened 2024-11-05
Refactor OpenCensus exporters lifecycle

kislaykishore:
Use a shutdown function to aggregate the shutdown functions for all running exporters and then shut them down. This refactoring will make toggling between OC and OTel exporters easier.

b/377410281

### Testing details
1. Manual - Verified that the exporters get started as expected and are shutdown on unmount.
2. Unit tests - NA
3. Integration tests - NA


internal/monitor/exporter.go
```diff
@@ -25,20 +24,72 @@ import (
...
 	"contrib.go.opencensus.io/exporter/prometheus"
 	"contrib.go.opencensus.io/exporter/stackdriver"
+	"github.com/googlecloudplatform/gcsfuse/v2/cfg"
+	"github.com/googlecloudplatform/gcsfuse/v2/common"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
 	"go.opencensus.io/stats/view"
 )
 
-var stackdriverExporter *stackdriver.Exporter
+func maybeStartStackdriverExporter(exporterIntervalSecs int64) common.ShutdownFn {
```
vadlakondaswetha, line 33: nit: mayBe in the method name sounds little different to me. I have this for library functions but in our case, its just about not failing the mount flow if the exporter is not started. Can we keep it simple startStackDriverExporter.

internal/monitor/exporter.go
```diff
@@ -25,20 +24,72 @@ import (
...
+	if exporterIntervalSecs <= 0 {
```
vadlakondaswetha, line 34: do you want to add some info logs here as well to know why the exporter hasn't started?

vadlakondaswetha (dismissed):

vadlakondaswetha (approved):

PR #2627, opened 2024-10-24
Reducing list API payload

anushka567:
### Description
Fetching only selected attributes using [query.SetAttrSelection](https://github.com/googleapis/google-cloud-go/blob/storage/v1.45.0/storage/storage.go#L1784) for ListObjects API which are sufficient for serving the request and reduce the payload for network call.
Also included :  necessary changes in the fake bucket tests.

Perf Metrics for listing : 
| Test Directory Size ( in number of files) | Mean Latency(Master) (in msec) | Mean Latency(reduced payload) (in msec) | Percentage Improvement |
| ------------------- | ---------------------- | --------------------------------- | ------------------------- |  
| 1000 | 61.153  | 56.325| 7.89% |
| 10000 | 462.775 | 410.701 | 11.25%|
| 50000 |  2373.293 |2103.235 | 11.38%|
| 100000 | 4677.908 | 4354.92| 6.90% |
| 200000 |9417.683 |8628.266 | 8.38% |

Note: More detailed metrics [here](https://docs.google.com/spreadsheets/d/101rbvNWjU55sTUaNP9p86D_g1A_2vxCjq7FSmyPDuzo/edit?usp=sharing)
### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

PR #2657, opened 2024-11-06
replacing gcs.Object with gcs.MinObject in gcs.Listing

anushka567:
### Description
We are fetching only MinObject attributes for reducing list API payload in this [PR](https://github.com/GoogleCloudPlatform/gcsfuse/pull/2627) , creating a gcs.Object using the attributes fetched, converting this gcs.Object to gcs.MinObject , and ultimately , storing minObject in cache for serving listobject requests, the conversion from gcs.Object to gcs.MinObject is redundant.

Hence, in this PR, the gcs.Listing struct is modified to store gcs.MinObject instead of gcs.Object so that  attributes fetched  can be used to directly populate the minObject and stored in cache.
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -85,6 +85,17 @@ func (b *fastStatBucket) insertMultiple(objs []*gcs.Object) {
 	}
 }
 
+// LOCKS_EXCLUDED(b.mu)
+func (b *fastStatBucket) insertMultipleMinObjects(minObjs []*gcs.MinObject) {
```
vadlakondaswetha, line 89: do we need the above method? if not, why not modify the above method?

internal/storage/storageutil/object_attrs.go
```diff
@@ -104,6 +104,27 @@ func ObjectAttrsToBucketObject(attrs *storage.ObjectAttrs) *gcs.Object {
 	}
 }
 
+func ObjectAttrsToBucketMinObject(attrs *storage.ObjectAttrs) *gcs.MinObject {
```
vadlakondaswetha, line 107: not sure why we call it BucketMinObject.
If there is no reason, can you call the method as objectAttrsToMinObject

internal/storage/caching/fast_stat_bucket.go
anushka567: The insertMultiple method, while not directly called, is still implicitly used through functions like createObject, composeObject, etc., which rely on the gcs.Object type.  Since listing operations use gcs.MinObject, we can directly utilize insertMultipleMinObjects.

Converting gcs.Object to gcs.MinObject and calling insertMultipleMinObjects is possible to get rid of insertMultiple method but it would effectively mean the same thing.

internal/storage/storageutil/object_attrs.go
anushka567: Updated.

internal/storage/caching/fast_stat_bucket.go
vadlakondaswetha: This becomes inconsistent. In cache we should either insert Minobject or object. Not both.

anushka567: In cache, we are always inserting minObject.. only the argument to the method differs 
insertMultiple.. takes in gcs.Object , converts it into minObject and then inserts into cache.
insertMultipleMin.. takes in gcs.MInObject and directly inserts into cache

vadlakondaswetha (approved):

PR #2666, opened 2024-11-08
Override non-HNS long haul test for gRPC

anushka567:
### Description
This PR configures non-HNS long haul tests to use gRPC for PyTorch V2 and TensorFlow (excluding PyTorch V1.12).
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/ml_tests/tf/resnet/setup_scripts/setup_container.sh
+++ b/perfmetrics/scripts/ml_tests/tf/resnet/setup_scripts/setup_container.sh
@@ -51,6 +51,7 @@ then
   DIR=${DIR}_${BUCKET_TYPE}
 elif [ ${BUCKET_TYPE} == "non-hns" ];
 then
+  #To validate the gRPC client, we have temporarily changed the non-HNS long haul test.
   echo "gcs-connection:
           client-protocol: grpc
          " >> /tmp/gcsfuse_config.yaml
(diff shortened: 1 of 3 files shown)


perfmetrics/scripts/ml_tests/tf/resnet/setup_scripts/setup_container.sh
```diff
@@ -49,6 +49,11 @@ then
   TEST_BUCKET="gcsfuse-ml-data-hns-central1"
   echo "enable-hns: true" >> /tmp/gcsfuse_config.yaml
   DIR=${DIR}_${BUCKET_TYPE}
+elif [ ${BUCKET_TYPE} == "non-hns" ];
```
raj-prince, line 52: How many types of bucket do we support currently? If only two "hns" and "non-hns" then we don't need to add elif right?

anushka567: Currently,we support only 2. Should I remove the elif , I added it just in case bucket type which is passed to this script  is any invalid value/nil

raj-prince: Okay, we can keep it. Just add a comment here, as it doesn't look logical to add gRPC config for non-hns bucket.

anushka567: Added.

raj-prince (approved):

PR #2667, opened 2024-11-08
Implement OTel Prometheus metrics exporter

kislaykishore:
Add support for OTel Prometheus exporter. We'll add OTel metrics support in a subsequent PR and as part of it add integration tests for the exporter.

b/378088556

### Testing details
1. Manual - Tested manually that the OC exporters continue to work. Verified that Prometheus exporter exposes the metrics at /metrics.
2. Unit tests - NA
3. Integration tests - TBD once the metrics are migrated to OTel.

### DIFF
--- a/cfg/validate.go
+++ b/cfg/validate.go
@@ -177,8 +177,9 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	if m.StackdriverExportInterval != 0 && m.CloudMetricsExportIntervalSecs != 0 {
 		return fmt.Errorf("exactly one of stackdriver-export-interval and cloud-metrics-export-interval-secs must be specified")
 	}
-	if m.PrometheusPort > 65535 {
-		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: 65535")
+	const maxPortNumber = 65535
+	if m.PrometheusPort > maxPortNumber {
+		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: 65535 but received: %d instead", m.PrometheusPort)
 	}
 	return nil
 }
--- a/cfg/validate.go
+++ b/cfg/validate.go
@@ -16,6 +16,7 @@ package cfg
 
 import (
 	"fmt"
+	"math"
 
 	cacheutil "github.com/googlecloudplatform/gcsfuse/v2/internal/cache/util"
 )
@@ -177,9 +178,9 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	if m.StackdriverExportInterval != 0 && m.CloudMetricsExportIntervalSecs != 0 {
 		return fmt.Errorf("exactly one of stackdriver-export-interval and cloud-metrics-export-interval-secs must be specified")
 	}
-	const maxPortNumber = 65535
+	const maxPortNumber = math.MaxUint16
 	if m.PrometheusPort > maxPortNumber {
-		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: 65535 but received: %d instead", m.PrometheusPort)
+		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: %d but received: %d instead", maxPortNumber, m.PrometheusPort)
 	}
 	return nil
 }
--- a/cfg/validate.go
+++ b/cfg/validate.go
@@ -177,6 +177,9 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	if m.StackdriverExportInterval != 0 && m.CloudMetricsExportIntervalSecs != 0 {
 		return fmt.Errorf("exactly one of stackdriver-export-interval and cloud-metrics-export-interval-secs must be specified")
 	}
+	if m.PrometheusPort > 65535 {
+		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: 65535")
+	}
 	return nil
 }
 
--- a/cfg/validate_test.go
+++ b/cfg/validate_test.go
@@ -539,6 +539,34 @@ func TestValidateMetrics(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name: "too_high_prom_port",
+			metricsConfig: MetricsConfig{
+				PrometheusPort: 100000,
+			},
+			wantErr: true,
+		},
+		{
+			name: "valid_prom_port",
+			metricsConfig: MetricsConfig{
+				PrometheusPort: 5550,
+			},
+			wantErr: false,
+		},
+		{
+			name: "prom_disabled_0",
+			metricsConfig: MetricsConfig{
+				PrometheusPort: 0,
+			},
+			wantErr: false,
+		},
+		{
+			name: "prom_disabled_less_than_0",
+			metricsConfig: MetricsConfig{
+				PrometheusPort: -21,
+			},
+			wantErr: false,
+		},
 	}
 	for _, tc := range testCases {
 		tc := tc
--- a/internal/monitor/otelexporters.go
+++ b/internal/monitor/otelexporters.go
@@ -54,7 +54,7 @@ func setupPrometheus(port int64) ([]metric.Option, common.ShutdownFn) {
 	}
 	exporter, err := prometheus.New(prometheus.WithoutUnits(), prometheus.WithoutCounterSuffixes(), prometheus.WithoutScopeInfo(), prometheus.WithoutTargetInfo())
 	if err != nil {
-		logger.Errorf("Error while creating prometheus exporter")
+		logger.Errorf("Error while creating prometheus exporter:%v", err)
 		return nil, nil
 	}
 	shutdownCh := make(chan context.Context)
@@ -62,7 +62,9 @@ func setupPrometheus(port int64) ([]metric.Option, common.ShutdownFn) {
 	go serveMetrics(port, shutdownCh, done)
 	return []metric.Option{metric.WithReader(exporter)}, func(ctx context.Context) error {
 		shutdownCh <- ctx
+		close(shutdownCh)
 		<-done
+		close(done)
 		return nil
 	}
 }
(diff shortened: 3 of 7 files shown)


internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,117 @@
...
+	return common.JoinShutdownFunc(meterProvider.Shutdown, shutdownFn)
+}
+
+func setupPrometheus(port int64) ([]metric.Option, common.ShutdownFn) {
+	if port <= 0 {
+		return nil, nil
+	}
+	exporter, err := prometheus.New(prometheus.WithoutUnits(), prometheus.WithoutCounterSuffixes(), prometheus.WithoutScopeInfo(), prometheus.WithoutTargetInfo())
+	if err != nil {
+		logger.Errorf("Error while creating prometheus exporter")
```
gargnitingoogle, line 57: Don't we want to print the value/message of err at least ?

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, nil
```
gargnitingoogle, line 58: Why is this error being ignored ?

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func serveMetrics(port int64, shutdownCh <-chan context.Context, done chan<- interface{}) {
+	logger.Infof("serving metrics at localhost:%d/metrics", port)
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+	prometheusServer := &http.Server{
+		Addr:           fmt.Sprintf(":%d", port),
+		Handler:        mux,
+		ReadTimeout:    10 * time.Second,
```
gargnitingoogle, line 77: nit: Define these value as constants at the top for ease of updating.

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,117 @@
...
+	} else {
+		options = append(options, metric.WithResource(res))
+	}
+	meterProvider := metric.NewMeterProvider(options...)
+	otel.SetMeterProvider(meterProvider)
+	return common.JoinShutdownFunc(meterProvider.Shutdown, shutdownFn)
+}
+
+func setupPrometheus(port int64) ([]metric.Option, common.ShutdownFn) {
+	if port <= 0 {
```
gargnitingoogle, line 52: Isn't there any upper limit on the port number ? I thought ports are generally limited ranges for avoiding clashes with other system-level processes/connections ?

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,117 @@
...
+	prometheusServer := &http.Server{
+		Addr:           fmt.Sprintf(":%d", port),
+		Handler:        mux,
+		ReadTimeout:    10 * time.Second,
+		WriteTimeout:   10 * time.Second,
+		MaxHeaderBytes: 1 << 20,
+	}
+	go func() {
+		if err := prometheusServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			logger.Errorf("Failed to start Prometheus server: %v", err)
```
gargnitingoogle, line 83: If server failed to start, would we still get the shutdown error callback ? If not, then the next go-routine the one which waits for the shutdownCh to be set will never move forward, which will cause a deadlock . If this is the case, then the two goroutines should be fused into one. I mean in that case, the waiting on shutDownCh should be started only if the server started successfully .

gargnitingoogle (changes requested):

internal/monitor/otelexporters.go
kislaykishore: We don't want mount to fail if prometheus doesn't come up.

internal/monitor/otelexporters.go
kislaykishore: Thanks! Done.

internal/monitor/otelexporters.go
kislaykishore: Added validations to enforce an upper limit on port.

cfg/validate.go
```diff
@@ -177,6 +177,9 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	if m.StackdriverExportInterval != 0 && m.CloudMetricsExportIntervalSecs != 0 {
 		return fmt.Errorf("exactly one of stackdriver-export-interval and cloud-metrics-export-interval-secs must be specified")
 	}
+	if m.PrometheusPort > 65535 {
```
raj-prince, line 180: Shouldn't we declare a constant for 65535? Also, we should print the prometheus-port received from user in the error.

kislaykishore: Yes. Thanks

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,119 @@
...
+		shutdownCh <- ctx
+		close(shutdownCh)
+		<-done
+		close(done)
+		return nil
+	}
+}
+
+func serveMetrics(port int64, shutdownCh <-chan context.Context, done chan<- interface{}) {
+	logger.Infof("serving metrics at localhost:%d/metrics", port)
```
raj-prince, line 73: nit: serving -> Serving

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func serveMetrics(port int64, shutdownCh <-chan context.Context, done chan<- interface{}) {
+	logger.Infof("Serving metrics at localhost:%d/metrics", port)
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+	prometheusServer := &http.Server{
+		Addr:           fmt.Sprintf(":%d", port),
+		Handler:        mux,
+		ReadTimeout:    10 * time.Second,
```
raj-prince, line 79: How did we come up with these defaults?

internal/monitor/otelexporters.go
kislaykishore: Done

raj-prince (approved):
LGTM.

internal/monitor/otelexporters.go
kislaykishore: I think these are just some sane values that were chosen. The same timeouts are currently being used in the OC exporters as well.

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,119 @@
...
+		if err := prometheusServer.Shutdown(ctx); err != nil {
+			logger.Errorf("Error while shutting down Prometheus exporter:%v", err)
+			return
+		}
+		logger.Info("Prometheus exporter shutdown")
+	}()
+	logger.Info("Prometheus collector exporter started")
+}
+
+func getAppName(c *cfg.Config) string {
```
charith87, line 101: Is there a chance that c.AppName is not populated?
This fallback is better to be there in config.

charith87 (dismissed):
Overall Approving.

internal/monitor/otelexporters.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.opentelemetry.io/contrib/detectors/gcp"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/prometheus"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+)
+
+// SetupOTelMetricExporters sets up the metrics exporters
+func SetupOTelMetricExporters(ctx context.Context, c *cfg.Config) (shutdownFn common.ShutdownFn) {
```
charith87, line 38: Pls check how we can add tests to check if the exporters are created correctly or not. Hopefully the metrics tests will catch these.

kislaykishore: Yeah, we plan to add integration tests for the same as part of/after adding metrics.

internal/monitor/otelexporters.go
kislaykishore: Makes sense - pinned it to gcsfuse.

cfg/validate.go
```diff
@@ -177,6 +177,10 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	if m.StackdriverExportInterval != 0 && m.CloudMetricsExportIntervalSecs != 0 {
 		return fmt.Errorf("exactly one of stackdriver-export-interval and cloud-metrics-export-interval-secs must be specified")
 	}
+	const maxPortNumber = 65535
```
gargnitingoogle, line 180: nit: Does it make sense to declare it as a variable and not literal if it's used right here and only here ?

cfg/validate.go
```diff
@@ -177,6 +177,10 @@ func isValidMetricsConfig(m *MetricsConfig) error {
...
+	if m.PrometheusPort > maxPortNumber {
+		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: 65535 but received: %d instead", m.PrometheusPort)
```
gargnitingoogle, line 182: If we've turned this into a variable, might as well as use it here.

```suggestion
		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: %v but received: %d instead", m.PrometheusPort, maxPortNumber)
```

cfg/validate_test.go
```diff
@@ -539,6 +539,34 @@ func TestValidateMetrics(t *testing.T) {
...
+			name: "prom_disabled_0",
+			metricsConfig: MetricsConfig{
+				PrometheusPort: 0,
+			},
+			wantErr: false,
+		},
+		{
+			name: "prom_disabled_less_than_0",
+			metricsConfig: MetricsConfig{
+				PrometheusPort: -21,
```
gargnitingoogle, line 566: So, negative port numbers are allowed ?

gargnitingoogle (approved):
2 Minor comments and 1 query. LGTM otherwise. PTAL.

cfg/validate.go
kislaykishore: This helps document the meaning of the number. Otherwise it'd look like a magic number.

cfg/validate_test.go
kislaykishore: Yes. It's already allowed in the current OC prometheus exporter. So, didn't want to introduce a breaking change.

cfg/validate.go
kislaykishore: Thanks!

charith87 (approved):

gargnitingoogle (approved):

PR #2695, opened 2024-11-20
Handle error in uploader go routine

ashmeenkaur:
### Description
This PR introduces an error channel `signalUploadFailure  in upload_handler.go which will close in case of failure during upload. Error is propagated to file inode via this channel and in case of failure in the uploader go routine, all follow up writes will result in error.

NA

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA


internal/fs/inode/file.go
```diff
@@ -681,17 +681,28 @@ func (f *FileInode) CreateEmptyTempFile() (err error) {
...
-	err = f.bwh.Write(data, offset)
+// writeToBuffer writes the given content to the in-memory buffer.
+func (f *FileInode) writeToBuffer(data []byte, offset int64) error {
+	if err := f.ensureBufferedWriteHandler(); err != nil {
+		return err
+	}
 
-	return
+	select {
+	case <-f.bwh.SignalUploadFailure():
```
vadlakondaswetha, line 703: can we have this logic inside buffered_write_handler itself. inode shouldn't understand any of buffered_write_handler.go logic. Only invoke and act if error is returned.

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -134,3 +134,7 @@ func (wh *BufferedWriteHandler) WriteFileInfo() WriteFileInfo {
 		Mtime:     wh.mtime,
 	}
 }
+
+func (wh *BufferedWriteHandler) SignalUploadFailure() chan error {
+	return wh.uploadHandler.signalUploadFailure
```
vadlakondaswetha, line 139: Not sure if this is good practice. UploadHandler private variables being exposed by another struct to other structs.

internal/bufferedwrites/upload_handler.go
```diff
@@ -96,9 +103,16 @@ func (uh *UploadHandler) uploader() {
...
-			logger.Errorf("upload failed: error in io.Copy: %v", err)
-			uh.wg.Done()
-			// TODO: handle failure scenario: finalize the upload and trigger edit flow.
+			logger.Errorf("buffered write upload failed: error in io.Copy: %v", err)
+			// close the channel to signal upload failure
+			close(uh.signalUploadFailure)
+
+			// Close the writer to finalize the object creation on GCS.
+			if closeErr := uh.writer.Close(); closeErr != nil {
+				logger.Errorf("Error in finalizing object: %v", closeErr)
```
vadlakondaswetha, line 112: should we print the object name in all the errors? or does it come in the server error response itself?

internal/bufferedwrites/upload_handler.go
```diff
@@ -123,6 +138,7 @@ func (uh *UploadHandler) Finalize() error {
 	err := uh.writer.Close()
 	if err != nil {
 		logger.Errorf("UploadHandler.Finalize(): %v", err)
+		close(uh.signalUploadFailure)
```
vadlakondaswetha, line 141: why do you need this in finalize? can't finalize return error directly?

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -137,12 +141,80 @@ func (t *UploadHandlerTest) TestFinalizeWithNoWriter_CreateObjectWriterFails() {
...
+	writer := &storagemock.Writer{}
 	t.mockBucket.On("CreateObjectChunkWriter", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(writer, nil)
 	assert.Nil(t.T(), t.uh.writer)
+	writer.On("Close").Return(fmt.Errorf("taco"))
 
 	err := t.uh.Finalize()
 
 	assert.Error(t.T(), err)
 	assert.ErrorContains(t.T(), err, "writer.Close")
+	select {
```
vadlakondaswetha, line 153: this can be extracted into seperate method.

internal/bufferedwrites/upload_handler.go
ashmeenkaur: We are returning the error as well (next line). Added this to ensure that any follow up writes on the same inode fail after this error.

internal/bufferedwrites/upload_handler.go
ashmeenkaur: added the object name

internal/bufferedwrites/buffered_write_handler.go
ashmeenkaur: moved to upload handler

internal/fs/inode/file.go
ashmeenkaur: My bad, this was earlier required when I was falling back to temp file flow from file inode in case of failure but no longer a requirement. I forgot to refactor this. Done now.

internal/bufferedwrites/buffered_write_handler_test.go
```diff
@@ -118,3 +118,18 @@ func (testSuite *BufferedWriteTest) TestFlushWithNilCurrentBlock() {
 
 	assert.NoError(testSuite.T(), err)
 }
+
+func (testSuite *BufferedWriteTest) TestWriteToLocalFile_StreamingWritesEnabled_SignalUploadFailure() {
```
vadlakondaswetha, line 122: i think you copied the test from file_test.go. can you rename it

internal/bufferedwrites/upload_handler.go
vadlakondaswetha: In what scenario do we get follow up writes. As soon as the finalize is triggered (success or failure) - bwh should be set to nil.

internal/bufferedwrites/upload_handler.go
```diff
@@ -96,9 +103,16 @@ func (uh *UploadHandler) uploader() {
...
+			close(uh.signalUploadFailure)
+
+			// Close the writer to finalize the object creation on GCS.
+			if closeErr := uh.writer.Close(); closeErr != nil {
+				logger.Errorf("Error in finalizing object %s: %v", uh.objectName, closeErr)
+				return
+			}
+			return
 		}
 		uh.wg.Done()
```
vadlakondaswetha, line 117: you are not marking it as done for failure case. what happens if finalize is triggered after failure?

internal/bufferedwrites/upload_handler.go
ashmeenkaur: Removed `close(uh.signalUploadFailure)` as discussed offline that we will not solve for multithreaded writes for now.
I will take care of making bwh nil in the follow up PR where I integrate flush with file inode sync changes.

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -71,6 +71,14 @@ func (wh *BufferedWriteHandler) Write(data []byte, offset int64) (err error) {
 		return fmt.Errorf("non sequential writes")
 	}
 
+	// Fail early if the uploadHandler has failed.
+	select {
```
vadlakondaswetha, line 75: please extract this into method as it is used at 2 places

vadlakondaswetha (approved):

internal/bufferedwrites/buffered_write_handler.go
ashmeenkaur: I'll take care of this in the sync/flush changes PR.

PR #2696, opened 2024-11-20
refactoring changes to make testing possible for inode package using testify mock bucket

ashmeenkaur:
### Description
This PR includes following refactoring changes:
-  testify mock bucket now implements SyncerBucketInterface
- file inode now uses SyncerBucketInterface instead of struct SyncerBucket

NA

### Testing details
1. Manual - NA
2. Unit tests - Updated
3. Integration tests - NA

### DIFF
--- a/internal/storage/mock/testify_mock_bucket.go
+++ b/internal/storage/mock/testify_mock_bucket.go
@@ -12,12 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package storage
+package mock
 
 import (
 	"context"
 	"io"
 
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 	"github.com/stretchr/testify/mock"
 )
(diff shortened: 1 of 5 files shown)


internal/storage/mock/testify_mock_bucket.go
```diff
@@ -124,3 +125,11 @@ func (m *TestifyMockBucket) CreateFolder(ctx context.Context, folderName string)
 	}
 	return nil, args.Error(1)
 }
+
+func (m *TestifyMockBucket) SyncObject(ctx context.Context, fileName string, srcObject *gcs.Object, content gcsx.TempFile) (o *gcs.Object, err error) {
```
vadlakondaswetha, line 129: MockBucket is for mocking bucket operations. Why are we adding custom operations to this file?

vadlakondaswetha, line 129: also fakes are preferred over mocks. Why do you want to have a mock for syncObject when the methods used by internally has fake implementation.

ashmeenkaur: Added this so that mock bucket also implements SyncerBucketInterface. I want to use it for file_test.go file which uses syncer bucket.

ashmeenkaur: In file_test.go we are creating the content (temp file) before hand for the tests. This won't work for buffered writes as I want to ensure that in case of failure temp file is created. That is why I needed mocks. Do you think there is a better way to do this?

ashmeenkaur:
Discussed offline that we don't need to handle failure in io.copy and fall back to temp file flow. So this PR is no longer required.

PR #2698, opened 2024-11-20
Emit metrics to Cloud monitoring

kislaykishore:
### Description
Implement Google Cloud exporter to emit metrics to Cloud monitoring.

b/380002762

### Testing details
1. Manual - Added an OTel metric and verified that it appeared on metrics explorer
2. Unit tests - NA
3. Integration tests - NA - will be added once metrics have been migrated to OTel.

### DIFF
--- a/internal/monitor/otelexporters.go
+++ b/internal/monitor/otelexporters.go
@@ -18,16 +18,20 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
+	cloudmetric "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric"
 	"github.com/googlecloudplatform/gcsfuse/v2/cfg"
 	"github.com/googlecloudplatform/gcsfuse/v2/common"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.opentelemetry.io/contrib/detectors/gcp"
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/prometheus"
 	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 	"go.opentelemetry.io/otel/sdk/resource"
 	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
 )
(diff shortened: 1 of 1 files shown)


internal/monitor/otelexporters.go
```diff
@@ -36,20 +40,57 @@ const serviceName = "gcsfuse"
...
+		logger.Errorf("Error while creating Google Cloud exporter:%v", err)
+		return nil, nil
+	}
+
+	r := metric.NewPeriodicReader(exporter, metric.WithInterval(time.Duration(secs)*time.Second))
+	return []metric.Option{metric.WithReader(r)}, r.Shutdown
+}
+
+func metricFormatter(m metricdata.Metrics) string {
+	return "custom.googleapis.com/gcsfuse/" + strings.ReplaceAll(m.Name, ".", "/")
```
charith87, line 92: Pls move the constants to top.

charith87, line 92: What is the replaceAll needed. If its our own metrics, could we make sure there is no "." in them

internal/monitor/otelexporters.go
```diff
@@ -36,20 +40,57 @@ const serviceName = "gcsfuse"
...
 }
 
+func setupCloudMonitoring(secs int64) ([]metric.Option, common.ShutdownFn) {
+	if secs <= 0 {
+		return nil, nil
+	}
+	options := []cloudmetric.Option{
+		cloudmetric.WithMetricDescriptorTypeFormatter(metricFormatter),
+		cloudmetric.WithFilteredResourceAttributes(func(kv attribute.KeyValue) bool {
+			// Ensure that PID is available as a metric label on metrics explorer.
```
charith87, line 76: Any more info regarding PID?

internal/monitor/otelexporters.go
kislaykishore: https://engdoc.corp.google.com/eng/doc/devguide/java/practices/constants.md?cl=head#when-to-use-a-constant-versus-an-inlined-literal - I think in this case an inlined literal makes more sense as opposed to a constant.

internal/monitor/otelexporters.go
kislaykishore: Can you elaborate on what information you are looking for?

charith87: Was trying to understand what is semconv.ProcessPIDKey and why we are setting this. Is this to differentiate different PIDs ( different mounts ) on the same node or is it for something else.
wanted to add metrics for the things which we can map to debugging and checking if we can reduce cardinality wherever possible.

kislaykishore: Yes, it's for differentiating between mounts.

internal/monitor/otelexporters.go
kislaykishore: It's just to be a bit more consistent. OTel metrics use the dot as the separator. I didn't want to create two separate conventions for GCSFuse-related metrics.

internal/monitor/otelexporters.go
charith87: I feel this will break the cardinality. PIDs can be a long number ( maybe 5 digits). And filtering support for them needs means they have to be bucketized and not sure if metrics pipeline can take it. Generally only a fixed set of values is what can be filtered. Here we already have bucket name, pod name etc . Can we recheck this if metric pipeline can take this or not.

kislaykishore: This is a different exporter i.e. Cloud monitoring (aka stackdriver) - unrelated to Prometheus. FWIW, the corresponding exporter in OpenCensus already exported PID.

internal/monitor/otelexporters.go
charith87: Okk

charith87: Seems fine.

charith87 (approved):

internal/monitor/otelexporters.go
charith87: We could take a separate task to audit the number of dimensions and cardinality for all metrics in the end of the feature. For now resolving this.

charith87 (approved):

internal/monitor/otelexporters.go
kislaykishore: Done

PR #2706, opened 2024-11-25
Set clobbered to true if an object is present on GCS and is local.

vipnydav:
### Description
Setting clobbered to true if an object is present on GCS and is local.

NA

### Testing details
1. Manual - Done
2. Unit tests - NA

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -108,17 +108,17 @@ var _ Inode = &FileInode{}
 // REQUIRES: len(m.Name) > 0
 // REQUIRES: m.Name[len(m.Name)-1] != '/'
 func NewFileInode(
-	id fuseops.InodeID,
-	name Name,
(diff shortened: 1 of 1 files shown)


internal/fs/inode/file.go
```diff
@@ -212,6 +212,12 @@ func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool, inclu
...
 
+	// For localFile, if object exists in GCS then it means file is clobbered.
+	if f.IsLocal() {
+		b = true
+		return
+	}
+
 	// We are clobbered iff the generation doesn't match our source generation.
 	oGen := Generation{o.Generation, o.MetaGeneration}
 	b = f.SourceGeneration().Compare(oGen) != 0
```
ashmeenkaur, line 223: Can you check if this is already taking care of the above condition we have added?

PR #2691, opened 2024-11-18
Add warning log for direct path connectivity

anushka567:
### Description
 Using  the [storage.CheckDirectConnectivitySupported](https://pkg.go.dev/cloud.google.com/go/storage@v1.47.0#CheckDirectConnectivitySupported) function to check for direct connectivity support, this PR
adds a warning log message if direct connectivity is unavailable, including the bucket name and the error message.
This logging is only for GRPC data client.

Sample log:
```
Direct path connectivity unavailable for <bucket-name>, reason: storage: direct connectivity not detected
```

Time observed for figuring out direct connectivity :
For vm located in us-west1-b: 10159 msec(issue actively tracked in internal bug)
For vm located in us-central1-f : 365 msec
### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

PR #2715, opened 2024-11-26
adding unit test for direct path detector functionality

anushka567:
### Description
Follow-up PR for https://github.com/GoogleCloudPlatform/gcsfuse/pull/2691

Added unit test for the following scenario:

- When client protocol is set to grpc, then directPathDetector is initialized and should be used for checking direct connectivity.
- When client protocol is set to http1, then directPathDetector is nil since we do not require checking direct connectivity. 

NA

### Testing details
1. Manual - Done on cloudtop.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/storage_handle_test.go
+++ b/internal/storage/storage_handle_test.go
@@ -203,6 +203,32 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleWithInvalidClientProtoco
 	assert.Contains(testSuite.T(), err.Error(), "invalid client-protocol requested: test-protocol")
 }
 
+func (testSuite *StorageHandleTest) TestNewStorageHandleWithNonNilDirectPathDetector() {
+	sc := storageutil.GetDefaultStorageClientConfig()
+	sc.ExperimentalEnableJsonRead = true
+	sc.ClientProtocol = "grpc"
+
+	handleCreated, err := NewStorageHandle(testSuite.ctx, sc)
+	storageClient, _ := handleCreated.(*storageClient)
+
+	assert.Nil(testSuite.T(), err)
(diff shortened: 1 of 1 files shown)


internal/storage/storage_handle_test.go
```diff
@@ -203,6 +203,32 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleWithInvalidClientProtoco
...
+
+	assert.Nil(testSuite.T(), err)
+	assert.NotNil(testSuite.T(), handleCreated)
+	assert.NotNil(testSuite.T(), storageClient.directPathDetector)
+}
+
+func (testSuite *StorageHandleTest) TestNewStorageHandleWithNilDirectPathDetector() {
+	sc := storageutil.GetDefaultStorageClientConfig()
+	sc.ExperimentalEnableJsonRead = true
+	sc.ClientProtocol = "http1"
```
raj-prince, line 222: Can we add for http2 also?

internal/storage/storage_handle_test.go
```diff
@@ -203,6 +203,32 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleWithInvalidClientProtoco
...
+
+	handleCreated, err := NewStorageHandle(testSuite.ctx, sc)
+	storageClient, _ := handleCreated.(*storageClient)
```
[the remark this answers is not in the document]
anushka567: Updated.

internal/storage/storage_handle_test.go
anushka567: Added  http2 case.

raj-prince (approved):

PR #2719, opened 2024-11-28
Remove error from ops_error_count metric labels.

kislaykishore:
ops_error_count cannot be grouped by actual error anymore. It was earlier removed to reduce cardinality.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


raj-prince (approved):
LGTM.

Would be good to update the description with earlier cardinality changes PR link.

PR #2723, opened 2024-11-29
Fix Version test

kislaykishore:
The expected version differs depending upon whether the installed package is being used. So, condition the test accordingly.

b/381357743

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/mounting/gcsfuse_test.go
+++ b/tools/integration_tests/mounting/gcsfuse_test.go
@@ -27,6 +27,8 @@ import (
 	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/canned"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/setup"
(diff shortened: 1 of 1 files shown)


tools/integration_tests/mounting/gcsfuse_test.go
```diff
@@ -230,10 +232,19 @@ func (t *GcsfuseTest) Version() {
...
+			assertContains("fake_version", string(output))
+		}
 	}
 }
 
+func assertContains(expected, actual string) {
+	if !strings.Contains(actual, expected) {
+		logger.Fatal("Actual: %s does not contain expected: %s", actual, expected)
+	}
+}
```
Tulsishah, line 247: nit: leave empty line after this.
Also did you check once the failure scenario is executing properly?

Tulsishah (approved):

tools/integration_tests/mounting/gcsfuse_test.go
kislaykishore: Yes, verified that test failure is captured.

Tulsishah (approved):

PR #2725, opened 2024-11-29
add benchmarking test package to e2e testing 

anushka567:
### Description
The benchmarking test package performs delete and stat benchmark, the test will fail if the time taken for the operation is beyond the threshold.Also, added this to e2e test scripts(both kokoro and louhi pipeline).
NA

### Testing details
1. Manual - on cloudtop and GCE VM
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -203,6 +203,9 @@ function run_parallel_tests() {
 
   for test_dir_p in "${test_array[@]}"
   do
+    # Unlike regular tests,benchmark tests are not executed by default when using go test .
+    # The -bench flag yells go test to run the benchmark tests and report their results by
+    # enabling the benchmarking framework.
     if [ $test_dir_p == "benchmarking" ]; then
         benchmark_flags="-bench=."
     fi
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -157,6 +157,7 @@ TEST_DIR_PARALLEL=(
   "log_content"
   "kernel_list_cache"
   "concurrent_operations"
+  "benchmarking"
 )
 
 # These tests never become parallel as they are changing bucket permissions.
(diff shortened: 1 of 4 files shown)


tools/integration_tests/benchmarking/benchmark_delete_test.go
```diff
@@ -25,6 +25,10 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/setup"
 )
 
+const (
+	deleteThreshold int = 450
```
raj-prince, line 29: Can't we change it to time.Duration?

tools/integration_tests/benchmarking/benchmark_stat_test.go
```diff
@@ -53,6 +57,9 @@ func (s *benchmarkStatTest) Benchmark_Stat(b *testing.B) {
 			b.Errorf("testing error: %v", err)
 		}
 	}
+	if timeTaken := b.Elapsed().Milliseconds(); timeTaken > int64(statThreshold*b.N) {
+		b.Errorf("Test failed due to timeout, time taken:%d msec", timeTaken)
```
raj-prince, line 61: Let's print the expected time too, in the error log.

Instead of saying `Test failed due to timeout` -> `StatFile took more time (timeTaken) than expected (expectedTime).`  or something similar.

tools/cd_scripts/e2e_test.sh
```diff
@@ -198,15 +199,20 @@ function run_parallel_tests() {
   local -n test_array=$1
   local BUCKET_NAME=$2
   local pids=()
+  local benchmark_flags=""
+
   for test_dir_p in "${test_array[@]}"
   do
+    if [ $test_dir_p == "benchmarking" ]; then
+        benchmark_flags="-bench=."
```
Tulsishah, line 207: Please write comment above why this variable is require.

anushka567: added.

Tulsishah (approved):

raj-prince (approved):

PR #2727, opened 2024-11-29
Modify fake bucket implementation to throw accurate error.

vipnydav:
### Description
Modify fake bucket implementation to throw same error that is thrown by GCS Bucket.

NA

### Testing details
1. Manual - Done
2. Unit tests - NA

### DIFF
--- a/internal/storage/fake/bucket.go
+++ b/internal/storage/fake/bucket.go
@@ -442,7 +442,7 @@ func (b *bucket) newReaderLocked(
 	// Find the object with the requested name.
 	index = b.objects.find(req.Name)
 	if index == len(b.objects) {
-		err = storage.ErrObjectNotExist
+		err = fmt.Errorf("object %s not found: %w", req.Name, storage.ErrObjectNotExist)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


internal/storage/fake/bucket.go
```diff
@@ -442,10 +442,7 @@ func (b *bucket) newReaderLocked(
 	// Find the object with the requested name.
 	index = b.objects.find(req.Name)
 	if index == len(b.objects) {
-		err = &gcs.NotFoundError{
-			Err: fmt.Errorf("object %s not found", req.Name),
-		}
-
+		err = storage.ErrObjectNotExist
```
gargnitingoogle, line 445: This loses information as this error can no longer convey the name of the missing object.

gargnitingoogle: Why do we need this change and isn't there any alternative ?

vipnydav: Updated the error to wrap the storage.ErrObjectNotExist with additional context.

vipnydav: We need to change this because we recieve storage.ErrObjectNotExist while creating a new reader for real bucket and fake bucket should also return same error.

vipnydav: Closing this PR as it is better to map storage.ErrObjectNotExist to gcs.NotFoundError in bucket_handle.go

PR #2732, opened 2024-12-02
Made changes to support setting mtime and getting when buffered writes are enabled

vadlakondaswetha:
### Description
Updated SetInodeattributes and getAttributes for handling mtime for local files and empty GCS files.
invoking bufferedWrites flow for empty gcs files.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1685,13 +1685,10 @@ func (fs *fileSystem) createLocalFile(
 	}
 	child = fs.mintInode(core)
 	fs.localFileInodes[child.Name()] = child
-	// For buffered writes, we don't create temp files on disk.
-	if !fs.newConfig.Write.ExperimentalEnableStreamingWrites {
-		// Empty file is created to be able to set attributes on the file.
-		fileInode := child.(*inode.FileInode)
-		if err := fileInode.CreateEmptyTempFile(); err != nil {
-			return nil, err
-		}
+	// Empty file is created to be able to set attributes on the file.
+	fileInode := child.(*inode.FileInode)
+	if err := fileInode.CreateEmptyTempFile(); err != nil {
+		return nil, err
 	}
 	fs.mu.Unlock()
 
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1687,7 +1687,7 @@ func (fs *fileSystem) createLocalFile(
 	fs.localFileInodes[child.Name()] = child
 	// Empty file is created to be able to set attributes on the file.
 	fileInode := child.(*inode.FileInode)
-	if err := fileInode.CreateEmptyTempFile(); err != nil {
+	if err := fileInode.CreateBufferedOrTempWriter(); err != nil {
 		return nil, err
 	}
 	fs.mu.Unlock()
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -706,14 +706,6 @@ func (f *FileInode) CreateEmptyTempFile() (err error) {
 	return
 }
 
-// writeToBuffer writes the given content to the in-memory buffer.
-func (f *FileInode) writeToBuffer(data []byte, offset int64) (err error) {
-
-	err = f.bwh.Write(data, offset)
-
-	return
-}
-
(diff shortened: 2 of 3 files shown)


internal/fs/inode/file.go
```diff
@@ -414,6 +414,14 @@ func (f *FileInode) Attributes(
 		}
 	}
 
+	bwEnabled, err := f.shouldTriggerBufferedWrites()
+	if err != nil {
+		return
+	}
+	if bwEnabled {
+		attrs.Mtime = f.bwh.WriteFileInfo().Mtime
```
ashmeenkaur, line 422: Should we take size also from bwh?

internal/fs/inode/file.go
```diff
@@ -681,17 +706,28 @@ func (f *FileInode) CreateEmptyTempFile() (err error) {
...
 		}
+		f.bwh.SetMtime(f.mtimeClock.Now())
 	}
 
-	err = f.bwh.Write(data, offset)
+	return nil
+}
 
-	return
+func (f *FileInode) shouldTriggerBufferedWrites() (bool, error) {
```
ashmeenkaur, line 722: Nit
```suggestion
func (f *FileInode) bufferedWritesEnabled() (bool, error) {
```

internal/fs/fs.go
```diff
@@ -1685,13 +1685,10 @@ func (fs *fileSystem) createLocalFile(
...
-	// For buffered writes, we don't create temp files on disk.
-	if !fs.newConfig.Write.ExperimentalEnableStreamingWrites {
-		// Empty file is created to be able to set attributes on the file.
-		fileInode := child.(*inode.FileInode)
-		if err := fileInode.CreateEmptyTempFile(); err != nil {
-			return nil, err
-		}
+	// Empty file is created to be able to set attributes on the file.
+	fileInode := child.(*inode.FileInode)
+	if err := fileInode.CreateEmptyTempFile(); err != nil {
```
ashmeenkaur, line 1690: This looks like a tech debt.. is there no way we can set attributes in bwh?

I think this also allows truncating the file in SetInodeAttributes. We don't want to allow that, right? Shall we throw an error / log a warning explicitly?

internal/fs/inode/file.go
vadlakondaswetha: Thanks. Since i focussed only on mtime, missed the size. updated now

internal/fs/inode/file.go
```diff
@@ -673,6 +696,15 @@ func (f *FileInode) CacheEnsureContent(ctx context.Context) (err error) {
 }
 
 func (f *FileInode) CreateEmptyTempFile() (err error) {
```
ashmeenkaur, line 698: Shall we rename this function to `CreateWriteHandler()` or `CreateBufferedOrTempWriter()`

internal/fs/inode/file_test.go
```diff
@@ -1004,6 +1066,59 @@ func (t *FileTest) MultipleWritesToLocalFileWhenStreamingWritesAreEnabled() {
 	err = t.in.Write(t.ctx, []byte("hello"), 0)
 	AssertEq(nil, err)
 	AssertEq(7, t.in.bwh.WriteFileInfo().TotalSize)
+	// The inode should agree about the new mtime.
```
ashmeenkaur, line 1069: shall we check for mtime too? same for WriteToEmptyGCSFileWhenStreamingWritesAreEnabled

internal/fs/inode/file.go
```diff
@@ -673,6 +696,15 @@ func (f *FileInode) CacheEnsureContent(ctx context.Context) (err error) {
...
+	// Skip creating empty file when streaming writes are enabled
+	if f.local && f.writeConfig.ExperimentalEnableStreamingWrites {
```
ashmeenkaur, line 700: Should this be for size 0 too?

vadlakondaswetha: This method is invoked only in CreateFile flow and not open file.
Size 0 files come in open file flow.

ashmeenkaur (approved):

PR #2736, opened 2024-12-03
Reduce test-parallelism to 1

kislaykishore:
* Currently, tests get timed-out probably due to resource constraints on the machine. Reducing the parallelism to 1 so that the resource requirements reduces.
* Keeping the cache tests and 

b/380798155

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,10 +43,8 @@ jobs:
         CGO_ENABLED=0 go build ./...
         go install ./tools/build_gcsfuse
         build_gcsfuse . /tmp ${GITHUB_SHA}
-    - name: Test all except caching parallely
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         go install ./tools/build_gcsfuse
         build_gcsfuse . /tmp ${GITHUB_SHA}
     - name: Test all except caching parallely
```
ashmeenkaur, line 46: Please change the stage name

kislaykishore: That was quick :)

ashmeenkaur (approved):

PR #2749, opened 2024-12-06
Add file-cache metrics to the OC views

kislaykishore:
This fixes a bug due to which the file-cache metrics were not getting exported.

NA

### Testing details
1. Manual - Verified that the all the metrics are showing up now in the Prometheus export
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/common/oc_metrics.go
+++ b/common/oc_metrics.go
@@ -212,6 +212,28 @@ func NewOCMetrics() (MetricHandle, error) {
 			Description: "The cumulative distribution of file system operation latencies",
 			Aggregation: ochttp.DefaultLatencyDistribution,
 			TagKeys:     []tag.Key{tag.MustNewKey(FSOp)},
+		},
+		// File cache related metrics
+		&view.View{
+			Name:        "file_cache/read_count",
+			Measure:     fileCacheReadCount,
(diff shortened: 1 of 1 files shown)


common/oc_metrics.go
```diff
@@ -212,6 +212,28 @@ func NewOCMetrics() (MetricHandle, error) {
...
+			Description: "The cumulative number of bytes read from file cache along with read type - Sequential/Random",
+			Aggregation: view.Sum(),
+			TagKeys:     []tag.Key{tag.MustNewKey(ReadType)},
+		},
+		&view.View{
+			Name:        "file_cache/read_latencies",
+			Measure:     fileCacheReadLatency,
+			Description: "The cumulative distribution of the file cache read latencies along with cache hit - true/false",
+			Aggregation: ochttp.DefaultLatencyDistribution,
+			TagKeys:     []tag.Key{tag.MustNewKey(CacheHit)},
```
ankitaluthra1, line 236: should we also have unit test to see if View is created correctly ? please add in subsequent PR if possible, approving this so bug fix is merged

ankitaluthra1 (approved):

common/oc_metrics.go
kislaykishore: Yes, we intend to add integration tests as part of OTel implementation that'll verify this.

PR #2755, opened 2024-12-09
[Write-stall] Create chunk transfer timeout flag

Tulsishah:
### Description
- Adding hidden flag to change the value of chunkTransferTimeout in gcs-retries.
- Default value set to 10s

NA

### Testing details
1. Manual - Manually verified that hidden flag is not visible.
2. Unit tests - NA
3. Integration tests - NA


cfg/params.yaml
```diff
@@ -628,6 +628,17 @@
   default: 64 #TODO: revisit default value after perf testing.
   hide-flag: true
 
+- config-path: "write.chunk-transfer-timeout"
+  flag-name: "chunk-transfer-timeout"
```
kislaykishore, line 632: We are suffixing the unit for other flags e.g. "kernel-list-cache-ttl-secs" and create an int flag instead of duration. Maybe we should follow the same convention here.

Tulsishah: Done

kislaykishore (approved):

cfg/validate.go
```diff
@@ -185,6 +185,13 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	return nil
 }
 
+func isValidChunkTransferTimeoutForRetriesConfig(chunkTransferTimeoutSecs *int64) error {
```
kislaykishore, line 188: Don't use pointer types for int64.

Tulsishah: Done

cfg/validate.go
```diff
@@ -185,6 +185,13 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	return nil
 }
 
+func isValidChunkTransferTimeoutForRetriesConfig(chunkTransferTimeoutSecs int64) error {
+	if chunkTransferTimeoutSecs < 0 || chunkTransferTimeoutSecs > maxSupportedTTLInSeconds {
```
kislaykishore, line 189: Why do we need the maxSupportedTTLInSeconds check?
```suggestion
	if chunkTransferTimeoutSecs < 0 {
```

kislaykishore (commented):
Apologies for the back and forth.

cfg/validate.go
Tulsishah: maxSupportedTTLInSeconds represents maximum multiple of seconds representable by time.Duration.

To check upper limit.

kislaykishore (approved):

PR #2764, opened 2024-12-09
Prevent unit test failure due to flag parse

Tulsishah:
### Description
Previously, unit tests were failing when running alongside end-to-end (e2e) tests. The error message indicated an undefined flag: -integrationTest. This issue is resolved by adding the command that parse tests within the TestMain function and return if integrationTest flag passed.

**Note**: We can't use [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L264) as it accepts *testing.T but in TestMain we have *testing.M. If we want to use this function then we have to add this in each and every unit test.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


raj-prince (approved):
LGTM. Assuming there will not be an issue in running only unit tests.

PR #2767, opened 2024-12-10
[Write-stall] Add e2e tests - 1

Tulsishah:
### Description
- E2e tests for with no chunk-transfer-timeout
- This test verifies that write operations stall for the expected duration when a write stall is induced while uploading the first chunk. It creates a file, writes data to it, and then calls Sync() to ensure the data is written to GCS. The test measures the time taken for the Sync() operation and asserts that it is greater than or equal to the configured stall time.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA


tools/integration_tests/emulator_tests/write_stalls/chunk_transfer_timeout_infinity_test.go
```diff
@@ -0,0 +1,116 @@
...
+	assert.GreaterOrEqual(t, elapsedTime, stallTime)
+}
+
+////////////////////////////////////////////////////////////////////////
+// Test Function (Runs once before all tests)
+////////////////////////////////////////////////////////////////////////
+
+func TestChunkTransferTimeoutInfinity(t *testing.T) {
+	ts := &chunkTransferTimeoutInfinity{}
+	proxyEndpoint := fmt.Sprintf("http://localhost:%d/storage/v1/b?project=test-project/b?bucket=%s", port, setup.TestBucket())
```
raj-prince, line 104: Why are we providing bucket-name as proxyEndpoint? Shouldn't it take as it is via GCSFuse mount option?

tools/integration_tests/emulator_tests/write_stalls/write_stall_test.go
```diff
@@ -0,0 +1,52 @@
...
+const port = 8020
+
+var (
+	testDirPath string
+	mountFunc   func([]string) error
+	// root directory is the directory to be unmounted.
+	rootDir string
+)
+
+func TestMain(m *testing.M) {
```
raj-prince, line 35: Can we make TestMain() common for all the emulator based test?

And then a file like `write_stall_test.go` for different test suites. So, if we need to add another test tomorrow then we will just implement `read_stall_test.go`. That's it.

If it's possible then we can remove this intermediate folder `write_stalls`.

tools/integration_tests/emulator_tests/write_stalls/chunk_transfer_timeout_infinity_test.go
Tulsishah: Hmm makes sense updated. Although in my testing I used same type of URL and it was working fine.

raj-prince (approved):

PR #2769, opened 2024-12-10
Fixing read for bufferedwrites and handling it for empty GCS File

vadlakondaswetha:
### Description
For an emptyGCS file after first writes, reads are not allowed since buffered writes come into picture. Handling it and fixing the bug in earlier PR/

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/file_test.go
+++ b/internal/fs/inode/file_test.go
@@ -1028,20 +1028,73 @@ func (t *FileTest) TestUnlinkLocalFile() {
 	assert.Equal(t.T(), "gcs.NotFoundError: object test not found", err.Error())
 }
 
-func (t *FileTest) TestReadLocalFileWhenStreamingWritesAreEnabled() {
-	// Create a local file inode.
-	t.createInodeWithLocalParam("test", true)
-	t.in.writeConfig = getWriteConfig()
-	err := t.in.Write(t.ctx, []byte("hi"), 0)
-	assert.Nil(t.T(), err)
-	assert.Equal(t.T(), int64(2), t.in.bwh.WriteFileInfo().TotalSize)
+func (t *FileTest) TestReadFileWhenStreamingWritesAreEnabled() {
+	tbl := []struct {
+		name         string
+		localFile    bool
+		emptyFile    bool
+		performWrite bool
+	}{
+		{
+			name:         "LocalFileWithWrite",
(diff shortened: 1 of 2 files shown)


internal/fs/inode/file_test.go
```diff
@@ -1028,20 +1028,73 @@ func (t *FileTest) TestUnlinkLocalFile() {
...
 }
 
-func (t *FileTest) TestReadLocalFileWhenStreamingWritesAreEnabled() {
-	// Create a local file inode.
-	t.createInodeWithLocalParam("test", true)
-	t.in.writeConfig = getWriteConfig()
-	err := t.in.Write(t.ctx, []byte("hi"), 0)
-	assert.Nil(t.T(), err)
-	assert.Equal(t.T(), int64(2), t.in.bwh.WriteFileInfo().TotalSize)
+func (t *FileTest) TestReadFileWhenStreamingWritesAreEnabled() {
```
ashmeenkaur, line 1031:
```suggestion
func (t *FileTest) TestReadFileThrowsErrorWhenStreamingWritesAreInProgress() {
```

internal/fs/inode/file_test.go
```diff
@@ -1028,20 +1028,73 @@ func (t *FileTest) TestUnlinkLocalFile() {
...
+			n, err := t.in.Read(t.ctx, data, 0)
+
+			assert.Equal(t.T(), 0, n)
+			assert.NotNil(t.T(), err)
+			assert.Equal(t.T(), "cannot read a file when upload in progress", err.Error())
+		})
+	}
+}
+
+func (t *FileTest) TestReadEmptyGCSFileWhenStreamingWritesAreEnabled() {
```
ashmeenkaur, line 1090:
```suggestion
func (t *FileTest) TestReadEmptyGCSFileSucceedsWhenStreamingWritesAreNotInProgress() {
```
Also:
t.in.writeConfig = getWriteConfig()

internal/fs/inode/file_test.go
vadlakondaswetha: For local file when streaming writes are enabled, we dont support reads irrespective of whether write happened or not. Thats why not calling as in progress

ashmeenkaur (approved):

PR #2770, opened 2024-12-10
Update Flush call to return an object

ashmeenkaur:
### Description
This PR changes flush call to return an object.
Also ClearFreeBlockChannel is included in Flush call to clear up all the buffers.

NA

### Testing details
1. Manual - NA
2. Unit tests - Updated
3. Integration tests - NA

### DIFF
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -38,7 +38,7 @@ type UploadHandler struct {
 	freeBlocksCh chan block.Block
 
 	// writer to resumable upload the blocks to GCS.
-	writer io.WriteCloser
+	writer gcs.Writer
 
 	// signalUploadFailure channel will propagate the upload error to file
 	// inode. This signals permanent failure in the buffered write job.
@@ -116,7 +116,7 @@ func (uh *UploadHandler) uploader() {
 }
 
 // Finalize finalizes the upload.
-func (uh *UploadHandler) Finalize() error {
+func (uh *UploadHandler) Finalize() (*gcs.Object, error) {
 	uh.wg.Wait()
 	close(uh.uploadCh)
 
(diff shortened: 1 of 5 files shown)


internal/bufferedwrites/upload_handler.go
```diff
@@ -125,16 +125,16 @@ func (uh *UploadHandler) Finalize() error {
...
 		if err != nil {
-			return fmt.Errorf("createObjectWriter failed for object %s: %w", uh.objectName, err)
+			return nil, fmt.Errorf("createObjectWriter failed for object %s: %w", uh.objectName, err)
 		}
 	}
 
-	err := uh.writer.Close()
+	obj, err := uh.bucket.FinalizeUpload(context.Background(), uh.writer)
 	if err != nil {
 		logger.Errorf("UploadHandler.Finalize(%s): %v", uh.objectName, err)
```
vadlakondaswetha, line 134: if we dont log and return wont it get logged somewhere? its not mandatory to log when returning the error.

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -111,46 +115,56 @@ func (t *UploadHandlerTest) TestUpload_CreateObjectWriterFails() {
...
+	obj, err := t.uh.Finalize()
 
 	require.Error(t.T(), err)
 	assert.ErrorContains(t.T(), err, "taco")
 	assert.ErrorContains(t.T(), err, "createObjectWriter")
+	assert.Nil(t.T(), obj)
 }
 
-func (t *UploadHandlerTest) TestFinalize_WriterCloseFails() {
+func (t *UploadHandlerTest) TestFinalize_FinalizeUploadFails() {
```
vadlakondaswetha, line 155: some tests we are having _ and others not. Lets be consistent and not use _ anywhere.

vadlakondaswetha (approved):

internal/bufferedwrites/upload_handler.go
ashmeenkaur: removed, thanks. The other failure log in this file is for uploader go routine (async) so not removing it.

vadlakondaswetha (approved):

PR #2772, opened 2024-12-10
Enforce file cache availability for parallel downloads through config…

codechanges:
… validation

### Description
Parallel download support requires file cache to be configured. From [documentation](https://cloud.google.com/storage/docs/cloud-storage-fuse/file-caching#parallel-downloads), `To use parallel downloads, you must first enable and configure the file cache.`

This fix will add a config validation to make parallel downloads support dependent on file cache config.

NA

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA


cfg/validate.go
```diff
@@ -232,5 +241,9 @@ func ValidateConfig(v isSet, config *Config) error {
 		return fmt.Errorf("error parsing metrics config: %w", err)
 	}
 
+	if err = isValidParallelDownloadConfig(config); err != nil {
```
raj-prince, line 244: Shouldn't we check emptiness of cacheDir outside and pass only FileCacheConfig to validate parallelDownload parameters?

cfg/validate.go
```diff
@@ -43,29 +44,37 @@ func isValidURL(u string) error {
...
+		if config.FileCache.MaxParallelDownloads == 0 {
+			return errors.New("the value of max-parallel-downloads for file-cache must not be 0 when enable-parallel-downloads is true")
+		}
+		if config.FileCache.WriteBufferSize < CacheUtilMinimumAlignSizeForWriting {
+			return errors.New("the value of write-buffer-size for file-cache can't be less than 4096")
+		}
+		if (config.FileCache.WriteBufferSize % CacheUtilMinimumAlignSizeForWriting) != 0 {
+			return errors.New("the value of write-buffer-size for file-cache should be in multiple of 4096")
+		}
+		if string(config.CacheDir) == "" {
```
raj-prince, line 58: This suggests - there are two conditions to enable file-cache:
https://github.com/GoogleCloudPlatform/gcsfuse/blob/94a9357fce70096f37a334511bb6d3d58f50148c/cfg/config_util.go#L27

And here we are only checking for CacheDir?

cfg/validate.go
codechanges: I did thought about this. But I found that, it is cleaner to pass top level object and encapsulate details inside a function. Since there is one part of this config is already outside of file cache (not ideal!) there is a possibility of something else coming at the top level hence I found this approach better than polluting this function.

cfg/validate.go
codechanges: Thanks. I based this on `Note that you enable file caching by passing a directory to cache-dir field.` which is mention [here](https://cloud.google.com/storage/docs/cloud-storage-fuse/file-caching#configure).

Is this not accurate? Else I would change this.

raj-prince (approved):

PR #2777, opened 2024-12-10
[Emulator tests] Automate tests in kokoro and add helper to run proxy server

Tulsishah:
### Description
- created helper function which will start proxy server.
- Also created function to kill the process
- Automated tests in kokoro
- Proxy server logs will get throw on kokoro log bucket.

NA

### Testing details
1. Manual - Tested
2. Unit tests - NA
3. Integration tests - NA


tools/integration_tests/emulator_tests/util/test_helper.go
```diff
@@ -0,0 +1,124 @@
...
+//
+// The function also starts a simple HTTP server to keep the process running and provides an
+// endpoint to check the proxy status.
+//
+// When receiving SIGINT or SIGTERM signals, it gracefully shuts down the proxy server by:
+//   - Sending SIGINT to the proxy process.
+//   - Killing any processes listening on given port.
+func StartProxyServer(port int, configPath string) {
+	// Start the proxy in the background
+	cmd := exec.Command("go", "run", "../proxy_server/.", "--config-path="+configPath)
```
raj-prince, line 45: This requires go in the path right. Do we have other way?

tools/integration_tests/emulator_tests/util/test_helper.go
```diff
@@ -0,0 +1,124 @@
...
+	log.Printf("Proxy server logs are generated with specific filename %s: ", logFileForProxyServer.Name())
+	cmd.Stdout = logFileForProxyServer
+	cmd.Stderr = logFileForProxyServer
+	err = cmd.Start()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Handle SIGINT and SIGTERM signals to gracefully shut down the proxy
+	sigs := make(chan os.Signal, 1)
```
raj-prince, line 59: Why do we need all these? Server already implements SIGINT and SIGTERM - right?

raj-prince (commented):
Looks good overall. One comment about duplicate SIGINT handler.

tools/integration_tests/emulator_tests/util/test_helper.go
Tulsishah: This is with respect to test package directory. So the path will be always same.
e.g.
```
emulator_tests/write_stall
emulator_tests/proxy_server/
```

tools/integration_tests/emulator_tests/util/test_helper.go
Tulsishah: I added the signal handling because the proxy server runs in the background, and I thought it was necessary in case our end-to-end tests were terminated mid-process. However, after further testing, I realized that it's no longer needed.

tools/integration_tests/emulator_tests/util/test_helper.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	log.Printf("Proxy server logs are generated with specific filename %s: ", logFileForProxyServer.Name())
+	cmd.Stdout = logFileForProxyServer
+	cmd.Stderr = logFileForProxyServer
+	err = cmd.Start()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Start a simple HTTP server to keep the process running
```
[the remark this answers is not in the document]
Tulsishah: It was required due to the program continues running, keeping the process alive, but the logic changes from then so it is not required any more. Removed.

raj-prince (approved):

PR #2778, opened 2024-12-10
[Do not review] Sync/flush changes for buffered writes

ashmeenkaur:
### Description
This PR includes changes to 
- integrate buffered write handler with sync/flush file flow.
- Fall back to temp file flow in case of out of order writes.
- Finalize the object in case of error during write.

Some refactoring is done to reuse the code.

**Note:** Flush empty file changes required precondition to be propagated to gcs request and will be taken up in the follow up PR. Unit tests are not added for it.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -721,20 +721,15 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 		err = fmt.Errorf("SyncObject: %w", err)
 		return
 	}
-
+	minObj := storageutil.ConvertObjToMinObject(newObj)
 	// If we wrote out a new object, we need to update our state.
-	f.updateInodeStateAfterSync(newObj)
+	f.updateInodeStateAfterSync(minObj)
 	return
 }
 
-func (f *FileInode) updateInodeStateAfterSync(newObj *gcs.Object) {
-	if newObj != nil && !f.localFileCache {
-		var minObj gcs.MinObject
-		minObjPtr := storageutil.ConvertObjToMinObject(newObj)
-		if minObjPtr != nil {
-			minObj = *minObjPtr
-		}
-		f.src = minObj
+func (f *FileInode) updateInodeStateAfterSync(minObj *gcs.MinObject) {
+	if minObj != nil && !f.localFileCache {
+		f.src = *minObj
 		// Convert localFile to nonLocalFile after it is synced to GCS.
 		if f.IsLocal() {
 			f.local = false
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -500,19 +500,26 @@ func (f *FileInode) Read(
 func (f *FileInode) Write(
 	ctx context.Context,
 	data []byte,
-	offset int64) (err error) {
+	offset int64) error {
 	// For empty GCS files also we will trigger bufferedWrites flow.
 	if f.src.Size == 0 && f.config.Write.ExperimentalEnableStreamingWrites {
-		err = f.ensureBufferedWriteHandler(ctx)
+		err := f.ensureBufferedWriteHandler(ctx)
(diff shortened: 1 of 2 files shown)


internal/fs/inode/file.go
```diff
@@ -530,6 +537,49 @@ func (f *FileInode) Write(
...
+	}
+
+	return err
+}
+
+// Helper function to flush buffered writes handler and update inode state with
+// new object.
+//
+// LOCKS_REQUIRED(f.mu)
+func (f *FileInode) flushBufferedWriteHandlerAndUpdateInode() error {
```
vadlakondaswetha, line 565: Can you call it flushUsingBufferedWriteHandler

internal/fs/inode/file.go
```diff
@@ -623,10 +673,15 @@ func (f *FileInode) SetMtime(
...
 func (f *FileInode) Sync(ctx context.Context) (err error) {
 	// If we have not been dirtied, there is nothing to do.
-	if f.content == nil {
+	if f.content == nil && f.bwh == nil {
 		return
 	}
 
+	if f.bwh != nil {
+		// Finalize the object.
+		return f.flushBufferedWriteHandlerAndUpdateInode()
```
vadlakondaswetha, line 682: we shouldn't flush on sync call.
from LLD:

> Sync() can be called multiple times during the lifetime of the file write. Once the resumable upload is finalized we will not be able to append anymore data to it. But the intention of sync() call from the application side, is to persist all the data to the backend device. So we will block the syncFile() call and upload all pending buffers to GCS and free up the free buffers. We will not finalize the upload.

internal/fs/inode/file.go
```diff
@@ -667,6 +722,11 @@ func (f *FileInode) Sync(ctx context.Context) (err error) {
 	}
 
 	// If we wrote out a new object, we need to update our state.
+	f.updateInodeStateAfterSync(newObj)
+	return
+}
+
+func (f *FileInode) updateInodeStateAfterSync(newObj *gcs.Object) {
```
vadlakondaswetha, line 729: Sorry i missed this earlier, to avoid more dependencies on getting gcs.Object from bucket_handle whereas we use only gcs.MinObject everywhere, can you return gcs.MinObject directly for finalizeUpload method.

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -125,30 +125,33 @@ func (wh *BufferedWriteHandler) Sync() (err error) {
...
 		return nil, fmt.Errorf("BufferedWriteHandler.Flush(): %w", err)
 	}
+
 	err = wh.blockPool.ClearFreeBlockChannel()
 	if err != nil {
 		// Only logging an error in case of resource leak as upload succeeded.
 		logger.Errorf("blockPool.ClearFreeBlockChannel() failed: %v", err)
 	}
+
+	// Return an error along with object if the uploadHandler failed in between.
```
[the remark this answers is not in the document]
ashmeenkaur: We can't return without an error if all blocks were not uploaded properly, right? If the error happened after write and before flush,  errors will be silently ignored.

internal/fs/inode/file.go
ashmeenkaur: Done in separate PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/2795

internal/fs/inode/file.go
bk-rubrik: Hey, I wanted to follow this. Is it possible to get the access to the LLD doc?

bhawesh-khandelwal: Hey, I wanted to follow this. Is it possible to get access to the LLD?

PR #2794, opened 2024-12-13
update buffered write sync and flush methods

ashmeenkaur:
### Description

This PR includes changes to:

**BufferedWriteHandler:**
- The Sync call now waits for all in-progress blocks to be processed.
- The Flush call finalizes the object as possible in case of an upload failure. It will return an error along with the object if a failure occurs.
**UploadHandler:**
- The uploader goroutine now frees all blocks, even in the event of an upload failure.
- `AwaitBlocksUpload` has been added to wait for all blocks to be uploaded.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler_test.go
+++ b/internal/bufferedwrites/buffered_write_handler_test.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/fake"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/operations"
 	"github.com/jacobsa/timeutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


vadlakondaswetha (dismissed):

internal/bufferedwrites/buffered_write_handler_test.go
```diff
@@ -186,5 +187,61 @@ func (testSuite *BufferedWriteTest) TestFlush_SignalUploadFailureDuringWrite() {
...
+	// Write 5 blocks.
+	for i := 0; i < 5; i++ {
+		err = testSuite.bwh.Write(buffer, int64(blockSize*i))
+		require.Nil(testSuite.T(), err)
+	}
+
+	// Wait for 5 blocks to upload successfully.
+	err = testSuite.bwh.Sync()
+
+	assert.NoError(testSuite.T(), err)
```
vadlakondaswetha, line 229: is it possible to check if all blocks are uploaded from the uploadHandler.uploadCh?

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -235,3 +245,31 @@ func TestSignalUploadFailure(t *testing.T) {
...
+		err := t.uh.Upload(b)
+		require.NoError(t.T(), err)
+	}
+
+	// AwaitBlocksUpload.
+	t.uh.AwaitBlocksUpload()
+
+	// The blocks should be available on the free channel for reuse.
+	for _, expect := range blocks {
+		got := <-t.uh.freeBlocksCh
```
vadlakondaswetha, line 271: can we simply the check the count on uploadCh and freeBlockCh?

internal/bufferedwrites/buffered_write_handler_test.go
ashmeenkaur: Added these two checks: 
```
	assert.Equal(testSuite.T(), 0, len(testSuite.bwh.uploadHandler.uploadCh))
	assert.Equal(testSuite.T(), 5, len(testSuite.bwh.blockPool.FreeBlocksChannel()))
```

vadlakondaswetha (approved):

PR #2796, opened 2024-12-13
Enable-gVNIC in gke scripts

gargnitingoogle:
### Description
This addresses internal bug id 383980261 .

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


raj-prince (approved):
Assuming we have tested this command once.

gargnitingoogle:
> Assuming we have tested this command once.

For testing, I created a new cluster with the script, and it got created with NIC type gVNIC . So this works as expected.

PR #2813, opened 2024-12-17
Adds Negative cache TTL

ankitaluthra1:
### Description
This PR adds a new hidden flag for negative cache TTL.
The value defaults to 5 secs.
The changes include unit tests and are manually verified on GKE as well.
Integration tests will be added in subsequent PR.

NA

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA


cfg/config.go
```diff
@@ -200,6 +200,8 @@ type MetadataCacheConfig struct {
 
 	TtlSecs int64 `yaml:"ttl-secs"`
 
+	NegativeTtlSecs int64 `yaml:"negative-ttl-secs"`
```
Tulsishah, line 203: You missed initializing flag in [params.yaml](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/cfg/params.yaml)

Tulsishah (commented):
All over it looks good. Do you think we should add e2e/composite tests related to this in this PR?

cfg/config.go
ankitaluthra1: added

Tulsishah (approved):

cfg/config.go
```diff
@@ -196,6 +196,8 @@ type MetadataCacheConfig struct {
 
 	ExperimentalMetadataPrefetchOnMount string `yaml:"experimental-metadata-prefetch-on-mount"`
 
+	NegativeTtlSecs int64 `yaml:"negative-ttl-secs"`
```
charith87, line 199: Could you add a integration test for testing negative-ttl-secs.  You could maybe have a test that the 2nd read after Write expects a FileNotFound error with high negative-ttl enabled.

ankitaluthra1: yes I am planning to add 3 basic cases - ttl: 0, ttl -1 and default value in the subsequent PR.

Issue #2792, opened 2024-12-13
Kernel List Cache TTL is not working on the linux with kernel version 6.9.x onwards


raj-prince:
**Describe the issue**
Kernel list cache TTL doesn't work with the kernel linux version 6.9.x onwards. That means GCSFuse will return the stale listing after expiry of TTL unless eviction happens because of creation, renaming, or deletion of new files or folders.

**System & Version (please complete the following information):**
 - OS: Ubuntu: 2410 (with linux version 6.9.x onwards)
 - Platform all
 - Version: all kernel-list-cache supported version.

**SLO:**
We strive to respond to all bug reports within 24 business hours provided the information mentioned above is included.

raj-prince:
Kernel has fixed this issue with [commit](https://github.com/torvalds/linux/commit/03f275adb8fbd7b4ebe96a1ad5044d8e602692dc), and should be available in v6.13.0 onwards.

PR #2814, opened 2024-12-17
skip kernel list cache test for unsupported kernel version

ashmeenkaur:
### Description
skip kernel list cache test for unsupported kernel version 6.9.x - Ref issue #2792

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/kernel_list_cache_test.go
+++ b/internal/fs/kernel_list_cache_test.go
@@ -26,6 +26,7 @@ package fs_test
 import (
 	"os"
 	"path"
+	"regexp"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/fs/kernel_list_cache_test.go
```diff
@@ -120,6 +122,12 @@ func (t *KernelListCacheTestWithPositiveTtl) SetupSuite() {
 }
 
 func TestKernelListCacheTestWithPositiveTtlSuite(t *testing.T) {
+	kernelVersion := operations.KernelVersion(t)
```
Tulsishah, line 125: Can we create function for this instead of repeating?

Tulsishah (approved):

internal/fs/kernel_list_cache_test.go
```diff
@@ -120,6 +122,12 @@ func (t *KernelListCacheTestWithPositiveTtl) SetupSuite() {
...
+	matched, err := regexp.MatchString(UnsupportedKernelVersion, kernelVersion)
```
charith87, line 126: If possible, we have a utils class we could have a top level function, isKernelListCacheEnabled() and then it makes the tests more readable and one place to update accordingly if other linux versions come or needs updation.

You can also add a todo in the code if we want to unblock currently and do it later.

ashmeenkaur: I think the best place to keep this would be in fstest setupTestSuite but currently it does not support taking Testing.T. I have created a TODO bug to fix it once we fully migrate to stretchr testify.

charith87 (approved):

PR #2819, opened 2024-12-18
fix bug in csi driver build

gargnitingoogle:
### Description
This fixes a bug unintentionally introduced in #2818 . In that changeset, the bash function `uuid` was printing out some more information other than the uuid value itself because of the installation step. In bash, the printed log becomes the output of a function, hence the installation inside this function was changing the output of the `uuid` function breaking the caller. So, I have moved the installation out to wherever we're invoking this function.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/testing_on_gke/examples/run-gke-tests.sh
+++ b/perfmetrics/scripts/testing_on_gke/examples/run-gke-tests.sh
@@ -514,13 +514,6 @@ function ensureGcsFuseCsiDriverCode() {
 }
 
 uuid() {
-  if ! which uuidgen; then
-    # try to install uuidgen
-    sudo apt-get update && sudo apt-get install -y uuid-runtime
-    # confirm that it got installed.
-    which uuidgen
-  fi
-
   echo $(uuidgen) | sed -e "s/\-//g" ;
 }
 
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/testing_on_gke/examples/run-gke-tests.sh
```diff
@@ -565,6 +558,12 @@ function createCustomCsiDriverIfNeeded() {
     make generate-spec-yaml
     printf "\nBuilding a new custom CSI driver using the above GCSFuse binary ...\n\n"
     registry=gcr.io/${project_id}/${USER}/${cluster_name}
+    if ! which uuidgen; then
+      # try to install uuidgen
```
raj-prince, line 562: Should we move this to installDependencies section?

raj-prince (commented):
Good to update description with - what was the issue? By looking at the code I don't see much difference.

perfmetrics/scripts/testing_on_gke/examples/run-gke-tests.sh
gargnitingoogle: That was my first thought, but this utility isn't needed unless user wants to build custom driver. Hence, I have kept it here to avoid unnecessary installation/checking/log.

In the future, I'll moving around more of these installations to wherever they are needed.

gargnitingoogle:
> Good to update description with - what was the issue? By looking at the code I don't see much difference.

I've updated description to explain the issue.

PR #2822, opened 2024-12-19
Add negative cache entry after deleting object/folderCache entry update

codechanges:
### Description
After deleting an object or folder populate negative cache immediately.
This will save call/s to a backend to verify existence of a said object/folder resulting in perf gains.

Added invalidate logic in case of error to better handle all scenarios in case delete[object/folder] fails.

NA

### Testing details
1. Manual - NA
2. Unit tests - Yes
3. Integration tests - Yes

### DIFF
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -378,6 +378,9 @@ func (b *fastStatBucket) DeleteObject(
 
 func (b *fastStatBucket) DeleteFolder(ctx context.Context, folderName string) error {
 	err := b.wrapped.DeleteFolder(ctx, folderName)
+	// In case of an error; invalidate the cached entry. This will make sure that
+	// gcsfuse is not caching possibly erroneous status of the folder and next
+	// call will hit GCS backend to probe the latest status.
 	if err != nil {
 		b.invalidate(folderName)
 	} else {
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -367,18 +367,22 @@ func (b *fastStatBucket) UpdateObject(
 func (b *fastStatBucket) DeleteObject(
 	ctx context.Context,
 	req *gcs.DeleteObjectRequest) (err error) {
-	b.invalidate(req.Name)
 	err = b.wrapped.DeleteObject(ctx, req)
+	if err != nil {
+		b.invalidate(req.Name)
+	} else {
+		b.addNegativeEntry(req.Name)
+	}
 	return
 }
 
 func (b *fastStatBucket) DeleteFolder(ctx context.Context, folderName string) error {
 	err := b.wrapped.DeleteFolder(ctx, folderName)
 	if err != nil {
-		return err
+		b.invalidate(folderName)
+	} else {
+		b.addNegativeEntryForFolder(folderName)
 	}
-	// TODO: Caching negative entries for both objects and folders will be implemented together due to test failures.
-	b.invalidate(folderName)
 	return err
 }
 
--- a/tools/integration_tests/implicit_dir/local_file_test.go
+++ b/tools/integration_tests/implicit_dir/local_file_test.go
@@ -37,7 +37,7 @@ var (
 // //////////////////////////////////////////////////////////////////////
 
 func TestNewFileUnderImplicitDirectoryShouldNotGetSyncedToGCSTillClose(t *testing.T) {
-	testDirPath = setup.SetupTestDirectory(testDirName)
+	testDirPath = setup.SetupTestDirectory(path.Join(testDirName, operations.GetRandomName(t)))
 	CreateImplicitDir(ctx, storageClient, testDirName, t)
 	fileName := path.Join(ImplicitDirName, FileName1)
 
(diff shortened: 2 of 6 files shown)


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -367,18 +367,22 @@ func (b *fastStatBucket) UpdateObject(
 func (b *fastStatBucket) DeleteObject(
 	ctx context.Context,
 	req *gcs.DeleteObjectRequest) (err error) {
-	b.invalidate(req.Name)
 	err = b.wrapped.DeleteObject(ctx, req)
+	if err != nil {
+		b.invalidate(req.Name)
```
Tulsishah, line 372: Why are we invalidating if entry is not deleted

I think it should be like this
```
If err != nil {
return
}
b.addNegativeEntry(req.Name)
```

Tulsishah (changes requested):
Please add the integration test label to validate that all tests are functioning correctly. Also, please include a composite test for this scenario.

internal/storage/caching/fast_stat_bucket.go
codechanges: Thanks for your comment.

My thinking behind this was to not to assume that DeleteObject's failure means object is not deleted rather it might leave object's existence in precarious state from GCSFuse perspective for e.g. call to GCS is succeeded in the backend but somehow GCSFuse did not receive confirmation(momentary network loss) so in this scenario maintaining it in a cache is not a right thing to do as this makes an assumption that object is not deleted.

If we erase such entry from a cache then these assumptions regarding whether the object is deleted or not does not matter and next access to such objects will flow to GCS. 

Also, I do not think this should hamper performance as we are invalidating entry only in case of errors which I guess should be far and few. Still, I would like to hear from others' opinions about such a logic.

Tulsishah: Hmm okay.. then can you please write comment on it about the same.

tools/integration_tests/kernel_list_cache/infinite_kernel_list_cache_test.go
```diff
@@ -527,7 +527,7 @@ func TestInfiniteKernelListCacheTest(t *testing.T) {
 
 	// Define flag set to run the tests.
 	flagsSet := [][]string{
-		{"--kernel-list-cache-ttl-secs=-1"},
+		{"--kernel-list-cache-ttl-secs=-1", "--metadata-cache-ttl-secs=0"},
```
Tulsishah, line 530: Add TODO here , that we can change this with negative-cache-ttl 0 when it will introduce.

codechanges: Added

internal/storage/caching/fast_stat_bucket.go
codechanges: Added

tools/integration_tests/kernel_list_cache/infinite_kernel_list_cache_test.go
```diff
@@ -526,8 +526,9 @@ func TestInfiniteKernelListCacheTest(t *testing.T) {
 	}
 
 	// Define flag set to run the tests.
+	// TODO: Replace metadata-cache-ttl-secs with something better
```
raj-prince, line 529: Should we write high-level why are we disabling the metadata cache for this testsuite?

internal/storage/caching/fast_stat_bucket.go
raj-prince: Still don't see the comment, seems you haven't pushed the changes?

raj-prince (dismissed):
LGTM.

tools/integration_tests/kernel_list_cache/infinite_kernel_list_cache_test.go
codechanges: Added

codechanges:
> Please add the integration test label to validate that all tests are functioning correctly. Also, please include a composite test for this scenario.

Added integration test label. I think that, not handling error properly is a wider issue and needs to be taken care of in a different PR which should include composite test. I will raise a separate bug for the same and keep the scope of this PR limited for now.

Tulsishah (approved):

tools/integration_tests/util/setup/setup.go
```diff
@@ -382,6 +382,18 @@ func SetupTestDirectory(testDirName string) string {
 	return testDirPath
 }
 
+// SetupTestDirectoryRecursive recursively creates a testDirectory in the mounted directory and cleans up
+// any content present in it.
+func SetupTestDirectoryRecursive(testDirName string) string {
```
raj-prince, line 387: Should we pass another boolean variable (recursive) in the existing method `SetupTestDirectory` ?

Although, this looks more readable to me on the cost of duplicity. So, we can keep as it is.

tools/integration_tests/implicit_dir/local_file_test.go
```diff
@@ -37,21 +37,23 @@ var (
 // //////////////////////////////////////////////////////////////////////
 
 func TestNewFileUnderImplicitDirectoryShouldNotGetSyncedToGCSTillClose(t *testing.T) {
-	testDirPath = setup.SetupTestDirectory(testDirName)
-	CreateImplicitDir(ctx, storageClient, testDirName, t)
+	dirName := path.Join(testDirName, operations.GetRandomName(t))
```
raj-prince, line 40: Should we rename it something like, testCaseBaseDirName or similar? - dirName looks very generic to me, and makes the code less readable.

Also, should we log the name in the test for debuggability purpose?

tools/integration_tests/util/setup/setup.go
codechanges: I thought about that but, as you mentioned, I did not want to clutter the existing function hence I chose to write a different one. Lets keep it as it is for clarity purpose.

tools/integration_tests/implicit_dir/local_file_test.go
codechanges: Renamed variables, logging can be added later when we take up larger refactoring of these test cases.

raj-prince (approved):

PR #2823, opened 2024-12-19
Setting bwh to nil after all fileHandles are closed

vadlakondaswetha:
### Description
Setting bwh to nil after all fileHandles are closed in case of error.
Follow up PR will be raised to clear the buffers

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -53,12 +53,8 @@ type FileHandle struct {
 	readOnly bool
 }
 
-func NewFileHandle(
-	inode *inode.FileInode,
-	fileCacheHandler *file.CacheHandler,
-	cacheFileForRangeRead bool,
-	metricHandle common.MetricHandle,
-	readOnly bool) (fh *FileHandle) {
+// LOCKS_REQUIRED(fh.inode.mu)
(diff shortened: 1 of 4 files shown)


internal/fs/handle/file.go
```diff
@@ -47,16 +47,27 @@ type FileHandle struct {
...
 	cacheFileForRangeRead bool
 	metricHandle          common.MetricHandle
+	// For now, we will consider the files which are open in append mode also as write,
+	// as we are not doing anything special for append. When required we will
+	// define an enum instead of boolean to hold the type of open.
+	readOnly bool
 }
 
-func NewFileHandle(inode *inode.FileInode, fileCacheHandler *file.CacheHandler, cacheFileForRangeRead bool, metricHandle common.MetricHandle) (fh *FileHandle) {
+func NewFileHandle(
```
ashmeenkaur, line 56: Nit: Let's add a comment that this assumes inode lock. Same for Destroy. `LOCKS_REQUIRED(fh.inode)`

Also method declaration should be on a single line to avoid indentation confusion https://google.github.io/styleguide/go/decisions#function-formatting

ashmeenkaur (commented):
1 NIT comment, LGTM otherwise - assuming we will clean up the buffers in follow up PR.

ashmeenkaur (approved):

Issue #2828, opened 2024-12-22
GCSFuse is extremely slow for `StableDiffusionPipeline.from_single_file`


wlhee:
**Describe the issue**
It takes 13mins+ for `StableDiffusionPipeline.from_single_file` to load 7.2GB weights (`stable-diffusion-v1-5/stable-diffusion-v1-5/v1-5-pruned.safetensors`).

**System & Version (please complete the following information):**
 - OS: [e.g. Ubuntu 20.04]: N/A
 - Platform [GCE VM, GKE, Vertex AI]: Cloud Run
 - Version [Gcsfuse version and GKE version]: HEAD at repo

**Steps to reproduce the behavior with following information:**
1. Please share Mount command including all command line or config flags used to mount the bucket.

```
gcsfuse -o ro --implicit-dirs --client-protocol=http1 --max-conns-per-host=100 \
          $BUCKET_NAME $MOUNT_PATH
```

4. Please rerun with --log-severity=TRACE --foreground as additional flags to enable debug logs.

6. Monitor the logs and please capture screenshots or copy the relevant logs to a file (can use --log-format and --log-file as well).
7. Attach the screenshot or the logs file to the bug report here.
[downloaded-logs-20241222-111514.csv](https://github.com/user-attachments/files/18222999/downloaded-logs-20241222-111514.csv)

**Additional context**
1. Why file-cache feature is not used? It's because Cloud Run doesn't support local SSD or PD, and it's local filesystem is backed by memory. Hence having a file-cache would consume too much RAM due to the large model files
2. What is the problem? The application `StableDiffusionPipeline.from_single_file` uses safetensors that uses mmap under the hood (see https://github.com/huggingface/diffusers/pull/10305). It is not a pure sequential read pattern. The reads have the following characteristics: a. multiple processes share a single fd and file handle, which means there are seeking back and forth for the same file handle; b. for each pid, it's also somewhat seeky. [More investigation can be found in Google-internal bug ID: 381955920]
3. We could make GCSFuse much faster to handle this case by (a) have a per pid GCS reader / stream within the same file handle, so that each pid can seek independently; (b) GCSFuse can be less aggressive when it determines the range for each stream, so that each pid is not trapped in a 1MB stream forever. 
4. Early results shows the optimization can reduce the read time from 13mins to 1min.

**SLO:**
We strive to respond to all bug reports within 24 business hours provided the information mentioned above is included.

abhishek10004:
[...]

PR #2829, opened 2024-12-22
Support `enable-per-pid-reader` and `min-sequential-read-size-mb` to accelerate mmap reads

wlhee:
### Description
Support `enable-per-pid-reader` and `min-sequential-read-size-mb` to accelerate mmap reads.

1. For each file handle, creates per pid GCS reader stream and stats (start, limit, read bytes count), so that each pid can seek forward and backward independently without interfering with each other
2. Allow configuring the minimum sequential-read-size-mb so that each GCS stream can have a minimum bytes that it can support.
3. This PR firstly refactored the random_reader to move all the GCS reading logics from `randomReader` to `objectRangeReader` so that each pid can have its own `objectRangeReader` and use it indepently.

Initial manual results show it reduced loading time from 13mins to 1min by setting `perPidReader: true` and `minSequentialReadSizeMb: 200`

Before:
Screenshot 2024-12-22 at 11 54 19 AM

After:
Screenshot 2024-12-22 at 11 54 27 AM

https://github.com/GoogleCloudPlatform/gcsfuse/issues/2828

### Testing details
1. Manual - NA
5. Unit tests - TODO
6. Integration tests - TODO

### DIFF
--- a/internal/gcsx/random_reader.go
+++ b/internal/gcsx/random_reader.go
@@ -84,85 +84,30 @@ type RandomReader interface {
 // reads using the given bucket.
 func NewRandomReader(o *gcs.MinObject, bucket gcs.Bucket, sequentialReadSizeMb int32, fileCacheHandler *file.CacheHandler, cacheFileForRangeRead bool, metricHandle common.MetricHandle) RandomReader {
 	return &randomReader{
-		object:                o,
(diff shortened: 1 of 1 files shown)


wlhee:
Note: I haven't added flags or tests. Sent out this PR first to get early feedback.

PR #2839, opened 2024-12-24
Added destroy method to free up the memory buffers

vadlakondaswetha:
### Description
Added destroy method which will clean up the memory buffers when bwh is unInitialized.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -174,7 +174,7 @@ func (uh *UploadHandler) Destroy() {
 	for {
 		select {
 		case currBlock, ok := <-uh.uploadCh:
-			// Not ok means channel not closed. Return.
+			// Not ok means channel closed. Return.
 			if !ok {
 				return
 			}
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -186,6 +186,15 @@ func (uh *UploadHandler) Destroy() {
 			uh.wg.Done()
 		}
 	}
+
+	// This code is safer when not executed from multiple go-routines at once.
+	// Since destroy takes a inode lock in fileHandle, this is called by
+	// only one goroutine at once.
+	select {
+	case <-uh.uploadCh:
+	default:
+		close(uh.uploadCh)
+	}
 }
 
 // TODO: Move this method to util and add unit tests.
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"io"
 	"sync"
-	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/block"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
@@ -187,23 +186,5 @@ func (uh *UploadHandler) Destroy() {
 			close(uh.uploadCh)
 			return
 		}
-	}
-}
-
-// TODO: Move this method to util and add unit tests.
-// waitTimeout waits for the waitGroup for the specified max timeout.
-// Returns true if waiting timed out.
-func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
-	c := make(chan struct{})
-	go func() {
-		defer close(c)
-		wg.Wait()
-	}()
-
-	select {
-	case <-c:
-		return false // completed normally
-	case <-time.After(timeout):
-		return true // timed out
 	}
 }
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -171,29 +171,22 @@ func (uh *UploadHandler) AwaitBlocksUpload() {
 }
 
 func (uh *UploadHandler) Destroy() {
-	// Waiting for upload routine to move all blocks to freeChannel.
-	timedOut := waitTimeout(&uh.wg, 10*time.Second)
-
-	// TimedOut means there are some blocks which are still in uploadChannel,
-	// either because upload is stuck in uploading a chunk or the upload
-	// go-routine crashed. Copying all pending blocks to freeBlock channel for cleanup.
-	// We can clean up from uploadChannel also, but to ensure clean up happens
-	// from one place we are copying them to freeBlock channel.
-	if timedOut {
-		for currBlock := range uh.uploadCh {
+	// Move all pending blocks to freeBlockCh and close the channel if not done.
+	for {
+		select {
+		case currBlock, ok := <-uh.uploadCh:
+			// Not ok means channel not closed. Return.
+			if !ok {
+				return
+			}
 			uh.freeBlocksCh <- currBlock
 			// Marking as wg.Done to ensure any waiters are unblocked.
 			uh.wg.Done()
+		default:
+			// This will get executed when there are no blocks pending in uploadCh and its not closed.
+			close(uh.uploadCh)
+			return
 		}
-	}
-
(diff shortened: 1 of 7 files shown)


internal/bufferedwrites/upload_handler.go
```diff
@@ -154,3 +155,39 @@ func (uh *UploadHandler) SignalUploadFailure() chan error {
...
+func (uh *UploadHandler) Destroy() {
+	// Waiting for upload routine to move all blocks to freeChannel.
+	timedOut := waitTimeout(&uh.wg, 10*time.Second)
+
+	// TimedOut means there are some blocks which are still in uploadChannel,
+	// either because upload is stuck in uploading a chunk or the upload
+	// go-routine crashed. Copying all pending blocks to freeBlock channel for cleanup.
+	// We can clean up from uploadChannel also, but to ensure clean up happens
+	// from one place we are copying them to freeBlock channel.
+	if timedOut {
```
charith87, line 168: Do we want to add a test case to test the timedOut scenario?

internal/fs/handle/file.go
```diff
@@ -71,9 +71,13 @@ func NewFileHandle(inode *inode.FileInode, fileCacheHandler *file.CacheHandler,
 
 // Destroy any resources associated with the handle, which must not be used
 // again.
+// LOCK_FUNCTION(fh.inode.mu)
+// UNLOCK_FUNCTION(fh.inode.mu)
 func (fh *FileHandle) Destroy() {
 	// Deregister the fileHandle with the inode.
+	fh.inode.Lock()
```
charith87, line 78: Trying to understand if locking is necessary here or not?

vadlakondaswetha: Yes lock is required for DeRegisterFileHandle. This is a new method which we added in the last PR only, at that time lock was taken at fs level. Now i am lowering it to inode level.

internal/bufferedwrites/upload_handler.go
vadlakondaswetha: As mentioned in the comment, its a crash scenario which i am able to replicate in the unit test. So not able to test case.

internal/fs/handle/file.go
```diff
@@ -71,9 +71,13 @@ func NewFileHandle(inode *inode.FileInode, fileCacheHandler *file.CacheHandler,
...
 	fh.inode.DeRegisterFileHandle(fh.readOnly)
+	fh.inode.Unlock()
 	if fh.reader != nil {
```
ashmeenkaur, line 81: do we need file handle lock here?

internal/bufferedwrites/upload_handler.go
```diff
@@ -154,3 +155,39 @@ func (uh *UploadHandler) SignalUploadFailure() chan error {
 func (uh *UploadHandler) AwaitBlocksUpload() {
 	uh.wg.Wait()
 }
+
+func (uh *UploadHandler) Destroy() {
+	// Waiting for upload routine to move all blocks to freeChannel.
```
ashmeenkaur, line 160: I think we need something like this here:

```
	// Attempt to close the upload channel in case it is not closed.
	func() {
		defer func() {
			_ = recover()
		}()
		close(uh.uploadCh)
	}()
```

if we won't close the upload channel, then the cleanup for loop will never exit.
```
for currBlock := range uh.uploadCh {
			uh.freeBlocksCh <- currBlock
			// Marking as wg.Done to ensure any waiters are unblocked.
			uh.wg.Done()
		}
```

internal/bufferedwrites/upload_handler.go
ashmeenkaur: We can add this simulated test case but the timeout is too large here:


```
func (t *UploadHandlerTest) TestDestroyWithTimeout() {
	// Simulate timeout.
	t.uh.wg.Add(1)
	b, err := t.blockPool.Get()
	require.NoError(t.T(), err)
	t.uh.uploadCh <- b

	t.uh.Destroy()

	assertAllBlocksProcessed(t.T(), t.uh)
	assert.Equal(t.T(), 1, len(t.uh.freeBlocksCh))
	assert.Equal(t.T(), 0, len(t.uh.uploadCh))
}
```

internal/fs/handle/file.go
charith87: I couldnt see the fs level lock being removed in this commit. Was in a previous commit?

vadlakondaswetha: it is in this PR only. in fs.go changes.

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -340,3 +344,54 @@ func (t *UploadHandlerTest) TestCreateObjectChunkWriterIsCalledWithCorrectReques
...
+				b, err := t.blockPool.Get()
+				require.NoError(t.T(), err)
+				blocks = append(blocks, b)
+			}
+			// CreateObjectChunkWriter -- should be called once.
+			writer := &storagemock.Writer{}
+			t.mockBucket.On("CreateObjectChunkWriter", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(writer, nil)
+			// Upload the blocks.
+			for _, b := range blocks {
+				err := t.uh.Upload(b)
```
ashmeenkaur, line 377: Instead of calling uploader go routine, shall we manually add to uploadCh and increment wait group. That way we will be sure that the buffers are  being cleared via Destroy

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -340,3 +344,54 @@ func (t *UploadHandlerTest) TestCreateObjectChunkWriterIsCalledWithCorrectReques
...
+		{
+			name:           "UploadChClosed",
+			uploadChClosed: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func() {
+			// Create some blocks.
+			var blocks []block.Block
```
ashmeenkaur, line 366: This is reused at multiple places. Shall we create a test helper to create 5 blocks?

internal/fs/handle/file.go
```diff
@@ -71,9 +71,13 @@ func NewFileHandle(inode *inode.FileInode, fileCacheHandler *file.CacheHandler,
 
 // Destroy any resources associated with the handle, which must not be used
 // again.
```
ashmeenkaur, line 73: NIT:

```suggestion
// again.
// LOCKS_REQUIRED(fh.mu)
```

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -340,3 +344,54 @@ func (t *UploadHandlerTest) TestCreateObjectChunkWriterIsCalledWithCorrectReques
...
+			if tc.uploadChClosed {
+				close(t.uh.uploadCh)
+			}
+
+			t.uh.Destroy()
+
+			assertAllBlocksProcessed(t.T(), t.uh)
+			assert.Equal(t.T(), 5, len(t.uh.freeBlocksCh))
+			assert.Equal(t.T(), 0, len(t.uh.uploadCh))
+
```
ashmeenkaur, line 389: NIT:
```suggestion
```

ashmeenkaur (approved):
Minor comments

ashmeenkaur (approved):

internal/bufferedwrites/upload_handler.go
vadlakondaswetha: removed the timeout flow.

internal/bufferedwrites/upload_handler.go
vadlakondaswetha: Changed the logic to handle it via select statement.

charith87 (approved):

PR #2831, opened 2024-12-23
[Move Object] Integrating MoveObject in all the bucket layers

Tulsishah:
### Description
Integrating MoveObject in all the bucket layers.
- Bucket handle
- Prefix bucket
- Throttle bucket
- debug bucket
- testify bucket
- fake bucket
- mock bucket
- Fast Stat bucket

### Link to the issue in case of a bug fix.
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

PR #2840, opened 2024-12-25
[Move Object] Bucket handle integration

Tulsishah:
### Description
- MoveObject bucket handle integration
- Unit tests are not covered as we're currently using a fake storage server for the rest of our storage client API unit tests. To add unit tests for the MoveObject function, we have two options:
  1. Add a moveObject API to the [fake GCS server repository](https://github.com/fsouza/fake-gcs-server/blob/main/fakestorage/object.go,). This would involve extending the fake server's functionality to include support for the move operation. This seems like a more complex solution and time consuming.
  2. Mock the MoveObject API. Mocking is not possible because the MoveObject function takes a [private](https://github.com/googleapis/google-cloud-go/blob/main/storage/http_client.go#L596) struct from storage library as an argument.  

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -37,6 +37,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
+	"google.golang.org/grpc/codes"
 )
 
 const FullFolderPathHNS = "projects/_/buckets/%s/folders/%s"
@@ -536,6 +537,13 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
 	return err
 }
 
+func isPreconditionFailed(err error) bool {
+	var gapiErr *googleapi.Error
+	var apiErr *apierror.APIError
+	return (errors.As(err, &gapiErr) && gapiErr.Code == http.StatusPreconditionFailed) ||
+		(errors.As(err, &apiErr) && apiErr.GRPCStatus().Code() == codes.FailedPrecondition)
+}
+
 func (bh *bucketHandle) MoveObject(ctx context.Context, req *gcs.MoveObjectRequest) (*gcs.Object, error) {
 	var o *gcs.Object
 	var err error
@@ -565,17 +573,17 @@ func (bh *bucketHandle) MoveObject(ctx context.Context, req *gcs.MoveObjectReque
 
 	// If storage object does not exist, httpclient is returning ErrObjectNotExist error instead of googleapi error
 	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L516
-	switch ee := err.(type) {
-	case *googleapi.Error:
-		if ee.Code == http.StatusPreconditionFailed {
-			err = &gcs.PreconditionError{Err: ee}
-		}
-	default:
-		if err == storage.ErrObjectNotExist {
+	if err != nil {
+		if isPreconditionFailed(err) {
+			var gapiErr *googleapi.Error
+			errors.As(err, &gapiErr)
+			err = &gcs.PreconditionError{Err: gapiErr}
+		} else if errors.Is(err, storage.ErrObjectNotExist) {
 			err = &gcs.NotFoundError{Err: storage.ErrObjectNotExist}
 		} else {
 			err = fmt.Errorf("error in updating object: %w", err)
 		}
+		return nil, err
 	}
 
 	return o, nil
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -549,11 +549,18 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
 	return err
 }
 
-func isPreconditionFailed(err error) bool {
+func isPreconditionFailed(err error) (bool, error) {
 	var gapiErr *googleapi.Error
+	if errors.As(err, &gapiErr) && gapiErr.Code == http.StatusPreconditionFailed {
+		return true, &gcs.PreconditionError{Err: gapiErr}
+	}
+
 	var apiErr *apierror.APIError
-	return (errors.As(err, &gapiErr) && gapiErr.Code == http.StatusPreconditionFailed) ||
-		(errors.As(err, &apiErr) && apiErr.GRPCStatus().Code() == codes.FailedPrecondition)
+	if errors.As(err, &apiErr) && apiErr.GRPCStatus().Code() == codes.FailedPrecondition {
+		return true, &gcs.PreconditionError{Err: apiErr}
+	}
+
+	return false, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/storage/bucket_handle.go
```diff
@@ -537,8 +537,48 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
...
+
+	attrs, err := obj.Move(ctx, dstMoveObject)
+	if err == nil {
+		// Converting objAttrs to type *Object
+		o = storageutil.ObjectAttrsToBucketObject(attrs)
+	}
+
+	// If storage object does not exist, httpclient is returning ErrObjectNotExist error instead of googleapi error
+	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L516
+	switch ee := err.(type) {
```
tritone, line 568: This should use errors.As. Also, for gRPC, the error type will be different. See https://pkg.go.dev/cloud.google.com/go#hdr-Inspecting_errors for how to unpack.

internal/storage/bucket_handle.go
```diff
@@ -537,8 +537,48 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
 }
 
 func (bh *bucketHandle) MoveObject(ctx context.Context, req *gcs.MoveObjectRequest) (*gcs.Object, error) {
```
tritone, line 539: How are we limiting this to only be called for HNS buckets? Move object will not work for non-HNS.

I looked at the other PR #2831 but didn't see it.

internal/storage/bucket_handle.go
```diff
@@ -537,8 +537,48 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
...
+		}
+	default:
+		if err == storage.ErrObjectNotExist {
+			err = &gcs.NotFoundError{Err: storage.ErrObjectNotExist}
+		} else {
+			err = fmt.Errorf("error in updating object: %w", err)
+		}
+	}
+
+	return o, nil
```
tritone, line 581: It looks like err is never actually returned even it it's non-nil?

internal/storage/bucket_handle.go
```diff
@@ -537,8 +537,48 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
...
+	}
+
+	// Putting a condition that the metaGeneration of source should match *req.SrcMetaGenerationPrecondition for move operation to occur.
+	if req.SrcMetaGenerationPrecondition != nil {
+		obj = obj.If(storage.Conditions{MetagenerationMatch: *req.SrcMetaGenerationPrecondition})
+	}
+
+	dstMoveObject := storage.MoveObjectDestination{
+		Object:     req.DstName,
+		Conditions: nil,
```
tritone, line 557: Is there not a way to supply destination preconditions?

tritone (commented):
Can you add an integration test? It should work in prod now as long as your project or bucket prefix are allowlisted.

internal/storage/bucket_handle.go
Tulsishah: This condition will come when we do fs level integration. e.g. refer [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/fs.go#L1989)

Tulsishah:
> Can you add an integration test? It should work in prod now as long as your project or bucket prefix are allowlisted.

I did POC and ran integration tests with MoveObject, it is working fine. To write integration test here, I have to do full integration(fs layer). I'll cover it in later PRs.

internal/storage/bucket_handle.go
Tulsishah: Returned from err!=nil fix this.

internal/storage/bucket_handle.go
Tulsishah: Done. Thanks

internal/storage/bucket_handle.go
Tulsishah: Conditions constrain methods to act on specific generations of objects. For MoveObject, if the destination object does not exist, it will be created with a new generation and metageneration. On the other hand, if it exists, it will overwrite the generation and metageneration values. This behavior aligns with how POSIX file systems work. Therefore, I don't believe any destination conditions are required. Similarly, in CopyObject (Rewrite), we haven't defined any conditions on the destination either.

vadlakondaswetha: I guess the kernel verifies if the destination already exists and asks if we have to overwrite. Did you check this ?

Tulsishah: In a POSIX file system, the command mv a.txt b.txt (where both a.txt and b.txt exist) will overwrite b.txt with the contents of a.txt without prompting for confirmation. This behavior is consistent with my POC of MoveObject with GCSFuse.

internal/storage/bucket_handle.go
```diff
@@ -549,9 +549,61 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
...
+
+	// If storage object does not exist, httpclient is returning ErrObjectNotExist error instead of googleapi error
+	// https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/vendor/cloud.google.com/go/storage/http_client.go#L516
+	if err != nil {
+		if ok, preCondErr := isPreconditionFailed(err); ok {
+			err = preCondErr
+		} else if errors.Is(err, storage.ErrObjectNotExist) {
+			err = &gcs.NotFoundError{Err: storage.ErrObjectNotExist}
+		} else {
+			err = fmt.Errorf("error in updating object: %w", err)
```
vadlakondaswetha, line 601: error in moving not updating

internal/storage/bucket_handle.go
```diff
@@ -549,9 +549,61 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
...
+
+	dstMoveObject := storage.MoveObjectDestination{
+		Object:     req.DstName,
+		Conditions: nil,
+	}
+
+	attrs, err := obj.Move(ctx, dstMoveObject)
+	if err == nil {
+		// Converting objAttrs to type *Object
+		o = storageutil.ObjectAttrsToBucketObject(attrs)
```
vadlakondaswetha, line 590: you can return from here too. return o, nil
and then you dont need if statement below

if err == nil {
  return
}

 // handle error code

vadlakondaswetha (approved):

internal/storage/bucket_handle.go
```diff
@@ -549,9 +549,58 @@ func (bh *bucketHandle) DeleteFolder(ctx context.Context, folderName string) (er
 	return err
 }
 
+func isPreconditionFailed(err error) (bool, error) {
```
vadlakondaswetha, line 552: you can write unit tests for this method.

Tulsishah: Thanks for suggestion. added.

vadlakondaswetha (approved):

PR #2853, opened 2025-01-03
Fix kernel cache ttl test failure for GKE

Tulsishah:
### Description
Tests began failing due to execution sequence issues stemming from the addition of a negative cache upon deletion. To address this, the affected tests were moved to a separate file and configured with a metadata cache TTL of 0.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA


tools/integration_tests/kernel_list_cache/infinite_kernel_list_cache_deletion_test.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2025 Google LLC
```
raj-prince, line 1: Can we change the name to `infinite_kernel_list_cache_delete_dir` test?

raj-prince (approved):

PR #2858, opened 2025-01-06
Disable write back cache when streaming writes are enabled

ashmeenkaur:
### Description
- Disable write back cache when streaming writes are enabled.
- Resolve block size config to MB.

b/387914550

### Testing details
1. Manual - NA
2. Unit tests - Updated
3. Integration tests - NA

### DIFF
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -34,6 +34,8 @@ const (
 	Parallel   = "Parallel"
 
 	MaxMiBsInUint64 uint64 = math.MaxUint64 >> 20
+	MaxMiBsInInt64  int64  = math.MaxInt64 >> 20
+	MiB                    = 1024 * 1024
 
 	// HeapSizeToRssConversionFactor is a constant factor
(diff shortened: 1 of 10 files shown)


internal/util/util.go
```diff
@@ -34,6 +34,8 @@ const (
 	Parallel   = "Parallel"
 
 	MaxMiBsInUint64 uint64 = math.MaxUint64 >> 20
+	MaxMiBsInInt64  int64  = math.MaxInt64 >> 20
```
vadlakondaswetha, line 37: What are we trying to achieve here?

ashmeenkaur: Created this constant to prevent integer overflow when converting MiB to bytes. This constant, derived by dividing math.MaxInt64 by 2^20, represents the maximum MiB value that can be safely stored in an int64 variable.

vadlakondaswetha (approved):

vadlakondaswetha (approved):

PR #2860, opened 2025-01-06
Handle global max blocks semaphore properly

ashmeenkaur:
### Description

we are creating at least 1 block for each inode even if we are out of global max blocks (tracked by `globalMaxBlocksSem` semaphore), but at the time of releasing blocks we are not checking if the semaphore was acquired. This leads to `panic("semaphore: released more than held")`.
This PR fixes this panic by keeping track of whether the block was created by acquiring semaphore.

b/385057665

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA


internal/block/block_pool.go
```diff
@@ -71,6 +76,9 @@ func (bp *BlockPool) Get() (Block, error) {
 			// calls to a single file are serialized because of inode.lock().
 			if bp.totalBlocks < bp.maxBlocks {
 				freeSlotsAvailable := bp.globalMaxBlocksSem.TryAcquire(1)
+				if freeSlotsAvailable {
```
vadlakondaswetha, line 79: Consider this scenario:
1. First block is created when there are no slots in globalMaxBlocksSem
2. 2nd block is created where there are free slots in globalMaxBlocksSem. FreeSlotsAvailable is set to true.
When we release 2 blocks it will fail right?

If yes, can we keep it simple. Always create the 1st block without checking globalMaxBlocks and don't release the last block to globalMaxBlocks.

ashmeenkaur: Thanks, made the changes.

internal/block/block_pool_test.go
```diff
@@ -222,3 +285,70 @@ func (t *BlockPoolTest) TestBlockPool_FreeBlocksChannel() {
 	assert.NotNil(t.T(), ch)
 	assert.Equal(t.T(), freeBlocksCh, ch)
 }
+
+func (t *BlockPoolTest) TestBlockPool_canAllocateBlock() {
```
vadlakondaswetha, line 289: TestCanAllocateBlock is suficient.

vadlakondaswetha (approved):

internal/block/block_pool_test.go
ashmeenkaur: Will do it in a separate PR as discussed offline.

PR #2867, opened 2025-01-06
Set the precondition-errors flag to default true as feature is complete.

vipnydav:
### Description
Set the precondition-errors flag to default true as feature is complete.

NA

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA


tools/integration_tests/emulator_tests/proxy-od8hc
ashmeenkaur, line 1: is this mistakenly added?

vipnydav: Yes

ashmeenkaur (approved):

PR #2868, opened 2025-01-06
Composite tests for negative entry in stat cache after object/directory deletion

codechanges:
### Description
Add composite test for negative entry in stat cache after object/directory deletion

* Tests are added for both HNS and non-HNS type buckets
* Tests are added for both file and directory types.

http://b/387403444

### Testing details
1. Composite tests - Yes


internal/fs/hns_bucket_test.go
```diff
@@ -379,3 +385,137 @@ func (t *HNSBucketTests) TestCreateLocalFileInSamePathAfterDeletingParentDirecto
...
+		&cacheClock,
+		uncachedHNSBucket)
+
+	// Enable directory type caching.
+	t.serverCfg.DirTypeCacheTTL = ttl
+	t.serverCfg.ImplicitDirectories = false
+	t.serverCfg.NewConfig = &cfg.Config{
+		EnableHns: true,
+		FileCache: defaultFileCacheConfig(),
+		MetadataCache: cfg.MetadataCacheConfig{
```
Tulsishah, line 423: Can't we add this to the existing setup? I don't think it will cause any issues to existing tests

codechanges: I intentionally kept cached flow different. If we move this configuration to the existing one then that flow would start using the stat cache bucket and not the uncached one. Right now, both of these flows are being tested so I would suggest that lets keep them separate.

internal/fs/hns_bucket_test.go
```diff
@@ -379,3 +385,137 @@ func (t *HNSBucketTests) TestCreateLocalFileInSamePathAfterDeletingParentDirecto
...
+// ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+// --------------------- Test for delete object -------------------
+// Create directory
+// Create object in directory
+// Stat the object
+// Delete object
+// Create object using other client
+// Stat the object immediately, It should return not found error although object present
+// Stat the object after TTL expiry and it should appear
+func (t *HNSCachedBucketMountTest) TestLocalFileIsInaccessibleAfterDeleteObjectButPresentRemotely() {
```
Tulsishah, line 457: nit: Can we write the test in arrange, act , assert form?

Tulsishah (approved):

internal/fs/hns_bucket_test.go
codechanges: The test cases are simple and procedural enough, imo hence I would avoid making any changes which can impact readability. "Arrange, Act, Assert" can be applied test cases which are amenable to daisy chaining of blocks.

raj-prince (approved):
Relying on the Tulsi's review.

PR #2872, opened 2025-01-07
Refactor: ValidateStaleNFSFileHandleError for ARM64 Compatibility

vipnydav:
### Description
Updated the ValidateStaleNFSFileHandleError function to dynamically fetch the error message associated with syscall.ESTALE. This change ensures that the function correctly identifies stale file handle errors across different architectures, including ARM64 and AMD64, which may return different error messages for the same syscall error.

NA

### Testing details
1. Manual - Done
2. Unit tests - NA


ashmeenkaur (approved):
Hope you have tested on arm64 machine

PR #2878, opened 2025-01-08
[Move Object] Rename benchmark e2e perf test

Tulsishah:
### Description
- Adding rename tests in benchmarking e2e tests to validate execution time for rename operation.

[b/388391082](https://b.corp.google.com/issues/388391082)

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/benchmarking/setup_test.go
+++ b/tools/integration_tests/benchmarking/setup_test.go
@@ -37,7 +37,26 @@ var (
 	storageClient *storage.Client
 	ctx           context.Context
 	testDirPath   string
+	mountFunc     func([]string) error
+	// mount directory is where our tests run.
+	mountDir string
+	// root directory is the directory to be unmounted.
+	rootDir string
 )
+
+////////////////////////////////////////////////////////////////////////
+// Helpers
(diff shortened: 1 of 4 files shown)


tools/integration_tests/benchmarking/setup_test.go
```diff
@@ -58,8 +68,8 @@ func TestMain(m *testing.M) {
 	setup.SetUpTestDirForTestBucketFlag()
 
 	flagsSet := [][]string{
-		{"--stat-cache-ttl=0"},
-		{"--client-protocol=grpc", "--stat-cache-ttl=0"},
+		{"--stat-cache-ttl=0", "--enable-atomic-rename-object=true"},
```
raj-prince, line 71: Instead, can we do similar to `managed_folder` e2e test, where all the file (test_suites) will have their own mount configuration and expected latency?

Tulsishah: It required changes in all the files that's why I avoid initially. Anyways fixed it for all the existing tests also.

raj-prince (approved):

PR #2887, opened 2025-01-09
Negative cache integration tests

ankitaluthra1:
### Description
This PR adds e2e integration tests for negative stat cache, scenarios include:
1. Happy path - TTL set to a finite positive integer.
2. Disabled Cache - TTL set to 0
3. Infinite Cache - TTL set to -1

b/387905752

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Done

### DIFF
--- a/tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
+++ b/tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
+++ b/tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
@@ -57,7 +57,7 @@ func (s *disabledNegativeStatCacheTest) TestNegativeStatCacheDisabled(t *testing
 	assert.ErrorContains(t, err, "explicit_dir/file1.txt: no such file or directory")
 
 	// Adding the object with same name
-	client.CreateObjectInGCSTestDir(ctx, storageClient, testDirName, targetFile, "some-content", t)
+	client.CreateObjectInGCSTestDir(ctx, storageClient, testDirName, "explicit_dir/file1.txt", "some-content", t)
 
 	// File should be returned, as call will be served from GCS and gcsfuse should not return from cache
 	f, err := os.OpenFile(targetFile, os.O_RDONLY, os.FileMode(0600))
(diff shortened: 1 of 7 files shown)


tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
```diff
@@ -0,0 +1,91 @@
...
+	assert.NotNil(t, err)
+	// Assert the underlying error is File Not Exist
+	assert.ErrorContains(t, err, "explicit_dir/file1.txt: no such file or directory")
+
+	// Adding the object with same name
+	f1 := operations.CreateFile(targetFile, setup.FilePermission_0600, t)
+	operations.CloseFile(f1)
+
+	// File should be returned, as call will be served from GCS and gcsfuse should not return from cache
+	f, err := os.OpenFile(targetFile, os.O_RDONLY, os.FileMode(0600))
```
Tulsishah, line 63: operations.CreateFile will create file through GCSFuse so the positive entry will get injected and response will come from positive cache. To check this scenario we should create file from other client.

tools/integration_tests/negative_stat_cache/infinite_negative_stat_cache_test.go
```diff
@@ -0,0 +1,90 @@
...
+	// Create test directory
+	operations.CreateDirectory(targetDir, t)
+	targetFile := path.Join(targetDir, "file1.txt")
+
+	// Error should be returned as file does not exist
+	_, err := os.OpenFile(targetFile, os.O_RDONLY, os.FileMode(0600))
+
+	assert.NotNil(t, err)
+	// Assert the underlying error is File Not Exist
+	assert.ErrorContains(t, err, "explicit_dir/file1.txt: no such file or directory")
```
Tulsishah, line 57: can we create var or const for `explicit_dir/file1.txt` ?

Tulsishah (commented):
Add package name in [CI](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/run_e2e_tests.sh#L66) and [CD](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/cd_scripts/e2e_test.sh#L144) script.

tools/integration_tests/negative_stat_cache/infinite_negative_stat_cache_test.go
ankitaluthra1: for assertions literal constants are preferred over variable.. sos would keep that same in assertions.. rest of the places we are using variable targetFile only

tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
ankitaluthra1: fixed

Tulsishah (approved):

Tulsishah (approved):

PR #2904, opened 2025-01-16
Skipping Kernel List Cache E2E Test on the Unsupported Kernel Version

raj-prince:
### Description
1. Refactoring - moving the common utility b/w gcsfuse/internal and e2e tests in common package.
2. Added check to skip the e2e kernel list cache test for unsupported kernel version.

Ref: https://github.com/GoogleCloudPlatform/gcsfuse/issues/2792

### Testing details
1. Manual - Tested manually on unsupported kernel version.
4. Unit tests - Automation
5. Integration tests - Automation

### DIFF
--- a/common/util_test.go
+++ b/common/util_test.go
@@ -1,0 +1,88 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


common/util_test.go
```diff
@@ -0,0 +1,88 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			originalKernelVersion := kernelVersionForTest
+			kernelVersionForTest = func() (string, error) { return tc.mockKernelVersion, nil }
+			defer func() { kernelVersionForTest = originalKernelVersion }()
+
+			skip, err := IsKLCacheEvictionUnSupported()
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
```
ashmeenkaur, line 84:
```suggestion
			require.NoError(t, err)
```

raj-prince:
> LGTM, do we need to skip list large dir tests too? That also uses kernel list cache https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/list_large_dir/list_large_dir_test.go

Nope, because these tests don't perform or expect any TTL based eviction.

PR #2921, opened 2025-01-20
Skip monitoring tests from cd

kislaykishore:
### Description
Monitoring tests fail with installed_package. So, skip them from cd. They are already covered by ci tests.
b/391097316

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


Tulsishah (commented):
As discussed in on-call meeting, can we remove benchmarking package as well?

Tulsishah (approved):

PR #2922, opened 2025-01-20
Enable dynamic creation of storage clients based on bucket type and specified protocol

abhishek10004:
Enable dynamic creation of storage clients based on bucket type and specified protocol

Reason for not using mutex in getClient:
Multiple clients can be created in dynamic mounting case only. But in this case, the root dir will actually be a base_dir & we take a lock before calling LookUpChild for base_dir. So, this will ensure that call to BucketHandle (which means calls to getClient) will be serialized.

Reason for not adding a new config EnableBidiConfig in StorageClientConfig (& instead passing a separate parameter for bidi config while creating grpc client):
This config is created from the user provided config at the start (& should reflect that) and I would not like to keep a transient field to it (i.e. would not like to keep a field which has to be set every time before grpc client creation happens and does not serve any other purpose apart from reducing an extra parameter to grpc client creation call).

Testing: Manually tested mounting/listing/reading on flat, hns and zb bucket with this change using dynamic mounting (with both grpc as well as http)

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -112,7 +112,6 @@ func createClientOptionForGRPCClient(clientConfig *storageutil.StorageClientConf
 }
 
 func setRetryConfig(sc *storage.Client, clientConfig *storageutil.StorageClientConfig) {
-
 	if sc == nil || clientConfig == nil {
 		logger.Fatal("setRetryConfig: Empty storage client or clientConfig")
 		return
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -59,11 +59,12 @@ type StorageHandle interface {
 }
 
 type storageClient struct {
-	httpClient           *storage.Client
-	grpcClient           *storage.Client
-	clientProtocol       cfg.Protocol
-	storageControlClient StorageControlClient
-	directPathDetector   *gRPCDirectPathDetector
+	httpClient               *storage.Client
+	grpcClient               *storage.Client
+	grpcClientWithBidiConfig *storage.Client
+	clientConfig             storageutil.StorageClientConfig
+	storageControlClient     StorageControlClient
+	directPathDetector       *gRPCDirectPathDetector
 }
 
 type gRPCDirectPathDetector struct {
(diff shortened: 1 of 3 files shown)


internal/storage/storage_handle.go
```diff
@@ -261,35 +280,32 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
...
+		clientConfig:         clientConfig,
+		directPathDetector:   &gRPCDirectPathDetector{clientOptions: clientOpts},
+	}
+	return
+}
+
+func (sh *storageClient) getClient(ctx context.Context, isbucketZonal bool) (sc *storage.Client, err error) {
+	if isbucketZonal {
+		if sh.grpcClientWithBidiConfig == nil {
+			sh.grpcClientWithBidiConfig, err = createGRPCClientHandle(ctx, &sh.clientConfig, true)
```
vadlakondaswetha, line 294: instead of some elseif cases, can you simply it a bit. improves readability and easy to extend in future.

if isBucketZonal {
if sh.grpcClientWithBidiConfig == nil {
			sh.grpcClientWithBidiConfig, err = createGRPCClientHandle(ctx, &sh.clientConfig, true)
return sh.grpcClientWithBidiConfig, nil
}

if sh.clientConfig.ClientProtocol == cfg.GRPC {
		if sh.grpcClient == nil {
			sh.grpcClient, err = createGRPCClientHandle(ctx, &sh.clientConfig, false)
		}
	return sh.grpcClient, nil
	}

internal/storage/storage_handle_test.go
```diff
@@ -556,7 +556,7 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleWithCustomEndpointAndEna
 func (testSuite *StorageHandleTest) TestCreateClientOptionForGRPCClient() {
 	sc := storageutil.GetDefaultStorageClientConfig()
 
-	clientOption, err := createClientOptionForGRPCClient(&sc)
+	clientOption, err := createClientOptionForGRPCClient(&sc, false)
```
vadlakondaswetha, line 559: should we add one test case which tests for bidiConfig=true scenario?

vadlakondaswetha (approved):

PR #2928, opened 2025-01-22
Skip test when testing with installed package

kislaykishore:
### Description
This is a temporary fix. We would like to get it working with installed package too but disabling it for now since metrics is currently not being worked upon.

b/391498868

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


ashmeenkaur (commented):
is this a temporary fix? why can't we run these tests with an installed package?

kislaykishore:
> is this a temporary fix? why can't we run these tests with an installed package?

Yes, it's a temporary fix. Logged b/391754854 for a permanent fix.

ashmeenkaur (approved):

PR #2929, opened 2025-01-22
Handle write edge case when first write is out of order write

ashmeenkaur:
### Description

Handle write edge case when first write is out of order write followed by write at offset 0.

Issue: when we get out of order write, we flush buffered write handler (to finalize whatever we have) and also set bwh to nil
First out of order write creates a source object of size 0, so even after bwh is cleared, it is recreated during the write call. Later when a write at offset 0 happens, it will go to bwh. Partial writes to bwh and temp file can cause errors.

Solution: If temp file is in use, don't re-create buffered write handler.

### Link to the internal issue
b/391545203

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -559,23 +559,13 @@ func (f *FileInode) Write(
 	data []byte,
 	offset int64) error {
 
-	isTempFileInUse, err := f.isTempFileInUse()
+	err := f.initBufferedWriteHandlerIfEligible(ctx)
 	if err != nil {
 		return err
 	}
 
-	// For empty GCS files also we will trigger bufferedWrites flow.
-	if f.src.Size == 0 && f.config.Write.EnableStreamingWrites && !isTempFileInUse {
-		err := f.ensureBufferedWriteHandler(ctx)
-		if err != nil {
-			return err
-		}
-	}
-
 	if f.bwh != nil {
-		err := f.writeUsingBufferedWrites(ctx, data, offset)
-		logger.Errorf("after buffered write handler = %v", f.bwh)
-		return err
+		return f.writeUsingBufferedWrites(ctx, data, offset)
 	}
 
 	return f.writeUsingTempFile(ctx, data, offset)
@@ -879,17 +869,9 @@ func (f *FileInode) Truncate(
 	ctx context.Context,
 	size int64) (err error) {
 
-	isTempFileInUse, err := f.isTempFileInUse()
+	err = f.initBufferedWriteHandlerIfEligible(ctx)
 	if err != nil {
 		return err
-	}
-
-	// For empty GCS files also, we will trigger bufferedWrites flow.
-	if f.src.Size == 0 && f.config.Write.EnableStreamingWrites && !isTempFileInUse {
-		err = f.ensureBufferedWriteHandler(ctx)
-		if err != nil {
-			return
-		}
 	}
 
 	if f.bwh != nil {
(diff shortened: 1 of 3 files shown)


internal/fs/inode/file.go
```diff
@@ -558,21 +558,41 @@ func (f *FileInode) Write(
...
+	if f.src.Size == 0 && f.config.Write.EnableStreamingWrites && !isTempFileInUse {
 		err := f.ensureBufferedWriteHandler(ctx)
 		if err != nil {
 			return err
 		}
 	}
 
-	if f.bwh != nil {
-		return f.writeUsingBufferedWrites(ctx, data, offset)
+	if f.bwh != nil && !isTempFileInUse {
```
vadlakondaswetha, line 575: why do we need isTempFileInUse check here. Do you any scenario where bwh != nil but we need to use disk?

ashmeenkaur: Was not required, thanks! ensured that we never create bwh when temp file is in use (truncate was the other such scenario).

internal/fs/inode/file.go
```diff
@@ -558,21 +558,41 @@ func (f *FileInode) Write(
...
 	offset int64) error {
+
+	isTempFileInUse, err := f.isTempFileInUse()
+	if err != nil {
+		return err
+	}
+
 	// For empty GCS files also we will trigger bufferedWrites flow.
-	if f.src.Size == 0 && f.config.Write.EnableStreamingWrites {
+	if f.src.Size == 0 && f.config.Write.EnableStreamingWrites && !isTempFileInUse {
```
meet2mky, line 568: Can we simplify like this?
`(f.src.Size == 0 && offset == 0 && f.config.Write.EnableStreamingWrites)` 
BWH should get initialized only when it's a new file or empty file and writes are sequential and they must start with offset == 0 to be sequential at start.

ashmeenkaur: For out of order writes, we might get offset 0 again after a non 0 offset. This will initialize the buffered write handler again in that scenario and will have the same problem of a some part of writes going to bwh and some part of writes going to temp file.

meet2mky: Got it.

internal/fs/inode/file.go
```diff
@@ -558,21 +558,41 @@ func (f *FileInode) Write(
 	ctx context.Context,
 	data []byte,
 	offset int64) error {
+
+	isTempFileInUse, err := f.isTempFileInUse()
+	if err != nil {
+		return err
```
vadlakondaswetha, line 564: Please move all this logic into common function. 561-574

internal/fs/inode/file.go
```diff
@@ -558,21 +558,41 @@ func (f *FileInode) Write(
...
 		err := f.ensureBufferedWriteHandler(ctx)
 		if err != nil {
 			return err
 		}
 	}
 
 	if f.bwh != nil {
-		return f.writeUsingBufferedWrites(ctx, data, offset)
+		err := f.writeUsingBufferedWrites(ctx, data, offset)
+		logger.Errorf("after buffered write handler = %v", f.bwh)
```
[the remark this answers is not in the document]
ashmeenkaur: Oops sorry, thanks for pointing

internal/fs/inode/file.go
```diff
@@ -573,6 +571,18 @@ func (f *FileInode) Write(
 	return f.writeUsingTempFile(ctx, data, offset)
 }
 
+func (f *FileInode) isTempFileInUse() (bool, error) {
+	if f.content == nil {
+		return false, nil
+	}
+	statRes, err := f.content.Stat()
```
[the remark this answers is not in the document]
ashmeenkaur: I was concerned about the possibility of leftover temp file after unlinking.  The size check was added to address that.  However, after reviewing the code and considering various scenarios, I haven't been able to identify a situation where `f.content` would be non-nil with 0 size. Removed this  redundant size check.  Thanks for pointing this out!

vadlakondaswetha (approved):

PR #2931, opened 2025-01-23
Stopping use of timer based cleanup logic for MRD

abhishek10004:
Making changes to stop using timer based cleanup function for MRD (Created b/391508479 to track restoring it).

Making this change against b/388177094

### DIFF
--- a/internal/gcsx/multi_range_downloader_wrapper.go
+++ b/internal/gcsx/multi_range_downloader_wrapper.go
@@ -32,7 +32,7 @@ import (
 
 // Timeout value which determines when the MultiRangeDownloader will be closed after
 // it's refcount reaches 0.
-// TODO b/391508479 : Reset it back to an appropriate value
+// TODO (b/391508479): Reset it back to an appropriate value & rename it to multiRangeDownloaderCleanupTimeout.
 const multiRangeDownloaderTimeout = 0
 
 func NewMultiRangeDownloaderWrapper(bucket gcs.Bucket, object *gcs.MinObject) MultiRangeDownloaderWrapper {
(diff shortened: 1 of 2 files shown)


internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -116,8 +117,9 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) cleanupMultiRangeDownloader() {
 	closeMRD := func(ctx context.Context) {
 		select {
 		case <-mrdWrapper.clock.After(multiRangeDownloaderTimeout):
-			mrdWrapper.mu.Lock()
-			defer mrdWrapper.mu.Unlock()
+			// TODO (b/391508479): Restore locks when closeMRD is called asynchronously.
```
vadlakondaswetha, line 120: how about we just do mrdWrapper.Wrapped.Close()
mrdWrapped.Wrapped = nil at line 109 
and comment out mrdWrapped.cleanupMultiRangeDownloader()

abhishek10004: but isn't that just removing the cleanup logic?

vadlakondaswetha: sorry for the confusion. I meant changing this 
mrdWrapper.cleanupMultiRangeDownloader() to

mrdWrapper.Wrapped.Close()
mrdWrapper.Wrapped=nil
// TODO: (b/) Add it when recreation of MRD is handled
//mrdWrapper.cleanupMultiRangeDownloader()


This way its easy to revert back when required.

vadlakondaswetha (approved):

PR #2932, opened 2025-01-23
Fixing reader growth logic & logging errors in MRD::Read

abhishek10004:
1. Changes to ensure reader size does not stop growing in random reads scenario.
2. Logging error in MRD::Read

b/390543202, b/391753515

### DIFF
--- a/internal/gcsx/multi_range_downloader_wrapper.go
+++ b/internal/gcsx/multi_range_downloader_wrapper.go
@@ -191,15 +191,15 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) Read(ctx context.Context, buf []b
 		}()
 
 		if e != nil && e != io.EOF {
-			e = fmt.Errorf("MultiRangeDownloaderWrapper::Read: Error in Add Call: %w", e)
+			e = fmt.Errorf("Error in Add Call: %w", e)
 		}
 	})
 
 	select {
 	case <-time.After(timeout):
-		err = fmt.Errorf("MultiRangeDownloaderWrapper::Read: Timeout")
+		err = fmt.Errorf("Timeout")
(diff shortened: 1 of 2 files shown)


gargnitingoogle:
@abhishek10004 could you share examples of how the before-change and after-change logs compare? I don't want to lose useful context from the logs.

abhishek10004:
> @abhishek10004 could you share examples of how the before-change and after-change logs compare? I don't want to lose useful context from the logs.

We are just adding an extra error log in MRD::Read function which was not there. Earlier, the error string was logged, along with time, in a trace log. Now, in case of an error, I've added an error log as well.

The errors that we were returning to the caller function will still remain same. There's no change there.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -209,6 +209,8 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) Read(ctx context.Context, buf []b
 	errDesc := "OK"
 	if err != nil {
 		errDesc = err.Error()
+		err = fmt.Errorf("MultiRangeDownloaderWrapper::Read: %w", err)
```
vadlakondaswetha, line 212: all these 3 lines can be clubbed into single line:

logger.Errorf("MultiRangeDownloaderWrapper::Read: %w", err")

vadlakondaswetha (approved):

internal/gcsx/multi_range_downloader_wrapper.go
abhishek10004: I had to keep these separate because:

1. I want to log error in the trace call below without adding the method prefix again.
2. The error that I return to the caller function should have the method name as the prefix for better debuggability.

PR #2933, opened 2025-01-24
Fix crash when the file name being statted is just "\n"

kislaykishore:
### Description
GCSFuse crashes when a stat call is made for a path where the file name is just "\n". Instead, return an ENOENT error.

b/391795883

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Added.

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -980,7 +980,10 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 	parent inode.DirInode,
 	childName string) (child inode.Inode, err error) {
 	// First check if the requested child is a localFileInode.
-	child = fs.lookUpLocalFileInode(parent, childName)
+	child, err = fs.lookUpLocalFileInode(parent, childName)
+	if err != nil {
+		return child, err
+	}
 	if child != nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


internal/fs/fs.go
```diff
@@ -980,7 +980,9 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 	parent inode.DirInode,
 	childName string) (child inode.Inode, err error) {
 	// First check if the requested child is a localFileInode.
-	child = fs.lookUpLocalFileInode(parent, childName)
+	if child, err = fs.lookUpLocalFileInode(parent, childName); err != nil {
+		return child, err
```
Tulsishah, line 984: nit: return nil, err ?

internal/fs/fs.go
```diff
@@ -1036,8 +1038,11 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 // LOCKS_EXCLUDED(parent)
 // UNLOCK_FUNCTION(fs.mu)
 // LOCK_FUNCTION(child)
-func (fs *fileSystem) lookUpLocalFileInode(parent inode.DirInode, childName string) (child inode.Inode) {
+func (fs *fileSystem) lookUpLocalFileInode(parent inode.DirInode, childName string) (child inode.Inode, err error) {
 	// Trim the suffix assigned to fix conflicting names.
+	if childName == inode.ConflictingFileNameSuffix {
+		return child, syscall.ENOENT
```
Tulsishah, line 1044: return nil, syscall.ENOENT

internal/fs/fs.go
```diff
@@ -980,7 +980,9 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 	parent inode.DirInode,
 	childName string) (child inode.Inode, err error) {
 	// First check if the requested child is a localFileInode.
-	child = fs.lookUpLocalFileInode(parent, childName)
+	if child, err = fs.lookUpLocalFileInode(parent, childName); err != nil {
+		return child, err
+	}
 	if child != nil {
```
vadlakondaswetha, line 986: should we just combine both the conditions into one line.
if child != nil | err != nil {
return child, err
}

vadlakondaswetha (approved):

internal/fs/fs.go
kislaykishore: I'd find that hard to read - since it captures both a success scenario as well as a failure scenario.

PR #2936, opened 2025-01-24
Fixing MinObject related issues in MRDWrapper

abhishek10004:
1. Ensure MRD's minObject is updated when inode's minObject is updated.

2. Ensure that we set MRD to nil if there were any errors while creating it. (We observed that NewMultiRangeDownloader returned a non nil value for MRD even in case of error)

b/391534052

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -160,7 +160,7 @@ func NewFileInode(
 		globalMaxWriteBlocksSem: globalMaxBlocksSem,
 	}
 	var err error
-	f.MRDWrapper, err = gcsx.NewMultiRangeDownloaderWrapper(bucket, &minObj)
+	f.MRDWrapper, err = gcsx.NewMultiRangeDownloaderWrapper(bucket, &f.src)
 	if err != nil {
 		logger.Errorf("NewFileInode: Error in creating MRDWrapper %v", err)
 	}
@@ -705,6 +705,7 @@ func (f *FileInode) SetMtime(
 			minObj = *minObjPtr
 		}
 		f.src = minObj
+		f.updateMRDWrapper()
 		return
 	}
 
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -861,7 +861,10 @@ func (f *FileInode) updateInodeStateAfterSync(minObj *gcs.MinObject) {
 // Updates the min object stored in MRDWrapper corresponding to the inode.
 // Should be called when minObject associated with inode is updated.
 func (f *FileInode) updateMRDWrapper() {
-	f.MRDWrapper.SetMinObject(&f.src)
+	err := f.MRDWrapper.SetMinObject(&f.src)
+	if err != nil {
+		logger.Errorf("FileInode::updateMRDWrapper Error in setting minObject %v", err)
+	}
 }
 
 // Truncate the file to the specified size.
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -629,7 +629,7 @@ func (f *FileInode) flushUsingBufferedWriteHandler() error {
 	// bwh can return a partially synced object along with an error so updating
 	// inode state before returning error.
 	f.updateInodeStateAfterSync(obj)
-	f.updateMRDWrapper(obj)
+	f.updateMRDWrapper()
 	if err != nil {
 		return fmt.Errorf("f.bwh.Flush(): %w", err)
 	}
@@ -808,7 +808,7 @@ func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
 	minObj := storageutil.ConvertObjToMinObject(newObj)
 	// If we wrote out a new object, we need to update our state.
 	f.updateInodeStateAfterSync(minObj)
-	f.updateMRDWrapper(minObj)
+	f.updateMRDWrapper()
 	return
 }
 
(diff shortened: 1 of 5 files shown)


internal/fs/inode/file.go
```diff
@@ -629,6 +629,7 @@ func (f *FileInode) flushUsingBufferedWriteHandler() error {
 	// bwh can return a partially synced object along with an error so updating
 	// inode state before returning error.
 	f.updateInodeStateAfterSync(obj)
+	f.updateMRDWrapper()
```
vadlakondaswetha, line 632: move this also to updateInodeState, so its clear what needs to be done after flush

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -56,6 +59,7 @@ type MultiRangeDownloaderWrapper struct {
 	Wrapped gcs.MultiRangeDownloader
 
 	// Bucket and object details for MultiRangeDownloader.
+	// Object should not be nil.
```
vadlakondaswetha, line 62: the comment is not clear. We are saying object shouldnt be nil but we are setting it to nil during init.

internal/fs/inode/file.go
abhishek10004: I deliberately did not put it in updateInodeState because of its name. But I understand the reasoning. What do you think about renaming updateInodeStateAfterSync to updateDependenciesAfterSync or updateStateAfterSync?

vadlakondaswetha: updateInodeStateAfterSync or updateStateAfterSync() is fine.
updateInodestateAfterSync -> This implies that updating dependencies after inode state is updated.

abhishek10004: Okay. Keeping it same, since it signifies updating dependencies as well.

internal/gcsx/multi_range_downloader_wrapper.go
abhishek10004: Since, MRDWrapper creation did not return an error originally, I had put a warning over there in case a nil minObject was passed.
I have now started returning an error from there to remove the ambiguity. 
I still do not want to panic code in inode creation if an error is returned (currently, it would never receive an error). Hence, just logged the error there.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -34,16 +34,19 @@ import (
...
 	return NewMultiRangeDownloaderWrapperWithClock(bucket, object, clock.RealClock{})
 }
 
-func NewMultiRangeDownloaderWrapperWithClock(bucket gcs.Bucket, object *gcs.MinObject, clock clock.Clock) MultiRangeDownloaderWrapper {
+func NewMultiRangeDownloaderWrapperWithClock(bucket gcs.Bucket, object *gcs.MinObject, clock clock.Clock) (MultiRangeDownloaderWrapper, error) {
+	if object == nil {
+		return MultiRangeDownloaderWrapper{}, fmt.Errorf("NewMultiRangeDownloaderWrapperWithClock: Missing MinObject")
+	}
 	return MultiRangeDownloaderWrapper{
 		clock:  clock,
```
vadlakondaswetha, line 46: object is getting initialized via setMinObject. When are bucket and clock variables getting initialized?
The earlier implementation was fine except the comment. you can just update the comment saying object will be nil first for local fileInodes and should be initialized before mrd.Read is called.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -136,6 +149,10 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) cleanupMultiRangeDownloader() {
 
 // Ensures that MultiRangeDownloader exists, creating it if it does not exist.
 func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader() (err error) {
+	if mrdWrapper.object == nil || mrdWrapper.bucket == nil {
```
vadlakondaswetha, line 152: Please add new test cases for all new if clauses added.

internal/gcsx/random_reader_stretchr_test.go
```diff
@@ -642,12 +642,14 @@ func (t *RandomReaderStretchrTest) Test_ReadAt_MRDRead() {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func() {
+			var err error
```
vadlakondaswetha, line 645: you dont need this seperate declaration since fakeMRDWrapper below is not a instance variable on the struct.

internal/gcsx/multi_range_downloader_wrapper.go
abhishek10004: Discussed offline

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -69,6 +75,19 @@ type MultiRangeDownloaderWrapper struct {
...
+// Sets the gcs.MinObject stored in the wrapper to passed value, only if it's non nil.
+func (mrdWrapper *MultiRangeDownloaderWrapper) SetMinObject(minObj *gcs.MinObject) error {
+	if minObj == nil {
+		return fmt.Errorf("MultiRangeDownloaderWrapper::SetMinObject: Missing MinObject")
+	}
+	mrdWrapper.object = minObj
+	return nil
+}
+
+func (mrdWrapper *MultiRangeDownloaderWrapper) GetMinObject() *gcs.MinObject {
```
vadlakondaswetha, line 87: add a comment that this method is only for unit tests.

vadlakondaswetha (approved):

vadlakondaswetha (approved):

PR #2943, opened 2025-01-27
[Readability fix] Be explicit that nil child is being returned.

kislaykishore:
b/391795883

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -980,7 +980,8 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 	parent inode.DirInode,
 	childName string) (child inode.Inode, err error) {
 	// First check if the requested child is a localFileInode.
-	if child, err = fs.lookUpLocalFileInode(parent, childName); err != nil {
+	child, err = fs.lookUpLocalFileInode(parent, childName)
+	if err != nil {
 		return nil, err
 	}
 	if child != nil {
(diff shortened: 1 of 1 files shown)


Tulsishah (approved):

internal/fs/fs.go
```diff
@@ -981,7 +981,7 @@ func (fs *fileSystem) lookUpOrCreateChildInode(
 	childName string) (child inode.Inode, err error) {
 	// First check if the requested child is a localFileInode.
 	if child, err = fs.lookUpLocalFileInode(parent, childName); err != nil {
```
charith87, line 983: Can we change this line to initialize above and then use it below since child is used even after the if block
go/go-style/decisions#indent-error-flow 
go/gotip/episodes/1
e.g.
child, err = fs.lookUpLocalFileInode(parent, childName); 
if err != nil {

kislaykishore: I don't think it violates those principles -
go/go-style/decisions#indent-error-flow - terminal condition is being handled before proceeding with the rest of the code and there is no else block.
go/gotip/episodes/1 - The happy path is already on the least indented level.

charith87: i saw `child` being used below and returning that am assuming is a happy path.
the lookupLocalFileInode if its not indented its better i assume since code below also is using .
We can do it later also. Not blocking this review. Since this was not directly in this PR.

charith87 (dismissed):

charith87 (approved):

PR #2948, opened 2025-01-28
Finalize object for any error in buffered writes flow

ashmeenkaur:
### Description
We were finalizing only for upload failure and out of order writes but it makes sense to finalize the object in case of any failure.

b/392834488

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -57,9 +57,9 @@ type BufferedWriteHandler interface {
 	Unlink()
 }
 
-// BufferedWriteHandlerImpl is responsible for filling up the buffers with the data
+// bufferedWriteHandlerImpl is responsible for filling up the buffers with the data
 // as it receives and handing over to uploadHandler which uploads to GCS.
-type BufferedWriteHandlerImpl struct {
+type bufferedWriteHandlerImpl struct {
 	current       block.Block
 	blockPool     *block.BlockPool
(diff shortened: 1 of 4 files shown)


internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -29,9 +29,37 @@ import (
...
+	Destroy() error
+
+	// Unlink cancels the ongoing upload and free up the buffers.
+	Unlink()
+}
+
+// BufferedWriteHandlerImpl is responsible for filling up the buffers with the data
 // as it receives and handing over to uploadHandler which uploads to GCS.
-type BufferedWriteHandler struct {
+type BufferedWriteHandlerImpl struct {
```
vadlakondaswetha, line 62: This need not be public anymore, if we are exposing interface.

internal/fs/inode/file_streaming_writes_test.go
```diff
@@ -588,3 +590,79 @@ func (t *FileStreamingWritesTest) TestDeRegisterFileHandle() {
 		})
 	}
 }
+
+// TestBufferedWriteHandler is a test double for BufferedWriteHandler.
+type TestBufferedWriteHandler struct {
+	WriteFunc func(data []byte, offset int64) error
```
vadlakondaswetha, line 596: should we call it FakeBufferedWriteHandler?

vadlakondaswetha (approved):

PR #2949, opened 2025-01-29
Add streaming writes config to useragent

ashmeenkaur:
### Description
To track usage of streaming writes config, added bitset for streaming writes to user agent. This is a temporary change until we find a better way to track usage.

NA

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -113,7 +113,11 @@ func getConfigForUserAgent(mountConfig *cfg.Config) string {
 	if cfg.IsParallelDownloadsEnabled(mountConfig) {
 		isParallelDownloadsEnabled = "1"
 	}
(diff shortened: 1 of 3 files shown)


cmd/legacy_main.go
```diff
@@ -113,7 +113,11 @@ func getConfigForUserAgent(mountConfig *cfg.Config) string {
 	if cfg.IsParallelDownloadsEnabled(mountConfig) {
 		isParallelDownloadsEnabled = "1"
 	}
-	return fmt.Sprintf("%s:%s:%s", isFileCacheEnabled, isFileCacheForRangeReadEnabled, isParallelDownloadsEnabled)
+	areStreamingWritesEnabled := "0"
+	if cfg.AreStreamingWritesEnabled(mountConfig) {
```
vadlakondaswetha, line 117: can we check the config directly here instead of seperate method
c.Write.EnableStreamingWrites?
why to write a seperate method for checking one condition.

vadlakondaswetha (approved):

PR #2958, opened 2025-01-31
Doing flush for pending writes on rename file 

vadlakondaswetha:
### Description
Doing this only when streaming writes is enabled to be safe. Will do it later for non-streaming writes.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2024,7 +2024,8 @@ func (fs *fileSystem) Rename(
 	if err != nil {
 		return err
 	}
-	if localChild != nil {
+	// For streaming writes, we will finalize localChild and do rename.
+	if localChild != nil && !fs.newConfig.Write.EnableStreamingWrites {
 		fs.unlockAndDecrementLookupCount(localChild, 1)
 		return fmt.Errorf("cannot rename open file %q: %w", op.OldName, syscall.ENOTSUP)
 	}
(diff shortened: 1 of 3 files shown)


internal/fs/fs.go
```diff
@@ -2052,10 +2052,57 @@ func (fs *fileSystem) Rename(
 		}
 		return fs.renameNonHierarchicalDir(ctx, oldParent, op.OldName, newParent, op.NewName)
 	}
+
```
ashmeenkaur, line 2055: shall we remove the condition which stops local files to be renamed when streaming writes are enabled?

internal/fs/fs.go
```diff
@@ -2052,10 +2052,57 @@ func (fs *fileSystem) Rename(
...
+	fs.mu.Lock()
+	existingInode, ok := fs.generationBackedInodes[child.FullName]
+	fs.mu.Unlock()
+	// If this is not an existing file, then nothing to do.
+	// We shouldn't hit scenario ideally since we checked for local file and
+	// explicit-dirs (flat bucket) before reaching this step.
+	if !ok {
+		logger.Warnf("Encountered a non-fileInode in rename file path %d", existingInode.ID())
+		return
+	}
```
ashmeenkaur, line 2092: Shall we instead remove the condition which stops local files to be renamed when streaming writes are enabled? (line 2022-2030)

internal/fs/fs.go
```diff
@@ -2052,10 +2052,57 @@ func (fs *fileSystem) Rename(
...
+
+	fileInode, isFileInode := existingInode.(*inode.FileInode)
```
ashmeenkaur, line 2094: Nit: shall we use fileInodeOrDie?

vadlakondaswetha: Converted to error message. avoiding panic.

internal/fs/fs.go
vadlakondaswetha: yep. doing finalize for local files incase of streaming writes.

ashmeenkaur (approved):

PR #2970, opened 2025-02-05
Recreate MRD in case of any error

abhishek10004:
Changes to ensure MRD is recreated if any error is thrown from the Add call as well. Currently, after successful  recreation of MRD, we are not recreating it.

Using a new variable to determine creation of MRD (instead of setting it to nil) to simplify the code (i.e. to prevent an additional mutex lock before calling add)

Performance testing: Tested random read with 50 threads for 1M/128K block size and there is no regression. Also tested the parallel read scenario (50 threads randomly reading from the same file) and there's no regression in this scenario as well.

b/394500981

### DIFF
--- a/internal/gcsx/multi_range_downloader_wrapper.go
+++ b/internal/gcsx/multi_range_downloader_wrapper.go
@@ -45,9 +45,10 @@ func NewMultiRangeDownloaderWrapperWithClock(bucket gcs.Bucket, object *gcs.MinO
 	// In case of a local inode, MRDWrapper would be created with an empty minObject (i.e. with a minObject without any information)
 	// and when the object is actually created, MRDWrapper would be updated using SetMinObject method.
 	return MultiRangeDownloaderWrapper{
-		clock:  clock,
-		bucket: bucket,
-		object: object,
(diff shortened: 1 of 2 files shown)


internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -220,6 +232,9 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) Read(ctx context.Context, buf []b
 
 		if e != nil && e != io.EOF {
 			e = fmt.Errorf("Error in Add Call: %w", e)
+			mrdWrapper.muCreateMRD.Lock()
```
vadlakondaswetha, line 235: Couple of things:
1. We shouldn't create mrd on every error. Should be created only when mrd is closed by go-sdk.
2. If there are x callbacks failed on the mrd, then we end up creating mrd x number of times even with the muCreateMRD lock. This is the thing which we wanted to avoid in the first place and didn't take this route.

Lets work with go-sdk in coming up with the right solution.

abhishek10004:
Closing this as it's obsolete now.

PR #2971, opened 2025-02-05
Maintain readType for random reader

abhishek10004:
We collect readType as part of our read metrics and hence need to ensure that proper readType value is being maintained and shared for metrics collection.

b/394500505

### DIFF
--- a/internal/gcsx/random_reader.go
+++ b/internal/gcsx/random_reader.go
@@ -541,7 +541,6 @@ func (rr *randomReader) getReadInfo(
 	// optimise for random reads. Random reads will read data in chunks of
 	// (average read size in bytes rounded up to the next MB).
 	end = int64(rr.object.Size)
-	rr.readType = util.Sequential
 	if rr.seeks >= minSeeksForRandom {
 		rr.readType = util.Random
 		averageReadBytes := rr.totalReadBytes / rr.seeks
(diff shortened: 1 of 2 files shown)


internal/gcsx/random_reader.go
```diff
@@ -539,9 +541,9 @@ func (rr *randomReader) getReadInfo(
 	// optimise for random reads. Random reads will read data in chunks of
 	// (average read size in bytes rounded up to the next MB).
 	end = int64(rr.object.Size)
-	readType = util.Sequential
+	rr.readType = util.Sequential
```
vadlakondaswetha, line 544: do we need this assignment here?

internal/gcsx/random_reader.go
```diff
@@ -150,6 +151,9 @@ type randomReader struct {
 	seeks          uint64
 	totalReadBytes uint64
 
+	// ReadType of the reader. Will be sequential by default.
+	readType string
```
vadlakondaswetha, line 155: What about unit tests to check if readType is preserved across calls to ReadAt()

internal/gcsx/random_reader.go
abhishek10004: Not needed. Removed.

internal/gcsx/random_reader.go
abhishek10004: Added UT for validating readType.

vadlakondaswetha (approved):

PR #2974, opened 2025-02-06
Use grpc client always for zonal buckets in e2e tests

gargnitingoogle:
### Description
* Always use client-type grpc for zonal bucket in e2e tests
* Always set append flag in the storage writer objects for zonal bucket in e2e tests

[b/393050972](http://b/393050972) - Create and pass test flag for zonal bucket

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Enabled through presubmit

### DIFF
--- a/tools/integration_tests/readonly/readonly_test.go
+++ b/tools/integration_tests/readonly/readonly_test.go
@@ -17,6 +17,7 @@ package readonly_test
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"os"
 	"path"
@@ -55,7 +56,7 @@ var (
 	cacheDir      string
 )
 
-func createTestDataForReadOnlyTests(ctx context.Context, storageClient *storage.Client) {
+func createTestDataForReadOnlyTests(ctx context.Context, storageClient *storage.Client) error {
 	// Define the text to write and the files to create
 	files := []struct {
 		fileContent string
@@ -74,17 +75,23 @@ func createTestDataForReadOnlyTests(ctx context.Context, storageClient *storage.
 		filePath := path.Join(dirPath, file.filePath)
 		// Create a storage writer for the destination object
 		object := bucketHandle.Object(filePath)
-		writer := client.NewWriterExt(ctx, object, storageClient)
+		writer, err := client.NewWriterExt(ctx, object, storageClient)
+		if err != nil {
+			return fmt.Errorf("Error opening writer for object %s: %w\n", file.filePath, err)
+		}
 
 		// Write the text to the object
-		if _, err := writer.Write([]byte(file.fileContent + "\n")); err != nil {
+		if _, err = writer.Write([]byte(file.fileContent + "\n")); err != nil {
 			log.Printf("Error writing to object %s: %v\n", file.filePath, err)
 		}
-		err := writer.Close()
+		err = writer.Close()
 		if err != nil {
 			log.Printf("Error in closing writer: %v", err)
+			return err
 		}
 	}
+
+	return nil
 }
 
 func checkErrorForObjectNotExist(err error, t *testing.T) {
@@ -139,7 +146,10 @@ func TestMain(m *testing.M) {
 	}
 
 	// Create test data.
-	createTestDataForReadOnlyTests(ctx, storageClient)
+	if err := createTestDataForReadOnlyTests(ctx, storageClient); err != nil {
+		log.Printf("Failed creating test data for readonly tests: %v", err)
+		os.Exit(1)
+	}
 
 	// Run tests for mountedDirectory only if --mountedDirectory flag is set.
 	setup.RunTestsForMountedDirectoryFlag(m)
--- a/tools/integration_tests/readonly/readonly_test.go
+++ b/tools/integration_tests/readonly/readonly_test.go
@@ -75,7 +75,7 @@ func createTestDataForReadOnlyTests(ctx context.Context, storageClient *storage.
 		filePath := path.Join(dirPath, file.filePath)
 		// Create a storage writer for the destination object
 		object := bucketHandle.Object(filePath)
-		writer, err := client.NewWriterExt(ctx, object, storageClient)
+		writer, err := client.NewWriterWithSupportForZB(ctx, object, storageClient)
 		if err != nil {
 			return fmt.Errorf("Error opening writer for object %s: %w\n", file.filePath, err)
 		}
--- a/tools/integration_tests/readonly/readonly_test.go
+++ b/tools/integration_tests/readonly/readonly_test.go
@@ -74,7 +74,7 @@ func createTestDataForReadOnlyTests(ctx context.Context, storageClient *storage.
 		filePath := path.Join(dirPath, file.filePath)
 		// Create a storage writer for the destination object
 		object := bucketHandle.Object(filePath)
-		writer := object.NewWriter(ctx)
+		writer := client.NewWriterExt(ctx, object, storageClient)
 
 		// Write the text to the object
 		if _, err := writer.Write([]byte(file.fileContent + "\n")); err != nil {
--- a/tools/integration_tests/readonly/readonly_test.go
+++ b/tools/integration_tests/readonly/readonly_test.go
@@ -75,7 +75,7 @@ func createTestDataForReadOnlyTests(ctx context.Context, storageClient *storage.
 		filePath := path.Join(dirPath, file.filePath)
 		// Create a storage writer for the destination object
 		object := bucketHandle.Object(filePath)
-		writer, err := client.NewWriterWithSupportForZB(ctx, object, storageClient)
+		writer, err := client.NewWriter(ctx, object, storageClient)
 		if err != nil {
 			return fmt.Errorf("Error opening writer for object %s: %w\n", file.filePath, err)
 		}
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -130,10 +130,12 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
 
 func NewWriterExt(ctx context.Context, o *storage.ObjectHandle, client *storage.Client) *storage.Writer {
 	wc := o.NewWriter(ctx)
-	if setup.IsZonalBucketRun() {
-		attrs, _ := client.Bucket(o.BucketName()).Attrs(ctx)
-		if attrs.StorageClass == "RAPID" {
+	attrs, _ := client.Bucket(o.BucketName()).Attrs(ctx)
+	if attrs.StorageClass == "RAPID" {
+		if setup.IsZonalBucketRun() {
 			wc.Append = true
+		} else {
+			panic(fmt.Sprintf("Found zonal bucket %q in non-zonal e2e test run (--zonal=false)", o.BucketName()))
 		}
 	}
 	return wc
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -128,6 +128,17 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
 	return string(content), nil
 }
 
+func NewWriterExt(ctx context.Context, o *storage.ObjectHandle, client *storage.Client) *storage.Writer {
+	wc := o.NewWriter(ctx)
+	if setup.IsZonalBucketRun() {
+		attrs, _ := client.Bucket(o.BucketName()).Attrs(ctx)
+		if attrs.StorageClass == "RAPID" {
+			wc.Append = true
+		}
+	}
+	return wc
+}
+
 func WriteToObject(ctx context.Context, client *storage.Client, object, content string, precondition storage.Conditions) error {
 	bucket, object := setup.GetBucketAndObjectBasedOnTypeOfMount(object)
 
@@ -137,13 +148,7 @@ func WriteToObject(ctx context.Context, client *storage.Client, object, content 
 	}
 
 	// Upload an object with storage.Writer.
-	wc := o.NewWriter(ctx)
-	if setup.IsZonalBucketRun() {
-		attrs, _ := client.Bucket(setup.TestBucket()).Attrs(ctx)
-		if attrs.StorageClass == "RAPID" {
-			wc.Append = true
-		}
-	}
+	wc := NewWriterExt(ctx, o, client)
 	if _, err := io.WriteString(wc, content); err != nil {
 		return fmt.Errorf("io.WriteSTring: %w", err)
 	}
@@ -290,7 +295,7 @@ func StatObject(ctx context.Context, client *storage.Client, object string) (*st
 func UploadGcsObject(ctx context.Context, client *storage.Client, localPath, bucketName, objectName string, uploadGzipEncoded bool) error {
 	// Create a writer to upload the object.
 	obj := client.Bucket(bucketName).Object(objectName)
-	w := obj.NewWriter(ctx)
+	w := NewWriterExt(ctx, obj, client)
 	defer func() {
 		if err := w.Close(); err != nil {
 			log.Printf("Failed to close GCS object gs://%s/%s: %v", bucketName, objectName, err)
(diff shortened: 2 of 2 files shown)


Tulsishah (commented):
We also need to add condition for append [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/readonly/readonly_test.go#L77)

tools/integration_tests/util/client/storage_client.go
```diff
@@ -133,6 +138,9 @@ func WriteToObject(ctx context.Context, client *storage.Client, object, content
 
 	// Upload an object with storage.Writer.
 	wc := o.NewWriter(ctx)
+	if setup.IsZonalBucketRun() {
+		wc.Append = true // setting true for zonal buckets
```
Tulsishah, line 142: This will cause an issue for dynamic mounting.  In dynamic mounting, we first run tests on a provided bucket (--testBucket) and then on another bucket we create during the test (a "flat" bucket). Here, if we run tests for a zonal configuration, the value will be true for the flat bucket, causing the tests to fail.
We have to attach this type of fix.

```
 attrs, _ := client.Bucket(setup.TestBucket()).Attrs(ctx)
+       if attrs.StorageClass == "RAPID" {
+               wc.Append = true
+       }
```

We have to do the same above as well..

gargnitingoogle: thanks. done. update all the places.

gargnitingoogle:
The e2e tests failed with the latest commit with failures in benchmarking tests, with error of the form., `RenameFile took more time (XXX msec) than expected (YYY msec)` . So I have given it a re-run.

gargnitingoogle:
> We also need to add condition for append [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/readonly/readonly_test.go#L77)

Handled this place and one more place that I found.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -123,6 +128,17 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
 	return string(content), nil
 }
 
+func NewWriterExt(ctx context.Context, o *storage.ObjectHandle, client *storage.Client) *storage.Writer {
+	wc := o.NewWriter(ctx)
+	if setup.IsZonalBucketRun() {
```
abhishek10004, line 133: This would not work as we have some tests for dynamic mounting as well wherein we create buckets dynamically. These buckets are standard buckets and hence appendable writer would not work for them.

gargnitingoogle: addressed as discussed offline.

gargnitingoogle:
> The e2e tests failed with the latest commit with failures in benchmarking tests, with error of the form., `RenameFile took more time (XXX msec) than expected (YYY msec)` . So I have given it a re-run.

Even on [re-run](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/d87f7535-4b46-43ba-bb92-1f648436daa3/log) the same failure repeated, so this is definitely a test environment or bug in gcsfuse library, so this needs to be debugged.

gargnitingoogle:
> > The e2e tests failed with the latest commit with failures in benchmarking tests, with error of the form., `RenameFile took more time (XXX msec) than expected (YYY msec)` . So I have given it a re-run.
> 
> Even on [re-run](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/d87f7535-4b46-43ba-bb92-1f648436daa3/log) the same failure repeated, so this is definitely a test environment or bug in gcsfuse library, so this needs to be debugged.

I [re-ran](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/02c2aeaf-d690-4e2e-8cfe-0cd9d4ae26f9/details) e2e tests once more to test with the fix made for dynamic bucket, and this time benchmark tests failed again along with `TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing` and `TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing`. Full error logs for these two new failures:

```
=== RUN   TestReadOnlyTest
=== RUN   TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing
    failure_during_file_sync_test.go:60: Expected error: permission denied, Got Error: <nil>
    failure_during_file_sync_test.go:53: Expected /tmp/gcsfuse_readwrite_test_3550593935/mnt/ReadOnlyCredsTest directory to be empty: [- fileName.txt]
=== RUN   TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing
    failure_during_file_sync_test.go:60: Expected error: permission denied, Got Error: <nil>
    failure_during_file_sync_test.go:53: Expected /tmp/gcsfuse_readwrite_test_3550593935/mnt/ReadOnlyCredsTest directory to be empty: [- fileName.txt]
--- FAIL: TestReadOnlyTest (0.89s)
    --- FAIL: TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing (0.53s)
    --- FAIL: TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing (0.36s)
FAIL
```

gargnitingoogle:
> > > The e2e tests failed with the latest commit with failures in benchmarking tests, with error of the form., `RenameFile took more time (XXX msec) than expected (YYY msec)` . So I have given it a re-run.
> > 
> > 
> > Even on [re-run](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/d87f7535-4b46-43ba-bb92-1f648436daa3/log) the same failure repeated, so this is definitely a test environment or bug in gcsfuse library, so this needs to be debugged.
> 
> I [re-ran](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/02c2aeaf-d690-4e2e-8cfe-0cd9d4ae26f9/details) e2e tests once more to test with the fix made for dynamic bucket, and this time benchmark tests failed again along with `TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing` and `TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing`. Full error logs for these two new failures:
> 
> ```
> === RUN   TestReadOnlyTest
> === RUN   TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing
>     failure_during_file_sync_test.go:60: Expected error: permission denied, Got Error: <nil>
>     failure_during_file_sync_test.go:53: Expected /tmp/gcsfuse_readwrite_test_3550593935/mnt/ReadOnlyCredsTest directory to be empty: [- fileName.txt]
> === RUN   TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing
>     failure_during_file_sync_test.go:60: Expected error: permission denied, Got Error: <nil>
>     failure_during_file_sync_test.go:53: Expected /tmp/gcsfuse_readwrite_test_3550593935/mnt/ReadOnlyCredsTest directory to be empty: [- fileName.txt]
> --- FAIL: TestReadOnlyTest (0.89s)
>     --- FAIL: TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing (0.53s)
>     --- FAIL: TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing (0.36s)
> FAIL
> ```

I and @abhishek10004 discussed the failure in benchmark-tests. It's possible that the latency has gone up because the presubmit run machine is now being created in us-west4 as opposed to us-central1 where it used to be created earlier, which is a recent change. So, we might have to increase the duration for passing criteria for the benchmark tests.

gargnitingoogle:
> > > > The e2e tests failed with the latest commit with failures in benchmarking tests, with error of the form., `RenameFile took more time (XXX msec) than expected (YYY msec)` . So I have given it a re-run.
> > > 
> > > 
> > > Even on [re-run](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/d87f7535-4b46-43ba-bb92-1f648436daa3/log) the same failure repeated, so this is definitely a test environment or bug in gcsfuse library, so this needs to be debugged.
> > 
> > 
> > I [re-ran](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/02c2aeaf-d690-4e2e-8cfe-0cd9d4ae26f9/details) e2e tests once more to test with the fix made for dynamic bucket, and this time benchmark tests failed again along with `TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing` and `TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing`. Full error logs for these two new failures:
> > ```
> > === RUN   TestReadOnlyTest
> > === RUN   TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing
> >     failure_during_file_sync_test.go:60: Expected error: permission denied, Got Error: <nil>
> >     failure_during_file_sync_test.go:53: Expected /tmp/gcsfuse_readwrite_test_3550593935/mnt/ReadOnlyCredsTest directory to be empty: [- fileName.txt]
> > === RUN   TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing
> >     failure_during_file_sync_test.go:60: Expected error: permission denied, Got Error: <nil>
> >     failure_during_file_sync_test.go:53: Expected /tmp/gcsfuse_readwrite_test_3550593935/mnt/ReadOnlyCredsTest directory to be empty: [- fileName.txt]
> > --- FAIL: TestReadOnlyTest (0.89s)
> >     --- FAIL: TestReadOnlyTest/TestEmptyCreateFileFails_FailedFileNotInListing (0.53s)
> >     --- FAIL: TestReadOnlyTest/TestNonEmptyCreateFileFails_FailedFileNotInListing (0.36s)
> > FAIL
> > ```
> 
> I and @abhishek10004 discussed the failure in benchmark-tests. It's possible that the latency has gone up because the presubmit run machine is now being created in us-west4 as opposed to us-central1 where it used to be created earlier, which is a recent change. So, we might have to increase the duration for passing criteria for the benchmark tests.

The tests have now passed but I had to increase the timeout values for stat and rename tests in the benchmarking package.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -123,6 +128,17 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
 	return string(content), nil
 }
 
+func NewWriterExt(ctx context.Context, o *storage.ObjectHandle, client *storage.Client) *storage.Writer {
```
abhishek10004, line 131: IMO we should log an error when we get a zonal bucket here when the zonal flag is not set. Will help us in catching incorrect usage.

abhishek10004: P.S. I was looking for something similar to what you've done at https://github.com/GoogleCloudPlatform/gcsfuse/pull/2979/files#diff-6f7c497fa7d6409c46108caba011760f35575256bc2e3169948a2d686f780c5d.
But instead of panic, we can log an error. Let me know what you think.

Also, can we think of a better name for this function? How about NewConfiguredWriter?

tools/integration_tests/benchmarking/benchmark_rename_test.go
```diff
@@ -31,7 +31,7 @@ type benchmarkRenameTest struct {
 }
 
 const (
-	expectedRenameLatency time.Duration = 700 * time.Millisecond
+	expectedRenameLatency time.Duration = 1200 * time.Millisecond
```
abhishek10004, line 34: We have a bug open for this - b/383948519. Please check if someone has looked into it

tools/integration_tests/util/client/storage_client.go
```diff
@@ -123,6 +128,17 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
...
+	wc := o.NewWriter(ctx)
+	if setup.IsZonalBucketRun() {
+		attrs, _ := client.Bucket(o.BucketName()).Attrs(ctx)
```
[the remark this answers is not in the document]
gargnitingoogle: done

tools/integration_tests/util/client/storage_client.go
gargnitingoogle: > I was looking for something similar to what you've done at

yes, that is what I meant by my next PR.

> But instead of panic, we can log an error

The reason I kept a panic instead of the returning of returning error, was to keep the same signature as that of NewWriter, which also doesn't return error. Anyway, I have changed it to returning error now.

tools/integration_tests/benchmarking/benchmark_rename_test.go
gargnitingoogle: rebased on top of #2922 which is the fix for b/383948519 .

tools/integration_tests/util/client/storage_client.go
```diff
@@ -123,6 +128,23 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
...
 }
 
+func NewWriterWithSupportForZB(ctx context.Context, o *storage.ObjectHandle, client *storage.Client) (wc *storage.Writer, err error) {
+	var attrs *storage.BucketAttrs
+	wc = o.NewWriter(ctx)
+	attrs, err = client.Bucket(o.BucketName()).Attrs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get attributes for bucket %q: %w", o.BucketName(), err)
+	}
+	if attrs.StorageClass == "RAPID" {
```
vadlakondaswetha, line 138: why can't we make this method generic. If storageClass is RAPID set wc.append=true, otherwise dont set it.

tools/integration_tests/readonly/readonly_test.go
```diff
@@ -74,17 +75,23 @@ func createTestDataForReadOnlyTests(ctx context.Context, storageClient *storage.
 		filePath := path.Join(dirPath, file.filePath)
 		// Create a storage writer for the destination object
 		object := bucketHandle.Object(filePath)
-		writer := object.NewWriter(ctx)
+		writer, err := client.NewWriterWithSupportForZB(ctx, object, storageClient)
```
vadlakondaswetha, line 78: Is this method used only for ZB? IMO we should create generic methods unless its too complicated.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -123,6 +128,23 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
...
+		if setup.IsZonalBucketRun() {
```
Tulsishah, line 139: I think this if-else condition is unnecessary.  If storage class is rapid than we should believe it as zonal only.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -123,6 +128,23 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
 	return string(content), nil
 }
 
+func NewWriterWithSupportForZB(ctx context.Context, o *storage.ObjectHandle, client *storage.Client) (wc *storage.Writer, err error) {
```
Tulsishah, line 131: Instead of this we can create method to add "wc.append=true" on rapid storage class and not create full writer.

gargnitingoogle: If a function just does `wc.append=true`, then we might as well just do `wc.append=true` in place of that function. The goal of this wrapper over NewWriter is to hide and de-dup the if-else and the error-handling that is needed for zonal buckets whenever NewWriter is invoked.

tools/integration_tests/util/client/storage_client.go
gargnitingoogle: If the name `NewWriterWithSupportForZB` is confusing, please suggest a different name.

gargnitingoogle: @vadlakondaswetha @Tulsishah @abhishek10004 ^

Tulsishah: I think `CreateWriter` or `NewWriter` should be sufficient. We can add comment about zonal where we are adding append=true.

gargnitingoogle: done. Added a few comments also for clarification.

PR #2978, opened 2025-02-10
Fix performance degradation due to metrics

kislaykishore:
### Description
The gcs/read_bytes_count metric is single-handedly responsible for perf degradation due when metrics are enabled. With this change, we are now using an atomic Int64 which efficiently computes the metric. The Int64Counter that uses OTel for metric computation has now been replaced with Int64ObservableCounter that uses the aforementioned atomic Int64 to report the same metric.

Perf:
| Master |  0.25MiB   | 528.52MiB/s  | 1.32MiB/s  |  74.95MiB/s  |  1.15MiB/s   |
|   PR   |  0.25MiB   | 547.94MiB/s  | 1.34MiB/s  |  78.84MiB/s  |  1.31MiB/s   |
| Master | 48.828MiB  | 3587.54MiB/s | 76.99MiB/s | 1485.06MiB/s |  84.21MiB/s  |
|   PR   | 48.828MiB  | 3659.14MiB/s | 86.45MiB/s | 1483.11MiB/s |  84.15MiB/s  |
| Master | 976.562MiB | 3635.5MiB/s  | 77.45MiB/s | 760.01MiB/s  |  29.29MiB/s  |
|   PR   | 976.562MiB | 3559.9MiB/s  | 78.1MiB/s  | 921.71MiB/s  |  28.44MiB/s  |

b/386226641

### Testing details
1. Manual - Verified that the metric is being correctly computed. Also verified that the performance is the same with and without metrics being enabled.
2. Unit tests - NA - the functionality of the module is not changing - only the internal implementation changes.
3. Integration tests - NA - the functionality is not changing - only the internal implementation changes.


common/oc_metrics.go
```diff
@@ -91,8 +91,8 @@ func attrsToAddOption(attrs []MetricAttr) []metric.AddOption {
 	return otelOptions
 }
 
-func (o *ocMetrics) GCSReadBytesCount(ctx context.Context, inc int64, attrs []MetricAttr) {
-	recordOCMetric(ctx, o.gcsReadBytesCount, inc, attrs, "GCS read bytes count")
+func (o *ocMetrics) GCSReadBytesCount(ctx context.Context, inc int64) {
```
ashmeenkaur, line 94: Shall we keep the attrs parameter? Although we are not using it, I think it is good to keep that consistency for all ocMetrics methods.

ashmeenkaur: Noticed that we are using WithAttributes instead of WithAttributeSet. The function documentation suggests to use WithAttributeSet for performance sensitive code. Have we tried using WithAttributeSet too?

Ref: 
- Otel reference: https://github.com/open-telemetry/opentelemetry-go/blob/dfc43f0d61524042665d6c5768bf6ff8cc006d46/metric/instrument.go#L345
- GCSFuse code: https://github.com/GoogleCloudPlatform/gcsfuse/blob/078d892227c588f48efdfb79d8653a83932ef6b2/common/oc_metrics.go#L89
https://github.com/GoogleCloudPlatform/gcsfuse/blob/078d892227c588f48efdfb79d8653a83932ef6b2/common/oc_metrics.go#L81

IMO this will be a cleaner solution if it works. Maybe this is faster because we removed WithAttributeSet?

common/otel_metrics.go
```diff
@@ -46,8 +47,8 @@ type otelMetrics struct {
 	fileCacheReadLatency    metric.Float64Histogram
 }
 
-func (o *otelMetrics) GCSReadBytesCount(ctx context.Context, inc int64, attrs []MetricAttr) {
-	o.gcsReadBytesCount.Add(ctx, inc, attrsToAddOption(attrs)...)
+func (o *otelMetrics) GCSReadBytesCount(_ context.Context, inc int64) {
+	o.gcsReadBytesCountInt.Add(inc)
```
ashmeenkaur, line 51: Noticed that we are using WithAttributes instead of WithAttributeSet. The function documentation suggests to use WithAttributeSet for performance sensitive code. Have we tried using WithAttributeSet too?

Ref: 
- Otel reference: https://github.com/open-telemetry/opentelemetry-go/blob/dfc43f0d61524042665d6c5768bf6ff8cc006d46/metric/instrument.go#L345
- GCSFuse code: https://github.com/GoogleCloudPlatform/gcsfuse/blob/078d892227c588f48efdfb79d8653a83932ef6b2/common/oc_metrics.go#L89
https://github.com/GoogleCloudPlatform/gcsfuse/blob/078d892227c588f48efdfb79d8653a83932ef6b2/common/oc_metrics.go#L81

IMO this will be a cleaner solution if it works. Maybe this is faster because we removed WithAttributeSet? I think we should test if this helps. Even if let's say this doesn't help GCSReadBytesCount metrics, I think we should do it for other metrics regardless.

kislaykishore: Yes, we did try that - it didn't help. We are planning for a right design for the implementation that'll encompass all metrics and be generic enough to handle attributes too.

ashmeenkaur (approved):

common/otel_metrics.go
```diff
@@ -122,12 +132,13 @@ func NewOTelMetrics() (MetricHandle, error) {
...
 		return nil, err
 	}
+
 	return &otelMetrics{
 		fsOpsCount:              fsOpsCount,
 		fsOpsErrorCount:         fsOpsErrorCount,
 		fsOpsLatency:            fsOpsLatency,
 		gcsReadCount:            gcsReadCount,
-		gcsReadBytesCount:       gcsReadBytesCount,
+		gcsReadBytesCountInt:    &gcsReadBytesCountInt,
```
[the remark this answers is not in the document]
kislaykishore: Makes sense. I'll do that in a follow up PR to avoid running the integration tests again.

PR #2986, opened 2025-02-10
POC: Map based adaptive prefetch

raj-prince:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/prefetch/prefetcher.go
+++ b/internal/prefetch/prefetcher.go
@@ -27,6 +27,10 @@ import (
 	"golang.org/x/net/context"
 )
 
+const (
+	min_prefetch = 10
+)
+
 type PrefetchConfig struct {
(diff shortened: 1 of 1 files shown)


internal/prefetch/block_pool.go
```diff
@@ -0,0 +1,178 @@
...
+	zeroBlock *Block
+
+	// channel to reset the data in a block
+	resetBlockCh chan *Block
+
+	// Wait group to wait for resetBlock() thread to finish
+	wg sync.WaitGroup
+
+	// Size of each block this pool holds
+	blockSize uint64
```
mkatiyar, line 40: Can you update the comment. It is not clear what is the unit of this ? Is it MB, KB, bytes ?? Why does it need to be uint64 ?

PR #2988, opened 2025-02-11
Add Streaming Writes failure scenario tests for local files.

meet2mky:
### Description

Adding Streaming writes failure scenario tests for write/truncate/sync/close operations.
 * Added check to stop the docker container if it's already running due to previous run fails.
 * Added `testOnCustomEndpoint` endpoint flag as the validation side in tests required storage client to be created with custom endpoint rather than default global endpoint.
 * Fixed the logic in proxy server to only kill the process directly listening on port 8020. This caused issues when there are other process with forwarded connections on the same port which is the case when developing on VSCode using SSH and causes test process to SIGINT.

b/388720840

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - Done


tools/integration_tests/emulator_tests/streaming_writes_failure/streaming_writes_failure_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+var (
+	proxyEndpoint = fmt.Sprintf("http://localhost:%d/storage/v1/b?project=test-project", port)
+	mountFunc     func([]string) error
+	// root directory is the directory to be unmounted.
+	rootDir string
+)
+
+func TestMain(m *testing.M) {
```
Tulsishah, line 39: This file is similar to write_stall_test.go. Why can't we keep these at the same level of write_stall?

It will reduce package building  and all basic setup time.

meet2mky: It will then remove the ability to run the test packages in parallel or selectively running test packages.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// Write next 1 MB block C may succeed based on the status of block B.
+	// Fuse:[C] -> Go-SDK:[B]-> GCS[A]
+	_, _ = t.fh1.WriteAt(data[2*operations.MiB:3*operations.MiB], 2*operations.MiB)
+
+	// Sync now reports failure from B block upload.
+	// Fuse:[] -> Go-SDK:[C]-> GCS[A, B -> upload fails]
+	operations.SyncFileShouldThrowError(t.fh1, t.T())
+	// Opening new file handle fails.
+	_, err = os.OpenFile(t.filePath, os.O_RDWR, FilePerms)
```
vadlakondaswetha, line 226: why did this fail

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// Close fails as it sees error on the finalize.
+	err = t.fh1.Close()
+
+	assert.NotNil(t.T(), err)
+	// Verify that Object is not found on GCS.
+	ValidateObjectNotFoundErrOnGCS(t.ctx, t.storageClient, testDirName, FileName1, t.T())
+	// Opening new file handle and writing to file succeeds.
+	fh3 := operations.CreateFile(t.filePath, FilePerms, t.T())
+	_, err = fh3.WriteAt(data, 0)
```
vadlakondaswetha, line 292: Can you move this into a seperate method and reuse it every where. 
Even cases where you are validating truncate bigger and sync, you can do those steps in that seperate method always. I meant there is no harm in doing sync for all test cases.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,302 @@
...
+	assert.NoError(t.T(), emulator_tests.KillProxyServerProcess(port))
+}
+
+func (t *defaultFailureTestSuite) TestStreamingWritesFailsOnSecondChunkUploadFailure() {
+	// Generate 5 MB random data.
+	data, err := operations.GenerateRandomData(5 * operations.MiB)
+	if err != nil {
+		t.T().Fatalf("Error in generating data: %v", err)
+	}
+	// Write first 2 MB (say A,B) block to file succeeds.
```
vadlakondaswetha, line 89: succeeds in the comment is misleading. 
Write first 2MB (say A, B) blocks. No error is returned since the B upload to GCS is not triggered yet.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,302 @@
...
+	// Fuse:[B] -> Go-SDK:[A]-> GCS[]
+	_, err = t.fh1.WriteAt(data[:2*operations.MiB], 0)
+	assert.NoError(t.T(), err)
+	// Write again 2MB (C, D) may or may not fail based on the status of block B upload but it ensures the block B
```
vadlakondaswetha, line 93: here also. 
Ensures the block B is uploaded is not misleading. 1st block B is not uploaded, upload is triggered. Also we are not sure if the upload gets triggered/

Restrict the comment  to
// Write again 2MB (C, D). The write may or may not fail based on the status of block B upload .

These test cases are complex and comments are the way to understand . Sorry for being more particular on the comments.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,302 @@
...
+	// Fuse:[D] -> Go-SDK:[C] -> GCS[A, B -> upload fails]
+	_, _ = t.fh1.WriteAt(data[2*operations.MiB:4*operations.MiB], 2*operations.MiB)
+
+	// Write 5th 1MB results in errors.
+	_, err = t.fh1.WriteAt(data[4*operations.MiB:5*operations.MiB], 4*operations.MiB)
+
+	assert.Error(t.T(), err)
+	// Opening new file handle succeeds.
+	fh2 := operations.OpenFile(t.filePath, t.T())
+	// Writes with fh2 also fails.
```
vadlakondaswetha, line 104: This piece of code of validating that other fileHandle is not able to write, can you move it to seperate function and reuse.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,302 @@
...
+	fh1                *os.File
+	storageClient      *storage.Client
+	closeStorageClient func() error
+	ctx                context.Context
+}
+
+// //////////////////////////////////////////////////////////////////////
+// Tests
+// //////////////////////////////////////////////////////////////////////
+func (t *defaultFailureTestSuite) SetupSuite() {
```
vadlakondaswetha, line 53: If fileHandle writes results in an error, any filehandle opened after that tries to write it fails and all fileHandles needs to be closed to recover.

If there is a fileHandle opened in read mode, we need not close that to recover. Please add test case for this in seperate PR

vadlakondaswetha (commented):
added comments for streaming writes testcases. Other refactoring and setup changes i will rely on tulsi and ashmeen's review

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: Done!

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: It's alright! Because of complexity I had put long comments. I guess, I got the wording wrong. I meant the write call succeeds as we don't see error at the time of writing but future async upload of B will result in failure. When we write more data (C, D) they may or may not see error as the Block B upload may be in any state. But the write of C and D ensures an attempt to upload the block B must be made by now and any of the future writes will see that failure in their write calls. Updated comments to be more succinct.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: Updated the comments.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: Done!

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,302 @@
...
+	fh2 := operations.OpenFile(t.filePath, t.T())
+	// Writes with fh2 also fails.
+	_, err = fh2.WriteAt(data[4*operations.MiB:5*operations.MiB], 4*operations.MiB)
+	assert.Error(t.T(), err)
+	// Closing first file handle.
+	operations.CloseFileShouldThrowError(t.fh1, t.T())
+	// Creating new file with same name will fail unless all file handles are closed.
+	// TODO(mohitkyadav): Uncomment when b/395029462 is fixed.
+	// _, err = os.OpenFile(filePath, os.O_CREATE, FilePerms)
+	// assert.Error(t.T(), err)
```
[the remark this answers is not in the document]
meet2mky: Thanks! Removed, Will add this test with the fix of this.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: SyncFile and FlushFile both seems to delete the FileINode entry from the LocalFileINode on the first error from the GCS and in current test this https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/fs.go#L1204 line deletes that entry.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
vadlakondaswetha: thats right. We will support reads on unfinalized objects in ZB, so read fileHandles shouldn't effect bwh from recovering.

vadlakondaswetha (commented):
minor comments. otherwise LGTM

tools/integration_tests/emulator_tests/emulator_tests.sh
```diff
@@ -59,7 +59,7 @@ if [ -f /etc/debian_version ]; then
     $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
     sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
   sudo apt-get update
-  sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
+  sudo apt-get install -y docker-ce docker-ce-cli containerd.io # TODO(mohitkyadav): These dependencies do not work on Cloudtop docker-buildx-plugin docker-compose-plugin
```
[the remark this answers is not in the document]
meet2mky: They seem to work in the GCE VM but not on the cloudtop even though both are Debian machines. Removed the TODO.

tools/integration_tests/util/setup/setup.go
```diff
@@ -329,6 +334,7 @@ func SetUpTestDirForTestBucketFlag() {
 		log.Printf("setUpTestDir: %v\n", err)
 		os.Exit(1)
 	}
+
```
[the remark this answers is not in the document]
meet2mky: it came due to lint issue.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,265 @@
...
+	// Opening a new file handle succeeds.
+	fh := operations.OpenFile(t.filePath, t.T())
+	// Writes with this file handle fails.
+	_, err := fh.WriteAt(data, off)
+	assert.Error(t.T(), err)
+	// Closing the file handle returns error.
+	operations.CloseFileShouldThrowError(fh, t.T())
+}
+
+func (t *defaultFailureTestSuite) WritingAfterBwhReinitializationSucceeds() {
```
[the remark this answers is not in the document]
meet2mky: Done!

tools/integration_tests/emulator_tests/emulator_tests.sh
```diff
@@ -112,5 +120,8 @@ curl -X POST --data-binary @test.json \
     "$STORAGE_EMULATOR_HOST/storage/v1/b?project=test-project"
 rm test.json
 
-# Run specific test suite
-go test ./tools/integration_tests/emulator_tests/... --integrationTest -v --testbucket=test-bucket -timeout 10m --testInstalledPackage=$RUN_E2E_TESTS_ON_PACKAGE
+# Run Write Stall Tests.
```
ashmeenkaur, line 123: what issue were we facing on running all emulator tests here? Just trying to avoid this manual step.

tools/integration_tests/util/setup/setup.go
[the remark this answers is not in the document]
ashmeenkaur: strange.. what was the lint error? Haven't seen something like this before

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (t *defaultFailureTestSuite) SetupSuite() {
+	t.configPath = "../proxy_server/configs/second_chunk_upload_returns412.yaml"
+	t.flags = []string{"--enable-streaming-writes=true", "--write-block-size-mb=1", "--write-max-blocks-per-file=1", "--custom-endpoint=" + proxyEndpoint}
+	// Generate 5 MB random data.
+	data, err := operations.GenerateRandomData(5 * operations.MiB)
+	t.data = data
+	if err != nil {
+		t.T().Fatalf("Error in generating data: %v", err)
+	}
```
ashmeenkaur, line 63: nit: 
```suggestion
	require.NoError(t.T(), err)
```

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,256 @@
...
+	setup.UnmountGCSFuse(rootDir)
+	err := t.closeStorageClient()
+	if err != nil {
+		log.Fatalf("closeStorageClient failed: %v", err)
+	}
+	assert.NoError(t.T(), emulator_tests.KillProxyServerProcess(port))
+}
+
+func (t *defaultFailureTestSuite) writingWithNewFileHandleAlsoFails(data []byte, off int64) {
+	// Opening a new file handle succeeds.
```
ashmeenkaur, line 92: add t.Helper for debugability in future to all helper functions

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,256 @@
...
+	_, err := t.fh1.WriteAt(t.data[:2*operations.MiB], 0)
+	assert.NoError(t.T(), err)
+	// Write again 2MB (C, D) will trigger B upload.
+	// Fuse:[D] -> Go-SDK:[C] -> GCS[A, B -> upload fails]
+	_, _ = t.fh1.WriteAt(t.data[2*operations.MiB:4*operations.MiB], 2*operations.MiB)
+
+	// Write 5th 1MB results in errors.
+	_, err = t.fh1.WriteAt(t.data[4*operations.MiB:5*operations.MiB], 4*operations.MiB)
+
+	assert.Error(t.T(), err)
```
ashmeenkaur, line 129:
```suggestion
	require.Error(t.T(), err)
```

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,256 @@
...
+	// Writing from new file handles also fails.
```
ashmeenkaur, line 130: can we remove redundant comments where command and function are same?

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
ashmeenkaur, line 1: any plans to add tests for synced file in follow up PR?

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,256 @@
...
+	// Fuse:[B] -> Go-SDK:[A] -> GCS[]
+	_, err := t.fh1.WriteAt(t.data[:2*operations.MiB], 0)
+	assert.NoError(t.T(), err)
+
+	// Close fails as it sees error from B block upload.
+	err = t.fh1.Close()
+
+	assert.NotNil(t.T(), err)
+	// Close file handle to reinitialize bwh.
+	operations.CloseFileShouldThrowError(t.fh1, t.T())
```
ashmeenkaur, line 232: we already closed t.fh1, why close it again? also how is the scenario different from TestStreamingWritesWhenFinalizeObjectFailure from test perspective?

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -579,6 +579,12 @@ func CloseFileShouldNotThrowError(file *os.File, t *testing.T) {
 	}
 }
 
+func CloseFileShouldThrowError(file *os.File, t *testing.T) {
```
ashmeenkaur, line 582: testing.T should be the first parameter: go/go-style/decisions#mark-test-helpers

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
ashmeenkaur, line 1: all tests look similar, why not create table driven tests:

go/go-style/decisions#table-driven-tests

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: Done!

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (t *defaultFailureTestSuite) TearDownTest() {
+	// CleanUp MntDir before unmounting GCSFuse.
+	setup.CleanUpDir(rootDir)
+	setup.UnmountGCSFuse(rootDir)
+	err := t.closeStorageClient()
+	if err != nil {
+		log.Fatalf("closeStorageClient failed: %v", err)
+	}
```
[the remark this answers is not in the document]
meet2mky: Done!

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: added in current helper functions also. Thanks!

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: Done!

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: Thanks for noticing. No need to close again, removed it.

tools/integration_tests/util/operations/file_operations.go
meet2mky: Thanks, updated the parameter ordering.

tools/integration_tests/emulator_tests/emulator_tests.sh
meet2mky: I think, it's possible to run using the single comment but we need to pass the `-p 1` to build and test one package tests binaries sequentially as both of these tests packages are using the same port and same test-bucket on the test bench server. This can cause unexpected behaviour as one test can kill proxy server for other test. Also streaming package requires --testOnCustomEndpoint flag to be passed to test binary from which it creates storage client from proxy server endpoint. In future PRs will make both the packages run in parallel when proxy server and test-bucket are separated per package.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: Sorry, I didn't reply to the other question. There is a difference in both tests on when the GCS side failure is returned. In the current test it is returned on the second chunk upload failure but in the test `TestStreamingWritesWhenFinalizeObjectFailure` it's returned on the finalize Object call to GCS.

tools/integration_tests/emulator_tests/streaming_writes_failure/local_file_failure_test.go
meet2mky: They look similar on the first few lines upto we introduce failure but in later part of assertions tests have different scenarios for write/close/sync. I have built another PR based on this so will keep in mind next time to check which of the cases would fit better.

vadlakondaswetha (approved):

ashmeenkaur:
Can you revert this and do squash & merge instead @meet2mky

PR #2989, opened 2025-02-11
do not do force stat for local files during flush

ashmeenkaur:
### Description
During Flush, we make a stat call to fetch all the attributes of an object which are required by create object call. For local files, this call can be avoided.

b/394759110

### Testing details
1. Manual - NA
2. Unit tests - added
4. Perf tests -
5RPYpZFGwbMcpkX

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -781,12 +781,13 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
 // object, syncs the content and updates the inode state.
 //
 // LOCKS_REQUIRED(f.mu)
-func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
+func (f *FileInode) syncUsingContent(ctx context.Context) error {
 	var latestGcsObj *gcs.Object
 	if !f.local {
+		var err error
 		latestGcsObj, err = f.fetchLatestGcsObject(ctx)
 		if err != nil {
-			return
+			return err
 		}
 	}
 
@@ -797,21 +798,19 @@ func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
 
 	var preconditionErr *gcs.PreconditionError
 	if errors.As(err, &preconditionErr) {
-		err = &gcsfuse_errors.FileClobberedError{
+		return &gcsfuse_errors.FileClobberedError{
 			Err: fmt.Errorf("SyncObject: %w", err),
 		}
-		return
 	}
 
 	// Propagate other errors.
 	if err != nil {
-		err = fmt.Errorf("SyncObject: %w", err)
-		return
+		return fmt.Errorf("SyncObject: %w", err)
 	}
 	minObj := storageutil.ConvertObjToMinObject(newObj)
 	// If we wrote out a new object, we need to update our state.
 	f.updateInodeStateAfterSync(minObj)
-	return
+	return nil
 }
 
 // Flush writes out contents to GCS. If this fails due to the generation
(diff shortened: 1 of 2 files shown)


internal/fs/inode/file.go
```diff
@@ -782,9 +782,12 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
...
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
-	latestGcsObj, err := f.fetchLatestGcsObject(ctx)
-	if err != nil {
-		return
+	var latestGcsObj *gcs.Object
+	if !f.local {
+		latestGcsObj, err = f.fetchLatestGcsObject(ctx)
+		if err != nil {
+			return
```
Tulsishah, line 789: nit: return err

Tulsishah (dismissed):

Tulsishah (dismissed):

internal/fs/inode/file_mock_bucket_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// Create the inode.
+	t.createInode(fileName, localFile)
+}
+
+func (t *FileMockBucketTest) TearDownTest() {
+	t.in.Unlock()
+}
+
+func (t *FileMockBucketTest) SetupSubTest() {
```
raj-prince, line 68: Do we need this, IIUC this is used for SubTests?

internal/fs/inode/file_mock_bucket_test.go
```diff
@@ -0,0 +1,155 @@
...
+func (t *FileMockBucketTest) SetupTest() {
+	// Enabling invariant check for all tests.
+	syncutil.EnableInvariantChecking()
+	t.ctx = context.Background()
+	t.clock.SetTime(time.Date(2012, 8, 15, 22, 56, 0, 0, time.Local))
+	t.bucket = new(storagemock.TestifyMockBucket)
+	t.bucket.On("BucketType").Return(gcs.BucketType{Hierarchical: false, Zonal: false})
+
+	// Create the inode.
+	t.createInode(fileName, localFile)
```
raj-prince, line 61: nit: Can we make the t.in.Lock() explicit in the SetupTest()? - This will improve the readability of TearDownTest().

internal/fs/inode/file_mock_bucket_test.go
```diff
@@ -0,0 +1,155 @@
...
+		Return(&gcs.Object{Name: fileName}, nil)
+
+	err := t.in.Flush(t.ctx)
+
+	require.NoError(t.T(), err)
+	t.bucket.AssertExpectations(t.T())
+}
+
+func (t *FileMockBucketTest) TestFlushSyncedFileForceFetchObjectFromGCS() {
+	t.bucket.On("CreateObject", t.ctx, mock.AnythingOfType("*gcs.CreateObjectRequest")).
```
raj-prince, line 141: nit: please add one liner comment, why is this CreateObject planted.

internal/fs/inode/file_mock_bucket_test.go
ashmeenkaur: Yes, I just added this as boilerplate. Will remove until we use SetupSubTest

internal/fs/inode/file_mock_bucket_test.go
ashmeenkaur: this would mean acquiring inode lock explicitly in case this is used outside of setup and can be easily missed. WDYT?

raj-prince: Then we can keep as it is.

raj-prince (approved):

PR #2993, opened 2025-02-12
Reducing the number of metric updates

kislaykishore:
* Wait for the entire buffer to be filled up before updating the metrics. This helps reduce the number of updates to the gcs/read_bytes_count metric by a factor of 1000 there by improving performance and reducing CPU and memory usage.

b/386226641

### Testing details
1. Manual - Verified reader behavior by checking that the md5 checksum of a file doesn't change when copied into a GCSFuse bucket.
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/monitor/full_read_closer.go
+++ b/internal/monitor/full_read_closer.go
@@ -1,0 +1,52 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package monitor
+
+import (
+	"io"
+
+	storagev2 "cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+)
+
+// gcsFullReadCloser wraps a gcs.StorageReader and ensures that the Read call returns:
+// if size of the buffer is more than the response, then the entire response is returned,
+// or if size of the buffer is less than or equal to the response, then the entire buffer is filled,
+// or an error is encountered.
+type gcsFullReadCloser struct {
+	wrapped gcs.StorageReader
+}
+
+func newGCSFullReadCloser(reader gcs.StorageReader) gcs.StorageReader {
+	return gcsFullReadCloser{wrapped: reader}
+}
+
+// Read reads exactly len(buf) bytes from the wrapped StorageReader into buf.
+// It returns the number of bytes copied and an error if fewer bytes were read.
+// The error is EOF only if no bytes were read.
+// If an EOF happens after reading some but not all the bytes, Read returns ErrUnexpectedEOF.
+// On return, n == len(buf) if and only if err == nil.
+// If StorageReader returns an error having read at least len(buf) bytes, the error is dropped.
+func (frc gcsFullReadCloser) Read(buf []byte) (n int, err error) {
+	return io.ReadFull(frc.wrapped, buf)
+}
+
+func (frc gcsFullReadCloser) ReadHandle() (rh storagev2.ReadHandle) {
+	return frc.wrapped.ReadHandle()
+}
+
+func (frc gcsFullReadCloser) Close() (err error) {
+	return frc.wrapped.Close()
+}
--- a/internal/monitor/full_read_closer.go
+++ b/internal/monitor/full_read_closer.go
@@ -21,10 +21,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 )
 
-// gcsFullReadCloser wraps a gcs.StorageReader and ensures that the Read call returns:
-// 1. entire response, if buffer size > response size
-// 2. entire buffer is filled, if buffer size <= response size
-// 3. error
+// gcsFullReadCloser wraps a gcs.StorageReader and ensures that the Read call reads the entire response up to the buffer size even if the wrapped read returns data in smaller chunks.
 type gcsFullReadCloser struct {
 	wrapped gcs.StorageReader
 }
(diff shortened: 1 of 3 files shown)


ashmeenkaur (approved):

internal/monitor/full_read_closer_test.go
```diff
@@ -0,0 +1,115 @@
...
+		expectedNumBytesRead int
+	}{
+		{
+			name:                 "large_buffer",
+			data:                 []byte("0123"),
+			bufSize:              5,
+			maxReadBytes:         10,
+			expectedData:         []byte("0123"),
+			expectedErr:          io.ErrUnexpectedEOF,
+			expectedNumBytesRead: 4,
```
ashmeenkaur, line 66: shall we remove this and use expectedData length?

internal/monitor/full_read_closer_test.go
```diff
@@ -0,0 +1,115 @@
...
+		maxReadBytes         int
+		expectedData         []byte
+		expectedErr          error
+		expectedNumBytesRead int
+	}{
+		{
+			name:                 "large_buffer",
+			data:                 []byte("0123"),
+			bufSize:              5,
+			maxReadBytes:         10,
```
ashmeenkaur, line 63: Since we are testing partial reads, shall we just hardcode this to a small value like 1 or 2? Trying to reduce test variables.

internal/monitor/full_read_closer_test.go
```diff
@@ -0,0 +1,106 @@
...
+			data:         []byte("0123"),
+			bufSize:      4,
+			expectedData: []byte("0123"),
+			expectedErr:  nil,
+		},
+		{
+			name:         "partial_read_full_data_returned",
+			data:         []byte("0123"),
+			bufSize:      10,
+			maxReadBytes: 2,
```
ashmeenkaur, line 83: Nit: I think you missed to remove here

internal/monitor/full_read_closer_test.go
```diff
@@ -0,0 +1,106 @@
...
+		{
+			name:         "equal_buffer",
+			data:         []byte("0123"),
+			bufSize:      4,
+			expectedData: []byte("0123"),
+			expectedErr:  nil,
+		},
+		{
+			name:         "partial_read_full_data_returned",
+			data:         []byte("0123"),
```
ashmeenkaur, line 81: This is also same throughout so can be removed.

internal/monitor/full_read_closer.go
```diff
@@ -0,0 +1,50 @@
...
+
+	storagev2 "cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+)
+
+// gcsFullReadCloser wraps a gcs.StorageReader and ensures that the Read call returns:
+// 1. entire response, if buffer size > response size
+// 2. entire buffer is filled, if buffer size <= response size
+// 3. error
+type gcsFullReadCloser struct {
```
gargnitingoogle, line 28: IMO, this new struct belongs in internal/storage/gcs or internal/storage because this is wrapping over gcs.StorageReader, and can be reused (if needed) by other packages. Even the name **gcs**FullReadCloser indicates the same need. 
Though, I can see that you've put it in monitoring because it's needed only in this package right now.

internal/monitor/full_read_closer_test.go
```diff
@@ -0,0 +1,106 @@
...
+	return n, err
+}
+
+func TestFullReaderCloser(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name         string
+		bufSize      int
+		data         []byte
+		maxReadBytes int
```
gargnitingoogle, line 54: this parameter is declared and set in one of the tests, but not used anywhere.

internal/monitor/full_read_closer_test.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		{
+			name:         "equal_buffer",
+			data:         []byte("0123"),
+			bufSize:      4,
+			expectedData: []byte("0123"),
+			expectedErr:  nil,
+		},
+		{
+			name:         "partial_read_full_data_returned",
```
gargnitingoogle, line 80: What is the difference between this case and the 'large_buffer' case ?

internal/monitor/full_read_closer.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"io"
+
+	storagev2 "cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+)
+
+// gcsFullReadCloser wraps a gcs.StorageReader and ensures that the Read call returns:
+// 1. entire response, if buffer size > response size
```
gargnitingoogle, line 25: The description of the behavior of `Read call returns` here is unnecessary here as well as slightly different from the description give at the Read call below (description of where it'll return error). Might just leave it at saying that it wraps over gcs.StorageReader and reads data up to the capacity of the passed buffer or something like that.

internal/monitor/full_read_closer_test.go
gargnitingoogle: This combined with the unused maxReadBytes gives me the impression that you forgot to add something in this test function that you were planning on adding.

gargnitingoogle (changes requested):
Added some comments

internal/monitor/full_read_closer.go
kislaykishore: Makes sense.

internal/monitor/full_read_closer_test.go
kislaykishore: yeah, given that maxReadBytes is removed, I think this is not needed now.

internal/monitor/full_read_closer_test.go
kislaykishore: Removed

internal/monitor/full_read_closer_test.go
kislaykishore: I would like to keep this here - gives a clear picture of the input and the output.

internal/monitor/full_read_closer.go
kislaykishore: This is internal (package private). If it needs to be exported in the future, we can do that then. Let me know if you think otherwise.

internal/monitor/full_read_closer_test.go
kislaykishore: Thanks!

ashmeenkaur (approved):

gargnitingoogle (approved):
LGTM

PR #2999, opened 2025-02-13
[Refactoring] Common method to wrap error as Precondition or NotFound

raj-prince:
### Description
- Created a single method which wraps the error returned by go-sdk under gcs.NotFound or gcs.PreconditionFailed. This takes care of error coming from http, grpc and control client.
- Updated all the methods of bucket_handle.go to use the new method for conversion.
- I'll enable the streaming write test for GRPC protocol as a separate PR (with more test-packages).

b/395511337

### Testing details
1. Manual - Tested the streaming write test manually by enabling that for gRPC. `GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/streaming_writes/... -p 1 --integrationTest -v --testbucket=<bucket_name> -run=TestDefaultMountLocalFileTest/TestCreateNewFileWhenSameFileExistsOnGCS`
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -115,9 +115,9 @@ func (bh *bucketHandle) DeleteObject(ctx context.Context, req *gcs.DeleteObjectR
 	}
 
 	err := obj.Delete(ctx)
-	var converted bool
-	err, converted = gcs.WrapGCSFuseError(err)
-	if err != nil && !converted {
+	var wrapped bool
+	err, wrapped = gcs.WrapGCSFuseError(err)
+	if err != nil && !wrapped {
 		err = fmt.Errorf("error in deleting object: %w", err)
 	}
 	return err
@@ -129,10 +129,10 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 	// Retrieving object attrs through Go Storage Client.
 	attrs, err = bh.bucket.Object(req.Name).Attrs(ctx)
 
-	var converted bool
-	err, converted = gcs.WrapGCSFuseError(err)
+	var wrapped bool
+	err, wrapped = gcs.WrapGCSFuseError(err)
 	if err != nil {
-		if !converted {
+		if !wrapped {
 			err = fmt.Errorf("error in fetching object attributes: %w", err)
 		}
 		return
@@ -202,9 +202,9 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 	// We can't use defer to close the writer, because we need to close the
 	// writer successfully before calling Attrs() method of writer.
 	if err = wc.Close(); err != nil {
-		var converted bool
-		err, converted = gcs.WrapGCSFuseError(err)
-		if !converted {
+		var wrapped bool
+		err, wrapped = gcs.WrapGCSFuseError(err)
+		if !wrapped {
 			err = fmt.Errorf("error in closing writer : %w", err)
 		}
 		return
@@ -235,9 +235,9 @@ func (bh *bucketHandle) CreateObjectChunkWriter(ctx context.Context, req *gcs.Cr
 
 func (bh *bucketHandle) FinalizeUpload(ctx context.Context, w gcs.Writer) (o *gcs.MinObject, err error) {
 	if err = w.Close(); err != nil {
-		var converted bool
-		err, converted = gcs.WrapGCSFuseError(err)
-		if !converted {
+		var wrapped bool
+		err, wrapped = gcs.WrapGCSFuseError(err)
+		if !wrapped {
 			err = fmt.Errorf("error in closing writer : %w", err)
 		}
 		return
@@ -266,9 +266,9 @@ func (bh *bucketHandle) CopyObject(ctx context.Context, req *gcs.CopyObjectReque
 	objAttrs, err := dstObj.CopierFrom(srcObj).Run(ctx)
 
 	if err != nil {
-		var converted bool
-		err, converted = gcs.WrapGCSFuseError(err)
-		if !converted {
+		var wrapped bool
+		err, wrapped = gcs.WrapGCSFuseError(err)
+		if !wrapped {
 			err = fmt.Errorf("error in copying object: %w", err)
 		}
 		return
@@ -405,9 +405,9 @@ func (bh *bucketHandle) UpdateObject(ctx context.Context, req *gcs.UpdateObjectR
 		return
 	}
 
-	var converted bool
-	err, converted = gcs.WrapGCSFuseError(err)
-	if !converted {
+	var wrapped bool
+	err, wrapped = gcs.WrapGCSFuseError(err)
+	if !wrapped {
 		err = fmt.Errorf("error in updating object: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


internal/storage/bucket_handle.go
```diff
@@ -124,24 +115,12 @@ func (bh *bucketHandle) DeleteObject(ctx context.Context, req *gcs.DeleteObjectR
...
-		default:
-			if err == storage.ErrObjectNotExist {
-				err = &gcs.NotFoundError{Err: storage.ErrObjectNotExist}
-			} else {
-				err = fmt.Errorf("error in deleting object: %w", err)
-			}
-		}
+	var wrapped bool
+	err, wrapped = gcs.WrapUnderCommonGCSError(err)
+	if err != nil && !wrapped {
```
ashmeenkaur, line 120: Do we really need wrapped? [go/go-style/best-practices#error-extra-info](go/go-style/best-practices#error-extra-info) (equivalent public ink: https://google.github.io/styleguide/go/best-practices.html#adding-information-to-errors)

internal/storage/gcs/errors.go
```diff
@@ -35,3 +45,50 @@ type PreconditionError struct {
 func (pe *PreconditionError) Error() string {
 	return fmt.Sprintf("gcs.PreconditionError: %v", pe.Err)
 }
+
+// WrapUnderCommonGCSError converts an error returned by go-sdk into gcsfuse specific common gcs error.
+// It returns wrapped error and a boolean indicating whether a wrapped or not.
```
ashmeenkaur, line 50: Nit: I think we should call it handled error instead if we decide to keep it. WDYT?
```suggestion
// It returns wrapped error and a boolean indicating whether error was handled or not.
```

internal/storage/gcs/errors.go
```diff
@@ -35,3 +45,50 @@ type PreconditionError struct {
...
+	if rpcErr, ok := status.FromError(err); ok {
+		switch rpcErr.Code() {
+		case codes.NotFound:
+			return &NotFoundError{Err: err}, true
+		case codes.FailedPrecondition:
+			return &PreconditionError{Err: err}, true
+		}
+	}
+
+	if err == storage.ErrObjectNotExist {
```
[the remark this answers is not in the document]
raj-prince: Seems, go-sdk does for both http as well as grpc client. Ref: http://shortn/_CY9Jyqf2wF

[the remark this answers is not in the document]
ashmeenkaur: Does that mean we don't need codes.NotFound?

internal/storage/gcs/errors.go
```diff
@@ -35,3 +45,50 @@ type PreconditionError struct {
...
+			return &PreconditionError{Err: err}, true
+		}
+	}
+
+	// RPC error.
+	if rpcErr, ok := status.FromError(err); ok {
+		switch rpcErr.Code() {
+		case codes.NotFound:
+			return &NotFoundError{Err: err}, true
+		case codes.FailedPrecondition:
```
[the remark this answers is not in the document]
raj-prince: I  see `FailedPrecondition` error for httpClient & grpcClient, since precondition is set in both cases.
```
CreateObject(\"OnlyDirMountNegativeStatCache/NegativeStatCacheTest/testAlreadyExistFolder/\") (95.279462ms): gcs.PreconditionError: error in closing writer : rpc error: code = FailedPrecondition desc = At least one of the pre-conditions you specified did not hold."}
```

I see `AlreadyExist` for `controlClient`, as there is no concept of Precondition in control client. 
```
0x8: -> CreateFolder(\"NegativeStatCacheTest/testAlreadyExistFolder/\") (81.373324ms): rpc error: code = AlreadyExists desc = The folder you tried to create already exists."}
```
Currently, this is handled in fs/error_mapping.go, I'll create a separate PR to handle this inconsistency in control-client. As it requires more discussion.

internal/storage/bucket_handle.go
raj-prince: Discussed offline, we will keep the wrapped information in all the flows. And `WrapUnderCommonGCSError` will return only converted error.

internal/storage/gcs/errors.go
raj-prince: Not required, as removed the second wrapped field.

internal/storage/gcs/errors.go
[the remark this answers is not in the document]
raj-prince: This is only for object. Go-sdk doesn't formatError for folder resources api, so for now this is required for control client error. I'll discuss this with Go-sdk, to provide consistency through out the clients.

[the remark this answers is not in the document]
ashmeenkaur: Thanks! Can we add this as comment for future reference?

internal/storage/bucket_handle.go
```diff
@@ -224,24 +202,6 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
...
-			}
-		}
-		var gErr *googleapi.Error
-		if errors.As(err, &gErr) {
-			if gErr.Code == http.StatusPreconditionFailed {
-				err = &gcs.PreconditionError{Err: err}
-				return
-			}
-		}
 		err = fmt.Errorf("error in closing writer : %w", err)
```
ashmeenkaur, line 205: why not simply return fmt.Errorf("error in closing writer : %w", gcs.WrapUnderCommonGCSError(err)) and return nil in other cases?

internal/storage/gcs/errors.go
```diff
@@ -35,3 +44,39 @@ type PreconditionError struct {
 func (pe *PreconditionError) Error() string {
 	return fmt.Sprintf("gcs.PreconditionError: %v", pe.Err)
 }
+
+// WrapUnderCommonGCSError converts an error returned by go-sdk into gcsfuse specific common gcs error.
+func WrapUnderCommonGCSError(err error) error {
```
abhishek10004, line 49: Can we think of a better error name? Maybe GetGCSError or GCSErrorWrapper?

abhishek10004 (approved):
LGTM

internal/storage/gcs/errors.go
raj-prince: GetGCSError seems better.

internal/storage/bucket_handle.go
raj-prince: Corrected in FinalizeUpload() method. Here, I didn't because we return error in case any issue in copying the content to write.

internal/storage/gcs/errors.go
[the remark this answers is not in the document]
ashmeenkaur: Thanks for checking this, I got confused between folder and object resource.

internal/storage/bucket_handle.go
```diff
@@ -450,25 +419,15 @@ func (bh *bucketHandle) UpdateObject(ctx context.Context, req *gcs.UpdateObjectR
 		return
```
ashmeenkaur, line 419: apologies for the back & forth. We should instead indent the error flow: [go/go-style/decisions#indent-error-flow](go/go-style/decisions#indent-error-flow)

abhishek10004 (approved):

ashmeenkaur (approved):

PR #3010, opened 2025-02-19
Adding UTs for negative cases in MRD

abhishek10004:
1. Adding UTs for negative cases
2. Updated handling of EOF in file_handle
3. Updated FakeMRD code to return bytes read as the 2nd parameter in the callback. Next release of go-sdk will have the changes from their side in MRD as well

### DIFF
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -15,6 +15,7 @@
 package handle
 
 import (
+	"errors"
 	"fmt"
 	"io"
 
@@ -124,7 +125,8 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 		var objectData gcsx.ObjectData
 		objectData, err = fh.reader.ReadAt(ctx, dst, offset)
 		switch {
-		case err == io.EOF:
+		case errors.Is(err, io.EOF):
+			err = io.EOF
 			return
 
 		case err != nil:
--- a/internal/gcsx/multi_range_downloader_wrapper_test.go
+++ b/internal/gcsx/multi_range_downloader_wrapper_test.go
@@ -204,7 +204,7 @@ func (t *mrdWrapperTest) Test_Read_EOF() {
 
 	_, err := t.mrdWrapper.Read(context.Background(), make([]byte, t.object.Size), 0, int64(t.object.Size), t.mrdTimeout, common.NewNoopMetrics())
 
-	assert.ErrorContains(t.T(), err, "EOF")
+	assert.ErrorIs(t.T(), err, io.EOF)
 }
 
 func (t *mrdWrapperTest) Test_Read_Error() {
--- a/internal/gcsx/multi_range_downloader_wrapper_test.go
+++ b/internal/gcsx/multi_range_downloader_wrapper_test.go
@@ -165,55 +165,36 @@ func (t *mrdWrapperTest) Test_Read() {
 	}
 }
 
-func (t *mrdWrapperTest) Test_Read_ErrorHandling() {
-	testCases := []struct {
-		name               string
-		start              int
-		end                int
-		expectedErrKeyword string
-	}{
-		{
-			name:               "ErrorInCreatingMRD",
-			start:              0,
-			end:                100,
-			expectedErrKeyword: "MultiRangeDownloader",
-		},
-		{
-			name:               "TimeoutError",
-			start:              0,
-			end:                100,
-			expectedErrKeyword: "Timeout",
-		},
-		{
-			name:               "ContextCancelled",
-			start:              0,
-			end:                100,
-			expectedErrKeyword: "Context Cancelled",
-		},
-	}
+func (t *mrdWrapperTest) Test_Read_ErrorInCreatingMRD() {
+	t.mrdWrapper.Wrapped = nil
+	t.mockBucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(nil, fmt.Errorf("Error in creating MRD")).Once()
 
-	for _, tc := range testCases {
-		t.Run(tc.name, func() {
-			ctx := context.Background()
-			buf := make([]byte, tc.end-tc.start)
-			t.mrdWrapper.Wrapped = nil
-			if tc.name == "ErrorInCreatingMRD" {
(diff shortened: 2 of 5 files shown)


internal/gcsx/random_reader_stretchr_test.go
```diff
@@ -805,47 +805,59 @@ func (t *RandomReaderStretchrTest) Test_ReadFromMultiRangeReader_ReadChunk() {
...
+				t.rr.wrapped.mrdWrapper = &fakeMRDWrapper
+				t.mockBucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(fake.NewFakeMultiRangeDownloaderWithSleep(t.object, testContent, tc.sleepTime)).Once()
+			}
+			t.mockBucket.On("BucketType", mock.Anything).Return(gcs.BucketType{Zonal: true}).Once()
 			buf := make([]byte, tc.dataSize)
 
 			bytesRead, err := t.rr.wrapped.readFromMultiRangeReader(t.rr.ctx, buf, 0, int64(t.object.Size), TestTimeoutForMultiRangeRead)
 
-			if i == len(testCases)-1 && bytesRead != 0 {
+			if tc.name == "TimeoutValue" && bytesRead != 0 {
```
vadlakondaswetha, line 854: only 1 test case matches this name. why do you need 2nd condition?

internal/gcsx/random_reader_stretchr_test.go
```diff
@@ -805,47 +805,59 @@ func (t *RandomReaderStretchrTest) Test_ReadFromMultiRangeReader_ReadChunk() {
...
 			t.rr.wrapped.reader = nil
 			t.rr.wrapped.isMRDInUse = false
 			t.object.Size = uint64(tc.dataSize)
 			testContent := testutil.GenerateRandomBytes(int(t.object.Size))
-			fakeMRDWrapper, err := NewMultiRangeDownloaderWrapperWithClock(t.mockBucket, t.object, &clock.FakeClock{})
-			assert.Nil(t.T(), err, "Error in creating MRDWrapper")
-			t.rr.wrapped.mrdWrapper = &fakeMRDWrapper
-			t.mockBucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(fake.NewFakeMultiRangeDownloaderWithSleep(t.object, testContent, tc.sleepTime)).Times(1)
-			t.mockBucket.On("BucketType", mock.Anything).Return(gcs.BucketType{Zonal: true}).Times(1)
+			if tc.name == "NilMRDWrapper" {
```
vadlakondaswetha, line 841: Its preferred to extract testcases in seperate methods if you need if-else. It becomes difficult to maintain and extend in future.
For eg, NilMRDWrapper doesn't require lot of this code. can you move it as seperate test method..

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -160,6 +165,57 @@ func (t *mrdWrapperTest) Test_Read() {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func() {
+			ctx := context.Background()
+			buf := make([]byte, tc.end-tc.start)
+			t.mrdWrapper.Wrapped = nil
+			if tc.name == "ErrorInCreatingMRD" {
+				t.mockBucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(nil, fmt.Errorf("Error in creating MRD")).Once()
```
vadlakondaswetha, line 201: Please extract these also into seperate test methods. Its okay to duplicate code in test methods for readability.

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -160,6 +165,57 @@ func (t *mrdWrapperTest) Test_Read() {
 	}
 }
 
+func (t *mrdWrapperTest) Test_Read_ErrorHandling() {
```
vadlakondaswetha, line 168: Other cases are when callback returns EOF and non EOF errors.

internal/gcsx/random_reader_stretchr_test.go
abhishek10004: Because when the sleep time is exactly same as timeout value, we can't be sure of the behavior. The read can go through or timeout can happen as well.
So, for this case, I use this second condition to determine the behavior.

internal/gcsx/multi_range_downloader_wrapper_test.go
abhishek10004: Done

vadlakondaswetha (dismissed):

internal/fs/handle/file.go
```diff
@@ -124,7 +125,8 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 		var objectData gcsx.ObjectData
 		objectData, err = fh.reader.ReadAt(ctx, dst, offset)
 		switch {
-		case err == io.EOF:
+		case errors.Is(err, io.EOF):
```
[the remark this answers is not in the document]
abhishek10004: This change is to ensure that even if EOF is wrapped under another error, we handle it properly. So, grpc/http should not matter here.

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -165,55 +166,55 @@ func (t *mrdWrapperTest) Test_Read() {
...
+	assert.Equal(t.T(), 0, bytesRead)
+}
+
+func (t *mrdWrapperTest) Test_Read_EOF() {
+	t.mrdWrapper.Wrapped = nil
+	t.mockBucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(fake.NewFakeMultiRangeDownloaderWithSleepAndDefaultError(t.object, t.objectData, time.Microsecond, io.EOF), nil).Once()
+
+	_, err := t.mrdWrapper.Read(context.Background(), make([]byte, t.object.Size), 0, int64(t.object.Size), t.mrdTimeout, common.NewNoopMetrics())
+
+	assert.ErrorContains(t.T(), err, "EOF")
```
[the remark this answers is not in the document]
abhishek10004: Done

vadlakondaswetha (approved):

PR #3011, opened 2025-02-19
Add failure scenario integration tests for empty gcs file for streaming writes.

meet2mky:
### Description
This PR extends the existing failure tests for local files to also cover scenarios with empty GCS files.

* Introduced a validator interface to check files from Google Cloud Storage (GCS) for both empty GCS files and newly created local files.
* Fixed the customEndpoint format as Bucket.Attr call was failing when creating empty gcs file because it expected the endpoint in the `http://localhost:8020/storage/v1/` format.
* Test directory cleanup is also not working on the TestBench server. To resolve this, we create randomized test directory path (testDirPath) for each test (b/397575791).

b/388720840

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - Done


tools/integration_tests/emulator_tests/streaming_writes_failure/new_local_file_failure_test.go
```diff
@@ -0,0 +1,65 @@
...
+	var err error
+	t.data, err = operations.GenerateRandomData(5 * operations.MiB)
+	require.NoError(t.T(), err)
+	log.Printf("Running tests with flags for new local files: %v", t.flags)
+}
+
+func (t *newLocalFileFailureTestSuite) SetupTest() {
+	t.setupTest()
+	// Create local file.
+	t.filePath, t.fh1 = CreateLocalFileInTestDir(t.ctx, t.storageClient, t.testDirPath, FileName1, t.T())
```
ashmeenkaur, line 52: Can we move `ValidateObjectNotFoundErrOnGCS()` here and remove the interface and test usage of `validateGcsObject()` method for simplicity.

tools/integration_tests/emulator_tests/streaming_writes_failure/empty_gcs_file_failure_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+type emptyGcsFileFailureTestSuite struct {
+	commonFailureTestSuite
+}
+
+// //////////////////////////////////////////////////////////////////////
+// Helpers
+// //////////////////////////////////////////////////////////////////////
+
+func (t *emptyGcsFileFailureTestSuite) SetupSuite() {
```
ashmeenkaur, line 40: SetupSuite function is identical for both the local and empty GCS test suites. Could we consolidate this into the common SetupSuite function in the shared commonFailureTestSuite, and set t.configPath within the SetupTest functions of the respective wrapper suites?

tools/integration_tests/emulator_tests/streaming_writes_failure/new_local_file_failure_test.go
meet2mky: `func CreateLocalFileInTestDir` already does this validation `ValidateObjectNotFoundErrOnGCS()`. The interface and `validateGcsObject()` is to validate the object from gcs in the `assert` part of the tests after the failure from emulator is introduced in both the cases when the object should be of empty size and when it shouldn't be present.

ashmeenkaur: sorry my bad, didn't realize that this was being used in a helper function. LGTM

tools/integration_tests/emulator_tests/streaming_writes_failure/empty_gcs_file_failure_test.go
meet2mky: Done!

ashmeenkaur (approved):

PR #3015, opened 2025-02-20
Use different ports for Prometheus metrics tests

kislaykishore:
### Description
* Ports don't quickly become available after use. So, this can lead to
bind errors when one tries to use the same port for multiple tests. Fix
that by using different ports for different tests.
* Use non-HNS bucket when tests are being run in non-HNS mode.

b/391754854

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/monitoring/prom_test.go
+++ b/tools/integration_tests/monitoring/prom_test.go
@@ -36,12 +36,33 @@ import (
 )
 
 const (
-	testBucket    = "gcsfuse_monitoring_test_bucket"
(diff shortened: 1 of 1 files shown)


Tulsishah (approved):
```
func (testSuite *PromTest) TearDownSuite() {
	os.RemoveAll(setup.TestDir())
}
```

Can you please remove this thing in this or next PR? I think due to this only failure logs are not sending on kokoro bucket. Bcz Logs are getting deleted with this.

PR #3024, opened 2025-02-21
Fix flaky TestConcurrentReads_ReadIsTreatedNonSequentialAfterFileIsRemovedFromCache

ashmeenkaur:
### Description
This PR addresses flaky `TestConcurrentReads_ReadIsTreatedNonSequentialAfterFileIsRemovedFromCache` test, which verifies GCSFuse's cache eviction behavior under concurrent read scenarios. Previously, concurrent reads triggered unpredictable situations like:
1. premature file removal from the cache, even before the initial read operation completed. 
2. logs order inconsistency.
This made the test flaky and unreliable.

To resolve this, the PR modifies the test to perform sequential partial reads one by one. By eliminating concurrent operations, we ensure a deterministic execution flow, allowing the cache to operate as expected. 

b/390309943

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
+++ b/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
@@ -116,7 +116,9 @@ func (s *cacheFileForRangeReadFalseTest) TestReadIsTreatedNonSequentialAfterFile
 	randomReadChunkCount := fileSizeSameAsCacheCapacity / chunkSizeToRead
 	readTillChunk := randomReadChunkCount / 2
 	fh1 := operations.OpenFile(path.Join(testDirPath, testFileNames[0]), t)
+	defer operations.CloseFile(fh1)
 	fh2 := operations.OpenFile(path.Join(testDirPath, testFileNames[1]), t)
+	defer operations.CloseFile(fh2)
 
 	// Use file handle 1 to read file 1 partially.
 	expectedOutcome[0] = readFileBetweenOffset(t, fh1, 0, int64(readTillChunk*chunkSizeToRead))
--- a/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
+++ b/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
@@ -16,7 +16,6 @@ package read_cache
 
 import (
 	"context"
-	"io"
 	"log"
 	"os"
 	"path"
@@ -74,19 +73,8 @@ func readFileBetweenOffset(t *testing.T, file *os.File, startOffset, endOffSet i
 		expected.BucketName = setup.DynamicBucketMounted()
 	}
 
-	chunkRead := make([]byte, chunkSizeToRead)
-	var readData []byte
-	for startOffset < endOffSet {
-		_, err := file.ReadAt(chunkRead, startOffset)
-		if err != nil && err != io.EOF {
-			t.Errorf("Failed to read file chunk at offset %d: %v", startOffset, err)
-		}
-		readData = append(readData, chunkRead...)
-		startOffset = startOffset + chunkSizeToRead
-	}
-
+	expected.content = operations.ReadFileBetweenOffset(t, file, startOffset, endOffSet, chunkSizeToRead)
 	expected.EndTimeStampSeconds = time.Now().Unix()
-	expected.content = string(readData)
 	return expected
 }
 
--- a/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
+++ b/tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
@@ -16,11 +16,13 @@ package read_cache
 
 import (
 	"context"
+	"io"
 	"log"
+	"os"
 	"path"
 	"strings"
-	"sync"
 	"testing"
+	"time"
 
 	"cloud.google.com/go/storage"
 	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/client"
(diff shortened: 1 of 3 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -73,8 +73,8 @@ func readFileAndGetExpectedOutcome(testDirPath, fileName string, readFullFile bo
 	return expected
 }
 
-func validate(expected *Expected, logEntry *read_logs.StructuredReadLogEntry,
-	isSeq, cacheHit bool, chunkCount int, t *testing.T) {
+func validate(expected *Expected, logEntry *read_logs.StructuredReadLogEntry, isSeq, cacheHit bool, chunkCount int, t *testing.T) {
```
vipnydav, line 76: Shouldn't t *testing.T be first parameter according to Go Style Guide?

tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
```diff
@@ -86,39 +108,43 @@ func (s *cacheFileForRangeReadFalseTest) TestRangeReadsWithCacheMiss(t *testing.
 	validateFileIsNotCached(testFileName, t)
 }
 
-func (s *cacheFileForRangeReadFalseTest) TestConcurrentReads_ReadIsTreatedNonSequentialAfterFileIsRemovedFromCache(t *testing.T) {
+func (s *cacheFileForRangeReadFalseTest) TestReadIsTreatedNonSequentialAfterFileIsRemovedFromCache(t *testing.T) {
 	var testFileNames [2]string
-	var expectedOutcome [2]*Expected
+	var expectedOutcome [4]*Expected
```
vipnydav, line 113: Instead of increasing the size to 4, wouldn't it have been better to pass a `WaitGroup` into the `readFileBetweenOffset` method, which is locked twice? This would allow us to use goroutines to read both files concurrently. After reading the first half of each file, we could unlock the lock and wait for it to be fully unlocked before proceeding with the second half. Once the `WaitGroup` is unlocked, the cache would be evicted, as we've read half of both files. This approach would result in cleaner code.

vipnydav (dismissed):
One comment, otherwise LGTM.

tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
ashmeenkaur: > Instead of increasing the size to 4, wouldn't it have been better to pass a WaitGroup into the readFileBetweenOffset method, which is locked twice

You are right, this would reduce LOC and the code would look cleaner but I think will hamper the test readability. WDYT?

tools/integration_tests/read_cache/cache_file_for_range_read_false_test.go
```diff
@@ -61,11 +63,31 @@ func (s *cacheFileForRangeReadFalseTest) Teardown(t *testing.T) {
...
+		ObjectName:            path.Join(testDirName, path.Base(file.Name())),
+	}
+	if setup.DynamicBucketMounted() != "" {
+		expected.BucketName = setup.DynamicBucketMounted()
+	}
+
+	chunkRead := make([]byte, chunkSizeToRead)
+	var readData []byte
+	for startOffset < endOffSet {
+		_, err := file.ReadAt(chunkRead, startOffset)
```
gargnitingoogle, line 80: We are not using endOffSet inside the loop anywhere. If the goal of function is to read from `[startOffset, endOffSet)` of given file, and `endOffset-startOffset` isn't a multiple of `chunkSizeToRead`, then either ReadAt should change itself or we should take a slice of readData later at `readData = append(readData, chunkRead...)` . 

The same goes for `startOffset = startOffset + chunkSizeToRead`, we should increase it only by the actual number of bytes read

ashmeenkaur: Thanks, did some refactoring to make the code generic and reusable. Please take another look!

vipnydav (dismissed):

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -316,6 +316,25 @@ func ReadChunkFromFile(filePath string, chunkSize int64, offset int64, flag int)
...
+func ReadFileBetweenOffset(t *testing.T, file *os.File, startOffset, endOffset, chunkSize int64) string {
+	t.Helper()
+	chunk := make([]byte, chunkSize)
+	var readData []byte
+
+	for startOffset < endOffset {
+		readSize := min(chunkSize, endOffset-startOffset)
+
+		n, err := file.ReadAt(chunk[:readSize], startOffset)
+		if err != nil && err != io.EOF {
```
gargnitingoogle, line 328: This will run into an infinite loop when `endOffser > size` of file, because then error will be EOF and n will be 0 and the loop will keep trying to read the next chunk or remaining bytes. 

We should check here if error is EOF and then terminate based on it. Or, we could fail in the beginning of the function if `endOffser > size`, but requires us to first figure out the file size, which is more expensive.

gargnitingoogle: this may not affect the test being modified here, but since this function is in a generic util package, its behavior should cater to all boundary cases.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -325,8 +325,17 @@ func ReadFileBetweenOffset(t *testing.T, file *os.File, startOffset, endOffset,
 		readSize := min(chunkSize, endOffset-startOffset)
 
 		n, err := file.ReadAt(chunk[:readSize], startOffset)
-		if err != nil && err != io.EOF {
-			t.Errorf("Failed to read file chunk at offset %d: %v", startOffset, err)
+		if err != nil {
+			if err == io.EOF && startOffset < endOffset {
```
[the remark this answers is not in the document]
gargnitingoogle: This also makes the comment below a bit redundant.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -325,8 +325,17 @@ func ReadFileBetweenOffset(t *testing.T, file *os.File, startOffset, endOffset,
...
+				// Reached EOF before endOffset.
+				if n > 0 {
+					readData = append(readData, chunk[:n]...)
+				}
+				break
+			} else if err != io.EOF {
+				t.Errorf("Failed to read file chunk at offset %d: %v", startOffset, err)
+				return "" //Return empty string on error.
```
gargnitingoogle, line 337: nit: comment seems unnecessary as the code is self-explanatory.

gargnitingoogle (dismissed):
added couple of nit comments about a redundant check and unnecessary comments.

charith87 (dismissed):

PR #3029, opened 2025-02-24
Parallel downloads default on with experimental flag

ankitaluthra1:
### Description
1. Adds new experimental flag file-cache-experimental-parallel-downloads-default-on with default value as false
2. Increase the frequency of updating marker while parallel downloads if experimental flag is true
3. Waits for download in parallel downloads if experimental flag is true.

b/399257714

### Testing details
1. Manual - done
4. Unit tests - done
5. Integration tests - done


cfg/validate.go
```diff
@@ -47,6 +47,7 @@ func isValidURL(u string) error {
 
 func isValidParallelDownloadConfig(config *Config) error {
 	if config.FileCache.EnableParallelDownloads {
+		// TODO: remove this check, once parallel downloads is enabled by default hence this check is redundant
```
charith87, line 50: This is not needed.

ankitaluthra1: done

charith87 (approved):
LGTM.

internal/cache/file/downloader/parallel_downloads_job.go
```diff
@@ -66,15 +66,60 @@ func (job *Job) downloadRange(ctx context.Context, dstWriter io.Writer, start, e
...
-		_, err = cacheutil.CopyUsingMemoryAlignedBuffer(ctx, newReader, dstWriter, end-start,
-			job.fileCacheConfig.WriteBufferSize)
-		// If context is canceled while reading/writing in CopyUsingMemoryAlignedBuffer
-		// then it returns error different from context cancelled (invalid argument),
-		// and we need to report that error as context cancelled.
-		if !errors.Is(err, context.Canceled) && errors.Is(ctx.Err(), context.Canceled) {
-			err = errors.Join(err, ctx.Err())
+		if job.IsExperimentalParallelDownloadsDefaultOn() {
+			for start < end {
+				writeSize := min(end-start, ReadChunkSize)
```
charith87, line 92: One thing would be to add a Unit Test for if ReadChunk is happening correctly , like a 40MB file is getting updated 5 times instead of 1

ankitaluthra1: sure..will add in subsequent PR

PR #3044, opened 2025-02-28
Fix the default value of temp-dir in --help

gargnitingoogle:
### Description
Fix the default value of temp-dir in --help

Show the default value of temp-dir, which is
"/tmp" in code, rather than keeping it
ambiguous.

[b/397660108](http://b/397660108)

### Testing details
1. Manual
   ```sh
   $ git switch master && go run . --help | grep "\/tmp"
         --temp-dir string                              Path to the temporary directory where writes are staged prior to upload to Cloud Storage. (default: system default, likely /tmp)
   $ git switch - && go run . --help | grep "\/tmp"
         --temp-dir string                              Path to the temporary directory where writes are staged prior to upload to Cloud Storage. (default: "/tmp")
   ```
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -547,7 +547,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.StringP("temp-dir", "", "", "Path to the temporary directory where writes are staged prior to upload to Cloud Storage. (default: system default, likely /tmp)")
+	flagSet.StringP("temp-dir", "", "", "Path to the temporary directory where writes are staged prior to upload to Cloud Storage. (default: \"/tmp\")")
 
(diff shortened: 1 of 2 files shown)


kislaykishore:
Thanks for the link, @gargnitingoogle. We also use tempdir to create anonymous files ([link](https://github.com/GoogleCloudPlatform/gcsfuse/blob/3b5e2165163a304c7ce73bb1d9da77ec325ed434/cmd/mount.go#L52)). That tempdir is resolved as per [this](https://cs.opensource.google/go/go/+/master:src/os/file_unix.go;l=400;drc=0044bc614af52d3f9acdd4db18e053e635c25ccd) that uses the TMPDIR environment variable if available and otherwise falls back to /tmp.
So, the ideal fixes here would be to align on using the same tempdir everywhere, infer it programmatically and then show it in the usage doc.

PR #3054, opened 2025-03-04
Add separate zonal bucket e2e test run

gargnitingoogle:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
```diff
@@ -27,9 +27,6 @@ readonly RUN_TEST_ON_TPC_ENDPOINT=false
 # This flag, if set true, will indicate to underlying script to customize for a presubmit run.
 readonly RUN_TESTS_WITH_PRESUBMIT_FLAG=true
 
-# This flag, if set true, will indicate to underlying script to also run for zonal buckets.
```
raj-prince, line 30: Why didn't we change the value to false - that will be more readable, what are being passed to run_e2e_test.sh?

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -451,68 +453,65 @@ function main(){
...
 
-  # Run tpc test and exit in case RUN_TEST_ON_TPC_ENDPOINT is true.
-  if [ $RUN_TEST_ON_TPC_ENDPOINT == true ]; then
-       run_e2e_tests_for_tpc
-  fi
-
   #run integration tests
+  exit_code=0
 
   if ${RUN_TESTS_WITH_ZONAL_BUCKET}; then
```
raj-prince, line 459: We should rename it to - RUN_TESTS_ONLY_WITH_ZONAL_BUCKET.

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -451,68 +453,65 @@ function main(){
 
   set +e
 
-  # Run tpc test and exit in case RUN_TEST_ON_TPC_ENDPOINT is true.
-  if [ $RUN_TEST_ON_TPC_ENDPOINT == true ]; then
```
raj-prince, line 455: Similarly, we can rename it to RUN_TEST_ONLY_ON_TPC_ENDPOINT.

Also, returning from start looks more readable than the if/else.

Similarly, we can do for the zonal also.

raj-prince (commented):
Added few comments.

Also, I felt run_e2e_tests.sh is hard to maintain. (a) already 6 arguments to change the behavior of scripts (b) some logic to divide parallel/non-parallel package (c) running tests in parallel and waiting for them

Given above points, we should switch to object oriented scripting language (either go or python) to make things simpler and maintainable. WDUT?

tools/integration_tests/run_e2e_tests.sh
gargnitingoogle: I don't think that RUN_TEST_ONLY_ON_TPC_ENDPOINT is any less ambiguous. I think it is more misleading, because it sounds like when it's false, it'll run on both TPC and non-TPC environment.

gargnitingoogle: I agree with the point about returning from the start though, but that will need more restructuring in this PR than what this PR is intended for.

perfmetrics/scripts/presubmit_test/pr_perf_test/build.sh
gargnitingoogle: That was the original version, but then I somehow felt that someone will come and comment about removing the variable because it's used exactly once. I'll put back the variable.

gargnitingoogle:
> WDUT?

I am not in favour of switching it to a python script because that I think will just be longer code than it will be in bash. But I agree with the need to restructure this file, especially de-dup the logic repeated a few places.

gargnitingoogle:
> Also, I don't think changing it right here in this PR is a good idea because there are other dependent PRs and it extends the scope of this PR and the task unnecessarily. I'll capture that in a separate ticket and open a separate PR for that. PLMK what you think.

@raj-prince I've captured this in [b/401468784](http://b/401468784) for now.

tools/integration_tests/run_e2e_tests.sh
raj-prince: `but that will need more restructuring in this PR than what this PR is intended for.` - Can we do as a separate PR?

raj-prince (approved):

PR #3057, opened 2025-03-05
Moving MRDWrapper to storage package

abhishek10004:
Moving MRDWrapper to storage package.

Currently MRDWrapper is in gcsx package which is not the proper location for this. MRDWrapper implements the read functionality using MRD interface provided by go sdk. Hence, moving it to the storage package.

There are no additional changes required for this move.

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -28,6 +28,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/fs/gcsfuse_errors"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/storageutil"
 	"github.com/jacobsa/fuse/fuseops"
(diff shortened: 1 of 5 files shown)


vadlakondaswetha (commented):
Can you add more details to the PR. what was the original problem, will it get solved with this PR or are we having followup PRs on this?

internal/fs/inode/file.go
```diff
@@ -94,12 +95,8 @@ type FileInode struct {
 	// Represents if local file has been unlinked.
 	unlinked bool
 
-	// Wrapper object for multi range downloader. Needed as we will create the MRD in
-	// random reader and we can't pass fileInode object to random reader as it
-	// creates a cyclic dependency.
-	// Todo: Investigate if cyclic dependency can be removed by removing some unused
-	// code.
-	MRDWrapper gcsx.MultiRangeDownloaderWrapper
+	// Wrapper object for multi range downloader.
```
abhishek10004, line 98: We cannot pass an inode to random reader due to cyclic dependency. Hence, we created an MRDWrapper so that we can pass MRDWrapper's pointer to random_reader to store the created MRD ensuring reuse across all file handles for the inode. (The comment reflects the reason for creating this wrapper instead of storing MRD here and passing the inode down to random_reader).

But now the MRDWrapper is an important structure which stores important information about the MRD (like refcount, backing object details etc.) and is necessary for providing the required functionality. Hence, updating the comment.

gargnitingoogle (approved):
Added a minor non-blocking comment.

internal/fs/inode/file.go
vadlakondaswetha: Storage is just a wrapper over go-sdk with minimal logic. This file encapsulates all read logic for serving reads from MRD. IMO this should be in gcsx package only.

PR #3066, opened 2025-03-07
Refactor methods for saving log files to KOKORO in case of test failures. 

meet2mky:
### Description
* Refactor logic for saving log files to kokoro, If the test run is not on kokoro we save the log files in TestDir to not dirty the CWD during development.
* Enabled logging for proxy server by default to ensure the logs are present in saved files.
* Minor fix in the read_stall test as mount and unmount for emulator tests need to be at Test level.
[b/398808494](http://b/398808494)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/emulator_tests/proxy_server/config.go
+++ b/tools/integration_tests/emulator_tests/proxy_server/config.go
@@ -60,9 +60,5 @@ func parseConfigFile(configPath string) (*Config, error) {
 		return nil, fmt.Errorf("unable to decode into struct, %v", err)
 	}
 
-	if *fDebug {
-		printConfig(config)
-	}
-
 	return &config, nil
 }
--- a/tools/integration_tests/emulator_tests/read_stall/read_stall_test.go
+++ b/tools/integration_tests/emulator_tests/read_stall/read_stall_test.go
@@ -44,26 +44,26 @@ type readStall struct {
 	port           int
 	proxyProcessId int
 	flags          []string
+	configPath     string
 	suite.Suite
 }
 
 func (r *readStall) SetupSuite() {
-	configPath := "../proxy_server/configs/read_stall_5s.yaml"
-	var err error
-	r.port, r.proxyProcessId, err = emulator_tests.StartProxyServer(configPath, setup.ProxyServerLogFile(r.T()))
-	require.NoError(r.T(), err)
-	setup.AppendProxyEndpointToFlagSet(&r.flags, r.port)
-	setup.MountGCSFuseWithGivenMountFunc(r.flags, mountFunc)
+	r.configPath = "../proxy_server/configs/read_stall_5s.yaml"
 }
 
 func (r *readStall) SetupTest() {
+	var err error
+	r.port, r.proxyProcessId, err = emulator_tests.StartProxyServer(r.configPath, setup.ProxyServerLogFile())
+	require.NoError(r.T(), err)
+	setup.AppendProxyEndpointToFlagSet(&r.flags, r.port)
+	setup.MountGCSFuseWithGivenMountFunc(r.flags, mountFunc)
 	testDirPath = setup.SetupTestDirectory(r.T().Name())
 }
 
-func (r *readStall) TearDownSuite() {
-	setup.UnmountGCSFuse(rootDir)
+func (r *readStall) TearDownTest() {
 	assert.NoError(r.T(), emulator_tests.KillProxyServerProcess(r.proxyProcessId))
-	setup.SaveLogFilesInCaseOfFailure(r.T())
+	setup.UnmountGCSFuseAndSaveLogFilesInCaseOfFailure(r.T(), rootDir)
 }
 
 ////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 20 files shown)


tools/integration_tests/emulator_tests/read_stall/read_stall_test.go
```diff
@@ -44,25 +44,26 @@ type readStall struct {
...
-
-func (r *readStall) SetupTest() {
 	testDirPath = setup.SetupTestDirectory(r.T().Name())
 }
 
-func (r *readStall) TearDownSuite() {
-	setup.UnmountGCSFuse(rootDir)
+func (r *readStall) TearDownTest() {
 	assert.NoError(r.T(), emulator_tests.KillProxyServerProcess(r.proxyProcessId))
+	setup.UnmountGCSFuseAndSaveLogFilesInCaseOfFailure(r.T(), rootDir)
```
Tulsishah, line 66: I think first we have to unmount and then we can kill the endpoint.

Tulsishah: But it is incorrect. You are killing endpoint where GCSFuse is already running. It will crash GCSFuse process imo.

tools/integration_tests/util/setup/setup.go
```diff
@@ -244,10 +248,51 @@ func SaveLogFileToKOKOROArtifact(artifactName string) {
...
+	} else {
+		log.Printf("Log file saved: %s", destination)
+	}
+
+}
+
+// saveLogFilesInCaseOfFailure saves log files (GCSFuse log file, Proxy Server log file)
+// only in case of test failures. For tests running on KOKORO it saves on KOKORO artifacts
+// otherwise saves in local TestDir.
+func saveLogFilesInCaseOfFailure(tb testing.TB) {
```
Tulsishah, line 279: Seems like there are many functions doing the same job. How it is different from `SaveLogFileInCaseOfFailure`

tools/integration_tests/util/setup/setup.go
```diff
@@ -244,10 +248,51 @@ func SaveLogFileToKOKOROArtifact(artifactName string) {
...
+	if tb.Failed() {
+		logDir := os.Getenv("KOKORO_ARTIFACTS_DIR")
```
Tulsishah, line 281: why can't we use `SaveLogFileToKOKOROArtifact` here?

tools/integration_tests/emulator_tests/proxy_server/config.go
```diff
@@ -60,9 +60,5 @@ func parseConfigFile(configPath string) (*Config, error) {
 		return nil, fmt.Errorf("unable to decode into struct, %v", err)
 	}
```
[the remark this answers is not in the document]
meet2mky: Already being logged in the `main.go` at line 235, We shouldn't log it twice and also when we log from here it goes to stdout instead of log file.

tools/integration_tests/emulator_tests/read_stall/read_stall_test.go
meet2mky: It didn't crash in integration tests but just to be safe, I have updated the code to unmount first and then kill the proxy server.

tools/integration_tests/util/setup/setup.go
meet2mky: It was saving only in KOKORO and if the run is not on KOKORO then it was saving in the current working directory.
Refactored both the functions and updated the code. PTAL!

Tulsishah (approved):

tools/integration_tests/emulator_tests/read_stall/read_stall_test.go
```diff
@@ -41,28 +41,33 @@ const (
...
+	proxyProcessId     int
+	proxyServerLogFile string
+	flags              []string
+	configPath         string
 	suite.Suite
 }
 
 func (r *readStall) SetupSuite() {
-	configPath := "../proxy_server/configs/read_stall_5s.yaml"
+	r.configPath = "../proxy_server/configs/read_stall_5s.yaml"
```
vipnydav, line 53: Shouldn't this be in SetupTest()? As different tests might need different configs?

vipnydav (approved):
Left a minor comment. Otherwise LGTM.

tools/integration_tests/emulator_tests/read_stall/read_stall_test.go
meet2mky: Partially yes, If different test need different config then you have to have different values for config but all tests part of this suite will run the SetupTest() so you have to create different suite structure with testify. An example is there in streaming writes.

meet2mky: I am keeping it here as Adding Tests with different config would require refactoring of suite structs.

meet2mky: Based on Test Name within SetupTest() we can configure the config path. Updated the code.

vipnydav (approved):

Tulsishah (approved):

PR #3076, opened 2025-03-11
Set appropriate defaults for a high performance node

codechanges:
### Description
This PR would implement this feature. Config changes would depends on machine type which can be obtained via:
1. `machine-type` flag
2. Querying metadata service

This feature can be disabled via `disable-autoconfig` flag.

Note: I temporarily increased time out to 20 mins as I added a new tests as well as since this is a default change it is changing behavior of scores of other tests (tests are running marginally higher time than 15 mins and it is tedious to change all of them as part of this change). I will raise a bug to fix this as a tech debt.

http://b/402273935

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA


kislaykishore:
Closing as stale.

PR #3086, opened 2025-03-13
Random read optimization by requesting application-read-size for the first read

raj-prince:
NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/random_reader.go
+++ b/internal/gcsx/random_reader.go
@@ -65,25 +65,25 @@ const TimeoutForMultiRangeRead = time.Hour
 // If gap b/w current read start and previously read content is greater than gRandomReadThreshold,
 // read is assumed as random read.
 const gRandomReadThreshold = 1 * MB
-const g3rdOrFurtherRead = HOLBlockingOffset(-2)
-const g1stRead = HOLBlockingOffset(-1)
+const gDisabled = randomReadOptimizationOffset(-2)
+const gFirstRead = randomReadOptimizationOffset(-1)
 
-type HOLBlockingOffset int64
+type randomReadOptimizationOffset int64
 
-func (h HOLBlockingOffset) offset() int64 {
+func (h randomReadOptimizationOffset) offset() int64 {
 	return int64(h)
 }
 
-func (h HOLBlockingOffset) isFirstRead() bool {
-	return h.offset() == -1
+func (h randomReadOptimizationOffset) isFirstRead() bool {
+	return h == gFirstRead
 }
 
-func (h HOLBlockingOffset) isSecondRead() bool {
+func (h randomReadOptimizationOffset) isSecondRead() bool {
 	return h.offset() >= 0
 }
 
-func (h HOLBlockingOffset) isThirdOrFurtherRead() bool {
-	return h.offset() == -2
+func (h randomReadOptimizationOffset) isDisabled() bool {
+	return h == gDisabled
 }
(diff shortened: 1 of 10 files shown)


internal/gcsx/random_reader.go
```diff
@@ -339,6 +386,25 @@ func (rr *randomReader) ReadAt(
 		return
 	}
 
+	// Special optimization for random read flow:
+	// Instead of starting a sequential read upto 200MiB, make a gcs request of size GCSFuse receives from kernel.
+	// This stops -
+	// (a) spamming the server because of large 200 MiB request (prefetch).
+	// (b) Head of line blocking issue for gRPC in case multiple reader creates a 200 MiB reader connection and
+	// don't read from that.
+	if !rr.randReadOptimizerState.isDisabled() {
```
abhishek10004, line 395: Also, by keeping it here, we are bypassing the additional checks done later in the function (like offset is within the range etc.). Please ensure that we handle all the edge cases and the behavior (in terms of error returned) remains same.

internal/gcsx/random_reader.go
```diff
@@ -339,6 +386,25 @@ func (rr *randomReader) ReadAt(
...
+		if rr.randReadOptimizerState.isFirstRead() {
+			objectData.Size, err = rr.readFromRangeReader(ctx, p, offset, offset+int64(len(p)), util.Random)
+			rr.randReadOptimizerState = randReadOptimizerState(offset + int64(len(p)))
```
Tulsishah, line 398: what if err is not equal to nil? Then also we will reset this parameter?

internal/gcsx/random_reader.go
```diff
@@ -339,6 +386,25 @@ func (rr *randomReader) ReadAt(
...
+			return
```
Tulsishah, line 399: Also we should make explicit return here. e.g.` return Obj, nil` or `nil, obj` to make it more readable.

internal/gcsx/random_reader.go
```diff
@@ -339,6 +386,25 @@ func (rr *randomReader) ReadAt(
 		return
 	}
 
+	// Special optimization for random read flow:
+	// Instead of starting a sequential read upto 200MiB, make a gcs request of size GCSFuse receives from kernel.
```
Tulsishah, line 390: If I understand correctly, this solution only addresses the issue for the initial two reads if they are random. If the first two reads are sequential, followed by a random pattern, this solution will not be effective?

internal/gcsx/random_reader.go
```diff
@@ -180,6 +216,12 @@ type randomReader struct {
 	isMRDInUse bool
 
 	metricHandle common.MetricHandle
+
+	// Stores three read-states:
+	// -1 represents the 1st read, to make the gcs request size GCSFuse receives from kernel.
+	// non-negative offset represents the 2nd read, used to decide the read-pattern relative to the 1st read.
+	// -2 represents the 3rd read (or disabled state), to disable this logic.
+	randReadOptimizerState randReadOptimizerState
```
Tulsishah, line 224: Will this cause any performance issues for sequential reads? I think yes, and if we're only addressing this issue for random reads, why don't we directly route to the random reader implementation when the flag is true, instead of rechecking if it's a sequential read? I mean if user has pass this flag true mean they are running random read workload.

PR #3088, opened 2025-03-17
Remove unnecessary allocation of memory in random reader

raj-prince:
### Description
If a fileHandle has a reader from [8MiB,100MiB), and application request comes from [13MiB, 14MiB):
GCSFuse skips the [8MiB, 13MiB) content by reading from the reader. GCSFuse allocates a temporary memory buffer up to 8 MiB to read the content from the reader, which can be avoided by discarding the content.

NA

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/random_reader.go
+++ b/internal/gcsx/random_reader.go
@@ -350,9 +350,8 @@ func (rr *randomReader) ReadAt(
 	// is a 15-20x improvement in throughput: 150-200 MB/s instead of 10 MB/s.
 	if rr.reader != nil && rr.start < offset && offset-rr.start < maxReadSize {
 		bytesToSkip := offset - rr.start
-		p := make([]byte, bytesToSkip)
-		n, _ := io.ReadFull(rr.reader, p)
-		rr.start += int64(n)
+		discardedBytes, _ := io.CopyN(io.Discard, rr.reader, int64(bytesToSkip))
+		rr.start += discardedBytes
 	}
 
 	// If we have an existing reader, but it's positioned at the wrong place,


abhishek10004 (approved):

internal/gcsx/random_reader.go
```diff
@@ -350,9 +350,8 @@ func (rr *randomReader) ReadAt(
 	// is a 15-20x improvement in throughput: 150-200 MB/s instead of 10 MB/s.
 	if rr.reader != nil && rr.start < offset && offset-rr.start < maxReadSize {
 		bytesToSkip := offset - rr.start
-		p := make([]byte, bytesToSkip)
-		n, _ := io.ReadFull(rr.reader, p)
-		rr.start += int64(n)
+		discardedBytes, _ := io.CopyN(io.Discard, rr.reader, int64(bytesToSkip))
+		rr.start += discardedBytes
```
Tulsishah, line 354: Should we handle error here? Although it is not harmful as we are anyway discarding the reader if start does not meet offset but I think it is good practice to not avoid error handling.

vadlakondaswetha: +1. Just a log line would help us understand if something goes wrong. otherwise we end up making calls to gcs and never know this failed.

vadlakondaswetha (approved):

internal/gcsx/random_reader.go
raj-prince: Make sense, added a warning log for the error other than io.EOF.

Tulsishah (approved):

internal/gcsx/random_reader.go
```diff
@@ -350,9 +350,12 @@ func (rr *randomReader) ReadAt(
 	// is a 15-20x improvement in throughput: 150-200 MB/s instead of 10 MB/s.
 	if rr.reader != nil && rr.start < offset && offset-rr.start < maxReadSize {
 		bytesToSkip := offset - rr.start
-		p := make([]byte, bytesToSkip)
-		n, _ := io.ReadFull(rr.reader, p)
-		rr.start += int64(n)
+		discardedBytes, copyError := io.CopyN(io.Discard, rr.reader, int64(bytesToSkip))
+		// io.EOF is expected when reader is shorter than required offset to read.
```
Tulsishah, line 354: Not blocking this PR but I could see this warning 
```
The double modal "required offset" is nonstandard (only accepted in certain dialects). Consider to be offset.
```
required offset --> offset.

raj-prince:
image

vadlakondaswetha (approved):

PR #3098, opened 2025-03-18
e2e tests for validation grpc directpath

anushka567:
### Description
Follow up of [this](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3062) PR . E2E tests to check for the following scenarios :

- gRPC DirectPath Activation in case the test VM and the single region test bucket are colocated
- gRPC DirectPath Activation **Failure** in case the test VM and the single region test bucket are **not colocated**
- gRPC DirectPath Activation in case the test VM and the multi region test bucket are colocated
- gRPC DirectPath Activation **Failure** in case the test VM and the multi region test bucket are **not colocated**

Based on the GCSFuse mount logs, we are verifying directpath successful or unsuccessful activation.

- For successful activation , the mount logs include the following INFO log:
`Successfully connected over gRPC DirectPath for <test-bucket>`
- For unsuccessful scenario , the mount logs include the following warning log:
`Direct path connectivity unavailable for <test-bucket>, reason: <reason>`
[b/376372141](b/376372141)

### Testing details
1. Manual - Tested on a GCE VM
```
=== RUN   TestGRPCValidationSuite
=== RUN   TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsSuccessfulWithMultiRegionBucket
=== RUN   TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsSuccessfulWithSingleRegionBucket
=== RUN   TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsUnsuccessfulWithMultiRegionBucket
=== RUN   TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsUnsuccessfulWithSingleRegionBucket
--- PASS: TestGRPCValidationSuite (12.49s)
    --- PASS: TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsSuccessfulWithMultiRegionBucket (0.62s)
    --- PASS: TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsSuccessfulWithSingleRegionBucket (0.59s)
    --- PASS: TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsUnsuccessfulWithMultiRegionBucket (2.47s)
    --- PASS: TestGRPCValidationSuite/TestGRPCDirectPathConnectionIsUnsuccessfulWithSingleRegionBucket (1.00s)
PASS
ok      github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/grpc_validation       12.509s
```
3. Unit tests - NA
4. Integration tests - NA


tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	err = os.Remove(g.logFile)
+	assert.NoError(g.T(), err)
+}
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+// Single Region Bucket tests
+func (g *gRPCValidation) TestGRPCDirectPathConnectionIsUnsuccessfulWithSingleRegionBucket() {
```
raj-prince, line 146: Can we improve the test-name? - seems a bit verbose and hard to follow

Maybe we can rename like - `TestSingleRegionBucket_DirectPath_Failure` (all the tests)?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+		g.T().Fatalf("Failed to mount the GCS Bucket using GCSFuse. Shutting down tests..")
+	}
+
+	defer func() {
+		if err := util.Unmount(g.mountPoint); err != nil {
+			g.T().Logf("Warning: unmount failed: %v", err)
+		}
+	}()
+
+	file, err := os.Open(g.logFile)
```
raj-prince, line 187: We can refactor Log File Parsing logic, which can be utilized at other places.

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	for scanner.Scan() {
+		if strings.Contains(scanner.Text(), expectedLog) {
+			directPathLogFound = true
+			break
+		}
+	}
+	require.True(g.T(), directPathLogFound, "gRPC DirectPath connection log message not found in log file")
+}
+
+// Multi Region Bucket tests
```
raj-prince, line 204: Can we combine these tests in the table format? As logic seems same for all the tests? Something like (below code is AI generated).

```go
func (g *gRPCValidation) TestGRPCDirectPathConnections() {
	testCases := []struct {
		name                     string
		bucketName               string
		expectedSuccess          bool
		expectedLogSubstring     string
		isMultiRegion            bool
	}{
		{
			name:                     "SingleRegion_Success",
			bucketName:               g.singleRegionBucketForGRPCSuccess,
			expectedSuccess:          true,
			expectedLogSubstring:     fmt.Sprintf("Successfully connected over gRPC DirectPath for %s", g.singleRegionBucketForGRPCSuccess),
			isMultiRegion:            false,
		},
		{
			name:                     "SingleRegion_Failure",
			bucketName:               g.singleRegionBucketForGRPCFailure,
			expectedSuccess:          false,
			expectedLogSubstring:     fmt.Sprintf("Direct path connectivity unavailable for %s, reason: ", g.singleRegionBucketForGRPCFailure),
			isMultiRegion:            false,
		},
		{
			name:                     "MultiRegion_Success",
			bucketName:               g.multiRegionBucketForGRPCSuccess,
			expectedSuccess:          true,
			expectedLogSubstring:     fmt.Sprintf("Successfully connected over gRPC DirectPath for %s", g.multiRegionBucketForGRPCSuccess),
			isMultiRegion:            true,
		},
		{
			name:                     "MultiRegion_Failure",
			bucketName:               g.multiRegionBucketForGRPCFailure,
			expectedSuccess:          false,
			expectedLogSubstring:     fmt.Sprintf("Direct path connectivity unavailable for %s, reason: ", g.multiRegionBucketForGRPCFailure),
			isMultiRegion:            true,
		},
	}

		g.T().Run(tc.name, func(t *testing.T) {
			args := []string{"--client-protocol=grpc", "--log-severity=TRACE", fmt.Sprintf("--log-file=%s", g.logFile), tc.bucketName, g.mountPoint}
			err := mounting.MountGcsfuse(g.gcsfusePath, args)
			if err != nil {
				if tc.expectedSuccess {
					t.Errorf("Unexpected mount failure: %v", err)
				}
				//If expected failure, no need to check log file.
				g.TearDownTest()
				return
			}

			defer func() {
				if err := util.Unmount(g.mountPoint); err != nil {
					t.Logf("Warning: unmount failed: %v", err)
				}
			}()

			success := g.checkLogFileForMessage(g.logFile, tc.expectedLogSubstring, tc.expectedSuccess)
			require.Equal(t, tc.expectedSuccess, success, fmt.Sprintf("Expected success: %t, but got: %t", tc.expectedSuccess, success))
		})
	}
}
```

tools/integration_tests/grpc_validation/setup_test.go
```diff
@@ -0,0 +1,189 @@
...
+	return nil
+}
+
+func TestMain(m *testing.M) {
+	// Parse flags from the setup.
+	setup.ParseSetUpFlags()
+
+	var err error
+
+	if setup.TestInstalledPackage() {
```
Tulsishah, line 160: Sorry why are we not using functions from utils/setup.go file?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	var err error
+	// Find the test environment, if run in cloudtop or issue fetching the test environment, skip all tests.
+	if g.testRegion, err = findTestExecutionEnvironment(g.ctx); g.testRegion == cloudtopProd || err != nil {
+		g.T().Logf("Skipping tests as environment is cloudtop.")
+		g.T().Skip()
+	}
+
+	setup.SetLogFile(fmt.Sprintf("/tmp/grpc_validation_mount_cmds_%d.txt", time.Now().UnixNano()))
+	if g.client, err = storage.NewClient(g.ctx); err != nil {
+		g.T().Fatalf("Failed to create storage client...")
```
Tulsishah, line 62: Not printing actual error here.

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	singleRegionBucketForGRPCSuccess string
+	singleRegionBucketForGRPCFailure string
+	multiRegionBucketForGRPCSuccess  string
+	multiRegionBucketForGRPCFailure  string
+	mountPoint                       string
+	gcsfusePath                      string
+	logFile                          string
+}
+
+func (g *gRPCValidation) SetupSuite() {
```
Tulsishah, line 50: Can we write comment, what is require in setupSuite. It is becoming quite long or we can break it?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	// Creating a common storage client for the test
+	g.ctx = context.Background()
+	var err error
+	// Find the test environment, if run in cloudtop or issue fetching the test environment, skip all tests.
+	if g.testRegion, err = findTestExecutionEnvironment(g.ctx); g.testRegion == cloudtopProd || err != nil {
+		g.T().Logf("Skipping tests as environment is cloudtop.")
+		g.T().Skip()
+	}
+
+	setup.SetLogFile(fmt.Sprintf("/tmp/grpc_validation_mount_cmds_%d.txt", time.Now().UnixNano()))
```
Tulsishah, line 60: Will we able to send it to kokoro bucket in case of failure. Also let's create Var for file name?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	singleRegionForGRPCSuccess := findSingleRegionForGRPCDirectPathSuccessCase(g.testRegion)
+	// Figure out the test region for single region test: failure case
+	singleRegionForGRPCFailure := pickFailureRegionFromListOfRegions(singleRegionForGRPCSuccess, single_regions)
+	// Figure out the test region for multi region test: success case
+	multiRegionForGRPCSuccess := findMultiRegionForGRPCDirectPathSuccessCase(g.testRegion)
+	// Figure out the test region for multi region test: failure case
+	multiRegionForGRPCFailure := pickFailureRegionFromListOfRegions(multiRegionForGRPCSuccess, multi_regions)
+
+	// Set the test bucket names
+	g.singleRegionBucketForGRPCSuccess = createTestBucketName(singleRegionForGRPCSuccess)
```
Tulsishah, line 76: What if last test crashed or someone have stopped the execution. The bucket will be still there right? And due to that this execution will fail. Can we fix those scenarios?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	// Specifying the path to gcsfuse binary for mounting
+	g.gcsfusePath = path.Join(gBuildDir, "bin/gcsfuse")
+}
+
+func (g *gRPCValidation) TearDownSuite() {
+	if g.client != nil {
+		// Delete the test buckets
+		if err := DeleteBucket(g.ctx, g.client, g.singleRegionBucketForGRPCSuccess); err != nil {
```
Tulsishah, line 102: Can we create for loop for this?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+			g.T().Logf("Failed to delete bucket %s: %v", g.singleRegionBucketForGRPCSuccess, err)
+		}
+		if err := DeleteBucket(g.ctx, g.client, g.multiRegionBucketForGRPCSuccess); err != nil {
+			g.T().Logf("Failed to delete bucket %s: %v", g.singleRegionBucketForGRPCSuccess, err)
+		}
+		if err := DeleteBucket(g.ctx, g.client, g.multiRegionBucketForGRPCFailure); err != nil {
+			g.T().Logf("Failed to delete bucket %s: %v", g.singleRegionBucketForGRPCSuccess, err)
+		}
+
+		g.client.Close()
```
Tulsishah, line 115: Can this be deferred? Additionally, since we aren't creating the client here, why are we closing it? It should be closed with a defer statement where it is created

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	logFile                          string
+}
+
+func (g *gRPCValidation) SetupSuite() {
+	// Creating a common storage client for the test
+	g.ctx = context.Background()
+	var err error
+	// Find the test environment, if run in cloudtop or issue fetching the test environment, skip all tests.
+	if g.testRegion, err = findTestExecutionEnvironment(g.ctx); g.testRegion == cloudtopProd || err != nil {
+		g.T().Logf("Skipping tests as environment is cloudtop.")
```
Tulsishah, line 56: Not printing err in case err != nil ?

tools/integration_tests/grpc_validation/setup_test.go
```diff
@@ -0,0 +1,189 @@
...
+	if err = bucket.Create(ctx, gcp_project, &storage.BucketAttrs{Location: testBucketRegion}); err != nil {
+		log.Printf("Error while creating bucket")
+		return err
+	}
+
+	return nil
+}
+
+// Delete the test bucket after the test is complete.
+func DeleteBucket(ctx context.Context, client *storage.Client, testBucketName string) (err error) {
```
Tulsishah, line 140: Can we generalize [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/mounting/dynamic_mounting/dynamic_mounting.go#L124) function instead of creating new one?

tools/integration_tests/grpc_validation/setup_test.go
```diff
@@ -0,0 +1,189 @@
...
+////////////////////////////////////////////////////////////////////////
+// Helper Functions
+////////////////////////////////////////////////////////////////////////
+
+// For both multi region buckets and single region buckets test, we need to decide failure case
+// region based on the region of the VM on which the test is running.
+func findTestExecutionEnvironment(ctx context.Context) (string, error) {
+	detectedAttrs, err := resource.New(ctx, resource.WithDetectors(gcp.NewDetector()))
+	if err != nil {
+		log.Printf("Error fetching the test environment.All tests will be skipped.")
```
Tulsishah, line 76: %v, err in log statement ?

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: This is not required as this log file simply contains the commands used to run the test and not the actual test logs.

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: While creating the bucket, we are appending unique suffix to prevent these errors.In the rare case when the tear down is skipped/ fails mid execution, the created buckets will have to be deleted manually.Please note that this will not interfere with subsequent test runs.

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: Initially we needed the created client to delete buckets i.e. after test execution so closing it only in the teardown made sense. 
However, now all client handling logic is in TestMain itself,so I have removed this close call from the teardown function.

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: Added the function to operations package.

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: Using testcase based approach.
The test names follow the following format: TestGRPCDirectPathConnections/{testcase_name}
For e.g. `TestGRPCDirectPathConnections/SingleRegion_Success`

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: Added a comment to explain the test scenarios and the setup.

tools/integration_tests/grpc_validation/setup_test.go
anushka567: Added this function to utils , made required changes in dependant test packages

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,157 @@
...
+				}
+				return
+			}
+
+			defer func() {
+				if err := util.Unmount(mountPoint); err != nil {
+					t.Logf("Warning: unmount failed: %v", err)
+				}
+				os.Remove(mountPoint)
+				os.Remove(logFile)
```
Tulsishah, line 151: Incase of failure also will if remove the file?

tools/integration_tests/grpc_validation/setup_test.go
```diff
@@ -0,0 +1,166 @@
...
+	// Parse flags from the setup.
+	var err error
+	setup.ParseSetUpFlags()
+	if err := setup.SetUpTestDir(); err != nil {
+		log.Fatalf("Failed to setup GCSFuse package. Error: %v", err)
+	}
+
+	// Creating a common storage client for the test
+	ctx = context.Background()
+	if client, err = storage.NewClient(ctx); err != nil {
```
Tulsishah, line 147: Can we use [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/client/storage_client.go#L43) function it self

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -779,3 +780,18 @@ func CloseLocalFile(t *testing.T, f **os.File) error {
 	*f = nil
 	return err
 }
+
+func CheckLogFileForMessage(t *testing.T, expectedLog, logFile string) bool {
+	fmt.Printf("logmessage: %s", expectedLog)
```
Tulsishah, line 785: Is this intended? Should it be `log.Println`?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,157 @@
...
+// Test Cases:
+// Test Case 1: single region test: success case , when test VM and bucket are colocated
+// Test Case 2: single region test: failure case , when test VM and bucket are out-of-region
+// Test Case 3: multi region test: success case , when test VM and bucket are colocated
+// Test Case 4: multi region test: failure case , when test VM and bucket are out-of-region
+func (g *gRPCValidation) SetupSuite() {
+
+	// Based on the test region which is initialized in the TestMain() function,
+	// we will pick out the test region for:
+	singleRegionForGRPCSuccess := findSingleRegionForGRPCDirectPathSuccessCase(testRegion)
```
Tulsishah, line 54: Can't we do this block by block per entry?
e.g.
```
singleRegionForGRPCSuccess := findSingleRegionForGRPCDirectPathSuccessCase(testRegion)
g.singleRegionBucketForGRPCSuccess = createTestBucketName(singleRegionForGRPCSuccess)

	// Create the test buckets
	if err := createTestBucket(singleRegionForGRPCSuccess, g.singleRegionBucketForGRPCSuccess); err != nil {
		g.T().Fatalf("Could not create bucket in the required region, err : %v", err)
	}

...
```

Tulsishah: Or apply for loop?

tools/integration_tests/grpc_validation/grpc_validation_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+			defer func() {
+				if err := util.Unmount(mountPoint); err != nil {
+					t.Logf("Warning: unmount failed: %v", err)
+				}
+				os.Remove(mountPoint)
+				os.Remove(logFile)
+			}()
+			success := operations.CheckLogFileForMessage(g.T(), tc.expectedLogSubstring, logFile)
+			require.Equal(t, true, success)
```
Tulsishah, line 154: I think it will be assert as we want to go to next test case even if current one failed?

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: For the bucket creation, the name and region vary for each entry. this means maintaining an additional data structure: map only for this use case. I suggest we avoid this.

tools/integration_tests/util/operations/file_operations.go
anushka567: No , this was not supposed to be here, removed it.

tools/integration_tests/grpc_validation/grpc_validation_test.go
anushka567: Since we are looping through the testcases,  the testcases are still executed in case any intermediate test case fail as in the scenario when require.Equal fails, it halts the test for that particular testcase only and not the subsequent testcases.

raj-prince (approved):

Tulsishah (approved):

PR #3103, opened 2025-03-20
Integrate flush writer with upload handler

ashmeenkaur:
### Description
Integrate flush writer with upload handler

b/404999845

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -171,6 +171,29 @@ func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
 	return obj, nil
 }
 
+// Flush uploads any data in the write buffer.
+func (uh *UploadHandler) Flush() (int64, error) {
+	uh.wg.Wait()
+
+	if uh.writer == nil {
+		// Writer may not have been created for empty file creation flow or for very
+		// small writes of size less than 1 block.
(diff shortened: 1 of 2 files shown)


internal/bufferedwrites/upload_handler.go
```diff
@@ -171,6 +171,29 @@ func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
 	return obj, nil
 }
 
+// Flush uploads any data in the write buffer.
+func (uh *UploadHandler) Flush() (int64, error) {
+	uh.wg.Wait()
+
+	if uh.writer == nil {
```
vadlakondaswetha, line 178: can you extract this into common method like ensureWriter and reuse it in finalize() & upload methods.

internal/bufferedwrites/upload_handler.go
```diff
@@ -171,6 +171,29 @@ func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
 	return obj, nil
 }
 
+// Flush uploads any data in the write buffer.
+func (uh *UploadHandler) Flush() (int64, error) {
```
vadlakondaswetha, line 175: Flush in fs.go has a different meaning. Can you rename this also to FlushPendingWrites

PR #3104, opened 2025-03-20
Refactor stale handle package and start running streaming writes tests for ZB.

meet2mky:
### Description
Refactoring stale handle package as both streaming writes and without streaming writes stale handle package have a lot of similarity. This also starts running stale handle tests for streaming writes for ZB.

[b/405033976](http://b/405033976)
[b/412949432](http://b/412949432)

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - Done


tools/integration_tests/stale_handle/stale_file_handle_common_test.go
```diff
@@ -30,39 +32,80 @@ import (
...
+}
+
 ////////////////////////////////////////////////////////////////////////
 // Tests
 ////////////////////////////////////////////////////////////////////////
 
 func (s *staleFileHandleCommon) TestClobberedFileSyncAndCloseThrowsStaleFileHandleError() {
+	// TODO(b/410698332): Remove skip condition once takeover support is ready.
+	if s.streamingWritesEnabled() && setup.IsZonalBucketRun() {
+		s.T().Skip("Skip the test as takeover support is required to overwrite the zonal objects.")
```
ashmeenkaur, line 76: IIUC, in this test we are creating a finalized object to clobber the file. I don't think we need takeover support to overwrite finalized objects. Maybe I am missing something trivial here.. could you explain why this isn't working for ZB?

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
```diff
@@ -30,39 +32,80 @@ import (
...
 	assert.NoError(s.T(), err)
+	operations.SyncFile(s.f1, s.T())
 	// Delete the file.
 	operations.RemoveFile(s.f1.Name())
 	// Verify unlink operation succeeds.
+
 	operations.ValidateNoFileOrDirError(s.T(), s.f1.Name())
 	// Attempt to write to file should not give any error.
-	operations.WriteWithoutClose(s.f1, Content2, s.T())
+	_, err = s.f1.WriteAt([]byte(s.data), int64(bytesWrote))
```
ashmeenkaur, line 106: WriteWithoutClose uses fh.Write, was it not working?

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
```diff
@@ -30,39 +32,80 @@ import (
...
 
+	s.validateESTALEErrorIfStreamingWritesDisabled(err)
```
ashmeenkaur, line 108: instead of adding if conditions for scenarios which are different, shall we create separate suites that test for that particular scenario with particular flagset? Some tests like renames are difficult to understand in one go

ashmeenkaur (commented):
I think there are scenarios untested for zonal buckets.. where sync creates a new object generation.. can we add them too?

tools/integration_tests/stale_handle/setup_test.go
```diff
@@ -54,15 +58,27 @@ func TestMain(m *testing.M) {
 	// Else run tests for testBucket.
 	// Set up test directory.
 	setup.SetUpTestDirForTestBucketFlag()
+	rootDir = setup.MntDir()
 
 	// Define flag set to run the tests.
 	flagsSet := [][]string{
 		{"--metadata-cache-ttl-secs=0", "--precondition-errors=true"},
+		{"--metadata-cache-ttl-secs=0", "--precondition-errors=true", "--enable-streaming-writes=true", "--write-block-size-mb=1", "--write-max-blocks-per-file=1"},
```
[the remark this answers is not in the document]
meet2mky: Done!

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
meet2mky: It is working, changed to this.

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
meet2mky: Created separate suite and validator, now streaming writes and non-streaming writes test run in different test suite.

meet2mky:
> I think there are scenarios untested for zonal buckets.. where sync creates a new object generation.. can we add them too?

We can add more tests in next PR when generation issue is fixed so it works in both streaming writes and non-streaming writes flow. This PR is only refactoring and running streaming writes package on ZB.

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
ashmeenkaur: Got it. Can we update the message to "Skip the test until unfinalized object overwrite is supported"

meet2mky: Done!

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
```diff
@@ -30,39 +32,58 @@ import (
...
+}
+
 ////////////////////////////////////////////////////////////////////////
 // Tests
 ////////////////////////////////////////////////////////////////////////
 
 func (s *staleFileHandleCommon) TestClobberedFileSyncAndCloseThrowsStaleFileHandleError() {
+	if s.isStreamingWritesEnabled && setup.IsZonalBucketRun() {
+		s.T().Skip("Skip test due to takeover support not available.")
+	}
```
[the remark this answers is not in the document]
ashmeenkaur: same comment for tools/integration_tests/stale_handle/stale_file_handle_synced_file_test.go too

tools/integration_tests/stale_handle_streaming_writes/stale_file_handle_streaming_writes_common_test.go
```diff
@@ -1,91 +0,0 @@
...
-	// Clobber file by replacing the underlying object with a new generation.
-	err = WriteToObject(ctx, storageClient, path.Join(testDirName, t.fileName), FileContents, storage.Conditions{})
-	assert.NoError(t.T(), err)
-	operations.SyncFile(t.f1, t.T())
-
-	// Closing the file/writer returns stale NFS file handle error.
-	err = t.f1.Close()
-
-	operations.ValidateESTALEError(t.T(), err)
-	ValidateObjectContentsFromGCS(ctx, storageClient, testDirName, t.fileName, FileContents, t.T())
```
ashmeenkaur, line 90: Can we add a check like this at the end for our refactored tests?

tools/integration_tests/stale_handle_streaming_writes/stale_file_handle_streaming_writes_empty_gcs_file_test.go
```diff
@@ -1,111 +0,0 @@
...
-
-	// Attempt first write to the file.
-	_, err = t.f1.WriteAt([]byte(t.data), 0)
-
-	operations.ValidateESTALEError(t.T(), err)
-	operations.SyncFile(t.f1, t.T())
-	operations.CloseFileShouldNotThrowError(t.T(), t.f1)
-}
-
-func (t *staleFileHandleStreamingWritesEmptyGcsFile) TestWriteOnRenamedFileThrowsStaleFileHandleError() {
```
ashmeenkaur, line 67: I think this test is missed after refactoring

tools/integration_tests/stale_handle/setup_test.go
```diff
@@ -38,7 +41,10 @@ var (
 func TestMain(m *testing.M) {
 	setup.ParseSetUpFlags()
 	setup.ExitWithFailureIfBothTestBucketAndMountedDirectoryFlagsAreNotSet()
-
+	if setup.MountedDirectory() != "" {
+		log.Printf("These tests will not run with mounted directory..")
```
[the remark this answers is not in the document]
meet2mky: These tests have different behaviour based on streaming writes is enabled or not and we validate those tests based on the streaming writes flag is On or Off. If we run these tests with mountedDir then by default `isStreamingWritesEnabed` variable will be false and the tests would run only for one type of behaviour even when the mountedDir is using streaming writes or not.

tools/integration_tests/stale_handle/stale_file_handle_rename_file_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
+	. "github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/client"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/setup"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRenamedFileSyncAndCloseThrowsStaleFileHandleError(t *testing.T) {
```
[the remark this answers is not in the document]
meet2mky: 1. Similar test already present at : https://github.com/GoogleCloudPlatform/gcsfuse/blob/3cb261e97f6f87879dd822862b357552b3182c55/tools/integration_tests/streaming_writes/rename_file_test.go#L66
2. Which synced_file package? Are you saying create a new package?
3. (Streaming Writes (Synced/Local) in above 1). Non-Streaming Writes Local will add in this test.

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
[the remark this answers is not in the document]
meet2mky: > Will we enable these tests after upgrading goSDK?
Not until takeover support is available for taking over on existing unfinalized object, finalizing it and creating a new generation of unfinalized object.
> Also takeover support will only apply for append use case which is not the case here I believe so we can update the error message accordingly.
This current test at it's core verifying that once object is clobbered our existing Sync/Close behaviour see the ESTALE Error. In terms of zonal buckets this could now be branched into two types ->
A) Clobber file by only appending the data to the end thus size mismatch from GCS/GCSFuse
B) Clobber file by finalizing existing unfinalized object and creating a new unfinalized Generation.
I think currently this test verifies B) but not A) so once the takeover support is available we need to revisit how we can test these scenarios exhaustively.

tools/integration_tests/stale_handle_streaming_writes/stale_file_handle_streaming_writes_common_test.go
meet2mky: Done!

tools/integration_tests/stale_handle/setup_test.go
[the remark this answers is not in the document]
ashmeenkaur: this will break existing GKE tests as they are already running this package. Seems like we need to inform them to disable these tests in that case.

tools/integration_tests/stale_handle/stale_file_handle_common_test.go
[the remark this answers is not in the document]
ashmeenkaur: makes sense. I think the core failure here is that we cannot overwrite an unfinalized object. Shall we update the Skip message accordingly so whenever that is supported, we can enable these tests?

tools/integration_tests/stale_handle/setup_test.go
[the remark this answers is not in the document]
meet2mky: Can you let me know what all places we are running mountedDir tests? If there is any place where we can see the GKE-GCSFuse tests or within GCSFuse which are running as part of automation?

tools/integration_tests/stale_handle/stale_file_handle_rename_file_test.go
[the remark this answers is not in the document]
ashmeenkaur: >1. Similar test already present at :

Nit: I think that test fits better in this package. WDYT?

> 2. Which synced_file package? Are you saying create a new package?

Sorry my bad, I meant the emptyGCSFile test suite. Updated my comment.

tools/integration_tests/stale_handle_streaming_writes/stale_file_handle_streaming_writes_empty_gcs_file_test.go
ashmeenkaur: tests in tools/integration_tests/stale_handle/stale_file_handle_rename_file_test.go are not running with streaming writes package. Let's continue the conversation in the same thread.

tools/integration_tests/stale_handle/stale_file_handle_rename_file_test.go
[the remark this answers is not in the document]
meet2mky: In terms of testing ESTALE ERROR after rename operation, I agree that this package makes more sense. We shouldn't duplicate test so if you strongly think that we should move that test here then we can remove that test from the streaming writes package and put it here and verify for all of the combinations. Wdyt?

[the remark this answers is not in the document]
meet2mky: Made the required tests. Thanks!

tools/integration_tests/stale_handle/setup_test.go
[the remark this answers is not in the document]
meet2mky: Now, Stale handle tests will run for non-streaming writes mountedDir tests only and this won't break the current GKE tests. Verified locally.

tools/integration_tests/stale_handle/stale_file_handle_synced_file_test.go
```diff
@@ -40,11 +40,11 @@ type staleFileHandleEmptyGcsFile struct {
 // //////////////////////////////////////////////////////////////////////
 
 func (s *staleFileHandleEmptyGcsFile) SetupTest() {
-	s.testDirPath = setup.SetupTestDirectory(s.T().Name())
```
ashmeenkaur: is s.testDirPath now unused?

ashmeenkaur (approved):

tools/integration_tests/stale_handle/stale_file_handle_synced_file_test.go
meet2mky: Instead of setting it again and again in `SetupTest`, moved it to `SetupSuite`. It's being used even now just changed initialization setup.

PR #3105, opened 2025-03-20
Experimental Parallel downloads flag default on

ankitaluthra1:
### Description
Makes default value to true for file-cache-experimental-parallel-downloads-default-on.
Updates parallel download unit tests for marker flag and experimental flag true.

b/404435879

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Done


internal/cache/file/downloader/parallel_downloads_job_test.go
```diff
@@ -81,30 +82,39 @@ func (dt *parallelDownloaderTest) Test_downloadRange() {
...
 	start, end = int64(5*util.MiB), int64(5*util.MiB)
 	offsetWriter = io.NewOffsetWriter(file, start)
-	_, err = dt.job.downloadRange(context.Background(), offsetWriter, start, end, nil, make(map[int64]int64))
+	_, err = dt.job.downloadRange(context.Background(), offsetWriter, start, end, nil, rangeMap)
 	AssertEq(nil, err)
 	verifyContentAtOffset(file, start, end)
+
+	//Verify marker map updated correctly
+	expectedMap := make(map[int64]int64)
+	expectedMap[0] = 4194304
```
charith87, line 114: Can we use annotations here 4*MiB and other way here. Its more readable

ankitaluthra1: Updated

cfg/rationalize.go
```diff
@@ -97,6 +97,14 @@ func resolveCloudMetricsUploadIntervalSecs(m *MetricsConfig) {
 	}
 }
 
+func resolveParallelDownloadsValue(v isSet, fc *FileCacheConfig, c *Config) {
+	// Parallel downloads should be default ON when file cache is enabled, in case
+	// it is explicitly set by the user, use that value.
+	if IsFileCacheEnabled(c) && !v.IsSet(FileCacheParallelDownloadsConfigKey) {
+		fc.EnableParallelDownloads = true
```
ashmeenkaur, line 104: Should we directly change EnableParallelDownloads flag's default to `true` in params.yaml? Any cons of doing that?

charith87 (approved):

cfg/rationalize.go
ankitaluthra1: this is needed because we still need [this condition](https://github.com/GoogleCloudPlatform/gcsfuse/blob/cce2741f9ee84a4cb6daeb72c26eff633012eb1f/cfg/validate.go#L50) to work

PR #3108, opened 2025-03-21
Integrate buffered write handler with writer.flush

ashmeenkaur:
### Description
Integrate buffered write handler with writer.flush

b/405337871

### Testing details
1. Manual - NA
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -179,8 +179,27 @@ func (wh *bufferedWriteHandlerImpl) appendBuffer(data []byte) (err error) {
 }
 
 func (wh *bufferedWriteHandlerImpl) Sync() (err error) {
-	// Upload all the pending buffers and release the buffers.
+	// Upload current block (for both regional and zonal buckets).
+	if wh.current != nil && wh.current.Size() != 0 {
+		err := wh.uploadHandler.Upload(wh.current)
+		if err != nil {
+			return err
+		}
+		wh.current = nil
+	}
+	// Upload all the pending buffers.
 	wh.uploadHandler.AwaitBlocksUpload()
+	// Flush writer in case of a zonal bucket.
+	if wh.uploadHandler.bucket.BucketType().Zonal {
+		n, err := wh.uploadHandler.Flush()
+		if err != nil {
+			return err
+		}
+		if n != wh.totalSize {
+			return fmt.Errorf("could not upload entire data, expected offset %d, Got %d", wh.totalSize, n)
+		}
+	}
+	// Release memory used by buffers.
 	err = wh.blockPool.ClearFreeBlockChannel()
 	if err != nil {
 		// Only logging an error in case of resource leak as upload succeeded.
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -189,9 +189,12 @@ func (wh *bufferedWriteHandlerImpl) Sync() (err error) {
 	}
 	// Upload all the pending buffers.
 	wh.uploadHandler.AwaitBlocksUpload()
-	// Flush writer in case of a zonal bucket.
+	// The FlushPendingWrites method synchronizes all bytes currently residing in
+	// the Writer's buffer to Cloud Storage, thereby making them available for
+	// other operations like read.
+	// This functionality is exclusively supported on zonal buckets.
 	if wh.uploadHandler.bucket.BucketType().Zonal {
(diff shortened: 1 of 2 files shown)


internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -179,8 +179,27 @@ func (wh *bufferedWriteHandlerImpl) appendBuffer(data []byte) (err error) {
...
+	if wh.current != nil && wh.current.Size() != 0 {
+		err := wh.uploadHandler.Upload(wh.current)
+		if err != nil {
+			return err
+		}
+		wh.current = nil
+	}
+	// Upload all the pending buffers.
 	wh.uploadHandler.AwaitBlocksUpload()
+	// Flush writer in case of a zonal bucket.
```
vadlakondaswetha, line 192: Can you add a comment what extra are we getting from Flush method?

internal/bufferedwrites/buffered_write_handler_test.go
```diff
@@ -289,6 +296,62 @@ func (testSuite *BufferedWriteTest) TestSync5InProgressBlocks() {
...
+	// Wait for 3 blocks to upload successfully.
+	err = testSuite.bwh.Sync()
+
+	assert.NoError(testSuite.T(), err)
+	bwhImpl := testSuite.bwh.(*bufferedWriteHandlerImpl)
+	// Current block should also be uploaded.
+	assert.Nil(testSuite.T(), bwhImpl.current)
+	assert.Equal(testSuite.T(), 0, len(bwhImpl.uploadHandler.uploadCh))
+	assert.Equal(testSuite.T(), 0, len(bwhImpl.blockPool.FreeBlocksChannel()))
+	// Read the file from back door.
```
vadlakondaswetha, line 320: add a comment that we can't read since the object is not finalized.

gargnitingoogle:
@ashmeenkaur Please enable the test label `execute-integration-tests-on-zb` for testing as that run the e2e tests on ZB/ rapid bucket as this test changes behavior for rapid buckets too .

internal/bufferedwrites/buffered_write_handler_test.go
```diff
@@ -289,6 +296,63 @@ func (testSuite *BufferedWriteTest) TestSync5InProgressBlocks() {
 	assert.Equal(testSuite.T(), 0, len(bwhImpl.blockPool.FreeBlocksChannel()))
 }
 
+func (testSuite *BufferedWriteTest) TestSyncPartialBlock() {
+	buffer, err := operations.GenerateRandomData(blockSize)
```
gargnitingoogle, line 300: nit: As most of the code is common between both these new test functions, the common logic can be pulled out into a common function like. This will be easier to maintain and less prone to errors.

```go
func (testSuite *BufferedWriteTest) testSyncPartialBlockCommon(useZB, shouldFailReadAfterSync bool) {
...
}
```

internal/bufferedwrites/buffered_write_handler_test.go
```diff
@@ -289,6 +296,63 @@ func (testSuite *BufferedWriteTest) TestSync5InProgressBlocks() {
...
+	var notFoundErr *gcs.NotFoundError
+	assert.ErrorAs(testSuite.T(), err, &notFoundErr)
+}
+
+func (testSuite *BufferedWriteTest) TestSyncPartialBlockOnZB() {
+	testSuite.setupTestWithBucketType(gcs.BucketType{Zonal: true})
+	buffer, err := operations.GenerateRandomData(blockSize)
+	assert.NoError(testSuite.T(), err)
+	// Write 2 blocks.
+	for i := 0; i < 2; i++ {
```
gargnitingoogle, line 333: Does it add any value to add another test having only an incomplete written buffer before sync, without the 2 completed buffers ?

gargnitingoogle (commented):
Added couple of comments.

internal/bufferedwrites/buffered_write_handler_test.go
ashmeenkaur: creating helper functions that test a core functionality is not a good test practice in go: https://google.github.io/styleguide/go/best-practices.html#leave-testing-to-the-test-function

I converted this to a table driven test instead.

internal/bufferedwrites/buffered_write_handler_test.go
ashmeenkaur: the path is same but added for the sake of completeness. Thanks!

vadlakondaswetha (approved):

PR #3110, opened 2025-03-24
cd script changes to run zb tests only based on flag passed

anushka567:
### Description
Added `RUN_ON_ZB_ONLY` flag to run integration tests only for ZB via this script.

Follow up CL:
* Changes in louhi pipeline to setup stages for the zb-only tests

### Motivation
ZB integration tests cannot be run in parallel with flat or HNS buckets. Hence, in our release pipeline , we are going to run these tests on a seperate VM altogether. To ensure this targeted test execution, we pass this field as a VM metadata argument , which is consumed the same VM's starter script (e2e_test.sh)

[b/405035152](http://b/405035152)

### Testing details
1. Manual - Checked script in cloned louhi pipeline. WAI.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -30,7 +30,7 @@ RUN_ON_ZB_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAM
 echo "RUN_ON_ZB_ONLY flag set to : \"${RUN_ON_ZB_ONLY}\""
 
 # Logging the tests being run on the active GCE VM
-if [[ "$RUN_ON_ZB_ONLY" == "false" ]]; then
+if [[ "$RUN_ON_ZB_ONLY" == "true" ]]; then
   echo "Running integration tests for Zonal bucket only..."
 else
   echo "Running integration tests for non-zonal buckets only..."
@@ -58,12 +58,12 @@ set -e
 # Print commands and their arguments as they are executed.
 set -x
 
-# Export the RUN_E2E_TESTS_FOR_ZB_ONLY variable so that it is available in the environment of the 'starterscriptuser' user.
+# Export the RUN_ON_ZB_ONLY variable so that it is available in the environment of the 'starterscriptuser' user.
 # Since we are running the subsequent script as 'starterscriptuser' using sudo, the environment of 'starterscriptuser' 
-# would not automatically have access to the environment variables set by the original user (i.e. $RUN_E2E_TESTS_FOR_ZB_ONLY).
-# By exporting this variable, we ensure that the value of RUN_E2E_TESTS_FOR_ZB_ONLY is passed into the 'starterscriptuser' script 
+# would not automatically have access to the environment variables set by the original user (i.e. $RUN_ON_ZB_ONLY).
+# By exporting this variable, we ensure that the value of RUN_ON_ZB_ONLY is passed into the 'starterscriptuser' script 
 # and can be used for conditional logic or decisions within that script.
-export RUN_E2E_TESTS_FOR_ZB_ONLY='$RUN_ON_ZB_ONLY'
+export RUN_ON_ZB_ONLY='$RUN_ON_ZB_ONLY'
 
 #Copy details.txt to starterscriptuser home directory and create logs.txt
 cd ~/
@@ -73,7 +73,7 @@ touch logs-hns.txt
 touch logs-zonal.txt
 GENERIC_LOG_FILE='~/logs.txt'
 
-if [[ "$RUN_E2E_TESTS_FOR_ZB_ONLY" == "true" ]]; then
+if [[ "$RUN_ON_ZB_ONLY" == "true" ]]; then
   GENERIC_LOG_FILE='~/logs-zonal.txt'
 fi
   
@@ -394,7 +394,7 @@ function gather_test_logs() {
   done
 }
 
-if [[ "$RUN_E2E_TESTS_FOR_ZB_ONLY" == "true" ]]; then
+if [[ "$RUN_ON_ZB_ONLY" == "true" ]]; then
   echo "Started integration tests for Zonal bucket ..."
   run_e2e_tests_for_zonal_bucket &
   e2e_tests_zonal_bucket_pid=$!
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -363,7 +363,7 @@ function run_e2e_tests_for_zonal_bucket(){
   non_parallel_tests_zonal_group_exit_code=$?
 
   if [ $parallel_tests_zonal_group_exit_code != 0 ] || [ $non_parallel_tests_zonal_group_exit_code != 0 ]; then
-  return 1
+    return 1
   fi
 }
 
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -21,21 +21,23 @@ set -e
 # Extract the metadata parameters passed, for which we need the zone of the GCE VM 
 # on which the tests are supposed to run.
 ZONE=$(curl -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/zone)
+echo "Got ZONE=\"${ZONE}\" from metadata server."
 # The format for the above extracted zone is projects/{project-id}/zones/{zone}, thus, from this
 # need extracted zone name.
 ZONE_NAME=$(basename $ZONE)
 # This parameter is passed as the GCE VM metadata at the time of creation.(Logic is handled in louhi stage script)
 RUN_ON_ZB_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAME" --format='get(metadata.run-on-zb-only)')
+echo "RUN_ON_ZB_ONLY flag set to : \"${RUN_ON_ZB_ONLY}\""
 
 # Logging the tests being run on the active GCE VM
-if [[ "$RUN_ON_ZB_ONLY" == "true" ]]; then
+if [[ "$RUN_ON_ZB_ONLY" == "false" ]]; then
   echo "Running integration tests for Zonal bucket only..."
 else
-  echo "Running integration tests for Flat buckets..."
+  echo "Running integration tests for non-zonal buckets only..."
 fi
 
 #details.txt file contains the release version and commit hash of the current release.
-gsutil cp  gs://gcsfuse-release-packages/version-detail/details.txt .
+gcloud storage cp  gs://gcsfuse-release-packages/version-detail/details.txt .
 # Writing VM instance name to details.txt (Format: release-test-<os-name>)
 curl http://metadata.google.internal/computeMetadata/v1/instance/name -H "Metadata-Flavor: Google" >> details.txt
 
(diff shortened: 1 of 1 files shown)


gargnitingoogle (changes requested):
added some comments

gargnitingoogle:
@anushka567 one of the biggest comments I have is the duplication of the code, as the code in tools/cd_scripts/e2e_test.sh is mostly a copy of the code in [tools/integration_tests/run_e2e_tests.sh](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/run_e2e_tests.sh) . This leads to a lot of duplication and added maintenance. We should handle this at a later point to avoid delaying your work. I have captured this in b/406171917 .

tools/cd_scripts/e2e_test.sh
```diff
@@ -40,14 +56,28 @@ set -e
 # Print commands and their arguments as they are executed.
 set -x
 
+# Export the RUN_E2E_TESTS_FOR_ZB_ONLY variable so that it is available in the environment of the 'starterscriptuser' user.
+# Since we are running the subsequent script as 'starterscriptuser' using sudo, the environment of 'starterscriptuser' 
+# would not automatically have access to the environment variables set by the original user (i.e. $RUN_E2E_TESTS_FOR_ZB_ONLY).
+# By exporting this variable, we ensure that the value of RUN_E2E_TESTS_FOR_ZB_ONLY is passed into the 'starterscriptuser' script 
+# and can be used for conditional logic or decisions within that script.
+export RUN_E2E_TESTS_FOR_ZB_ONLY='$RUN_ON_ZB_ONLY'
```
gargnitingoogle, line 64: Why do we need this 2nd variable? Couldn't we have just sufficed with the name RUN_ON_ZB_ONLY alone?

tools/cd_scripts/e2e_test.sh
```diff
@@ -40,14 +56,28 @@ set -e
...
 touch logs.txt
 touch logs-hns.txt
+touch logs-zonal.txt
+GENERIC_LOG_FILE='logs.txt'
+
+if [[ "$RUN_E2E_TESTS_FOR_ZB_ONLY" == "true" ]]; then
+  GENERIC_LOG_FILE='logs-zonal.txt'
+fi
+  
+echo User: $USER &>> ~/${GENERIC_LOG_FILE}
```
gargnitingoogle, line 78: Put echo parameters in quotes here for better readability.

```suggestion
echo "User: $USER" &>> ~/${GENERIC_LOG_FILE}
```

tools/cd_scripts/e2e_test.sh
```diff
@@ -289,6 +345,33 @@ function run_e2e_tests_for_hns_bucket(){
...
+  wait $parallel_tests_zonal_group_pid
+  parallel_tests_zonal_group_exit_code=$?
+  wait $non_parallel_tests_zonal_group_pid
+  non_parallel_tests_zonal_group_exit_code=$?
+
+  if [ $parallel_tests_zonal_group_exit_code != 0 ] || [ $non_parallel_tests_zonal_group_exit_code != 0 ];
+  then
+  return 1
+  fi
+  return 0
```
gargnitingoogle, line 371: nit: explicitly returning 0 is not necessary in shell functions. It's automatic.

tools/cd_scripts/e2e_test.sh
anushka567: Since there was a discussion regarding running only read_cache tests on certain VMs and rest of the test packages on the remaining, I introduced this variable to account for the case when we just want to run e2e tests except read cache , and only on zonal buckets. The corresponding variable for read_cache test would be something like `RUN_READ_CACHE_TESTS_FOR_ZB` defined as `$RUN_ON_ZB_ONLY && $RUN_READ_CACHE_TESTS_ONLY`

Since this selective testing of the read_cache test package is not implemented yet, hence I did not add any thing explicitly here.However, I can add a note here for future reference.

anushka567: To reduce the number of metadata arguments passed to the GCE VM's starter script, which originates from louhi stages, I've implemented this approach. This avoids creating a separate flag for each specific scenario.

tools/cd_scripts/e2e_test.sh
```diff
@@ -313,53 +398,75 @@ function gather_test_logs() {
...
-    touch success-hns.txt
-    gsutil cp success-hns.txt gs://gcsfuse-release-packages/v$(sed -n 1p ~/details.txt)/$(sed -n 3p ~/details.txt)/
-fi
-gsutil cp ~/logs-hns.txt gs://gcsfuse-release-packages/v$(sed -n 1p ~/details.txt)/$(sed -n 3p ~/details.txt)/
-
-if [ $e2e_tests_emulator_status != 0 ];
-then
-    echo "Test failures detected in emulator based tests." &>> ~/logs-emulator.txt
+if [[ "$RUN_E2E_TESTS_FOR_ZB_ONLY" == "true" ]]; then
+  echo "Running integration tests for Zonal bucket only..."
```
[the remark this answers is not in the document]
anushka567: The log at line 32 identifies the test scope as either zonal-only or not.
To enhance clarity regarding test initiation, this log, and its counterparts, have been modified to
`Started integration tests for Zonal/HNS/FLAT bucket ...`

tools/cd_scripts/e2e_test.sh
anushka567: Made similar changes to  similar pre-existing functions as well .

tools/cd_scripts/e2e_test.sh
```diff
@@ -282,13 +338,36 @@ function run_e2e_tests_for_hns_bucket(){
...
+  non_parallel_tests_zonal_group_pid=$!
+
+  # Wait for all tests to complete.
+  wait $parallel_tests_zonal_group_pid
+  parallel_tests_zonal_group_exit_code=$?
+  wait $non_parallel_tests_zonal_group_pid
+  non_parallel_tests_zonal_group_exit_code=$?
+
+  if [ $parallel_tests_zonal_group_exit_code != 0 ] || [ $non_parallel_tests_zonal_group_exit_code != 0 ]; then
+  return 1
```
gargnitingoogle, line 366:
```suggestion
    return 1
```

tools/cd_scripts/e2e_test.sh
```diff
@@ -313,53 +394,76 @@ function gather_test_logs() {
...
+      touch success-emulator.txt
+      gcloud storage cp success-emulator.txt gs://gcsfuse-release-packages/v$(sed -n 1p ~/details.txt)/$(sed -n 3p ~/details.txt)/
+  fi
+
+  gcloud storage cp ~/logs-emulator.txt gs://gcsfuse-release-packages/v$(sed -n 1p ~/details.txt)/$(sed -n 3p ~/details.txt)/
 fi
 
-gsutil cp ~/logs-emulator.txt gs://gcsfuse-release-packages/v$(sed -n 1p ~/details.txt)/$(sed -n 3p ~/details.txt)/
-'
+'
```
gargnitingoogle, line 469: Out of curiosity, what is this `'` closing ?

tools/cd_scripts/e2e_test.sh
gargnitingoogle: this is still not crystal clear to me. Let's resolve this offline.

tools/cd_scripts/e2e_test.sh
```diff
@@ -18,8 +18,26 @@ set -x
...
+ZONE_NAME=$(basename $ZONE)
+# This parameter is passed as the GCE VM metadata at the time of creation.(Logic is handled in louhi stage script)
+RUN_ON_ZB_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAME" --format='get(metadata.run-on-zb-only)')
+echo "RUN_ON_ZB_ONLY flag set to : \"${RUN_ON_ZB_ONLY}\""
+
+# Logging the tests being run on the active GCE VM
+if [[ "$RUN_ON_ZB_ONLY" == "false" ]]; then
+  echo "Running integration tests for Zonal bucket only..."
+else
+  echo "Running integration tests for non-zonal buckets only..."
```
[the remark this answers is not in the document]
anushka567: Yes, This logic is handled in the louhi stage itself due to which this script will always receive either true or false.

tools/cd_scripts/e2e_test.sh
anushka567: We are creating the user account `starterscriptuser` to run the e2e tests. This ` closes the starter script which is passed to this user.

tools/cd_scripts/e2e_test.sh
anushka567: As discussed offline, this variable is renamed to `RUN_ON_ZB_ONLY` to avoid creating a second variable for now.

If in the future the need arises to run selected test packages on a single VM for a particular bucket type, we can reintroduce this variable for efficient handling.

tools/cd_scripts/e2e_test.sh
```diff
@@ -18,8 +18,26 @@ set -x
...
+# The format for the above extracted zone is projects/{project-id}/zones/{zone}, thus, from this
+# need extracted zone name.
+ZONE_NAME=$(basename $ZONE)
+# This parameter is passed as the GCE VM metadata at the time of creation.(Logic is handled in louhi stage script)
+RUN_ON_ZB_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAME" --format='get(metadata.run-on-zb-only)')
+echo "RUN_ON_ZB_ONLY flag set to : \"${RUN_ON_ZB_ONLY}\""
+
+# Logging the tests being run on the active GCE VM
+if [[ "$RUN_ON_ZB_ONLY" == "false" ]]; then
+  echo "Running integration tests for Zonal bucket only..."
```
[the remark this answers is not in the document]
anushka567: Corrected this now,Sorry for the oversight.

gargnitingoogle (dismissed):
LGTM

gargnitingoogle:
@anushka567 is it possible to provide a link to the louhi pipeline run that you did with these changes that show that this set up works ?

tools/cd_scripts/e2e_test.sh
```diff
@@ -40,14 +58,28 @@ set -e
...
+# would not automatically have access to the environment variables set by the original user (i.e. $RUN_ON_ZB_ONLY).
+# By exporting this variable, we ensure that the value of RUN_ON_ZB_ONLY is passed into the 'starterscriptuser' script 
+# and can be used for conditional logic or decisions within that script.
+export RUN_ON_ZB_ONLY='$RUN_ON_ZB_ONLY'
+
 #Copy details.txt to starterscriptuser home directory and create logs.txt
 cd ~/
 cp /details.txt .
 touch logs.txt
 touch logs-hns.txt
```
vadlakondaswetha, line 72: how sets the log file as logs-hns?

tools/cd_scripts/e2e_test.sh
```diff
@@ -40,14 +58,28 @@ set -e
...
+touch logs-zonal.txt
+GENERIC_LOG_FILE='~/logs.txt'
```
vadlakondaswetha, line 74: nit: rename it to log_file

vadlakondaswetha (approved):

tools/cd_scripts/e2e_test.sh
anushka567: logs-hns.txt file will only contain test logs from the HNS bucket tests. 

Each of logs-flat/hns/zb contain the test logs for the respective bucket tests.
For writing out the generic info (containing GCSFuse version etc. ), we pick logs.txt in case of the VMs running non zonal tests and logs-zonal.txt in case of zonal tests.

Tulsishah: @vadlakondaswetha this is not GCSFuse logs and Pass fail tests logs.

These log files are intended to gather logs for pass/fail tests. In the CD pipeline tests, we are collecting:

1. Installation dependency logs and e2e tests for flat buckets in common `logs.txt`.
2. HNS e2e tests in `logs-hns.txt`.
3. Emulator e2e tests in `logs-emulator.txt`.

Now, for ZB, as it will run on a separate VM, we need to implement a similar logging mechanism as described in point 1. 

Alternatively, we could organize the logs such that all cases have an `installation.txt` file containing installation logs. The pass/fail logs for the remaining e2e tests would then reside in separate files named logs-flat.txt, logs-hns.txt, logs-zb.txt, etc. 

However, this approach would require re-testing the Louhi pipeline and would likely add an additional 1-2 days for the PR to be merged. Additionally it will require changes in louhi pipeline also.

anushka567: Resolving this conversation for now. Will take AI for optimization if required.

Tulsishah (approved):

PR #3119, opened 2025-03-24
Integrate default values functionality

codechanges:
### Description
This PR would implement part of this feature. This change integrates optimize functionality within config initialization flow.

http://b/402273935

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA


cfg/rationalize.go
```diff
@@ -39,38 +39,68 @@ func decodeURL(u string) (string, error) {
 
 // resolveMetadataCacheTTL returns the ttl to be used for stat/type cache based
 // on the user flags/configs.
-func resolveMetadataCacheTTL(v isSet, c *MetadataCacheConfig) {
+func resolveMetadataCacheTTL(v isSet, c *MetadataCacheConfig, optimizedFlags []string) {
```
raj-prince, line 42: Can we update the method docs, kind of summary? As method doesn't seem very readable, suggesting what I understood?

```
// resolveMetadataCacheTTL determines the final TTL values for metadata and negative caches.
//
// Precedence for TtlSecs:
// 1. User explicitly sets the new 'metadata-cache.ttl-secs' flag.
// 2. User explicitly sets deprecated 'stat-cache-ttl' or 'type-cache-ttl' flags (and not 'metadata-cache.ttl-secs').
// 3. Value set by machine-type optimization for 'metadata-cache.ttl-secs'.
// 4. Default value derived from the defaults of deprecated 'stat-cache-ttl' and 'type-cache-ttl'.
```

**Code**: Also, I would prefer less nesting, something like below.
```go
        if isTTLSetByUser {
		// 1. Highest precedence: User set the new flag.
		//    The value is already parsed into c.TtlSecs by viper/flags.
		finalTTL = c.TtlSecs
	} else if isStatTTLSetByUser || isTypeTTLSetByUser {
		// 2. Second precedence: User set only deprecated flags.
		finalTTL = ttlFromDeprecated
	} else if isTTLOptimized {
		// 3. Third precedence: Optimization applied.
		//    The value was already applied to c.TtlSecs by the optimization step.
		finalTTL = c.TtlSecs
	} else {
		// 4. Lowest precedence: No user settings, no optimization.
		//    Use the value derived from deprecated flags' defaults.
		finalTTL = ttlFromDeprecated
	}

```

cfg/rationalize.go
```diff
@@ -39,38 +39,68 @@ func decodeURL(u string) (string, error) {
...
-		if c.StatCacheMaxSizeMb == -1 {
-			c.StatCacheMaxSizeMb = int64(maxSupportedStatCacheMaxSizeMB)
+
+	// if any of two (StatCacheMaxSizeConfigKey, MetadataCacheStatCacheCapacityKey) are set then ->
+	// 	if StatCacheMaxSizeConfigKey is set then use that
+	// 	else use DeprecatedStatCacheCapacity
+	// else if optimization not applied set default value
+	// otherwise optimization would have been applied so mutate accordingly
+	optimizationAppliedToStatCacheMaxSize := isFlagPresent(optimizedFlags, StatCacheMaxSizeConfigKey)
+	if v.IsSet(StatCacheMaxSizeConfigKey) || v.IsSet(MetadataCacheStatCacheCapacityConfigKey) {
```
raj-prince, line 91: Similarly, can we simply the nesting?

```go
	if isSizeSetByUser {
		// 1. Highest precedence: User set the new size flag.
		finalSizeMB = c.StatCacheMaxSizeMb
	} else if isCapacitySetByUser {
		// 2. Second precedence: User set only the deprecated capacity flag.
		finalSizeMB = sizeFromDeprecatedCapacity
	} else if isSizeOptimized {
		// 3. Third precedence: Optimization applied.
		//    The value was already applied to c.StatCacheMaxSizeMb by the optimization step.
		finalSizeMB = c.StatCacheMaxSizeMb
	} else {
		// 4. Lowest precedence: No user settings, no optimization.
		//    Use the value derived from the deprecated capacity flag's default.
		finalSizeMB = sizeFromDeprecatedCapacity
	}
```

cfg/rationalize_test.go
```diff
@@ -304,14 +304,106 @@ func TestRationalizeMetadataCache(t *testing.T) {
...
+			},
+			expectedTTLSecs:         math.MaxInt64 / int64(time.Second), // Max supported ttl in seconds.
+			expectedNegativeTTLSecs: math.MaxInt64 / int64(time.Second), // Max supported ttl in seconds.
+			expectedStatCacheSize:   math.MaxUint64 >> 20,               // Max supported cache size in MiB.
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if assert.NoError(t, Rationalize(tc.flags, tc.config, []string{"metadata-cache.negative-ttl-secs", "metadata-cache.ttl-secs", "metadata-cache.stat-cache-max-size-mb", "metadata-cache.deprecated-stat-cache-capacity", "metadata-cache.deprecated-stat-cache-ttl", "metadata-cache.deprecated-type-cache-ttl"})) {
```
raj-prince, line 398: Instead of if, we can use require to stop the test on error.

```go
require.NoError(t, err)
assert.Equal(...)
assert.Equal(...)
assert.Equal(...)
```

raj-prince: Also, can we follow Arrange, Act, Assert?

cmd/root.go
```diff
@@ -86,7 +86,12 @@ of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 		if cfgErr = cfg.ValidateConfig(v, &configObj); cfgErr != nil {
 			return
 		}
-		if cfgErr = cfg.Rationalize(v, &configObj); cfgErr != nil {
+		var optimizedFlags []string
```
raj-prince, line 89: nit: empty line.

We can be consistent of spacing.

raj-prince (commented):
Overall looks good. 

Added few comments, mostly about readability.

cmd/root.go
codechanges: Fixed

cfg/rationalize_test.go
codechanges: The structure of this test mirrors other tests in the file and divergence would break uniformity. Lets create a tech debt item to take care of this for each test.

cfg/rationalize.go
codechanges: Yes, you are spot on regarding the precedence logic. I had following in mind w.r.t regarding the current code to create sections of precedence
1. Any flag is set
2. Optimization applied.

But I hear you and will try to make it simpler.

cfg/rationalize.go
```diff
@@ -39,38 +39,68 @@ func decodeURL(u string) (string, error) {
 
 // resolveMetadataCacheTTL returns the ttl to be used for stat/type cache based
 // on the user flags/configs.
-func resolveMetadataCacheTTL(v isSet, c *MetadataCacheConfig) {
+func resolveMetadataCacheTTL(v isSet, c *MetadataCacheConfig, optimizedFlags []string) {
 	// If metadata-cache:ttl-secs has been set, then it overrides both
```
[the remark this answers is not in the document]
codechanges: I believe this comment itself is not required now. Removed it.

cfg/rationalize.go
codechanges: Done

raj-prince (approved):

PR #3121, opened 2025-03-25
Use new writer flush before reads for zonal buckets with streaming writes.

meet2mky:
### Description
Implicitly Call Zonal Bucket writer.Flush() before read operation to ensure writes using streaming_writes are available for read from zonal buckets.

http://b/406160016

### Testing details
1. Manual - NA
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -108,10 +108,11 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
 	fh.inode.Lock()
-	err = fh.inode.SyncUsingBufferedWriteHandler()
+	// Ensure all pending writes to Zonal Buckets are flushed before issuing a read.
+	err = fh.inode.SyncPendingBufferedWrites()
 	if err != nil {
 		fh.inode.Unlock()
-		err = fmt.Errorf("fh.inode.SyncUsingBufferedWriteHandler: %w", err)
+		err = fmt.Errorf("fh.inode.SyncPendingBufferedWrites: %w", err)
 		return
 	}
 	err = fh.tryEnsureReader(ctx, sequentialReadSizeMb)
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -108,6 +108,12 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
 	fh.inode.Lock()
+	err = fh.inode.SyncUsingBufferedWriteHandler()
+	if err != nil {
+		fh.inode.Unlock()
+		err = fmt.Errorf("fh.inode.SyncUsingBufferedWriteHandler: %w", err)
+		return
+	}
 	err = fh.tryEnsureReader(ctx, sequentialReadSizeMb)
 	if err != nil {
 		fh.inode.Unlock()
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -378,9 +378,9 @@ func (f *FileInode) Source() *gcs.MinObject {
 func (f *FileInode) SourceGenerationIsAuthoritative() bool {
 	// Source generation is authoritative if:
 	//   1.  No pending writes exists on the inode (both content and bwh are nil).
-	//   2.  The bucket is zonal and there is no pending content write and
-	// 		 SyncUsingBufferedWriteHandler has been called on the fileInode
-	// 		 if using streaming writes.
+	//   2.  The bucket is zonal and there is no pending content write.
+	// 	 3.  The bucket is zonal with streaming writes and SyncUsingBufferedWriteHandler
+	// 		 has been called on f within same Inode lock.
 	// TODO(b/406160290): Check if this can be improved.
 	return (f.content == nil && f.bwh == nil) || (f.bucket.BucketType().Zonal && f.content == nil)
 }
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -376,9 +376,11 @@ func (f *FileInode) Source() *gcs.MinObject {
 //
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) SourceGenerationIsAuthoritative() bool {
-	// When streaming writes are enabled, writes are done via bufferedWritesHandler(bwh).
-	// Hence checking both f.content & f.bwh to be nil
-	return f.content == nil && f.bwh == nil
+	// Source generation is authoritative if:
+	//   1.  No pending writes exists on the inode (both content and bwh are nil).
+	//   2.  The bucket is zonal and there is no pending content write and SyncUsingBufferedWriteHandler has been called on the f.
+	// TODO(b/406160290): Check if this can be improved.
+	return (f.content == nil && f.bwh == nil) || (f.bucket.BucketType().Zonal && f.content == nil)
 }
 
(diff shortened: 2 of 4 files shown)


internal/fs/inode/file_streaming_writes_test.go
```diff
@@ -142,6 +155,42 @@ func (t *FileStreamingWritesTest) createInode(fileName string, fileType string)
...
+}
+
+func (t *FileStreamingWritesTest) TestSyncUsingBufferedWritesForNonZonalBuckets() {
+	assert.NoError(t.T(), t.in.Write(t.ctx, []byte("taco"), 0))
+
+	assert.NoError(t.T(), t.in.SyncUsingBufferedWriteHandler())
+	operations.ValidateObjectNotFoundErr(t.ctx, t.T(), t.bucket, t.in.Name().GcsObjectName())
+}
+
+func (t *FileStreamingWritesZonalBucketTest) TestSourceGenerationIsAuthoritativeReturnsTrueForZonalBuckets() {
```
vadlakondaswetha, line 175: dont mix fileStreamingWritesTests with ZB tests . Either keep all ZB tests at top or at bottom.

internal/fs/handle/file.go
```diff
@@ -108,6 +108,12 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
 	fh.inode.Lock()
+	err = fh.inode.SyncUsingBufferedWriteHandler()
```
vadlakondaswetha, line 111: Add a comment that we need to flush all pending writes before issuing a read.

internal/fs/inode/file_streaming_writes_test.go
```diff
@@ -142,6 +155,42 @@ func (t *FileStreamingWritesTest) createInode(fileName string, fileType string)
 // Tests
 ////////////////////////////////////////////////////////////////////////
 
+func (t *FileStreamingWritesTest) TestSourceGenerationIsAuthoritativeReturnsFalseForNonZonalBuckets() {
```
vadlakondaswetha, line 158: its not clear when there are no pending writes why is this false?

internal/fs/inode/file.go
```diff
@@ -376,9 +376,12 @@ func (f *FileInode) Source() *gcs.MinObject {
 //
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) SourceGenerationIsAuthoritative() bool {
-	// When streaming writes are enabled, writes are done via bufferedWritesHandler(bwh).
-	// Hence checking both f.content & f.bwh to be nil
-	return f.content == nil && f.bwh == nil
+	// Source generation is authoritative if:
+	//   1.  No pending writes exists on the inode (both content and bwh are nil).
+	//   2.  The bucket is zonal with non-streaming writes and there is no pending content write.
+	//   3.  The bucket is zonal with streaming writes and SyncUsingBufferedWriteHandler has been called on f within same Inode lock.
```
ashmeenkaur, line 382: Instead of this point(3) can you add this as a requirement of this function for now at the top of this function (documentation) along with the TOOD bug.

internal/fs/inode/file.go
```diff
@@ -642,6 +645,27 @@ func (f *FileInode) flushUsingBufferedWriteHandler() error {
 	return nil
 }
 
+// Helper function to sync buffered writes, for zonal buckets it flushes the writer
+// for writes to be available for read. This operation is no-op when BWH is nil.
+//
+// LOCKS_REQUIRED(f.mu)
+func (f *FileInode) SyncUsingBufferedWriteHandler() error {
```
ashmeenkaur, line 652: Nit:```suggestion
// SyncPendingBufferedWrites flushed any pending writes on the bwh to GCS.
// It is a no-op when bwh is nil.
// LOCKS_REQUIRED([f.mu](http://f.mu/)) 
func (f *FileInode) SyncPendingBufferedWrites() error {
```

File: internal/fs/inode/file_streaming_writes_test.go (hunk 1 above)
Comment (ashmeenkaur) [reply internal/fs/inode/file_streaming_writes_test.go:175]: +1, maybe a good idea to create a separate file for zb tests

Comment (meet2mky) [reply internal/fs/inode/file_streaming_writes_test.go]: Added Zonal buckets at the top and separated using a block comment. We can separate it to a different file once we add more composite tests for zb.

File: internal/fs/inode/file_streaming_writes_test.go (hunk 3 above)
Comment (meet2mky) [reply internal/fs/inode/file_streaming_writes_test.go]: This was because the t.in.bwh was initiated in the setup. Fixed it. Thanks!

File: internal/fs/inode/file.go (hunk 4 above)
Comment (meet2mky) [reply internal/fs/inode/file.go]: Done

File: internal/fs/handle/file.go (hunk 2 above)
Comment (meet2mky) [reply internal/fs/handle/file.go]: Done

File: internal/fs/inode/file.go [hunk 6]
```diff
@@ -376,9 +376,12 @@ func (f *FileInode) Source() *gcs.MinObject {
 //
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) SourceGenerationIsAuthoritative() bool {
-	// When streaming writes are enabled, writes are done via bufferedWritesHandler(bwh).
-	// Hence checking both f.content & f.bwh to be nil
-	return f.content == nil && f.bwh == nil
+	// Source generation is authoritative if:
+	//   1.  No pending writes exists on the inode (both content and bwh are nil).
+	//   2.  The bucket is zonal with non-streaming writes and there is no pending content write.
```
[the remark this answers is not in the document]
meet2mky: Done

internal/fs/inode/file.go
meet2mky: Done

internal/fs/inode/file.go
```diff
@@ -642,6 +646,27 @@ func (f *FileInode) flushUsingBufferedWriteHandler() error {
 	return nil
 }
 
+// SyncPendingBufferedWrites flushed any pending writes on the bwh to GCS.
+// It is a no-op when bwh is nil.
+//
+// LOCKS_REQUIRED((f.mu)
```
ashmeenkaur, line 652: Nit: sorry for the typo
```suggestion
// SyncPendingBufferedWrites flushes any pending writes on the bwh to GCS.
// It is a no-op when bwh is nil.
//
// LOCKS_REQUIRED(f.mu)
```

internal/fs/inode/file_test.go
```diff
@@ -173,6 +173,29 @@ func (t *FileTest) TestInitialSourceGeneration() {
...
+
+func (t *FileTest) TestSourceGenerationIsAuthoritativeReturnsFalseAfterWrite() {
+	assert.NoError(t.T(), t.in.Write(t.ctx, []byte("taco"), 0))
+
+	assert.False(t.T(), t.in.SourceGenerationIsAuthoritative())
+}
+
+func (t *FileTest) TestSyncPendingBufferedWritesReturnsNilAndNoOpForNonStreamingWrites() {
+	contents, err := storageutil.ReadObject(t.ctx, t.bucket, t.in.Name().GcsObjectName())
+	assert.Equal(t.T(), t.initialContents, string(contents))
```
ashmeenkaur, line 188: I would use `require.NoError` before checking content. At other places also where we are checking for errors, we should use "require" instead of "assert" and it should be done before asserting on other values returned from the function.

vadlakondaswetha (dismissed):

ashmeenkaur (approved):

internal/fs/inode/file_test.go
meet2mky: Thanks, Fixed this at multiple places in both the test files.

meet2mky:
Manually verified that tests are passing, additional details added on bug.

PR #3123, opened 2025-03-26
create new inode on change in length

ashmeenkaur:
### Description
keep check of length during `LookupInode` call, particularly useful for zonal buckets where object length can change keeping the inode generation and metageneration same.
With the changes in this PR, we will now create a new inode on change in object length. 

b/405000262

### Testing details
1. Manual - NA
2. Unit tests - added
3. Composite tests - added
4. Integration tests - via KOKORO
Failed test passed on manual retry:
3U6YDXMHsmt8FHb


internal/fs/inode/inode.go
```diff
@@ -84,27 +84,36 @@ type GenerationBackedInode interface {
...
+// Note: Ordering matters here, latest represents object's latest generation and
+// current represents inode cached object's current generation.
+func (latest Generation) Compare(current Generation) int {
 	// Compare first on object generation number.
-	switch {
-	case g.Object < other.Object:
-		return -1
-
-	case g.Object > other.Object:
+	if latest.Object != current.Object {
```
vadlakondaswetha, line 96: why is -1 condition removed?

internal/fs/inode/inode.go
```diff
@@ -84,27 +84,36 @@ type GenerationBackedInode interface {
...
+	case latest.Metadata > current.Metadata:
+		return 1
+	}
+
+	// Break ties on object size.
+	switch {
+	case latest.Size > current.Size:
 		return 1
+	case latest.Size < current.Size:
+		// We ignore this case as little staleness is expected on the latest
```
vadlakondaswetha, line 114: default we are returning 0 why do we need this condition.

internal/fs/inode/inode.go
```diff
@@ -84,27 +84,36 @@ type GenerationBackedInode interface {
...
 		return -1
 
-	case g.Metadata > other.Metadata:
+	case latest.Metadata > current.Metadata:
+		return 1
+	}
+
+	// Break ties on object size.
+	switch {
+	case latest.Size > current.Size:
```
vadlakondaswetha, line 111: can you add a comment that this happens only in ZB in append flow.

internal/fs/inode/file.go
```diff
@@ -385,8 +385,17 @@ func (f *FileInode) SourceGenerationIsAuthoritative() bool {
...
 // LOCKS_REQUIRED(f)
 func (f *FileInode) SourceGeneration() (g Generation) {
+	g.Size = f.src.Size
 	g.Object = f.src.Generation
 	g.Metadata = f.src.MetaGeneration
+	// If bwh is not nil, it's size takes precedence as that is being actively
+	// written to GCS.
+	// Since temporary file does not write to GCS on the go, it's size is not
+	// used as source object's size.
+	if f.bwh != nil {
```
vadlakondaswetha, line 395: why can't we call f.getAttributes directly?

vadlakondaswetha (commented):
yet to review unit test files

internal/fs/inode/inode.go
ashmeenkaur: GCS does not guarantee that generation numbers increase for successive versions, only that each new version has a unique generation number.
Ref - [cloud doc](https://cloud.google.com/storage/docs/metadata#generation-number:~:text=There%20is%20no%20guarantee%20that%20generation%20numbers%20increase%20for%20successive%20versions%2C%20only%20that%20each%20new%20version%20has%20a%20unique%20generation%20number.)

internal/fs/inode/inode.go
```diff
@@ -84,27 +84,36 @@ type GenerationBackedInode interface {
...
-
-	case g.Object > other.Object:
+	if latest.Object != current.Object {
 		return 1
 	}
 
 	// Break ties on meta-generation.
 	switch {
-	case g.Metadata < other.Metadata:
+	case latest.Metadata < current.Metadata:
```
meet2mky, line 102: Is it possible for an object to have cached metageneration number greater than latest metageneration on GCS? Metageneration increases from 1 so in gcsfuse inode cache we can only have less metageneration not greater, no?

internal/storage/fake/bucket.go
```diff
@@ -391,13 +392,31 @@ func preconditionChecks(b *bucket, req *gcs.CreateObjectRequest, contents []byte
...
+			existingObj.data = contents
+			existingObj.metadata.Size = uint64(len(contents))
+			if finalize {
+				existingObj.finalized = true
+			}
+			return &existingObj.metadata, nil
+		}
+	}
+
+	// Otherwise create new object and overwrite.
```
meet2mky, line 414: Nit: Please update comment, both comments are saying create object.

internal/fs/inode/file.go
ashmeenkaur: f.getAttributes will also return f.content (temporary file) size which is incorrect IMO.
Consider this edge case:
- GCSFuse temp file size is ahead of unfinalized object size on GCS (not yet uploaded).
- Some other process also writes to the same unfinalized object.

At the time of clobbered check, we will end up comparing GCS object length with temp file length (gcs object length < temp file length) and not consider it clobbered. While in reality, it was actually clobbered.

internal/fs/inode/inode.go
ashmeenkaur: I don't think this is possible (ideally) but since this change is only refactoring (receiver name change), I would not change in this PR.

internal/fs/inode/inode.go
```diff
@@ -84,28 +84,41 @@ type GenerationBackedInode interface {
...
 	Metadata int64
+	Size     uint64
 }
 
-// Compare returns -1, 0, or 1 according to whether g is less than, equal to, or greater
-// than other.
-func (g Generation) Compare(other Generation) int {
+// Compare returns -1, 0, or 1 according to whether src is less than, equal to,
+// or greater than existing.
+// Note: Ordering matters here, latest represents object's latest generation and
```
vadlakondaswetha, line 92: update the comment. latest represents the object fetched from GCS

internal/fs/inode/inode.go
```diff
@@ -84,28 +84,41 @@ type GenerationBackedInode interface {
...
 
+	// Break ties on object size.
+	// Because objects in zonal buckets can be appended without altering their
+	// generation or metageneration, the following case applies exclusively to
+	// zonal buckets.
+	if latest.Size > current.Size {
+		return 1
+	}
+	// We ignore latest.Size < current.Size case as little staleness is expected
+	// on the latest object's size.
```
vadlakondaswetha, line 121: latest object's size -> GCS object's size

internal/fs/append_zb_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/setup"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+type AppendsZBTest struct {
```
vadlakondaswetha, line 36: as discussed offline, these tests are not mimicking the user behaviour. Please add integration tests directly instead of composite tests.

vadlakondaswetha (approved):

PR #3139, opened 2025-04-02
add unfinalized object operations tests

ashmeenkaur:
### Description
Add e2e tests for operations on unfinalized object.

b/407959986

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - added

No

### DIFF
--- a/tools/integration_tests/unfinalized_object/setup_test.go
+++ b/tools/integration_tests/unfinalized_object/setup_test.go
@@ -1,0 +1,55 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package unfinalized_object
+
+import (
+	"log"
+	"os"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/mounting/static_mounting"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/setup"
+)
+
+const (
+	testDirName    = "UnfinalizedObjectTest"
+	onlyDirMounted = "OnlyDirMountedUnfinalizedObject"
+)
+
+var (
+	mountFunc func([]string) error
+)
+
+////////////////////////////////////////////////////////////////////////
+// TestMain
+////////////////////////////////////////////////////////////////////////
+
+func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+
+	// To run mountedDirectory tests, we need both testBucket and mountedDirectory
+	// flags to be set, as operations tests validates content from the bucket.
+	if setup.AreBothMountedDirectoryAndTestBucketFlagsSet() {
+		setup.RunTestsForMountedDirectoryFlag(m)
+	}
+
+	// Set up test directory.
+	setup.SetUpTestDirForTestBucketFlag()
+
+	log.Println("Running static mounting tests...")
+	mountFunc = static_mounting.MountGcsfuseWithStaticMounting
+	successCode := m.Run()
+	os.Exit(successCode)
+}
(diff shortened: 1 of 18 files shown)


tools/integration_tests/unfinalized_object/setup_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"os"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/mounting/static_mounting"
+	"github.com/googlecloudplatform/gcsfuse/v2/tools/integration_tests/util/setup"
+)
+
+const (
+	testDirName    = "UnfinalizedObjectTest"
+	onlyDirMounted = "OnlyDirMountedUnfinalizedObject"
```
meet2mky, line 28: Nit: It is not used in the test please remove.

tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (t *unfinalizedObjectOperations) TeardownTest() {}
+
+////////////////////////////////////////////////////////////////////////
+// Test scenarios
+////////////////////////////////////////////////////////////////////////
+
+func (t *unfinalizedObjectOperations) TestUnfinalizedObjectCreatedOutsideOfMountReports0Size() {
+	fileName := path.Base(t.T().Name()) + setup.GenerateRandomString(5)
```
meet2mky, line 56: Nit: We can move this to SetupTest. Same for other tests

tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
```diff
@@ -0,0 +1,176 @@
...
+		err := closeStorageClient()
+		if err != nil {
+			t.Errorf("closeStorageClient failed: %v", err)
+		}
+	}()
+
+	// Run tests for mounted directory if the flag is set.
+	if setup.AreBothMountedDirectoryAndTestBucketFlagsSet() {
+		suite.Run(t, ts)
+		return
```
meet2mky, line 159: This is redundant, If the mounted Dir test run from the TestMain then Mount and Unmount operations become no-op so current setup should work but I think it may fail for SaveGCSFuseLogs because we don't know gcsfuse logFile, Wdyt?

tools/integration_tests/util/client/gcs_helper.go
```diff
@@ -163,3 +165,18 @@ func GetCRCFromGCS(objectPath string, ctx context.Context, storageClient *storag
 	}
 	return attr.CRC32C, nil
 }
+
+func CreateUnfinalizedObject(ctx context.Context, t *testing.T, client *storage.Client, object string, size int64) *storage.Writer {
+	writer, err := AppendableWriter(ctx, client, object, storage.Conditions{})
+	require.NoError(t, err)
+
+	writeOffset, err := writer.Write([]byte(setup.GenerateRandomString(int(size))))
```
meet2mky, line 173: I think writer returns the bytes wrote and not offset, offset is only returned from Flush operation.
```
bytesWritten, err := writer.Write([]byte(setup.GenerateRandomString(int(size))))
```

tools/integration_tests/util/client/storage_client.go
```diff
@@ -381,3 +381,19 @@ func CopyFileInBucket(ctx context.Context, storageClient *storage.Client, srcfil
 		log.Fatalf("Error while copying file in bucket: %v", err)
 	}
 }
+
+func AppendableWriter(ctx context.Context, client *storage.Client, object string, precondition storage.Conditions) (*storage.Writer, error) {
```
meet2mky, line 385: Is this required? NewWriter does create appendable writer if supplied bucket is zonal.

meet2mky (commented):
Added some comments, PTAL.

tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
meet2mky: I think, We can add a check for mountedDir `SaveGCSFuseLogFileInCaseOfFailure` method similar to what we do in unmount methods.
```
func SaveGCSFuseLogFileInCaseOfFailure(tb testing.TB) {
	if !tb.Failed() || *mountedDirectory != "" {
		return
	}
	SaveLogFileAsArtifact(LogFile(), GCSFuseLogFilePrefix+strings.ReplaceAll(tb.Name(), "/", "_")+GenerateRandomString(5))
}
```

tools/integration_tests/util/client/storage_client.go
ashmeenkaur: Yes IMO this will still be useful.. in case we want to test scenarios where we don't want to flush the writer etc.

tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
ashmeenkaur: this is to skip steps like building GCSFuse package and running the same tests repeatedly when more flags are added to the flagset so that the test suite is run only on the provided mounted directory once and returned from there.

ashmeenkaur:
Facing some trouble rebasing this PR. Opened a new PR #3204

PR #3140, opened 2025-04-02
Add flag to guard metric renames

kislaykishore:
* The hidden flag allows one to switch between the existing metric renames and the new proposed ones.
* This flag is not supposed to be exposed to the end user but will allow to decouple GKE with GCSFuse metrics.

b/406112147

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -613,6 +613,7 @@
   usage: "Expose Prometheus metrics endpoint on this port and a path of /metrics."
   default: "0"
 
+
 - config-path: "metrics.stackdriver-export-interval"
   flag-name: "stackdriver-export-interval"
   type: "duration"
@@ -622,6 +623,13 @@
   default: "0s"
   deprecated: true
   deprecation-warning: "Please use --cloud-metrics-export-interval-secs instead."
+
+- config-path: "metrics.use-new-names"
+  flag-name: "metrics-use-new-names"
+  type: "bool"
+  usage: "Use the new metric names."
+  default: false
+  hide-flag: true
 
 - config-path: "monitoring.experimental-opentelemetry-collector-address"
   flag-name: "experimental-opentelemetry-collector-address"
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -925,6 +925,14 @@ func TestArgsParsing_MetricsFlags(t *testing.T) {
 				EnableOtel:                     true,
 			},
 		},
+		{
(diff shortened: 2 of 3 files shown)


cfg/params.yaml
```diff
@@ -613,6 +613,7 @@
   usage: "Expose Prometheus metrics endpoint on this port and a path of /metrics."
   default: "0"
 
+
```
gargnitingoogle, line 616: nit: this newline seems stray.

cfg/params.yaml
```diff
@@ -623,6 +624,13 @@
   deprecated: true
   deprecation-warning: "Please use --cloud-metrics-export-interval-secs instead."
 
+- config-path: "metrics.use-new-names"
+  flag-name: "metrics-use-new-names"
+  type: "bool"
+  usage: "Use the new metric names."
```
gargnitingoogle, line 630: nit: Should we give some examples here, or point to a link or something, to make this help message more helpful ?

cmd/root_test.go
```diff
@@ -925,6 +925,14 @@ func TestArgsParsing_MetricsFlags(t *testing.T) {
 				EnableOtel:                     true,
 			},
 		},
+		{
+			name: "use_new_metric_names",
+			args: []string{"gcsfuse", "--metrics-use-new-names=true", "abc", "pqr"},
```
gargnitingoogle, line 930: I think we should add another test case with `args: []string{"gcsfuse", "--metrics-use-new-names=true", "--enable-otel=false", "abc", "pqr"},` for higher coverage.

gargnitingoogle (changes requested):
One minor request to add a new test case.

cmd/root_test.go
kislaykishore: Per go/gcs-fuse-coding-guidelines#adding-new-param-in-gcsfuse - "Add one test-case in [root_test.go](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/cmd/root_test.go) to verify that the flag works - no need to test for different scenarios; a single test for a happy-case is sufficient.

cfg/params.yaml
kislaykishore: As discussed offline, this usage is not gonna show up in `--help`. So, it's fine to ignore.

charith87 (approved):

gargnitingoogle (approved):
LGTM

PR #3149, opened 2025-04-03
Timeout based closure for inactive range reader

raj-prince:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


raj-prince:
Implemented.

PR #3154, opened 2025-04-07
[Random Reader Refactoring] Initialise interfaces and req-resp struct

Tulsishah:
### Description
1. Interface declaration.
2. Struct for req and resp.

[b/408957421](https://b.corp.google.com/issues/408957421)

### Testing details
1. Manual - NA
3. Unit tests - NA
4. Integration tests - NA


internal/gcsx/reader.go
```diff
@@ -0,0 +1,85 @@
...
+// FallbackToAnotherReaderError is returned when data could not be retrieved
+// from the current reader, indicating that the caller should attempt to fall back
+// to an alternative reader.
+type FallbackToAnotherReaderError struct{}
+
+func (e *FallbackToAnotherReaderError) Error() string {
+	return "fallback to another reader is required."
+}
+
+var FallbackToAnotherReader = &FallbackToAnotherReaderError{}
```
raj-prince, line 32: Why not simply?

var FallbackToAnotherReader = errors.New("fallback to another reader is required.)

As we don't need additional data like, error code in the custom error.

internal/gcsx/reader.go
```diff
@@ -0,0 +1,85 @@
...
+// It includes the actual data read and its size.
+type ObjectRead struct {
+	// DataBuf contains the bytes read from the object.
+	DataBuf []byte
+
+	// Size indicates how many bytes were read into DataBuf.
+	Size int
+}
+
+// Reader is the base interface for all logical readers.
```
raj-prince, line 56: We can remove this comment, as interface is self-explanatory.

raj-prince (approved):

internal/gcsx/reader.go
```diff
@@ -0,0 +1,80 @@
...
+	// Offset specifies the starting position in the object from where data should be read.
+	Offset int64
+
+	// This determines GCS range request.
+	EndOffset int64
+}
+
+// ObjectRead represents the response returned as part of a ReadAt call.
+// It includes the actual data read and its size.
+type ObjectRead struct {
```
vadlakondaswetha, line 43: rename this to something like GCSReaderResponse

internal/gcsx/reader.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+)
+
+// FallbackToAnotherReader is returned when data could not be retrieved
+// from the current reader, indicating that the caller should attempt to fall back
+// to an alternative reader.
+var FallbackToAnotherReader = errors.New("fallback to another reader is required")
+
+// GCSReaderReq represents the request parameters needed to read a data from a GCS object.
+type GCSReaderReq struct {
```
vadlakondaswetha, line 30: Please dont use the short names (Req). Either use the full name or omit the work entirely. rename this to GCSReaderRequest.

internal/gcsx/reader.go
Tulsishah: This is not GCSReaderResponse right? File cache even Adaptive prefetch will use this as returned data.

PR #3166, opened 2025-04-08
MRD recreation in case of stream closed error

anushka567:
### Description
The bidi stream used to communicate with the GCS server, can get disconnected due to various reasons. If go-sdk determines the reason to be retryable, it recreates the stream otherwise the stream is closed and the MRD becomes unusable. If the application still continues to use the MRD, it will continue to get the stream closed error.

This PR includes code changes where we make use of the Error() method exposed by Go-sdk,  which returns error in case the existing MRD has become unusable, this functionality is used to recreate the MRD for serving the subsequent read calls. This will help eliminate the need to close and reopen file handles in such scenarios.

**NOTE:**
The dependencies for this functionality will go out with the next go sdk release. Hence, tests will continue to fail on this PR.
I have developed against HEAD and confirmed the correct usage .

Update : go-sdk version has been upgraded already .
```
=== RUN   TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader
=== RUN   TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader/ValidMinObject
--- PASS: TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader/ValidMinObject (0.00s)
=== RUN   TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader/NilMinObject
--- PASS: TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader/NilMinObject (0.00s)
=== RUN   TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader/NilBucket
--- PASS: TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader/NilBucket (0.00s)
--- PASS: TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader (0.00s)
=== RUN   TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader_UnusableExistingMRDTriggersRecreation
--- PASS: TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader_UnusableExistingMRDTriggersRecreation (0.00s)
=== RUN   TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader_UsableExistingMRDPreventsRecreation
--- PASS: TestMRDWrapperTestSuite/Test_EnsureMultiRangeDownloader_UsableExistingMRDPreventsRecreation (0.00s)
```
### Testing details
1. Manual - NA
2. Unit tests - Added for scenarios where :
* Existing MRD is in unusable state and hence, the MRD is recreated. 
* Exisiting MRD is in usable state, thus, no need for MRD recreation.
3. Integration tests - NA


internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -162,17 +162,22 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader() (err
...
-			Generation:     mrdWrapper.object.Generation,
-			ReadCompressed: mrdWrapper.object.HasContentEncodingGzip(),
-		})
-		if err == nil {
-			// Updating mrdWrapper.Wrapped only when MRD creation was successful.
-			mrdWrapper.Wrapped = mrd
+	// Create the MRD if it does not exist.
+	// In case the existing MRD is unusable due to closed stream, recreate the MRD.
+	if mrdWrapper.Wrapped == nil || mrdWrapper.Wrapped.Error() != nil {
+		mrdWrapper.mu.Lock()
```
abhishek10004, line 168: We should use a different mutex here (& not the existing one which is used for reference counting)

vadlakondaswetha: it should be the same mutex since we are setting mrdWrapper.Wrapper=nil when the refCount goes to 0 under the same mutext.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -162,17 +162,22 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader() (err
...
+		defer mrdWrapper.mu.Unlock()
+		if mrdWrapper.Wrapped == nil || mrdWrapper.Wrapped.Error() != nil {
```
vadlakondaswetha, line 170: nit: add a comment

Checking if the mrdWrapper state is same after taking the lock.

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -287,42 +287,81 @@ func (t *mrdWrapperTest) Test_EnsureMultiRangeDownloader() {
...
+				err := t.mrdWrapper.ensureMultiRangeDownloader()
+				if tc.err == nil {
+					assert.NoError(t.T(), err)
+					assert.NotNil(t.T(), t.mrdWrapper.Wrapped)
+				} else {
+					assert.Error(t.T(), err)
+					assert.EqualError(t.T(), err, tc.err.Error())
+					assert.Nil(t.T(), t.mrdWrapper.Wrapped)
+			}
+			}else{
```
vadlakondaswetha, line 353: we should parameterized tests only if the test code is all same. in this specific case 2 new tests which are added are going into else clause. No need to force fit them here. please create seperate tests.

also there is again if-else inside else. please create seperat tests.

internal/storage/debug_bucket.go
```diff
@@ -340,6 +340,13 @@ func (dmrd *debugMultiRangeDownloader) Wait() {
 	dmrd.wrapped.Wait()
 }
 
+func (dmrd *debugMultiRangeDownloader) Error() (err error) {
+	id, desc, start := dmrd.bucket.startRequest("MultiRangeDownloader.Error()")
+	defer dmrd.bucket.finishRequest(id, desc, start, &err)
```
vadlakondaswetha, line 345: do we really need this log?
@abhishek10004

internal/gcsx/multi_range_downloader_wrapper.go
anushka567: Resolving as the implementation is the same as the one ultimately expected.

internal/storage/debug_bucket.go
abhishek10004: Not really. We should remove it as it will 2 extra log statements per read call (making debugging experience worse) & in any case if there are errors, we will try to recreate MRD and will log the failure in recreation.

abhishek10004 (approved):
LGTM

ashmeenkaur (approved):
Relying on @abhishek10004's review

PR #3178, opened 2025-04-10
Correcting the periodic test failure for grpc_validation package

anushka567:
### Description
This PR fixes a subtle bug where the testRegion variable, intended to be shared across test files, was not being properly initialized due to variable shadowing within TestMain().
Due to this , the regions for the testcases were not being correctly configured , resulting in test failures.
Confirmed successful run manually:
image### Link to the issue in case of a bug fix.

### Testing details
1. Manual - On GCE VM
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/grpc_validation/setup_test.go
+++ b/tools/integration_tests/grpc_validation/setup_test.go
@@ -79,7 +79,7 @@ func findTestExecutionEnvironment(ctx context.Context) (string, error) {
 	if v, exists := attrs.Value("gcp.gce.instance.hostname"); exists && strings.Contains(strings.ToLower(v.AsString()), cloudtopProd) {
 		return cloudtopProd, nil
 	}
-	if v, exists := attrs.Value("cloud.region"); exists {
+	if v, exists := attrs.Value("cloud.availability_zone"); exists {
 		return v.AsString(), nil
 	}
 	return "", nil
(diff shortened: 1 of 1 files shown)


tools/integration_tests/grpc_validation/setup_test.go
```diff
@@ -79,7 +79,7 @@ func findTestExecutionEnvironment(ctx context.Context) (string, error) {
 	if v, exists := attrs.Value("gcp.gce.instance.hostname"); exists && strings.Contains(strings.ToLower(v.AsString()), cloudtopProd) {
 		return cloudtopProd, nil
 	}
-	if v, exists := attrs.Value("cloud.region"); exists {
+	if v, exists := attrs.Value("cloud.availability_zone"); exists {
```
raj-prince, line 82: Could you please add a comment, how did you figure out, which keywords needs to check?

raj-prince: Also, just realized, although slight out of scope of this PR.

Could you please convert - single_region, multi_regions and gcp_project - in camelCase?

anushka567: This was an unrelated change `cloud.region` excludes zone information while `cloud.availability_zone` returns region in the format us-west1-b.
Since the logic for deciding the test regions utilizies regex matching and the method itself assumed that the test region would be in the latter format. Thus, updated.

anushka567: Updated with all changes.

ashmeenkaur (approved):

PR #3181, opened 2025-04-10
CloseFile should not exit the tests

ashmeenkaur:
### Description
CloseFile test helper logs a fatal error in case of failure which exits the tests. Due to this, sometimes GCSFuse failure logs are also not retained. Changing to use assert library instead so that logs are retained.

b/406133834

### Testing details
1. Manual - NA
2. Unit tests - NA

NA

### DIFF
--- a/tools/integration_tests/util/operations/file_operations.go
+++ b/tools/integration_tests/util/operations/file_operations.go
@@ -180,6 +180,7 @@ func CloseFiles(t *testing.T, files []*os.File) {
 	}
 }
 
+// Deprecated: please use CloseFileShouldNotThrowError instead.
 func CloseFile(file *os.File) {
 	if err := file.Close(); err != nil {
 		log.Fatalf("error in closing: %v", err)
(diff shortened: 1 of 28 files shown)


tools/integration_tests/util/operations/file_operations.go
```diff
@@ -601,10 +602,9 @@ func WriteAt(content string, offset int64, fh *os.File, t testing.TB) {
 	}
 }
 
-func CloseFileShouldNotThrowError(file *os.File, t *testing.T) {
-	if err := file.Close(); err != nil {
-		t.Fatalf("file.Close() for file %s: %v", file.Name(), err)
-	}
+func CloseFileShouldNotThrowError(t testing.TB, file *os.File) {
```
Tulsishah, line 605: Why are we not changing above function itself?

ashmeenkaur: It is still being used by other functions were testing.TB is not available, would require more refactoring. Planning to move away from that function incrementally.

Tulsishah (approved):

charith87 (approved):

PR #3183, opened 2025-04-10
add retries in case ssh fails

ashmeenkaur:
### Description
SSH sometimes fails causing KOKORO failures so adding retries.

b/401861276

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
+++ b/perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
@@ -37,9 +37,22 @@ function initialize_ssh_key () {
     # This is required to avoid issue: https://github.com/kyma-project/test-infra/issues/93
     for i in $(sudo gcloud compute os-login ssh-keys list | grep -v FINGERPRINT); do sudo gcloud compute os-login ssh-keys remove --key $i; done
 
-    # Requires running first ssh command with --quiet option to initialize keys.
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
```diff
@@ -37,9 +37,22 @@ function initialize_ssh_key () {
...
+      # Otherwise it prompts for yes and no.
+      if sudo gcloud compute ssh "$VM_NAME" --zone "$ZONE_NAME" --internal-ip --quiet --command "echo 'Running from VM'"; then
+        return 0 # Exit function on success.
+      else
+        sleep "$RETRY_DELAY"
+        attempt=$((attempt + 1))
+      fi
+    done
+
+    return 1 # Return error status if all retries fail
```
Tulsishah, line 55: Are we capturing this return code at caller? I think no (line#111)

ashmeenkaur: This explicit check isn't required. When set -e is active, the script will automatically terminate if a function returns 1. Here's a playground example demonstrating this: https://onlinegdb.com/xeXQa_1o3

ashmeenkaur:
seems like this change was merged in another PR, closing it! :D

PR #3188, opened 2025-04-11
fix install test for patch release

ashmeenkaur:
### Description
Installation tests were failing when running against patch releases. This was because the test script assumed the version being released was the most recent.
This PR updates the test script to no longer rely on the release version being the absolute latest, making it compatible with patch releases.

internal bug : b/409958930

### Testing details
1. Manual - manually ran on CD pipeline
2. Unit tests - NA
3. Integration tests - NA

NA

### DIFF
--- a/tools/cd_scripts/install_test.sh
+++ b/tools/cd_scripts/install_test.sh
@@ -20,6 +20,7 @@ set -x
 gsutil cp  gs://gcsfuse-release-packages/version-detail/details.txt .
 # Writing VM instance name to details.txt (Format: release-test-<os-name>)
 vm_instance_name=$(curl http://metadata.google.internal/computeMetadata/v1/instance/name -H "Metadata-Flavor: Google")
+to_release_version=$(sed '1q' details.txt | tr -d '\n')
 echo $vm_instance_name >> details.txt
 touch ~/logs.txt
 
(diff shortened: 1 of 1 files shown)


tools/cd_scripts/install_test.sh
```diff
@@ -20,6 +20,7 @@ set -x
 gsutil cp  gs://gcsfuse-release-packages/version-detail/details.txt .
 # Writing VM instance name to details.txt (Format: release-test-<os-name>)
 vm_instance_name=$(curl http://metadata.google.internal/computeMetadata/v1/instance/name -H "Metadata-Flavor: Google")
+to_release_version=$(sed '1q' details.txt | tr -d '\n')
```
Tulsishah, line 23: The release version in details.txt is in the format vX.Y.Z, for example, v2.2.0. If this is the case, shouldn't we only capture 2.2.0 as the to_release_version? If not, could you please provide a comment indicating the expected format of the first line in details.txt.

tools/cd_scripts/install_test.sh
```diff
@@ -114,10 +115,10 @@ fi
 
 gcsfuse --version |& tee version.txt
 installed_version=$(echo $(sed -n 1p version.txt) | cut -d' ' -f3)
-if grep -q $installed_version details.txt; then
-    echo "GCSFuse successfully upgraded to latest version $installed_version." &>> ~/logs.txt
+if [[ "$(printf '%s\n%s\n' "$to_release_version" "$installed_version" | sort -V | tail -n 1)" == "$installed_version" ]]; then
```
Tulsishah, line 118: Can we write small comment on what this line is doing?

ashmeenkaur: Added

tools/cd_scripts/install_test.sh
ashmeenkaur: It is of the format 2.2.0, added a comment mentioning the same.

Tulsishah (approved):

PR #3190, opened 2025-04-11
Clean up buckets on e2e failure

gargnitingoogle:
### Description
It addresses two problems.
1. Deletion of buckets created by E2E run script. The code is there in place, but it doesn't fully work in the following ways
   - It deletes objects but not the bucket itself, at least not if the test run fails, as I have many times gone back and deleted buckets manually afterwards.
3. The deletion code doesn't get called if the script terminates early because of some test or environment issue causing a crash.

It also logs the bucket deletions if and when they are deleted.

[b/409363794](http://b/409363794)

### Testing details
1. Manual - NA
4. Unit tests - NA
5. Integration tests - NA

### DIFF
--- a/tools/integration_tests/run_e2e_tests.sh
+++ b/tools/integration_tests/run_e2e_tests.sh
@@ -178,13 +178,6 @@ function delete_buckets_listed_in_file() {
 	fi
 }
 
-# Clean-up for this program, which is called whenever this program exits.
-# Args: None.
-function clean_up() {
-	local bucketNamesFile=${1}
-	delete_buckets_listed_in_file "${bucketNamesFile}"
-}
-
 function upgrade_gcloud_version() {
   sudo apt-get update
   # Upgrade gcloud version.
@@ -461,7 +454,7 @@ function main(){
   # buckets to be cleaned-up while exiting this program.
   bucketNamesFile=$(realpath ./bucketNames)"-"$(tr -dc 'a-z0-9' < /dev/urandom | head -c $RANDOM_STRING_LENGTH)
   # Delete all these buckets when the program exits.
-  trap "clean_up ${bucketNamesFile}" EXIT
+  trap "delete_buckets_listed_in_file ${bucketNamesFile}" EXIT
 
   set -e
 


tools/integration_tests/run_e2e_tests.sh
```diff
@@ -155,6 +155,45 @@ TEST_DIR_NON_PARALLEL_FOR_ZB=(
 # Create a temporary file to store the log file name.
 TEST_LOGS_FILE=$(mktemp)
 
+# Delete contents of the bucket, and then the bucket itself.
+# Args: <bucket>
+function clean_up_bucket() {
```
Tulsishah, line 160: Is it necessary to do it in two steps? Can't we directly delete the bucket?
```
gcloud  storage rm -r --verbosity=none gs://${bucket}
```

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -155,6 +155,45 @@ TEST_DIR_NON_PARALLEL_FOR_ZB=(
...
+		echo "Deleting bucket ${bucket} ..."
+		if ! gcloud -q --no-user-output-enabled storage buckets delete gs://${bucket}; then
+			>&2 echo "Failed to delete bucket ${bucket} !"
+		fi
+	fi
+}
+
+# Delete contents of the buckets (and then the buckets themselves) whose names are in the passed file.
+# Args: <bucket-names-file>
+function clean_up_buckets_in_file() {
```
Tulsishah, line 174: Can we use a different function name? This seems unclear to me. Also, please add a comment explaining that, during testing, we create buckets, store the names in a file or array, and then delete them upon exiting the test.

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -318,41 +357,42 @@ function run_e2e_tests_for_flat_bucket() {
   bucketPrefix="golang-grpc-test-gcsfuse-np-e2e-tests-"
   bucket_name_non_parallel=$(create_bucket $bucketPrefix)
   echo "Bucket name for non parallel tests: "$bucket_name_non_parallel
+  echo ${bucket_name_non_parallel}>>"${bucketNamesFile}"
```
Tulsishah, line 360: Can't we use array instead of file name?

gargnitingoogle: We can't use an array or any shell variable because a variable modified in a sub-process doesn't show up modified in the parent process as it won't work. This is what I started with as well.

In this case, the array/variable will be populated/updated inside the functions like `run_e2e_tests_for_flat_bucket` which are run in a subshell/sub-process from the `main` function in the script. So, any bucket-name added in the `run_e2e_tests_for_flat_bucket` array/variable won't be there in it when it is used in the main function or in the main script. The easiest reliable way of communicating information between processes in a shell script is either exit-codes or files on disk. Exit-codes can't be used to convey string AFAIK. Hence, the file.

tools/integration_tests/run_e2e_tests.sh
gargnitingoogle: I am going with 'delete_buckets_listed_in_file' for now. If you have a suggestion, please share.

tools/integration_tests/run_e2e_tests.sh
gargnitingoogle: The split was mainly for better readability, reusability. Anyway, I've taken your suggestion and merged them into one.

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -155,6 +155,45 @@ TEST_DIR_NON_PARALLEL_FOR_ZB=(
...
+# The name of a file containing the names of all the 
+# buckets to be cleaned-up while exiting this program.
+bucketNamesFile=$(realpath ./bucketNames)"-"$(tr -dc 'a-z0-9' < /dev/urandom | head -c $RANDOM_STRING_LENGTH)
+
+# Clean-up for this program, which is called whenever this program exits.
+# Args: None.
+function clean_up() {
+	clean_up_buckets_in_file "${bucketNamesFile}"
+}
+trap clean_up EXIT
```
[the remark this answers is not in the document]
gargnitingoogle: I am not sure if a trap placed inside a function is a good idea, as a a crash caused by anything else outside the function won't be caught by it. People usually put trap inside the script directly rather than inside functions. But it seems to work in this case because everything happens inside the main function. So, I've made this change.

Tulsishah (approved):

PR #3191, opened 2025-04-11
[Random reader refactoring] Create interfaces and mock structure for unit test

Tulsishah:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/common/mock_telemetry.go
+++ b/common/mock_telemetry.go
@@ -1,0 +1,73 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+
+	"github.com/stretchr/testify/mock"
+)
+
+type MockMetricHandle struct {
+	mock.Mock
+}
+
+func (m *MockMetricHandle) GCSReadBytesCount(ctx context.Context, inc int64) {
+	m.Called(ctx, inc)
+}
+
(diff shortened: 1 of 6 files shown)


internal/cache/file/cache_handle.go
```diff
@@ -27,6 +27,12 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 )
 
+type CacheHandleInterface interface {
+	Read(ctx context.Context, bucket gcs.Bucket, object *gcs.MinObject, offset int64, dst []byte) (n int, cacheHit bool, err error)
+	IsSequential(currentOffset int64) bool
+	Close() (err error)
+}
+
 type CacheHandle struct {
```
vadlakondaswetha, line 36: why are both interface and struct public

internal/cache/file/cache_handler.go
```diff
@@ -27,6 +27,12 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
 )
 
+type CacheHandlerInterface interface {
```
vadlakondaswetha, line 30: same here also why are interface and struct public?

common/mock_telemetry.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+
+	"github.com/stretchr/testify/mock"
+)
+
+type MockMetricHandle struct {
+	mock.Mock
+}
+
+func (m *MockMetricHandle) GCSReadBytesCount(ctx context.Context, inc int64) {
```
vadlakondaswetha, line 27: why do we need mock for metricHandle. how are we testing so far?

internal/cache/file/cache_handle.go
Tulsishah: Interfaces cannot be made private because they need to be accessible in the `gcsx` module for file cache reading. The struct was public according to the read cache design, but since we are introducing an interface, we can now make the struct private.

common/mock_telemetry.go
Tulsishah: Current tests are composite tests(build on fake bucket) and not unit tests. For writing unit tests we have to mock metric handler behaviour.

e.g.  We have to mock [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/gcsx/random_reader.go#L254) line to write unit test for `tryReadingFromFileCache` function.

PR #3198, opened 2025-04-14
Ensure correct size of min object in fileInode and Random Reader

meet2mky:
### Description
This PR is a fix to ensure correct size of minObject is reflected in fileInode and Random Reader for the Zonal objects with streaming writes.

b/409192356

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -193,8 +193,8 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb 
 	// If we already have a reader, and it's at the appropriate generation, we
 	// can use it. Otherwise we must throw it away.
 	if fh.reader != nil {
-		srcGen := fh.inode.SourceGeneration()
-		if fh.reader.IsValid(srcGen.Object, srcGen.Size) {
+		if fh.reader.Object().Generation == fh.inode.SourceGeneration().Object {
+			fh.reader.Object().Size = fh.inode.SourceGeneration().Size
 			return
 		}
 		fh.reader.Destroy()
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -194,6 +194,7 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb 
 	// can use it. Otherwise we must throw it away.
 	if fh.reader != nil {
 		if fh.reader.Object().Generation == fh.inode.SourceGeneration().Object {
+			// Update reader object size to source object size.
 			fh.reader.Object().Size = fh.inode.SourceGeneration().Size
 			return
 		}
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -199,8 +199,8 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb 
 		}
 		fh.reader.Destroy()
 		fh.reader = nil
-
 	}
+
 	// Attempt to create an appropriate reader.
 	rr := gcsx.NewRandomReader(fh.inode.Source(), fh.inode.Bucket(), sequentialReadSizeMb, fh.fileCacheHandler, fh.cacheFileForRangeRead, fh.metricHandle, &fh.inode.MRDWrapper)
 
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -194,18 +194,12 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb 
 	// can use it. Otherwise we must throw it away.
 	if fh.reader != nil {
 		srcGen := fh.inode.SourceGeneration()
-		if fh.reader.IsStale(srcGen.Object) {
-			fh.reader.Destroy()
-			fh.reader = nil
-		} else {
-			// If we know the reader is not stale, we should update the reader's object
-			// size to f.src.Size. For Zonal Objects with streaming writes f.src.Size changes
-			// even when generation of Object doesn't change.
-			if fh.inode.Bucket().BucketType().Zonal && fh.inode.IsUsingBWH() {
-				fh.reader.UpdateReaderObjectSizeToSrcSize(srcGen.Size)
-			}
+		if fh.reader.IsValid(srcGen.Object, srcGen.Size) {
 			return
 		}
+		fh.reader.Destroy()
+		fh.reader = nil
+
 	}
 	// Attempt to create an appropriate reader.
 	rr := gcsx.NewRandomReader(fh.inode.Source(), fh.inode.Bucket(), sequentialReadSizeMb, fh.fileCacheHandler, fh.cacheFileForRangeRead, fh.metricHandle, &fh.inode.MRDWrapper)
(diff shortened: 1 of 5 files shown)


internal/fs/handle/file.go
```diff
@@ -193,13 +193,20 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
...
-		if fh.reader.Object().Generation == fh.inode.SourceGeneration().Object {
+		srcGen := fh.inode.SourceGeneration()
+		if fh.reader.IsStale(srcGen.Object) {
+			fh.reader.Destroy()
+			fh.reader = nil
+		} else {
+			// If we know the reader is not stale, we should update the reader's object
+			// size to f.src.Size. For Zonal Objects with streaming writes f.src.Size changes
+			// even when generation of Object doesn't change.
+			if fh.inode.Bucket().BucketType().Zonal && fh.inode.IsUsingBWH() {
```
vadlakondaswetha, line 204: what happens if we dont have this 2nd check? I am trying to avoid too many conditions in the code.

vadlakondaswetha: actually thinking more why do we even want to do this for zonalBucket() why can't we just do it for all buckets and add a comment that it will not get triggered for regional bucket.

internal/fs/inode/file.go
```diff
@@ -673,6 +673,11 @@ func (f *FileInode) SyncPendingBufferedWrites() error {
 	if err != nil {
 		return fmt.Errorf("f.bwh.Sync(): %w", err)
 	}
+	// Update the f.src.Size to the current Size of object
+	// on GCS after flushing pending buffers for Zonal Buckets.
+	if f.bucket.BucketType().Zonal {
+		f.src.Size = uint64(f.bwh.WriteFileInfo().TotalSize)
```
vadlakondaswetha, line 679: what if f.bwh is nil?

internal/fs/inode/file.go
```diff
@@ -673,6 +673,11 @@ func (f *FileInode) SyncPendingBufferedWrites() error {
 	if err != nil {
 		return fmt.Errorf("f.bwh.Sync(): %w", err)
 	}
+	// Update the f.src.Size to the current Size of object
+	// on GCS after flushing pending buffers for Zonal Buckets.
+	if f.bucket.BucketType().Zonal {
```
vadlakondaswetha, line 678: again why do we need this check?

internal/fs/handle/file.go
```diff
@@ -193,13 +193,20 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
 	// If we already have a reader, and it's at the appropriate generation, we
 	// can use it. Otherwise we must throw it away.
 	if fh.reader != nil {
-		if fh.reader.Object().Generation == fh.inode.SourceGeneration().Object {
+		srcGen := fh.inode.SourceGeneration()
+		if fh.reader.IsStale(srcGen.Object) {
```
vadlakondaswetha, line 197: isn't this method doing the same thing as existing condition. why do we need a new method?

internal/gcsx/random_reader.go
```diff
@@ -77,12 +77,16 @@ type RandomReader interface {
...
 
-	// Return the record for the object to which the reader is bound.
-	Object() (o *gcs.MinObject)
-
 	// Clean up any resources associated with the reader, which must not be used
 	// again.
 	Destroy()
+
+	// Updates reader object size to f.src object size.
+	// It can only be called when reader is not stale.
```
vadlakondaswetha, line 85: who guarantees this - It can only be called when reader is not stale?

Also can we pass the entire Generation object here and this mehtod will take of updating size if generation is same.

internal/gcsx/random_reader.go
```diff
@@ -77,12 +77,16 @@ type RandomReader interface {
...
+	UpdateReaderObjectSizeToSrcSize(srcSize uint64)
```
vadlakondaswetha, line 86: rename this method to updateObjectSize

internal/fs/handle/file.go
meet2mky: If we remove this check
```
if fh.inode.Bucket().BucketType().Zonal && fh.inode.IsUsingBWH() {}
```
If we do so then for Non Zonal Objects with streaming writes we would be continuously updating reader ObjectSize to src.Size. Which is not what is visible on GCS but since we don't support reads. It may not matter wdyt? @vadlakondaswetha 

We can completely remove this condition here and also similarly in the SyncPendingBufferedWrites.

internal/fs/inode/file.go
meet2mky: Replied in: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3198/files/481f82d9dc85677f43ec1d6d51c738fc64a0bdc4#r2046594963

internal/fs/inode/file.go
meet2mky: Lock on `f.mu` is required to for this method and condition for not nil `f.bwh` is checked at the top of method.

internal/fs/handle/file.go
meet2mky: Currently fh.reader is exposing it's minObject through Object() method which is not required as it's only used for checking if the reader is stale or not. I think it's better to not expose reader object unless required to prevent accidental modifications.

internal/gcsx/random_reader.go
meet2mky: Yes, That can be done, I will do the changes.

internal/gcsx/random_reader.go
meet2mky: Ack will do!

internal/fs/handle/file.go
meet2mky: I have removed the conditions, PTAL.

internal/fs/handle/file.go
```diff
@@ -193,7 +193,8 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
 	// If we already have a reader, and it's at the appropriate generation, we
 	// can use it. Otherwise we must throw it away.
 	if fh.reader != nil {
-		if fh.reader.Object().Generation == fh.inode.SourceGeneration().Object {
+		srcGen := fh.inode.SourceGeneration()
+		if fh.reader.IsValid(srcGen.Object, srcGen.Size) {
```
meet2mky, line 197: Note: I can't pass whole `fh.inode.SourceGeneration()` to `fh.reader.IsValid` as it creates a cyclic dependency on imports.

internal/fs/handle/file.go
vadlakondaswetha: Agreed, but dont change it now. It will change again as part of refactoring.
just keep the same condition and update the size below it

internal/fs/inode/file.go
```diff
@@ -673,6 +673,9 @@ func (f *FileInode) SyncPendingBufferedWrites() error {
 	if err != nil {
 		return fmt.Errorf("f.bwh.Sync(): %w", err)
 	}
+	// Update the f.src.Size to the current Size of object
+	// on GCS after flushing pending buffers.
+	f.src.Size = uint64(f.bwh.WriteFileInfo().TotalSize)
```
vadlakondaswetha, line 678: I think we will end up updating the whole object f.src with your next fix

meet2mky: I don't think so, Since they are going to fix creating empty unfinalized gcs object. We would flush immediately after creating writer and set the minObject accordingly.

vadlakondaswetha (approved):

PR #3201, opened 2025-04-15
Make MRDWrapper store copy of MinObject from inode to prevent data race.

meet2mky:
### Description
Holding a direct reference to an inode's MinObject within MRDWrapper could lead to unsynchronized access and potential data races on inode's minObject mutex-protected fields. To prevent this, MRDWrapper will now store a reference to a separate copy of the MinObject.

b/410751316

### Testing details
1. Manual - NA
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -160,7 +160,7 @@ func NewFileInode(
 		globalMaxWriteBlocksSem: globalMaxBlocksSem,
 	}
 	var err error
-	f.MRDWrapper, err = gcsx.NewMultiRangeDownloaderWrapper(bucket, &f.src)
+	f.MRDWrapper, err = gcsx.NewMultiRangeDownloaderWrapper(bucket, &minObj)
 	if err != nil {
 		logger.Errorf("NewFileInode: Error in creating MRDWrapper %v", err)
 	}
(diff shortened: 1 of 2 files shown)


internal/fs/inode/file.go
```diff
@@ -905,7 +905,7 @@ func (f *FileInode) updateInodeStateAfterSync(minObj *gcs.MinObject) {
 // Updates the min object stored in MRDWrapper corresponding to the inode.
 // Should be called when minObject associated with inode is updated.
 func (f *FileInode) updateMRDWrapper() {
-	err := f.MRDWrapper.SetMinObject(&f.src)
+	err := f.MRDWrapper.SetMinObject(f.Source())
```
abhishek10004, line 908: Please ensure updateMRDWrapper is called whenever we update the minObject in inode

internal/fs/inode/file_test.go
```diff
@@ -442,8 +443,10 @@ func (t *FileTest) TestWriteThenSync() {
 			// The generation should have advanced.
 			assert.Less(t.T(), t.backingObj.Generation, t.in.SourceGeneration().Object)
 
+			// Validate MinObject in inode and MRDWrapper points to different copy of MinObject.
+			assert.NotSame(t.T(), &t.in.src, t.in.MRDWrapper.GetMinObject())
 			// Validate MinObject in MRDWrapper is same as the MinObject in inode.
-			assert.Same(t.T(), &t.in.src, t.in.MRDWrapper.GetMinObject())
+			assert.True(t.T(), reflect.DeepEqual(&t.in.src, t.in.MRDWrapper.GetMinObject()))
```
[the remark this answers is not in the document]
meet2mky: yes, it works, Updated! Thanks!

internal/fs/inode/file.go
meet2mky: MRDWrapper only required Name & Generation field of the object and whenever these fields changes we call `updateMRDWrapper`.

abhishek10004 (approved):
LGTM

ashmeenkaur (approved):

PR #3202, opened 2025-04-15
[Random reader refactoring] File cache reader implementation - 1

Tulsishah:
### Description
1. Implemented the ReadAt method for the file cache reader, utilizing the helper function tryReadingFromCache.
2. Unit tests cover the following scenarios:  
   i. Nil file cache handler 
   ii. Cache hit and data population from cache on subsequent reads 
   iii. Cache hit and data population from cache on subsequent reads (next offset from the first read) 
   iv. Sequential range read 
   v. When cacheRangeRead is false, an ErrFallbackToGCS error is returned, triggering a fallback to another reader. 
   vi. When the cacheRangeRead flag is true, the file is downloaded into the cache. 
   vii. Receiving ErrFallbackToGCS error. 
   viii. Receiving ErrInvalidFileDownloadJob error.
   ix. Sequential to Random SubsequentRead Offset less than previous
   x. Sequential to Random SubsequentRead Offset More than ReadChunkSize

**Please note that I will add more unit tests in a follow-up pull request to manage the size of this current PR.**

[b/411086522](https://b.corp.google.com/issues/411086522)

### Testing details
1. Manual - NA
4. Unit tests - Added
5. Integration tests - NA

### DIFF
--- a/internal/gcsx/file_cache_reader.go
+++ b/internal/gcsx/file_cache_reader.go
@@ -117,7 +117,7 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 		if isSequential {
 			readType = util.Sequential
 		}
-		captureFileCacheMetrics(ctx, fc.metricHandle, readType, bytesRead, hit, executionTime)
+		fc.captureFileCacheMetrics(ctx, fc.metricHandle, readType, bytesRead, hit, executionTime)
 	}()
 
 	// Create fileCacheHandle if not already.
--- a/internal/gcsx/file_cache_reader.go
+++ b/internal/gcsx/file_cache_reader.go
@@ -85,13 +85,12 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 
 	// By default, consider read type random if the offset is non-zero.
 	isSequential := offset == 0
-	requestID := uuid.New()
 
 	var handleID uint64
 	if readOp, ok := ctx.Value(ReadOp).(*fuseops.ReadFileOp); ok {
 		handleID = uint64(readOp.Handle)
 	}
-
+	requestID := uuid.New()
 	logger.Tracef("%.13v <- FileCache(%s:/%s, offset: %d, size: %d, handle: %d)", requestID, fc.bucket.Name(), fc.obj.Name, offset, len(p), handleID)
 
 	startTime := time.Now()
@@ -134,7 +133,7 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 				isSequential = false
 				return 0, false, nil
 			default:
-				return 0, false, fmt.Errorf("tryReadingFromFileCache: getCacheHandle failed: %w", err)
+				return 0, false, fmt.Errorf("tryReadingFromFileCache: GetCacheHandle failed: %w", err)
 			}
 		}
 	}
--- a/internal/gcsx/file_cache_reader.go
+++ b/internal/gcsx/file_cache_reader.go
@@ -129,6 +129,9 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 				logger.Warnf("cache handle creation failed due to size constraints: %v", err)
 				return 0, false, nil
 			case errors.Is(err, cacheutil.ErrCacheHandleNotRequiredForRandomRead):
+				// Fall back to GCS if it is a random read, cacheFileForRangeRead is
+				// False and there doesn't already exist file in cache.
+				isSequential = false
 				return 0, false, nil
 			default:
 				return 0, false, fmt.Errorf("tryReadingFromFileCache: getCacheHandle failed: %w", err)
@@ -191,6 +194,7 @@ func (fc *FileCacheReader) ReadAt(ctx context.Context, p []byte, offset int64) (
 		return o, nil
 	}
 
+	// The cache is unable to serve data and requires a fallback to another reader.
 	err = FallbackToAnotherReader
 	return o, err
 }
--- a/internal/gcsx/file_cache_reader.go
+++ b/internal/gcsx/file_cache_reader.go
@@ -96,7 +96,7 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 
 	startTime := time.Now()
 	var bytesRead int
-	var hit bool
+	var cacheHit bool
 	var err error
 
 	defer func() {
@@ -108,7 +108,7 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 			if fc.fileCacheHandle != nil {
 				isSequential = fc.fileCacheHandle.IsSequential(offset)
 			}
-			requestOutput = fmt.Sprintf("OK (isSeq: %t, hit: %t) (%v)", isSequential, hit, executionTime)
+			requestOutput = fmt.Sprintf("OK (isSeq: %t, cacheHit: %t) (%v)", isSequential, cacheHit, executionTime)
 		}
 
 		logger.Tracef("%.13v -> %s", requestID, requestOutput)
@@ -117,7 +117,7 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 		if isSequential {
 			readType = util.Sequential
 		}
-		fc.captureFileCacheMetrics(ctx, fc.metricHandle, readType, bytesRead, hit, executionTime)
+		fc.captureFileCacheMetrics(ctx, fc.metricHandle, readType, bytesRead, cacheHit, executionTime)
 	}()
 
 	// Create fileCacheHandle if not already.
@@ -130,7 +130,7 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 				return 0, false, nil
 			case errors.Is(err, cacheUtil.ErrCacheHandleNotRequiredForRandomRead):
 				// Fall back to GCS if it is a random read, cacheFileForRangeRead is
-				// False and there doesn't already exist file in cache.
+				// false and there doesn't already exist file in cache.
 				isSequential = false
 				return 0, false, nil
 			default:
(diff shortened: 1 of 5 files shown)


internal/gcsx/file_cache_reader_test.go
```diff
@@ -81,3 +98,209 @@ func (t *FileCacheReaderTest) TestNewFileCacheReader() {
 	assert.Nil(t.T(), reader.metricHandle)
 	assert.Nil(t.T(), reader.fileCacheHandle)
 }
+
+func (t *fileCacheReaderTest) TestReadWithNilFileCacheHandler() {
```
raj-prince, line 102: Lets follow the same test-method naming pattern: Test_<method_name>_behavior().

internal/gcsx/file_cache_reader.go
```diff
@@ -40,12 +52,149 @@ type FileCacheReader struct {
...
+// fileHandle to file in cache. So, we will get the correct data from fileHandle
+// because Linux does not delete a file until open fileHandle count for a file is zero.
+func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte, offset int64) (int, bool, error) {
+	if fc.fileCacheHandler == nil {
+		return 0, false, nil
+	}
+
+	// By default, consider read type random if the offset is non-zero.
+	isSequential := offset == 0
+	requestID := uuid.New()
```
raj-prince, line 88: Lets create `requestID`, just before the request log.

internal/gcsx/file_cache_reader.go
```diff
@@ -40,12 +52,149 @@ type FileCacheReader struct {
...
+		fc.fileCacheHandle = nil
+	} else if !errors.Is(err, cacheUtil.ErrFallbackToGCS) {
+		return 0, false, fmt.Errorf("tryReadingFromFileCache: while reading via cache: %w", err)
+	}
+	err = nil
+
+	return 0, false, nil
+}
+
+func (fc *FileCacheReader) captureFileCacheMetrics(ctx context.Context, metricHandle common.MetricHandle, readType string, readDataSize int, cacheHit bool, readLatency time.Duration) {
```
raj-prince, line 165: Why this method is part of FileCacheReader? - I don't see we are accessing any reader attributes.

internal/gcsx/file_cache_reader.go
```diff
@@ -40,12 +52,149 @@ type FileCacheReader struct {
...
+			case errors.Is(err, lru.ErrInvalidEntrySize):
+				logger.Warnf("tryReadingFromFileCache: while creating CacheHandle: %v", err)
+				return 0, false, nil
+			case errors.Is(err, cacheUtil.ErrCacheHandleNotRequiredForRandomRead):
+				// Fall back to GCS if it is a random read, cacheFileForRangeRead is
+				// false and there doesn't already exist file in cache.
+				isSequential = false
+				return 0, false, nil
+			default:
+				return 0, false, fmt.Errorf("tryReadingFromFileCache: getCacheHandle failed: %w", err)
```
raj-prince, line 137: nit: getCacheHandle -> GetCacheHandle

internal/gcsx/file_cache_reader.go
Tulsishah: We are passing the cacheHit and readType attributes from the file cache read data and capturing metrics. This is being called in tryReadingFromCache method in defer function.

internal/gcsx/file_cache_reader.go
```diff
@@ -40,12 +52,149 @@ type FileCacheReader struct {
...
+
+	var handleID uint64
+	if readOp, ok := ctx.Value(ReadOp).(*fuseops.ReadFileOp); ok {
+		handleID = uint64(readOp.Handle)
+	}
+
+	logger.Tracef("%.13v <- FileCache(%s:/%s, offset: %d, size: %d, handle: %d)", requestID, fc.bucket.Name(), fc.obj.Name, offset, len(p), handleID)
+
+	startTime := time.Now()
+	var bytesRead int
```
raj-prince, line 98: What's the problem in adding these variable name in the method return only? - anyway these are kind of global variable for this method.

internal/gcsx/file_cache_reader.go
```diff
@@ -40,12 +52,149 @@ type FileCacheReader struct {
...
+	metricHandle.FileCacheReadLatency(ctx, float64(readLatency.Microseconds()), []common.MetricAttr{{Key: common.CacheHit, Value: strconv.FormatBool(cacheHit)}})
+}
+
+func (fc *FileCacheReader) ReadAt(ctx context.Context, p []byte, offset int64) (ReaderResponse, error) {
+	var err error
+	readerResponse := ReaderResponse{
+		DataBuf: p,
+		Size:    0,
+	}
+
```
raj-prince, line 181: Can we validate the offset? - as it was in the previous implementation.

offset >= objectSize -> return io.EOF

internal/gcsx/file_cache_reader.go
Tulsishah: In Go, the recommended practice is to return values directly from functions rather than using output parameters. Also, this approach can accidentally lead to the use of naked returns, which is generally considered poor practice.

https://screenshot.googleplex.com/AVZM8iRaMAsf3cc

internal/gcsx/file_cache_reader.go
Tulsishah: This logic will be implemented within the ReadManager. For example, if the ReadManager encounters this specific condition, it will immediately return it, preventing any further attempts to read data using other readers.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -81,3 +98,209 @@ func (t *FileCacheReaderTest) TestNewFileCacheReader() {
...
+func (t *fileCacheReaderTest) Test_ReadAt_NilFileCacheHandlerThrowFallBackError() {
+	reader := NewFileCacheReader(t.object, t.mockBucket, nil, true, nil)
+
+	readerResponse, err := reader.ReadAt(t.ctx, make([]byte, 10), 0)
+
+	assert.True(t.T(), errors.Is(err, FallbackToAnotherReader))
+	assert.Zero(t.T(), readerResponse.Size)
+}
+
+// Writing unit tests on tryReadingFromFileCache to check if cache hit is getting populated correctly.
```
raj-prince, line 111: nit: Writing unit tests? - we can remove this, test-method name conveys everything.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -81,3 +98,209 @@ func (t *FileCacheReaderTest) TestNewFileCacheReader() {
...
+	// Second read will be a cache hit.
+	n, cacheHit, err = t.reader.tryReadingFromFileCache(t.ctx, buf, 0)
+
+	assert.NoError(t.T(), err)
+	assert.True(t.T(), cacheHit)
+	assert.Equal(t.T(), n, len(buf))
+	t.mockBucket.AssertExpectations(t.T())
+}
+
+// Writing unit tests on tryReadingFromFileCache to check if cache hit is getting populated correctly.
```
raj-prince, line 133: nit: do you want to remove this comment?

internal/gcsx/file_cache_reader_test.go
Tulsishah: Generally, it is preferable to write unit tests for public methods rather than helper functions. To avoid confusion for the reader, I've kept this.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -81,3 +98,209 @@ func (t *FileCacheReaderTest) TestNewFileCacheReader() {
...
+	t.mockNewReaderWithHandleCallForTestBucket(0, t.object.Size, rd)
+	t.mockBucket.On("Name").Return("test-bucket")
+	buf := make([]byte, end-start)
+
+	readerResponse, err := t.reader.ReadAt(t.ctx, buf, int64(start))
+
+	assert.True(t.T(), errors.Is(err, FallbackToAnotherReader))
+	assert.Zero(t.T(), readerResponse.Size)
+	job := t.jobManager.GetJob(t.object.Name, t.mockBucket.Name())
+	assert.True(t.T(), job == nil || job.GetStatus().Name == downloader.Downloading)
```
raj-prince, line 215: Can we validate the non-nil cacheHandle also just to make sure cache got triggered?

As job == nil can happen when cache-path is not triggered.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -81,3 +98,209 @@ func (t *FileCacheReaderTest) TestNewFileCacheReader() {
...
+	job := t.jobManager.GetJob(t.object.Name, t.mockBucket.Name())
+	if job != nil {
+		jobStatus := job.GetStatus().Name
+		assert.True(t.T(), jobStatus == downloader.Downloading || jobStatus == downloader.Completed, fmt.Sprintf("the actual status is %v", jobStatus))
+	}
+
+	err = t.reader.fileCacheHandler.InvalidateCache(t.object.Name, t.mockBucket.Name())
+	assert.NoError(t.T(), err)
+
+	readerResponse, err = t.reader.ReadAt(t.ctx, buf, 0)
```
raj-prince, line 301: Lets follow the AAA - assuming we are testing the ReadAt() method, but seems we are testing InvalidateCache method.

internal/gcsx/file_cache_reader.go
raj-prince: I can understand if you want to keep a check at top level, but to make the fileCacheReader complete you should make sure it handles the offset correctly.

Or you need to explicitly add a comment, client needs to make sure pass the valid offset. I would prefer the previous option (adding a check).

internal/gcsx/file_cache_reader_test.go
```diff
@@ -81,3 +98,209 @@ func (t *FileCacheReaderTest) TestNewFileCacheReader() {
...
+}
+
+func (t *fileCacheReaderTest) Test_ReadAt_SequentialRangeRead() {
+	testContent := testutil.GenerateRandomBytes(int(t.object.Size))
+	rd := &fake.FakeReader{ReadCloser: getReadCloser(testContent)}
+	t.mockNewReaderWithHandleCallForTestBucket(0, t.object.Size, rd)
+	t.mockBucket.On("Name").Return("test-bucket")
+	start := 0
+	end := 10
+	assert.Less(t.T(), end, int(t.object.Size))
```
kislaykishore, line 169: Use require
```suggestion
	require.Less(t.T(), end, int(t.object.Size))
```

internal/gcsx/file_cache_reader.go
Tulsishah: As discussed offline kept in file cache reader but removed from part of struct.

raj-prince (approved):

PR #3205, opened 2025-04-16
add zonal buckets to periodic/presubmit checkpoint tests

ashmeenkaur:
### Description
add zonal buckets to periodic/presubmit checkpoint tests

b/411002074

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

NA


perfmetrics/scripts/ml_tests/checkpoint/Jax/run_checkpoints.sh
meet2mky, line 1: Nit: Do we know if we have zonal bytes quota for all Zones from where this test would be run?

ashmeenkaur: Yes, we do!

vadlakondaswetha (approved):

PR #3209, opened 2025-04-18
[Random reader refactoring] File cache reader implementation - 2

Tulsishah:
### Description
- Adding more unit tests on https://github.com/GoogleCloudPlatform/gcsfuse/pull/3202
- Removed unused(ErrInSeekingFileHandle) variable.
- Update variable name from MB to MiB.
- Unit tests coverd
      1. File size is greater than CacheSize
      2. Offset greater than FileSize will return EOF
      3. Cache populated and then cacheMiss Due to InvalidJob
      4. Cache populated and then cacheMiss due to Invalid fileHandle after then cacheHit with NewFileCacheHandle
      5. If cache file gets deleted
      6. If cache file gets deleted with cacheHandle Open
      7. failed job restart and cacheHit
      8. Negative offset should throw error
- Total code coverage - 97% .
- Only these two lines missed - (https://screenshot.googleplex.com/49TdQSoLTAYjJEe)

[b/411320312](http://b/411320312)

### Testing details
1. Manual - NA
9. Unit tests - Added
10. Integration tests - NA


internal/gcsx/file_cache_reader_test.go
```diff
@@ -85,7 +87,13 @@ func (t *fileCacheReaderTest) TearDown() {
 func (t *fileCacheReaderTest) mockNewReaderWithHandleCallForTestBucket(start uint64, limit uint64, rd gcs.StorageReader) {
 	t.mockBucket.On("NewReaderWithReadHandle", mock.Anything, mock.MatchedBy(func(rg *gcs.ReadObjectRequest) bool {
 		return rg != nil && (*rg.Range).Start == start && (*rg.Range).Limit == limit
-	})).Return(rd, nil).Maybe()
+	})).Return(rd, nil).Once()
```
Tulsishah, line 90: For reviewer: It was causing issue as again calling the method was reusing the old reader.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -208,12 +235,11 @@ func (t *fileCacheReaderTest) Test_ReadAt_RandomReadNotStartWithZeroOffsetWhenCa
 
 	readerResponse, err := t.reader.ReadAt(t.ctx, buf, int64(start))
 
-	assert.True(t.T(), errors.Is(err, FallbackToAnotherReader))
+	assert.True(t.T(), errors.Is(err, FallbackToAnotherReader), "expected %v error got %v", FallbackToAnotherReader, err)
 	assert.Zero(t.T(), readerResponse.Size)
 	job := t.jobManager.GetJob(t.object.Name, t.mockBucket.Name())
 	assert.True(t.T(), job == nil || job.GetStatus().Name == downloader.Downloading)
 	assert.NotNil(t.T(), t.reader.fileCacheHandle)
-	t.mockBucket.AssertExpectations(t.T())
```
Tulsishah, line 216: For reviewer: Removed this because test was completing before the NreRangeReader method got hit. Previously it was working because we had MayBe() in mock.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -301,7 +327,152 @@ func (t *fileCacheReaderTest) Test_ReadAt_CacheMissDueToInvalidJob() {
...
+	// Read via cache only, as we have old fileHandle open and linux
+	// doesn't delete the file until the fileHandle count for the file is zero.
+	readerResponse, err = t.reader.ReadAt(t.ctx, make([]byte, t.object.Size), 0)
+
+	assert.NoError(t.T(), err)
+	assert.Equal(t.T(), readerResponse.DataBuf, testContent)
+	t.mockBucket.AssertExpectations(t.T())
+}
+
+func (t *fileCacheReaderTest) Test_ReadAt_FailedJobRestartAndCacheHit() {
```
raj-prince, line 439: nit: Name is a bit misleading - Failed job is not restarted, but a new job is created for the new read.

In the first read: error during NewReaderWithReadHandle() -> Failed job -> invalid file-handle -> and make the file-handle = nil.

In the 2nd read, a new file-handle is created -> creates a new job .... -> finally cache hit.

The new name could be - Test_ReadAt_FailedJobNextReadCreatesNewJobAndCacheHit

internal/gcsx/file_cache_reader.go
```diff
@@ -33,6 +33,12 @@ import (
 	"github.com/jacobsa/fuse/fuseops"
 )
 
+const (
+	// ReadOp ("readOp") is the value used in read context to store pointer to the read operation.
+	ReadOp = "readOp"
+	MiB    = 1 << 20
```
raj-prince, line 39: Instead of keeping the constants here, you can create a constants.go file in the gcsx package, which will be used by all the readers.

raj-prince (commented):
Overall LGTM, couple of minor comments.

internal/gcsx/file_cache_reader.go
Tulsishah: I feel currently we only have two constants so we should not create separate file. If in future it will require then I'll create at that time.

raj-prince (approved):

PR #3211, opened 2025-04-18
Changes in mount_timeout package for ZB E2E

gargnitingoogle:
### Description
- Splits the existing MountTimeoutTest class to NonZBMountTimeoutTest and ZBMountTimeoutTest. 
- Creates and adds new ZBs for mount_timeout tests and mounts them in the test
- Disables the tests for mounting of non-ZB out of the E2E ZB runs
- Adds new tests for mounting the ZBs during the E2E ZB runs .

[b/411578929](http://b/411578929) , [b/410500613](http://b/410500613)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - as part of presubmit - [logs](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/a369417f-6667-4e9b-b529-e200bc260385/summary)

### DIFF
--- a/tools/integration_tests/mount_timeout/mount_timeout_test.go
+++ b/tools/integration_tests/mount_timeout/mount_timeout_test.go
@@ -63,19 +63,24 @@ const (
 	zonalUSCentral1ABucket         string        = "mount_timeout_test_bucket_zb_usc1a"
 	zonalUSWest4ABucket            string        = "mount_timeout_test_bucket_zb_usw4a"
 	zonalSameZoneExpectedMountTime time.Duration = 2500 * time.Millisecond
+	zonalCrossZoneExpectedMountTime time.Duration = 5000 * time.Millisecond
 	// Commont constants.
 	relaxedExpectedMountTime time.Duration = 8000 * time.Millisecond
 	logfilePathPrefix        string        = "/tmp/gcsfuse_mount_timeout_"
 )
 
-// findTestExecutionEnvironment determines the environment in which the tests are running.
+// findTestExecutionEnvironment determines the environment (region, zone) in which the tests are running.
 // It uses the GCP resource detector to identify the environment.
 //
 // If the tests are running on a GCE instance with a hostname containing non-gce.
 // it returns testEnvNonGCE since it implies that the tests are being run on cloudtop.
 //
-// If the tests are running on a VM in the "us-central" region, it returns gce-us-central .
+// If the tests are running on a VM in the "us-central" region, it returns gce-us-central.
 // Otherwise, if running in any other region, it returns gce-non-us-central.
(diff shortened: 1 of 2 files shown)


tools/integration_tests/mount_timeout/mount_timeout_test.go
```diff
@@ -55,37 +56,59 @@ const (
...
 	}
 	if v, exists := attrs.Value("cloud.region"); exists {
-		if strings.Contains(strings.ToLower(v.AsString()), "us-central") {
-			return testEnvGCEUSCentral
-		} else {
-			return testEnvGCENonUSCentral
+		region := strings.ToLower(v.AsString())
+		// The above method only ever returns region of the VM, and not the zone. So, I am assuming that when region is us-central1, then zone is us-central1-a, which may not be true always.
+		switch region {
+		case "us-central1":
```
anushka567, line 99: We can find out the exact zone of the test VM by using. `attrs.Value("availability_zone")`
ref: [link](https://opentelemetry.io/docs/specs/semconv/attributes-registry/cloud/#cloud-availability-zone)

gargnitingoogle: > We can find out the exact zone of the test VM by using. attrs.Value("availability_zone")

Used this to find out the VM's zone.

> If the GCE VM is in us-central1-b and the fixed zonal bucket as precreated , is in us-central1-a , will the threshold defined be correct?

Don't know for sure. But it should be same as or worse than the case where VM and bucket are in same zone. Now I've made the different between same-zone and cross-zone mounting cases.

anushka567 (approved):
LGTM.  I did notice one minor difference between the non-ZB and ZB test configurations. The ZB tests include two basic parameters—timeout and bucket-name—whereas the non-ZB tests only specify timeout.

To ensure consistency across both test types, I suggest we consider including the bucket-name parameter in the non-ZB test configuration as well.

gargnitingoogle:
> LGTM. I did notice one minor difference between the non-ZB and ZB test configurations. The ZB tests include two basic parameters—timeout and bucket-name—whereas the non-ZB tests only specify timeout.
> 
> To ensure consistency across both test types, I suggest we consider including the bucket-name parameter in the non-ZB test configuration as well.

thanks @anushka567 . I didn't want to increase the scope of this task by changing anything about the non-ZB tests, so I didn't change anything there. For now, I've captured the request in b/412536927 . I'll leave it as improvement task for whoever can take it up.

Tulsishah (approved):
Relying on @anushka567's review

PR #3215, opened 2025-04-20
Fix: Handle Intermittent SSH Connection Errors for long haul test VMs

anushka567:
### Description
Intermittent failures occur when establishing SSH connections to long haul test VMs, arising from a timing conflict between the removal of OS Login-managed SSH keys and subsequent SSH connection attempts. Specifically, `gcloud compute ssh ` may attempt to connect using a newly generated key before the OS Login key removal has fully propagated.This leads to the VM rejecting the authentication attempt , as a result, kokoro jobs which probe for the test status fail.
The ssh keys removal is necessary due to the size limit on os login profiles: [ref](https://github.com/kyma-project/test-infra/issues/93)

Solution -
Implemented a retry mechanism with exponential backoff for the SSH connection attempt. This allows sufficient time for the OS Login key removal to propagate, significantly reducing the frequency of authentication failures.
### Testing details
1. Manual - WAI, ref: 
image

2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
+++ b/perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
@@ -39,26 +39,26 @@ function initialize_ssh_key () {
     delay=1
     max_delay=10
     attempt=1
-    retries=5
-    while [ $attempt -le $retries ]; do
+    max_attempts=5
+    while true; do
         echo "Attempting SSH connection (attempt $attempt)..."
-        # Requires running ssh command with --quiet option to initialize keys.
+        # Requires running first ssh command with --quiet option to initialize keys.
         # Otherwise it prompts for yes or no.
         if sudo gcloud compute ssh "$VM_NAME" --zone "$ZONE_NAME" --internal-ip --quiet --command "echo 'Running from VM'"; then
-            echo "SSH connection successful."
             return 0
-        else
-            echo "SSH connection failed. Waiting $delay seconds before retrying..."
-            sleep $delay
-            delay=$((delay * 2)) # Exponential backoff
-            if [ "$delay" -gt "$max_delay" ]; then
-                delay="$max_delay" # Cap the delay
-            fi
-            attempt=$((attempt + 1))
         fi
+        if [ "$attempt" -ge "$max_attempts" ]; then
+            echo "All SSH connection attempts failed."
+            exit 1
+        fi
+        echo "SSH connection failed. Waiting $delay seconds before retrying..."
+        sleep $delay
+        delay=$((delay * 2))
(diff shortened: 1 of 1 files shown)


gargnitingoogle (approved):
LGTM. Minor comments.

perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
```diff
@@ -36,10 +36,29 @@ function initialize_ssh_key () {
...
+    retries=5
+    while [ $attempt -le $retries ]; do
+        echo "Attempting SSH connection (attempt $attempt)..."
+        # Requires running first ssh command with --quiet option to initialize keys.
+        # Otherwise it prompts for yes and no.
+        if sudo gcloud compute ssh "$VM_NAME" --zone "$ZONE_NAME" --internal-ip --quiet --command "echo 'Running from VM'"; then
+            echo "SSH connection successful."
+            return 0
+        else
+            echo "SSH connection failed. Waiting $delay seconds before retrying..."
```
gargnitingoogle, line 51: There is a slight inefficiency here. After the last failed attempt (i.e. `$attempt > $retries`), you are unnecessarily going to sleep here for no reason. We could simply break out of the loop here instead of going to sleep in that case.

perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
```diff
@@ -36,10 +36,29 @@ function initialize_ssh_key () {
...
     # This is required to avoid issue: https://github.com/kyma-project/test-infra/issues/93
     for i in $(sudo gcloud compute os-login ssh-keys list | grep -v FINGERPRINT); do sudo gcloud compute os-login ssh-keys remove --key $i; done
-
-    # Requires running first ssh command with --quiet option to initialize keys.
-    # Otherwise it prompts for yes and no.
-    sudo gcloud compute ssh $VM_NAME --zone $ZONE_NAME --internal-ip --quiet --command "echo 'Running from VM'"
+    delay=1
+    max_delay=10
+    attempt=1
+    retries=5
```
gargnitingoogle, line 42: Could we rename `retries` to `max_attempts` to make it a bit more intuitive ?

gargnitingoogle (changes requested):
minor comments

perfmetrics/scripts/continuous_test/ml_tests/run_and_manage_test.sh
anushka567: Updated to exit immediately after final attempt failure.

gargnitingoogle (approved):
LGTM

Tulsishah (approved):
Relying on @gargnitingoogle's review.

PR #3220, opened 2025-04-21
Update documentation for  custom-endpoint flag

Tulsishah:
### Description
We've removed `Should only be used for testing` from the custom endpoint description. This was added previously as the feature was developed for the test-bench server.

Currently, the custom endpoint is being used in the TPC environment, which is not a test server. Therefore, we are updating its description.

I'll also update cloud documentation as well after this.

[b/412284502](https://b.corp.google.com/issues/412284502)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -277,7 +277,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.BoolP("create-empty-file", "", false, "For a new file, it creates an empty file in Cloud Storage bucket as a hold.")
 
-	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data. The custom endpoint  must support the equivalent resources and operations as the GCS  JSON endpoint, https://storage.googleapis.com/storage/v1. If a custom endpoint is not specified,  GCSFuse uses the global GCS JSON API endpoint, https://storage.googleapis.com/storage/v1.")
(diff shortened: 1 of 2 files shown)


cfg/config.go
```diff
@@ -277,7 +277,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.BoolP("create-empty-file", "", false, "For a new file, it creates an empty file in Cloud Storage bucket as a hold.")
 
-	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data. Should only be used for testing.  The custom endpoint must support the equivalent resources and operations as the GCS  JSON endpoint, https://storage.googleapis.com/storage/v1. If a custom endpoint is not specified,  GCSFuse uses the global GCS JSON API endpoint, https://storage.googleapis.com/storage/v1.")
+	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data. The custom endpoint must support the equivalent resources and operations as the GCS JSON endpoint, https://storage.googleapis.com/storage/v1. If a custom endpoint is not specified, GCSFuse uses the global GCS JSON API endpoint, https://storage.googleapis.com/storage/v1.")
```
mustvicky, line 280: Should we specify more than that, something like: 

This can be used for testing among other use cases.  
cc: Ashish Tara, I have requested his attention to this PR as well. They might give private message feedback if not on github.

mustvicky (approved):
I don't have any blocking comment. It was more of a conversation starter to think about more things.

cfg/config.go
Tulsishah: I think we shouldn't write too much, as it's self-explanatory that this flag is used to pass any endpoint other than GCS. Explicitly mentioning testing makes it sound like it's only for testing purposes.

raj-prince (approved):

PR #3225, opened 2025-04-22
TPC support with HNS buckets

Tulsishah:
### Description
Previously, TPC did not support the control client at the time GCSFuse implementation for TPC was developed. However, control client support for TPC has been available since December 2024, necessitating modifications to enable GCSFuse to support TPC with HNS buckets.

Also added automated tests for HNS buckets on TPC.

[b/379063794](https://b.corp.google.com/issues/379063794)

### Testing details
1. Manual - Tested on TPC environment
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/run_e2e_tests.sh
+++ b/tools/integration_tests/run_e2e_tests.sh
@@ -428,7 +428,7 @@ function run_e2e_tests_for_zonal_bucket(){
 function run_e2e_tests_for_tpc() {
   bucket=$1
   # Clean bucket before testing.
-  gcloud storage rm -r gs://$bucket/**
+  gcloud storage rm -r gs://$bucket/*
 
   # Run Operations e2e tests in TPC to validate all the functionality.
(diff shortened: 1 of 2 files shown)


tools/integration_tests/run_e2e_tests.sh
```diff
@@ -425,24 +425,25 @@ function run_e2e_tests_for_zonal_bucket(){
    return 0
 }
 
-function run_e2e_tests_for_tpc_and_exit() {
+function run_e2e_tests_for_tpc() {
+  bucket=$1
   # Clean bucket before testing.
-  gcloud storage rm -r gs://gcsfuse-e2e-tests-tpc/**
+  gcloud storage rm -r gs://$bucket/*
```
gargnitingoogle, line 431: Also, this can be dangerous if bucket is `""`, so we should check for that.

raj-prince (approved):
LGTM.

Relying on @gargnitingoogle for e2e test script changes.

PR #3230, opened 2025-04-22
fix: Return complete object on flush writer 

ashmeenkaur:
### Description
Create object from attributes returned by the writer's `flush` operation, specifically for zonal buckets. The full GCS object metadata (including name, generation, metageneration, etc.) is now returned after a successful flush of a zero-byte file.
Also use the returned object to populate metadata cache.

b/412583244

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

NA

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -666,7 +666,7 @@ func (f *FileInode) SyncPendingBufferedWrites() error {
 	if f.bwh == nil {
 		return nil
 	}
-	err := f.bwh.Sync()
+	_, err := f.bwh.Sync()
 	var preconditionErr *gcs.PreconditionError
 	if errors.As(err, &preconditionErr) {
 		return &gcsfuse_errors.FileClobberedError{
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -260,7 +260,7 @@ func (bh *bucketHandle) FlushPendingWrites(ctx context.Context, w gcs.Writer) (o
 		err = gcs.GetGCSError(err)
 	}()
 
-	offset, err := w.Flush()
+	_, err = w.Flush()
 	if err != nil {
 		err = fmt.Errorf("error in FlushPendingWrites : %w", err)
 		return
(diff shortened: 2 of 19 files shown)


vadlakondaswetha (dismissed):

internal/fs/inode/file.go
```diff
@@ -666,7 +666,7 @@ func (f *FileInode) SyncPendingBufferedWrites() error {
 	if f.bwh == nil {
 		return nil
 	}
-	err := f.bwh.Sync()
+	_, err := f.bwh.Sync()
```
meet2mky, line 669: Can we propagate this minObject back to upto `fs.go` layer. The way, I see is that we would need to promote this `f` to generationBackedInodes as soon as first Sync() is called for local files/and empty gcs file for Zonal Objects.

internal/storage/bucket_handle.go
```diff
@@ -251,18 +251,25 @@ func (bh *bucketHandle) FinalizeUpload(ctx context.Context, w gcs.Writer) (o *gc
...
 	}
 
-	return offset, err
+	attrs := w.Attrs() // Retrieving the attributes of the created object.
+	// Converting attrs to type *MinObject.
+	o = storageutil.ObjectAttrsToMinObject(attrs)
+	if o == nil {
+		return nil, fmt.Errorf("FlushPendingWrites: nil object returned after w.Flush()")
+	}
+	o.Size = uint64(offset)
```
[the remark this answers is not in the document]
meet2mky: I think, we can remove this as attrs has the correct size which is same as offset in this case.

meet2mky, line 271:
```suggestion
```

meet2mky (approved):
Overall looks good. Some comments.

internal/fs/inode/file.go
ashmeenkaur: I think we should do this in a separate PR (which also takes care of promoting the inode to generation backed inode)

vadlakondaswetha (approved):

PR #3233, opened 2025-04-23
[Random Reader Refactoring] Range Reader implementation - 2

Tulsishah:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/client_readers/range_reader.go
+++ b/internal/gcsx/client_readers/range_reader.go
@@ -15,18 +15,26 @@
 package gcsx
 
 import (
+	"context"
+	"errors"
 	"fmt"
+	"io"
 
+	"github.com/googlecloudplatform/gcsfuse/v2/common"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/fs/gcsfuse_errors"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
 )
 
 type RangeReader struct {
 	// TODO: Add additional fields as needed.
 	gcsx.Reader
-	start int64
-	limit int64
+	object *gcs.MinObject
+	bucket gcs.Bucket
+	start  int64
+	limit  int64
 
 	// If non-nil, an in-flight read request and a function for cancelling it.
 	//
@@ -38,12 +46,18 @@ type RangeReader struct {
 	// checks.
 	readHandle []byte
 	cancel     func()
+
+	readType     string
+	metricHandle common.MetricHandle
 }
 
-func NewRangeReader() *RangeReader {
+func NewRangeReader(object *gcs.MinObject, bucket gcs.Bucket, metricHandle common.MetricHandle) *RangeReader {
 	return &RangeReader{
-		start: -1,
-		limit: -1,
+		object:       object,
+		bucket:       bucket,
+		metricHandle: metricHandle,
+		start:        -1,
+		limit:        -1,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


internal/gcsx/client_readers/range_reader.go
```diff
@@ -15,18 +15,25 @@
...
 )
 
 type RangeReader struct {
-	// TODO: Add additional fields as needed.
 	gcsx.Reader
-	start int64
-	limit int64
+	object *gcs.MinObject
+	bucket gcs.Bucket
+	start  int64
```
raj-prince, line 35: Can we provide the brief description of start, limit? What is it?

internal/gcsx/client_readers/range_reader.go
```diff
@@ -81,3 +94,158 @@ func (rr *RangeReader) closeReader() {
...
+		Size:    0,
+	}
+	var err error
+
+	readerResponse.Size, err = rr.readFromRangeReader(ctx, req.Buffer, req.Offset, req.EndOffset, rr.readType)
+
+	return readerResponse, err
+}
+
+// readFromRangeReader reads using the NewReader interface of go-sdk. Its uses
```
raj-prince, line 110: nit: its -> it

internal/gcsx/client_readers/range_reader.go
Tulsishah: I think it is self-explanatory, but since you felt we should add it, so did I

internal/gcsx/client_readers/range_reader.go
```diff
@@ -15,17 +15,38 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v2/common"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/fs/gcsfuse_errors"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v2/internal/util"
+)
+
+const (
+	MiB = 1 << 20
```
raj-prince, line 33: nit: newline before the comment.

internal/gcsx/client_readers/range_reader.go
```diff
@@ -15,17 +15,38 @@
...
+	// Max read size in bytes for random reads.
+	// If the average read size (between seeks) is below this number, reads will optimise for random access.
+	// We will skip forwards in a GCS response at most this many bytes.
+	// About 6 MiB of data is buffered anyway, so 8 MiB seems like a good round number.
```
raj-prince, line 37: nit: Comment doesn't provide the full context. Either remove the last line, or be a bit more descriptive about the buffering. Here, we are talking about tcp buffer.

Note: I know, this comment was already there.

internal/gcsx/client_readers/range_reader.go
```diff
@@ -81,3 +108,213 @@ func (rr *RangeReader) closeReader() {
...
+	}()
+
+	return io.ReadFull(rr.reader, p)
+}
+
+// Ensure that rr.reader is set up for a range for which [start, start+size) is
+// a prefix. Irrespective of the size requested, we try to fetch more data
+// from GCS defined by sequentialReadSizeMb flag to serve future read requests.
+func (rr *RangeReader) startRead(ctx context.Context, start int64, end int64) error {
+	ctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
raj-prince: We should create a new context, which will be valid for the lifecycle of reader (response stream).

If I understand correctly, existing one is per ReadFile() call from the kernel, so using that might cancel our reader instance in case parent is cancelled after a ReadFile() call.

internal/gcsx/client_readers/range_reader.go
```diff
@@ -81,3 +108,213 @@ func (rr *RangeReader) closeReader() {
...
+func (rr *RangeReader) invalidateReaderIfMisalignedOrTooSmall(offset int64, p []byte) bool {
+	rr.skipBytes(offset)
+
+	// If we have an existing reader, but it's positioned at the wrong place,
+	// clean it up and throw it away.
+	// We will also clean up the existing reader if it can't serve the entire request.
+	dataToRead := math.Min(float64(offset+int64(len(p))), float64(rr.object.Size))
+	if rr.reader != nil && (rr.start != offset || int64(dataToRead) > rr.limit) {
+		rr.closeReader()
+		rr.reader = nil
```
raj-prince, line 311: Can we do make the cancel function also nil?

rr.cancel = nil

internal/gcsx/client_readers/range_reader_test.go
```diff
@@ -159,11 +214,112 @@ func (t *rangeReaderTest) Test_CheckInvariants() {
...
+
+func (t *rangeReaderTest) Test_ReadAt_ReadFailsWithTimeoutError() {
+	content := "xxx"
+	r := iotest.OneByteReader(iotest.TimeoutReader(strings.NewReader(content)))
+	rc := &fake.FakeReader{ReadCloser: io.NopCloser(r)}
+	t.mockNewReaderWithHandleCallForTestBucket(0, uint64(len(content)), rc)
+
+	readerResponse, err := t.readAt(0, int64(len(content)))
+
+	assert.Error(t.T(), err)
```
raj-prince, line 234: I agree, we should generally not match the error.

But in this case, since we are validating if timeout scenario so good to add check for timeout error. We don't need to match the string completely.

internal/gcsx/client_readers/range_reader_test.go
```diff
@@ -159,11 +214,112 @@ func (t *rangeReaderTest) Test_CheckInvariants() {
...
+func (t *rangeReaderTest) Test_ReadAt_PartialReadWithEOF() {
+	offset := int64(0)
+	size := int64(10)                                       // Shorter than requested
+	partialReader := io.NopCloser(iotest.ErrReader(io.EOF)) // Simulates early EOF
+	r := &fake.FakeReader{ReadCloser: partialReader}
+	t.mockNewReaderWithHandleCallForTestBucket(uint64(offset), uint64(offset+size), r)
+
+	resp, err := t.readAt(offset, size)
+
+	assert.Error(t.T(), err)
```
raj-prince, line 263: Here, also. 

I feel, good to add check that err is EOF only to make sure. Simulation happened correctly.

internal/gcsx/client_readers/range_reader_test.go
```diff
@@ -159,11 +214,112 @@ func (t *rangeReaderTest) Test_CheckInvariants() {
...
+	t.mockBucket.On("NewReaderWithReadHandle", mock.Anything, mock.Anything).Return(nil, errors.New("network error")).Once()
+
+	resp, err := t.readAt(offset, size)
+
+	assert.Error(t.T(), err)
+	assert.Zero(t.T(), resp.Size)
+	t.mockBucket.AssertExpectations(t.T())
+}
+
+func (t *rangeReaderTest) Test_invalidateReaderIfMisalignedOrTooSmall_InvalidateReaderDueToWrongOffset() {
```
raj-prince, line 293: Can we write TDD test for invalidateReaderIfMisalignedOrTooSmall?

raj-prince (commented):
Added couple of comments.

internal/gcsx/client_readers/range_reader.go
Tulsishah: Removed last line.

internal/gcsx/client_readers/range_reader_test.go
Tulsishah: In the event of an EOF error, we are throwing a different error. 
```
if rr.reader != nil {
			err = fmt.Errorf("reader returned early by skipping %d bytes", rr.limit-rr.start)
			return 0, err
		}

		err = nil
```
While it's generally preferable in Go to avoid matching error strings, since you feel it's necessary in this instance, I've added an assertion to check the string.

internal/gcsx/client_readers/range_reader.go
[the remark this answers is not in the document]
Tulsishah: I used 
```
ctx, cancel := context.WithCancel(ctx)
```
which creates a child context from the parent. Canceling this child context will not cancel the parent context (in this case, the kernel's).

Using a parent context, rather than a new one, helps maintain a clear hierarchy. Child contexts inherit cancellation signals and values, ensuring they respect the parent's scope (like timeouts). This prevents accidental overrides and ensures related operations are tied together – if the parent is done, the children should also stop.

PR #3237, opened 2025-04-23
Support directpath for StorageControl

cjc25:
### Description
This allows the StorageControl client to use directpath/xDS endpoints. These mirror the default gRPC options in package cloud.google/com/go/storage.

See https://github.com/googleapis/google-cloud-go/blob/main/storage/grpc_client.go#L105-L109

### Related issues
[b/412688930](http://b/412688930)

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - Ran through presubmit

### DIFF
--- a/internal/storage/storageutil/control_client.go
+++ b/internal/storage/storageutil/control_client.go
@@ -24,6 +24,7 @@ import (
 	"github.com/googleapis/gax-go/v2"
 	"github.com/googlecloudplatform/gcsfuse/v2/internal/logger"
 	"google.golang.org/api/option"
+	"google.golang.org/api/option/internaloption"
 	"google.golang.org/grpc/codes"
 )
 
@@ -58,7 +59,8 @@ func CreateGRPCControlClient(ctx context.Context, clientOpts []option.ClientOpti
 		logger.Fatal("error setting direct path env var: %v", err)
 	}
 
-	controlClient, err = control.NewStorageControlClient(ctx, clientOpts...)
+	opts := append(clientOpts, internaloption.EnableDirectPath(true), internaloption.AllowNonDefaultServiceAccount(true), internaloption.EnableDirectPathXds())
+	controlClient, err = control.NewStorageControlClient(ctx, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("NewStorageControlClient: %w", err)
 	}


cjc25:
I don't think we need to have different options for the emulator, as cloud.google.com/go/storage does: it doesn't look like https://github.com/googleapis/storage-testbench supports StorageControl at all. But if we use the testbench in some workloads, we probably should do the same thing for StorageControl and Storage clients.

internal/storage/storageutil/control_client.go
```diff
@@ -58,7 +59,8 @@ func CreateGRPCControlClient(ctx context.Context, clientOpts []option.ClientOpti
 		logger.Fatal("error setting direct path env var: %v", err)
 	}
 
-	controlClient, err = control.NewStorageControlClient(ctx, clientOpts...)
+	opts := append(clientOpts, internaloption.EnableDirectPath(true), internaloption.AllowNonDefaultServiceAccount(true), internaloption.EnableDirectPathXds())
```
tritone, line 62: I think the internaloptions are only meant to be used inside google-cloud-go libraries. @codyoss @quartzmo is there a config we can set on the generator to get these into the storage/control GAPIC by default?

quartzmo: The `internaloption` values in a gRPC GAPIC are set [here](https://github.com/googleapis/gapic-generator-go/blob/main/internal/gengapic/gengrpc.go#L169-L201) in a pretty much static operation. The only conditional is if the client  supports gRPC.

> internaloptions are only meant to be used inside google-cloud-go libraries

The reason for this afaik is primarily the instability of the options. Most should be considered EXPERIMENTAL, and may be removed/replaced at any time. The maintainers of `internaloption` will not be on the lookout for usages outside of the core Go client libraries projects such as gapic-generator-go, google-cloud-go, and google-api-go-client. Using them here is risky.

tritone: +1, I think this is fine for now. I will file an internal FR to have a non-internal option available for enabling DP in a GAPIC (or allow service owners to force this option to be supplied via a generator config).

charith87 (approved):
LGTM from me. Adding @raj-prince and @abhishek10004 since it involves GRPC/ZB. 
cc @vadlakondaswetha

cjc25:
The tests passed, I just can't upload to codecov from my fork. Likewise for shadow reviewer I'd need some token valid for my fork.

Happy to do some setup if someone can send me instructions.

gargnitingoogle:
> The tests passed, I just can't upload to codecov from my fork. Likewise for shadow reviewer I'd need some token valid for my fork.
> 
> Happy to do some setup if someone can send me instructions.

I created https://github.com/GoogleCloudPlatform/gcsfuse/pull/3248 to unblock this crucial change.

charith87:
> The tests passed, I just can't upload to codecov from my fork. Likewise for shadow reviewer I'd need some token valid for my fork.
> 
> Happy to do some setup if someone can send me instructions.

@kislaykishore regarding codecov

gargnitingoogle:
Closing this in favor of https://github.com/GoogleCloudPlatform/gcsfuse/pull/3248 .

PR #3240, opened 2025-04-24
fix: Use random bucket name in each CreateTestBucketForDynamicMounting call instead of using  fixed random name.

meet2mky:
### Description

Current method can't create multiple random buckets for dynamic mounting if required for parallel execution of tests within same package test. Currently running two tests in a test package in sequential is also not possible if they create dynamic bucket and one of them fails to delete it. Thus other test will see that bucket already exists on GCS. This is part of larger effort in which I am trying to reduce test package run time.

- Use random bucket name in each CreateTestBucketForDynamicMounting call instead of using random fixed name.
- Use `metadata.ProjectIDWithContext(ctx)` instead of `deprecated metadata.ProjectID()`
- Improved error handling.

b/413231761

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - run as part of Presubmit run


Tulsishah (approved):
Please mention what level of parallelism in description. e.g. In this case it is flag level in package.

meet2mky:
> Please mention what level of parallelism in description. e.g. In this case it is flag level in package.

Updated the description of the PR.

PR #3243, opened 2025-04-24
[Random Reader Refactoring] Range reader implementation - 3

Tulsishah:
### Description
- Adding more unit tests on https://github.com/GoogleCloudPlatform/gcsfuse/pull/3233
  1. Test ReadFromRangeReader when reader returned more data
  2. ReadAt propagates cancellation
  3. ReadAt context cancel after read done skips cancel
  4. Reader No tExhausted
  5. Reader exhausted read finished
  6. When reader has less data than requested
  7. When reader has more data than requested
  8. When all data from reader is read
  9. Unappropriat offset
- Made destroy and checkInvariant method private.
- Made Range reader to implement GCSReader interface
- Adding check if offset is greater than object size

### Code coverage  of range reader - 97.8%
https://screenshot.googleplex.com/8ftVinFWZkHLc4c
https://screenshot.googleplex.com/9n9L3vz9mKpjVGn

[b/413258486](https://b.corp.google.com/issues/413258486)

### Testing details
1. Manual - NA
10. Unit tests - Added
11. Integration tests - NA


internal/gcsx/client_readers/range_reader_test.go
```diff
@@ -355,3 +370,145 @@ func (t *rangeReaderTest) Test_invalidateReaderIfMisalignedOrTooSmall() {
...
+	go cancel()
+
+	// Let the read complete
+	select {
+	case <-readReturned:
+		// Pass
+	case <-time.After(100 * time.Millisecond):
+		t.T().Fatal("Expected read to finish before cancellation")
+	}
+	assert.False(t.T(), cancelCalled, "Expected rr.cancel NOT to be called since read already finished")
```
raj-prince, line 499: This can fail, if cancel() method will be called before the read. I think, you need to make sure cancel() is called after the read-done.

What about just calling both ReadAt() and cancel() sequentially, instead of calling as background thread?

raj-prince (commented):
Overall LGTM, one minor comment.

internal/gcsx/client_readers/range_reader_test.go
Tulsishah: So there are two cases and I wrote this test specifically for second code snippet
```
case <-readDone:
			return
```
```
**case <-ctx.Done():
			select {
			case <-readDone:
				return**

			default:
				rr.cancel()
			}
		}
```

If I do sequential call then it will go through the first code snippet on the other side if I trigger cancel first then it is not triggering any case. So to trigger this case I did parallelism.

On cancel call, I realized we are not doing `rr.cancel` in such case so cancel function will not being called, so I removed that assertion.

Tulsishah: As we discussed offline, making it serial also triggers the test with the race condition because the channel can listen in any sequence since its listener is triggered as a go routine. Therefore, I removed the parallelism.

raj-prince (approved):

PR #3245, opened 2025-04-24
Fix error handing with gcsFullReadCloser

kislaykishore:
### Description
If an EOF is encountered before reading the full length of the buffer, ReadFull returns an ErrUnexpectedEOF error. This needs to be convered to EOF in order to have a consistent behavior (error) with and without gcsFullReadCloser.

b/413109300

### Testing details
1. Manual - NA
2. Unit tests - Updated
3. Integration tests - Executed after enabling metrics

### DIFF
--- a/internal/monitor/full_read_closer.go
+++ b/internal/monitor/full_read_closer.go
@@ -31,9 +31,8 @@ func newGCSFullReadCloser(reader gcs.StorageReader) gcs.StorageReader {
 }
 
 // Read reads exactly len(buf) bytes from the wrapped StorageReader into buf.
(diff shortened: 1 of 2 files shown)


internal/monitor/full_read_closer.go
```diff
@@ -35,7 +35,14 @@ func newGCSFullReadCloser(reader gcs.StorageReader) gcs.StorageReader {
 // 2. EOF only if no bytes were read.
```
ashmeenkaur, line 35: Shall we update this comment too?

ashmeenkaur (approved):

PR #3246, opened 2025-04-25
fix: promote zonal object to generation backed inode on sync

meet2mky:
### Description
Promote zonal bucket objects to generation backed inodes as soon as the Sync() is called for local inodes which does a implicit flush on zonal buckets and upload pending buffers for non-zonal buckets. In case of non-zonal buckets they are only updated to generation backed inodes once the Close() has been called on them which finalizes the object.
Note: Start running tests which were skipped as part of b/413296959
**Note: A follow up PR will be send to promote Zonal Object to generation backed Inode as soon as BWH is initialized.**

b/413296959
b/413587897

### Testing details
1. Manual - Tested skipped tests from b/413296959
2. Unit tests - Updated unit tests and tested.
3. Integration tests - Run as part of presubmit tests.

### DIFF
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -109,7 +109,7 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// because the inode is dirty).
 	fh.inode.Lock()
 	// Ensure all pending writes to Zonal Buckets are flushed before issuing a read.
-	err = fh.inode.SyncPendingBufferedWrites()
+	_, err = fh.inode.SyncPendingBufferedWrites()
 	if err != nil {
 		fh.inode.Unlock()
 		err = fmt.Errorf("fh.inode.SyncPendingBufferedWrites: %w", err)
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -109,6 +109,8 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// because the inode is dirty).
 	fh.inode.Lock()
 	// Ensure all pending writes to Zonal Buckets are flushed before issuing a read.
+	// Updating inode state is not required here because inode state for Zonal Buckets will
+	// be updated at time of BWH creation.
 	_, err = fh.inode.SyncPendingBufferedWrites()
 	if err != nil {
 		fh.inode.Unlock()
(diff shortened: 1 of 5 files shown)


internal/fs/handle/file.go
```diff
@@ -109,7 +109,7 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// because the inode is dirty).
 	fh.inode.Lock()
 	// Ensure all pending writes to Zonal Buckets are flushed before issuing a read.
-	err = fh.inode.SyncPendingBufferedWrites()
+	_, err = fh.inode.SyncPendingBufferedWrites()
```
ashmeenkaur, line 112: Can we add a comment here explaining why updating inode state after this sync call is not required here (because we will update inode state after bwh creation)

internal/fs/inode/file.go
```diff
@@ -653,7 +653,9 @@ func (f *FileInode) flushUsingBufferedWriteHandler() error {
 	if err != nil {
 		return fmt.Errorf("f.bwh.Flush(): %w", err)
 	}
-
+	// Set BWH to nil as as object has been finalized.
```
ashmeenkaur, line 656: Nit: Instead of doing this here, can we create 2 separate methods that updates inode state after sync and after flush? The current one is for flush so we can rename it to `updateInodeStateAfterFlush`

updateInodeStateAfterFlush can call updateInodeStateAfterSync

internal/fs/inode/file.go
```diff
@@ -813,9 +819,8 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
 	}
 
 	if f.bwh != nil {
-		// SyncPendingBufferedWrites does not finalize the upload,
-		// so return gcsSynced as false.
-		err = f.SyncPendingBufferedWrites()
+		// SyncPendingBufferedWrites returns gcsSynced as true for Zonal Buckets.
```
ashmeenkaur, line 822: nit: this comment is no longer required now as we are already explaining when gcsSynced is returned true in SyncPendingBufferedWrites method

internal/fs/handle/file.go
meet2mky: Done

internal/fs/inode/file.go
meet2mky: Done!

internal/fs/inode/file.go
meet2mky: Removed it. Thanks!

ashmeenkaur (approved):

PR #3249, opened 2025-04-28
Support GCE_METADATA_HOST in background mode

chengs:
### Description
`GCE_METADATA_HOST` is supported by google sdks to bypass GCE metadata service endpoint, for some cases (eg, container)
Documented in `https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/metadata.go`
And `google-cloud-go` is a dependency of `gcsfuse`, which means `GCE_METADATA_HOST` should be supported by `gcsfuse`.

However, only --foreground mode supports `GCE_METADATA_HOST`.
In background mode, the env does not check and include `GCE_METADATA_HOST`
It is a bug to me and this PR fixes it.

### Testing details
1. Manual - OK
2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -310,6 +310,13 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
 		if p, ok := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS"); ok {
 			env = append(env, fmt.Sprintf("GOOGLE_APPLICATION_CREDENTIALS=%s", p))
 		}
+
+		// Pass along GCE_MEATADATA_HOST, since it is supported by google-cloud-go
+		// https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/metadata.go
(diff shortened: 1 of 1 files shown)


raj-prince:
Apology @chengs, I missed it.

In the mean time, @gargnitingoogle has made this change - https://github.com/GoogleCloudPlatform/gcsfuse/pull/3253 and merged in the latest master.

gargnitingoogle:
@chengs @raj-prince this PR is now oudated and not required because of the fix already made in https://github.com/GoogleCloudPlatform/gcsfuse/pull/3253. That fix was also released back in [v3.0.0](https://github.com/GoogleCloudPlatform/gcsfuse/releases/tag/v3.0.0) . So, this can be now closed.

chengs:
@raj-prince no worries. btw I raised this bug mentioned in #3253 through Google Support using our company account.
Since it is fixed, I have using it for several months and it works well.
It is a bit late, but @gargnitingoogle thanks for the fix.

PR #3250, opened 2025-04-28
un-support read operations on unfinalized objects

ashmeenkaur:
### Description
Return unsupported error for reads on unfinalized objects.

b/414336188

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - added

NA

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/v2/cfg"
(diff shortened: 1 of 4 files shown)


raj-prince (approved):

internal/fs/inode/file.go
```diff
@@ -384,8 +385,7 @@ func (f *FileInode) Source() *gcs.MinObject {
 func (f *FileInode) SourceGenerationIsAuthoritative() bool {
 	// Source generation is authoritative if:
 	//   1.  No pending writes exists on the inode (both content and bwh are nil).
-	//   2.  The bucket is zonal and there are no pending writes in the temporary file.
-	return (f.content == nil && f.bwh == nil) || (f.bucket.BucketType().Zonal && f.content == nil)
+	return f.content == nil && f.bwh == nil
```
meet2mky, line 388: Nit: Please revert the comment added above (L379-382)as it doesn't make sense now if other condition is removed. Wdyt?

meet2mky (approved):
LGTM

internal/fs/inode/file.go
ashmeenkaur: Thanks for the catch! This is a temporary change and the original condition (and thus the comment) will be restored soon. To avoid another review cycle, I'll leave the comment as is for now and it will make sense again once the related changes are reverted.

PR #3251, opened 2025-04-29
Sync file Inode at the time of BWH initialization

meet2mky:
### Description
Ensure implicit sync file inode at the time bwh is initialized. This ensures as soon as bwh is created we have the correct generation from the server. With this change Local File Tests from local_file package which were running started failing for ZB with streaming writes as those tests validate local file but in this case file is not local and it is promoted to generation backed inode as soon as created. Removed local_file package tests from running as part of streaming writes test as those require a large refactoring and error checking for multiple scenarios.

Note: Manually verified that ZB tests are passing.

[b/414501988](http://b/414501988)

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - Tested stale handle skipped tests in this [PR](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3104)

### DIFF
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -48,7 +48,8 @@ type UploadHandler struct {
 	// uploadError stores atomic pointer to the error seen by uploader.
 	uploadError atomic.Pointer[error]
 	// CancelFunc persisted to cancel the uploads in case of unlink operation.
-	cancelFunc context.CancelFunc
+	cancelFunc    context.CancelFunc
+	startUploader sync.Once
 
 	// Parameters required for creating a new GCS chunk writer.
 	bucket               gcs.Bucket
@@ -87,18 +88,14 @@ func newUploadHandler(req *CreateUploadHandlerRequest) *UploadHandler {
 func (uh *UploadHandler) Upload(block block.Block) error {
 	uh.wg.Add(1)
 
-	if uh.writer == nil {
-		// Lazily create the object writer.
-		err := uh.createObjectWriter()
-		if err != nil {
-			// createObjectWriter can only fail here due to throttling, so we will not
-			// handle this error explicitly or fall back to temp file flow.
-			return fmt.Errorf("createObjectWriter failed for object %s: %w", uh.objectName, err)
-		}
-		// Start the uploader goroutine.
-		go uh.uploader()
+	err := uh.ensureWriter()
+	if err != nil {
+		return fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
-
+	// Start the uploader goroutine but only once.
+	uh.startUploader.Do(func() {
+		go uh.uploader()
+	})
 	uh.uploadCh <- block
 	return nil
 }
@@ -126,6 +123,7 @@ func (uh *UploadHandler) UploadError() (err error) {
 func (uh *UploadHandler) uploader() {
 	for currBlock := range uh.uploadCh {
 		if uh.UploadError() != nil {
+			uh.freeBlocksCh <- currBlock
 			uh.wg.Done()
 			continue
 		}
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -122,22 +122,19 @@ func (uh *UploadHandler) UploadError() (err error) {
 // uploader is the single-threaded goroutine that uploads blocks.
 func (uh *UploadHandler) uploader() {
 	for currBlock := range uh.uploadCh {
-		if uh.UploadError() != nil {
-			uh.freeBlocksCh <- currBlock
-			uh.wg.Done()
-			continue
-		}
-		_, err := io.Copy(uh.writer, currBlock.Reader())
-		if errors.Is(err, context.Canceled) {
-			// Context canceled error indicates that the file was deleted from the
-			// same mount. In this case, we suppress the error to match local
-			// filesystem behavior.
-			err = nil
-		}
-		if err != nil {
-			logger.Errorf("buffered write upload failed for object %s: error in io.Copy: %v", uh.objectName, err)
-			err = gcs.GetGCSError(err)
-			uh.uploadError.Store(&err)
+		if uh.UploadError() == nil {
+			_, err := io.Copy(uh.writer, currBlock.Reader())
+			if errors.Is(err, context.Canceled) {
+				// Context canceled error indicates that the file was deleted from the
+				// same mount. In this case, we suppress the error to match local
+				// filesystem behavior.
+				err = nil
+			}
+			if err != nil {
+				logger.Errorf("buffered write upload failed for object %s: error in io.Copy: %v", uh.objectName, err)
+				err = gcs.GetGCSError(err)
+				uh.uploadError.Store(&err)
+			}
 		}
 		// Put back the uploaded block on the freeBlocksChannel for re-use.
 		uh.freeBlocksCh <- currBlock
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -123,6 +123,7 @@ func (uh *UploadHandler) UploadError() (err error) {
 func (uh *UploadHandler) uploader() {
 	for currBlock := range uh.uploadCh {
 		if uh.UploadError() != nil {
+			uh.freeBlocksCh <- currBlock
 			uh.wg.Done()
 			continue
 		}
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -122,19 +122,21 @@ func (uh *UploadHandler) UploadError() (err error) {
 // uploader is the single-threaded goroutine that uploads blocks.
 func (uh *UploadHandler) uploader() {
 	for currBlock := range uh.uploadCh {
-		if uh.UploadError() == nil {
-			_, err := io.Copy(uh.writer, currBlock.Reader())
-			if errors.Is(err, context.Canceled) {
-				// Context canceled error indicates that the file was deleted from the
-				// same mount. In this case, we suppress the error to match local
-				// filesystem behavior.
-				err = nil
-			}
-			if err != nil {
-				logger.Errorf("buffered write upload failed for object %s: error in io.Copy: %v", uh.objectName, err)
-				err = gcs.GetGCSError(err)
-				uh.uploadError.Store(&err)
-			}
+		if uh.UploadError() != nil {
+			uh.wg.Done()
+			continue
+		}
+		_, err := io.Copy(uh.writer, currBlock.Reader())
+		if errors.Is(err, context.Canceled) {
+			// Context canceled error indicates that the file was deleted from the
+			// same mount. In this case, we suppress the error to match local
+			// filesystem behavior.
+			err = nil
+		}
+		if err != nil {
+			logger.Errorf("buffered write upload failed for object %s: error in io.Copy: %v", uh.objectName, err)
+			err = gcs.GetGCSError(err)
+			uh.uploadError.Store(&err)
 		}
 		// Put back the uploaded block on the freeBlocksChannel for re-use.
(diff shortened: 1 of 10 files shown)


internal/fs/fs.go
```diff
@@ -1745,19 +1770,26 @@ func (fs *fileSystem) createLocalFile(ctx context.Context, parentID fuseops.Inod
...
+
+	// We need to release the filesystem lock before acquiring the inode lock.
 	fs.mu.Unlock()
+	fileInode.Lock()
+	err = fs.initBufferedWriteHandlerAndSyncFileIfEligible(ctx, fileInode)
+	fileInode.Unlock()
+	if err != nil {
+		return
+	}
+	err = fileInode.CreateEmptyTempFile(ctx)
```
ashmeenkaur, line 1785: IIUC fileInode lock must be required for this method too. Could you check?

ashmeenkaur (commented):
Yet to review inode test files

internal/bufferedwrites/upload_handler.go
```diff
@@ -125,21 +122,19 @@ func (uh *UploadHandler) UploadError() (err error) {
...
-			// Context canceled error indicates that the file was deleted from the
-			// same mount. In this case, we suppress the error to match local
-			// filesystem behavior.
-			err = nil
-		}
-		if err != nil {
-			logger.Errorf("buffered write upload failed for object %s: error in io.Copy: %v", uh.objectName, err)
-			err = gcs.GetGCSError(err)
-			uh.uploadError.Store(&err)
+		if uh.UploadError() == nil {
```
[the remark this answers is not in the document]
meet2mky: Thanks, reverted these changes.

internal/fs/fs.go
```diff
@@ -1214,6 +1214,24 @@ func (fs *fileSystem) syncFile(
 	return nil
 }
 
+// Initializes Buffered Write Handler if Eligible and synchronizes the file inode to GCS if initialized succeds.
```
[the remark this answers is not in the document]
meet2mky: Done!

internal/fs/inode/file.go
```diff
@@ -980,24 +967,25 @@ func (f *FileInode) CreateBufferedOrTempWriter(ctx context.Context) (err error)
 	return
 }
 
-func (f *FileInode) initBufferedWriteHandlerIfEligible(ctx context.Context) error {
+// Initializes Buffered Write Handler if the file inode is eligible and returns
+// initialized as true when the new instance of buffered writer handler is created.
+func (f *FileInode) InitBufferedWriteHandlerIfEligible(ctx context.Context) (initalized bool, err error) {
```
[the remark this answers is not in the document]
meet2mky: Thanks, done

internal/fs/fs.go
```diff
@@ -2597,11 +2629,14 @@ func (fs *fileSystem) WriteFile(
 	in.Lock()
 	defer in.Unlock()
 
-	// Serve the request.
-	if err := in.Write(ctx, op.Data, op.Offset); err != nil {
-		return err
+	// Initialize BWH if eligible and Sync file inode.
```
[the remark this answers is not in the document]
meet2mky: Removed the comment.

internal/fs/fs.go
meet2mky: Added the lock for now to ensure no locking behaviour change.

internal/fs/fs.go
```diff
@@ -1214,6 +1214,24 @@ func (fs *fileSystem) syncFile(
...
+//
+// LOCKS_EXCLUDED(fs.mu)
+// LOCKS_REQUIRED(f.mu)
+func (fs *fileSystem) initBufferedWriteHandlerAndSyncFileIfEligible(ctx context.Context, f *inode.FileInode) error {
+	initialized, err := f.InitBufferedWriteHandlerIfEligible(ctx)
+	if err != nil {
+		return err
+	}
+	if initialized {
+		err = fs.syncFile(ctx, f)
```
ashmeenkaur, line 1227: Sync should be done only for bwh and only for zonal buckets (Flush call). I think we need a new sync method here.

internal/fs/fs.go
```diff
@@ -1745,19 +1770,27 @@ func (fs *fileSystem) createLocalFile(ctx context.Context, parentID fuseops.Inod
...
 	child = fs.mintInode(core)
 	fs.localFileInodes[child.Name()] = child
-	// Empty file is created to be able to set attributes on the file.
 	fileInode := child.(*inode.FileInode)
-	if err := fileInode.CreateBufferedOrTempWriter(ctx); err != nil {
-		return nil, err
+	// Use deferred locking on filesystem so that it is locked before the defer call that unlocks the mutex and it doesn't fail.
+	defer fs.mu.Lock()
+	if !fs.newConfig.Write.EnableStreamingWrites {
+		err = fileInode.CreateEmptyTempFile(ctx)
```
ashmeenkaur, line 1777: we still need to create a temp file if src object size is not 0 (even when streaming writes are enabled). Can we keep the flow same as how it was earlier? Init bwh first based on the conditions in initBufferedWriteHandlerIfEligible and after that if bwh was not initialized, init empty temp file.

internal/fs/fs.go
meet2mky: I think, this syncFile will work fine for all of the cases since we have fileInode lock and bwh is set so it will invoke in.bwh.Sync() which is going to create an empty Flush for ZB and no-op for Non-ZB (newly created bwh doesn't have any pending buffers to upload)

internal/fs/fs.go
meet2mky: Fixed the flow to try init bwh first.

internal/bufferedwrites/upload_handler.go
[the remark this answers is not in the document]
meet2mky: changes done.

internal/fs/fs.go
ashmeenkaur: Can we add a comment mentioning this?

internal/fs/inode/file.go
```diff
@@ -962,11 +951,9 @@ func (f *FileInode) CacheEnsureContent(ctx context.Context) (err error) {
...
-func (f *FileInode) CreateBufferedOrTempWriter(ctx context.Context) (err error) {
-	err = f.initBufferedWriteHandlerIfEligible(ctx)
-	if err != nil {
-		return err
-	}
+// CreateEmptyTempFile creates an empty file with no contents when
+// streaming writes are not in enabled.
+func (f *FileInode) CreateEmptyTempFile(ctx context.Context) (err error) {
 	// Skip creating empty file when streaming writes are enabled.
 	if f.bwh != nil {
```
ashmeenkaur, line 958: Skip creating temp file if it's already created too (can happen due to concurrent create file calls). Earlier we had fs lock so it was not a problem

tools/integration_tests/streaming_writes/default_mount_local_file_test.go
```diff
@@ -49,7 +47,6 @@ func (t *defaultMountLocalFile) createLocalFile() {
 // Executes all tests that run with single streamingWrites configuration for localFiles.
 func TestDefaultMountLocalFileTest(t *testing.T) {
 	s := new(defaultMountLocalFile)
-	s.CommonLocalFileTestSuite.TestifySuite = &s.Suite
```
ashmeenkaur, line 52: My understanding is that we only need to skip these tests only for zonal buckets. Could you please confirm if that's correct?

internal/fs/fs.go
```diff
@@ -1517,6 +1535,13 @@ func (fs *fileSystem) SetInodeAttributes(
 	defer in.Unlock()
 	file, isFile := in.(*inode.FileInode)
 
+	if isFile {
```
ashmeenkaur, line 1538: Let's do this only if op.Mtime != nil or op.Size != nil (no need to create bwh for atime or mode updates

internal/fs/inode/file_streaming_writes_test.go
```diff
@@ -163,8 +167,8 @@ func TestFileStreamingWritesWithZonalBucketTestSuite(t *testing.T) {
 	suite.Run(t, new(FileStreamingWritesZonalBucketTest))
 }
 
-func (t *FileStreamingWritesZonalBucketTest) TestSourceGenerationIsAuthoritativeReturnsTrueForZonalBuckets() {
-	assert.True(t.T(), t.in.SourceGenerationIsAuthoritative())
+func (t *FileStreamingWritesZonalBucketTest) TestSourceGenerationIsAuthoritativeReturnsFalseForZonalBuckets() {
```
ashmeenkaur, line 170: This is not right IMO, source generation should be Authoritative unless writes are done on the file. Could you check why this is happening?

internal/fs/inode/file_streaming_writes_test.go
```diff
@@ -151,7 +151,11 @@ func (t *FileStreamingWritesCommon) createInode(fileName string, fileType string
 		EnableStreamingWrites: true,
 		GlobalMaxBlocks:       10,
 	}}
-
+	// Initialize BWH for inode created above.
+	initialized, err := t.in.InitBufferedWriteHandlerIfEligible(t.ctx)
```
ashmeenkaur, line 155: Can we add this to specific tests where bwh will be initialized by fs layer to keep the behavior closer to real scenario?

internal/fs/fs.go
meet2mky: Done added comment.

internal/fs/inode/file_streaming_writes_test.go
meet2mky: I don't think we can do that here as this inode layer.

tools/integration_tests/streaming_writes/default_mount_local_file_test.go
meet2mky: Let's do separation in follow up PR.

internal/fs/inode/file.go
meet2mky: Done

internal/fs/fs.go
meet2mky: Done

internal/fs/fs.go
```diff
@@ -1745,19 +1774,25 @@ func (fs *fileSystem) createLocalFile(ctx context.Context, parentID fuseops.Inod
...
-		return nil, err
-	}
+	// Use deferred locking on filesystem so that it is locked before the defer call that unlocks the mutex and it doesn't fail.
+	// We need to release the filesystem lock before acquiring the inode lock.
 	fs.mu.Unlock()
+	defer fs.mu.Lock()
+	fileInode.Lock()
+	err = fs.initBufferedWriteHandlerAndSyncFileIfEligible(ctx, fileInode)
+	if err != nil {
+		return
```
ashmeenkaur, line 1785: in case of errors also we should release file inode lock

meet2mky: Done

internal/fs/inode/file_test.go
```diff
@@ -1618,9 +1673,13 @@ func (t *FileTest) TestSetMtimeOnEmptyGCSFileWhenStreamingWritesAreEnabled() {
 func (t *FileTest) TestSetMtimeOnEmptyGCSFileAfterWritesWhenStreamingWritesAreEnabled() {
 	t.createInodeWithEmptyObject()
 	t.in.config = &cfg.Config{Write: *getWriteConfig()}
-	assert.Nil(t.T(), t.in.bwh)
+	// Initialize BWH for inode created above.
+	initalized, err := t.in.InitBufferedWriteHandlerIfEligible(t.ctx)
+	require.NoError(t.T(), err)
+	assert.True(t.T(), initalized)
+	assert.NotNil(t.T(), t.in.bwh)
```
ashmeenkaur, line 1680: let's create a test helper function for this

meet2mky: Done

ashmeenkaur (approved):

PR #3253, opened 2025-04-29
Forward env variables `GCE_METADATA_*` to gcsfuse daemon process in case of background run

gargnitingoogle:
### Description
Forward env variables `GCE_METADATA_HOST`, `GCE_METADATA_ROOT`, `GCE_METADATA_IP` to gcsfuse daemon process in case of background run. `GCE_METADATA_HOST` is used [here](https://github.com/googleapis/google-cloud-go/blob/a6c85f6387ee6aa291e786c882637fb03f3302f4/compute/metadata/metadata.go#L461) to determine where to send the metadata requests.

[b/414377352](http://b/414377352)

### Testing details
1. Manual - Tested manually that after this change, the env vars GCE_METADATA_HOST, GCE_METADATA_ROOT and GCE_METADATA_IP are being passed down the gcsfuse daemon process.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -338,6 +338,17 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
 				p)
 		}
 
+		// Forward GCE_METADATA_HOST and other related environment variables.
+		for _, envvar := range []string{"GCE_METADATA_HOST"} {
+			if p, ok := os.LookupEnv(envvar); ok {
+				env = append(env, fmt.Sprintf("%s=%s", envvar, p))
+				fmt.Fprintf(
+					os.Stdout,
+					"Added environment %s: %s\n",
+					envvar, p)
+			}
+		}
+
 		// Pass the parent process working directory to child process via
 		// environment variable. This variable will be used to resolve relative paths.
 		if parentProcessExecutionDir, err := os.Getwd(); err == nil {
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -322,13 +322,14 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
 				p)
 		}
 
-		// Forward GCE_METADATA_HOST and other related environment variables.
-		// Pass along GOOGLE_APPLICATION_CREDENTIALS, since we document in
+		// Forward environment variables.
+		// Forward GOOGLE_APPLICATION_CREDENTIALS, since we document in
 		// mounting.md that it can be used for specifying a key file.
-		// Pass through the no_proxy environment variable. Whenever
+		// Forward the no_proxy environment variable. Whenever
 		// using the http(s)_proxy environment variables. This should
 		// also be included to know for which hosts the use of proxies
 		// should be ignored.
+		// Forward GCE_METADATA_HOST as it is used for mocked metadata services.
 		for _, envvar := range []string{"GOOGLE_APPLICATION_CREDENTIALS", "no_proxy", "GCE_METADATA_HOST"} {
 			if envval, ok := os.LookupEnv(envvar); ok {
 				env = append(env, fmt.Sprintf("%s=%s", envvar, envval))
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -301,7 +301,7 @@ func forwardedEnvVars() []string {
 
 	// Here, parent process doesn't pass the $HOME to child process implicitly,
 	// hence we need to pass it explicitly.
-	if homeDir, _ := os.UserHomeDir(); true {
+	if homeDir, err := os.UserHomeDir(); err == nil {
 		env = append(env, fmt.Sprintf("HOME=%s", homeDir))
 	}
 
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -249,6 +249,9 @@ func isDynamicMount(bucketName string) bool {
 	return bucketName == "" || bucketName == "_"
 }
 
+// forwardedEnvVars collects and returns all the environment
+// variables which should be sent to the gcsfuse daemon
+// process in case of background run.
 func forwardedEnvVars() []string {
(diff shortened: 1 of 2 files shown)


cmd/legacy_main.go
```diff
@@ -249,6 +249,67 @@ func isDynamicMount(bucketName string) bool {
...
+	// Pass the parent process working directory to child process via
+	// environment variable. This variable will be used to resolve relative paths.
+	if parentProcessExecutionDir, err := os.Getwd(); err == nil {
+		env = append(env, fmt.Sprintf("%s=%s", util.GCSFUSE_PARENT_PROCESS_DIR,
+			parentProcessExecutionDir))
+	}
+
+	// Here, parent process doesn't pass the $HOME to child process implicitly,
+	// hence we need to pass it explicitly.
+	if homeDir, _ := os.UserHomeDir(); true {
```
gargnitingoogle, line 302: The condition here was `err != nil` which was always true. Since this function-scope doesn't have an `err` variable, I replaced this redundant check with just `true` to avoid creating an additional boolean variable. An even cleaner code would be remove the if-else completely, but that'll make this code-change harder to recognize in this PR. So, keeping the if-else structure.

cmd/legacy_main.go
```diff
@@ -249,6 +249,67 @@ func isDynamicMount(bucketName string) bool {
 	return bucketName == "" || bucketName == "_"
 }
 
+func forwardedEnvVars() []string {
```
raj-prince, line 252: Can we write a comment? Something like,

```
// forwardedEnvVars returns all the enviornment variables which should be forwareded to
// daemon process when gcsfuse is mounted in background mode.
```

cmd/legacy_main.go
raj-prince:
```suggestion
	if homeDir, err := os.UserHomeDir(); err == nil {
```

cmd/legacy_main.go
```diff
@@ -249,6 +249,67 @@ func isDynamicMount(bucketName string) bool {
...
+			p)
+	} else if p, ok := os.LookupEnv("http_proxy"); ok {
+		env = append(env, fmt.Sprintf("http_proxy=%s", p))
+		fmt.Fprintf(
+			os.Stdout,
+			"Added environment http_proxy: %s\n",
+			p)
+	}
+
+	// Forward environment variables.
```
raj-prince, line 275:
```suggestion
```

cmd/legacy_main.go
```diff
@@ -249,6 +249,67 @@ func isDynamicMount(bucketName string) bool {
...
+	// Forward GOOGLE_APPLICATION_CREDENTIALS, since we document in
+	// mounting.md that it can be used for specifying a key file.
+	// Forward the no_proxy environment variable. Whenever
+	// using the http(s)_proxy environment variables. This should
+	// also be included to know for which hosts the use of proxies
+	// should be ignored.
+	// Forward GCE_METADATA_HOST as it is used for mocked metadata services.
```
raj-prince, line 282: Can we write about all three ROOT and IP here, why other two is required?

cmd/legacy_main.go
```diff
@@ -300,61 +361,7 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
...
-		// hence we need to pass it explicitly.
-		if homeDir, _ := os.UserHomeDir(); err == nil {
-			env = append(env, fmt.Sprintf("HOME=%s", homeDir))
-		}
-
-		// This environment variable will be helpful to distinguish b/w the main
-		// process and daemon process. If this environment variable set that means
-		// programme is running as daemon process.
-		env = append(env, fmt.Sprintf("%s=true", logger.GCSFuseInBackgroundMode))
+		env := forwardedEnvVars()
```
raj-prince, line 364: If possible, can we write unit test for this method. We don't need to write exhaustive test. Just for happy case.

Test set up would be.
(a) Set all the required environemnt variables.
(b) Call this method.
(c) Expect all the envs present in the returned array.

cmd/legacy_main.go
gargnitingoogle: The error returned from `os.UserHomeDir()` was ignored earlier, so this would be reintroducing that error-check. I don't know if this was ignored on purpose.

cmd/legacy_main.go
```diff
@@ -326,16 +321,23 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
...
 		// also be included to know for which hosts the use of proxies
 		// should be ignored.
-		if p, ok := os.LookupEnv("no_proxy"); ok {
-			env = append(env, fmt.Sprintf("no_proxy=%s", p))
-			fmt.Fprintf(
-				os.Stdout,
-				"Added environment no_proxy: %s\n",
-				p)
+		// Forward GCE_METADATA_HOST as it is used for mocked metadata services.
+		for _, envvar := range []string{"GOOGLE_APPLICATION_CREDENTIALS", "no_proxy", "GCE_METADATA_HOST", "GCE_METADATA_ROOT", "GCE_METADATA_IP"} {
```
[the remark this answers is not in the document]
gargnitingoogle: outdated.

cmd/legacy_main.go
gargnitingoogle: Oh, this was a miss on my part. Originally I was only adding GCE_METADATA_HOST, but added the other two later based on the full conversation in the customer support ticket. But forgot to update this comment. Will do now,

cmd/legacy_main.go
gargnitingoogle: done

cmd/legacy_main.go
gargnitingoogle: done

cmd/legacy_main.go
gargnitingoogle: I'm going with your suggestion because it makes sense.

raj-prince (approved):

PR #3254, opened 2025-04-29
Adding simulated clock implementation for time.After method

raj-prince:
### Description
- Adding a simulated clock for time.After() method.
- Current fake_clock implementation doesn't provide the SetTime() and AdvanceTime() support with time.After implementation.

b/414538309

### Testing details
1. Manual - NA
2. Unit tests - Automated
3. Integration tests - NA

### DIFF
--- a/internal/clock/simulated_clock.go
+++ b/internal/clock/simulated_clock.go
@@ -37,7 +37,7 @@ type SimulatedClock struct {
 func NewSimulatedClock(startTime time.Time) *SimulatedClock {
 	return &SimulatedClock{
 		t:       startTime,
-		pending: make([]*afterRequest, 0),
+		pending: nil,
 	}
 }
 
@@ -111,7 +111,5 @@ func (sc *SimulatedClock) processPending() {
 		}
 	}
 
-	if len(stillPending) < len(sc.pending) { // Optimization: only re-slice if necessary
-		sc.pending = stillPending
-	}
+	sc.pending = stillPending
 }
--- a/internal/clock/simulated_clock.go
+++ b/internal/clock/simulated_clock.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google Inc. All Rights Reserved.
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -68,8 +68,8 @@ func (sc *SimulatedClock) AdvanceTime(d time.Duration) {
 	sc.processPending()
 }
 
-// After returns a time channel and returns the time over the returned channel after
-// the provided duration.
+// After returns a time channel and the expectedFired time is sent over the returned
+// channel after the provided duration.
 // If the duration is zero or negative, it sends the current simulated time immediately.
 func (sc *SimulatedClock) After(d time.Duration) <-chan time.Time {
 	sc.mu.Lock()
--- a/internal/clock/simulated_clock.go
+++ b/internal/clock/simulated_clock.go
@@ -1,0 +1,119 @@
+// Copyright 2025 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clock
+
+import (
+	"sync"
+	"time"
+)
+
+// afterRequest holds the information for a pending After call in SimulatedClock.
+type afterRequest struct {
+	targetTime time.Time
+	ch         chan time.Time
+}
+
+// SimulatedClock is a clock that allows for manipulation of the time,
+// which does not change unless AdvanceTime or SetTime is called.
+// The zero value is a clock initialized to the zero time.
+type SimulatedClock struct {
+	mu      sync.RWMutex
+	t       time.Time       // GUARDED_BY(mu)
+	pending []*afterRequest // GUARDED_BY(mu)
+}
+
+func NewSimulatedClock(startTime time.Time) *SimulatedClock {
+	return &SimulatedClock{
+		t:       startTime,
+		pending: make([]*afterRequest, 0),
+	}
+}
+
+func (sc *SimulatedClock) Now() time.Time {
+	sc.mu.RLock()
+	defer sc.mu.RUnlock()
+
+	return sc.t
+}
+
+// SetTime sets the current time according to the clock.
+// It also processes any pending After calls that should fire.
+func (sc *SimulatedClock) SetTime(t time.Time) {
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+
+	sc.t = t
+	sc.processPending()
+}
+
+// AdvanceTime advances the current time according to the clock by the supplied duration.
+// It also processes any pending After calls that should fire.
+func (sc *SimulatedClock) AdvanceTime(d time.Duration) {
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+
+	sc.t = sc.t.Add(d)
+	sc.processPending()
+}
+
+// After returns a time channel and returns the time over the returned channel after
+// the provided duration.
+// If the duration is zero or negative, it sends the current simulated time immediately.
+func (sc *SimulatedClock) After(d time.Duration) <-chan time.Time {
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+
+	ch := make(chan time.Time, 1)
+	effectiveTargetTime := sc.t.Add(d)
+
+	// If duration is non-positive, fire immediately with the current time.
+	// Or if the target time is not after the current time (e.g. d <= 0)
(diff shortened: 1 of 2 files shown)


abhishek10004 (approved):
LGTM

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			clock := NewSimulatedClock(referenceTime)
+
+			ch := clock.After(tc.afterDuration)
+			require.NotNil(t, ch, "Channel should not be nil")
+
+			// Perform the action (if any) that might trigger the timer
+			if tc.action != nil {
```
kislaykishore, line 271: Let's remove this check if possible - I don't see this function being set to nil and the lesser the number of conditionals in test code, the better it is.

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,279 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			clock := NewSimulatedClock(referenceTime)
+			clockTimeAtAfterCall := clock.Now()
+
+			ch := clock.After(tc.afterDuration)
+			require.NotNil(t, ch, "Channel should not be nil")
+
+			// Perform the action (if any) that might trigger the timer
+			if tc.action != nil {
```
kislaykishore, line 221: Let's remove this check if possible - I don't see this function being set to nil and the lesser the number of conditionals in test code, the better it is.

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+
+			// Determine the expected time on the channel if it fires
+			// For zero/negative duration, it's the time After() was called.
+			// For positive, it's that time + duration.
+			var expectedFireTimeOnChannel time.Time
+			if tc.afterDuration <= 0 {
+				expectedFireTimeOnChannel = clockTimeAtAfterCall
+			} else {
+				expectedFireTimeOnChannel = clockTimeAtAfterCall.Add(tc.afterDuration)
```
kislaykishore, line 232: Can we have two separate tests instead of having if-else in test code.

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,279 @@
...
+		name         string
+		initialTime  time.Time
+		advanceBy    time.Duration
+		expectedTime time.Time
+	}{
+		{
+			name:         "AdvancePositiveDuration",
+			initialTime:  referenceTime,
+			advanceBy:    5 * time.Minute,
+			expectedTime: referenceTime.Add(5 * time.Minute),
```
kislaykishore, line 146: Can we make tests damp and dumb and not have logic in them - https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html and https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html.

I'd encourage to get rid of referenceTime and use actual time throughout.

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,279 @@
...
+	fireTestTimeout  = 50 * time.Millisecond // When expecting a channel to fire
+)
+
+// Helper to assert that a channel receives a specific time within a timeout.
+func assertReceivesTime(t *testing.T, ch <-chan time.Time, expectedTime time.Time, timeout time.Duration, msgAndArgs ...interface{}) {
+	t.Helper()
+	select {
+	case actualTime := <-ch:
+		assert.True(t, expectedTime.Equal(actualTime), "Received time %v, expected %v. %s", actualTime, expectedTime, msgAndArgs)
+	case <-time.After(timeout): // Using real time.After for test timeout
```
kislaykishore, line 38: While I understand the intention of this, Golang already comes with a test timeout feature with a default of 10 minutes. I'd argue that in the interest of making tests simple, we just have `<- ch` in the assert block.

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,279 @@
...
+		t.Fatalf("Timeout waiting for time on channel. Expected %v. %s", expectedTime, msgAndArgs)
+	}
+}
+
+// Helper to assert that a channel does NOT receive a time within a short duration.
+func assertNotReceivesTime(t *testing.T, ch <-chan time.Time, timeout time.Duration, msgAndArgs ...interface{}) {
```
kislaykishore, line 44: Should we move this select block to the main test? I don't think we need a separate method for this.

internal/clock/simulated_clock.go
```diff
@@ -0,0 +1,117 @@
...
+type SimulatedClock struct {
+	mu      sync.RWMutex
+	t       time.Time       // GUARDED_BY(mu)
+	pending []*afterRequest // GUARDED_BY(mu)
+}
+
+func NewSimulatedClock(startTime time.Time) *SimulatedClock {
+	return &SimulatedClock{
+		t:       startTime,
+		pending: make([]*afterRequest, 0),
```
kislaykishore, line 40: Why not start with a nil value?
```suggestion
		pending: nil,
```

internal/clock/simulated_clock.go
kislaykishore, line 1: Should we have this in a common place as opposed to internal? Perhaps in a testutil package at the top level?

internal/clock/simulated_clock.go
```diff
@@ -0,0 +1,117 @@
...
+			ar.ch <- ar.targetTime // Send the time it was scheduled to fire
+			// Do not close the channel, to mimic time.After behavior
+		} else {
+			stillPending = append(stillPending, ar)
+		}
+	}
+
+	if len(stillPending) < len(sc.pending) { // Optimization: only re-slice if necessary
+		sc.pending = stillPending
+	}
```
kislaykishore, line 116: I'm not sure of the optimization - slice assignment doesn't do an actual copy - it just assigns the pointer. So, I doubt it's worth the effort.

internal/clock/simulated_clock_test.go
raj-prince: I still feel, we should give per test timeout, other wise we will get to know after 10 minutes. Discussed offline, resolving this.

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,287 @@
...
+		{
+			name:             "SetFromZeroTime",
+			initialTimeSetup: func(sc *SimulatedClock) {},
+			timeToSet:        time.Date(2020, time.January, 1, 12, 0, 0, 0, time.UTC),
+			expectedAfter:    time.Date(2020, time.January, 1, 12, 0, 0, 0, time.UTC),
+		},
+		{
+			name: "OverwriteExistingTime",
+			initialTimeSetup: func(sc *SimulatedClock) {
+				sc.SetTime(time.Date(2020, time.January, 1, 12, 0, 0, 0, time.UTC).Add(-time.Hour)) // Start with a different time
```
kislaykishore, line 87:
```suggestion
				sc.SetTime(time.Date(2020, time.January, 1, 11, 0, 0, 0, time.UTC)
```

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			clock := NewSimulatedClock(time.Date(2020, time.January, 1, 12, 0, 0, 0, time.UTC))
+			tc.initialTimeSetup(clock)
+
+			now := clock.Now()
+
+			assert.True(t, now.Equal(tc.expectedTime), "clock.Now() returned %v, expected %v", now, tc.expectedTime)
```
kislaykishore, line 66: Ditto elsewhere. Apologies for not catching this in the first pass.
```suggestion
			assert.Equal(t, tc.expectedTime), now)
```

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,287 @@
...
+			afterDuration: 10 * time.Second,
+			action: func(sc *SimulatedClock) {
+				sc.AdvanceTime(5 * time.Second) // Advance, but not enough
+			},
+		},
+		{
+			name:          "PositiveDuration_DoesNotFire_WhenTimeSetBeforeDuration",
+			afterDuration: 10 * time.Second,
+			action: func(sc *SimulatedClock) {
+				sc.SetTime(time.Date(2020, time.January, 1, 12, 0, 0, 0, time.UTC).Add(5 * time.Second)) // Set time, but not enough
```
kislaykishore, line 263:
```suggestion
				sc.SetTime(time.Date(2020, time.January, 1, 12, 0, 0, 5, time.UTC)) // Set time, but not enough
```

internal/clock/simulated_clock_test.go
```diff
@@ -0,0 +1,287 @@
...
+			afterDuration: 10 * time.Second,
+			action: func(sc *SimulatedClock) {
+				sc.AdvanceTime(15 * time.Second) // Advance well past the duration
+			},
+		},
+		{
+			name:          "PositiveDuration_Fires_WhenTimeSetPastDuration",
+			afterDuration: 10 * time.Second,
+			action: func(sc *SimulatedClock) {
+				sc.SetTime(time.Date(2020, time.January, 1, 12, 0, 0, 0, time.UTC).Add(15 * time.Second)) // Set time well past the duration
```
kislaykishore, line 217:
```suggestion
				sc.SetTime(time.Date(2020, time.January, 1, 12, 0, 0, 15, time.UTC)) // Set time well past the duration
```

kislaykishore (approved):

PR #3256, opened 2025-04-29
Pass "x-goog-user-project" to GetStorageLayout call to support `requester-pays` buckets

gargnitingoogle:
### Description
Pass "x-goog-user-project" to GetStorageLayout call to support `requester-pays` buckets

[b/406090515](http://b/406090515), [b/405913611](http://b/405913611)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


abhishek10004:
You would have to ensure that all control client API calls handle billing project. I tried getFolder call and it failed with the same error.

gargnitingoogle:
This PR also needs changes for other APIs called through storage-control-client, so marking this draft for now. Will have to rethink this change.

gargnitingoogle:
For some weird reason, this change is failing emulator_tests consistently (failing to mount bucket), even though those tests don't pass `--billing-project` or use `requester-pays` buckets in any way. So, it's not sure what's wrong with those tests or this change.

gargnitingoogle:
> For some weird reason, this change is failing emulator_tests consistently (with the error `"Failed to mount GCSFuse: cannot mount gcsfuse: exit status 1"`), even though those tests don't pass `--billing-project` or use `requester-pays` buckets in any way. So, it's not sure what's wrong with those tests or this change.

The issue was identified and it was fixed last week.

internal/storage/storage_handle.go
```diff
@@ -261,12 +259,19 @@ func (sh *storageClient) getStorageLayout(bucketName string) (*controlpb.Storage
 	return stoargeLayout, err
 }
 
+func createStorageControlClientWithBillingProject(controlClient StorageControlClient, billingProject string) StorageControlClient {
+	if len(strings.TrimSpace(billingProject)) > 0 {
```
abhishek10004, line 263: Let's just check for empty string here (like we are doing at line 352). Our flag parsing logic should handle the extra spaces. If it's not doing that, then we should handle it there.

abhishek10004:
Have left a couple of minor comments.

Also, you mentioned the following in your PR description:
>Moving the interface StorageControlClient from package internal/storage to internal/storage/storageutil to avoid cyclicity in package importing

I don't see this in the changes. Could you please check & update the description.

internal/storage/storage_handle.go
gargnitingoogle: done

internal/storage/storage_handle.go
```diff
@@ -280,6 +285,8 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
 		if err != nil {
 			return nil, fmt.Errorf("could not create StorageControl Client: %w", err)
 		}
+		// special handling for requester-pays buckets and for mounts created with custom billing projects.
+		controlClient = createStorageControlClientWithBillingProject(controlClient, billingProject)
```
charith87, line 289: Overall looks fine with the approach. But should this function be appendBillingProject instead of createStorageControlClientWithBillingProject since there is no creation of Control Client inside it for Readability Sake.

charith87: or WrapBillingProject

gargnitingoogle: done

PR #3260, opened 2025-04-30
Make NewReaderWithReadHandle to return fullGCSReaderCloser

raj-prince:
### Description
**Problem:**
The NewReaderWithReadHandle function currently returns a storageReader that, when using gRPC, requires approximately 64 separate read operations to fill a 1 MiB buffer. This inefficiency stems from the underlying Go gRPC stream implementation, which yields data in small 4KB chunks per read call.

Furthermore, the issue can be magnified because the storageReader might be wrapped by intermediate layers (like monitoring, debugging, or throttling readers). Each of these wrappers potentially processes every one of the numerous small read calls, adding cumulative overhead. Any new addition of wrapper will cause the similar overhead.

**Solution:**
Although a previous fix [PR#2993](https://github.com/GoogleCloudPlatform/gcsfuse/pull/2993) introduced an io.Reader implementation intended to read the entire buffer in fewer calls, but that is only limited when monitoring is enabled, also intermediate wrapper in gcsfuse still face the similar issue.

Making the fix complete by moving the above mentioned reader instance as a first GCSFuse wrapper, so that subsequent wrappers (higher layers e.g., monitoring, debugging, throttling) doesn't need to do anything.

**Why Safe:**
The full_reader_closer implementation still follows the [io.Reader](https://pkg.go.dev/io#Reader) semantics other than returning quickly with partial data. So, there will not be any change from the caller perspective.

User of reader in gcsfuse:
(1) Random reader in generic read flow.
(2) Async job downloader in the file-cache read flow.
(3)  Inode downloads data on disk in the write (modify) flow.

**Cleanup post changes:**
- Currently random_reader.go performs [full_read](https://github.com/GoogleCloudPlatform/gcsfuse/blob/b4f20f865274d52fadf14860d9cbf50e5cd47af3/internal/gcsx/random_reader.go#L444) which might be simplified by just making the reader.Read() call.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


raj-prince:
Perf:

image

vadlakondaswetha (approved):

PR #3261, opened 2025-05-02
feat: inactive timeout support for range reader

raj-prince:
### Description

This PR introduces a new reader wrapper, inactiveTimeoutReader, to automatically close open GCS read streams after a timeout. Useful for gRPC client-protocol where total number of max-streams is limited and surpassing the limit leads to stuck/hang.

More concurrent tests will be implemented in the follow up PR.

b/412465587

### Testing details
1. Manual - Functional testing using a python script which starts reads, idle for more than timeout and then read again.
2. Unit tests - added.
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/inactive_timeout_reader.go
+++ b/internal/gcsx/inactive_timeout_reader.go
@@ -54,10 +54,10 @@ type inactiveTimeoutReader struct {
 	gcsReader gcs.StorageReader
 
 	// Total number of bytes successfully read so far.
-	seen int64
+	seen uint64
 
 	// Requested range [start, end) from this reader.
-	start, end int64
+	start, end uint64
 
 	// The read handle used for efficient reconnection for zonal bucket.
 	readHandle []byte
@@ -76,6 +76,8 @@ type inactiveTimeoutReader struct {
 
 	// Used for waiting for timeout (helps us in mocking the functionality).
 	clock clock.Clock
+
+	timeout time.Duration
 }
 
 var (
@@ -89,53 +91,53 @@ var (
 //
 // If the timeout duration is zero, it returns (nil, ErrZeroInactivityTimeout) as a zero timeout
 // defeats the purpose of this wrapper.
-func NewInactiveTimeoutReader(ctx context.Context, bucket gcs.Bucket, object *gcs.MinObject, readHandle []byte, start int64, end int64, timeout time.Duration) (gcs.StorageReader, error) {
-	if timeout == time.Duration(0) {
-		return nil, ErrZeroInactivityTimeout
-	}
-
-	return NewInactiveTimeoutReaderWithClock(ctx, bucket, object, readHandle, start, end, timeout, clock.RealClock{})
+func NewInactiveTimeoutReader(ctx context.Context, bucket gcs.Bucket, object *gcs.MinObject, readHandle []byte, byteRange gcs.ByteRange, timeout time.Duration) (gcs.StorageReader, error) {
+	return NewInactiveTimeoutReaderWithClock(ctx, bucket, object, readHandle, byteRange, timeout, clock.RealClock{})
 }
 
-func NewInactiveTimeoutReaderWithClock(ctx context.Context, bucket gcs.Bucket, object *gcs.MinObject, readHandle []byte, start int64, end int64, timeout time.Duration, clock clock.Clock) (gcs.StorageReader, error) {
-	if timeout == time.Duration(0) {
+func NewInactiveTimeoutReaderWithClock(ctx context.Context, bucket gcs.Bucket, object *gcs.MinObject, readHandle []byte, byteRange gcs.ByteRange, timeout time.Duration, clock clock.Clock) (gcs.StorageReader, error) {
+	if timeout == 0 {
 		return nil, ErrZeroInactivityTimeout
 	}
 
-	tsr := &inactiveTimeoutReader{
+	itr := &inactiveTimeoutReader{
 		object:        object,
 		bucket:        bucket,
-		start:         start,
-		end:           end,
+		start:         byteRange.Start,
+		end:           byteRange.Limit,
 		parentContext: ctx,
 		readHandle:    readHandle,
 		mu:            locker.New("inactiveTimeoutReader: "+object.Name, func() {}),
 		isActive:      false,
 		stopChan:      make(chan struct{}),
 		clock:         clock,
+		timeout:       timeout,
 	}
 
 	var err error
-	tsr.gcsReader, err = tsr.createGCSReader()
-	if err == nil {
-		go tsr.monitor(timeout)
+	if itr.gcsReader, err = itr.createGCSReader(); err != nil {
+		return nil, err
 	}
-	return tsr, err
+
+	// Start the background periodic routine.
+	go itr.monitor()
+
+	return itr, err
 }
 
-// createGCSReader is a helper method to create the underlined reader from tsr.start + tsr.seen offset.
-func (tsr *inactiveTimeoutReader) createGCSReader() (gcs.StorageReader, error) {
-	reader, err := tsr.bucket.NewReaderWithReadHandle(
-		tsr.parentContext,
+// createGCSReader is a helper method to create the underlined reader from itr.start + itr.seen offset.
+func (itr *inactiveTimeoutReader) createGCSReader() (gcs.StorageReader, error) {
+	reader, err := itr.bucket.NewReaderWithReadHandle(
+		itr.parentContext,
 		&gcs.ReadObjectRequest{
-			Name:       tsr.object.Name,
-			Generation: tsr.object.Generation,
+			Name:       itr.object.Name,
+			Generation: itr.object.Generation,
 			Range: &gcs.ByteRange{
-				Start: uint64(tsr.start + tsr.seen),
-				Limit: uint64(tsr.end),
+				Start: itr.start + itr.seen,
+				Limit: itr.end,
 			},
-			ReadCompressed: tsr.object.HasContentEncodingGzip(),
-			ReadHandle:     tsr.readHandle,
+			ReadCompressed: itr.object.HasContentEncodingGzip(),
+			ReadHandle:     itr.readHandle,
 		})
 	if err != nil {
 		return nil, fmt.Errorf("NewReaderWithReadHandle: %w", err)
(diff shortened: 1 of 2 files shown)


internal/gcsx/inactive_timeout_reader_test.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func (s *InactiveTimeoutReaderTestSuite) TearDownTest() {
+	if s.reader != nil {
+		// Close the wrapper reader, not the potentially nil internal one
+		s.reader.Close()
+		s.reader = nil
+	}
+	s.mockBucket.AssertExpectations(s.T())
+}
```
kislaykishore, line 73:
```suggestion
func (s *InactiveTimeoutReaderTestSuite) TearDownTest() {
	s.mockBucket.AssertExpectations(s.T())
	if s.reader == nil {
	    return
	 }
         // Close the wrapper reader, not the potentially nil internal one
	s.reader.Close()
	s.reader = nil
}
}
```

internal/gcsx/inactive_timeout_reader_test.go
```diff
@@ -0,0 +1,310 @@
...
+	initialFakeReader *fake.FakeReader
+	timeout           time.Duration
+	simulatedClock    *clock.SimulatedClock
+}
+
+func TestInactiveTimeoutReaderTestSuite(t *testing.T) {
+	suite.Run(t, new(InactiveTimeoutReaderTestSuite))
+}
+
+func (s *InactiveTimeoutReaderTestSuite) SetupTest() {
```
kislaykishore, line 53: [Avoid suites if possible.](https://google.github.io/styleguide/go/best-practices.html#keep-setup-code-scoped-to-specific-tests)

internal/gcsx/inactive_timeout_reader_test.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func (s *InactiveTimeoutReaderTestSuite) Test_NewInactiveTimeoutReader_ZeroTimeoutError() {
+	s.initialData = []byte("zero timeout")
+	s.timeout = 0 * time.Millisecond // Zero timeout
+
+	_, err := NewInactiveTimeoutReader(s.ctx, s.mockBucket, s.object, []byte{}, 0, 100, s.timeout)
+
+	s.Error(err)
+	s.Equal(ErrZeroInactivityTimeout, err)
```
kislaykishore, line 128: Ditto elsewhere
```suggestion
	s.EqualError(ErrZeroInactivityTimeout, err)
```

internal/gcsx/inactive_timeout_reader_test.go
```diff
@@ -0,0 +1,310 @@
...
+	// EOF
+	{
+		buf3 := make([]byte, 6)
+
+		n3, err3 := s.reader.Read(buf3)
+
+		s.Same(io.EOF, err3)
+		s.Equal(0, n3)
+	}
+}
```
kislaykishore, line 171: We are testing three different things here. Let's break it down as:
* Test_InitialRead_NoError
* Test_NoReadCloserWithinTimeout
* Test_ReadFull_Succeeds

internal/gcsx/inactive_timeout_reader_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+func (s *InactiveTimeoutReaderTestSuite) Test_Read_ReconnectFails() {
```
kislaykishore, line 173: Similar for this - let's break it down. Also, we don't need to verify the initial-read works in all tests btw.

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,238 @@
...
+		tsr.gcsReader, err = tsr.createGCSReader()
+
+		if err != nil {
+			return 0, err
+		}
+	}
+
+	n, err = tsr.gcsReader.Read(p)
+	tsr.seen += int64(n)
+	return
```
kislaykishore, line 173: Avoid naked returns.

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if tsr.gcsReader == nil {
+		return tsr.readHandle
+	}
+
+	return tsr.gcsReader.ReadHandle()
+}
+
+// monitor runs in a background goroutine, and checks for inactivity.
+func (tsr *inactiveTimeoutReader) monitor(timeout time.Duration) {
```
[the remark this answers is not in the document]
raj-prince: renamed it to itr, initially the name was timedStorageReader.

internal/gcsx/inactive_timeout_reader_test.go
raj-prince: Using ErrorIs() instead, as second argument expects an string.

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,256 @@
...
+	mu locker.Locker
+
+	// Flag set by Read and reset by the monitor goroutine to track activity within the timeout window.
+	isActive bool
+
+	// Channel used to signal the monitor goroutine to stop, typically during Close().
+	stopChan chan struct{}
+
+	// Used for waiting for timeout (helps us in mocking the functionality).
+	clock clock.Clock
```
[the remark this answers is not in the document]
raj-prince: Yeah, that's right. Initially, timeout was not part of struct. I kept it as part of struct to avoid passing the timeout in monitor(), handleTimeout() and closeGCSReader() method. Also, we will not save any memory here, as anyway these values will be for the complete lifecycle for the wrapper.

I feel, it's okay to keep as part of struct in favor of not passing the parameter via method, as anyway we will not get any memory improvements.

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,256 @@
...
+//     protected by an internal mutex.
+type inactiveTimeoutReader struct {
+	object *gcs.MinObject
+	bucket gcs.Bucket
+
+	// The underlying GCS storage reader; nil if closed due to inactivity.
+	gcsReader gcs.StorageReader
+
+	// Total number of bytes successfully read so far.
+	seen uint64
```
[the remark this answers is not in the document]
raj-prince: None, of the problem will hit in our case. Specially, tricky conversion mentioned in the above blog will not be an issue for go-lang.

The reason, I kept it uint64 as at the end we need to pass uint64 to gcs.ByteRange{}, so somewhere we need to convert into uint64 and still problem can occur in case negative int64.

internal/gcsx/inactive_timeout_reader.go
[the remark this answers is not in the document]
raj-prince: Discussed offline, removing the fields from the struct - as okay to pass the value in the internal methods (although looks weird semantically `closeGCSReader(timeout)`),

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,245 @@
...
+func (itr *inactiveTimeoutReader) monitor(clock clock.Clock, timeout time.Duration) {
+	timer := clock.After(timeout)
+	for {
+		select {
+		case <-timer:
+			itr.handleTimeout()
+			timer = clock.After(timeout)
+		case <-itr.ctx.Done():
+			return
+		}
```
kislaykishore, line 211: WDYT?
```suggestion
	for {
		select {
		case <-clock.After(timeout):
			itr.handleTimeout()
		case <-itr.ctx.Done():
			return
		}
```

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,245 @@
...
+	if itr.gcsReader == nil {
+		return nil
+	}
+
+	err = itr.gcsReader.Close()
+	itr.gcsReader = nil
+	if err != nil {
+		return fmt.Errorf("close reader: %w", err)
+	}
+	return err
```
Tulsishah, line 184: nit: return nil?

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,245 @@
...
+			Name:       itr.object.Name,
+			Generation: itr.object.Generation,
+			Range: &gcs.ByteRange{
+				Start: itr.reqRange.Start + itr.seen,
+				Limit: itr.reqRange.Limit,
+			},
+			ReadCompressed: itr.object.HasContentEncodingGzip(),
+			ReadHandle:     itr.readHandle,
+		})
+	if err != nil {
```
Tulsishah, line 131: Do we want to handle clobbered error here?
e.g.
```
var notFoundError *gcs.NotFoundError
	if errors.As(err, &notFoundError) {
		err = &gcsfuse_errors.FileClobberedError{
			Err: fmt.Errorf("NewReader: %w", err),
		}
		return
	}
```

internal/gcsx/inactive_timeout_reader.go
```diff
@@ -0,0 +1,245 @@
...
+	itr.isActive = true
+
+	if itr.gcsReader == nil {
+		if itr.gcsReader, err = itr.createGCSReader(); err != nil {
+			return 0, err
+		}
+	}
+
+	n, err = itr.gcsReader.Read(p)
+	itr.seen += uint64(n)
```
Tulsishah, line 161: I understood if err!= nil then n will be zero but should we explicitly return before if error is not nit for better readability? Also what if error is EOF?

abhishek10004 (approved):

internal/gcsx/inactive_timeout_reader.go
raj-prince: Nope, that will handled by the wrapper instance, that means in the random reader while integration.

internal/gcsx/inactive_timeout_reader.go
raj-prince: will take as part of next PR, concurrent-unit test PR.

internal/gcsx/inactive_timeout_reader.go
raj-prince: Semantically, there will not be any change from the underlined error to this wrapper. So, all the error will be handled similar to the underlined error.

Returning before will required an extra check for err != EOF, as Don't want to add an extra error check in the Read operation, unless important.

PR #3262, opened 2025-05-02
Early termination in Test_ListWithMoveDir on failure in renameFolder

gargnitingoogle:
### Description
- Switching failure in folder renames from soft-failures to fatal-asserts as they should be.
- Extended it to all failures in concurrent_operations where proceeding after failure is pointless.

This is to reduce unnecessary repeated failures after a fatal failure.

[b/414503184](http://b/414503184)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Tested with [presubmit](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/1f5807ab-c2bc-4598-b778-2fc9b8198b1c/summary) .

### DIFF
--- a/tools/integration_tests/concurrent_operations/concurrent_listing_test.go
+++ b/tools/integration_tests/concurrent_operations/concurrent_listing_test.go
@@ -94,10 +94,10 @@ func (s *concurrentListingTest) Test_OpenDirAndLookUp(t *testing.T) {
 		defer wg.Done()
 		for i := 0; i < iterationsForLightOperations; i++ {
 			f, err := os.Open(targetDir)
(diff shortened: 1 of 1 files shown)


Tulsishah (commented):
Should we do it for other tests as well?

gargnitingoogle:
> Should we do it for other tests as well?

Done. I was hoping to keep the scope of this change limited to only the test in question, but I have not expanded it to this whole file.

Tulsishah (approved):

PR #3271, opened 2025-05-07
Flag to support inactive timeout reader

raj-prince:
### Description
- Adding a new hidden flag `--read-inactive-stream-timeout`
- Default value - 0s (that means disabled)
- Currently only applicable for `--client-protocol=grpc`

412465859

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -654,8 +654,8 @@
   usage: "Mount only a specific directory within the bucket. See docs/mounting for more information"
   default: ""
 
-- config-path: "read.inactivity-timeout"
-  flag-name: "read-inactivity-timeout"
+- config-path: "read.inactive-stream-timeout"
+  flag-name: "read-inactive-stream-timeout"
   type: "duration"
   usage: >-
     Duration of inactivity after which an open GCS read stream is automatically closed.
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -300,7 +300,6 @@
     https://storage.googleapis.com/storage/v1.
   default: ""
 
-
 - config-path: "gcs-connection.experimental-enable-json-read"
   flag-name: "experimental-enable-json-read"
   type: "bool"
@@ -655,6 +654,17 @@
   usage: "Mount only a specific directory within the bucket. See docs/mounting for more information"
   default: ""
 
+- config-path: "read.inactivity-timeout"
+  flag-name: "read-inactivity-timeout"
+  type: "duration"
+  usage: >-
+    Duration of inactivity after which an open GCS read stream is automatically closed.
+    This helps conserve resources when a file handle remains open without active Read calls.
+    A value of '0s' disables this timeout.
+    Note: Currently only applies when using the 'grpc' client-protocol.
+  default: "0s"
+  hide-flag: true
+
 - config-path: "write.block-size-mb"
   flag-name: "write-block-size-mb"
   type: "int"
@@ -667,8 +677,7 @@
 - config-path: "write.create-empty-file"
   flag-name: "create-empty-file"
   type: "bool"
-  usage: "For a new file, it creates an empty file in Cloud Storage bucket as a
-  hold."
+  usage: "For a new file, it creates an empty file in Cloud Storage bucket as a hold."
   default: false
 
 - config-path: "write.enable-streaming-writes"
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -124,7 +124,7 @@ func resolveParallelDownloadsValue(v isSet, fc *FileCacheConfig, c *Config) {
 func resolveReadconfig(c *Config, r *ReadConfig) {
 	// Only enable for GRPC client protocol.
 	if c.GcsConnection.ClientProtocol != GRPC {
-		r.InactivityTimeout = 0
+		r.InactiveStreamTimeout = 0
 	}
 }
 
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -121,7 +121,7 @@ func resolveParallelDownloadsValue(v isSet, fc *FileCacheConfig, c *Config) {
 	}
 }
 
-func resolveReadconfig(c *Config, r *ReadConfig) {
+func resolveReadConfig(c *Config, r *ReadConfig) {
 	// Only enable for GRPC client protocol.
 	if c.GcsConnection.ClientProtocol != GRPC {
 		r.InactiveStreamTimeout = 0
@@ -148,7 +148,7 @@ func Rationalize(v isSet, c *Config, optimizedFlags []string) error {
 	resolveStatCacheMaxSizeMB(v, &c.MetadataCache, optimizedFlags)
 	resolveCloudMetricsUploadIntervalSecs(&c.Metrics)
 	resolveParallelDownloadsValue(v, &c.FileCache, c)
-	resolveReadconfig(c, &c.Read)
+	resolveReadConfig(c, &c.Read)
 
 	return nil
 }
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -121,6 +121,13 @@ func resolveParallelDownloadsValue(v isSet, fc *FileCacheConfig, c *Config) {
 	}
 }
 
+func resolveReadconfig(c *Config, r *ReadConfig) {
+	// Only enable for GRPC client protocol.
+	if c.GcsConnection.ClientProtocol != GRPC {
+		r.InactivityTimeout = 0
+	}
+}
+
 // Rationalize updates the config fields based on the values of other fields.
 func Rationalize(v isSet, c *Config, optimizedFlags []string) error {
 	var err error
@@ -141,6 +148,7 @@ func Rationalize(v isSet, c *Config, optimizedFlags []string) error {
 	resolveStatCacheMaxSizeMB(v, &c.MetadataCache, optimizedFlags)
 	resolveCloudMetricsUploadIntervalSecs(&c.Metrics)
 	resolveParallelDownloadsValue(v, &c.FileCache, c)
+	resolveReadconfig(c, &c.Read)
 
 	return nil
 }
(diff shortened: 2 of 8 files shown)


cfg/params.yaml
```diff
@@ -655,6 +654,17 @@
   usage: "Mount only a specific directory within the bucket. See docs/mounting for more information"
   default: ""
 
+- config-path: "read.inactive-stream-timeout"
+  flag-name: "read-inactive-stream-timeout"
+  type: "duration"
```
kislaykishore, line 659: We are moving away from having duration based flags. While I know it's a bummer, our most important flags are int flags with names that end with "-secs". If we ever plan on exposing this flag to public (unhiding it), then we should probably follow the "-secs" convention.

cfg/rationalize.go
```diff
@@ -121,6 +121,13 @@ func resolveParallelDownloadsValue(v isSet, fc *FileCacheConfig, c *Config) {
 	}
 }
 
+func resolveReadconfig(c *Config, r *ReadConfig) {
+	// Only enable for GRPC client protocol.
+	if c.GcsConnection.ClientProtocol != GRPC {
+		r.InactiveStreamTimeout = 0
```
kislaykishore, line 127: Wondering if we should use -1 to represent infinite inactive-stream-timeout.

cfg/rationalize.go
```diff
@@ -121,6 +121,13 @@ func resolveParallelDownloadsValue(v isSet, fc *FileCacheConfig, c *Config) {
 	}
 }
 
+func resolveReadconfig(c *Config, r *ReadConfig) {
```
kislaykishore, line 124:
```suggestion
func resolveReadConfig(c *Config, r *ReadConfig) {
```

cfg/rationalize.go
```diff
@@ -141,6 +148,7 @@ func Rationalize(v isSet, c *Config, optimizedFlags []string) error {
 	resolveStatCacheMaxSizeMB(v, &c.MetadataCache, optimizedFlags)
 	resolveCloudMetricsUploadIntervalSecs(&c.Metrics)
 	resolveParallelDownloadsValue(v, &c.FileCache, c)
+	resolveReadconfig(c, &c.Read)
```
kislaykishore, line 151:
```suggestion
	resolveReadConfig(c, &c.Read)
```

cfg/rationalize_test.go
```diff
@@ -579,3 +579,48 @@ func TestRationalize_ParallelDownloadsConfig(t *testing.T) {
...
+					InactiveStreamTimeout: 60 * time.Second,
+				},
+				GcsConnection: GcsConnectionConfig{
+					ClientProtocol: HTTP1,
+				},
+			},
+			expectedTimeout: 0,
+		},
+		{
+			name: "gRPC_client_protocol",
```
kislaykishore, line 603: super nit: I think the idea of naming subtests is so that you can easily execute it on a command line without putting a lot of cognitive overhead. That's why we recommend practices like don't have spaces in the test since then the user will have to replace spaces with underscore to run them via command line. In the same vein, I think it'd be useful to keep lowercase casing throughout so that user doesn't need to remember the casing nitty-gritty. Although I don't have a standard readability recommendation to back this up.
```suggestion
			name: "grpc_client_protocol",
```

cfg/rationalize_test.go
```diff
@@ -579,3 +579,48 @@ func TestRationalize_ParallelDownloadsConfig(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			if assert.NoError(t, Rationalize(&mockIsSet{}, tc.config, []string{})) {
+				assert.EqualValues(t, tc.expectedTimeout, tc.config.Read.InactiveStreamTimeout)
+			}
```
kislaykishore, line 623:
```suggestion
			require.NoError(t, Rationalize(&mockIsSet{}, tc.config, []string{}))
			
			assert.EqualValues(t, tc.expectedTimeout, tc.config.Read.InactiveStreamTimeout)
```

cmd/root_test.go
```diff
@@ -1282,3 +1282,39 @@ func TestArgParsing_GCSRetries(t *testing.T) {
...
+				return nil
+			})
+			require.Nil(t, err)
+			cmd.SetArgs(convertToPosixArgs([]string{"gcsfuse", fmt.Sprintf("--config-file=testdata/read_config/%s", tc.cfgFile), "abc", "pqr"}, cmd))
+
+			err = cmd.Execute()
+
+			if assert.NoError(t, err) {
+				assert.Equal(t, tc.expectedTimeout, gotConfig.Read.InactiveStreamTimeout)
+			}
```
kislaykishore, line 1317:
```suggestion
			err = cmd.Execute()
			
			require.NoError(t, err)
			assert.Equal(t, tc.expectedTimeout, gotConfig.Read.InactiveStreamTimeout)
```

cfg/rationalize.go
gargnitingoogle: I don't think we have a concept of a `-1` supported with duration-type flags. Anyway, as the the flag description defines it, the value 0 disables the flag i.e. no timeout on stream, which is the only behaviour today. So, no -1 needed in that case. In short, no change needed here.

cfg/params.yaml
gargnitingoogle: In fact, because this value never needs to be negative (`-1` not needed because `0` is equivalent to no timeout), we should make it an unsigned integer .

cfg/rationalize.go
```diff
@@ -121,6 +121,13 @@ func resolveParallelDownloadsValue(v isSet, fc *FileCacheConfig, c *Config) {
 	}
 }
 
+func resolveReadconfig(c *Config, r *ReadConfig) {
+	// Only enable for GRPC client protocol.
+	if c.GcsConnection.ClientProtocol != GRPC {
```
gargnitingoogle, line 126: Shouldn't we put a warning log here to indicate that this flag has been ignored to make it easier for the user to spot why this flag got ignored ?

cfg/rationalize.go
gargnitingoogle, line 127: This is a bit problematic for zonal/rapid buckets, where despite user passing `client-protocol=HTTP`, the storage-client uses GRPC. We change it under the hood. So, if user passes an inactiveStreamTimeout with a zonal bucket, client-protocol `HTTP` or doesn't pass `client-protocol` in which it becomes `HTTP` by default, the timeout won't apply. In fact, for zonal/rapid buckets, the user will almost never pass client-protocol GRPC as it's assumed and the timeout will never work for them.

cfg/rationalize_test.go
```diff
@@ -579,3 +579,48 @@ func TestRationalize_ParallelDownloadsConfig(t *testing.T) {
...
+			name: "gRPC_client_protocol",
+			config: &Config{
+				Read: ReadConfig{
+					InactiveStreamTimeout: 60 * time.Second,
+				},
+				GcsConnection: GcsConnectionConfig{
+					ClientProtocol: GRPC,
+				},
+			},
+			expectedTimeout: 60 * time.Second,
```
gargnitingoogle, line 612: We should add a default test case for when ReadConfig and ClientProtocol aren't passed as that is the most common scenario.

cmd/root_test.go
```diff
@@ -1282,3 +1282,39 @@ func TestArgParsing_GCSRetries(t *testing.T) {
...
+	}{
+		{
+			name:            "default",
+			cfgFile:         "empty.yaml",
+			expectedTimeout: 0,
+		},
+		{
+			name:            "override_default",
+			cfgFile:         "override.yaml",
+			expectedTimeout: 0,
```
gargnitingoogle, line 1300: Is this 0 because of the default client-protocol being `http` ? If so, we should add an override test case with client-protocol being `grpc`

cfg/rationalize_test.go
raj-prince: Skipping this, as have already added the test in the root_test.go. Trying to minimized the duplicate test.

cfg/rationalize.go
raj-prince: I wouldn't want to add a log here.
(a) It's a hidden flag, doesn't want user to know.
(b) Anyway we are printing the flags value after rationalization (if not we should do it).

cfg/rationalize.go
raj-prince: In the timeout world, generally 0 is used to represent no timeout. E.g., we use for the existing http-client-timeout flag: https://github.com/GoogleCloudPlatform/gcsfuse/blob/1b7f821e915472d09718b2825a59a3e638864414/cfg/params.yaml#L322

kislaykishore (approved):

cfg/rationalize.go
raj-prince: Thanks, in that case we will handle internally for zonal bucket.

That means timeout will be activated when client-protocol = `grpc` or `bucket-type` is `rapid`.

kislaykishore (approved):

cfg/rationalize.go
gargnitingoogle: > It's a hidden flag, doesn't want user to know.

The warning log should happen only if the value of this hidden flag is non-default, in which case it is useful for user to know why it got ignored.

> Anyway we are printing the flags value after rationalization (if not we should do it).

Agree. My point was not that the user knew what value they set, but that they knew why that value was ignored.

Anyway, this is a nice-to-have for the user (for the time being gcsfuse devs), and not a blocker for me. It's up to you.

cfg/rationalize.go
gargnitingoogle: Could we just add a note for now to say that this may not work for zonal buckets.

gargnitingoogle (approved):
LGTM. Couple of non-blocking comments open.

cfg/rationalize.go
gargnitingoogle: Or capture it in buganizer for future as I think this is easy to miss in the future.

kislaykishore (approved):

cfg/rationalize.go
raj-prince: Have captured in the design doc. Thanks!

PR #3272, opened 2025-05-07
Remove infinite TTL from MD cache

codechanges:
### Description
This change is made to ascertain that there is no negative side effect due to infinite TTL for MD cache.

b/402273935

### Testing details
1. Manual - NA
2. Unit tests - Yes
3. Integration tests - Yes


charith87:
Obsolete. Closing.

PR #3284, opened 2025-05-11
Integrate inactiveTimeoutReader in random-reader

raj-prince:
### Description
Includes:
- Changes to integrate the inactiveTimeoutReader in random reader, activated for positive inactive-stream-timeout.
- Unit test to validate no behavior change of random-reader and integration with inactiveTimeoutReader.

---
Currently, timeout is only enabled when user passes client-protocol as gRPC. In case of zonal bucket, when user doesn't pass the grpc as client protocol - this feature will be disabled. So, requires an internal discussion to support for zonal bucket,  we might need to create another kill-switch flag instead of using the same timeout.
> Discussed offline, we would need to enable it for http also, then we would not need any consideration about the zonal bucket. I'll create a separate PR to enable this flag for http client protocol.

b/416129654

### Testing details
1. Manual - NA
4. Unit tests - NA
5. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1839,7 +1839,7 @@ func (fs *fileSystem) CreateFile(
 	fs.nextHandleID++
 
 	// Creating new file is always a write operation, hence passing readOnly as false.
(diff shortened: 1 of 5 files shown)


raj-prince:
image

vadlakondaswetha (approved):

kislaykishore (approved):

PR #3285, opened 2025-05-11
E2E test for inactiveStreamTimeout Support

raj-prince:
### Description
1. WithTimeout test to validate - reader is closed and further reader works as expected.
2. WithoutTimeout test to validate - no such read closer after the timeout.

b/416943998

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/inactive_stream_timeout/setup_test.go
+++ b/tools/integration_tests/inactive_stream_timeout/setup_test.go
@@ -42,7 +42,7 @@ const (
 	kChunkSizeToRead                     = 128 * 1024       // 64 KiB
 	kTestFileName                        = "foo"
 	kDefaultInactiveReadTimeoutInSeconds = 1 // A short timeout for testing
-	kLogFileNameForMountedDirectoryTests = "/tmp/gcsfuse_inactive_read_timeout_test_logs/log.json"
+	kLogFileNameForMountedDirectoryTests = "/tmp/inactive_stream_timeout_logs/log.json"
 	kHTTP1ClientProtocol                 = "http1"
 	kGRPCClientProtocol                  = "grpc"
 )
@@ -146,8 +146,7 @@ func createConfigFile(flags *gcsfuseTestFlags) string {
 			"format":    "json", // Ensure JSON logs for easier parsing
 		},
 	}
-	filePath := setup.YAMLConfigFile(mountConfig, flags.fileName)
-	return filePath
+	return setup.YAMLConfigFile(mountConfig, flags.fileName)
 }
 
 func TestMain(m *testing.M) {
--- a/tools/integration_tests/inactive_stream_timeout/setup_test.go
+++ b/tools/integration_tests/inactive_stream_timeout/setup_test.go
@@ -1,0 +1,203 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


tools/integration_tests/inactive_stream_timeout/setup_test.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}()
+
+	setup.ExitWithFailureIfBothTestBucketAndMountedDirectoryFlagsAreNotSet()
+
+	if setup.MountedDirectory() != "" {
+		gMountDir = setup.MountedDirectory()
+		setup.SetLogFile(kLogFileNameForMountedDirectoryTests)
+		// Run tests for mounted directory if the flag is set.
+		os.Exit(m.Run())
```
Tulsishah, line 172: Why are we not using this [function](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L355)?

Also where are we using  gMountDir?

Tulsishah (commented):
Can we add step to run tests [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/run_tests_mounted_directory.sh)?

tools/integration_tests/inactive_stream_timeout/setup_test.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+const (
+	kTestDirName                         = "inactiveReadTimeoutTest"
+	kOnlyDirMounted                      = "onlyDirInactiveReadTimeout"
+	kFileSize                            = 10 * 1024 * 1024 // 10 MiB
+	kChunkSizeToRead                     = 128 * 1024       // 64 KiB
+	kTestFileName                        = "foo"
+	kDefaultInactiveReadTimeoutInSeconds = 1 // A short timeout for testing
+	kLogFileNameForMountedDirectoryTests = "/tmp/gcsfuse_inactive_read_timeout_test_logs/log.json"
```
Tulsishah, line 45: Will we able to send failure logs on kokoro bucket?
ref - https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L236

tools/integration_tests/inactive_stream_timeout/with_timeout_test.go
```diff
@@ -0,0 +1,141 @@
...
+	// 3. Wait for timeout
+	time.Sleep(2*timeoutDuration + 1*time.Second) // Add buffer
+	endTimeWait := time.Now()
+
+	// 4. "Closing reader" log should be present.
+	validateInactiveReaderClosedLog(t, setup.LogFile(), gcsFileName, true, endTimeRead, endTimeWait)
+
+	// 5. Further reads should work as it is.
+	_, err = fileHandle.ReadAt(buff, 8)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
raj-prince: Yes, updated the comment.

tools/integration_tests/inactive_stream_timeout/setup_test.go
raj-prince: Can't use directly, as we are setting the log-file for mountedDir test.

gMountDir is used in every TestSetup `mountGCSFuseAndSetupTestDir` in setup_test.go file.

tools/integration_tests/inactive_stream_timeout/setup_test.go
raj-prince: I see, logging is disabled for mountedDir test. Otherwise, it should work as it is.

Ref: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/setup/setup.go#L261

Tulsishah (approved):
Please share these tests to GKE so that they can add it in their pipleline.

PR #3286, opened 2025-05-12
[testing-on-gke] Fix bug in zb output download

gargnitingoogle:
### Description
This fixes the bug in downloading the fio outputs generated by gke ai/ml test tool for runs on ZB buckets (i.e. with `zonal=true`). This failure is specifically in a case where a single input bucket is used for runs of different file-size values in different fio workloads. This fix downloads only the fio outputs relevant to specific file-sizes while parsing outputs for that file-size.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/testing_on_gke/examples/run-gke-tests.sh
+++ b/perfmetrics/scripts/testing_on_gke/examples/run-gke-tests.sh
@@ -755,8 +755,8 @@ function downloadFioOutputsFromBucket() {
   dst_dir="${gcsfuse_src_dir}/perfmetrics/scripts/testing_on_gke/bin/fio-logs/${instance_id}/${fileSize}"
   if test -d "${src_dir}" ; then
     mkdir -pv "${dst_dir}"
-    echo "Copying files from \"${src_dir}\" to \"${dst_dir}/\" ... "
-    cp -rfvu "${src_dir}"/* "${dst_dir}"/
+    echo "Copying files of type ${fileSize}* from \"${src_dir}\" to \"${dst_dir}/\" ... "
+    cp -rfvu "${src_dir}"/${fileSize}* "${dst_dir}"/
   fi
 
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/testing_on_gke/examples/run-gke-tests.sh
```diff
@@ -755,8 +755,8 @@ function downloadFioOutputsFromBucket() {
   dst_dir="${gcsfuse_src_dir}/perfmetrics/scripts/testing_on_gke/bin/fio-logs/${instance_id}/${fileSize}"
   if test -d "${src_dir}" ; then
     mkdir -pv "${dst_dir}"
-    echo "Copying files from \"${src_dir}\" to \"${dst_dir}/\" ... "
-    cp -rfvu "${src_dir}"/* "${dst_dir}"/
+    echo "Copying files of type ${fileSize}* from \"${src_dir}\" to \"${dst_dir}/\" ... "
```
Tulsishah, line 758: Can't we add the same in line no. 754? This is stay consistent with dest dir.

gargnitingoogle: Technically we can, but it'll be a little misleading and confusing, as the variable src_dir=`"${mountpath}/fio-output/${instance_id}"` is a proper directory in which I get subdirectories such as 64K-xyz, 128K-xyz etc. and I need to copy only sub-dir `64K-xyz` to `${dst_dir}` ending with `64K`. Hence I've kept it in the cp command, rather than in the `src_dir` variable to keep it more accurate.

gargnitingoogle: @Tulsishah ^

Tulsishah (approved):

PR #3290, opened 2025-05-13
Adding takeover API implementation to fake bucket

anushka567:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/storage/fake/fake_object_writer.go
```diff
@@ -97,3 +97,7 @@ func NewFakeObjectWriter(b *bucket, req *gcs.CreateObjectRequest) (w gcs.Writer,
 
 	return wr, nil
 }
+
+func NewFakeAppendableObjectWriter(b *bucket, req *gcs.CreateObjectChunkWriterRequest) (w gcs.Writer, err error) {
```
vadlakondaswetha, line 101: Why we need a seperate method, why can't we use directly NewFakeObjectWriter if we are not doing anything specific

internal/gcsx/prefix_bucket_test.go
```diff
@@ -364,6 +364,67 @@ func (t *PrefixBucketTest) TestCreateObjectChunkWriterAndFlushPendingWrites() {
...
+	w, err := t.bucket.CreateAppendableObjectWriter(
+		t.ctx,
+		&gcs.CreateObjectChunkWriterRequest{
+			CreateObjectRequest: gcs.CreateObjectRequest{
+				Name: suffix,
+			},
+			ChunkSize: 1024,
+			Offset:    10,
+		})
+
```
vadlakondaswetha, line 382: Nit: remove hte new line here, this is part of test invocation only.

Same for below method

internal/storage/fake/fake_object_writer.go
anushka567: Agreed, we can directly use the NewFakeObjectWriter, update the code now.

vadlakondaswetha (approved):

PR #3305, opened 2025-05-14
[testing-on-gke] Don't print metadata value if no failure

gargnitingoogle:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


perfmetrics/scripts/testing_on_gke/examples/fio/parse_logs.py
```diff
@@ -229,7 +230,8 @@ def create_output_scenarios_from_downloaded_files(args: dict) -> dict:
       # Create a record for this key.
       r = record.copy()
       try:
-        r["pod_name"] = pod_name
+        for metadata, metadata_value in metadata_values.items():
```
Tulsishah, line 233: Previously you were assigning some values later, now you are doing before, will it be fine?

perfmetrics/scripts/testing_on_gke/examples/fio/parse_logs.py
```diff
@@ -262,7 +261,10 @@ def create_output_scenarios_from_downloaded_files(args: dict) -> dict:
         r["e2e_latency_ns_p99"] = clat_ns_percentile["99.000000"]
         r["e2e_latency_ns_p99.9"] = clat_ns_percentile["99.900000"]
       except Exception as e:
-        print(f"Failed to create following record with error: {e}")
+        print(
+            f"Failed to create following record with error: {e}, metadata:"
```
Tulsishah, line 265: What if an exception is not raised during the initial assignment of a metadata value, but then an exception occurs during a subsequent assignment? What value would be printed for metadata_values in that scenario?

perfmetrics/scripts/testing_on_gke/examples/fio/parse_logs.py
```diff
@@ -355,8 +357,6 @@ def write_records_to_bq_table(
     bq_dataset_id: str,
     bq_table_id: str,
 ):
-  fioBqExporter = FioBigqueryExporter(bq_project_id, bq_dataset_id, bq_table_id)
-
   # list of FioRowTable objects to be populated to be inserted into BigQuery
```
Tulsishah, line 360: Why did we move this below?

Tulsishah (commented):
We don't require any changes in unit  tests?

perfmetrics/scripts/testing_on_gke/examples/fio/parse_logs.py
gargnitingoogle: Previously you were assigning some values later, now you are doing before, will it be fine? Yes, not problem, because their values are being generated much before the for-loop over epoch json files. So, the order of these assignments won't cause an issue.

perfmetrics/scripts/testing_on_gke/examples/fio/parse_logs.py
gargnitingoogle: > What if an exception is not raised during the initial assignment of a metadata value, but then an exception occurs during a subsequent assignment? What value would be printed for metadata_values in that scenario?

In that case, the default values will be printed for the remaining metadata fields will be printed, which are all `""` (initialized at https://github.com/GoogleCloudPlatform/gcsfuse/blob/2d3f6e33224024fad12fbae834721d4fe23d7b38/perfmetrics/scripts/testing_on_gke/examples/fio/parse_logs.py#L36 ).

perfmetrics/scripts/testing_on_gke/examples/fio/parse_logs.py
gargnitingoogle: This is an expensive operation. It creates a Bigquery client, and possibly also a bigquery dataset and a bigquery table, and has a 2 min sleep in the code after that. All this unnecessary if for some reason, we don't get any rows to export to bigquery.

gargnitingoogle:
> We don't require any changes in unit tests?

We don't have any unit tests for this file as of today, as this is a test tool itself, and the effort for adding unit tests for the parser higher than the value it carries.

Tulsishah (approved):

PR #3309, opened 2025-05-15
set new defaults for write configs

ashmeenkaur:
### Description
This PR sets new defaults for write config:
1. `--write-global-max-blocks` config flag. 
- The default value is set to 4 blocks.
-  For high end machines, the default will be 1600 blocks.
2. `--enable-streaming-writes` config flag is set to true.

This PR also includes the following fixes:
1. Proper propagation of chunk transfer timeout value to go SDK.
2. Fix in prefix bucket where object name was not properly set in returned min object.

Test changes included to 
b/418687148

### Testing details
1. Manual - NA
2. Unit tests - updated
3. Integration tests - NA


meet2mky (approved):
Minor comments, overall looks good.

PR #3311, opened 2025-05-15
Script to find micro benchmarks for writing file with single thread 

Tulsishah:
### Description

1. Create function to create file of given size with single thread.
2. Calculate latency, bandwidth and log in big query table.
3. File size, Num of files can be passed through flags.

[b/415708057](https://b.corp.google.com/issues/415708057)

### Testing details
1. Manual - Manually tested script
4. Unit tests - Added
5. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/micro_benchmarks/write_single_thread.py
+++ b/perfmetrics/scripts/micro_benchmarks/write_single_thread.py
@@ -16,6 +16,7 @@ import os
 import argparse
 import time
 import helper
+import sys
 
 MOUNT_DIR = "gcs"
 FILE_PREFIX = "testfile"
@@ -28,11 +29,16 @@ def delete_existing_file(file_path):
       file_path (str): The full path to the file to delete.
 
   Returns:
-      None
+      bool: True if file deleted or didn't exist; False on error.
   """
-  if os.path.exists(file_path):
-    os.remove(file_path)
-    print(f"{file_path} existed and was cleared.")
+  try:
+    if os.path.exists(file_path):
+      os.remove(file_path)
+      print(f"{file_path} existed and was cleared.")
+    return True
+  except Exception as e:
+    print(f"Error deleting file {file_path}: {e}")
+    return False
 
 def write_random_file(file_path, file_size_in_bytes):
   """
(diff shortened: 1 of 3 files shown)


perfmetrics/scripts/micro_benchmarks/write_single_thread.py
```diff
@@ -0,0 +1,88 @@
...
+  """
+  total_bytes_written = 0
+  file_size_in_bytes = int(file_size_in_gb * (1000 ** 3))
+
+  for i in range(num_files):
+    file_path = os.path.join(MOUNT_DIR, f"{FILE_PREFIX}_{file_size_in_gb}_{i}.bin")
+
+    try:
+      if os.path.exists(file_path):
+        os.remove(file_path)
```
charith87, line 42: Should we remove the existing deleting in a separate function so total time doesn't include that.

perfmetrics/scripts/micro_benchmarks/write_single_thread.py
```diff
@@ -0,0 +1,135 @@
...
+  except RuntimeError as e:
+    print(f"Failed during file write: {e}")
+    helper.unmount_gcs_directory(MOUNT_DIR)
+    sys.exit(1)  # Exit with error status
+  duration = time.time() - start
+
+  helper.unmount_gcs_directory(MOUNT_DIR)
+
+  helper.log_to_bigquery(
+      duration_sec=duration,
```
charith87, line 128: dont we have a start time as well?

Tulsishah: We are just logging duration for reading/writing/listing.

Tulsishah: If we want to add start time, just to check viceroy or any other logs then I can take it as separate PR.

charith87: But without the start time how will we plot the dashboard.

charith87: Resolving it for now since current time is considered now but will be changed to start time in next PR.

charith87 (approved):
Approving. We can iterate as we find any issues on execution for new tests.

perfmetrics/scripts/micro_benchmarks/write_single_thread.py
Tulsishah: As discussed offline we have `timestamp` field which is having time at the time of logging, we can change it to time when we started read/write. I'll do it in next PR.

PR #3319, opened 2025-05-19
Add support for using single GCSFuse build for all test packages

codechanges:
### Description
Add support for using single GCSFuse build for all test packages
Currently, every package first builds its own GCSFuse binary and then uses it to run the test.
This not only adds extra latency but also increases space utilization.
This change will build GCSFuse as part of run e2e script and pass this information to the packages to they can use the same binary. This will increase performance and reduce space usage for integration tests.

Currently, this PR adds such a support for utils/setup.go only which covers majority of the test cases.

http://b/417330393

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - Yes

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -194,7 +194,7 @@ func SetUpTestDir() error {
 		// Set PATH to include the bin directory of the pre-built gcsfuse
 		err = os.Setenv(PathEnvVariable, filepath.Dir(binFile)+string(filepath.ListSeparator)+os.Getenv(PathEnvVariable))
 		if err != nil {
-			log.Printf("Error setting PATH for --gcsfuse_prebuilt_dir: %v", err)
+			return fmt.Errorf("error setting PATH for --gcsfuse_prebuilt_dir: %v", err.Error())
 		}
 	} else {
 		log.Printf("Building GCSFuse from source in the dir: %s ...", testDir)
@@ -210,7 +210,7 @@ func SetUpTestDir() error {
 		// Setting PATH so that executable is found in test directory.
 		err := os.Setenv(PathEnvVariable, path.Join(TestDir(), "bin")+string(filepath.ListSeparator)+os.Getenv(PathEnvVariable))
 		if err != nil {
-			log.Printf("Error in setting PATH environment variable: %v", err.Error())
+			return fmt.Errorf("error in setting PATH environment variable: %v", err.Error())
 		}
 	}
 


tools/integration_tests/run_e2e_tests.sh
```diff
@@ -155,6 +165,74 @@ TEST_DIR_NON_PARALLEL_FOR_ZB=(
 # Create a temporary file to store the log file name.
 TEST_LOGS_FILE=$(mktemp)
 
+# This variable will store the path if the script builds GCSFuse binaries (gcsfuse, mount.gcsfuse)
+BUILT_BY_SCRIPT_GCSFUSE_BUILD_DIR=""
+USE_PREBUILT_GCSFUSE_BINARY=""
+# This variable will store the path to the temporary directory holding the build_gcsfuse tool itself
+TEMP_BUILD_TOOL_DIR=""
+
+build_gcsfuse_once() {
```
meet2mky, line 174: Can we directly run this from project root directory?
```
go run "./tools/build_gcsfuse/main.go" . "${build_output_dir}" "e2e-$(date +%s)")
```
Instead of building `build_gcsfuse` tool first then building the gcsfuse from that?

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -52,6 +52,16 @@ if [[ $# -ge 6 ]] ; then
   fi
 fi
 
+# 7th parameter is to determine whether we want to disable build by the script
+# and let every test package build its own GCSFuse binary.
+BUILD_BINARY_IN_SCRIPT=true
```
Tulsishah, line 57: Why are we creating flag for this? Is it due to backward compatibility?

tools/integration_tests/run_e2e_tests.sh
codechanges: Thanks. I followed the exact mechanism mentioned in the buildGCsFuse function. This helps keeping the parity with that but let me see whether this can help too.

tools/integration_tests/util/setup/setup.go
```diff
@@ -168,7 +169,35 @@ func SetUpTestDir() error {
...
+		prebuiltDir := *gcsfusePreBuiltDir
+		log.Printf("Using GCSFuse from pre-built directory specified by --gcsfuse_prebuilt_dir flag: %s", prebuiltDir)
+		binFile = filepath.Join(prebuiltDir, "bin/gcsfuse")
+		sbinFile = filepath.Join(prebuiltDir, "sbin/mount.gcsfuse")
+
+		if _, statErr := os.Stat(binFile); statErr != nil {
+			return fmt.Errorf("gcsfuse binary from --gcsfuse_prebuilt_dir not found at %s: %w", binFile, statErr)
+		}
+		if _, statErr := os.Stat(sbinFile); statErr != nil {
+			log.Printf("Warning: Mount helper from --gcsfuse_prebuilt_dir not found at %s. Some tests might fail.", sbinFile)
```
Tulsishah, line 192: Shouldn't we throw this as error only?

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -438,7 +516,7 @@ function run_e2e_tests_for_tpc() {
   gcloud --verbosity=error storage rm -r gs://"$bucket"/*
 
   # Run Operations e2e tests in TPC to validate all the functionality.
-  GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/operations/... --testOnTPCEndPoint=$RUN_TEST_ON_TPC_ENDPOINT $GO_TEST_SHORT_FLAG $PRESUBMIT_RUN_FLAG --zonal=false -p 1 --integrationTest -v --testbucket="$bucket" --testInstalledPackage=$RUN_E2E_TESTS_ON_PACKAGE -timeout $INTEGRATION_TEST_TIMEOUT
+  GODEBUG=asyncpreemptoff=1 go test ./tools/integration_tests/operations/... --testOnTPCEndPoint=$RUN_TEST_ON_TPC_ENDPOINT $GO_TEST_SHORT_FLAG $PRESUBMIT_RUN_FLAG --zonal=false -p 1 --integrationTest -v --testbucket="$bucket" --testInstalledPackage=$RUN_E2E_TESTS_ON_PACKAGE $USE_PREBUILT_GCSFUSE_BINARY -timeout $INTEGRATION_TEST_TIMEOUT
```
Tulsishah, line 519: Should we pass it as `--gcsfuse_prebuilt_dir =USE_PREBUILT_GCSFUSE_BINARY`? to make consistency with rest of the flags?

tools/integration_tests/run_e2e_tests.sh
codechanges: Updated the code to use the command you mentioned. Thanks.

tools/integration_tests/run_e2e_tests.sh
codechanges: Yes, this will provide user an option to move to build binary per package as per the existing model.

tools/integration_tests/util/setup/setup.go
codechanges: I thought about it and not all test cases use this hence I though let make it a warning. Do you think it should be an error, I am open to suggestions.

Tulsishah (commented):
Should we create separate script for this?  this file is becoming large day by day and becoming hard to follow.

codechanges:
> Should we create separate script for this? this file is becoming large day by day and becoming hard to follow.

I agree but with everyone making changes to the script now. We should take a stock once all the changes are merged and how to refactor it in a better fashion.

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -155,6 +165,51 @@ TEST_DIR_NON_PARALLEL_FOR_ZB=(
 # Create a temporary file to store the log file name.
 TEST_LOGS_FILE=$(mktemp)
 
+# This variable will store the path if the script builds GCSFuse binaries (gcsfuse, mount.gcsfuse)
+BUILT_BY_SCRIPT_GCSFUSE_BUILD_DIR=""
+USE_PREBUILT_GCSFUSE_BINARY=""
```
Tulsishah, line 170: Can we add comment that it is GCSFuse flag?

tools/integration_tests/run_e2e_tests.sh
```diff
@@ -155,6 +165,51 @@ TEST_DIR_NON_PARALLEL_FOR_ZB=(
...
+
+build_gcsfuse_once() {
+  local build_output_dir # For the final gcsfuse binaries
+  build_output_dir=$(mktemp -d -t gcsfuse_e2e_run_build_XXXXXX)
+  echo "GCSFuse binaries will be built in ${build_output_dir}..."
+
+  local gcsfuse_src_dir
+  # Determine GCSFuse source directory
+  # If this script is in tools/integration_tests, project root is ../../
+  SCRIPT_DIR_REALPATH=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
```
Tulsishah, line 180: If I am running this script from gcsfuse directory then will it catch the path correctly?

tools/integration_tests/util/setup/setup.go
codechanges: Done

tools/integration_tests/run_e2e_tests.sh
codechanges: Yes, it will.

Tulsishah (approved):

PR #3324, opened 2025-05-19
cd script changes to run read cache tests on a seperate VM in louhi

anushka567:
…ouhi

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -18,16 +18,19 @@ set -x
 # Exit immediately if a command exits with a non-zero status.
 set -e
 
-# Extract the metadata parameters passed, for which we need the zone of the GCE VM 
+# Extract the metadata parameters passed, for which we need the zone of the GCE VM
 # on which the tests are supposed to run.
 ZONE=$(curl -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/zone)
 echo "Got ZONE=\"${ZONE}\" from metadata server."
 # The format for the above extracted zone is projects/{project-id}/zones/{zone}, thus, from this
 # need extracted zone name.
-ZONE_NAME=$(basename $ZONE)
+ZONE_NAME=$(basename "$ZONE")
 # This parameter is passed as the GCE VM metadata at the time of creation.(Logic is handled in louhi stage script)
 RUN_ON_ZB_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAME" --format='get(metadata.run-on-zb-only)')
+RUN_READ_CACHE_TESTS_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAME" --format='get(metadata.run-read-cache-only)')
 echo "RUN_ON_ZB_ONLY flag set to : \"${RUN_ON_ZB_ONLY}\""
+echo "RUN_READ_CACHE_TESTS_ONLY flag set to : \"${RUN_READ_CACHE_TESTS_ONLY}\""
+
 
 # Logging the tests being run on the active GCE VM
 if [[ "$RUN_ON_ZB_ONLY" == "true" ]]; then
@@ -35,6 +38,12 @@ if [[ "$RUN_ON_ZB_ONLY" == "true" ]]; then
 else
   echo "Running integration tests for non-zonal buckets only..."
 fi
+
+# Logging the tests being run on the active GCE VM
+if [[ "$RUN_READ_CACHE_TESTS_ONLY" == "true" ]]; then
+  echo "Running read cache test only..."
+fi
+
 
 #details.txt file contains the release version and commit hash of the current release.
 gcloud storage cp  gs://gcsfuse-release-packages/version-detail/details.txt .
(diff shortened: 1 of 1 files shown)


tools/cd_scripts/e2e_test.sh
```diff
@@ -240,21 +250,23 @@ TEST_LOGS_FILE=$(mktemp)
 INTEGRATION_TEST_TIMEOUT=240m
 
 function run_non_parallel_tests() {
-  local exit_code=0
-  local -n test_array=$1
-  local BUCKET_NAME=$2
-  local zonal=$3
+  local exit_code=0 # Initialize to 0 for success
+  local BUCKET_NAME=$1
+  local zonal=$2
```
Tulsishah, line 255: Can we add comment here and everywhere for readbility?

tools/cd_scripts/e2e_test.sh
```diff
@@ -263,111 +275,73 @@ function run_non_parallel_tests() {
...
- if [ $non_parallel_tests_exit_code != 0 ] || [ $parallel_tests_exit_code != 0 ]; then
-   return 1
- fi
-}
-
-function run_e2e_tests_for_hns_bucket(){
-  hns_bucket_name_non_parallel=$(sed -n 3p ~/details.txt)-hns
-  echo "HNS Bucket name to run tests sequentially: "$hns_bucket_name_non_parallel
+function run_e2e_tests() {
+  local testcase=$1
```
Tulsishah, line 307: Can we add if condition above that this expect 4 argument and the order?

tools/cd_scripts/e2e_test.sh
```diff
@@ -263,111 +275,73 @@ function run_non_parallel_tests() {
...
-  wait $non_parallel_tests_zonal_group_pid
-  non_parallel_tests_zonal_group_exit_code=$?
+  wait "$parallel_tests_pid"
+  local parallel_tests_exit_code=$?
+  wait "$non_parallel_tests_pid"
+  local non_parallel_tests_exit_code=$?
 
-  if [ $parallel_tests_zonal_group_exit_code != 0 ] || [ $non_parallel_tests_zonal_group_exit_code != 0 ]; then
-    return 1
+  if [ "$non_parallel_tests_exit_code" -ne 0 ]; then
```
Tulsishah, line 337: if `parallel_tests_exit_code -ne 0` ?

tools/cd_scripts/e2e_test.sh
```diff
@@ -392,76 +366,87 @@ function gather_test_logs() {
...
-  wait $e2e_tests_flat_bucket_pid
-  e2e_tests_flat_bucket_status=$?
+    run_e2e_tests "zonal" read_cache_test_dir_parallel read_cache_test_dir_non_parallel true
+    exit_status["zonal"]=$?
 
-  wait $e2e_tests_hns_bucket_pid
-  e2e_tests_hns_bucket_status=$?
+else
+    if [[ "$RUN_ON_ZB_ONLY" == "true" ]]; then
+        run_e2e_tests "zonal" TEST_DIR_PARALLEL_ZONAL TEST_DIR_NON_PARALLEL_ZONAL true
```
Tulsishah, line 430: won't it run in background ?

anushka567: The `if` block gets executed only if the RUN_ON_ZB_ONLY is set to true and READ_CACHE_TESTS is set to false, which means the VM will essentially run only zonal bucket e2e tests excluding read cache. Thus, running in background is not necessary since there are no other tests being run.


I will update the code with comments to be able to follow through.

anushka567:
Tested the changes in cloned louhi pipeline. Stage failures due to internal test failures. Pipeline execution is working as intended.

tools/cd_scripts/e2e_test.sh
```diff
@@ -239,135 +244,130 @@ TEST_LOGS_FILE=$(mktemp)
...
+  if [[ -z $3 ]]; then
+    return 1 # The name of the test array cannot be empty.
+  fi
+  local -n test_array=$3 # Create a nameref to this array.
   local pids=()
 
   for test_dir_p in "${test_array[@]}"
   do
     test_path_parallel="./tools/integration_tests/$test_dir_p"
-    # To make it clear whether tests are running on a flat or HNS bucket, We kept the log file naming
```
Tulsishah, line 289: Why did we remove this comment?

tools/cd_scripts/e2e_test.sh
```diff
@@ -239,135 +244,130 @@ TEST_LOGS_FILE=$(mktemp)
...
   done
   return $exit_code
 }
 
-function run_e2e_tests_for_flat_bucket() {
-  flat_bucket_name_non_parallel=$(sed -n 3p ~/details.txt)
-  echo "Flat Bucket name to run tests sequentially: "$flat_bucket_name_non_parallel
+#Common method to invoke e2e tests on different types of buckets: flat, HNS or Zonal
+# Arguments:
+#   $1: TESTCASE (flat/hns/zonal)
```
Tulsishah: nit: BucketType?

tools/cd_scripts/e2e_test.sh
```diff
@@ -18,37 +18,43 @@ set -x
...
 # on which the tests are supposed to run.
 ZONE=$(curl -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/zone)
 echo "Got ZONE=\"${ZONE}\" from metadata server."
 # The format for the above extracted zone is projects/{project-id}/zones/{zone}, thus, from this
 # need extracted zone name.
-ZONE_NAME=$(basename $ZONE)
+ZONE_NAME=$(basename "$ZONE")
 # This parameter is passed as the GCE VM metadata at the time of creation.(Logic is handled in louhi stage script)
 RUN_ON_ZB_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAME" --format='get(metadata.run-on-zb-only)')
+RUN_READ_CACHE_TESTS_ONLY=$(gcloud compute instances describe "$HOSTNAME" --zone="$ZONE_NAME" --format='get(metadata.run-read-cache-only)')
```
Tulsishah, line 39: `RUN_READ_CACHE_TESTS_ONLY` is true, the script hardcodes the test directories to be run, with an empty parallel list and "read_cache" as the only non-parallel test. This is logical for the current goal, but it reduces flexibility. If another test needed to be isolated in the future, the script's core logic would require modification rather than just changing the input arrays.

Can we change the const value to some generic for future extension.

Tulsishah (commented):
Overall it looks good to me. Some nits. Will approve after the comments get resolved.

tools/cd_scripts/e2e_test.sh
anushka567: Unintended. Will revert.

tools/cd_scripts/e2e_test.sh
anushka567: From what i understand, in case testcase X has to be run on a single VM that would mean, introducing another VM metadata parameter, introducing another list and then another if case for handling the test.

For that, we might have to tweak a lot of exisiting logic in louhi as well in terms of Vm reservations,etc. just in order to identify the VM which is supposed to run a particular test package.

While this recommendation is very much beneficial, I believe that a config based approach for the problem is much better than aiming generality at this point here . I have a proposal for that. Happy to discuss.

PR #3327, opened 2025-05-20
 [Dont review]Fixoffset

vadlakondaswetha:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/random_reader.go
+++ b/internal/gcsx/random_reader.go
@@ -176,6 +176,10 @@ type randomReader struct {
 	metricHandle common.MetricHandle
 
 	config *cfg.ReadConfig
+
+	// Specifies the next expected offset for the reads. Used to distinguish between
+	// sequential and random reads.
+	expectedOffset int64
 }
 
 func (rr *randomReader) CheckInvariants() {
@@ -336,12 +340,13 @@ func (rr *randomReader) ReadAt(
 	// is a 15-20x improvement in throughput: 150-200 MiB/s instead of 10 MiB/s.
 	if rr.reader != nil && rr.start < offset && offset-rr.start < maxReadSize {
 		bytesToSkip := offset - rr.start
-		discardedBytes, copyError := io.CopyN(io.Discard, rr.reader, int64(bytesToSkip))
+		discardedBytes, copyError := io.CopyN(io.Discard, rr.reader, bytesToSkip)
 		// io.EOF is expected if the reader is shorter than the requested offset to read.
 		if copyError != nil && !errors.Is(copyError, io.EOF) {
 			logger.Warnf("Error while skipping reader bytes: %v", copyError)
 		}
 		rr.start += discardedBytes
+		rr.expectedOffset = rr.start
 	}
 
 	// If we have an existing reader, but it's positioned at the wrong place,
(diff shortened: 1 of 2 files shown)


raj-prince (commented):
Overall, LGTM.

One minor behavior change I see, before and after. When the file is repeatedly read completely from the same file handle. Now, every start read will increase the seek (as expectedOffset will be equal to objectSize and fresh read will start from 0 offset), which was not the case earlier.

This might affect the performance due to extra seek, so good to think a bit here.

[Edited]
- More I can see, when completely read through a reader and then a random offset will lead to a increase in seek (behavior change).

raj-prince:
CC: @Tulsishah we need to incorporate this change as part of random reader refactoring.

internal/gcsx/random_reader.go
```diff
@@ -352,19 +356,19 @@ func (rr *randomReader) ReadAt(
...
 	}
 
 	if rr.reader != nil {
 		objectData.Size, err = rr.readFromRangeReader(ctx, p, offset, -1, rr.readType)
 		return
 	}
 
+	// If the data can't be served from the existing reader, then we need to update the seeks.
+	// If current offset is not same as expected offset, its a random read.
+	if rr.expectedOffset != 0 && rr.expectedOffset != offset {
```
Tulsishah, line 368: Why do we have condition `rr.expectedOffset != 0 ` ?  If offset went from 10, 50, 15, 0 then also we will count it as random right?

internal/gcsx/random_reader.go
```diff
@@ -666,3 +672,7 @@ func (rr *randomReader) closeReader() {
 		logger.Warnf("error while closing reader: %v", err)
 	}
 }
+
+func (rr *randomReader) updateExpectedOffset(offset int64) {
+	rr.expectedOffset = offset
```
abhishek10004, line 677: Should we set it to (offset % rr.object.Size) instead to ensure that it gets set back to zero when object has been read till the end?

vadlakondaswetha: Agreed its not random, my opinion on increasing seek vs not increasing:

- Increasing seek:
 in case of actual random reads, increasing seek makes sense. 
In case of sequential reads, increasing seek shouldn't effect anything since we are determing the size to read based on the total ReadBytes. so just updating seeks once for every file read shouldn't effect the calculation.
Also i am not sure what it is the scenario where the same file is read multiple times using the same file handle

- Not increasing seeks
We will miss the actual random scenario.

vadlakondaswetha:
> Overall, LGTM.
> 
> One minor behavior change I see, before and after. When the file is repeatedly read completely from the same file handle. Now, every start read will increase the seek (as expectedOffset will be equal to objectSize and fresh read will start from 0 offset), which was not the case earlier.
> 
> This might affect the performance due to extra seek, so good to think a bit here.
replied here: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3327#discussion_r2101693935
> 
> [Edited]
> 
> * More I can see, when completely read through a reader and then a random offset will lead to a increase in seek (behavior change).

Agreed its a behaviour change. IMO Seeks (sequential vs random) shouldn't be updated irrespective whether the reader is completely read or not. If you see the perf results for large files, we are actually seeing perf improvement in case of regional buckets too

PR #3328, opened 2025-05-20
routing writes to inode via fileHandle

anushka567:
### Description
The main changes include:
* Added Write() method to fileHandle, which will pass along the mode in which the filehandle has been opened when Write on inode is invoked.

**Testing:**
Existing unit tests have been updated and pass. This PR focuses on the structural changes to the write path. Further functional tests for append behavior will be covered as the feature is fully implemented.

[b/417137089](b/417137089)

### Testing details
1. Manual - Tested affected write/append flow manually
2. Unit tests - Included 


internal/fs/handle/file.go
```diff
@@ -165,6 +165,17 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	return
 }
 
+// To propagate the mode in which file has been opened while calling the Write()
```
vadlakondaswetha, line 168: the comment is not clear. what are you trying to explain.
Why we are adding Write() method in fileHandle.

Mention that all writes will be done at indoe level. Method is added in fileHandle to be able to pass the fileOpenMode which is used for determining the write path.
For eg: in case of append mode for ZB unfinalized objects, streaming writes flow is used.

internal/fs/fs.go
```diff
@@ -2651,7 +2662,7 @@ func (fs *fileSystem) WriteFile(
 	}
 
 	// Serve the request.
-	err = in.Write(ctx, op.Data, op.Offset)
+	err = in.Write(ctx, op.Data, op.Offset, util.Write)
```
vadlakondaswetha, line 2665: line no: 2659 in a new change and not incorporated in the new flow. We need to fix this. Can you create a task for this, so we dont forget.

internal/fs/fs.go
```diff
@@ -2637,6 +2638,16 @@ func (fs *fileSystem) WriteFile(
...
+
+	if fs.newConfig.Write.ExperimentalEnableRapidAppends {
+		fs.mu.Lock()
+		fh := fs.handles[op.Handle].(*handle.FileHandle)
+		fs.mu.Unlock()
+
+		if err := fh.Write(ctx, op.Data, op.Offset); err != nil {
+			return err
+		}
+	}
```
vadlakondaswetha, line 2650: either you need to put the logic from line no: 2651 in else block or add a return statement in the if block

internal/fs/handle/file.go
```diff
@@ -165,6 +165,17 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
...
+// method on file inode, required as the appends to unfinalized objects in zonal
+// buckets should trigger the streaming writes flow.
+// LOCKS_EXCLUDED(fh.inode)
+func (fh *FileHandle) Write(ctx context.Context, data []byte, offset int64) error {
```
vadlakondaswetha, line 172: unit tests?

anushka567: Added.

vadlakondaswetha (approved):

PR #3336, opened 2025-05-21
Enable reads when writes are in progress with streaming writes for Regional Buckets.

meet2mky:
### Description
This PR enables reads when writes are in-progress for streaming writes for Regional buckets. We will be finalizing the streaming writes upload for regional bucket as soon as we receive a read operation for the file and then the further write flow for the file will be performed using staged writes.

b/417358797

### Testing details
1. Manual - Yes
2. Unit tests - Updated
3. Integration tests - Yes

NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1152,13 +1152,13 @@ func (fs *fileSystem) promoteToGenerationBacked(f *inode.FileInode) {
 }
 
 // Flushes the supplied file inode to GCS, updating the index as
-// appropriate.
+// appropriate. If flushForRead is true it only flushes streaming writes files for Regional buckets.
 //
 // LOCKS_EXCLUDED(fs.mu)
 // LOCKS_REQUIRED(f)
 func (fs *fileSystem) flushFile(
 	ctx context.Context,
-	f *inode.FileInode) error {
+	f *inode.FileInode, flushForRead bool) error {
 	// FlushFile mirrors the behavior of native filesystems by not returning an error
 	// when file to be synced has been unlinked from the same mount.
 	if f.IsUnlinked() {
@@ -1166,7 +1166,16 @@ func (fs *fileSystem) flushFile(
 	}
 
 	// Flush the inode.
-	err := f.Flush(ctx)
+	var err error
+	// Do not flush for reads if bucket type is zonal.
+	if flushForRead && f.Bucket().BucketType().Zonal {
+		return nil
+	}
+	if flushForRead {
+		err = f.FlushUsingBufferedWriteHandler()
+	} else {
+		err = f.Flush(ctx)
+	}
 	if err != nil {
 		err = fmt.Errorf("FileInode.Sync: %w", err)
 		// If the inode was local file inode, treat it as unlinked.
@@ -2124,7 +2133,7 @@ func (fs *fileSystem) flushPendingWrites(ctx context.Context, fileInode *inode.F
 		return
 	}
 	// Try to flush if there are any pending writes.
-	err = fs.flushFile(ctx, fileInode)
+	err = fs.flushFile(ctx, fileInode, false)
 	minObject = fileInode.Source()
 	return
 }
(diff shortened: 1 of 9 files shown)


meet2mky:
Rename test failed in unfinalized package, retrying execution of e2e regional bucket.

tools/integration_tests/streaming_writes/common_streaming_writes_suite_test.go
```diff
@@ -50,17 +50,26 @@ func (t *StreamingWritesSuite) TearDownSuite() {
...
+	if setup.IsZonalBucketRun() && !t.fallbackToDiskCase {
+		operations.ValidateEOPNOTSUPPError(t.T(), err)
 	} else {
-		require.Error(t.T(), err)
-		assert.ErrorContains(t.T(), err, syscall.ENOTSUP.Error())
+		require.NoError(t.T(), err)
+		assert.Equal(t.T(), len(content), n)
+		// assert.Equal(t.T(), content, string(readContent))
 	}
 }
```
ashmeenkaur, line 75: aren't these 2 functions same? Can we create only 1?

tools/integration_tests/streaming_writes/read_file_test.go
```diff
@@ -41,12 +41,25 @@ func (t *StreamingWritesSuite) TestReadBeforeFileIsFlushed() {
...
 }
 
+func (t *StreamingWritesSuite) TestReadBeforeSyncThenWriteAgainAndRead() {
+	// Write data to file.
+	operations.WriteAt(t.data, 0, t.f1, t.T())
+
+	t.validateReadCall(t.f1, t.data)
+
+	operations.WriteAt(t.data, int64(len(t.data)), t.f1, t.T())
+
```
ashmeenkaur, line 57: Nit: 
```suggestion
```

tools/integration_tests/interrupt/interrupt_test.go
```diff
@@ -65,7 +65,10 @@ func TestMain(m *testing.M) {
 		{"--implicit-dirs=true"},
 		{"--config-file=" + setup.YAMLConfigFile(yamlContent1, "ignore_interrupts.yaml")},
 		{"--config-file=" + setup.YAMLConfigFile(yamlContent2, "default_ignore_interrupts.yaml")}}
-
+	// TODO(b/410698332): Enable this test for Zonal Bucket also once read start working.
```
ashmeenkaur, line 68:
```suggestion
	// TODO(b/417136852): Enable this test for Zonal Bucket also once read start working.
```

internal/fs/fs.go
```diff
@@ -2595,8 +2595,16 @@ func (fs *fileSystem) ReadFile(
 	fs.mu.Unlock()
 
 	fh.Lock()
+	fh.Inode().Lock()
 	defer fh.Unlock()
-
+	// Flush Pending streaming writes file for regional bucket and issue read within same inode lock.
+	if fh.Inode().IsUsingBWH() && !fh.Inode().Bucket().BucketType().Zonal {
```
ashmeenkaur, line 2601: As discussed, I think we should keep Sync/FlushPendingBufferedWrites separate (potentially in a separate method). That way we don't need to expose IsUsingBWH and will also allow us to use a separate utility when both bwh and temp file needs to be flushed including temporary file (for eg - before renames).

ashmeenkaur (commented):
can we enable local_file test package tests for streaming writes which are now working

tools/integration_tests/streaming_writes/common_streaming_writes_suite_test.go
meet2mky: Nope, In the first one the `os.ReadFile()` does a read and then closes the file handle. However in the second one it is using the provided fileHandle.
Test `TestReadBeforeSyncThenWriteAgainAndRead()` fails if the file handle is closed when read is performed. So created another method. Validating read through symlink needs to be using filePath/symlink only.

internal/fs/fs.go
ashmeenkaur, line 2601: NIt: add TODO to remove zonal bucket condition when we start leaving the objects in unfinalized state for zb

meet2mky: Done!

tools/integration_tests/interrupt/interrupt_test.go
meet2mky: Updated.

ashmeenkaur (approved):

PR #3340, opened 2025-05-21
Add MountID in UserAgent

charith87:
Add a 6 digit MountID in UserAgent to differentiate between mounts

b/417402055

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -20,8 +20,10 @@
 package cmd
 
 import (
+	"crypto/rand"
 	"fmt"
 	"io/fs"
+	"math/big"
 	"os"
 	"os/signal"
 	"path"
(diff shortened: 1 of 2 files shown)


cmd/legacy_main.go
```diff
@@ -87,13 +89,24 @@ func registerTerminatingSignalHandler(mountPoint string, c *cfg.Config) {
 
 func getUserAgent(appName string, config string) string {
 	gcsfuseMetadataImageType := os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE")
+	mountId := getUniqueMountID()
```
raj-prince, line 92: I feel, we should make caller's responsibility to pass the unique ID. Other wise, different invocation with the same parameter will return different user-agent string.

cmd/legacy_main.go
```diff
@@ -87,13 +89,24 @@ func registerTerminatingSignalHandler(mountPoint string, c *cfg.Config) {
 
 func getUserAgent(appName string, config string) string {
```
raj-prince, line 90:
```suggestion
func getUserAgent(appName string, config string, mountId string) string {
```

I feel, we should make caller's responsibility to pass the unique ID. Other wise, different invocation with the same parameter will return different user-agent string.

charith87:
We have added this feature already. Obsolete. Closing this.

PR #3355, opened 2025-05-26
Turn on the flag for Move object

Tulsishah:
### Description
Turn on the flag for Move object

[b/398707266](https://buganizer.corp.google.com/issues/398707266), [b/420534227](https://buganizer.corp.google.com/issues/420534227)

### Testing details
1. Manual - NA
2. Unit tests - NA
4. Checkpointing - Ran checkpointing tests for zonal and HNS to check the issue is gone.

### DIFF
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -82,7 +82,7 @@
   flag-name: "enable-atomic-rename-object"
   type: "bool"
   usage: "Enables support for atomic rename object operation on HNS bucket."
-  default: false
+  default: true
   hide-flag: true
 
 - config-path: "enable-hns"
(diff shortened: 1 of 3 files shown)


cfg/params.yaml
```diff
@@ -82,7 +82,7 @@
   flag-name: "enable-atomic-rename-object"
   type: "bool"
   usage: "Enables support for atomic rename object operation on HNS bucket."
-  default: false
+  default: true
```
raj-prince, line 85: Good to add unit test in [root_test.go](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/cmd/root_test.go) to validate the default behavior.

PR #3365, opened 2025-05-28
adding Finalized obj attribute to gcs.MinObject

anushka567:
### Description
This change introduces a `Finalized` attribute to the gcs.MinObject struct. This is essential for enabling GCSFuse to correctly handle read and append operations on unfinalized objects residing in Zonal Buckets (ZB).

The `Finalized` status is frequently checked within the read and write flows to determine the appropriate operational path  Adding this attribute directly to gcs.MinObject was chosen over placing it in extendedObjectAttributes for the following reasons:
* Performance: It avoids the overhead of an extra method invocation that would be required to fetch extended attributes during critical decision points in the code flow
* Efficiency: It prevents the retrieval of other, potentially unneeded, extended attributes, thereby being more resource-efficient

While the general principle is to keep gcs.MinObject lean to manage GCSFuse memory utilization, the frequent access and performance implications in this specific case make a direct attribute the more optimal solution

Due to the inclusion of this `Finalized` attribute in gcs.minObj , the average stat cache entry size has increased and hence , necessary changes have been made accordingly. 

[b/421166762](b/421166762)

### Testing details
1. Manual - Manually tested that the finalized attribute is getting populated
2. Unit tests - Existing unit tests updated.
3. Integration tests - Automated via PR label ( Tested both ZB and regional )


cfg/constants.go
```diff
@@ -67,7 +67,7 @@ const (
 	// meant for two purposes.
 	// 1. for conversion from stat-cache-capacity to stat-cache-max-size-mb.
 	// 2. internal testing.
-	AverageSizeOfPositiveStatCacheEntry uint64 = 1400
+	AverageSizeOfPositiveStatCacheEntry uint64 = 1464
```
anushka567: The addition of the Finalized attribute to the gcs.MinObject struct has increased the average size of a positive stat cache entry by 24 bytes, reflecting the size of this new time object field.

cfg/constants.go
```diff
@@ -67,7 +67,9 @@ const (
 	// meant for two purposes.
 	// 1. for conversion from stat-cache-capacity to stat-cache-max-size-mb.
 	// 2. internal testing.
-	AverageSizeOfPositiveStatCacheEntry uint64 = 1400
+	// Please note that with every added attribute of size x bytes, the heapsize
```
gargnitingoogle: This is wrong. It should be something like: For every x bytes increase on heap (structures), the RSS (Resident Set Size) which is closer to the actual memory usage increases by approximately 2x . E.g. if you add a boolean (8-bytes) to a stat-cache entry, increase this value by 16.

anushka567: Updated the comment. PTAL .

internal/storage/bucket_handle.go
```diff
@@ -367,7 +367,16 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 		IncludeFoldersAsPrefixes: req.IncludeFoldersAsPrefixes,
 		//MaxResults: , (Field not present in storage.Query of Go Storage Library but present in ListObjectsQuery in Jacobsa code.)
 	}
-	err = query.SetAttrSelection([]string{"Name", "Size", "Generation", "Metageneration", "Updated", "Metadata", "ContentEncoding", "CRC32C"})
+	minObjAttrs := []string{"Name", "Size", "Generation", "Metageneration", "Updated", "Metadata", "ContentEncoding", "CRC32C"}
+	if bh.BucketType().Zonal {
+		// If the bucket type is regional, then requesting this attribute via
+		// partial response API results in failure.
```
vadlakondaswetha: what is partial response api mean?
Also can you inline bug number in line 2 to reduce lengthy comments

i.e, partial response api results in failure (b/<bug_number>)

anushka567: Partial API only populates the requested attributes in the returned response, as we use in our listing flow.

vadlakondaswetha (approved):

PR #3366, opened 2025-05-28
feat: support rename of local file

ashmeenkaur:
### Description
Support Rename Operation for local files by flushing any pending writes before rename operation.
Also update e2e tests for the same.

b/402335988

### Testing details
1. Manual - NA
2. Unit tests - Updated
3. Integration tests - Updated

NA


internal/fs/local_file_test.go
```diff
@@ -93,7 +94,7 @@ func (t *LocalFileTest) createLocalFile(fileName string) (filePath string, f *os
 	t.T().Helper()
 	// Creating a file shouldn't create file on GCS.
 	filePath = path.Join(mntDir, fileName)
-	f, err := os.Create(filePath)
+	f, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC|syscall.O_DIRECT, 0655)
```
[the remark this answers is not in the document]
ashmeenkaur: Intended change for test predictability

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -496,7 +496,7 @@ func ClearCacheControlOnGcsObject(gcsObjPath string) error {
 
 func CreateFile(filePath string, filePerms os.FileMode, t testing.TB) (f *os.File) {
 	// Creating a file shouldn't create file on GCS.
-	f, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, filePerms)
+	f, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC|syscall.O_DIRECT, filePerms)
```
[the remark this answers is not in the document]
ashmeenkaur: Intended change for test predictability

tools/integration_tests/local_file/setup_test.go
```diff
@@ -67,10 +67,13 @@ func TestMain(m *testing.M) {
 	// Running these tests with streaming writes disabled because local file tests are already running in streaming_writes test package.
 	flagsSet := [][]string{
 		{"--implicit-dirs=true", "--rename-dir-limit=3", "--enable-streaming-writes=false"},
-		{"--implicit-dirs=false", "--rename-dir-limit=3", "--enable-streaming-writes=false"}}
+		{"--implicit-dirs=false", "--rename-dir-limit=3", "--enable-streaming-writes=false"},
```
Tulsishah, line 70: How does `--implicit-dirs` flag matter for this? also we made change to support rename for streaming writes then shouldn't we have test with `-enable-streaming-writes=true`?

tools/integration_tests/util/operations/validation_helper.go
```diff
@@ -48,13 +48,10 @@ func ValidateObjectNotFoundErr(ctx context.Context, t *testing.T, bucket gcs.Buc
...
 
 func ValidateESTALEError(t *testing.T, err error) {
-	require.Error(t, err)
-	assert.Regexp(t, syscall.ESTALE.Error(), err.Error())
-}
+	t.Helper()
 
-func ValidateEOPNOTSUPPError(t *testing.T, err error) {
 	require.Error(t, err)
-	assert.Regexp(t, syscall.EOPNOTSUPP.Error(), err.Error())
```
Tulsishah, line 57: why did we remove this?

tools/integration_tests/local_file/setup_test.go
ashmeenkaur: > How does --implicit-dirs flag matter for this

The tests would work regardless of implicit-dirs flag, just kept the same flag set.

> also we made change to support rename for streaming writes then shouldn't we have test with -enable-streaming-writes=true?

These tests were running with streaming writes only for non-zonal buckets because of some behavior difference between the 2 buckets. I've kept the same behavior by running the tests for streaming writes (default true) if it is a non zonal bucket run. Ref: line 74 below and
https://github.com/GoogleCloudPlatform/gcsfuse/blob/337c65a887d6782cfd642995b2c98872784be755/tools/integration_tests/streaming_writes/local_file_test.go#L75

tools/integration_tests/util/operations/validation_helper.go
ashmeenkaur: ValidateEOPNOTSUPPError function is removed as it was no longer required.

internal/fs/fs.go
```diff
@@ -2289,11 +2289,6 @@ func (fs *fileSystem) renameFile(ctx context.Context, op *fuseops.RenameOp, chil
 
 	switch c := child.(type) {
 	case *inode.FileInode:
-		// TODO(b/402335988): Fix rename flow for local files when streaming writes is disabled.
-		// If object to be renamed is a local file inode and streaming writes are disabled, rename operation is not supported.
-		if c.IsLocal() && !fs.newConfig.Write.EnableStreamingWrites {
-			return fmt.Errorf("cannot rename open file %q: %w", op.OldName, syscall.ENOTSUP)
-		}
```
meet2mky, line 2296: One think I wanted to understand here is how the fs map is updated when a file becomes gcsSynced due to rename. Do we need fs map update or not? I think it's not needed but wanted to confirm with you!

tools/integration_tests/local_file/setup_test.go
Tulsishah: Then I think we should remove implicit-dir dependency. This will decrease no. of flag set to verify the test.

tools/integration_tests/streaming_writes/local_file_test.go
```diff
@@ -51,30 +50,3 @@ func TestStreamingWritesLocalFileTestSuite(t *testing.T) {
...
-	setup.UnmountGCSFuse(rootDir)
-	setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
-}
-
-// Executes all tests that run with single streamingWrites configuration for localFiles.
-func TestExistingLocalFileTest(t *testing.T) {
-	if !setup.IsZonalBucketRun() {
-		s := new(existingLocalFileTestSuite)
-		s.CommonLocalFileTestSuite.TestifySuite = &s.Suite
-		suite.Run(t, s)
```
meet2mky, line 78: Local File Suite is now running with streaming writes by default so we don't it here right?

tools/integration_tests/local_file/setup_test.go
ashmeenkaur: I'll remove it in the refactoring PR to keep the changes segregated.

tools/integration_tests/local_file/read_dir_test.go
```diff
@@ -58,7 +58,7 @@ func readingDirNTimesShouldNotThrowError(n int, wg *sync.WaitGroup, t *testing.T
 // Tests
 // //////////////////////////////////////////////////////////////////////
 
-func (t *CommonLocalFileTestSuite) TestReadDir() {
+func (t *LocalFileTestSuite) TestReadDir() {
```
meet2mky, line 61: Can we make this unexported as it's not required in other packages now.

internal/fs/fs.go
ashmeenkaur: Rename operation leads to new object generation on GCS and hence a new inode in the map.
However, we do promote the inode to gcs synced inode in the flushPendingWrites() method which will then eventually be updated after GSFuse discovers the fresh generation of file.

https://github.com/GoogleCloudPlatform/gcsfuse/blob/337c65a887d6782cfd642995b2c98872784be755/internal/fs/fs.go#L1216

tools/integration_tests/local_file/read_dir_test.go
ashmeenkaur: Yes, I'll do it in the refactoring PR

internal/fs/fs.go
meet2mky: got it. Thanks!

tools/integration_tests/streaming_writes/local_file_test.go
ashmeenkaur: Yes, thats correct!

Peiqi0714:
I wonder why at first gcsfuse only support rename a file under buffered write? Is there any historical reasons? Thans a lot!! @ashmeenkaur @Tulsishah

ashmeenkaur:
@Peiqi0714 This has been a tech debt for some time. No historical reasons.

PR #3371, opened 2025-05-29
Add support for truncate to lower for streaming writes and fix promotion to generation backed Inode in Write and Truncate Flow.

meet2mky:
### Description
Add support for truncate to lower size than current file size during streaming writes. We are going to finalise the file and fallback to stage disk. This PR also fixes the promotion to generation backed Inodes for both out of order write flow and truncate to lower size scenario.

b/421367385

### Testing details
1. Manual - Done
2. Unit tests - Updated.

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -261,10 +261,9 @@ func (wh *bufferedWriteHandlerImpl) SetMtime(mtime time.Time) {
 
 func (wh *bufferedWriteHandlerImpl) Truncate(size int64) error {
 	if size < wh.totalSize {
-		logger.Errorf("BufferedWriteHandler.TruncateToLowerSizeError for object: %s, file size: %d, truncate size: %d",
-			wh.uploadHandler.objectName, wh.totalSize, size)
 		return ErrTruncateSizeLessThanFileSize
 	}
+
 	wh.truncatedSize = size
 	return nil
 }
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -88,7 +88,6 @@ type WriteFileInfo struct {
 }
 
 var ErrOutOfOrderWrite = errors.New("outOfOrder write detected")
-var ErrTruncateSizeLessThanFileSize = errors.New("truncate size less than file size")
 
 type CreateBWHandlerRequest struct {
 	Object                   *gcs.Object
@@ -261,7 +260,7 @@ func (wh *bufferedWriteHandlerImpl) SetMtime(mtime time.Time) {
 
 func (wh *bufferedWriteHandlerImpl) Truncate(size int64) error {
 	if size < wh.totalSize {
-		return ErrTruncateSizeLessThanFileSize
+		return ErrOutOfOrderWrite
 	}
 
 	wh.truncatedSize = size
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -88,6 +88,7 @@ type WriteFileInfo struct {
 }
 
 var ErrOutOfOrderWrite = errors.New("outOfOrder write detected")
+var ErrTruncateSizeLessThanFileSize = errors.New("truncate size less than file size")
 
 type CreateBWHandlerRequest struct {
 	Object                   *gcs.Object
@@ -260,9 +261,10 @@ func (wh *bufferedWriteHandlerImpl) SetMtime(mtime time.Time) {
 
 func (wh *bufferedWriteHandlerImpl) Truncate(size int64) error {
 	if size < wh.totalSize {
-		return fmt.Errorf("cannot truncate to lesser size when upload is in progress")
+		logger.Errorf("BufferedWriteHandler.TruncateToLowerSizeError for object: %s, file size: %d, truncate size: %d",
+			wh.uploadHandler.objectName, wh.totalSize, size)
+		return ErrTruncateSizeLessThanFileSize
 	}
-
 	wh.truncatedSize = size
 	return nil
 }
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1572,7 +1572,12 @@ func (fs *fileSystem) SetInodeAttributes(
 		if err != nil {
 			return
 		}
-		err = file.Truncate(ctx, int64(*op.Size))
+		gcsSynced, err := file.Truncate(ctx, int64(*op.Size))
+		// Sync the inode if finalize during truncate is successful
+		// even if the truncate operation later resulted error.
+		if gcsSynced {
+			fs.promoteToGenerationBacked(file)
+		}
 		if err != nil {
 			err = fmt.Errorf("truncate: %w", err)
 			return err
@@ -2654,7 +2659,7 @@ func (fs *fileSystem) WriteFile(
 		fs.mu.Unlock()
 
 		//TODO: Initialize BWH before invoking write()
-		if err := fh.Write(ctx, op.Data, op.Offset); err != nil {
+		if _, err := fh.Write(ctx, op.Data, op.Offset); err != nil {
 			return err
 		}
 		return
@@ -2673,7 +2678,12 @@ func (fs *fileSystem) WriteFile(
 	}
 
 	// Serve the request.
-	err = in.Write(ctx, op.Data, op.Offset, util.Write)
+	gcsSynced, err := in.Write(ctx, op.Data, op.Offset, util.Write)
+	// Sync the inode if finalize during write is successful
+	// even if the write operation later resulted in error.
+	if gcsSynced {
+		fs.promoteToGenerationBacked(in)
+	}
 	return
 }
 
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -161,7 +161,7 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 // unfinalized objects in zonal buckets, streaming writes is used.
 // Note that the writes are still done at the inode level.
 // LOCKS_EXCLUDED(fh.inode)
-func (fh *FileHandle) Write(ctx context.Context, data []byte, offset int64) error {
+func (fh *FileHandle) Write(ctx context.Context, data []byte, offset int64) (bool, error) {
 	fh.inode.Lock()
 	defer fh.inode.Unlock()
 
--- a/internal/fs/handle/file_test.go
+++ b/internal/fs/handle/file_test.go
@@ -116,7 +116,7 @@ func TestFileHandleWrite(t *testing.T) {
 	ctx := context.Background()
 	data := []byte("hello")
 
-	err := fh.Write(ctx, data, 0)
+	_, err := fh.Write(ctx, data, 0)
 
 	assert.Nil(t, err)
 	// Validate that write is successful at inode.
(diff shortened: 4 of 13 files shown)


internal/fs/inode/file.go
```diff
@@ -934,7 +934,22 @@ func (f *FileInode) Truncate(
 	size int64) (err error) {
 
 	if f.bwh != nil {
-		return f.bwh.Truncate(size)
+		err = f.bwh.Truncate(size)
+		if err == nil {
+			return
+		}
+	}
+	if err != nil && !errors.Is(err, bufferedwrites.ErrTruncateSizeLessThanFileSize) {
```
ashmeenkaur, line 942: We can remove the redundant err != nil check, as we would have already returned if err were nil. This check should also only occur when bwh != nil. Perhaps we should refactor this into a separate method.

Following the Go style guide for error handling, I recommend ordering the checks as follows:
- errors.Is(err, bufferedwrites.ErrTruncateSizeLessThanFileSize)
- err != nil
- The normal flow (not indented)

internal/fs/inode/file.go
```diff
@@ -934,7 +934,22 @@ func (f *FileInode) Truncate(
...
+		return fmt.Errorf("got unexpected error from bwh.Truncate(): %w", err)
+	}
+	// If truncate size is less than total size, flush and fall back to temp file.
+	if errors.Is(err, bufferedwrites.ErrTruncateSizeLessThanFileSize) {
+		logger.Infof("Falling back to staged writes on disk for file %s (inode %d) due to err: %v.", f.Name(), f.ID(), err.Error())
+		// Finalize the object.
+		err = f.flushUsingBufferedWriteHandler()
```
ashmeenkaur, line 949: Don't we have to promote the inode to generation backed inode if it was local? Seems like this is missing in out of order write flow as well.

internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -260,7 +261,9 @@ func (wh *bufferedWriteHandlerImpl) SetMtime(mtime time.Time) {
 
 func (wh *bufferedWriteHandlerImpl) Truncate(size int64) error {
 	if size < wh.totalSize {
-		return fmt.Errorf("cannot truncate to lesser size when upload is in progress")
+		logger.Errorf("BufferedWriteHandler.TruncateToLowerSizeError for object: %s, file size: %d, truncate size: %d",
```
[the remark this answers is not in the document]
meet2mky: Removed duplicate log from here and added warning log in file.go.

internal/fs/inode/file.go
```diff
@@ -934,7 +934,22 @@ func (f *FileInode) Truncate(
...
+		if err == nil {
+			return
+		}
+	}
+	if err != nil && !errors.Is(err, bufferedwrites.ErrTruncateSizeLessThanFileSize) {
+		return fmt.Errorf("got unexpected error from bwh.Truncate(): %w", err)
+	}
+	// If truncate size is less than total size, flush and fall back to temp file.
+	if errors.Is(err, bufferedwrites.ErrTruncateSizeLessThanFileSize) {
+		logger.Infof("Falling back to staged writes on disk for file %s (inode %d) due to err: %v.", f.Name(), f.ID(), err.Error())
```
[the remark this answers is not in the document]
meet2mky: Logging warning log here only, removed from buffered_write_handler.go

internal/fs/inode/file.go
meet2mky: I have fixed the flow as per review. I think refactoring we can do for both the Truncate and Write in a separate PR as both have similar error handling now.

internal/fs/inode/file.go
meet2mky: Made all the changes along with truncate method refactoring, integration tests and fs_tests. PTAL

internal/fs/fs.go
```diff
@@ -2654,7 +2659,7 @@ func (fs *fileSystem) WriteFile(
 		fs.mu.Unlock()
 
 		//TODO: Initialize BWH before invoking write()
-		if err := fh.Write(ctx, op.Data, op.Offset); err != nil {
+		if _, err := fh.Write(ctx, op.Data, op.Offset); err != nil {
```
meet2mky, line 2662: This change is under experimental flag and work in progress so I haven't changed it's code for this.

internal/fs/handle/file.go
```diff
@@ -161,7 +161,7 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 // unfinalized objects in zonal buckets, streaming writes is used.
 // Note that the writes are still done at the inode level.
 // LOCKS_EXCLUDED(fh.inode)
-func (fh *FileHandle) Write(ctx context.Context, data []byte, offset int64) error {
+func (fh *FileHandle) Write(ctx context.Context, data []byte, offset int64) (bool, error) {
```
meet2mky, line 164: This change is under experimental flag and work in progress so I haven't changed it's code for this. Will talk to anushka to align on this.

internal/fs/handle/file_test.go
```diff
@@ -116,7 +116,7 @@ func TestFileHandleWrite(t *testing.T) {
 	ctx := context.Background()
 	data := []byte("hello")
 
-	err := fh.Write(ctx, data, 0)
+	_, err := fh.Write(ctx, data, 0)
```
meet2mky, line 119: This change is under experimental flag and work in progress so I haven't changed it's code for this. Will talk to anushka to align on this.

internal/fs/inode/file_test.go
```diff
@@ -1052,22 +1070,20 @@ func (t *FileTest) TestTruncateDownwardWhenStreamingWritesAreEnabled() {
...
-				assert.Equal(t.T(), int64(7), t.in.bwh.WriteFileInfo().TotalSize)
-				// Fetch the attributes and check if the file size reflects the write.
-				attrs, err := t.in.Attributes(t.ctx)
-				require.NoError(t.T(), err)
-				assert.Equal(t.T(), uint64(7), attrs.Size)
-			}
-			t.createBufferedWriteHandler(!tc.performWrite)
-
-			err = t.in.Truncate(t.ctx, tc.truncateSize)
+			t.createBufferedWriteHandler(true)
```
ashmeenkaur, line 1083: Can we keep test case for how it was earlier - covering the edge case when truncate is done before initializing bwh?

can we also add a scenario for negative truncate size which should result in an error (as a separate test)

internal/fs/inode/file_test.go
```diff
@@ -1052,22 +1070,20 @@ func (t *FileTest) TestTruncateDownwardWhenStreamingWritesAreEnabled() {
...
+			attrs, err = t.in.Attributes(t.ctx)
+			require.NoError(t.T(), err)
+			assert.Equal(t.T(), uint64(7), attrs.Size)
+			gcsSynced, err = t.in.Truncate(t.ctx, tc.truncateSize)
 
-			require.Error(t.T(), err)
-			assert.ErrorContains(t.T(), err, "cannot truncate")
+			require.NoError(t.T(), err)
+			assert.True(t.T(), gcsSynced)
+			t.createBufferedWriteHandler(false)
```
ashmeenkaur, line 1096: why is this required?

tools/integration_tests/local_file/stat_file.go
```diff
@@ -56,12 +56,13 @@ func (t *CommonLocalFileTestSuite) TestStatOnLocalFileWithConflictingFileNameSuf
...
-func (t *localFileTestSuite) TestTruncateLocalFileToSmallerSize() {
+func (t *CommonLocalFileTestSuite) TestTruncateLocalFileToSmallerSize() {
 	testDirPath = setup.SetupTestDirectory(testDirName)
 	// Create a local file.
-	filePath, fh := CreateLocalFileInTestDir(ctx, storageClient, testDirPath, FileName1, t.T())
+	fileName := FileName1 + setup.GenerateRandomString(5)
+	filePath, fh := CreateLocalFileInTestDir(ctx, storageClient, testDirPath, fileName, t.T())
 	// Writing contents to local file .
-	WritingToLocalFileShouldNotWriteToGCS(ctx, storageClient, fh, testDirName, FileName1, t.T())
+	operations.WriteWithoutClose(fh, FileContents, t.T())
```
ashmeenkaur, line 65: This change is because of ZB, is it?

tools/integration_tests/streaming_writes/truncate_file_test.go
```diff
@@ -84,51 +86,116 @@ func (t *StreamingWritesSuite) TestWriteAfterTruncate() {
...
+			operations.WriteWithoutClose(t.f1, tc.writeContent, t.T())
+
+			operations.VerifyStatFile(t.filePath, int64(len(tc.finalContent)), FilePerms, t.T())
+			// Close the file and validate that the file is created on GCS.
+			CloseFileAndValidateContentFromGCS(ctx, storageClient, t.f1, testDirName, t.fileName, tc.finalContent, t.T())
+		})
+	}
+}
 
+func (t *StreamingWritesSuite) TestTruncateToLowerSyncsFileToGcsForStreamingWrites() {
```
ashmeenkaur: This overlaps with TestWriteAndTruncate, can we revert the changes in TestWriteAndTruncate and keep the 2 scenarios different like how it was earlier.

tools/integration_tests/streaming_writes/truncate_file_test.go
```diff
@@ -84,51 +86,116 @@ func (t *StreamingWritesSuite) TestWriteAfterTruncate() {
...
+	require.NoError(t.T(), err)
+	operations.VerifyStatFile(t.filePath, 3, FilePerms, t.T())
+	if !t.fallbackToDiskCase {
+		ValidateObjectContentsFromGCS(ctx, storageClient, testDirName, t.fileName, "foobar", t.T())
+	}
+	CloseFileAndValidateContentFromGCS(ctx, storageClient, t.f1, testDirName, t.fileName, "foo", t.T())
 }
 
-func (t *StreamingWritesSuite) TestTruncateToLowerSizeAfterWrite() {
+func (t *StreamingWritesSuite) TestTruncateToLowerSyncsFileToGcsAndDeletingFileDeletsFileFromGcs() {
```
ashmeenkaur: NIT:
```suggestion
func (t *StreamingWritesSuite) TestTruncateDownAndDeleteFile() {
```

tools/integration_tests/local_file/stat_file.go
meet2mky: yes, writing to local file can create file on GCS in case of ZB for streaming writes.

internal/fs/inode/file_test.go
meet2mky: Truncate is never done before trying to initialize bwh as that logic is at fs layer. This test case before also we have initialized bwh before every inode.Truncate call to ensure similar behaviour of fs layer.
Added negative size truncate tests to both the streaming writes file and current file which should cover all the cases.

internal/fs/inode/file_test.go
meet2mky: This is required to ensure that after truncate to size lower than current size or even zero size. We won't reinitialise bwh for it.

tools/integration_tests/streaming_writes/truncate_file_test.go
meet2mky: Done.

tools/integration_tests/streaming_writes/truncate_file_test.go
meet2mky: Removed this test as it is covered as part of other test. I am keeping the TestWriteAndTruncate as current one is more readable and covers both scenario for truncate to lower and upper.

ashmeenkaur:
@anushka567 FYI

ashmeenkaur (approved):

anushka567:
Noted. I will take care of the required changes when I am working on this. I have created a [bug](b/422075967) for the same to track it.

PR #3372, opened 2025-05-29
fix CVE-2025-27144 and upgrade direct dependencies

ashmeenkaur:
### Description
fix CVE-2025-27144 and upgrade direct dependencies

b/420967819

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Passed - failed interrupt test passed on retry, error happened due to server error (context deadline exceeded for GetFolder API call)
4. Perf tests - 
AYXXya9jXkYwESP

NA


abhishek10004:
I see a large improvement in the randread performance with 976MB file. Is this expected or just a random variation between runs?

abhishek10004 (approved):

ashmeenkaur:
> I see a large improvement in the randread performance with 976MB file. Is this expected or just a random variation between runs?

This is likely a setup issue, we've been seeing this for quite some runs and have an internal bug to find the root cause.

PR #3373, opened 2025-05-29
Refactor go installation to separate script.

meet2mky:
### Description
Refactor go installation to separate script which can be reused across other scripts where go is required.
b/423357486

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - Part of presubmit run.


Tulsishah (approved):
One comment

PR #3374, opened 2025-05-29
Disable grpc direct path for TPC

Tulsishah:
### Description

- TPC doesn't support a direct path in gRPC. 
- Currently, if the custom endpoint includes "tpc," we'll set `GOOGLE_CLOUD_DISABLE_DIRECT_PATH=true`. This conditional check is necessary because TPC environments presently require a custom endpoint.

[b/420794069](https://b.corp.google.com/issues/420794069), [b/421107527](https://b.corp.google.com/issues/421107527)

### Testing details
1. Manual - 
    i. Tested e2e tests with grpc cloud path on TPC
    ii. Tested e2e tests with http client.
    iii. Non tpc env will get tested with automated e2e tests.
3. Unit tests - NA

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -144,6 +144,10 @@ func setRetryConfig(sc *storage.Client, clientConfig *storageutil.StorageClientC
 	}
 }
 
+func isTPCEnv(endpoint string) bool {
+	return strings.Contains(endpoint, "tpc")
+}
+
 // Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
 func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig, enableBidiConfig bool) (sc *storage.Client, err error) {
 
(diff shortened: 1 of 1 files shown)


internal/storage/storage_handle.go
```diff
@@ -146,6 +151,20 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		logger.Fatal("error setting direct path env var: %v", err)
 	}
 
+	// TPC does not support direct path in gRPC.(b/420794069#comment8)
+	// Sets GOOGLE_CLOUD_DISABLE_DIRECT_PATH to "true" if the CustomEndpoint contains "tpc".
+	// TODO: Remove this check once the direct path is available in TPC environment.
+	if strings.Contains(clientConfig.CustomEndpoint, "tpc") {
+		if err := os.Setenv(disableDirectPath, "true"); err != nil {
+			logger.Fatal("error setting disable direct path env var: %v", err)
+		}
```
charith87, line 160: Should this logic be moved inside isDirectPathPossible() and the error appropriately thrown will be handled in Line 355. So line 353 change of again checking the same "tpc" is not needed

Tulsishah:
This is not require as per [b/420794069](https://buganizer.corp.google.com/issues/420794069)

PR #3377, opened 2025-06-02
[Random Reader Refactoring] Add seek improvement in new reader

Tulsishah:
### Description
fetch changes from this PR - https://github.com/GoogleCloudPlatform/gcsfuse/pull/3327/files
### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/client_readers/gcs_reader.go
+++ b/internal/gcsx/client_readers/gcs_reader.go
@@ -65,6 +65,10 @@ type GCSReader struct {
 
 	sequentialReadSizeMb int32
 
+	// Specifies the next expected offset for the reads. Used to distinguish between
+	// sequential and random reads.
+	expectedOffset int64
+
 	// seeks represents the number of random reads performed by the reader.
 	seeks uint64
 
@@ -101,6 +105,12 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx.R
 		gr.seeks++
 	}
 
+	// If the data can't be served from the existing reader, then we need to update the seeks.
+	// If current offset is not same as expected offset, its a random read.
+	if gr.expectedOffset != 0 && gr.expectedOffset != offset {
+		gr.seeks++
+	}
+
 	readReq := &gcsx.GCSReaderRequest{
 		Buffer:    p,
 		Offset:    offset,
@@ -127,11 +137,13 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx.R
 	if readerType == RangeReaderType {
 		readerResponse, err = gr.rangeReader.ReadAt(ctx, readReq)
 		gr.totalReadBytes += uint64(readerResponse.Size)
+		gr.updateExpectedOffset(offset + int64(readerResponse.Size))
 		return readerResponse, err
 	}
 
 	readerResponse, err = gr.mrr.ReadAt(ctx, readReq)
 	gr.totalReadBytes += uint64(readerResponse.Size)
+	gr.updateExpectedOffset(offset + int64(readerResponse.Size))
 
 	return readerResponse, err
 }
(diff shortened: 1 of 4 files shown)


internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -101,6 +105,12 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx.R
 		gr.seeks++
```
raj-prince: Now, we need to revert this seek increase, otherwise double counting will happen.

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -127,11 +137,13 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx.R
 	if readerType == RangeReaderType {
 		readerResponse, err = gr.rangeReader.ReadAt(ctx, readReq)
 		gr.totalReadBytes += uint64(readerResponse.Size)
+		gr.updateExpectedOffset(offset + int64(readerResponse.Size))
```
raj-prince: Should we do this as part of defer statement? - As we always want to update the expected offset.

internal/gcsx/client_readers/gcs_reader.go
raj-prince: Just saw the below change, where the above method will false in that case.

Now, invalidateReaderIfMisalignedOrTooSmall method always returns false, can we simply the method by removing the return type?

internal/gcsx/client_readers/gcs_reader_test.go
```diff
@@ -571,3 +572,45 @@ func (t *gcsReaderTest) Test_ReadAt_WithAndWithoutReadConfig() {
 		})
 	}
 }
+
```
raj-prince, line 584: I see validation of seek for some unit test-cases are missing? - E.g., Test_ReadAt_ValidateReadType
Good to add those test-cases as part of this PR only.

Tulsishah: Added for this `Test_ReadAt_ValidateReadType` test. Other tests were asserting for mrr or range reader. with the refactoring it is being part of GCS Reader so it is not required to assert there.

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -116,6 +114,9 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx.R
 		Offset:    offset,
 		EndOffset: -1,
 	}
+	defer func() {
+		gr.updateExpectedOffset(offset + int64(readerResponse.Size))
+	}()
 
 	readerResponse, err := gr.rangeReader.readFromExistingReader(ctx, readReq)
```
raj-prince: Why are we creating the readerResponse again, as we have already declared on the top? This might cause confusion for readerResponse value in the defer function.

internal/gcsx/client_readers/range_reader.go
```diff
@@ -317,11 +317,7 @@ func (rr *RangeReader) skipBytes(offset int64) {
 // Parameters:
 //   - offset: the starting byte position of the requested read.
 //   - p: the buffer representing the size of the requested read.
-//
```
raj-prince, line 320: You need to correct the above comment also. - `If the reader is discarded due to misalignment, the method
// returns true to signal that a seek should be recorded.`

internal/gcsx/client_readers/gcs_reader.go
Tulsishah: I used := for brevity to declare both variables in a single line without needing to initialize err separately, but I agree with your point regarding clarity.

raj-prince (approved):

PR #3379, opened 2025-06-04
[[Draft PR]] Testing perf with atomics/mutex for updating random reader shared state

abhishek10004:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -186,7 +186,7 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 
 		var objectData gcsx.ObjectData
 		readType := fh.reader.ReadType()
-		if !fh.inode.Bucket().BucketType().Zonal || readType == util.Sequential {
+		if !fh.inode.Bucket().BucketType().Zonal || readType == common.ReadTypeSequential {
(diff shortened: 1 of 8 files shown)


internal/fs/handle/file.go
```diff
@@ -128,7 +128,12 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 		fh.inode.Unlock()
 
 		var objectData gcsx.ObjectData
-		objectData, err = fh.reader.ReadAt(ctx, dst, offset)
+		readType := fh.reader.ReadType()
+		if !fh.inode.Bucket().BucketType().Zonal || readType == util.Sequential {
```
Tulsishah, line 132: Why can't we put only one condition `!fh.inode.Bucket().BucketType().Zonal`?

internal/fs/handle/file.go
```diff
@@ -128,7 +128,12 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 		fh.inode.Unlock()
```
Tulsishah, line 185: tryEnsureReader requires fh.lock. Won't that cause issue?

PR #3381, opened 2025-06-04
[Random reader refactoring] Integrate a read manager instance with a file handle to support the new read flow, controlled by a hidden flag

Tulsishah:
### Description
- Integrate a read manager instance with a file handle to support the new read flow, controlled by a hidden flag.
- This will enable new reader implementation.
- Small fixes in file cache reader.

[b/422343678](https://b.corp.google.com/issues/422343678)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Tested
4. Perf tests - 
BR5mLXS5YRs4j2u

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2015 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -196,6 +196,7 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		metricHandle:               serverCfg.MetricHandle,
 		enableAtomicRenameObject:   serverCfg.NewConfig.EnableAtomicRenameObject,
 		globalMaxWriteBlocksSem:    semaphore.NewWeighted(serverCfg.NewConfig.Write.GlobalMaxBlocks),
+		enableNewReader:            serverCfg.NewConfig.EnableNewReader,
 	}
 
 	// Set up root bucket
@@ -492,6 +493,8 @@ type fileSystem struct {
 	// Limits the max number of blocks that can be created across file system when
 	// streaming writes are enabled.
 	globalMaxWriteBlocksSem *semaphore.Weighted
+
+	enableNewReader bool
 }
 
 ////////////////////////////////////////////////////////////////////////
@@ -2613,7 +2616,7 @@ func (fs *fileSystem) ReadFile(
 		}
 	}
 	// Serve the read.
-	op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb)
+	op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb, fs.enableNewReader)
 
 	// As required by fuse, we don't treat EOF as an error.
 	if err == io.EOF {
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1,4 +1,4 @@
-// Copyright 2015 Google LLC
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -196,7 +196,6 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		metricHandle:               serverCfg.MetricHandle,
 		enableAtomicRenameObject:   serverCfg.NewConfig.EnableAtomicRenameObject,
 		globalMaxWriteBlocksSem:    semaphore.NewWeighted(serverCfg.NewConfig.Write.GlobalMaxBlocks),
-		enableNewReader:            serverCfg.NewConfig.EnableNewReader,
 	}
 
 	// Set up root bucket
@@ -493,9 +492,6 @@ type fileSystem struct {
 	// Limits the max number of blocks that can be created across file system when
 	// streaming writes are enabled.
 	globalMaxWriteBlocksSem *semaphore.Weighted
-
-	// If enableNewReader flag is set, we will take new reader implementation for reading file.
-	enableNewReader bool
 }
 
 ////////////////////////////////////////////////////////////////////////
@@ -2617,7 +2613,12 @@ func (fs *fileSystem) ReadFile(
 		}
 	}
 	// Serve the read.
-	op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb, fs.enableNewReader)
+
+	if fs.newConfig.EnableNewReader {
+		op.Dst, op.BytesRead, err = fh.ReadWithReadManager(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb)
+	} else {
+		op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb)
+	}
 
 	// As required by fuse, we don't treat EOF as an error.
 	if err == io.EOF {
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -494,6 +494,7 @@ type fileSystem struct {
 	// streaming writes are enabled.
 	globalMaxWriteBlocksSem *semaphore.Weighted
 
+	// If enableNewReader flag is set, we will take new reader implementation for reading file.
 	enableNewReader bool
 }
 
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -119,28 +119,35 @@ func (fh *FileHandle) Unlock() {
 // LOCKS_REQUIRED(fh.inode.mu)
 // UNLOCK_FUNCTION(fh.inode.mu)
 func (fh *FileHandle) ReadWithReadManager(ctx context.Context, dst []byte, offset int64, sequentialReadSizeMb int32) ([]byte, int, error) {
-	// Attempt to initialize or validate the read manager
+	// fh.inode.mu is already locked to ensure that we have a readManager for its current
+	// state, or clear fh.readManager if it's not possible to create one (probably
+	// because the inode is dirty).
 	err := fh.tryEnsureReadManager(ctx, sequentialReadSizeMb)
 	if err != nil {
 		fh.inode.Unlock()
 		return nil, 0, fmt.Errorf("tryEnsureReadManager: %w", err)
 	}
 
-	// If read manager is available, try reading through it
+	// If we have an appropriate readManager, unlock the inode and use that. This
+	// allows reads to proceed concurrently with other operations; in particular,
+	// multiple reads can run concurrently. It's safe because the user can't tell
+	// if a concurrent write started during or after a read.
 	if fh.readManager != nil {
 		fh.inode.Unlock()
 
-		readerResponse, readErr := fh.readManager.ReadAt(ctx, dst, offset)
-		if errors.Is(readErr, io.EOF) {
-			// Unexpected EOF — log and propagate
-			logger.Warnf("Unexpected EOF error encountered while reading, err: %v type: %T", readErr, readErr)
+		var readerResponse gcsx.ReaderResponse
+		readerResponse, err = fh.readManager.ReadAt(ctx, dst, offset)
+		switch {
+		case errors.Is(err, io.EOF):
+			if err != io.EOF {
+				logger.Warnf("Unexpected EOF error encountered while reading, err: %v type: %T ", err, err)
+			}
 			return nil, 0, io.EOF
-		}
-		if readErr != nil {
-			return nil, 0, fmt.Errorf("fh.readManager.ReadAt: %w", readErr)
+
+		case err != nil:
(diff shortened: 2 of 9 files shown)


internal/fs/fs.go
```diff
@@ -492,6 +493,9 @@ type fileSystem struct {
 	// Limits the max number of blocks that can be created across file system when
 	// streaming writes are enabled.
 	globalMaxWriteBlocksSem *semaphore.Weighted
+
```
raj-prince: Why are we storing this field as part of fs struct? As we'll have this value in the serverCfg.

internal/fs/fs.go
```diff
@@ -2613,7 +2617,7 @@ func (fs *fileSystem) ReadFile(
 		}
 	}
 	// Serve the read.
-	op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb)
+	op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb, fs.enableNewReader)
```
raj-prince: Instead of passing this boolean as part of every Read, you can pass while creation of FileHandle struct during the OpenFile call.

internal/fs/handle/file.go
```diff
@@ -109,51 +118,61 @@ func (fh *FileHandle) Unlock() {
...
+	// fh.inode.mu is already locked to ensure that we have a reader/readManager for its current
+	// state, or clear fh.reader/fh.readManager if it's not possible to create one (probably
 	// because the inode is dirty).
-	err = fh.tryEnsureReader(ctx, sequentialReadSizeMb)
-	if err != nil {
-		fh.inode.Unlock()
-		err = fmt.Errorf("tryEnsureReader: %w", err)
-		return
+	// If enableReadManager flag is set, we will take new reader implementation.
+	if enableReadManager {
```
raj-prince: Instead of adding if/else check in the existing method can we create a separate method for read-manager, which will be called if read-manager-reader is enabled. Later we can easily remove the unused code.

internal/fs/handle/file.go
```diff
@@ -221,3 +240,57 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
 	fh.reader = rr
 	return
 }
+
+// If possible, ensure that fh.readManager is set to an appropriate random reader
```
raj-prince: nit: random reader?

internal/fs/handle/file.go
```diff
@@ -221,3 +240,57 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
...
+	// If a read manager already exists and its object generation matches the inode's
+	// source generation, it's still valid. Just update its size and return.
+	if fh.readManager != nil && fh.readManager.Object().Generation == fh.inode.SourceGeneration().Object {
+		fh.readManager.Object().Size = fh.inode.SourceGeneration().Size
+		return nil
+	}
+
+	// If we reached here, either no read manager exists, or the existing one is outdated.
+	// Destroy any old read manager before creating a new one.
+	fh.destroyReadManager()
```
raj-prince, line 311: Why aren't we moving this insde `if fh.readManager != nil` check?

internal/fs/handle/file.go
```diff
@@ -221,3 +240,57 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
...
+	// Create a new read manager for the current inode state.
+	rm := read_manager.NewReadManager(fh.inode.Source(), fh.inode.Bucket(), &read_manager.ReadManagerConfig{
+		SequentialReadSizeMB:  sequentialReadSizeMb,
+		FileCacheHandler:      fh.fileCacheHandler,
+		CacheFileForRangeRead: fh.cacheFileForRangeRead,
+		MetricHandle:          fh.metricHandle,
+		MrdWrapper:            &fh.inode.MRDWrapper,
+		ReadConfig:            fh.readConfig,
+	})
+	fh.readManager = rm
```
raj-prince:
```suggestion
	fh.readManager = read_manager.NewReadManager(fh.inode.Source(), fh.inode.Bucket(), &read_manager.ReadManagerConfig{
		SequentialReadSizeMB:  sequentialReadSizeMb,
		FileCacheHandler:      fh.fileCacheHandler,
		CacheFileForRangeRead: fh.cacheFileForRangeRead,
		MetricHandle:          fh.metricHandle,
		MrdWrapper:            &fh.inode.MRDWrapper,
		ReadConfig:            fh.readConfig,
	})
```

internal/fs/handle/file.go
```diff
@@ -221,3 +240,57 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
...
+	return nil
+}
+
+// destroyReadManager is a helper function to safely destroy and nil the readManager.
+// This assumes the necessary locks (fh.mu, fh.inode.mu) are already held by the caller.
+func (fh *FileHandle) destroyReadManager() {
+	if fh.readManager != nil {
+		fh.readManager.Destroy()
+		fh.readManager = nil
+	}
```
raj-prince:
```suggestion
    if fh.readManager == nil {
        return
    }

    fh.readManager.Destroy()
    fh.readManager = nil
```

raj-prince (changes requested):

internal/fs/handle/file.go
Tulsishah: We are doing it in side the function.

internal/fs/fs.go
Tulsishah: Not require this as we are making check in fs read and opting new method in case of readmanager is enable via config.

internal/fs/handle/file.go
```diff
@@ -103,6 +112,51 @@ func (fh *FileHandle) Unlock() {
...
+		return readerResponse.DataBuf, readerResponse.Size, nil
+	}
+
+	// If read manager is not available, fall back to reading via inode
+	defer fh.inode.Unlock()
+
+	n, err := fh.inode.Read(ctx, dst, offset)
+
+	if errors.Is(err, io.EOF) {
+		err = nil
```
Tulsishah, line 153: We are anyway doing it in caller of fs.go, but doing it here makes sense as for file handle reads also we are doing it in this function only. 

Let me know if you find any concerns @raj-prince .

internal/fs/handle/file.go
```diff
@@ -103,6 +112,47 @@ func (fh *FileHandle) Unlock() {
...
+// UNLOCK_FUNCTION(fh.inode.mu)
+func (fh *FileHandle) ReadWithReadManager(ctx context.Context, dst []byte, offset int64, sequentialReadSizeMb int32) ([]byte, int, error) {
+	// Attempt to initialize or validate the read manager
+	err := fh.tryEnsureReadManager(ctx, sequentialReadSizeMb)
+	if err != nil {
+		fh.inode.Unlock()
+		return nil, 0, fmt.Errorf("tryEnsureReadManager: %w", err)
+	}
+
+	// If read manager is available, try reading through it
```
raj-prince: Why have we removed lock related comment?

```
// If we have an appropriate reader, unlock the inode and use that. This
	// allows reads to proceed concurrently with other operations; in particular,
	// multiple reads can run concurrently. It's safe because the user can't tell
	// if a concurrent write started during or after a read.
```

internal/fs/handle/file.go
```diff
@@ -103,6 +112,47 @@ func (fh *FileHandle) Unlock() {
...
+	if fh.readManager != nil {
+		fh.inode.Unlock()
+
+		readerResponse, readErr := fh.readManager.ReadAt(ctx, dst, offset)
+		if errors.Is(readErr, io.EOF) {
```
raj-prince: Any strong reason to change the switch case into if-else? If not I would suggest to keep same.

internal/fs/handle/file.go
```diff
@@ -221,3 +272,57 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
...
+	// If the inode is dirty (its source generation isn't authoritative),
+	// we cannot create a read manager based on the cloud object.
+	// Invalidate any existing read manager and return.
+	if !fh.inode.SourceGenerationIsAuthoritative() {
+		fh.destroyReadManager()
+		return nil
+	}
+
+	// If a read manager already exists and its object generation matches the inode's
+	// source generation, it's still valid. Just update its size and return.
```
raj-prince: I think, this size update is required only for zonal bucket, good to convey in the comment.

internal/gcsx/file_cache_reader.go
```diff
@@ -98,7 +98,7 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 		handleID = uint64(readOp.Handle)
 	}
 	requestID := uuid.New()
-	logger.Tracef("%.13v <- FileCache(%s:/%s, offset: %d, size: %d, handle: %d)", requestID, fc.bucket.Name(), fc.object.Name, offset, len(p), handleID)
```
raj-prince, line 101: Why this change?

internal/gcsx/file_cache_reader.go
```diff
@@ -133,14 +133,17 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 			switch {
 			case errors.Is(err, lru.ErrInvalidEntrySize):
 				logger.Warnf("tryReadingFromFileCache: while creating CacheHandle: %v", err)
+				err = nil
```
raj-prince, line 138: Can we update all the variable, cacheHit, bytesRead as these can be used by defer statment?

internal/fs/fs.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 Google LLC
+// Copyright 2025 Google LLC
```
raj-prince: Why are we changing this? Generally, we keep the copyright same year when we add the source file.

internal/fs/handle/file_test.go
```diff
@@ -102,31 +136,210 @@ func createFileInode(
...
+			fh := NewFileHandle(in, nil, false, common.NewNoopMetrics(), util.Read, &cfg.ReadConfig{})
+			buf := make([]byte, len(tc.expectedData))
+			fh.inode.Lock()
+			var (
+				output []byte
+				n      int
+				err    error
+			)
+
+			if tc.enableReadManager {
```
raj-prince: We should avoid adding if-else in the test. Consider breaking it into two tests, one for reader other for read-manager.

internal/fs/handle/file_test.go
```diff
@@ -102,31 +136,210 @@ func createFileInode(
...
+
+// Test_Read_Success validates successful read behavior when using either
+// the readManager or the random reader. It verifies that the correct data
+// is returned with no error.
+func (t *fileTest) Test_Read_Success() {
+	type testCase struct {
+		name              string
+		enableReadManager bool
+		expectedData      []byte
+		expectErr         bool
```
raj-prince: Where are we use expectErr? If not please remove.

internal/fs/handle/file_test.go
```diff
@@ -102,31 +136,210 @@ func createFileInode(
...
+			assert.Equal(t.T(), len(tc.expectedData), n)
+			assert.Equal(t.T(), tc.expectedData, output)
+		})
+	}
+}
+
+// Test_Read_ErrorScenarios verifies that both io.EOF and other error conditions
+// are correctly handled by ReadWithReadManager and Read, returning the right
+// byte count, output, and error values.
+func (t *fileTest) Test_Read_ErrorScenarios() {
```
raj-prince, line 236: Same here, break it into two test-case one of reader other for read-manager. And for other tests also.

internal/fs/handle/file_test.go
```diff
@@ -102,31 +137,238 @@ func createFileInode(
...
+	for _, tc := range tests {
+		t.Run(tc.name, func() {
+			t.SetupTest()
+			fh := tc.mockSetup(t, dst)
+
+			output, n, err := fh.Read(t.ctx, dst, 0, 200, tc.useReadManager)
+
+			assert.NoError(t.T(), err)
+			assert.Equal(t.T(), len(objectData), n)
+			assert.Equal(t.T(), objectData, output[:n])
```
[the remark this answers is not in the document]
raj-prince: Please write it as comment in the test.

raj-prince (commented):
Suggested few nits.

internal/gcsx/file_cache_reader.go
Tulsishah: some e2e test was failing due to additional `,`.

internal/fs/handle/file.go
Tulsishah: I think with some refactoring it got removed. Added it.

internal/fs/handle/file.go
Tulsishah: I think a switch statement would be faster than nested if-else blocks. In this case, there were only two conditions, so I used if-else, but a switch statement would also work.

internal/fs/fs.go
Tulsishah: I was having lint issue pointing to this so I changed it, not sure why reverting it again is not showing up.

internal/fs/handle/file_test.go
Tulsishah: I thought I can avoid same setup by utilizing TDD format, but if you think so I split it in two.

internal/gcsx/file_cache_reader.go
raj-prince: Ohh, okay, that's the old log.

We need to change the log as well as e2e test to adapt this change. But we can take that as another task.

PR #3387, opened 2025-06-06
zero byte read to fetch size for unfinalized obj

anushka567:
### Description
Currently, stat calls incorrectly report the size of unfinalized objects as zero. This poses a significant blocker for unfinalized object support, as our read/write operations depend on accurate size information. As a temporary workaround, we're implementing a zero-byte reader. This is because the reader's attributes correctly reflect the unfinalized object's size.

Caveat: Listing is slightly slow.
[b/422944268](b/422944268)

### Testing details
1. Manual - Manual ls, stat show correct object size.
2. Unit tests - Skipping as this is a temporary fix.
3. Integration tests - [Automated](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/e4580def-3dfa-405a-9d62-7bb8d5caaea2/summary)

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -161,7 +161,7 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 
 // Note: This is not production ready code and will be removed once StatObject
 // requests return correct attr values for appendable objects.
-func (bh *bucketHandle) updateObjectSizeFromZeroByteReader(ctx context.Context, attrs *storage.ObjectAttrs) error {
+func (bh *bucketHandle) fetchLatestSizeOfUnfinalizedObject(ctx context.Context, attrs *storage.ObjectAttrs) error {
 	if bh.BucketType().Zonal && bh.enableRapidAppends {
 		// Get object handle
 		obj := bh.bucket.Object(attrs.Name)
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -143,9 +143,9 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 		return
 	}
 	if attrs.Finalized.IsZero() {
-		if err = bh.updateObjectSizeFromZeroByteReader(ctx, attrs); err != nil {
-		  err = fmt.Errorf("failed to fetch the latest size of unfinalized object %q: %w", attrs.Name, err)
-		  return
+		if err = bh.fetchLatestSizeOfUnfinalizedObject(ctx, attrs); err != nil {
+			err = fmt.Errorf("failed to fetch the latest size of unfinalized object %q: %w", attrs.Name, err)
+			return
 		}
 	}
 
@@ -168,8 +168,7 @@ func (bh *bucketHandle) fetchLatestSizeOfUnfinalizedObject(ctx context.Context, 
 		// Create a new reader
 		reader, err := obj.NewRangeReader(ctx, 0, 0)
 		if err != nil {
-			return fmt.Errorf("failed to create zero-byte reader for object %q: %w", attrs.Name, err)
-			return err
+			return fmt.Errorf("failed to create zero-byte reader for object %q: %v", attrs.Name, err)
 		}
 		err = reader.Close()
 		if err != nil {
@@ -433,7 +432,7 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 			return
 		}
 		if attrs.Finalized.IsZero() {
-			err = bh.updateObjectSizeFromZeroByteReader(ctx, attrs)
+			err = bh.fetchLatestSizeOfUnfinalizedObject(ctx, attrs)
 		}
 
 		// Prefix attribute will be set for the objects returned as part of Prefix[] array in list response.
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -39,10 +39,11 @@ const FullBucketPathHNS = "projects/_/buckets/%s"
 
 type bucketHandle struct {
 	gcs.Bucket
-	bucket        *storage.BucketHandle
-	bucketName    string
-	bucketType    *gcs.BucketType
-	controlClient StorageControlClient
+	bucket             *storage.BucketHandle
+	bucketName         string
+	bucketType         *gcs.BucketType
+	controlClient      StorageControlClient
+	enableRapidAppends bool
 }
 
 func (bh *bucketHandle) Name() string {
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -433,7 +433,8 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 		}
 		if attrs.Finalized.IsZero() {
 			if err = bh.fetchLatestSizeOfUnfinalizedObject(ctx, attrs); err != nil {
-			  err = fmt.Errorf("failed to fetch the latest size of unfinalized object %q: %w", attrs.Name, err)
+				err = fmt.Errorf("failed to fetch the latest size of unfinalized object %q: %w", attrs.Name, err)
+				return
 			}
 		}
 
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -165,7 +165,7 @@ func (bh *bucketHandle) updateObjectSizeFromZeroByteReader(ctx context.Context, 
 		// Create a new reader
 		reader, err := obj.NewRangeReader(ctx, 0, 0)
 		if err != nil {
-			logger.Debugf("failed to create zero-byte reader for object %q: %v", attrs.Name, err)
+			return fmt.Errorf("failed to create zero-byte reader for object %q: %w", attrs.Name, err)
 			return err
 		}
 		err = reader.Close()
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -165,11 +165,12 @@ func (bh *bucketHandle) updateObjectSizeFromZeroByteReader(ctx context.Context, 
 		// Create a new reader
 		reader, err := obj.NewRangeReader(ctx, 0, 0)
 		if err != nil {
-			return fmt.Errorf("failed to create zero byte reader: %w", err)
+			logger.Debugf("failed to create zero-byte reader for object %q: %v", attrs.Name, err)
+			return err
 		}
 		err = reader.Close()
 		if err != nil {
-			logger.Debugf("Failed to closed zero-byte reader for %s", attrs.Name)
+			logger.Warnf("failed to close zero-byte reader for object %q: %v", attrs.Name, err)
 		}
 
 		// Set the size
@@ -420,9 +421,6 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 		var attrs *storage.ObjectAttrs
 
 		attrs, err = itr.Next()
-		if attrs.Finalized.IsZero() {
-			err = bh.updateObjectSizeFromZeroByteReader(ctx, attrs)
-		}
 		if err == iterator.Done {
 			err = nil
 			break
@@ -430,6 +428,9 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 		if err != nil {
 			err = fmt.Errorf("error in iterating through objects: %w", err)
 			return
+		}
+		if attrs.Finalized.IsZero() {
+			err = bh.updateObjectSizeFromZeroByteReader(ctx, attrs)
 		}
 
 		// Prefix attribute will be set for the objects returned as part of Prefix[] array in list response.
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -142,6 +142,9 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 		err = fmt.Errorf("error in fetching object attributes: %w", err)
 		return
 	}
+	if attrs.Finalized.IsZero() {
+		err = bh.updateObjectSizeFromZeroByteReader(ctx, attrs)
+	}
 
 	// Converting attrs to type *Object
 	o := storageutil.ObjectAttrsToBucketObject(attrs)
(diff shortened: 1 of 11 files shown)


internal/storage/storageutil/zero_byte_reader.go
```diff
@@ -0,0 +1,37 @@
...
+package storageutil
+
+import (
+	"context"
+	"fmt"
+
+	"cloud.google.com/go/storage"
+)
+
+func GetObjectSizeFromZeroByteReader(ctx context.Context, bh *storage.BucketHandle, objectName string) (int64, error) {
```
vadlakondaswetha: No unit tests for this and ideally this should be bucket_handle.go

If we are expecting this to be temporary and will be removed before AGA, can you add a comment that this is not production ready code

internal/storage/bucket_handle.go
```diff
@@ -141,6 +141,13 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 		err = fmt.Errorf("error in fetching object attributes: %w", err)
 		return
 	}
+	// TODO(anushkadhn): read only for unfinalized object. To be done after PR#3365 is merged
+	if req.ReadWhileStat {
```
vadlakondaswetha: so many changes to get readWhileStat to this file.

Can we simply check if the bucket is zonal and enableRapidAppends are true, make a call. You need to pass enableRapidAppends flag to bucket_handle only during initialization.

internal/storage/bucket_handle.go
```diff
@@ -391,6 +415,12 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 		var attrs *storage.ObjectAttrs
 
 		attrs, err = itr.Next()
+		if bh.BucketType().Zonal && bh.enableRapidAppends && attrs.Finalized.IsZero() {
```
gargnitingoogle: This logic (new block) is exactly the same as in StatObject. It's best to extract this out into a smaller private utility function, which can then also be unit tested more easily potentially.

internal/storage/bucket_handle.go
```diff
@@ -141,6 +142,12 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 		err = fmt.Errorf("error in fetching object attributes: %w", err)
 		return
 	}
+	if bh.BucketType().Zonal && bh.enableRapidAppends && attrs.Finalized.IsZero() {
+		size, err := bh.getObjectSizeFromZeroByteReader(ctx, req.Name)
+		if err == nil {
+			attrs.Size = size
+		}
```
gargnitingoogle, line 149: Why aren't we returning here if err is not nil?

Beside that, we should log error if err is not nil, even if we don't return on it. At least that way, it'll make debugging failures easier.

anushka567: Thanks for pointing out, What should be the ideal thing to do when - we have created the reader successfully, extracted the size attribute as well but while closing the reader, an error is encountered?

internal/storage/storage_handle.go
```diff
@@ -65,6 +65,7 @@ type storageClient struct {
 	clientConfig             storageutil.StorageClientConfig
 	storageControlClient     StorageControlClient
 	directPathDetector       *gRPCDirectPathDetector
+	enableRapidAppends       bool
```
gargnitingoogle: I don't think we need to store it in storageClient. We just need it in bucketHandle. 
For passing this to bucketHandle, we just need to pass it from https://github.com/GoogleCloudPlatform/gcsfuse/blob/6a3dd66d5e737f09c55f0098d2e5a7eb15518827/internal/gcsx/bucket_manager.go#L170 . This can reduce the code changes a lot.

internal/storage/bucket_handle.go
```diff
@@ -152,6 +159,23 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 	return
 }
 
+// Note: This code is not production ready and will be removed once StatObject
+// requests return correct object size for unfinalized objects.
+func (bh *bucketHandle) getObjectSizeFromZeroByteReader(ctx context.Context, objectName string) (int64, error) {
+	// Get object handle
+	obj := bh.bucket.Object(objectName)
```
gargnitingoogle: nit: We can avoid this look-up if we directly pass the objectHandle to this function instead of objectName, which is already available in the caller function.

internal/storage/storage_handle.go
anushka567: For this approach,I'll have to update the bucketCfg itself since we need to access the mount config object  here based on which the bucketHandle will be created, Considering that, the amount of code changes is nearly the same for both.

internal/storage/bucket_handle.go
```diff
@@ -152,6 +159,23 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 	return
 }
 
+// Note: This code is not production ready and will be removed once StatObject
+// requests return correct object size for unfinalized objects.
+func (bh *bucketHandle) getObjectSizeFromZeroByteReader(ctx context.Context, objectName string) (int64, error) {
```
gargnitingoogle: Is it possible to unit test this ?

internal/storage/bucket_handle.go
gargnitingoogle: > What should be the ideal thing to do when - we have created the reader successfully, extracted the size attribute as well but while closing the reader, an error is encountered?

I'd say log an error and continue if it's the failure is only in Close. Error in closing the reader is not necessarily a functional blocker for us, but it can create a resource-leak and problems if these accumulate.
Failure in creating a ranged-reader itself is a functional issue and should probably stop the flow.

internal/storage/storage_handle.go
gargnitingoogle: I tried out this change. PTAL at https://github.com/GoogleCloudPlatform/gcsfuse/compare/78dda6d0dc819ffa439a584893e7f5ac6336a7a3...1d78cb5bfb3fa80735c6788dd176978a53f257af . Net code change will be near-zero or slightly better, but it's structurally better to keep this flag only at the bucketHandle where it's needed.

gargnitingoogle:
@anushka567 Also PTAL at why the linux-tests job is failing.

internal/storage/storage_handle.go
anushka567: Updated.

internal/storage/bucket_handle.go
anushka567: This method is invoked at two points-
- StatObject: object handle available in caller function
- ListObject: object attrs directly available and not object handle

Hence, i dont think it is feasible.

internal/storage/bucket_handle.go
```diff
@@ -152,6 +156,29 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
...
+		// Get object handle
+		obj := bh.bucket.Object(attrs.Name)
+		// Create a new reader
+		reader, err := obj.NewRangeReader(ctx, 0, 0)
+		if err != nil {
+			return fmt.Errorf("failed to create zero byte reader: %w", err)
+		}
+		err = reader.Close()
+		if err != nil {
+			logger.Debugf("Failed to closed zero-byte reader for %s", attrs.Name)
```
gargnitingoogle:
```suggestion
			logger.Warnf("Failed to close zero-byte reader for object %q: %v", attrs.Name, err)
```

internal/storage/bucket_handle.go
```diff
@@ -152,6 +156,29 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
...
+// Note: This is not production ready code and will be removed once StatObject
+// requests return correct attr values for appendable objects.
+func (bh *bucketHandle) updateObjectSizeFromZeroByteReader(ctx context.Context, attrs *storage.ObjectAttrs) error {
+	if bh.BucketType().Zonal && bh.enableRapidAppends {
+		// Get object handle
+		obj := bh.bucket.Object(attrs.Name)
+		// Create a new reader
+		reader, err := obj.NewRangeReader(ctx, 0, 0)
+		if err != nil {
+			return fmt.Errorf("failed to create zero byte reader: %w", err)
```
gargnitingoogle:
```suggestion
			return fmt.Errorf("failed to create zero-byte reader for object %q: %w", attrs.Name, err)
```

internal/storage/bucket_handle.go
```diff
@@ -391,6 +418,9 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 		var attrs *storage.ObjectAttrs
 
 		attrs, err = itr.Next()
+		if attrs.Finalized.IsZero() {
```
gargnitingoogle: This new block should be after the next two checks (i.e. after line number 431). The next 2 checks are for checking if the iterator got the next objects attributes correctly. So, this check for finalized and size calculation should be after them.

internal/storage/bucket_handle.go
anushka567: Not feasible.

anushka567:
> @anushka567 Also PTAL at why the linux-tests job is failing.

The tests are passing now.

internal/storage/bucket_handle.go
```diff
@@ -141,6 +142,9 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 		err = fmt.Errorf("error in fetching object attributes: %w", err)
 		return
 	}
+	if attrs.Finalized.IsZero() {
```
gargnitingoogle, line 145: nit: readability

The check `attrs.Finalized.IsZero()` can be absorbed in the function bh.updateObjectSizeFromZeroByteReader to avoid duplication and to make the caller code cleaner.

internal/storage/bucket_handle.go
```diff
@@ -152,6 +156,30 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 	return
 }
 
+// Note: This is not production ready code and will be removed once StatObject
+// requests return correct attr values for appendable objects.
+func (bh *bucketHandle) updateObjectSizeFromZeroByteReader(ctx context.Context, attrs *storage.ObjectAttrs) error {
```
gargnitingoogle: nit: readability

```suggestion
func (bh *bucketHandle) fetchLatestSizeOfUnfinalizedObject(ctx context.Context, attrs *storage.ObjectAttrs) error {
```

internal/storage/bucket_handle.go
```diff
@@ -141,6 +142,9 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 		err = fmt.Errorf("error in fetching object attributes: %w", err)
 		return
 	}
+	if attrs.Finalized.IsZero() {
+		err = bh.updateObjectSizeFromZeroByteReader(ctx, attrs)
```
gargnitingoogle: Return here if `err != nil` .

internal/storage/bucket_handle.go
```diff
@@ -152,6 +156,30 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
...
+// Note: This is not production ready code and will be removed once StatObject
+// requests return correct attr values for appendable objects.
+func (bh *bucketHandle) updateObjectSizeFromZeroByteReader(ctx context.Context, attrs *storage.ObjectAttrs) error {
+	if bh.BucketType().Zonal && bh.enableRapidAppends {
+		// Get object handle
+		obj := bh.bucket.Object(attrs.Name)
+		// Create a new reader
+		reader, err := obj.NewRangeReader(ctx, 0, 0)
+		if err != nil {
+			logger.Debugf("failed to create zero-byte reader for object %q: %v", attrs.Name, err)
```
gargnitingoogle: Not sure why you chose to just log this error string, and not propagate in the error object .

```suggestion
			return fmt.Errorf("failed to create zero-byte reader for object %q: %w", attrs.Name, err)
```

internal/storage/bucket_handle.go
gargnitingoogle:
```suggestion
		if err = bh.updateObjectSizeFromZeroByteReader(ctx, attrs); err != nil {
		  err = fmt.Errorf("failed to fetch the latest size of unfinalized object %q: %w", attrs.Name, err)
		  return
		}
```

gargnitingoogle (commented):
Some minor comments (readability and error handling). Functionally LGTM .

internal/storage/bucket_handle.go
anushka567: Given that the method is `fetchLatestSizeOfUnfinalizedObject`, I believe it's more appropriate to keep this check in the calling context. This ensures that we only invoke `fetchLatestSizeOfUnfinalizedObject` for objects that are indeed unfinalized, thereby avoiding unnecessary method calls for objects that are already finalized.

internal/storage/bucket_handle.go
```diff
@@ -401,6 +431,9 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 			err = fmt.Errorf("error in iterating through objects: %w", err)
 			return
 		}
+		if attrs.Finalized.IsZero() {
+			err = bh.fetchLatestSizeOfUnfinalizedObject(ctx, attrs)
```
gargnitingoogle: Minor miss. Same fix as in StatObject .

```suggestion
			if err = bh.fetchLatestSizeOfUnfinalizedObject(ctx, attrs); err != nil {
			  err = fmt.Errorf("failed to fetch the latest size of unfinalized object %q: %w", attrs.Name, err)
			}
```

internal/storage/bucket_handle.go
```diff
@@ -141,6 +142,12 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 		err = fmt.Errorf("error in fetching object attributes: %w", err)
 		return
 	}
+	if attrs.Finalized.IsZero() {
+		if err = bh.fetchLatestSizeOfUnfinalizedObject(ctx, attrs); err != nil {
```
vadlakondaswetha, line 146: Ideally you should return the size from this method and update the attrs here. 
Since this is not production ready code, i am okay with this change.

internal/storage/bucket_handle_test.go
```diff
@@ -1496,7 +1496,7 @@ func (testSuite *BucketHandleTest) TestBucketHandleWithError() {
 	var err error
 	// Test when the client returns an error.
 	testSuite.mockClient.On("GetStorageLayout", mock.Anything, mock.Anything, mock.Anything).Return(x, errors.New("mocked error"))
-	testSuite.bucketHandle, err = testSuite.storageHandle.BucketHandle(context.Background(), TestBucketName, "")
+	testSuite.bucketHandle, err = testSuite.storageHandle.BucketHandle(context.Background(), TestBucketName, "", false)
```
vadlakondaswetha, line 1499: can you add a test case for this new method/flow

gargnitingoogle: @vadlakondaswetha Do you mean with value `true` ?

gargnitingoogle (approved):
LGTM

internal/storage/bucket_handle_test.go
anushka567: Added a test for successful creation of buckethandle with rapid appends set to true.

gargnitingoogle (approved):

vadlakondaswetha (approved):

PR #3388, opened 2025-06-06
setting generation for object handle before creating takeover writer

anushka567:
### Description
The object handle for creating the takeover writer must be set before method invocation.

[b/422892135](b/422892135)
### Testing details
1. Manual - Tested as part of POC, that the writer is successfully getting created. 
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -152,14 +152,14 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 	return
 }
 
-func (bh *bucketHandle) getObjectHandleWithGenerationAndPreconditionSet(ctx context.Context, req *gcs.CreateObjectChunkWriterRequest) *storage.ObjectHandle {
+func (bh *bucketHandle) getObjectHandleWithGenerationAndPreconditionSet(ctx context.Context, req *gcs.CreateObjectChunkWriterRequest) (*storage.ObjectHandle, error) {
 	objHandle := bh.getObjectHandleWithPreconditionsSet(&req.CreateObjectRequest)
 	attrs, err := bh.bucket.Object(req.Name).Attrs(ctx)
 	if err != nil {
 		err = fmt.Errorf("error in fetching object attributes: %w", err)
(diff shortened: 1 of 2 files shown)


internal/storage/bucket_handle.go
```diff
@@ -152,6 +152,16 @@ func (bh *bucketHandle) StatObject(ctx context.Context,
 	return
 }
 
+func (bh *bucketHandle) getObjectHandleWithGenerationAndPreconditionSet(ctx context.Context, req *gcs.CreateObjectChunkWriterRequest) (*storage.ObjectHandle, error) {
+	objHandle := bh.getObjectHandleWithPreconditionsSet(&req.CreateObjectRequest)
+	attrs, err := bh.bucket.Object(req.Name).Attrs(ctx)
```
vadlakondaswetha: wont this line(157) result in a network call to GCS?

anushka567: Yes, this would have resulted in a network call , which is unnecessary . Removed this call, we will pass the object generation as part of the request itself since `Generation` field can be populated from min object attributes itself.

internal/storage/bucket_handle.go
```diff
@@ -245,7 +245,7 @@ func (bh *bucketHandle) CreateObjectChunkWriter(ctx context.Context, req *gcs.Cr
 
 func (bh *bucketHandle) CreateAppendableObjectWriter(ctx context.Context,
 	req *gcs.CreateObjectChunkWriterRequest) (gcs.Writer, error) {
-	obj := bh.getObjectHandleWithPreconditionsSet(&req.CreateObjectRequest)
+	obj := bh.getObjectHandleWithPreconditionsSet(&req.CreateObjectRequest).Generation(req.Generation)
```
vadlakondaswetha: please split it into next line to improve readability.
and add comment why this is required.

anushka567: Updated.

vadlakondaswetha (approved):

PR #3389, opened 2025-06-09
appending to unfinalized objects in ZB using streaming writes flow with takeover writer

anushka567:
### Description
This PR enables streaming writes support for append to unfinalized objects on ZB. 
Changes are made behind hidden flag: `--write-experimental-enable-rapid-appends`

**Key changes include**

* File Open Mode: The file's openMode is now passed down from the filesystem layer to the inode layer to determine if a write is an append.

* Conditional BWH Initialization: A new check, areBufferedWritesSupported, has been added to the file inode to decide if buffered writes can be used for appends.

* Appendable Object Writer: When appending, the upload_handler now creates takeover writer with an offset set to the current size of the object, ensuring data is appended rather than overwritten.

* BWH Size Initialization: A new SetTotalSize() method initializes the BWH's tracked size to the size of the existing object, ensuring correct offset calculations for subsequent writes (to avoid being treated as out-of-order write)

* Unit Tests: Comprehensive unit tests have been added at all relevant levels to validate the new logic for append scenarios.

Behavior Analysis Doc : [link](https://docs.google.com/document/d/1rgum0-p6ESz79nIrPDADjuXW-GZMeOdhJzGJ2LRaLOI/edit?usp=sharing)

[b/422075967](b/422075967)
[b/423517346](b/423517346)

### Testing details
1. Manual - Manual testing conclusion: Writes successful for appendable objects using takeover writer. Written data persists in GCS bucket.
2. Unit tests - Included as part of PR, and in follow up PR [#3424](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3424)
3. Integration tests - Automated, [results](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/0a6efc05-aec0-482a-ba63-454077af3954/summary).

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -58,9 +58,6 @@ type BufferedWriteHandler interface {
 
 	// Unlink cancels the ongoing upload and free up the buffers.
 	Unlink()
-
-	// SetTotalSize sets the totalSize attr of BWH to size of the object.
-	SetTotalSize()
 }
 
 // bufferedWriteHandlerImpl is responsible for filling up the buffers with the data
@@ -108,6 +105,10 @@ func NewBWHandler(req *CreateBWHandlerRequest) (bwh BufferedWriteHandler, err er
 	if err != nil {
 		return
 	}
+	var size int64
+	if req.Object != nil {
+		size = int64(req.Object.Size)
+	}
 
 	bwh = &bufferedWriteHandlerImpl{
 		current:   nil,
@@ -121,7 +122,7 @@ func NewBWHandler(req *CreateBWHandlerRequest) (bwh BufferedWriteHandler, err er
 			BlockSize:                req.BlockSize,
 			ChunkTransferTimeoutSecs: req.ChunkTransferTimeoutSecs,
 		}),
-		totalSize:     0,
+		totalSize:     size,
 		mtime:         time.Now(),
 		truncatedSize: -1,
 	}
@@ -311,8 +312,4 @@ func (wh *bufferedWriteHandlerImpl) Unlink() {
 		// Only logging an error in case of resource leak.
 		logger.Errorf("blockPool.ClearFreeBlockChannel() failed: %v", err)
 	}
-}
-
-func (wh *bufferedWriteHandlerImpl) SetTotalSize() {
-	wh.totalSize = int64(wh.uploadHandler.obj.Size)
 }
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -58,6 +58,9 @@ type BufferedWriteHandler interface {
 
 	// Unlink cancels the ongoing upload and free up the buffers.
 	Unlink()
+
+	// SetTotalSize sets the totalSize attr of BWH to size of the object.
+	SetTotalSize()
 }
 
 // bufferedWriteHandlerImpl is responsible for filling up the buffers with the data
@@ -308,4 +311,8 @@ func (wh *bufferedWriteHandlerImpl) Unlink() {
 		// Only logging an error in case of resource leak.
 		logger.Errorf("blockPool.ClearFreeBlockChannel() failed: %v", err)
 	}
+}
+
+func (wh *bufferedWriteHandlerImpl) SetTotalSize() {
+	wh.totalSize = int64(wh.uploadHandler.obj.Size)
 }
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -111,6 +111,18 @@ func (uh *UploadHandler) createObjectWriter() (err error) {
 	return
 }
 
+func (uh *UploadHandler) createAppendableObjectWriter() (err error) {
+	cReq := gcs.NewCreateObjectRequest(uh.obj, uh.objectName, nil, uh.chunkTransferTimeout)
+	req := gcs.CreateObjectChunkWriterRequest{
+		CreateObjectRequest: *cReq,
+		Offset:              int64(uh.obj.Size),
+	}
+	var ctx context.Context
+	ctx, uh.cancelFunc = context.WithCancel(context.Background())
+	uh.writer, err = uh.bucket.CreateAppendableObjectWriter(ctx, &req)
+	return err
+}
+
 func (uh *UploadHandler) UploadError() (err error) {
 	if uploadError := uh.uploadError.Load(); uploadError != nil {
 		err = *uploadError
@@ -168,7 +180,12 @@ func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
 
 func (uh *UploadHandler) ensureWriter() error {
 	if uh.writer == nil {
-		err := uh.createObjectWriter()
+		var err error
+		if uh.obj.Size != 0 {
+			err = uh.createAppendableObjectWriter()
+		} else {
+			err = uh.createObjectWriter()
+		}
 		if err != nil {
 			return fmt.Errorf("createObjectWriter failed for object %s: %w", uh.objectName, err)
 		}
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -116,7 +116,6 @@ func (uh *UploadHandler) createAppendableObjectWriter() (err error) {
 	req := gcs.CreateObjectChunkWriterRequest{
 		CreateObjectRequest: *cReq,
 		Offset:              int64(uh.obj.Size),
-		Generation:          uh.obj.Generation,
 	}
 	var ctx context.Context
 	ctx, uh.cancelFunc = context.WithCancel(context.Background())
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -116,6 +116,7 @@ func (uh *UploadHandler) createAppendableObjectWriter() (err error) {
 	req := gcs.CreateObjectChunkWriterRequest{
 		CreateObjectRequest: *cReq,
 		Offset:              int64(uh.obj.Size),
+		Generation:          uh.obj.Generation,
 	}
 	var ctx context.Context
 	ctx, uh.cancelFunc = context.WithCancel(context.Background())
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -110,6 +110,7 @@ func (uh *UploadHandler) createObjectWriter() (err error) {
 	if uh.obj != nil && uh.obj.Size != 0 {
 		chunkWriterReq := gcs.CreateObjectChunkWriterRequest{
 			CreateObjectRequest: *req,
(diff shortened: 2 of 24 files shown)


internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -58,6 +58,9 @@ type BufferedWriteHandler interface {
 
 	// Unlink cancels the ongoing upload and free up the buffers.
 	Unlink()
+
+	// SetTotalSize sets the totalSize attr of BWH to size of the object.
+	SetTotalSize()
```
vadlakondaswetha: No need of seperate method. this should come via constructor. Any concerns in doing so?

internal/bufferedwrites/upload_handler.go
```diff
@@ -107,7 +107,16 @@ func (uh *UploadHandler) createObjectWriter() (err error) {
 	// (and context will be cancelled) by the time complete upload is done.
 	var ctx context.Context
 	ctx, uh.cancelFunc = context.WithCancel(context.Background())
-	uh.writer, err = uh.bucket.CreateObjectChunkWriter(ctx, req, int(uh.blockSize), nil)
+	if uh.obj != nil && uh.obj.Size != 0 {
```
vadlakondaswetha: does uh.object.size has the correct value?
in that case bwh also has the object size right? why do we need seperate method?

internal/fs/fs.go
```diff
@@ -2661,13 +2661,27 @@ func (fs *fileSystem) WriteFile(
...
-		if _, err := fh.Write(ctx, op.Data, op.Offset); err != nil {
+		in.Lock()
+		err = fs.initBufferedWriteHandlerAndSyncFileIfEligible(ctx, in, fh.OpenMode())
+		in.Unlock()
+		if err != nil {
+			return
+		}
+
+		// Serve the request.
+		gcsSynced, err := fh.Write(ctx, op.Data, op.Offset)
```
vadlakondaswetha: all the logic is duplicated across if-else. the only logic which should be behind if -else is

if enableRapidAppends {
   fh.write()
} else {
   inode.write
}

internal/fs/inode/file.go
```diff
@@ -1045,7 +1052,20 @@ func (f *FileInode) InitBufferedWriteHandlerIfEligible(ctx context.Context) (boo
...
 		return true, nil
 	}
 	return false, nil
 }
+
+func (f *FileInode) areBufferedWritesSupported(openMode util.OpenMode, obj *gcs.Object) bool {
+	if f.src.Size == 0 {
+		return true
+	}
+	if openMode == util.Append && f.bucket.BucketType().Zonal && obj.Finalized.IsZero() {
```
vadlakondaswetha, line 1065: why are you checking finalized on latest object and size on existing object. incase of ZB, both can change after inode is cached. we can rely on the server to send the correct error incase both changed

internal/fs/inode/file.go
```diff
@@ -1006,16 +1006,18 @@ func (f *FileInode) CreateEmptyTempFile(ctx context.Context) (err error) {
...
 	// bwh already initialized, do nothing.
 	if f.bwh != nil {
 		return false, nil
 	}
 
 	tempFileInUse := f.content != nil
-	if f.src.Size != 0 || !f.config.Write.EnableStreamingWrites || tempFileInUse {
-		// bwh should not be initialized under these conditions.
-		return false, nil
+	if !f.config.Write.ExperimentalEnableRapidAppends {
```
vadlakondaswetha: why we need to add a condition on experimentalEnableRapidAppends. why not same as what is mentioned in design doc. only change is you need to check for the flag also in areBufferedWritesSupported method

internal/fs/fs.go
```diff
@@ -2661,13 +2661,27 @@ func (fs *fileSystem) WriteFile(
 	if fs.newConfig.Write.ExperimentalEnableRapidAppends {
```
vadlakondaswetha: when this flag is enabled, we are switching to fh.write for regional buckets too. we need to ensure we have test coverage for that also

internal/bufferedwrites/buffered_write_handler.go
anushka567: Updated this to initialize size based on latest gcs object if returned, else 0 during the time of bwh creation .

internal/fs/inode/file.go
anushka567: Updated.

internal/fs/inode/file.go
anushka567: Updated.

internal/fs/fs.go
anushka567: Will take it up in a follow-up PR

internal/bufferedwrites/upload_handler.go
```diff
@@ -107,7 +107,16 @@ func (uh *UploadHandler) createObjectWriter() (err error) {
 	// (and context will be cancelled) by the time complete upload is done.
 	var ctx context.Context
 	ctx, uh.cancelFunc = context.WithCancel(context.Background())
-	uh.writer, err = uh.bucket.CreateObjectChunkWriter(ctx, req, int(uh.blockSize), nil)
+	if uh.obj == nil || !uh.obj.Finalized.IsZero() {
```
vadlakondaswetha: we should check hte bucket type too just incase finalized is not correctly set on regional buckets.

internal/fs/inode/file.go
```diff
@@ -1049,3 +1053,17 @@ func (f *FileInode) InitBufferedWriteHandlerIfEligible(ctx context.Context) (boo
...
 	return false, nil
 }
+
+func (f *FileInode) areBufferedWritesSupported(openMode util.OpenMode, obj *gcs.Object) bool {
+	// For new files and existing files of size 0, buffered writes are always supported.
+	if f.local || obj.Size == 0 {
+		return true
+	}
+
+	if f.config.Write.ExperimentalEnableRapidAppends {
```
vadlakondaswetha: we should always do short-circuits instead of having nested if-else

if !f.config.write.experimentalEnableRapidAppends {
   return false.
}

internal/fs/fs.go
```diff
@@ -2658,32 +2658,31 @@ func (fs *fileSystem) WriteFile(
...
 	fs.mu.Lock()
+	fh := fs.handles[op.Handle].(*handle.FileHandle)
 	in := fs.fileInodeOrDie(op.Inode)
 	fs.mu.Unlock()
 
-	in.Lock()
-	defer in.Unlock()
+	var gcsSynced bool
+	if fs.newConfig.Write.ExperimentalEnableRapidAppends {
+		err = fs.initBufferedWriteHandlerAndSyncFileIfEligible(ctx, in, fh.OpenMode())
```
vadlakondaswetha: why can't it be always fh.openMode() ?? that way we can move this line out of if-else

internal/fs/fs.go
```diff
@@ -2658,32 +2658,31 @@ func (fs *fileSystem) WriteFile(
...
+		if err != nil {
+			return
+		}
+		// Serve the request via the file handle.
+		gcsSynced, err = fh.Write(ctx, op.Data, op.Offset)
+	} else {
+		in.Lock()
```
vadlakondaswetha: lock and unlock is not done in the if clause.
do we need lock for initBufferedWRiteHAndlerandSyncFileIfEligible method??

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -67,6 +71,90 @@ func (t *UploadHandlerTest) SetupSubTest() {
 	t.SetupTest()
 }
 
+func (t *UploadHandlerTest) createUploadHandlerWithNonNilObjectOfGivenSize(size uint64, finalized time.Time) {
```
vadlakondaswetha: can you rename this to createUploadHandlerWithObjectOfGivenSize
the object is nonnil always

internal/fs/inode/file_test.go
```diff
@@ -180,6 +180,81 @@ func (t *FileTest) TestName() {
...
+		{
+			name:       "WriteToNonNilObj",
+			content:    nonNilContents,
+			bucketType: gcs.BucketType{},
+			finalized:  finalizedTime,
+			openMode:   util.Write,
+			supported:  false,
+		},
+		{
+			name:       "WriteToNilObj",
```
vadlakondaswetha: Nil word is little confusing. WriteToEmptyObject

internal/fs/fs.go
anushka567: Inode Lock is taken before invoking inode.Write().

In the if condition, we are performing writes via filehandle, and in fileHandle.Write() , we already take the inode lock in the method itself.
So only in else clause, taking the inode lock is required.

internal/bufferedwrites/upload_handler.go
anushka567: Outdated condition.

internal/bufferedwrites/upload_handler.go
anushka567: Updated condition.

internal/bufferedwrites/upload_handler.go
```diff
@@ -107,7 +107,16 @@ func (uh *UploadHandler) createObjectWriter() (err error) {
 	// (and context will be cancelled) by the time complete upload is done.
 	var ctx context.Context
 	ctx, uh.cancelFunc = context.WithCancel(context.Background())
-	uh.writer, err = uh.bucket.CreateObjectChunkWriter(ctx, req, int(uh.blockSize), nil)
+	if uh.bucket.BucketType().Zonal && (uh.obj != nil && uh.obj.Finalized.IsZero()) {
```
anushka567, line 110: Currently we do not create takeover writer for a new object on ZB , as that will introduce unfinalized object via GCSFuse , and the support for it is yet not complete.

Thus, takeover writer only created for unfinalized object on ZB of size >= 0.

internal/fs/fs.go
vadlakondaswetha: initBufferedWriteHandlerAndSyncFileIfEligible() method definition says it requires a inode lock. Internally it calls syncFile() which needs a inode lock. please check again.

anushka567: Noted, 
We need to hold the inode lock when initBWH... is. called.
After that,
when the write is via the fileHandle, inode must be unlocked before fh.Write() is called as the fh.Write() takes inode lock within the method.

when the write is via inode , we have to explicitly hold the inode before inode.Write().. So i am changing the code to
the following, if we want to keep the initBWH common , then this is required, PTAL @vadlakondaswetha 
```
	in.Lock()
	defer in.Unlock()
	if err = fs.initBufferedWriteHandlerAndSyncFileIfEligible(ctx, in, fh.OpenMode()); err != nil {
		return
	}
	if fs.newConfig.Write.ExperimentalEnableRapidAppends {
		// Serve the request via the file handle.
		in.Unlock()
		defer in.Lock()
		gcsSynced, err = fh.Write(ctx, op.Data, op.Offset)
	} else {
		// Serve the request.
		gcsSynced, err = in.Write(ctx, op.Data, op.Offset, util.Write)
	}
```

PR #3401, opened 2025-06-10
[read-unfinalized-object] [part1] Relax file-cache checks for unfinalized objects

gargnitingoogle:
### Description
The change is based on [design doc](https://docs.google.com/document/d/1eKAsVbvmiBYtPs8h3jbapN8jP_Pmegb4_AW8WlTdNts/edit?usp=sharing&resourcekey=0-z2SvePxO8kTVnL64-qHq5w)

Changes in behavior for unfinalized objects:
- Don't invalidate file-cache entry if size of completed download entry >= intended downloaded size
- Don't fall back to GCS from file-cache read if size of completed download entry >= intended upper limit of offset to be read.
- Fallback to GCS while reading from cache_handle if requested read isn't available from the cached file-info data size. If partial read is possible from the cache, then go ahead and start the download job.

Additional changes
- Added a simple utility for easily checking if an object is unfinalized.
- Enable zonal bucket unit tests for TestFileCacheReader, TestRandomReader and TestReadManager.
- Add new unit tests corresponding to the above code changes.
- Update existing unit tests corresponding to the above code changes.

~Dependencies~
- ~This builds on top of #3387 and is thus blocked for merge until that PR is merged~

Out of scope:
- This does not add support for reading unfinalized objects itself. That's covered in the in https://github.com/GoogleCloudPlatform/gcsfuse/pull/3405 .
- The changes in this don't touch any non-zonal non-unfinalized objects in any way.

### Link to the issue in case of a bug fix.
[b/421294827](http://b/421294827)

### Testing details
1. Manual - Tested manually.
2. Unit tests - Tested automatically by this PR. Add more UTs for the changes in the code.
3. Integration tests - ran with presubmit ([only zonal](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/b21fc64b-349f-4287-bff6-2bc06371fc02/details), [e2e zonal, non-zonal and perf test run](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/fa4f953d-79d2-445e-93a7-3306010e4d72))
4. Perf tests results
```md
+--------+------------+--------------+------------+--------------+--------------+
| Branch | File Size  |   Read BW    |  Write BW  | RandRead BW  | RandWrite BW |
+--------+------------+--------------+------------+--------------+--------------+
[...]

PR #3405, opened 2025-06-11
[read-unfinalized-object] [part2] Enable read of unfinalized objects while being written via streaming

gargnitingoogle:
### Description
**What this change does**:
- ~This builds on top of #3401 . So, it shouldn't be merged before that one.~
- This pull request enables reading from GCS objects while they are being written to via streaming.
- This change is particularly relevant for zonal buckets. 
- The changes primarily involves modifying the conditions under which a read is permitted and ensuring data is flushed to GCS before a read occurs.
- The associated unit and integration tests have been updated to validate this new behavior.
- Undoes changes from #3250 

**What it does not do**:
- It doesn't add new e2e tests for testing reads of unfinalized objects. That task is separately tracked in [b/421361399](http://421361399) .

[b/425828097](http://b/425828097)
[b/410698332](http://b/410698332)
[b/422355580](http://b/422355580)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Ran ([run](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/4b6912a9-8436-49bb-b598-0f093450db4f), [run](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/0c1664e4-b451-4949-b3e0-d80e69279077), [run](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/3a579676-35b9-44c8-8e7f-5d3342cd5239), [run](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/007fbb49-62b5-4cb8-9b26-e6f20824bf99)) as part of presubmit .

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2593,9 +2593,8 @@ func (fs *fileSystem) ReadFile(
 		// when we start leaving zonal bucket objects unfinalized.
 		if !fh.Inode().Bucket().BucketType().Zonal {
 			err = fs.flushFile(ctx, fh.Inode())
-		} else if fh.Inode().Source().IsUnfinalized() {
-			// Flush, but don't finalized for unfinalized objects
-			// in zonal buckets.
+		} else {
+			// Flush but don't finalize, in zonal bucket.
 			err = fs.syncFile(ctx, fh.Inode())
 		}
 		if err != nil {
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2587,10 +2587,17 @@ func (fs *fileSystem) ReadFile(
 	fh.Lock()
 	fh.Inode().Lock()
 	defer fh.Unlock()
-	// TODO(b/417136852): Remove bucket type check when we start leaving zonal bucket objects unfinalized.
-	// Flush Pending streaming writes file for regional bucket and issue read within same inode lock.
-	if fh.Inode().IsUsingBWH() && !fh.Inode().Bucket().BucketType().Zonal {
-		err = fs.flushFile(ctx, fh.Inode())
+	if fh.Inode().IsUsingBWH() {
+		// Flush Pending streaming writes and issue read within same inode lock.
+		// TODO(b/417136852): Remove bucket type check and call only flushFile
+		// when we start leaving zonal bucket objects unfinalized.
+		if !fh.Inode().Bucket().BucketType().Zonal {
+			err = fs.flushFile(ctx, fh.Inode())
+		} else if fh.Inode().Source().IsUnfinalized() {
+			// Flush, but don't finalized for unfinalized objects
+			// in zonal buckets.
+			err = fs.syncFile(ctx, fh.Inode())
+		}
 		if err != nil {
 			fh.Inode().Unlock()
 			return err
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -170,18 +170,6 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// fh.inode.mu is already locked to ensure that we have a reader for its current
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
-
-	// TODO: Remove this if-block once we stop finalizing
-	// objects in zonal bucket on writes by default (b/426512250).
-	if fh.inode.Bucket().Bucket.BucketType().Zonal && fh.inode.Source().IsUnfinalized() {
-		_, err = fh.inode.SyncPendingBufferedWrites()
-		if err != nil {
-			fh.inode.Unlock()
-			err = fmt.Errorf("fh.inode.SyncPendingBufferedWrites: %w", err)
-			return
-		}
-	}
-
 	err = fh.tryEnsureReader(ctx, sequentialReadSizeMb)
 	if err != nil {
 		fh.inode.Unlock()
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -171,11 +171,13 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
 
-	_, err = fh.inode.SyncPendingBufferedWrites()
-	if err != nil {
-		fh.inode.Unlock()
-		err = fmt.Errorf("fh.inode.SyncPendingBufferedWrites: %w", err)
-		return
+	if fh.inode.Bucket().Bucket.BucketType().Zonal && fh.inode.Source().IsUnfinalized() {
+		_, err = fh.inode.SyncPendingBufferedWrites()
+		if err != nil {
+			fh.inode.Unlock()
+			err = fmt.Errorf("fh.inode.SyncPendingBufferedWrites: %w", err)
+			return
+		}
 	}
 
 	err = fh.tryEnsureReader(ctx, sequentialReadSizeMb)
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -171,6 +171,8 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
 
+	// TODO: Remove this if-block once we stop finalizing
+	// objects in zonal bucket on writes by default (b/426512250).
 	if fh.inode.Bucket().Bucket.BucketType().Zonal && fh.inode.Source().IsUnfinalized() {
 		_, err = fh.inode.SyncPendingBufferedWrites()
 		if err != nil {
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -170,6 +170,14 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// fh.inode.mu is already locked to ensure that we have a reader for its current
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
+
+	_, err = fh.inode.SyncPendingBufferedWrites()
+	if err != nil {
+		fh.inode.Unlock()
+		err = fmt.Errorf("fh.inode.SyncPendingBufferedWrites: %w", err)
+		return
+	}
+
 	err = fh.tryEnsureReader(ctx, sequentialReadSizeMb)
 	if err != nil {
 		fh.inode.Unlock()
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/v3/cfg"
@@ -394,7 +393,8 @@ func (f *FileInode) Source() *gcs.MinObject {
 func (f *FileInode) SourceGenerationIsAuthoritative() bool {
 	// Source generation is authoritative if:
 	//   1.  No pending writes exists on the inode (both content and bwh are nil).
-	return f.content == nil && f.bwh == nil
+	//   2.  The bucket is zonal and there are no pending writes in the temporary file.
+	return (f.content == nil && f.bwh == nil) || (f.bucket.BucketType().Zonal && f.content == nil)
 }
 
 // Equivalent to the generation returned by f.Source().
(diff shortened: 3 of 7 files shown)


internal/fs/inode/file.go
```diff
@@ -551,11 +550,6 @@ func (f *FileInode) Read(
 	ctx context.Context,
 	dst []byte,
 	offset int64) (n int, err error) {
-	// It is not nil when streaming writes are enabled and bucket type is Zonal.
```
gargnitingoogle, line 554: Don't completely remove this condition. Disable it only for zb, unfinalized objects.

gargnitingoogle: done

internal/fs/handle/file.go
```diff
@@ -170,6 +170,14 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// fh.inode.mu is already locked to ensure that we have a reader for its current
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
+
+	_, err = fh.inode.SyncPendingBufferedWrites()
```
gargnitingoogle: Add comment to describe why this is needed.

tools/integration_tests/streaming_writes/common_streaming_writes_suite_test.go
```diff
@@ -54,7 +54,7 @@ func (t *StreamingWritesSuite) validateReadCall(fh *os.File, content string) {
 	readContent := make([]byte, len(content))
 	n, err := fh.ReadAt(readContent, 0)
 	// TODO(b/417136852): Fix validation once zb reads start working.
-	if setup.IsZonalBucketRun() && !t.fallbackToDiskCase {
+	if false /*setup.IsZonalBucketRun() && !t.fallbackToDiskCase*/ {
```
gargnitingoogle: Remove this `false` case altogether.

internal/fs/handle/file.go
```diff
@@ -171,11 +171,13 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
 
-	_, err = fh.inode.SyncPendingBufferedWrites()
-	if err != nil {
-		fh.inode.Unlock()
-		err = fmt.Errorf("fh.inode.SyncPendingBufferedWrites: %w", err)
-		return
+	if fh.inode.Bucket().Bucket.BucketType().Zonal && fh.inode.Source().IsUnfinalized() {
```
gargnitingoogle: Add comment to describe this block.

tools/integration_tests/streaming_writes/common_streaming_writes_suite_test.go
gargnitingoogle: done

internal/fs/handle/file.go
meet2mky: This would be required for now but once we set the `--finalizeOnClose=false` on the client then the flush would also behave like Sync() for ZB and we can just remove Zonal bucket condition check in the fs.go ReadFile method.

gargnitingoogle: thanks @meet2mky . Should I add a comment/TODO here then ?

internal/fs/inode/file.go
```diff
@@ -553,8 +553,11 @@ func (f *FileInode) Read(
 	offset int64) (n int, err error) {
 	// It is not nil when streaming writes are enabled and bucket type is Zonal.
 	if f.bwh != nil {
-		err = fmt.Errorf("cannot read a file when upload in progress: %w", syscall.ENOTSUP)
-		return
+		// Allow reading from unfinalized objects (which are supported only in zonal buckets as of now).
+		if f.bucket == nil || !f.bucket.BucketType().Zonal || !f.src.IsUnfinalized() {
```
meet2mky: - This seems redundant right? `f.bucket == nil` f.bucket must not be nil here.

- Also as per comment on top if `f.bwh != nil` then this is ZONAL bucket guaranteed so we may remove the redundant check.
- Also this check shouldn't come here right? streaming writes when in-progress can't be read through inode and they need to go through GCS only?

meet2mky (commented):
Take a look at this PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3250
I think some changes in SourceGenerationIsAuthoritative needs to be restored. It is discussed in the review comments of that PR.

internal/fs/handle/file.go
meet2mky: yeah, I think it's better to add a TODO because then this particular call will become redundant and no-op.

internal/fs/inode/file.go
gargnitingoogle: Done.

gargnitingoogle:
> Take a look at this PR: #3250 I think some changes in SourceGenerationIsAuthoritative needs to be restored. It is discussed in the review comments of that PR.

Undone all the changes from PR #3250 here in consultation with @meet2mky .

internal/fs/inode/file.go
```diff
@@ -551,11 +551,6 @@ func (f *FileInode) Read(
 	ctx context.Context,
 	dst []byte,
 	offset int64) (n int, err error) {
-	// It is not nil when streaming writes are enabled and bucket type is Zonal.
-	if f.bwh != nil {
```
meet2mky, line 555: Should we add a error log here, As this is our invariant that reads with streaming writes are not supposed to come through FileInode as behaviour is undefined. @ashmeenkaur Wdyt?
```
	if f.bwh != nil {
		err = fmt.Errorf("unexpected read call to FileInode %v Read() when streaming writes in progress for file: %s", f.ID(), f.Name().LocalName())
		return
	}
```

meet2mky (approved):
LGTM, One comment, Overall looks good.

internal/fs/inode/file.go
gargnitingoogle: I've added this error back as discussed offline with @meet2mky .

meet2mky: We can remove the comment on top of it. error shouldn't be `SYSCALL.ENOTSUP`. Because it shouldn't even come to this function.

internal/fs/inode/file.go
```diff
@@ -551,6 +552,11 @@ func (f *FileInode) Read(
 	ctx context.Context,
 	dst []byte,
 	offset int64) (n int, err error) {
+	// It is not nil when streaming writes are enabled and bucket type is Zonal.
```
meet2mky: Nit: This comment is not needed.

internal/fs/inode/file.go
```diff
@@ -551,6 +552,11 @@ func (f *FileInode) Read(
...
+	if f.bwh != nil {
+		err = fmt.Errorf("Unexpected read call for %q when steaming write is in progress for it: %w", f.Name().LocalName(), syscall.ENOTSUP)
```
meet2mky: Nit: Error string shouldn't be captialized. go/go-style/decisions#error-strings

meet2mky: Nit: typo `streaming`

meet2mky (approved):
LGTM, minor comments.

internal/fs/inode/file.go
gargnitingoogle: done

internal/fs/inode/file.go
gargnitingoogle: done

internal/fs/inode/file.go
gargnitingoogle: done

gargnitingoogle: done

meet2mky (approved):

internal/fs/handle/file.go
```diff
@@ -170,6 +170,18 @@ func (fh *FileHandle) Read(ctx context.Context, dst []byte, offset int64, sequen
 	// fh.inode.mu is already locked to ensure that we have a reader for its current
 	// state, or clear fh.reader if it's not possible to create one (probably
 	// because the inode is dirty).
+
+	// TODO: Remove this if-block once we stop finalizing
```
ashmeenkaur: IIUC, this check should remain because there are no plans to revert finalized objects to an un-finalized state in the long run. Shall we remove this TODO?

ashmeenkaur: On a separate node, what happens when we have a bwh on an empty object if it is already finalized? creating a bwh on it and doing a flush will change the object generation IIUC. So first flush should leave the object in unfinalized state but should update the object generation and inode state (from local to gen backed) in fs maps. We do handle something similar for regional buckets in the fs.go ReadFile method. Can we do something similar here?

meet2mky: If you do a flush on a finalized empty Object it creates a new unfinalized empty object.

meet2mky: > We do handle something similar for regional buckets in the fs.go ReadFile method. Can we do something similar here?

Can you explain this a bit more, I didn't get what you mean.

ashmeenkaur: > Can you explain this a bit more, I didn't get what you mean.

I was referring to how we update the file system (FS) maps within the fs.go's ReadFile method by finalizing objects for regional buckets. You can see this at: https://github.com/GoogleCloudPlatform/gcsfuse/blob/e1c05a86ec7289bf0bad6de76f887ccdcedcf637/internal/fs/fs.go#L2593

My main concern is to confirm that the inode in the FS maps is also updated with this new object generation in such cases. (It might already be handled, but I want to be certain.)

meet2mky: This is already taken care within this method for ZB.
https://github.com/GoogleCloudPlatform/gcsfuse/blob/e1c05a86ec7289bf0bad6de76f887ccdcedcf637/internal/fs/fs.go#L1234-L1255

internal/fs/fs.go
```diff
@@ -2587,10 +2587,17 @@ func (fs *fileSystem) ReadFile(
...
-	if fh.Inode().IsUsingBWH() && !fh.Inode().Bucket().BucketType().Zonal {
-		err = fs.flushFile(ctx, fh.Inode())
+	if fh.Inode().IsUsingBWH() {
+		// Flush Pending streaming writes and issue read within same inode lock.
+		// TODO(b/417136852): Remove bucket type check and call only flushFile
+		// when we start leaving zonal bucket objects unfinalized.
+		if !fh.Inode().Bucket().BucketType().Zonal {
+			err = fs.flushFile(ctx, fh.Inode())
+		} else if fh.Inode().Source().IsUnfinalized() {
+			// Flush, but don't finalized for unfinalized objects
```
gargnitingoogle: nit: typo

internal/fs/handle/file.go
gargnitingoogle: @ashmeenkaur @meet2mky I've addressed this comment as discussed offline. PTAL.

internal/fs/fs.go
abhishek10004: nit: finalize

abhishek10004 (approved):

internal/fs/fs.go
```diff
@@ -2587,10 +2587,17 @@ func (fs *fileSystem) ReadFile(
...
-	// Flush Pending streaming writes file for regional bucket and issue read within same inode lock.
-	if fh.Inode().IsUsingBWH() && !fh.Inode().Bucket().BucketType().Zonal {
-		err = fs.flushFile(ctx, fh.Inode())
+	if fh.Inode().IsUsingBWH() {
+		// Flush Pending streaming writes and issue read within same inode lock.
+		// TODO(b/417136852): Remove bucket type check and call only flushFile
+		// when we start leaving zonal bucket objects unfinalized.
+		if !fh.Inode().Bucket().BucketType().Zonal {
+			err = fs.flushFile(ctx, fh.Inode())
+		} else if fh.Inode().Source().IsUnfinalized() {
```
ashmeenkaur: Why not simply use else? If BWH is in use/created, `inode.Source` be unfinalized always.

internal/fs/fs.go

internal/fs/fs.go
gargnitingoogle: fair point.

ashmeenkaur (approved):

PR #3407, opened 2025-06-12
Reduce CPU and memory usage incurred by metrics

kislaykishore:
Reduce metric CPU usage from ~10% for single threaded reads to ~4% and memory allocation from 10% to ~1%.

The `metric.WithAttributes` method is inefficient and is not recommended for performance-sensitive scenarios. This PR moves it to use `metric.WithAttributeSet` method that takes an attribute-set as input and doesn't have those performance concerns. To optimize it further, we use a cache that takes attribute-keys as input, and returns a corresponding attribute.Set object. This reduces memory-allocations by creating an attribute.Set object just once.

There are more improvements that could be performed and will be taken up in the subsequent PRs.

### Description
Before fix: http://shortn/_zNJRaWzRB4
After fix: http://shortn/_nPnS1bb5Vy
b/424403628

Perf:

| Master |  0.25MiB   | 551.34MiB/s  | 1.26MiB/s  |  74.65MiB/s  |  1.17MiB/s   |
|   PR   |  0.25MiB   | 540.25MiB/s  | 1.18MiB/s  |  77.54MiB/s  |  1.21MiB/s   |
| Master | 48.828MiB  | 3630.12MiB/s | 85.51MiB/s | 1530.37MiB/s |  82.58MiB/s  |
|   PR   | 48.828MiB  | 3629.67MiB/s | 85.17MiB/s | 1485.41MiB/s |  85.75MiB/s  |
| Master | 976.562MiB | 3622.02MiB/s | 78.52MiB/s | 764.61MiB/s  |  39.29MiB/s  |
|   PR   | 976.562MiB | 3755.86MiB/s | 77.99MiB/s | 1129.05MiB/s |  38.71MiB/s  |

### Testing details
1. Manual - Verified
2. Unit tests - NA
3. Integration tests - Covered by existing integration tests.

### DIFF
--- a/common/otel_metrics.go
+++ b/common/otel_metrics.go
@@ -26,41 +26,34 @@ import (
 	"go.opentelemetry.io/otel/metric"
 )
 
-const (
-	// ioMethodKey annotates the event that opens or closes a connection or file.
-	ioMethodKey = "io_method"
-
-	// gcsMethodKey annotates the method called in the GCS client library.
-	gcsMethodKey = "gcs_method"
-
-	// fsOpKey annotates the file system op processed.
-	fsOpKey = "fs_op"
-
-	// fsErrCategoryKey reduces the cardinality of FSError by grouping errors together.
-	fsErrCategoryKey = "fs_error_category"
-
-	// readTypeKey annotates the read operation with the type - Sequential/Random
-	readTypeKey = "read_type"
-
-	// cacheHitKey annotates the read operation from file cache with true or false.
-	cacheHitKey = "cache_hit"
-)
-
 var (
 	fsOpsMeter     = otel.Meter("fs_op")
 	gcsMeter       = otel.Meter("gcs")
 	fileCacheMeter = otel.Meter("file_cache")
 
-	fsOpsAttributeSet,
-	readTypeAttributeSet,
-	ioMethodAttributeSet,
-	gcsMethodAttributeSet,
-	cacheHitAttributeSet,
-	cacheHitReadTypeAttributeSet,
-	fsOpsErrorCategoryAttributeSet sync.Map
+	// ioMethodKeyName specifies whether the GCS reader is being opened or closed.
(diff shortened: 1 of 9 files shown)


common/otel_metrics.go
```diff
@@ -25,31 +26,90 @@ import (
...
+	ioMethodOptionCache,
+	gcsMethodOptionCache,
+	cacheHitOptionCache,
+	cacheHitReadTypeOptionCache,
+	fsOpsErrorCategoryOptionCache sync.Map
+)
 
-	// FSOp annotates the file system op processed.
-	FSOp = "fs_op"
+func loadOrStoreAttrOption[K comparable](mp *sync.Map, key K, attrSetGenFunc func() attribute.Set) metric.MeasurementOption {
```
vadlakondaswetha, line 57: why is this method taking a function as argument ? why can't we pass attribute.NewSet as param?

common/otel_metrics.go
```diff
@@ -25,31 +26,90 @@ import (
 	"go.opentelemetry.io/otel/metric"
 )
 
-const (
-	// IOMethod annotates the event that opens or closes a connection or file.
-	IOMethod = "io_method"
+var (
```
vadlakondaswetha, line 29: can you add comment here - what is meter, attribute ad cache

common/otel_metrics.go
kislaykishore: The attribute.NewSet() does a bit of computation as well like sorting, deduplication of attribute keys [implementation](https://github.com/open-telemetry/opentelemetry-go/blob/a85ae98dcedc0761078518a715dea53e519b4846/attribute/set.go#L183).
If I were to pass an attribute.NewSet as a param then that computation would happen every time this function is called and so, it would render the cache useless. My objective here is to do the computation once and not to create new attribute sets allocation for every metric update.

PR #3399, opened 2025-06-09
Update public benchmarks for gcsfuse version v3.0.0

meet2mky:
### Description
Update benchmarking numbers for GCSFuse Version 3.0.0 at [commitID](https://github.com/GoogleCloudPlatform/gcsfuse/tree/a651d01424df33dbcfb0b19dc7cd6827336183e6)

- For each read benchmark metadata cache was prefetched using `ls -R "$MOUNT_POINT"` 
- For each write benchmark we ran it twice to ensure first benchmark creates files and then we delete files to ensure negative stat cache and then results of second write benchmarks were taken.

### Link to the issue in case of a bug fix.
b/415727864

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

PR #3408, opened 2025-06-12
Update benchmark results for version v3.0.0

meet2mky:
### Description
Update benchmarking numbers for GCSFuse Version v3.0.0 at tag `v3.0.0`

- For each read benchmark metadata cache was prefetched using `ls -R "$MOUNT_POINT"` 
- For each write benchmark we ran it twice to ensure first benchmark creates files and then we delete files to ensure negative stat cache and then results of second write benchmarks were taken.
- Converted GiB and MiB to GB and MB for bandwidth calculation.
- Used numjob=48 for 1G write benchmarks instead of numjob=112 (Ref: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3399)
b/415727864

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


docs/benchmarks.md
```diff
@@ -89,87 +89,88 @@ to GCS on close/fsync.
...
-| 100M |1M | 10 | 13.2 | 13.5k | 33 |
-| 200M  | 1M | 10  | 12.4 |  12.7k| 38 |
-| 1G | 1M | 10 | 14.5 | 14.8k  | 60 |
-
-
+| 128K | 128K | 30 | 0.791 | 6.04K | 16.52ms |
+| 256K | 128K | 30 | 1.878 | 14.33K | 7.85ms |
+| 1MB | 1M | 30 | 6.092 | 5.81K | 19.20ms |
+| 5MB | 1M | 20 | 11.166 | 10.65K | 10.84ms |
+| 10MB | 1M | 20 | 12.912 | 12.31K | 10.39ms |
```
charith87: Probably we can keep it to single decimal for the values greater than 1.0 similar to existing. ( e.g. 6.092 to 6.0 12.912 to 12.9 ) Helps in readability.

docs/benchmarks.md
```diff
@@ -89,87 +89,88 @@ to GCS on close/fsync.
...
+| 50MB | 1M | 20 | 19.033 | 18.15K | 12.66ms |
+| 100MB | 1M | 10 | 16.132 | 15.38K | 26.86ms |
+| 200MB | 1M | 10 | 18.945 | 18.07K | 32.89ms |
```
charith87: You can check if we really need ms again since ms is already mentioned in the table header

docs/benchmarks.md
```diff
@@ -89,87 +89,88 @@ to GCS on close/fsync.
...
-| 1M | 1M  | 30  |  683 | 682 | 1.23 |
-| 50M | 1M | 20 | 3429 | 3429 | 2.88 |
-| 100M |1M | 10 | 3519 | 3518 | 11.83 |
-| 1G | 1M | 2 | 1892 | 1891 | 45.40  |
+| 256K | 16K | 30 | 260.902 | 15.92K | 2.80ms |
+| 1M | 1M | 30 | 954.284 | 0.91K | 51.50ms |
+| 50M | 1M | 20 | 3166.878 | 3.02K | 15.49ms |
+| 100M | 1M | 10 | 2835.631 | 2.70K | 21.63ms |
+| 1G | 1M | 2 | 2502.955 | 2.39K | 14.02ms |
+
```
charith87, line 173: Please add if any new steps are added below in steps to benchmark. I see its mentioned you had prefetch etc. good to mention in the steps so customers can redo it as well.

docs/benchmarks.md
```diff
@@ -89,87 +89,88 @@ to GCS on close/fsync.
...
 ### Sequential Writes
-| File Size | BlockSize | nrfiles |Bandwidth in (MiB/sec) | IOPs  |  Avg Latency (msec)  |
+| File Size | BlockSize | nrfiles | Bandwidth in (MB/sec) | IOPs | IOPs Avg Latency (ms) |
 |---|---|---|---|---|---|
-| 256K  | 16K  | 30  | 215 | 13.76k | 0.23 |
-| 1M | 1M  | 30  |  718 | 717 | 1.12 |
-| 50M | 1M | 20 | 3592 | 3592 | 2.35 |
-| 100M |1M | 10 | 4549 | 4549 | 7.04 |
-| 1G | 1M | 2 | 2398 | 2398 | 37.07  |
+| 256K | 16K | 30 | 257.922 | 15.74K | 2.67ms |
```
charith87: Also assuming this bump up is due to GiB to GB conversion for other ones. Here it is due to streaming writes.
Hoping the numbers are validated by @ashmeenkaur or @vadlakondaswetha

docs/benchmarks.md
meet2mky: Thanks did standard rounding upto single digit after decimal.

docs/benchmarks.md
meet2mky: Done, removed.

docs/benchmarks.md
meet2mky: These benchmarks are not reproducible at a single row level so even if mention it there may be variations, We have logged a bug (b/423764522) to fix this in future. Let me know if we need to do for V3.0.0 benchmark update as it was decided in a thread to not do for V3.0.0
CC(folks who are involved in the discussion): @mustvicky @anushka567 @ashmeenkaur

meet2mky: I think, we can add the conditions for the benchmarks though. wdyt? @anushka567

PR #3267, opened 2025-05-06
Add GCSFuse release version in flag parse crash

Tulsishah:
### Description
Add GCSFuse release version in flag parse crash

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

PR #3413, opened 2025-06-13
Print GCSFuse version to stdout stream only in case of error

codechanges:
### Description
The GCSFuse version was being printed to stdout unconditionally [refer](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3267).

This was causing two issues:
* in case of a happy path this message was being printed to stdout and rest of the logs to logs stream (can be stdout) hence giving rise to divergence.
* http://b/424383265 expects that GCSFuse version to be printed on the first line and in a fixed format. 

https://b.corp.google.com/issues/424383265

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - NA


cmd/root.go
```diff
@@ -155,10 +155,10 @@ func convertToPosixArgs(args []string, c *cobra.Command) []string {
 var ExecuteMountCmd = func() {
 	rootCmd, err := newRootCmd(Mount)
 	if err != nil {
-		log.Fatalf("Error occurred while creating the root command: %v", err)
+		log.Fatalf("Error occurred while creating the root command on gcsfuse/%s: %v", common.GetVersion(), err)
 	}
 	rootCmd.SetArgs(convertToPosixArgs(os.Args, rootCmd))
 	if err := rootCmd.Execute(); err != nil {
-		log.Fatalf("Error occurred during command execution: %v", err)
+		log.Fatalf("Error occurred during command execution on gcsfuse/%s: %v", common.GetVersion(), err)
```
Tulsishah, line 162: Are these only two Fatal logs before mounting succeed ? Can you please check once.

codechanges: Yes, it looks fine.

PR #3414, opened 2025-06-13
add resource usage script and add it to e2e test runs to print cpu/mem/disk usage after runs.

meet2mky:
### Description
Add resource usage cpu/mem/disk usage after every e2e test runs. 
#### Output in fusion:
AfzYdSj36iGbixp
4ugpThhP7UY5sYb

b/418902965

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - ran as part of presubmit.


tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -530,9 +536,16 @@ main() {
     fi
     log_info "Script built GCSFuse at: ${BUILT_BY_SCRIPT_GCSFUSE_BUILD_DIR}"
   fi
-    
+
   # Reset SECONDS to 0
   SECONDS=0
+
+  # Start collecting system resource usage in background.
+  log_info "Starting resource usage collection process."
```
codechanges, line 556: can we make it optional to start with?

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -556,7 +569,19 @@ main() {
   elapsed_min=$(((SECONDS + 60) / 60))
   log_info "------ E2E test packages complete run took ${elapsed_min} minutes ------"
   log_info ""
+
+  # Print package runtime stats table.
   ./tools/integration_tests/create_package_runtime_table.sh "$PACKAGE_RUNTIME_STATS"
+
+  # Kill resource usage background PID and print resource usage.
+  log_info "Stopping resource usage collection process: $RESOURCE_USAGE_PID"
+  if kill "$RESOURCE_USAGE_PID"; then
```
codechanges: It is always safe to verify whether PID and process name matches otherwise OS can recycle PID (if resource usage job is already dead) and you would end up killing the wrong process.

tools/integration_tests/resource_usage.sh
```diff
@@ -0,0 +1,228 @@
...
+
+    # Calculate ceiling of (100 * busy_diff / total_diff) using integer arithmetic
+    local cpu_usage=$(((100 * busy_diff + total_diff - 1) / total_diff))
+
+    echo "$cpu_usage"
+}
+
+# Calculates memory usage as a ceiling integer percentage from /proc/meminfo.
+get_mem_usage() {
+    local line mem_total=0 mem_available=0
```
codechanges: can we simplify memory and CPU gathering by using `top` or `ps` command? parsing raw files seems too cumbersome for these tasks?

meet2mky: Thanks updated the code.

tools/integration_tests/improved_run_e2e_tests.sh
meet2mky: This process will always outlive the e2e test runs (3 hrs) so it's fine to kill the process because it won't be dead unless killed in trap or by script itself.

tools/integration_tests/improved_run_e2e_tests.sh
meet2mky: I am keeping this as it for now, I will add one more flag when we do `getopts` otherwise # flags would increase more :)

tools/integration_tests/improved_run_e2e_tests.sh
codechanges: I am talking about cases where the script dies because of some error it faced. In that case, it is possible that PID will be reused.

meet2mky: I see! Thank you! Made the changes for this edge case.

codechanges (approved):

Tulsishah (approved):

PR #3420, opened 2025-06-17
[Migrate auth] Add new auth functions

Tulsishah:
### Description
- Create new function to find credentials from new auth library 

[b/425580693](https://b.corp.google.com/issues/425580693)

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA


internal/auth/credentials.go
```diff
@@ -0,0 +1,56 @@
...
+//
+// This method requests `storagev1.DevstorageFullControlScope` for Google Cloud Storage access.
+//
+// Args:
+// keyFile: Path to a service account key file. Pass an empty string if not used.
+//
+// Returns:
+//
+//	*auth.Credentials: Discovered authentication credentials.
+//	error: An error if credential detection fails.
```
vadlakondaswetha, line 59: looks like the formatting is wrong here. please check once.

internal/auth/credentials.go
```diff
@@ -0,0 +1,56 @@
...
+
+const scope = storage.ScopeFullControl
+
+// GetCredentials detects default Google Cloud credentials.
+//
+// It prioritizes a service account key file if `keyFile` is provided. If `keyFile` is
+// empty, it automatically attempts to detect Application Default Credentials (ADC)
+// and checks the metadata server for credentials.
+//
+// This method requests `storagev1.DevstorageFullControlScope` for Google Cloud Storage access.
```
vadlakondaswetha: why does it requests for this scope?

internal/auth/credentials.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+
+	"cloud.google.com/go/auth"
+	"cloud.google.com/go/auth/credentials"
+	"cloud.google.com/go/storage"
+)
+
+var detectCredentials = credentials.DetectDefault
```
vadlakondaswetha: why is this required? for unit tests?

Tulsishah: Yes for unit tests.

internal/auth/credentials.go
Tulsishah: It seems correct. `go fmt` formats it this way. Although updated little bit.

internal/auth/credentials.go
Tulsishah: The function requests storagev1.DevstorageFullControlScope to ensure the most comprehensive permissions for GCS. This allows subsequent operations using these credentials to have full read, write, and administrative control over GCS resources. 

We've maintained this [scope](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/auth/auth.go#L95) from our older authentication method to avoid breaking any existing scenarios.

vadlakondaswetha: Thanks for the clarification.

vadlakondaswetha (dismissed):

internal/auth/credentials.go
```diff
@@ -0,0 +1,59 @@
...
```
raj-prince: Please avoid using global variable, instead I would suggest to create a private method which take this method.

```go
func getCredentials(keyFile string, detectCredential typeof(credentials.DetectDefault)) (*auth.Credentials, error) {}

func GetCredentials(keyFile string) (*auth.Credentials, error) {
    return getCredentials(keyFile, credentials.DetectDefault)
}
```
And write unit test for private method.

internal/auth/credentials_test.go
```diff
@@ -0,0 +1,108 @@
...
+	if args.Get(0) == nil {
+		return nil, args.Error(1)
+	}
+	return args.Get(0).(*googleAuth.Credentials), args.Error(1)
+}
+
+// //////////////////////////////////////////////////////////////////////
+// Boilerplate
+// //////////////////////////////////////////////////////////////////////
+type credentialsTestSuite struct {
```
raj-prince: Do we need suite? If not then please avoid, as it unnecessarily increases complexity by introducing an additional layer of abstraction.

internal/auth/credentials.go
Tulsishah: It is private variable only so I think it is not harmful, but we can follow your approach as well.

internal/auth/credentials.go
```diff
@@ -0,0 +1,62 @@
...
+//
+//	keyFile: Path to a service account key file. Pass an empty string to use ADC.
+//
+// Returns:
+//
+//	*auth.Credentials: Discovered authentication credentials.
+//	error: An error if credential detection fails.
+func GetCredentials(keyFile string) (*auth.Credentials, error) {
+	creds, err := getCredentials(keyFile, credentials.DetectDefault)
+	if err != nil {
```
raj-prince: Add these checks in the private helper - keep this (public) method just a dispatcher. That's why safe to not write unit test for dispatcher.

raj-prince (approved):

PR #3422, opened 2025-06-18
Generic Queue Implementation for Buffered Read

raj-prince:
### Description
- Implementing generic queue implementation for buffered read.
- Why not use container/list? - extra memory because of doubly linked list.
- Why not implemented using slice - double copy cost in the slice based queue implementation.

<todo>

### Testing details
1. Manual - Yes.
2. Unit tests - Added unit test.
3. Integration tests - NA


common/queue.go
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		q.end.next = n
+		q.end = n
+	}
+	q.size++
+}
+
+// Pop removes and returns the front item from the queue.
+func (q *Queue[T]) Pop() T {
+	if q.size == 0 {
```
vadlakondaswetha, line 84: shouldn't popping from empty queue throw an error?

raj-prince: Yeah, ideally.

But not returning an error as that requires handling everytime, so panicking in this situation. Now, client needs to make sure queue is non-empty, while calling Pop() or Peek().

raj-prince:
> If we know the maximum size of the queue in advance - could we instead use a buffered channel for this purpose? It'd make the queue thread-safe as well.

Primarily due to two reasons, 
(a) We will not get Peek() functionality (at least easily).  Without Peek we need to manage the first block in the buffered reader, will make the implementation a bit complex.
(b) Channel overhead (until thread safety is required).

[Discussed offline] - We will go with the existing queue, as hard to implement Peek() functionality with buffered channel.

common/queue.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// Queue is a generic queue implementation using a linked list.
+type Queue[T any] struct {
+	start, end *node[T]
+	size       int
+}
+
+// NewQueue creates a new empty queue.
+func NewQueue[T any]() *Queue[T] {
```
kislaykishore: There are two ways to instantiate a Queue object currently:
* `common.Queue{}`
* `common.NewQueue()`

Both do the same thing currently, but it's something that can cause confusion as in what's the right way. And, it's also difficult to ensure that both the approaches are always semantically equivalent in the future. So, either remove the NewQueue method altogether or ensure that Queue cannot be initialized by another means except NewQueue as in:

```
type Queue[T any] interface {
	Enqueue(a T)
   ... other methods ...
}

type linkedListBasedQueue[T any] struct {
  ...
}

func (l *linkedListBasedQueue[T]) Enqueue(a T) {
	fmt.Println("Hello")
}

func NewQueue[T any]() Queue[T] {
	return &linkedListBasedQueue[T]{}
}

```

common/queue_test.go
```diff
@@ -0,0 +1,153 @@
...
+	q := NewQueue[int]()
+
+	q.Push(4)
+	q.Push(5)
+
+	assert.Equal(t, 4, q.Peek())
+	assert.False(t, q.IsEmpty())
+}
+
+func TestQueue_Pop(t *testing.T) {
```
kislaykishore: This test could be simplified - it has multiple "Act" blocks for instance. I'd recommend splitting it into two tests.

common/queue_test.go
```diff
@@ -0,0 +1,153 @@
...
+	assert.Equal(t, 5, q.Peek())
+
+	val = q.Pop()
+	assert.Equal(t, 5, val)
+	assert.True(t, q.IsEmpty())
+}
+
+func TestQueue_PopNewQueue(t *testing.T) {
+	q := NewQueue[int]()
+	assert.True(t, q.IsEmpty(), "New queue should be empty.")
```
kislaykishore: The fact that NewQueue is empty should have a test of its own. Let's not have this assert here.

common/queue_test.go
```diff
@@ -0,0 +1,153 @@
...
+		q.Pop()
+	}, "Pop should panic when called on an empty queue.")
+}
+
+func TestQueue_PopEmptyQueue(t *testing.T) {
+	q := NewQueue[int]()
+	q.Push(3)
+	val := q.Pop()
+	assert.Equal(t, 3, val)
+	assert.True(t, q.IsEmpty())
```
kislaykishore: There is a lot of redundancy in these tests. For instance, the fact that 3 should be on top should be tested by a test meant for "Peek" functionality and not this. I'd recommend making similar changes elsewhere.
```suggestion
```

common/queue_test.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func TestQueue_PopNewQueue(t *testing.T) {
+	q := NewQueue[int]()
+	assert.True(t, q.IsEmpty(), "New queue should be empty.")
+
+	assert.Panics(t, func() {
+		q.Pop()
+	}, "Pop should panic when called on an empty queue.")
+}
```
kislaykishore:
```suggestion
func TestQueue_PopEmptyQueue(t *testing.T) {
	assert.Panics(t, func() {
		NewQueue[int]().Pop()
	}, "Pop should panic when called on an empty queue.")
}
```

common/queue_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+func TestQueue_Peek(t *testing.T) {
+	q := NewQueue[int]()
+	q.Push(4)
+
+	val := q.Peek()
+	assert.Equal(t, 4, val)
+
+	// Peek again to ensure it doesn't remove the element.
```
kislaykishore: This comment is what this test is about I guess - the test name IMO should be TestQueuePeekIdempotent. I would recommend another test to verify that q.Peek() returns 4. Alternatively, don't test the idempotency and just end the test at `assert.Equal(t, 4, val)`

common/queue_test.go
raj-prince: Changing assert -> require.

common/queue_test.go
raj-prince: Instead of invoking the Peek() again to validate no side change, asserted over length - before and after length should be unchanged.

vadlakondaswetha (dismissed):

vadlakondaswetha (approved):

common/queue_test.go
kislaykishore: Please refer to these guidelines: https://screenshot.googleplex.com/A42iSeX3o8MRhGR

raj-prince: Thanks, Got, it!
Not required here, as deleted this test case.

PR #3423, opened 2025-06-18
[Buffered reader] Implement static workerpool

raj-prince:
### Description
- Added WorkerPool and Task Interface
- Implementation of StaticWorkerPool  & Unit test

b/426060537

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/workerpool/static_worker_pool.go
+++ b/internal/workerpool/static_worker_pool.go
@@ -15,7 +15,6 @@
 package workerpool
 
 import (
-	"context"
 	"fmt"
 	"sync"
 
@@ -30,9 +29,8 @@ type staticWorkerPool struct {
 	priorityWorker uint32 // Number of priority workers in this pool.
 	normalWorker   uint32 // Number of normal workers in this pool.
 
-	// Context to close all the workers.
-	ctx    context.Context
-	cancel context.CancelFunc
+	// Stop channel to notify all the workers to stop.
+	stop chan bool
 
 	// Wait group to wait for all workers to finish.
 	wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


internal/workerpool/static_worker_pool.go
```diff
@@ -0,0 +1,132 @@
...
+	if totalWorkers == 0 {
+		return nil, fmt.Errorf("staticWorkerPool: can't create with 0 workers, priority: %d, normal: %d", priorityWorker, normalWorker)
+	}
+
+	logger.Infof("staticWorkerPool: creating with %d normal, and %d priority workers.", normalWorker, priorityWorker)
+
+	return &staticWorkerPool{
+		priorityWorker: priorityWorker,
+		normalWorker:   normalWorker,
+		close:          make(chan int, totalWorkers),
```
vadlakondaswetha: can we have a non buffered channel for signaling close for all goroutines?
I meant something like this:
stop() {
 close(swp.close)
 close(swp.priorityCh)
 close(swp.normalCh)
  
  swp.wg.Wait()
}

internal/workerpool/static_worker_pool.go
```diff
@@ -0,0 +1,141 @@
...
+	if priority {
+		// Worker only listens to the priority channel.
+		for {
+			select {
+			case <-swp.ctx.Done():
+				return
+			default:
+				select {
+				case <-swp.ctx.Done():
+					return
```
vadlakondaswetha, line 114: why did you add default case with swp.ctx.Done(). cant you simply have this:

select {
 case <- swp.ctx.Done():
    return
case task := <-swp.priorityCh;
  task.execute

}

raj-prince: `select` statement chooses pseudo randomly if multiple channels are available for read. 

Adding default with swp.ctx.Done() will ensure priority of cancellation if there is some task in the task channel.

vadlakondaswetha (approved):

PR #3424, opened 2025-06-19
Unit tests for Appends to unfinalized object using BW

anushka567:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/fs/inode/file_test.go
```diff
@@ -803,6 +805,40 @@ func (t *FileTest) TestAppendThenSync() {
...
+	assert.Nil(t.T(), err)
+	object.Finalized = time.Time{}
+	t.backingObj = storageutil.ConvertObjToMinObject(object)
+	t.createInode()
+	t.in.config = &cfg.Config{Write: *getWriteConfigWithEnabledRapidAppends()}
+	assert.Nil(t.T(), t.in.content)
+	t.createBufferedWriteHandler(true, AppendOpenMode)
+	assert.NotNil(t.T(), t.in.bwh)
+
+	gcsSynced, err := t.in.Write(t.ctx, []byte("juice"), int64(len("lychee")), AppendOpenMode)
```
vadlakondaswetha: you can use t.initialcontents here instead of len("lychee")

internal/fs/inode/file_test.go
```diff
@@ -58,6 +58,8 @@ const fileMode os.FileMode = 0641
 const Delta = 30 * time.Minute
 const LocalFile = "Local"
 const EmptyGCSFile = "EmptyGCS"
+const WriteOpenMode = util.Write
```
vadlakondaswetha: why are these changes required?

internal/storage/fake/bucket.go
```diff
@@ -392,18 +392,32 @@ func preconditionChecks(b *bucket, req *gcs.CreateObjectRequest, contents []byte
...
+		var content []byte
+		if b.bucketType.Zonal {
+			// If the bucket type is zonal, then we will update the fake object with the appended content.
+			b.mu.Unlock()
+			existingContent, err := storageutil.ReadObject(context.Background(), b, req.Name)
+			b.mu.Lock()
+			if err != nil {
+				return nil, fmt.Errorf("error while reading existing content in fake object : %v", err)
+			}
+			content = append(existingContent, contents...)
```
vadlakondaswetha, line 408: this becomes tricky, how will you test the overwrite scenarios of finalized objects?

internal/fs/inode/file_test.go
anushka567: Not required since util.Write is a constant anyways, Reverted for both Append and Write.

internal/storage/fake/bucket.go
anushka567: For overwrite scenarios, we'll have to modify the flows for fakebucket implementation to differentiate between write to new/ overwrite and append scenarios for the flow itself.

We'll have to split this method into createFakeObject and UpdateFakeObject (with a flag to denote append or overwrite)

vadlakondaswetha (approved):

PR #3426, opened 2025-06-19
Add Recommended Sample GCSFuse config YAML for different workflows

codechanges:
### Description
We already provide recommended samples for GKE. On the similar lines, this PR will add samples for GCSFuse config in general which can be used directly with GCSFuse or any infrastructure provisioning tools.

https://buganizer.corp.google.com/issues/423797260

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - NA


samples/config/gpu/config_file/checkpointing.yaml
charith87, line 1: Consider moving this to a directory structure like "samples/gce_gcsfuse_config" and then tpu and gpu.

There is no need to again a config folder underneath.

In future, we can have a "samples/gce_cli_config" and then tpu and gpu.

samples/config/gpu/config_file/checkpointing.yaml
```diff
@@ -0,0 +1,13 @@
+implicit-dirs: true
```
charith87: Consider creating a README file with these folders which can be used when ppl look at these samples e.g. https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/samples/gke-csi-yaml/gpu/README.md

samples/config/gpu/config_file/checkpointing.yaml
codechanges: Thanks but it can be used beyond GCE like a private machine or SLURM cluster so we may not want to use gce nomenclature.

samples/config/gpu/config_file/checkpointing.yaml
```diff
@@ -0,0 +1,13 @@
...
+cache-dir: /tmp # set to appropriate directory
+metadata-cache:
+  negative-ttl-secs: 0
+  ttl-secs: -1 # change as per workload needs
+  stat-cache-max-size-mb: -1
```
charith87: We could add comments after it so it serves as a short version of the choice that they are making.
Similar things were done for the yamls . For e.g. 0 and -1 are not intuitive at a glance for new customers. https://github.com/GoogleCloudPlatform/gcsfuse/blob/0cb3f5745a38b41157bd51cc3f7ce72f214ec68e/samples/gke-csi-yaml/gpu/checkpointing-pv.yaml#L17

samples/config/gpu/config_file/training.yaml
```diff
@@ -0,0 +1,11 @@
+implicit-dirs: true
+metadata-cache:
+  negative-ttl-secs: 0
+  ttl-secs: -1 # change as per workload needs
+  stat-cache-max-size-mb: -1
+  type-cache-max-size-mb: -1
+# if enabling the file cache, uncomment out to use #
+# cache-dir: /tmp
```
charith87: For TPU, you could mention its supposed to be a RAMDisk entry.  /tmp may very well be the local PD which needs a change from the customer.

For creation of a RAM Disk, you may want to add that either here or in the README.

samples/config/tpu/config_file/training.yaml
```diff
@@ -0,0 +1,11 @@
+implicit-dirs: true
```
charith87: You may also want to add the ls -R part in the README. which is currently there in gke yaml but not there in gce yaml. to do  ls -R > /dev/null . The info should be there in perf cloud docs. We could also point to it.

samples/config/gpu/config_file/checkpointing.yaml
charith87: probably signifying its "gcsfuse_config" will help instead of just config.
The Readme could have details about SLURM cluster etc.

samples/config/gpu/config_file/checkpointing.yaml
codechanges: Good suggestion, thanks.

samples/config/gpu/config_file/checkpointing.yaml
codechanges: Agreed, will make the changes.

samples/config/gpu/config_file/training.yaml
codechanges: Added it in the config file itself for easy discover-ability

samples/config/tpu/config_file/training.yaml
codechanges: Added the details.

samples/config/gpu/config_file/checkpointing.yaml
codechanges: Added README, ptal

samples/config/gpu/config_file/checkpointing.yaml
codechanges: I have added README which alludes to SLURM and GCE but intentionally kept it simple to maintain generic nature of the document and samples.

samples/gcsfuse_config/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Here is an [example](https://github.com/GoogleCloudPlatform/cluster-toolkit/blob/51c51f2c83383a8f241cd0ef8a8998413393bff5/examples/hypercompute_clusters/a3u-slurm-ubuntu-gcs/a3u-slurm-ubuntu-gcs.yaml#L193) of using config with Ansible.
+
+## **Key Considerations**
+
+* **File Cache:**
+
+  * For GPU-based workloads, it is highly recommended to use a Local SSD (LSSD) for the file cache directory for optimal performance.  
+  * For TPU-based workloads, using a RAM disk for the file cache can provide a significant speed advantage.  
+* **Metadata Cache TTL:**
```
charith87, line 46: The metadata cache ttl is not needed since we are already adding the field in the configs with infinite ttl. We can keep to the things which manually the customers must minimal steps for a good default behavior. Which involves the ls -R

samples/gcsfuse_config/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+  * The sample configurations may use a long or infinite Time-to-Live (TTL) for the metadata cache. While this can improve performance for static datasets, be aware of the implications. If the data in your GCS bucket changes frequently, you may need to adjust the `ttl-secs` setting in the configuration file to avoid stale metadata.  
+* **Pre-populating Metadata Cache:**
+
+  * To improve the performance of subsequent file and directory lookups, you can pre-populate the metadata cache after mounting the bucket by running the following command. More details can be found [here](https://cloud.google.com/storage/docs/cloud-storage-fuse/performance#pre-populate-the-metadata-cache):
+
+```
+      ls -R <mount-point> > /dev/null
```
charith87, line 54: Pls check if there are other features which are missing and add them for GKE to GCE translation since i also see the read_ahead_kb missing.

samples/gcsfuse_config/README.md
codechanges: This is needed from SLURM perspective. Those mounts generally have similar lifetime as that of node and infinite timeout can be detrimental to them. I got this comment from the respective team itself and that's why I chose to spell it out here.

samples/gcsfuse_config/README.md
codechanges: Ok, `read_ahead_kb` is not supported directly hence I did not add it. I have included that in readme now. I have taken care of all flags already and wanted to avoid GKE specific stuff like `ls -R` and `read_ahead_kb` but adding them in readme looks ok.

samples/gcsfuse_config/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+* **Tuning `read-ahead-kb`** 
+
+  * The `read-ahead-kb` flag controls the size of the read-ahead buffer that GCSFuse uses to prefetch data from GCS. Tuning this value can significantly improve performance for sequential reads of large files.
```
charith87: We might need to change this. Read_ahead controls the kernel request from fuse. Reduces the number of requests from application to gcsfuse.

samples/gcsfuse_config/README.md
charith87: Yes, Okk. But we already have it in the config with the comment. So was thinking the use of it here again if its repeating itself. 
I am guessing this is for High End machines not small machines.
Also, we need to change the text data in your bucket changes frequently can be misleading a bit. Its bad only for overwriting files and out of mount files deletions, not for creation of new files.So not strictly static datasets ( write to new is also fine). Most of the reasons are written here as well https://cloud.google.com/storage/docs/cloud-storage-fuse/performance#increase-metadata-cache-values

samples/gcsfuse_config/README.md
codechanges: Thanks, good catch. Let me rephrase it.

samples/gcsfuse_config/README.md
codechanges: Thanks, let me rephrase it. The comment in the config is more about what it is doing and section is here is about implication of the same. I will point to the link you mentioned with lite text.

samples/gcsfuse_config/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## **Key Considerations**
+
+* **File Cache:**
+
+  * For GPU-based workloads, it is highly recommended to use a Local SSD (LSSD) for the file cache directory for optimal performance.  
+  * For TPU-based workloads, using a RAM disk for the file cache can provide a significant speed advantage.  
+* **Metadata Cache TTL:**
+
+  * The sample configurations may use a long or infinite Time-to-Live (TTL) for the metadata cache. While this can improve performance for static datasets, be aware of the implications. More details can be found [here](https://cloud.google.com/storage/docs/cloud-storage-fuse/performance#increase-metadata-cache-values).
```
charith87: I would still remove the static datasets since write to new modes also works.

charith87 (approved):
Overall fine with a minor suggested edit.

samples/gcsfuse_config/README.md
codechanges: ok

PR #3428, opened 2025-06-19
[Random Reader Refactoring] Increase totalBytes for existing reader

Tulsishah:
### Description
- Inc totalBytes in case reading from existing reader as well.

[b/426166667](https://b.corp.google.com/issues/426166667)

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - Done
4. Perf tests

Performance tests were conducted by averaging results from multiple runs.
  i. Scenario 1: 200 MiB file, 1 thread, 1 MiB block size
   - Without this fix: 180 MiB/s
   - With this fix: 250 MiB/s (same as master)

  ii. Scenario 2: 200 MiB file, 1 thread, 4 KiB block size
   - Without this fix: 20 MiB/s
   - With this fix: 50 MiB/s (same as master)

### DIFF
--- a/internal/gcsx/client_readers/gcs_reader.go
+++ b/internal/gcsx/client_readers/gcs_reader.go
@@ -116,6 +116,7 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx.R
 	}
 	defer func() {
 		gr.updateExpectedOffset(offset + int64(readerResponse.Size))
+		gr.totalReadBytes += uint64(readerResponse.Size)
 	}()
 
 	var err error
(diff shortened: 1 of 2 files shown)


internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -116,6 +116,7 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx.R
 	}
 	defer func() {
 		gr.updateExpectedOffset(offset + int64(readerResponse.Size))
+		gr.totalReadBytes += uint64(readerResponse.Size)
```
raj-prince, line 119: Good find. 

Is it possible to write an e2e test which basically validates number of GCS calls for a sequential read?

Tulsishah: I'll handle that as a separate task(will a create bug), as it seems to require significant effort. Until then, we can consider an FIO test with a 200MiB file, using both 1MiB and 4KiB block sizes, a good verification test.

raj-prince (approved):

PR #3429, opened 2025-06-19
Changing Block Interface to Follow io.Writer Interface

raj-prince:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/block/block_pool_test.go
```diff
@@ -82,7 +82,9 @@ func (t *BlockPoolTest) TestGetWhenBlockIsAvailableForReuse() {
 	b, err := createBlock(2)
 	require.Nil(t.T(), err)
 	content := []byte("hi")
-	err = b.Write(content)
+	var n int
+	n, err = b.Write(content)
```
kislaykishore: Nit: 
```suggestion
	n, err := b.Write(content)
```

Ditto elsewhere

internal/block/block_test.go
```diff
@@ -50,19 +52,24 @@ func (testSuite *MemoryBlockTest) TestMemoryBlockWriteWithDataGreaterThanCapacit
...
 func (testSuite *MemoryBlockTest) TestMemoryBlockWriteWithMultipleWrites() {
 	mb, err := createBlock(12)
 	require.Nil(testSuite.T(), err)
-	err = mb.Write([]byte("hi"))
+	var n int
+	n, err = mb.Write([]byte("hi"))
 	assert.Nil(testSuite.T(), err)
-	err = mb.Write([]byte("hello"))
+	assert.Equal(testSuite.T(), 2, n)
+	n, err = mb.Write([]byte("hello"))
```
kislaykishore: I think everything in the "Arrange" block should be `require` and not `assert`. The functionality is not really being tested in "Arrange" so any precondition error should be fatal for the test run which is achieved via `require`.

PR #3431, opened 2025-06-20
Removing files created on disk after the test

abhishek10004:
### Description
Currently in TestReadFilesConcurrently, we don't delete the local files created on disk after the test is finished. Making changes to remove the local files after the test. 

b/422632324

### Testing details
1. Manual - Manually ran the test package and tests are passing
2. Unit tests - NA
3. Integration tests - NA

No


meet2mky (approved):
LGTM

meet2mky:
You can attach the release bug as this issue was discovered during that analysis or create a separate bug if you like.

Tulsishah (approved):

PR #3435, opened 2025-06-23
Prefetch Task for WorkerPool

raj-prince:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/block/prefetch_block.go
```diff
@@ -0,0 +1,169 @@
...
+	// - BlockStatusDownloadFailed: Download of this block has failed.
+	// - BlockStatusDownloadCancelled: Download of this block has been cancelled.
+	NotifyReady(val int)
+
+	// GetAbsStartOff returns the absolute start offset of the block.
+	GetAbsStartOff() int64
+
+	// SetAbsStartOff sets the absolute start offset of the block.
+	// This should be called only once just after getting the block from the pool.
+	SetAbsStartOff(startOff int64)
```
raj-prince, line 64: Ideally, this should be set while getting the block from the block-pool and should be read only forever.

raj-prince:
Implemented.

PR #3438, opened 2025-06-23
Add new flag --enable-readdirplus for enabling readdirplus

aditimittal2003:
### Description
Add new flag --enable-readdirplus for enabling readdirplus

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


cfg/params.yaml
```diff
@@ -183,6 +183,13 @@
   default: false
   hide-flag: true
 
+- config-path: "file-system.enable-readdirplus"
+  flag-name: "enable-readdirplus"
+  type: "bool"
+  usage: "This flag is used to enable ReadDirPlus capability"
```
kislaykishore:
```suggestion
  usage: "Enables ReadDirPlus capability"
```

cfg/params.yaml
```diff
@@ -183,6 +183,13 @@
   default: false
   hide-flag: true
 
+- config-path: "file-system.enable-readdirplus"
```
kislaykishore: Double check if this flag should be in file-system or in list.

ashmeenkaur: I think file-system is okay because list only has `enable-empty-managed-folders` config right now which is hidden

cfg/config.go
```diff
@@ -112,6 +112,8 @@ type FileSystemConfig struct {
 
 	DisableParallelDirops bool `yaml:"disable-parallel-dirops"`
 
+	EnableReaddirplus bool `yaml:"enable-readdirplus"`
```
ashmeenkaur: can we add experimental prefix here per https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/dev_guide.md

cfg/params.yaml
```diff
@@ -183,6 +183,13 @@
   default: false
   hide-flag: true
 
+- config-path: "file-system.experimental-enable-readdirplus"
+  flag-name: "experimental-enable-readdirplus"
+  type: "bool"
+  usage: "Experimental: Enables ReadDirPlus capability"
```
ashmeenkaur:
```suggestion
  usage: "Enables ReadDirPlus capability"
```

ashmeenkaur (dismissed):

kislaykishore (approved):

PR #3446, opened 2025-06-25
Add new experimental flag --experimental-enable-dentry-cache to file-system config

aditimittal2003:
### Description
Add a new experimental flag --experimental-enable-dentry-cache to file-system config . The default value of this flag is false.
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


cfg/params.yaml
```diff
@@ -190,6 +190,13 @@
   default: false
   hide-flag: true
 
+- config-path: "file-system.experimental-enable-entry-timeout"
+  flag-name: "experimental-enable-entry-timeout"
+  type: "bool"
+  usage: "Enables timeout-controlled caching of entries in the name lookup cache."
```
kislaykishore: Sorry, never heard of this "name-lookup" cache. What's that?

aditimittal2003: By name lookup, I meant looking up a name (eg: file name). Eg: Stat filename.txt
And entry-timeout determines whether to use the cached entry for the name lookup (if within the specified timeout window) or make a LookUpInode call.

raj-prince: We can change usage doc, something like,
```
When enabled, it sets the Dentry cache entry timeout same as metadata-cache-ttl. This enables kernel to use cached entry to map the file paths to inodes, instead of making LookUpInode calls to GCSFuse.
```

Also name we can change something - `experimental-enable-dentry-cache`

PR #3448, opened 2025-06-25
fix: remove extra go routine in read flow when interrupts are disabled

ashmeenkaur:
### Description
In the read flow when we read data from range reader into the buffer, GCSFuse creates a goroutine which takes care of checking for context cancelled while waiting for data from network. Since interrupts are disabled by default, there is no need of creating this goroutine always for every read request. Create this go routine only if user wants interrupts to be handled.

b/425859120

### Testing details
1. Manual - 

Manual profiline results:
Before optimization:
7FGsbXhaQGDZNPC

After optimization:
5wowPCA4LAvNGNz

Bandwidth for 50G file
1. 4K IO: 
- Before the fix: 67.4MB/s
- After the fix: 131MB/s (~2x improvement)
2. 1M IO:
- Before the fix: 765MB/s
- After the fix: 779MB/s (no significant improvement)

No significant improvement with 1M IO ( <3%)

2. Unit tests - Updated
3. Integration tests - NA

NA

### DIFF
--- a/internal/gcsx/client_readers/gcs_reader_test.go
+++ b/internal/gcsx/client_readers/gcs_reader_test.go
@@ -351,7 +351,7 @@ func (t *gcsReaderTest) Test_ReadAt_ValidateReadType() {
 
 func (t *gcsReaderTest) Test_ReadAt_PropagatesCancellation() {
 	t.gcsReader = NewGCSReader(t.object, t.mockBucket, &GCSReaderConfig{
-		MetricHandle:         common.NewNoopMetrics(),
+		MetricHandle:         metrics.NewNoopMetrics(),
 		MrdWrapper:           nil,
 		SequentialReadSizeMb: sequentialReadSizeInMb,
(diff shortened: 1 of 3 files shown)


vadlakondaswetha (dismissed):

internal/gcsx/random_reader.go
```diff
@@ -427,28 +427,30 @@ func (rr *randomReader) Destroy() {
...
-	readDone := make(chan struct{})
-	defer close(readDone)
-
-	go func() {
-		select {
-		case <-readDone:
-			return
-
-		case <-ctx.Done():
+	if rr.config != nil && !rr.config.FileSystem.IgnoreInterrupts {
```
vadlakondaswetha, line 430: do we need to make same changes in new readers added as part of ReadManager?

ashmeenkaur: Good catch! thanks!

PR #3458, opened 2025-06-27
update pv, pod sample specs for metadata prefetch mem overrides

saikat-royc:
### Description
The standard samples for training/serving/checkpoints use the `gcsfuseMetadataPrefetchOnMount` csi volumeAttributes. Following the standard practice of gke-gcsfuse-sidecar, we are providing guidance for setting unlimited memory and cpu for the `gke-gcsfuse-metadata-prefetch` sidecar container to avoid cpu throttle and oom kils.

We are also internally working to make this defaults in the CSI driver, however until that is available, the guidance to set unlimited resources is valid.

### Testing details
1. Manual - a) Tested manually in clusters which support metadata prefetch gke container b) also tested in a cluster version `1.32.2-gke.1297002` which supports gke metadata prefetch (but does not support the cotnainer resource override annotations), to ensure it is safe to simply set the annotations. The annotation would simply be ignored in such clusters.

3. Unit tests - NA
4. Integration tests - NA

The annotations override only take effect in gke clusters `1.32.3-gke.1717000` or later (see https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-sidecar#configure-sidecar-resources)

### DIFF
--- a/samples/gke-csi-yaml/gpu/checkpointing-pod.yaml
+++ b/samples/gke-csi-yaml/gpu/checkpointing-pod.yaml
@@ -5,7 +5,8 @@ metadata:
   namespace: <customer-namespace>
   annotations:
     gke-gcsfuse/volumes: "true"
-
+    # gke-gcsfuse/metadata-prefetch-memory-limit: "0" # min GKE version: `1.32.3-gke.1717000` for this annotation to take effect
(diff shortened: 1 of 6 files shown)


saikat-royc:
cc @charith87

charith87 (approved):

PR #3469, opened 2025-07-03
Add ReadEntryCores method to DirInode interface

aditimittal2003:
### Description
Add the ReadEntryCores method to the DirInode interface and its implementations. This method allows for reading a batch of directory entries and returning them as a map of inode.Core objects.

This change is a foundational step for the future implementation of the ReadDirPlus FUSE operation.

### Testing details
1. Manual - NA
2. Unit tests - Added unit tests in base_dir_test.go and dir_test.go
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/base_dir.go
+++ b/internal/fs/inode/base_dir.go
@@ -191,6 +191,18 @@ func (d *baseDirInode) ReadEntries(
 	return nil, "", syscall.ENOTSUP
 }
 
+// LOCKS_REQUIRED(d)
+func (d *baseDirInode) ReadEntryCores(
+	ctx context.Context,
+	tok string) (cores map[Name]*Core, newTok string, err error) {
+
+	// The subdirectories of the base directory should be all the accessible
+	// buckets. Although the user is allowed to visit each individual
+	// subdirectory, listing all the subdirectories (i.e. the buckets) can be
+	// very expensive and currently not supported.
+	return nil, "", syscall.ENOTSUP
+}
+
 ////////////////////////////////////////////////////////////////////////
 // Forbidden Public interface
 ////////////////////////////////////////////////////////////////////////
--- a/internal/fs/inode/base_dir.go
+++ b/internal/fs/inode/base_dir.go
@@ -192,9 +192,7 @@ func (d *baseDirInode) ReadEntries(
 }
 
 // LOCKS_REQUIRED(d)
-func (d *baseDirInode) ReadEntryCores(
-	ctx context.Context,
-	tok string) (cores map[Name]*Core, newTok string, err error) {
+func (d *baseDirInode) ReadEntryCores(ctx context.Context, tok string) (cores map[Name]*Core, newTok string, err error) {
 
 	// The subdirectories of the base directory should be all the accessible
 	// buckets. Although the user is allowed to visit each individual
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -100,9 +100,7 @@ type DirInode interface {
 	// empty. Otherwise it will be non-empty. There is no guarantee about the
 	// number of entries returned; it may be zero even with a non-empty
 	// continuation token.
-	ReadEntryCores(
-		ctx context.Context,
-		tok string) (cores map[Name]*Core, newTok string, err error)
+	ReadEntryCores(ctx context.Context, tok string) (cores map[Name]*Core, newTok string, err error)
 
 	// Create an empty child file with the supplied (relative) name, failing with
 	// *gcs.PreconditionError if a backing object already exists in GCS.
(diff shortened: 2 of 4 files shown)


internal/fs/inode/dir.go
```diff
@@ -91,6 +91,19 @@ type DirInode interface {
...
+	// used to pick up the read operation where it left off.
+	// Supply the empty token on the first call.
+	//
+	// At the end of the directory, the returned continuation token will be
+	// empty. Otherwise it will be non-empty. There is no guarantee about the
+	// number of entries returned; it may be zero even with a non-empty
+	// continuation token.
+	ReadEntryCores(
+		ctx context.Context,
+		tok string) (cores map[Name]*Core, newTok string, err error)
```
ashmeenkaur: Per [go/go-style/decisions#func-formatting](go/go-style/decisions#func-formatting) (externally - https://google.github.io/styleguide/go/decisions#function-formatting):
```suggestion
	ReadEntryCores(ctx context.Context, tok string) (cores map[Name]*Core, newTok string, err error)
```

Follow at other places too

internal/fs/inode/dir.go
```diff
@@ -801,6 +814,20 @@ func (d *dirInode) ReadEntries(
 	return
 }
 
+func (d *dirInode) ReadEntryCores(
```
ashmeenkaur: Shall we re-use this function in `ReadEntries` method above?

internal/fs/inode/base_dir.go
```diff
@@ -191,6 +191,18 @@ func (d *baseDirInode) ReadEntries(
 	return nil, "", syscall.ENOTSUP
 }
 
+// LOCKS_REQUIRED(d)
+func (d *baseDirInode) ReadEntryCores(
```
ashmeenkaur: Same here as well, shall we re-use it in the above ReadEntries method?

internal/fs/inode/dir_test.go
```diff
@@ -177,6 +177,28 @@ func (t *DirTest) readAllEntries() (entries []fuseutil.Dirent, err error) {
 	return
 }
 
+// Read all of the entry cores
+func (t *DirTest) readAllEntryCores() (cores map[Name]*Core, err error) {
+	cores = make(map[Name]*Core)
+	tok := ""
+	for {
+		var tmp map[Name]*Core
```
ashmeenkaur: Nit: 
```suggestion
		var fetchedCores map[Name]*Core
```

internal/fs/inode/dir_test.go
```diff
@@ -177,6 +177,28 @@ func (t *DirTest) readAllEntries() (entries []fuseutil.Dirent, err error) {
...
+		tmp, tok, err = t.in.ReadEntryCores(t.ctx, tok)
+		for name, core := range tmp {
```
ashmeenkaur: maps.Copy(cores, tmp) could also have been used

internal/fs/inode/dir_test.go
```diff
@@ -177,6 +177,28 @@ func (t *DirTest) readAllEntries() (entries []fuseutil.Dirent, err error) {
...
+			cores[name] = core
+		}
+		if err != nil {
```
ashmeenkaur, line 188: any reason for not checking the error first before copying? I think the idiomatic way is to check and return the error first before processing other outputs. [go/go-style/decisions#handle-errors](go/go-style/decisions#handle-errors)

internal/fs/inode/base_dir.go
aditimittal2003: We can reuse it in ReadEntries method in dir.go but resusing here doesn't make sense because both the methods here are currently just placeholders that return an error, there is no logic to reuse yet.

internal/fs/inode/dir_test.go
```diff
@@ -960,6 +982,181 @@ func (t *DirTest) ReadEntries_TypeCaching() {
...
+	ExpectEq(metadata.RegularFileType, core.Type())
+	ExpectEq(metadata.RegularFileType, t.getTypeFromCache("file"))
+
+	// symlink
+	name = NewFileName(t.in.Name(), "symlink")
+	core, ok = cores[name]
+	AssertTrue(ok, "entry for symlink not found")
+	ExpectEq(symlinkName, core.FullName.GcsObjectName())
+	ExpectEq(metadata.SymlinkType, core.Type())
+	ExpectEq(metadata.SymlinkType, t.getTypeFromCache("symlink"))
```
ashmeenkaur: do you want to create a helper function for this instead of repeating for each core?

internal/fs/inode/dir_test.go
```diff
@@ -216,6 +237,21 @@ func (t *DirTest) getLocalDirentKey(in Inode) string {
 	return path.Base(in.Name().LocalName())
 }
 
+func (t *DirTest) validateCore(cores map[Name]*Core, entryName string, isDir bool, expectedType metadata.Type, expectedFullName string) {
+	var name Name
```
ashmeenkaur, line 241: Nit: add t.Helper()
(you can do it in the follow up PR)

PR #3473, opened 2025-07-04
ci(read unfinalized objects): Add new e2e tests for reads of unfinalized objects while being appended

meet2mky:
### Description
- Adding basic functional tests of unfinalized objects for reading while being appended. 
- This is the first step in the new tests to be added for [test plan doc](https://docs.google.com/document/d/1f41hYxHdSfR7eM_EftkA2ajq0p7aV-vze-ALicYHa9Y/edit?usp=sharing)
- Additional Fix: CreateUnfinalizedObject to accept content instead of size for creating unfinalized Objects.

#### Not done here
- Test configurations e.g. metadata cache on/off, file cache on/off, streaming-write on/off, sequential/random writes - will be covered in future PRs start with #3539 

#### Future improvements pending
- Make the code and tests for multi-mount scenarios cleaner
- Merge this package with unfinalized_object package to keep a single uniform package.

### Link to the issue in case of a bug fix
b/429498114

~~Note: These tests would pass when this PR is rebased on top of #3405~~  (Rebased)

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - Passed e2e ZB as part of [presubmit](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/af5ae340-ab70-4670-83c7-3482f6a993f2/summary) .


tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -221,7 +221,7 @@ TEST_PACKAGES_COMMON=(
 # Test packages for regional buckets.
 TEST_PACKAGES_FOR_RB=("${TEST_PACKAGES_COMMON[@]}" "inactive_stream_timeout" "cloud_profiler")
 # Test packages for zonal buckets.
-TEST_PACKAGES_FOR_ZB=("${TEST_PACKAGES_COMMON[@]}" "unfinalized_object")
+TEST_PACKAGES_FOR_ZB=("${TEST_PACKAGES_COMMON[@]}" "unfinalized_object" "unfinalized_appends")
```
gargnitingoogle: Let's rename this package to rapid_appends.

tools/integration_tests/unfinalized_appends/setup_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"testing"
+
+	"cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/client"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting/static_mounting"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup"
+)
+
+const (
+	testDirName = "AppendFileTest"
```
gargnitingoogle:
```suggestion
	testDirName = "RapidAppendsTest"
```

tools/integration_tests/unfinalized_appends/setup_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+var (
+	// Flags for the mount options for gRootDir and gOtherRootDir
+	gFlags []string
```
gargnitingoogle: nit: Why do we have `g` prefix in every variable?

tools/integration_tests/unfinalized_appends/setup_test.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}()
+
+	// Set up test directory for secondary mount.
+	setup.SetUpTestDirForTestBucketFlag()
+	gOtherRootDir = setup.MntDir()
+	gOtherLogFilePath = setup.LogFile()
+	// For reads to work for unfinalized object from secondary mount metadata cache ttl must be set to 0.
+	// and rapid appends should be enabled.
+	secondarMountFlags := []string{"--metadata-cache-ttl-secs=0", "--write-experimental-enable-rapid-appends=true"}
```
gargnitingoogle:
```suggestion
	secondaryMountFlags := []string{"--metadata-cache-ttl-secs=0", "--write-experimental-enable-rapid-appends=true"}
```

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+	t.fileSize = SizeOfFileContents
+
+	t.appendFileHandle, err = os.OpenFile(path.Join(gTestDirPath, t.fileName), os.O_APPEND|os.O_WRONLY|syscall.O_DIRECT, operations.FilePermission_0600)
+	require.NoError(t.T(), err)
+	initialContent, err := operations.ReadFile(path.Join(gTestDirPath, t.fileName))
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), SizeOfFileContents, len(initialContent))
+	t.initialContent = string(initialContent)
+}
+
```
anushka567: Add TearDownTest() to close the writer and file handle.

tools/integration_tests/unfinalized_appends/setup_test.go
```diff
@@ -0,0 +1,123 @@
...
+		gTestDirPath = setup.SetupTestDirectory(testDirName)
+		setup.RunTestsForMountedDirectoryFlag(m)
+	}
+
+	// Set up test directory for primary mount.
+	setup.SetUpTestDirForTestBucketFlag()
+
+	// Define flag set to run the tests.
+	flagsSet := [][]string{
+		{"--metadata-cache-ttl-secs=0", "--write-experimental-enable-rapid-appends=true"},
```
gargnitingoogle: Let's add more configurations here. "--write-experimental-enable-rapid-appends=true" will be common to all configurations. The configurations will vary over the following configurations.


1. Metadata cache enable/disable: "--metadata-cache-ttl-secs=0" Or "--metadata-cache-ttl-secs=60"
2. File cache enable/disable: {"file-cache-max-size-mb=-1", "cache-dir=..."} OR {}

tools/integration_tests/unfinalized_appends/setup_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+////////////////////////////////////////////////////////////////////////
+// TestMain
+////////////////////////////////////////////////////////////////////////
+
+func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+
+	// Create storage client before running tests.
```
gargnitingoogle: Let's add a check here to skip if `setup.IsZonalBucketRun()` is false.

tools/integration_tests/unfinalized_appends/setup_test.go
```diff
@@ -0,0 +1,123 @@
...
+	gCtx = context.Background()
```
gargnitingoogle: Let's add a check here to fail if mountedDirectory is passed as it's not supported here.

tools/integration_tests/unfinalized_appends/setup_test.go
```diff
@@ -0,0 +1,123 @@
...
+	testDirName = "AppendFileTest"
+)
+
+var (
+	// Flags for the mount options for gRootDir and gOtherRootDir
+	gFlags []string
+	// Mount function to be used for the mounting.
+	gMountFunc func([]string) error
+
+	// Globals for primary mount.
```
gargnitingoogle: Let's add a little more detailed description of what primary mount is, and what secondary mount is. From the code, looks like primary mount is what is used to write to the unfinalized objects, and secondary mount is what is used to read from unfinalized objects.

nit: I'd suggest switching the definition of primary and secondary. Let secondary be the mount to write the objects, and primary be the mount to read the objects. This is so because we'd need to run multiple configurations on the read mount.

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+// //////////////////////////////////////////////////////////////////////
+// Boilerplate
+// //////////////////////////////////////////////////////////////////////
+
+type UnfinalizedAppendsSuite struct {
```
gargnitingoogle:
```suggestion
type RapidAppendsTestSuite struct {
```

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+func (t *UnfinalizedAppendsSuite) SetupSuite() {
+	setup.MountGCSFuseWithGivenMountFunc(gFlags, gMountFunc)
+	gRootDir = setup.MntDir()
+	gLogFilePath = setup.LogFile()
+	gTestDirPath = setup.SetupTestDirectory(testDirName)
+}
+
+func (t *UnfinalizedAppendsSuite) TearDownSuite() {
+	setup.UnmountGCSFuse(gRootDir)
+	setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
```
gargnitingoogle: SHouldn't this also be inside the `t.T().Failed()` check ?

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+	initialContent, err := operations.ReadFile(path.Join(gTestDirPath, t.fileName))
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), SizeOfFileContents, len(initialContent))
+	t.initialContent = string(initialContent)
+}
+
+// AppendToFile appends "appendContent" using
+// existing appendFileHandle in the first mount.
+func (t *UnfinalizedAppendsSuite) AppendToFile(appendContent string) {
+	n, err := t.appendFileHandle.WriteString(appendContent)
```
gargnitingoogle:
```suggestion
    t.T().Helper()
	n, err := t.appendFileHandle.WriteString(appendContent)
```

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+	assert.NoError(t.T(), err)
```
gargnitingoogle:
```suggestion
	require.NoError(t.T(), err)
```

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+func (t *UnfinalizedAppendsSuite) TestAppendsFromSameMount() {
+	expectedContent := t.initialContent
+	for range 3 {
+		t.AppendToFile(FileContents)
+		expectedContent += FileContents
+
+		// Read content of file from same mount.
+		gotContent, err := operations.ReadFile(path.Join(gTestDirPath, t.fileName))
+
+		assert.NoError(t.T(), err)
```
gargnitingoogle:
```suggestion
		require.NoError(t.T(), err)
```

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+		assert.Equal(t.T(), t.fileSize, len(gotContent))
```
gargnitingoogle: nit: This isn't required if we are comparing the full string in the next line anyway.

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+	expectedContent := t.initialContent
+	for range 3 {
+		t.AppendToFile(FileContents)
+		expectedContent += FileContents
+		operations.SyncFile(t.appendFileHandle, t.T())
+
+		// Read content of file from differnt mount.
+		gotContent, err := operations.ReadFile(path.Join(gOtherTestDirPath, t.fileName))
+
+		assert.NoError(t.T(), err)
```
gargnitingoogle:
```suggestion
		require.NoError(t.T(), err)
```

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
gargnitingoogle: Same goes in the next test function.

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+func (t *UnfinalizedAppendsSuite) SetupTest() {
+	t.fileName = FileName1 + setup.GenerateRandomString(5)
+	var err error
+	// Create unfinalized object.
+	t.appendableWriter = CreateUnfinalizedObject(gCtx, t.T(), gStorageClient, path.Join(testDirName, t.fileName), SizeOfFileContents)
+	t.fileSize = SizeOfFileContents
+
+	t.appendFileHandle, err = os.OpenFile(path.Join(gTestDirPath, t.fileName), os.O_APPEND|os.O_WRONLY|syscall.O_DIRECT, operations.FilePermission_0600)
+	require.NoError(t.T(), err)
+	initialContent, err := operations.ReadFile(path.Join(gTestDirPath, t.fileName))
```
gargnitingoogle: Why do we need to read the file in the setup ?

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
```diff
@@ -0,0 +1,133 @@
...
+////////////////////////////////////////////////////////////////////////
+
+func (t *UnfinalizedAppendsSuite) TestAppendsFromSameMount() {
+	expectedContent := t.initialContent
+	for range 3 {
+		t.AppendToFile(FileContents)
+		expectedContent += FileContents
+
+		// Read content of file from same mount.
+		gotContent, err := operations.ReadFile(path.Join(gTestDirPath, t.fileName))
```
gargnitingoogle: Switch this to `operations.ReadFileSequentially(path.Join(gTestDirPath, t.fileName), util.MiB)` for breaking the reads down into controllable chunk sizes. The same goes for the other test function.

tools/integration_tests/improved_run_e2e_tests.sh
meet2mky: Done!

tools/integration_tests/unfinalized_appends/setup_test.go
meet2mky: Done!

tools/integration_tests/unfinalized_appends/setup_test.go
meet2mky: I observed the use of such naming conventions recently within this [package](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/inactive_stream_timeout/setup_test.go#L52). Based on moma search, this approach appears to deviate from the recommended Go naming conventions, as outlined in the [Go style guide](go/go-style/decisions#variable-names). Consequently, I have reverted this change.

tools/integration_tests/unfinalized_appends/setup_test.go
meet2mky: I have added a fatal log statement to explicitly error the package. This mechanism will ensure failure detection if regional buckets are inadvertently used in any configuration.

tools/integration_tests/unfinalized_appends/setup_test.go
meet2mky: I have added a fatal log statement to explicitly error the package. This mechanism will ensure failure detection if mountedDir runs are performed on this package.

tools/integration_tests/unfinalized_appends/setup_test.go
meet2mky: Done

tools/integration_tests/unfinalized_appends/setup_test.go
meet2mky: Done!

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
meet2mky: Done

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
meet2mky: This is already checked inside the `setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())`.

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
meet2mky: `CreateUnfinalizedObject(gCtx, t.T(), gStorageClient, path.Join(testDirName, t.fileName), SizeOfFileContents)` This method takes only FileSize and generated random bytes object of that size on GCS. To verify reads we initially read the object to ensure appended content to the unfinalized object is returned correctly along with initial random bytes.

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
meet2mky: Done!

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
meet2mky: Done!

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
meet2mky: I don't see how that's useful here?

tools/integration_tests/unfinalized_appends/setup_test.go
meet2mky: I would like to put it like this~
This test package is named `unfinalized_appends` so the primary mount should do the appends.
Secondary mount is for verifying reads. If we want to verify reads with multiple scenarios then we have to extend flagSet mapping between primary mount and secondary mount which is many to one as of now.

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
meet2mky: Done!

tools/integration_tests/unfinalized_appends/unfinalized_appends_test.go
gargnitingoogle: For more complex testing, we need to switch this to reading this in non-uniform chunk sizes. But we can do that in a future PR.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,134 @@
...
+	setup.MountGCSFuseWithGivenMountFunc(flags, mountFunc)
+	rootDir = setup.MntDir()
+	logFilePath = setup.LogFile()
+	testDirPath = setup.SetupTestDirectory(testDirName)
+}
+
+func (t *RapidAppendsSuite) TearDownSuite() {
+	setup.UnmountGCSFuse(rootDir)
+	setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	if t.T().Failed() {
```
gargnitingoogle, line 58: nit: Add a comment here to indicate that the following code is needed specifically only for the 2nd mount.

gargnitingoogle: I am also wondering how we'll be able to differentiate between these two log files in the gcsfuse-kokoro-logs gcs bucket because they'll have similar names and be in the same folder . Any thoughts ?

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (t *RapidAppendsSuite) SetupTest() {
+	t.fileName = FileName1 + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	_ = CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), SizeOfFileContents)
+	t.fileSize = SizeOfFileContents
+	// Create append file handle.
+	var err error
+	t.appendFileHandle, err = os.OpenFile(path.Join(testDirPath, t.fileName), os.O_APPEND|os.O_WRONLY|syscall.O_DIRECT, operations.FilePermission_0600)
+	require.NoError(t.T(), err)
```
gargnitingoogle: nit

```suggestion
	require.NoError(t.T(), err)
    // We'll be reusing the same file handle for appending in the test code as well.
```

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// Root directory which is mounted by gcsfuse.
+	rootDir string
+	// Stores test directory path in the mounted path for rootDir.
+	testDirPath string
+	// Stores log file path for the mount rootDir.
+	logFilePath string
+
+	// Globals for secondary mount which is used to verify reads.
+	// Other Root directory which is mounted by gcsfuse for multi-mount scenarios.
+	otherRootDir string
```
gargnitingoogle: nit:

For readability, let's rename `other<Xyz>` to `secondaryMount<Xyz>` .

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: Thanks, I have added the logs to indicate primary and secondary mount.

tools/integration_tests/rapid_appends/setup_test.go
meet2mky: Done, Explicitly added primary and secondary to dictate the same.

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: This has changed now, PTAL. We are creating file handle within test body as @anushka567 requires more complex file handle management in test body for her tests.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/stretchr/testify/suite"
+)
+
+// //////////////////////////////////////////////////////////////////////
+// Boilerplate
+// //////////////////////////////////////////////////////////////////////
+
+type RapidAppendsSuite struct {
+	fileName       string
+	fileSize       int
```
gargnitingoogle: This is an unnecessary variable and can be removed.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+	initialContent string
```
gargnitingoogle: readability:


```suggestion
	fileContent string
```

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+	suite.Suite
```
gargnitingoogle, line 42: nit: it's preferable to keep `suite.Suite` as first entry in the struct `RapidAppendsSuite ` to signal composition, as this is the base class/struct for RapidAppendsSuite .

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// //////////////////////////////////////////////////////////////////////
+// Helpers
+// //////////////////////////////////////////////////////////////////////
+
+func (t *RapidAppendsSuite) SetupSuite() {
+	setup.MountGCSFuseWithGivenMountFunc(flags, mountFunc)
+	secondaryMntTestDirPath = setup.SetupTestDirectory(testDirName)
```
gargnitingoogle, line 53: This and the corresponding unmount in tear-down is unnecessary in the same-mount tests and should be done only in that test. For now, I am okay with a TODO to handle this later for optimization.


```suggestion
    // TODO: this should be done only for the different-mount test(s).
	secondaryMntTestDirPath = setup.SetupTestDirectory(testDirName)
```
The same goes for the unmount and the log-saving in the TearDownSuite .

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+		log.Println("Primary mount log file:")
+		setup.SetLogFile(primaryMntLogFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+
+func (t *RapidAppendsSuite) SetupTest() {
+	t.fileName = FileName1 + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	_ = CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), SizeOfFileContents)
```
gargnitingoogle: We should have a deletion (`os.RemoveFile`) of `path.Join(testDirName, t.fileName)` in `TeardownTest()` for cleanup .

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+	t.fileSize = SizeOfFileContents
+	initialContent, err := operations.ReadFile(path.Join(primaryMntTestDirPath, t.fileName))
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), SizeOfFileContents, len(initialContent))
+	t.initialContent = string(initialContent)
+}
+
+// AppendToFile appends "appendContent" to the given file.
+func (t *RapidAppendsSuite) AppendToFile(file *os.File, appendContent string) {
```
gargnitingoogle: Can be private

```suggestion
func (t *RapidAppendsSuite) appendToFile(file *os.File, appendContent string) {
```

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (t *RapidAppendsSuite) SetupTest() {
+	t.fileName = FileName1 + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	_ = CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), SizeOfFileContents)
+	t.fileSize = SizeOfFileContents
+	initialContent, err := operations.ReadFile(path.Join(primaryMntTestDirPath, t.fileName))
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), SizeOfFileContents, len(initialContent))
+	t.initialContent = string(initialContent)
```
gargnitingoogle: nit: This package is testing reads of unfinalized objects, so we shouldn't just assume that the read here is reading the data correctly. 

Is it possible to create the unfinalized object with predefined content and then expect that content here as the read-output ? 

If not, please add a TODO here, and we can add it later. It's not a blocker right now as there are other tests to cover it.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+		t.AppendToFile(appendFileHandle, FileContents)
+		expectedContent += FileContents
+
+		// Read content of file from same mount.
+		gotContent, err := operations.ReadFile(path.Join(primaryMntTestDirPath, t.fileName))
+
+		require.NoError(t.T(), err)
+		assert.Equal(t.T(), expectedContent, string(gotContent))
+	}
+	err := appendFileHandle.Close()
```
gargnitingoogle: Do this close as a deferred call after successful open . Without this, there might be a resource leak if ReadFile fails.

gargnitingoogle: Same goes in the next function.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+	require.NoError(t.T(), err)
+}
+
+func (t *RapidAppendsSuite) TestAppendsAndSequentialReadFromSecondaryMount() {
```
gargnitingoogle:
```suggestion
func (t *RapidAppendsSuite) TestAppendsAndReadsFromDifferentMounts() {
```

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,129 @@
...
+	assert.NoError(t.T(), err)
+	assert.Equal(t.T(), len(appendContent), n)
+	t.fileSize += n
+}
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *RapidAppendsSuite) TestAppendsAndSequentialReadFromPrimaryMount() {
```
gargnitingoogle:
```suggestion
func (t *RapidAppendsSuite) TestAppendsAndReadsFromSameMount() {
```

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -0,0 +1,132 @@
...
+func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+	if !setup.IsZonalBucketRun() {
+		log.Fatalf("This package is not supposed to be run with Regional Buckets.")
+	}
+	// TODO(b/431926259): Add support for mountedDir tests as this
+	// package has multi-mount scenario tests and currently we only
+	// pass single mountedDir to test package.
+	if setup.AreBothMountedDirectoryAndTestBucketFlagsSet() {
+		log.Fatalf("This package doesn't support mountedDir tests.")
```
gargnitingoogle: nit

```suggestion
		log.Fatalf("This package doesn't support --mountedDirectory option currently.")
```

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -0,0 +1,132 @@
...
+		if err != nil {
+			log.Fatalf("closeStorageClient failed: %v", err)
+		}
+	}()
+
+	// Set up test directory for primary mount.
+	setup.SetUpTestDirForTestBucketFlag()
+	primaryMntRootDir = setup.MntDir()
+	primaryMntLogFilePath = setup.LogFile()
+	primaryMountFlags := []string{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-global-max-blocks=-1"}
```
gargnitingoogle, line 92: For bookkeeping

```suggestion
    // TODO: `--write-global-max-blocks=-1` is needed right now because of a bug
    // (b/432179045) . Remove it when it is fixed.
	primaryMountFlags := []string{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-global-max-blocks=-1"}
```

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -0,0 +1,132 @@
...
+	secondaryMntRootDir = setup.MntDir()
+	secondaryMntLogFilePath = setup.LogFile()
+	rapidAppendsCacheDir, err := os.MkdirTemp("", "rapid_appends_cache_dir_*")
+	if err != nil {
+		log.Fatalf("Failed to create cache dir for rapid append tests: %v", err)
+	}
+	defer func() {
+		err := os.RemoveAll(rapidAppendsCacheDir)
+		if err != nil {
+			log.Fatalf("Error while cleaning up cache dir: %v", err)
```
gargnitingoogle:
```suggestion
			log.Fatalf("Error while cleaning up cache dir %q: %v", rapidAppendsCacheDir, err)
```

gargnitingoogle (changes requested):
Some comments.

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: > This package is testing reads of unfinalized objects, so we shouldn't just assume that the read here is reading the data correctly.

This read will be as good GCSFuse read of Unfinalized Object as `CreateUnfinalizedObject` does a ZB flush API.

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: I think, it's okay to have secondary mount setup and teardown at one place at the cost of few test not utilising this mount than writing repeated code in multiple tests which would need this setup. Wdyt?

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,137 @@
...
+		log.Println("Primary mount log file:")
+		setup.SetLogFile(primaryMntLogFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+
+func (t *RapidAppendsSuite) SetupTest() {
+	t.fileName = FileName1 + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	_ = CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), SizeOfFileContents)
```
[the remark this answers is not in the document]
meet2mky: it doesn't return err, it returns a writer we don't require as of now.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,137 @@
...
+package rapid_appends
+
+import (
+	"log"
+	"os"
+	"path"
+	"syscall"
+	"testing"
+
+	. "github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/client"
```
gargnitingoogle: Let's not use `.` alias here. This makes the code harder to read by not using the package name in front of function calls.

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: Ok, I am fine with it for now. But please add a TODO here to change/duplicate CreateUnfinalizedObject to take in content instead of size, so we could compare the content after reading it back.

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: Sure, my point is not about doing it at one place or two. But about not doing a second mount when it's not needed, which adds unnecessary complexity and runtime.

meet2mky: added a TODO to split suite.

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: changed the method itself.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,136 @@
...
+// Helpers
+// //////////////////////////////////////////////////////////////////////
+
+func (t *RapidAppendsSuite) SetupSuite() {
+	setup.MountGCSFuseWithGivenMountFunc(flags, mountFunc)
+	secondaryMntTestDirPath = setup.SetupTestDirectory(testDirName)
+}
+
+func (t *RapidAppendsSuite) TearDownSuite() {
+	setup.UnmountGCSFuse(secondaryMntRootDir)
```
gargnitingoogle, line 57: nit: add a comment line here to say why we aren't doing unmount for primary mount here. This will help the reader.

gargnitingoogle (approved):
LGTM

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: We aren't mounting here so I don't see any value, Any way once we split the suite then mounting and unmounting would be more straightforward.

gargnitingoogle: Sounds good.

gargnitingoogle:
@meet2mky Sharing a suggestion for a future improvement. No action in this PR.

In case of different-mount tests, we should make the appends and reads truely concurrent (running in two different go-routines) and not being done in round-robin fashion in the same goroutine, like we're doing right now. This change is needed to make the testing more real-world-like.

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -221,7 +221,7 @@ TEST_PACKAGES_COMMON=(
 # Test packages for regional buckets.
 TEST_PACKAGES_FOR_RB=("${TEST_PACKAGES_COMMON[@]}" "inactive_stream_timeout" "cloud_profiler")
 # Test packages for zonal buckets.
-TEST_PACKAGES_FOR_ZB=("${TEST_PACKAGES_COMMON[@]}" "unfinalized_object")
```
vadlakondaswetha, line 224: unfinalized is also feature of rapid. why are we adding 2 different packages here. why can't we put all rapid specific features under rapid folder

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,136 @@
...
+	fileContent string
+}
+
+// //////////////////////////////////////////////////////////////////////
+// Helpers
+// //////////////////////////////////////////////////////////////////////
+
+func (t *RapidAppendsSuite) SetupSuite() {
+	setup.MountGCSFuseWithGivenMountFunc(flags, mountFunc)
+	secondaryMntTestDirPath = setup.SetupTestDirectory(testDirName)
```
vadlakondaswetha, line 53: we are doing 2 mounts always for the tests which required 1 mount also? is that intentional

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -0,0 +1,136 @@
...
+	// TODO(b/432179045): `--write-global-max-blocks=-1` is needed right now because of a bug in global semaphore release.
+	// Remove this flag once bug is fixed.
+	primaryMountFlags := []string{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-global-max-blocks=-1"}
+	err := static_mounting.MountGcsfuseWithStaticMounting(primaryMountFlags)
+	if err != nil {
+		log.Fatalf("Unable to mount primary mount: %v", err)
+	}
+	// Setup Package Test Directory for primary mount.
+	primaryMntTestDirPath = setup.SetupTestDirectory(testDirName)
+	defer func() {
```
vadlakondaswetha: you can inline it.
defer setup.unmountGCSfuse

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,136 @@
...
+	assert.Equal(t.T(), len(appendContent), n)
+}
+
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *RapidAppendsSuite) TestAppendsAndSequentialReadFromSameMount() {
+	appendFileHandle := operations.OpenFileInMode(t.T(), path.Join(primaryMntTestDirPath, t.fileName), os.O_APPEND|os.O_WRONLY|syscall.O_DIRECT)
+	defer func() {
```
vadlakondaswetha: can we use this method here: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/util/operations/file_operations.go#L615

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	defer func() {
+		err := os.RemoveAll(rapidAppendsCacheDir)
+		if err != nil {
+			log.Fatalf("Error while cleaning up cache dir %q: %v", rapidAppendsCacheDir, err)
+		}
+	}()
+	// Define flag set for secondary mount to run the tests.
+	flagsSet := [][]string{
+		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0"},
```
vadlakondaswetha, line 117: do we want to test all scenarios with file-cache on and off?

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+		// Read content of file from same mount.
+		gotContent, err := operations.ReadFile(path.Join(primaryMntTestDirPath, t.fileName))
+
+		require.NoError(t.T(), err)
+		assert.Equal(t.T(), expectedContent, string(gotContent))
+	}
+}
+
+func (t *RapidAppendsSuite) TestAppendsAndSequentialReadFromDifferentMount() {
```
vadlakondaswetha: entire logic is same except syncfile and secondarymount path. can we combine this as single test with multiple testcases.

tools/integration_tests/rapid_appends/setup_test.go
gargnitingoogle: More scenarios are covered in my dependent PR https://github.com/GoogleCloudPlatform/gcsfuse/pull/3539 .

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: It'll be unnecessarily complex code to put conditional secondary-mount in a single suite. @meet2mky and I talked about it and decided to split the suite into two, one for single-mount and double-mount, both reusing common test logic apart from multi-mount-related differences. It'll be done as a refactor in a future PR.

gargnitingoogle: In this planned change, all the secondary-mount related data itself will be moved to the multi-mount test suite to minimize unnecessary use.

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: Covered in https://github.com/GoogleCloudPlatform/gcsfuse/pull/3473#discussion_r2213841972 .

tools/integration_tests/improved_run_e2e_tests.sh
gargnitingoogle: The purpose for a different package was primarily the handling of the multi-mount in some of the tests needed for reads. One we've refactored the tests into an a multi-mount test-suite (like I mentioned in https://github.com/GoogleCloudPlatform/gcsfuse/pull/3473#discussion_r2213841972) and a single-mount test suite, then we can merge the two packages into one.

tools/integration_tests/rapid_appends/setup_test.go
meet2mky: Done!

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: Done!

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: I made it table driven tests, @gargnitingoogle you can extend it for random reads within the same table. It would be less repeated code.

meet2mky:
Removed file cleanup from the test as in setup_test.go we eventually delete all the objects including testDir for the package.
CC: @gargnitingoogle

tools/integration_tests/improved_run_e2e_tests.sh
gargnitingoogle: After @meet2mky's recent changes, we are a bit closer to this now. If we could move all the primary/secondary mount related logic also to the test-suite file, then we could even merge the these tests to the unfinalized_object package.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -62,17 +62,19 @@ func (t *RapidAppendsSuite) TearDownSuite() {
...
 	// Create unfinalized object.
 	t.fileContent = setup.GenerateRandomString(client.SizeOfFileContents)
 	_ = client.CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), t.fileContent)
 }
 
-func (t *RapidAppendsSuite) TearDownTest() {
-	err := os.Remove(path.Join(primaryMntTestDirPath, t.fileName))
-	require.NoError(t.T(), err)
+func (t *RapidAppendsSuite) SetupTest() {
+	t.createUnfinalizedObject()
```
gargnitingoogle: Isn't this redundant if we're doing the same in SetupSubTest ?

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,140 @@
...
+	setup.UnmountGCSFuse(secondaryMntRootDir)
+	if t.T().Failed() {
+		log.Println("Secondary mount log file:")
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+		log.Println("Primary mount log file:")
+		setup.SetLogFile(primaryMntLogFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+func (t *RapidAppendsSuite) createUnfinalizedObject() {
```
gargnitingoogle: nit: keep a line gap between two functions to be consistent.

tools/integration_tests/rapid_appends/rapid_appends_test.go
meet2mky: right now yes, because I don't have any Test at suite level but would be required if you a Test at suite level.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,140 @@
...
+	t.fileName = client.FileName1 + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	t.fileContent = setup.GenerateRandomString(client.SizeOfFileContents)
+	_ = client.CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), t.fileContent)
+}
+
+func (t *RapidAppendsSuite) SetupTest() {
+	t.createUnfinalizedObject()
```
gargnitingoogle: This seems redundant given that we're doing the same in SetupSubTest .

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,140 @@
...
+		log.Println("Primary mount log file:")
+		setup.SetLogFile(primaryMntLogFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+func (t *RapidAppendsSuite) createUnfinalizedObject() {
+	t.fileName = client.FileName1 + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	t.fileContent = setup.GenerateRandomString(client.SizeOfFileContents)
+	_ = client.CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), t.fileContent)
```
gargnitingoogle: nit: I don't think we need to assign to _ here.

```suggestion
	client.CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), t.fileContent)
```

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,140 @@
...
+			appendFileHandle := operations.OpenFileInMode(t.T(), path.Join(primaryMntTestDirPath, t.fileName), os.O_APPEND|os.O_WRONLY|syscall.O_DIRECT)
+			defer operations.CloseFileShouldNotThrowError(t.T(), appendFileHandle)
+			for range numAppends {
+				t.appendToFile(appendFileHandle, client.FileContents)
+				// Sync the file if the test case requires it.
+				if tc.syncNeeded {
+					operations.SyncFile(appendFileHandle, t.T())
+				}
+
+				readPath := path.Join(tc.readMountPath, t.fileName)
```
gargnitingoogle: We can pull this outside of the `for range numAppends` loop.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,140 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func() {
+			// Open the file for appending on the primary mount.
+			appendFileHandle := operations.OpenFileInMode(t.T(), path.Join(primaryMntTestDirPath, t.fileName), os.O_APPEND|os.O_WRONLY|syscall.O_DIRECT)
+			defer operations.CloseFileShouldNotThrowError(t.T(), appendFileHandle)
+			for range numAppends {
+				t.appendToFile(appendFileHandle, client.FileContents)
```
gargnitingoogle: Why are we using `client.FileContents` here ? Instead of this, we should use some variable specific to this package/file which we can independently configure/maintain .

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,140 @@
...
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+		log.Println("Primary mount log file:")
+		setup.SetLogFile(primaryMntLogFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+func (t *RapidAppendsSuite) createUnfinalizedObject() {
+	t.fileName = client.FileName1 + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	t.fileContent = setup.GenerateRandomString(client.SizeOfFileContents)
```
gargnitingoogle: Why are we using client.SizeOfFileContents here ? Instead of this, we should use some variable specific to this package/file which we can independently configure/maintain .

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -0,0 +1,140 @@
...
+	if t.T().Failed() {
+		log.Println("Secondary mount log file:")
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+		log.Println("Primary mount log file:")
+		setup.SetLogFile(primaryMntLogFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+func (t *RapidAppendsSuite) createUnfinalizedObject() {
+	t.fileName = client.FileName1 + setup.GenerateRandomString(5)
```
gargnitingoogle: let's not use `client.FileName1` . Instead use some variable defined in this package/file. Otherwise any change in client.FileName1 will change the bahviour in this test silently.

gargnitingoogle:
> Removed file cleanup from the test as in setup_test.go we eventually delete all the objects including testDir for the package. CC: @gargnitingoogle

While it functionally works, I still suggest that keep deleting individual's test's/subtest's files in their own clean-up, for (a) faster clean-up, (b) easier debugging if something fails.

meet2mky:
Addressed comments @gargnitingoogle PTAL.

gargnitingoogle (approved):
LGTM

tools/integration_tests/improved_run_e2e_tests.sh
meet2mky: Agreed to @gargnitingoogle and the current rapid_appends package doesn't support --mountedDirectory option so merging can be done once it's supported.

tools/integration_tests/rapid_appends/setup_test.go
vadlakondaswetha: Its not clear what is going in each PR. Can we just add what is required for this PR.

tools/integration_tests/improved_run_e2e_tests.sh
vadlakondaswetha: I am not able to follow. Why are we creating new work as we write tests. I would suggest spending time in thinking how we want to write tests, how should the packages be and then proceed.

tools/integration_tests/rapid_appends/setup_test.go
meet2mky: Added as part of review comments, https://github.com/GoogleCloudPlatform/gcsfuse/pull/3473#discussion_r2206416558 It is not required in this PR.

gargnitingoogle: @vadlakondaswetha 

> Its not clear what is going in each PR.

This PR only adds a basic read while append test for unfinalized object. My PRs (starting with #3539 ) will add more test configurations including metadata-cache on/off, streaming
on/off, file-cache on/off etc.

> Can we just add what is required for this PR.

Sure, will update the PR description accordingly for this PR.

tools/integration_tests/improved_run_e2e_tests.sh
gargnitingoogle: @vadlakondaswetha 

> Why are we creating new work as we write tests

It's not more work. We're reusing the same utilities (e.g. CreateUnfinalizedObject) as we used in the other package. The main reason we had to split out into this package was because of the support for (a) having two different mounts and being able to switch between them for read/operations, (b) not being able to support `--mountedDirectory` for these read tests again because of multi-mount.

> I would suggest spending time in thinking how we want to write tests, how should the packages be and then proceed.

I agree that we didn't consider what the most optimal approach for the code structure would be in this case, and decided to just start with tests, and refactor later if needed.

This PR has already taken a long time and a lot of iterations, so I am hoping that we can merge this in and do restructuring of test files later, to not block further work on it anymore. But if you insist, we'd modify the test structure. Please let me know.

tools/integration_tests/rapid_appends/setup_test.go
meet2mky: Resolving as per offline discussion.

PR #3474, opened 2025-07-07
Make PR review reminder default with Auto applying label on new/open/repopened PR(s).

meet2mky:
### Description
This change adds a actionable which adds `remind-reviewers` label to auto remind for reviews on stale PR(s). If author wants to disable the reminders they can remove the `remind-reviewers` label manually.
* Updated cron to run this job every hour from 9AM to 5PM to remove 48 hrs wait on reminder if the PR is opened just after the reminder job was run.

b/427343327

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


.github/workflows/pr-reminder.yml
```diff
@@ -4,8 +4,8 @@ on:
   # Allow manual runs.
   workflow_dispatch:
   schedule:
-    # Runs at 9:30 UTC, which is 2:30 PM IST weekdays.
-    - cron: '30 9 * * 1-5'
+    # Runs every hour from 9AM to 5PM IST weekdays.
```
charith87, line 7: I see this is hourly, It still checks for PR thats are >24h only right. 
In case Reviewer has not taken a look at 9 AM, will it still notify at 10AM, 11 AM, etc.

meet2mky: This is a frequency of workflow run. Earlier it was like: it would run only at 2:30 PM IST so let's say you open a PR at 3 PM today now when it runs tomorrow then it would run at 2:30 and find the PR is not stale and then run after next to next day would add a reminder but this change would ensure that a run at 4 PM would add a comment for reminder. Subsequent runs would not add reminders because the PR is already reminded once so it would be considered as an activity on the PR for next 24 hours.

charith87: Sounds good. Thanks for this.

PR #3475, opened 2025-07-07
Block Pool: Release Block Flow Improvement for Buffered Write

raj-prince:
### Description
- Replacing blockPool.FreeBlockChannel() with blockPool.Release(block)

b/429960425

### Testing details
1. Manual - Locally tested, working fine as expected.
2. Unit tests - Automation

### DIFF
--- a/internal/block/block_pool.go
+++ b/internal/block/block_pool.go
@@ -113,6 +113,15 @@ func (bp *BlockPool) FreeBlocksChannel() chan Block {
 	return bp.freeBlocksCh
 }
 
+// Release puts the block back into the free blocks channel for reuse.
+func (bp *BlockPool) Release(b Block) {
+	select {
+	case bp.freeBlocksCh <- b:
+	default:
+		panic("Block pool's free blocks channel is full, this should never happen")
+	}
+}
+
 // BlockSize returns the block size used by the blockPool.
 func (bp *BlockPool) BlockSize() int64 {
 	return bp.blockSize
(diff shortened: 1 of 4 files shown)


internal/bufferedwrites/upload_handler_test.go
```diff
@@ -184,7 +184,9 @@ func (t *UploadHandlerTest) TestMultipleBlockUpload() {
 	assert.Equal(t.T(), mockObj, obj)
 	// The blocks should be available on the free channel for reuse.
 	for _, expect := range blocks {
-		got := <-t.uh.freeBlocksCh
+
```
ashmeenkaur: Nit: unecessary line, shall we also assert on free block channel length?

internal/block/block_pool.go
```diff
@@ -113,6 +113,15 @@ func (bp *BlockPool) FreeBlocksChannel() chan Block {
 	return bp.freeBlocksCh
 }
 
+// Release puts the block back into the free blocks channel for reuse.
+func (bp *BlockPool) Release(b Block) {
+	select {
```
ashmeenkaur, line 118: Shall we simplify this to bp.freeBlocksCh <- b? We've observed that select statements, even small ones, can sometimes introduce performance overhead based on our past experience. Alternatively, we can try running profiling on this to see if there is any performance overhead. WDYT? (I don't have a very strong opinion on this though)

ashmeenkaur (dismissed):

internal/block/block_pool.go
raj-prince: We wouldn't see such overhead for default block-size, we need to come up with the workload for this.

Anyway, there is a plan to perform profiler test in the end, will include for buffered-write too, to make sure no such regression.

ashmeenkaur (approved):

PR #3476, opened 2025-07-07
Block: Making block interface a reader

raj-prince:
### Description
- Making the block interface also a reader.
- Passing directly block instead while uploading the content to gcs instead of reader.

b/429969295

### Testing details
1. Manual - Tested locally, working as expected.
2. Unit tests - Automated

### DIFF
--- a/internal/block/block.go
+++ b/internal/block/block.go
@@ -91,17 +91,32 @@ func (m *memoryBlock) Read(bytes []byte) (int, error) {
 	return n, nil
 }
 
+// Seek sets the readSeek position in the block.
+// It returns the new readSeek position and an error if any.
+// The whence argument specifies how the offset should be interpreted:
+//   - io.SeekStart: offset is relative to the start of the block.
+//   - io.SeekCurrent: offset is relative to the current readSeek position.
+//   - io.SeekEnd: offset is relative to the end of the block.
+// It returns an error if the whence value is invalid or if the new
+// readSeek position is out of bounds.
 func (m *memoryBlock) Seek(offset int64, whence int) (int64, error) {
+	newReadSeek := m.readSeek
 	switch whence {
 	case io.SeekStart:
 		m.readSeek = m.offset.start + offset
 	case io.SeekCurrent:
-		m.readSeek += offset
+		newReadSeek += offset
 	case io.SeekEnd:
-		m.readSeek = m.offset.end + offset
+		newReadSeek = m.offset.end + offset
 	default:
 		return 0, fmt.Errorf("invalid whence value: %d", whence)
 	}
+
+	if newReadSeek < m.offset.start || newReadSeek > m.offset.end {
+		return 0, fmt.Errorf("new readSeek position %d is out of bounds", newReadSeek)
+	}
+
+	m.readSeek = newReadSeek
 	return m.readSeek, nil
 }
 
--- a/internal/block/block.go
+++ b/internal/block/block.go
@@ -97,6 +97,7 @@ func (m *memoryBlock) Read(bytes []byte) (int, error) {
 //   - io.SeekStart: offset is relative to the start of the block.
 //   - io.SeekCurrent: offset is relative to the current readSeek position.
 //   - io.SeekEnd: offset is relative to the end of the block.
+//
 // It returns an error if the whence value is invalid or if the new
 // readSeek position is out of bounds.
 func (m *memoryBlock) Seek(offset int64, whence int) (int64, error) {
--- a/internal/block/block.go
+++ b/internal/block/block.go
@@ -15,7 +15,6 @@
 package block
 
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"syscall"
@@ -23,6 +22,13 @@ import (
 
 // Block represents the buffer which holds the data.
 type Block interface {
+	// ReadSeeker is the interface that groups the basic Read and Seek methods.
+	// It is used to read data from the block.
+	io.ReadSeeker
+
+	// Writer provides a way to write data to the block.
+	io.Writer
+
 	// GenBlock defines reuse and deallocation of the block.
 	GenBlock
 
@@ -35,10 +41,6 @@ type Block interface {
 
 	// Write writes the given data to block.
 	Write(bytes []byte) (n int, err error)
-
-	// Reader interface helps in copying the data directly to storage.writer
-	// while uploading to GCS.
-	Reader() io.Reader
 }
 
 // TODO: check if we need offset or just storing end is sufficient. We might need
@@ -51,6 +53,9 @@ type memoryBlock struct {
 	Block
 	buffer []byte
 	offset offset
+
+	// readSeek is used to track the position for reading data.
+	readSeek int64
 }
 
 func (m *memoryBlock) Reuse() {
@@ -58,6 +63,7 @@ func (m *memoryBlock) Reuse() {
 
 	m.offset.end = 0
 	m.offset.start = 0
+	m.readSeek = 0
 }
 
 func (m *memoryBlock) Size() int64 {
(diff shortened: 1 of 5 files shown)


internal/block/block_test.go
```diff
@@ -144,3 +153,86 @@ func (testSuite *MemoryBlockTest) TestMemoryBlockDeAllocate() {
...
+	}{
+		{io.SeekStart, 0, "hello", 0},   // After this, readSeek = 5
+		{io.SeekCurrent, 1, "world", 6}, // After this readSeek = 11
+		{io.SeekEnd, -6, " worl", 5},
+	}
+
+	for _, tt := range tests {
+		testSuite.T().Run(fmt.Sprintf("whence=%d, offset=%d, expectedOutput:%s", tt.whence, tt.offset, tt.expectedOutput), func(t *testing.T) {
+			offset, err := mb.Seek(tt.offset, tt.whence)
+			t.Logf("Seek returned offset: %d, expected: %d, readSeek: %d", offset, tt.expectedOffset, mb.(*memoryBlock).readSeek)
```
ashmeenkaur: Nit: unecessary log

internal/block/block_test.go
```diff
@@ -144,3 +153,86 @@ func (testSuite *MemoryBlockTest) TestMemoryBlockDeAllocate() {
...
+}
+
+func (testSuite *MemoryBlockTest) TestMemoryBlockReadBeyondEnd() {
+	mb, err := createBlock(12)
+	require.Nil(testSuite.T(), err)
+	content := []byte("hello world")
+	n, err := mb.Write(content)
+	require.Nil(testSuite.T(), err)
+	require.Equal(testSuite.T(), len(content), n)
+	readBuffer := make([]byte, 20)
```
ashmeenkaur, line 197: Not sure if the error happening because of read buffer being larger than block or because of seek offset. Can we create 2 separate tests for this?

internal/block/block.go
```diff
@@ -15,28 +15,23 @@
...
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"syscall"
 )
 
 // Block represents the buffer which holds the data.
 type Block interface {
+	io.ReadSeeker
```
codechanges, line 27: A short comment would be helpful here.

internal/block/block.go
```diff
@@ -50,18 +45,54 @@ type memoryBlock struct {
...
 	m.offset.start = 0
+	m.readSeek = 0
 }
 
 func (m *memoryBlock) Size() int64 {
 	return m.offset.end - m.offset.start
 }
+
+// Read reads data from the block into the provided byte slice.
+// Please make sure to call Seek before calling Read if you want to read from a specific position.
```
codechanges, line 78: Do you think adding a helper function SeekThenRead(...) would avoid missing calling seek before reading?

internal/bufferedwrites/upload_handler.go
```diff
@@ -135,7 +135,11 @@ func (uh *UploadHandler) uploader() {
 			uh.wg.Done()
 			continue
 		}
-		_, err := io.Copy(uh.writer, currBlock.Reader())
+		off, err := currBlock.Seek(0, io.SeekStart) // Reset the readSeek to 0 before uploading.
+		if err != nil || off != 0 {
+			logger.Errorf("buffered write upload failed for object %s: error in block.Seek: %v", uh.objectName, err)
+		}
+		_, err = io.Copy(uh.writer, currBlock)
```
codechanges: [Possible issue] if currBlock.Seek() fails, an error is logged, but the execution continues to io.Copy(). This could result in copying from an incorrect offset or an unpredictable state. It would be safer to skip the io.Copy if the Seek operation fails.

internal/block/block.go
raj-prince: Yeah, that will make sense, but adding this method in the interface doesn't look good, we need to add both the method.

internal/bufferedwrites/upload_handler.go
raj-prince: Good catch, thanks!

internal/block/block.go
codechanges: May be in utils somewhere then. Just wanted to enforce contract through a helper method. This is a P2 comment though

codechanges (dismissed):

internal/block/block.go
```diff
@@ -115,19 +148,3 @@ func createBlock(blockSize int64) (Block, error) {
 	}
 	return &mb, nil
 }
-
-// ReadAt reads data from the block at the specified offset.
```
raj-prince, line 119: For reviewer: this method is currently unused and missed in this PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3503

internal/bufferedwrites/upload_handler.go
```diff
@@ -135,18 +135,25 @@ func (uh *UploadHandler) uploader() {
...
-			// Context canceled error indicates that the file was deleted from the
-			// same mount. In this case, we suppress the error to match local
-			// filesystem behavior.
-			err = nil
-		}
-		if err != nil {
-			logger.Errorf("buffered write upload failed for object %s: error in io.Copy: %v", uh.objectName, err)
-			err = gcs.GetGCSError(err)
+		// Reset the readSeek to 0 before uploading.
+		if off, err := currBlock.Seek(0, io.SeekStart); err == nil && off == 0 {
```
ashmeenkaur: nit: we should indent the error flow per the go style guide - go/go-style/decisions#indent-error-flow

I think it makes sense to return early from here. We can refactor by creating a function for releasing the current block and reuse it at line 134, here, and at line 158:
```
uh.blockPool.Release(currBlock)
uh.wg.Done()
```

internal/block/prefetch_block_test.go
```diff
@@ -52,7 +52,10 @@ func (testSuite *PrefetchMemoryBlockTest) TestPrefetchMemoryBlockReuse() {
 
 	pmb.Reuse()
 
-	output, err = io.ReadAll(pmb.Reader())
+	off, err := pmb.Seek(0, io.SeekStart) // Reset the read position to the start of the block.
```
ashmeenkaur: isn't Reuse function setting the seek offset to 0? why are we re-setting it? Shall we just assert that seek offset is 0 instead after Reuse is called?

ashmeenkaur: same at other places too where seek is not required. for eg - after write.

internal/block/block.go
```diff
@@ -68,6 +74,37 @@ func (m *memoryBlock) Cap() int64 {
...
+		return 0, io.EOF
+	}
+
+	n := copy(bytes, m.buffer[m.readSeek:m.offset.end])
+	m.readSeek += int64(n)
+
+	return n, nil
+}
+
+func (m *memoryBlock) Seek(offset int64, whence int) (int64, error) {
```
ashmeenkaur, line 103: Nit: add documentation

internal/block/block_test.go
```diff
@@ -162,3 +170,97 @@ func (testSuite *MemoryBlockTest) TestMemoryBlockCapAfterWrite() {
...
+
+	n, err = mb.Read(readBuffer)
+
+	require.NoError(testSuite.T(), err)
```
ashmeenkaur, line 201: shouldn't this also return EOF error?

raj-prince: Depends on the implementation of Reader interface.

```
/ An instance of this general case is that a Reader returning
// a non-zero number of bytes at the end of the input stream may
// return either err == EOF or err == nil. The next Read should
// return 0, EOF.
```

In our implementation, it will return err == nil, if n > 0. But in the subsequent Read() it will definitely return 0, EOF.

internal/block/block.go
raj-prince: Currently, we are using only at one place, will create utility in case required.

codechanges: sure

internal/block/block_test.go
raj-prince: Discussed offline, will take this as separate PR. As returning EOF early will save one call to reader.

PR #3483, opened 2025-07-08
Upgrade gcloud version in scripts for CD pipeline

Tulsishah:
### Description
During the [initial setup](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/cd_scripts/e2e_test.sh#L40) stage for running tests on arm64 VMs, the gcloud storage cp command fails with a Python module loading error. This issue blocks the execution of our test suites on these architectures.

The failure has been observed on both Debian 11 arm64 and Rocky Linux arm64 environments.

The full error message is as follows:
```
ERROR: gcloud failed to load (gcloud.storage.cp): Problem loading gcloud.storage.cp: No module named 'grpc'.

This usually indicates corruption in your gcloud installation or problems with your Python interpreter.

Please verify that the following is the path to a working Python 3.8+ executable:
    /bin/python3
If it is not, please set the CLOUDSDK_PYTHON environment variable to point to a working Python 3.8+ executable.

If you are still experiencing problems, please run the following command to reinstall:
    $ gcloud components reinstall

If that command fails, please reinstall the Google Cloud CLI using the instructions here:
    https://cloud.google.com/sdk/
```

Python was having latest version only so upgrading gcloud version fixed the issue.

[b/430036204](https://b.corp.google.com/issues/430036204)

### Testing details
1. Manual - Tested manually  by creating VM
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -67,6 +67,7 @@ set -e
 # Print commands and their arguments as they are executed.
 set -x
 
+# Since we are now operating as the starterscriptuser, we need to set the environment variable for this user again.
 export PATH=/usr/local/google-cloud-sdk/bin:$PATH
 
 # Export the RUN_ON_ZB_ONLY variable so that it is available in the environment of the 'starterscriptuser' user.
--- a/tools/cd_scripts/install_test.sh
+++ b/tools/cd_scripts/install_test.sh
@@ -25,7 +25,7 @@ export PATH=/usr/local/google-cloud-sdk/bin:$PATH
 gcloud version && rm gcloud.tar.gz
 
 #details.txt file contains the release version and commit hash of the current release.
-gsutil cp  gs://gcsfuse-release-packages/version-detail/details.txt .
+gcloud storage cp  gs://gcsfuse-release-packages/version-detail/details.txt .
 # Writing VM instance name to details.txt (Format: release-test-<os-name>)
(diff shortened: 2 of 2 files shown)


ashmeenkaur (approved):

tools/cd_scripts/e2e_test.sh
```diff
@@ -58,6 +67,8 @@ set -e
 # Print commands and their arguments as they are executed.
 set -x
 
+export PATH=/usr/local/google-cloud-sdk/bin:$PATH
```
raj-prince, line 71: Seems duplicate, you have already done above.

Tulsishah: It is specifically for startup script user. We are changing the user to run e2e tests so we have to export it again.

raj-prince (approved):

tools/cd_scripts/install_test.sh
```diff
@@ -16,8 +16,16 @@
 # Print commands and their arguments as they are executed.
 set -x
 
+echo "Upgrade gcloud version"
```
raj-prince, line 19: Do we have a way to re-use this?

I understand a single script is passed as a start-up script to VM, so would required more effort to re-use.

raj-prince (approved):

tools/cd_scripts/install_test.sh
Tulsishah: > I understand a single script is passed as a start-up script to VM, so would required more effort to re-use.

Yes exactly. I thought so but we are adding one file in VM metadata and running that file only. We can think it later. resolving it for now as it is blocking to run tests on those VMs.

PR #3490, opened 2025-07-10
Dir Handle changes to support Readdirplus operation

aditimittal2003:
### Description
This PR supports ReadDirPlus implementation by adding the logic to fetch directory entry cores and populate the FUSE response buffer. To support this new functionality efficiently and avoid code duplication, the underlying directory entry processing logic has been refactored.

Key Changes:
A new FetchEntryCores function has been added to retrieve the core inode data for all entries within a directory from GCS.
The ReadDirPlus method uses this data to populate the response buffer with both the entry names and their attributes.

Refactoring to Avoid Duplication:
A generic DirEntry interface has been introduced to abstract common operations over the different directory entry types (fuseutil.Dirent and fuseutil.DirentPlus).
A generic helper function, resolveAndSortEntries, now centralizes the logic for sorting entries, resolving name conflicts, and setting offsets for both the ReadDir and ReadDirPlus code paths.

### Testing details
1. Manual - NA
2. Unit tests - Added comprehensive unit tests in dir_handle_test.go to validate the new ReadDirPlus flow, conflict resolution, and core data fetching logic.
3. Integration tests - NA

### DIFF
--- a/internal/fs/handle/dir_handle.go
+++ b/internal/fs/handle/dir_handle.go
@@ -450,7 +450,6 @@ func (dh *DirHandle) FetchEntryCores(ctx context.Context, op *fuseops.ReadDirPlu
 // ReadDirPlus populates the FUSE response buffer using a pre-processed list
 // of directory entries.
 // LOCKS_REQUIRED(dh.Mu)
-// LOCKS_EXCLUDED(dh.in)
 func (dh *DirHandle) ReadDirPlus(op *fuseops.ReadDirPlusOp, entries []fuseutil.DirentPlus, localEntries map[string]fuseutil.DirentPlus) (err error) {
 	// Sort, resolve conflicts, and set offsets.
 	entries, err = sortAndResolveEntries(entries, localEntries, func(e fuseutil.DirentPlus) *direntPlus { dp := direntPlus(e); return &dp }, func(w *direntPlus) fuseutil.DirentPlus { return fuseutil.DirentPlus(*w) })
--- a/internal/fs/handle/dir_handle.go
+++ b/internal/fs/handle/dir_handle.go
@@ -15,9 +15,10 @@
 package handle
 
 import (
+	"cmp"
 	"fmt"
 	"maps"
-	"sort"
+	"slices"
 
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/fs/inode"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/locker"
@@ -30,35 +31,26 @@ import (
 // DirEntry is a generic interface for directory entries that expose
 // name, type, and offset. It is implemented by fuseutil.Dirent and fuseutil.DirentPlus.
 type DirEntry interface {
-	GetName() string
+	EntryName() string
 	SetName(name string)
-	GetType() fuseutil.DirentType
+	EntryType() fuseutil.DirentType
 	SetOffset(offset fuseops.DirOffset)
 }
 
-// Dirent wraps fuseutil.Dirent to provide method implementations required by the DirEntry interface.
-// Used to allow generic operations over directory entries.
-type Dirent struct {
-	fuseutil.Dirent
-}
-
-// DirentPlus wraps fuseutil.DirentPlus to provide method implementations required by the DirEntry interface.
-// Used to allow generic operations over directory entries.
-type DirentPlus struct {
-	fuseutil.DirentPlus
-}
+type dirent fuseutil.Dirent
+type direntPlus fuseutil.DirentPlus
 
 // For Dirent
-func (d *Dirent) GetName() string                    { return d.Name }
-func (d *Dirent) SetName(name string)                { d.Name = name }
-func (d *Dirent) GetType() fuseutil.DirentType       { return d.Type }
-func (d *Dirent) SetOffset(offset fuseops.DirOffset) { d.Offset = offset }
+func (d *dirent) EntryName() string                  { return d.Name }
+func (d *dirent) SetName(name string)                { d.Name = name }
+func (d *dirent) EntryType() fuseutil.DirentType     { return d.Type }
+func (d *dirent) SetOffset(offset fuseops.DirOffset) { d.Offset = offset }
 
 // For DirentPlus
-func (dp *DirentPlus) GetName() string                    { return dp.Dirent.Name }
-func (dp *DirentPlus) SetName(name string)                { dp.Dirent.Name = name }
-func (dp *DirentPlus) GetType() fuseutil.DirentType       { return dp.Dirent.Type }
-func (dp *DirentPlus) SetOffset(offset fuseops.DirOffset) { dp.Dirent.Offset = offset }
+func (dp *direntPlus) EntryName() string                  { return dp.Dirent.Name }
+func (dp *direntPlus) SetName(name string)                { dp.Dirent.Name = name }
+func (dp *direntPlus) EntryType() fuseutil.DirentType     { return dp.Dirent.Type }
+func (dp *direntPlus) SetOffset(offset fuseops.DirOffset) { dp.Dirent.Offset = offset }
 
 // DirHandle is the state required for reading from directories.
 type DirHandle struct {
@@ -124,13 +116,6 @@ func NewDirHandle(
 // Helpers
 ////////////////////////////////////////////////////////////////////////
 
-// Directory entries, sorted by name.
-type SortedDirEntries[T DirEntry] []T
-
-func (p SortedDirEntries[T]) Len() int           { return len(p) }
-func (p SortedDirEntries[T]) Less(i, j int) bool { return p[i].GetName() < p[j].GetName() }
-func (p SortedDirEntries[T]) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
-
 func (dh *DirHandle) checkInvariants() {
 	// INVARIANT: For each i, entries[i+1].Offset == entries[i].Offset + 1
 	for i := 0; i < len(dh.entries)-1; i++ {
@@ -165,6 +150,12 @@ func (dh *DirHandle) checkInvariants() {
 	}
 }
 
+// compareEntriesByName provides a comparison function for sorting directory entries
+// by name.
+func compareEntriesByName[T DirEntry](a, b T) int {
+	return cmp.Compare(a.EntryName(), b.EntryName())
+}
+
 // Resolve name conflicts between file objects and directory objects (e.g. the
 // objects "foo/bar" and "foo/bar/") by appending U+000A, which is illegal in
 // GCS object names, to conflicting file names.
@@ -172,7 +163,7 @@ func (dh *DirHandle) checkInvariants() {
 // Input must be sorted by name.
 func fixConflictingNames[T DirEntry](entries []T, localEntries map[string]T) (output []T, err error) {
 	// Sanity check.
-	if !sort.IsSorted(SortedDirEntries[T](entries)) {
+	if !slices.IsSortedFunc(entries, compareEntriesByName) {
 		err = fmt.Errorf("expected sorted input")
 		return
 	}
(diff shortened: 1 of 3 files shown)


internal/fs/handle/dir_handle.go
```diff
@@ -77,11 +125,11 @@ func NewDirHandle(
 ////////////////////////////////////////////////////////////////////////
 
 // Directory entries, sorted by name.
-type sortedDirents []fuseutil.Dirent
+type SortedDirEntries[T DirEntry] []T
```
ashmeenkaur: what is the need to make it public (exported)? Can we instead use slices.sort?

internal/fs/handle/dir_handle.go
```diff
@@ -26,6 +27,39 @@ import (
...
+type DirEntry interface {
+	GetName() string
+	SetName(name string)
+	GetType() fuseutil.DirentType
+	SetOffset(offset fuseops.DirOffset)
+}
+
+// Dirent wraps fuseutil.Dirent to provide method implementations required by the DirEntry interface.
+// Used to allow generic operations over directory entries.
+type Dirent struct {
```
ashmeenkaur: Instead of creating wrapper, can we use alias instead so the two types can be type casted (This will be a zero-cost operation in terms of memory allocation. It's a compile-time instruction to treat the underlying data as a different type)

internal/fs/handle/dir_handle.go
```diff
@@ -26,6 +27,39 @@ import (
 	"golang.org/x/net/context"
 )
 
+// DirEntry is a generic interface for directory entries that expose
+// name, type, and offset. It is implemented by fuseutil.Dirent and fuseutil.DirentPlus.
+type DirEntry interface {
+	GetName() string
```
ashmeenkaur: avoid get prefix in golang - [go/go-style/best-practices#function-name-conventions](go/go-style/best-practices#function-name-conventions)

internal/fs/handle/dir_handle.go
```diff
@@ -99,71 +147,139 @@ func (dh *DirHandle) checkInvariants() {
...
 		if eIsDir {
-			prev.Name += inode.ConflictingFileNameSuffix
+			prev.SetName(prev.GetName() + inode.ConflictingFileNameSuffix)
 		} else {
-			e.Name += inode.ConflictingFileNameSuffix
+			e.SetName(e.GetName() + inode.ConflictingFileNameSuffix)
 		}
 
-		output = append(output, *e)
+		output[len(output)-1] = prev
```
ashmeenkaur: Can you explain why this is needed? 
I see we are using pointer receivers only so should be good without this? https://go.dev/play/p/hB-LhiTRqdZ

internal/fs/handle/dir_handle.go
```diff
@@ -99,71 +147,139 @@ func (dh *DirHandle) checkInvariants() {
...
+// directory entries, sorts them, resolves name conflicts, and sets their
+// offsets.
+//
+// This generic function supports both fuseutil.Dirent and fuseutil.DirentPlus
+// by wrapping/ unwrapping them into DirEntry interface-compatible types.
+func sortAndResolveEntries[Entry any, WrappedEntry DirEntry](entries []Entry, localEntries map[string]Entry, wrap func(Entry) WrappedEntry, unwrap func(WrappedEntry) Entry) ([]Entry, error) {
+	// Append local file entries (not synced to GCS).
+	for _, localEntry := range localEntries {
+		entries = append(entries, localEntry)
+	}
```
ashmeenkaur: instead of 2 loops, can we do bothc(line 247) in the same for loop

internal/fs/handle/dir_handle.go
```diff
@@ -99,71 +147,139 @@ func (dh *DirHandle) checkInvariants() {
...
+	// Fix up offset fields.
+	for i, fe := range fixedEntries {
+		fe.SetOffset(fuseops.DirOffset(i) + 1)
+	}
+
+	// Unwrap
+	finalEntries := make([]Entry, 0, len(fixedEntries))
+	for _, fe := range fixedEntries {
+		finalEntries = append(finalEntries, unwrap(fe))
+	}
```
ashmeenkaur: this as well - can we do in a single loop?

internal/fs/handle/dir_handle.go
```diff
@@ -26,6 +28,30 @@ import (
...
+	SetName(name string)
+	EntryType() fuseutil.DirentType
+	SetOffset(offset fuseops.DirOffset)
+}
+
+type dirent fuseutil.Dirent
+type direntPlus fuseutil.DirentPlus
+
+// For Dirent
+func (d *dirent) EntryName() string                  { return d.Name }
```
ashmeenkaur, line 44: Nit: can we just use Name() and Type() instead

aditimittal2003: No we can't. Because Dirent has Name and Type fields. In Go, when a method and a field have the same name, the method takes precedence.Therefore, inside Name() method, the expression d.Name on the right-hand side refers to the method itself, not the field.

internal/fs/handle/dir_handle.go
```diff
@@ -311,3 +417,71 @@ func (dh *DirHandle) ReadDir(
...
+		dh.in.Unlock()
+	}
+
+	return
+}
+
+// ReadDirPlus populates the FUSE response buffer using a pre-processed list
+// of directory entries.
+// LOCKS_REQUIRED(dh.Mu)
+// LOCKS_EXCLUDED(dh.in)
```
ashmeenkaur: LOCKS_EXCLUDED means that this lock should not be acquired before calling this function to avoid deadlock usually because the called function acquires that lock. I don't this that is the case here. WDYT?
refer - https://releases.llvm.org/3.5.1/tools/clang/docs/ThreadSafetyAnalysis.html#locks-excluded

aditimittal2003: Yes, correct. It's not needed here.

internal/fs/handle/dir_handle.go
```diff
@@ -239,6 +318,33 @@ func readAllEntries(
...
+// LOCKS_REQUIRED(in)
+func readAllEntryCores(ctx context.Context, in inode.DirInode) (cores map[inode.Name]*inode.Core, err error) {
+	// Read entries from GCS.
+	// Read one batch at a time.
+	var tok string
+	cores = make(map[inode.Name]*inode.Core)
+	for {
+		// Read a batch from GCS
+		var batch map[inode.Name]*inode.Core
+		batch, tok, err = in.ReadEntryCores(ctx, tok)
```
ashmeenkaur, line 332: Can we add `// LOCKS_REQUIRED(d)` annotation to in.ReadEntryCores and in.ReadEntried implementation in dir.go

internal/fs/handle/dir_handle_test.go
```diff
@@ -278,3 +319,131 @@ func (t *DirHandleTest) EnsureEntriesWithOneLocalFile() {
...
+	AssertEq(2, len(cores))
+
+	entry1, ok := cores[inode.NewFileName(t.dh.in.Name(), "gcsObject1")]
+	AssertTrue(ok, "Core for gcsObject1 not found")
+	t.validateCore(entry1, "gcsObject1", metadata.RegularFileType, "testDir/gcsObject1")
+
+	entry2, ok := cores[inode.NewFileName(t.dh.in.Name(), "gcsObject2")]
+	AssertTrue(ok, "Core for gcsObject2 not found")
+	t.validateCore(entry2, "gcsObject2", metadata.RegularFileType, "testDir/gcsObject2")
+}
```
ashmeenkaur: Nit: use AAA format https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/dev_guide.md#how-to-write-composite-tests

```suggestion
	// validations
	AssertEq(nil, err)
	AssertEq(2, len(cores))
	entry1, ok := cores[inode.NewFileName(t.dh.in.Name(), "gcsObject1")]
	AssertTrue(ok, "Core for gcsObject1 not found")
	t.validateCore(entry1, "gcsObject1", metadata.RegularFileType, "testDir/gcsObject1")
	entry2, ok := cores[inode.NewFileName(t.dh.in.Name(), "gcsObject2")]
	AssertTrue(ok, "Core for gcsObject2 not found")
	t.validateCore(entry2, "gcsObject2", metadata.RegularFileType, "testDir/gcsObject2")
}
```

internal/fs/handle/dir_handle_test.go
```diff
@@ -116,6 +119,44 @@ func (t *DirHandleTest) validateEntry(entry fuseutil.Dirent, name string, filety
...
+			Type: dtype,
+		},
+		Entry: fuseops.ChildInodeEntry{
+			Child:      childInodeID,
+			Attributes: attrs,
+		},
+	}
+}
+
+func (t *DirHandleTest) validateEntryPlus(entry fuseutil.DirentPlus, name string, fileType fuseutil.DirentType, childInodeID fuseops.InodeID) {
```
ashmeenkaur: add expected prefix for parameters you are asserting with in all helper functions

internal/fs/handle/dir_handle_test.go
```diff
@@ -278,3 +319,131 @@ func (t *DirHandleTest) EnsureEntriesWithOneLocalFile() {
...
+	t.dh.entriesPlus = []fuseutil.DirentPlus{{}}
+	t.dh.entriesPlusValid = true
+	op := &fuseops.ReadDirPlusOp{
+		ReadDirOp: fuseops.ReadDirOp{Offset: 1},
+	}
+
+	cores, err := t.dh.FetchEntryCores(t.ctx, op)
+
+	AssertEq(nil, err)
+	AssertEq(nil, cores)
```
ashmeenkaur, line 378: should it be 0 or nil cores?

internal/fs/handle/dir_handle_test.go
```diff
@@ -278,3 +319,131 @@ func (t *DirHandleTest) EnsureEntriesWithOneLocalFile() {
...
+	_, err := storageutil.CreateObject(t.ctx, t.bucket, "testDir/fetchThis", nil)
+	AssertEq(nil, err)
+	op := &fuseops.ReadDirPlusOp{
+		ReadDirOp: fuseops.ReadDirOp{Offset: 1},
+	}
+
+	cores, err := t.dh.FetchEntryCores(t.ctx, op)
+
+	AssertEq(nil, err)
+	AssertEq(1, len(cores))
```
ashmeenkaur, line 393: NIT: assert on the core returned too

internal/fs/handle/dir_handle_test.go
ashmeenkaur: same at other places too

internal/fs/handle/dir_handle_test.go
aditimittal2003: cores is a map[], it is not numeric.

internal/fs/handle/dir_handle.go
```diff
@@ -311,3 +417,69 @@ func (dh *DirHandle) ReadDir(
...
 }
+
+// FetchEntryCores retrieves the core inode data for all entries within the directory from GCS.
+//
+// Special case: If the request offset is zero, it assumes the directory is being read from the
+// beginning and resets the cached list of entries.
+//
+// LOCKS_REQUIRED(dh.Mu)
+// LOCKS_EXCLUDED(dh.in)
+func (dh *DirHandle) FetchEntryCores(ctx context.Context, op *fuseops.ReadDirPlusOp) (cores map[inode.Name]*inode.Core, err error) {
```
ashmeenkaur, line 428: Thinking a bit more about this, `dh.Mu` here should also guards against double-fetching data. We currently hold the mutex only until the data is fetched, but we don't mark the entries as being fetched or valid. This means any potential follow-up request will try to fetch the data again until entriesPlusValid is marked true. Both of these actions should be performed under the same mutex lock. Can you confirm if this understanding is correct?

It will be hard to ensure this if we have separate methods. Can we combine the two methods somehow so this consistency is ensured?

internal/fs/handle/dir_handle_test.go
ashmeenkaur: My bad, what I wanted to verify is that cached data is being used. Probably needs a combined method test (composite test) so we know `t.dh.entriesPlus = []fuseutil.DirentPlus{{}}` is used

internal/fs/handle/dir_handle.go
ashmeenkaur: We discussed offline that this will be handled in a separate PR to unblock other PRs for now, while @aditimittal2003 investigates further.

PR #3496, opened 2025-07-10
fix(bash installation): Fix bash installation script to install gcc/make dependencies before installing bash.

meet2mky:
### Description
Fixing bash installation script to install gcc and make dependencies if they are not present on the machine.
b/430459406

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - As part of presubmit run.

No


perfmetrics/scripts/install_bash.sh
```diff
@@ -29,6 +29,25 @@ fi
...
 
+# Function to install dependencies like gcc and make if not present
+install_dependencies() {
+    if ! command -v gcc &>/dev/null || ! command -v make &>/dev/null; then
+        echo "GCC or make not found. Attempting to install build tools..."
+        if command -v apt-get &>/dev/null; then
+            sudo apt-get update && sudo apt-get install -y build-essential
+        elif command -v dnf &>/dev/null; then
+            sudo dnf install -y gcc make
+        elif command -v yum &>/dev/null; then
```
Tulsishah, line 40: will it work on debian machines as well?

meet2mky: apt-get is the package manager used by Debian and its derivatives. If apt-get is found, the script installs the build-essential package, which contains gcc and make.

Tulsishah (approved):

PR #3502, opened 2025-07-13
feat(buffered-read): passing error for failed state via notification channel

raj-prince:
### Description
- Includes changes to pass the error for the failed state.
- Cancelled notification will be removed in the separate PR.

b/426060431

### Testing details
1. Manual - NA
2. Unit tests - Automation
3. Integration tests - NA


internal/block/block_status.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (b blockStatusDownloadCancelled) Error() error {
+	return b.err
+}
+
+// BlockStatus represents the status of a block download.
+// It provides methods to retrieve the current state of the block and any error
+// that may have occurred during the download process specifically for Failed state.
+type BlockStatus interface {
```
vadlakondaswetha: why do you need an interface and multiple structs.
you can have either interface and 1 struct
or one public struct

Error will be nil incase of success.

I would prefer the 2nd option. public struct. We dont need an interface unless there is a compelling need.

raj-prince: Initially I thought, keeping interface will help create less memory error field will not be there for Success scenario. But later realized interface comes with extra 8 byte to keep the type. 

So, switching back to public struct.

internal/bufferedread/download_task.go
```diff
@@ -68,13 +68,13 @@ func (p *DownloadTask) Execute() {
 	defer func() {
 		if err == nil {
 			logger.Tracef("Download: -> block (%s, %v) completed in: %v.", p.object.Name, blockId, time.Since(stime))
-			p.block.NotifyReady(block.BlockStatusDownloaded)
+			p.block.NotifyReady(block.BlockStatus{State: block.BlockStateDownloaded, Err: nil})
```
vadlakondaswetha: do you need to set it to nil explicitly. isn't it nil by default?

PR #3503, opened 2025-07-13
refactor(block-pool): making the block-pool implementation generic

raj-prince:
### Description
- Making block-pool implementation generic to use it for different block interfaces.

NA

### Testing details
1. Manual - NA
2. Unit tests - Automation.
3. Integration tests - NA

### DIFF
--- a/internal/block/block_pool.go
+++ b/internal/block/block_pool.go
@@ -23,10 +23,18 @@ import (
 
 var CantAllocateAnyBlockError error = errors.New("cant allocate any streaming write block as global max blocks limit is reached")
 
-// BlockPool handles the creation of blocks as per the user configuration.
-type BlockPool struct {
+type GenBlock interface {
+	// Reuse resets the block for reuse.
+	Reuse()
+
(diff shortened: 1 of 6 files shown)


raj-prince:
image

internal/block/block_pool.go
```diff
@@ -23,10 +23,10 @@ import (
 
 var CantAllocateAnyBlockError error = errors.New("cant allocate any streaming write block as global max blocks limit is reached")
 
-// BlockPool handles the creation of blocks as per the user configuration.
-type BlockPool struct {
+// Pool handles the creation of blocks as per the user configuration.
+type Pool[T Block] struct {
```
vadlakondaswetha: why not call it BlockPool

raj-prince: Discussed offline.

Reverting back to GenBlock and GenBlockPool.

Block and PrefetchBlock will inherit GenBlock, where PrefetchBlock inherit Block and add new functionality notification, offset.

vadlakondaswetha (approved):

PR #3521, opened 2025-07-15
test(metrics): Add unit-tests for metrics.

kislaykishore:
### Description
Add unit tests for metrics. This is needed to ensure that after the metrics performance revamp, we can verify that the two implementations are functionally equivalent.

b/431939061

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

No

### DIFF
--- a/common/otel_metrics_test.go
+++ b/common/otel_metrics_test.go
@@ -1,0 +1,1168 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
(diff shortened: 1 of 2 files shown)


common/otel_metrics_test.go
```diff
@@ -0,0 +1,877 @@
...
+	m, err := NewOTelMetrics()
+	require.NoError(t, err)
+	return m, reader
+}
+
+// gatherHistogramMetrics collects all histogram metrics from the reader.
+// It returns a map where the key is the metric name, and the value is another map.
+// The inner map's key is the set of attributes,
+// and the value is the HistogramDataPoint.
+func gatherHistogramMetrics(ctx context.Context, t *testing.T, rd *metric.ManualReader) map[string]map[attribute.Set]metricdata.HistogramDataPoint[int64] {
```
charith87, line 51: This seems fine to test the oTel handler and export.
Shoudl we also add tests to verify that various actions correctly increments all the required metrics. like e.g.
Read of a File should increment FsOps, GCS ReadCount, ReadBytes, etc.

kislaykishore: We have a test for [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/monitoring/prom_test.go#L4). While it doesn't really tests whether it's exactly correct but it verifies that those metrics have been incremented.

charith87: Yea we'll need to create a task for the accuracy/correctness as well as discussed. Just to start off on the main workflows like read, write of an expected size. 
This is probably outside of this CL, But pls create a task for it.

kislaykishore: Created: b/431955973

charith87 (approved):

PR #3525, opened 2025-07-15
perf: MRD improvements

ashmeenkaur:
### Description
This PR optimizes CPU runtime by removing unused and unnecessary code paths within the Multi Range Downloader (MRD) flow.

During CPU profiling of the random reads flow (50GB file, 4KB I/O), we identified three key areas consuming significant CPU time that could be optimized:
1. Duplicate Tracing: Redundant tracing logging within the MRD flow.
2. Excessive UUID Generation: Frequent UUID.New() calls for log IDs.
3. Unnecessary Interrupt Select Statement: A select statement for interrupt handling that was always active, even when not needed.

This PR addresses these issues with the following changes:

1. Streamlined Tracing: Trace logs have been removed from the MRD wrapper. The debug bucket is now only included as a wrapper when TRACE logs are explicitly enabled, preventing unnecessary overhead.
2. Eliminated UUID Generation: UUID.New() calls are no longer performed in the MRD wrapper, as logging is no longer conducted at this level.
3. Conditional Interrupt Handling: The select statement for interrupt handling is now conditionally enabled, activating only when a user explicitly configures interrupt handling.

NOTE: bug [b/432639555](b/432639555) was identified during the implementation. This will be fixed in subsequent PR.

[b/423600392](b/423600392)

### Testing details
1. Manual - NA
2. Unit tests - updated
4. Integration tests - NA
5. Profiling/Perf:
**Profiling:**
Before:
87V2StUjyGV9WZ8

After:
8U8hWZavY9KSVNH

**Bandwidth:**
50G 4K IO:
Before changes: 3976kB/s
After changes: 4343kB/s (~10% improvement)

### DIFF
--- a/internal/gcsx/multi_range_downloader_wrapper.go
+++ b/internal/gcsx/multi_range_downloader_wrapper.go
@@ -99,7 +99,7 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
 	return mrdWrapper.refCount
 }
 
-// IncrementRefCount increment the refcount and cancel any running cleanup function.
+// IncrementRefCount increments the refcount and cancel any running cleanup function.
 // This method should be called exactly once per user of this wrapper.
 // It has to be called before using the MultiRangeDownloader.
 func (mrdWrapper *MultiRangeDownloaderWrapper) IncrementRefCount() {


internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -229,27 +230,30 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) Read(ctx context.Context, buf []b
...
 
-	select {
-	case <-time.After(timeout):
-		err = fmt.Errorf("Timeout")
-	case <-ctx.Done():
-		err = fmt.Errorf("Context Cancelled: %w", ctx.Err())
-	case res := <-done:
+	if !mrdWrapper.config.FileSystem.IgnoreInterrupts {
+		select {
+		case <-time.After(timeout):
```
vadlakondaswetha, line 239: we can remove the timeout since its currently set to a very higher value @abhishek10004

abhishek10004: Yeah, it's fine to remove it for now.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -84,19 +87,19 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) SetMinObject(minObj *gcs.MinObjec
...
-// Returns current refcount.
+// GetRefCount returns current refcount.
 func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
 	mrdWrapper.mu.Lock()
 	defer mrdWrapper.mu.Unlock()
 	return mrdWrapper.refCount
 }
 
-// Increment the refcount and cancel any running cleanup function.
+// IncrementRefCount increment the refcount and cancel any running cleanup function.
```
abhishek10004: nit: increments

internal/gcsx/random_reader.go
```diff
@@ -631,7 +631,7 @@ func (rr *randomReader) readFromRangeReader(ctx context.Context, p []byte, offse
 		// if the reader peters out early. That's fine, but it means we should
 		// have hit the limit above.
 		if rr.reader != nil {
-			err = fmt.Errorf("Reader returned early by skipping %d bytes", rr.limit-rr.start)
+			err = fmt.Errorf("random reader returned early by skipping %d bytes", rr.limit-rr.start)
```
abhishek10004, line 634: nit: range

ashmeenkaur: the other struct name is RangeReader. Added this log here and at internal/gcsx/client_readers/range_reader.go so we can differentiate between the two flows.

internal/gcsx/multi_range_downloader_wrapper.go
ashmeenkaur:
```suggestion
// IncrementRefCount increments the refcount and cancel any running cleanup function.
```

PR #3529, opened 2025-07-16
feat(entry-timeout): Set EntryExpiration for dentry caching

aditimittal2003:
### Description
This change enables dentry (directory entry) caching by setting the entry expiration timeout when the experimental-enable-dentry-cache flag is set.
This change allows the kernel to cache the mapping of file paths to inodes.
The expiration time is set on entries returned by LookUpInode and ReadDirPlus operations.

b/423824806

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/read_dir_plus_test.go
+++ b/internal/fs/read_dir_plus_test.go
@@ -43,7 +43,7 @@ func TestReadDirPlusTestSuite(t *testing.T) {
 func (t *ReadDirPlusTest) SetupSuite() {
 	t.mountCfg.EnableReaddirplus = true
 	t.serverCfg.ImplicitDirectories = true
-	t.serverCfg.InodeAttributeCacheTTL = 60 * time.Second
+	t.serverCfg.InodeAttributeCacheTTL = 5 * time.Second
 	t.serverCfg.NewConfig = &cfg.Config{
 		FileSystem: cfg.FileSystemConfig{
 			ExperimentalEnableDentryCache: true,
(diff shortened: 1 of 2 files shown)


internal/fs/read_dir_plus_test.go
```diff
@@ -122,7 +129,11 @@ func TestLocalFileEntriesReadDirPlusTestSuite(t *testing.T) {
 
 func (t *LocalFileEntriesReadDirPlusTest) SetupSuite() {
 	t.mountCfg.EnableReaddirplus = true
+	t.serverCfg.InodeAttributeCacheTTL = 60 * time.Second
 	t.serverCfg.NewConfig = &cfg.Config{
+		FileSystem: cfg.FileSystemConfig{
+			ExperimentalEnableDentryCache: true,
+		},
```
ashmeenkaur, line 169: Can we add a test by setting a high value to InodeAttributeCacheTTL, modifying the file directly on GCS and validating that on stat after readdirplus, old data cached is served

internal/fs/read_dir_plus_test.go
```diff
@@ -114,6 +114,39 @@ func (t *ReadDirPlusTest) TestDirectoryWithVariousEntryTypes() {
...
+	// This should return the same attributes as Readdirplus.
+	fileInfo, err := os.Stat(filePath)
+
+	// Check that the stat returns the old attributes.
+	assert.Nil(t.T(), err)
+	assert.Equal(t.T(), testFileName, fileInfo.Name())
+	assert.Equal(t.T(), int64(len(initialContent)), fileInfo.Size())
+	// Check stat after cache expiry.
+	// Wait for a duration longer than the metadata cache TTL.
+	time.Sleep(6 * time.Second)
```
ashmeenkaur: 5 and 6 seconds are still a lot for a unit test. I would recommend keeping it in milliseconds. Probably expire after 500 ms and then sleep for a second and validate fresh attributes are served.

ashmeenkaur (dismissed):
One small comment, LGTM otherwise. Please don't run KOKORO tests again after this change as it is only a UT change.

ashmeenkaur (approved):

PR #3530, opened 2025-07-16
test(rapid appends): validate content appended in non-append mode available on GCS only after close()

anushka567:
### Description
This functional test covers the scenario where an unfinalized object is opened in a non-append mode to be written to. According to the heuristics that GCSFuse has in place, such a scenario does not trigger the buffered writes flow, instead falling back to the legacy path, hence, the appends that are done to the unfinalized object will not be visible in GCS in real time. 

The test includes:
- A file handle opened to the unfinalized object in non append mode and "appending" to it.
- Reads from the secondary mount against the same object to ensure the correct behavior.

*NOTE* :  To unskip the test once support for CreateObject() on unfinalized object is in place.

[b/432130679](b/432130679)

### Testing details
1. Manual - Yes. The expected path is followed. However, due to the lack of support of CreateObject() API on unfinalized object, the test will fail due to its invocation in the legacy path. 
2. Unit tests - NA
3. Integration tests - NA


tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -170,6 +170,37 @@ func (t *RapidAppendsSuite) TestAppendSessionInvalidatedByAnotherClientUponTakeo
 	assert.Equal(t.T(), expectedContent, string(content))
 }
 
+func (t *RapidAppendsSuite) TestContentAppendedInNonAppendModeNotVisibleTillClose() {
+	// Skipping test for now until CreateObject() is supported for unfinalized objects.
+	// Ref: b/424253611
+	t.T().Skip()
```
vadlakondaswetha, line 175: same comment as previous PR

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -32,6 +32,8 @@ const (
 	fileNamePrefix = "rapid-append-file-"
 	initialContent = "dummy content"
 	appendContent  = "appended content"
+	// Minimum content size to write in order to trigger block upload while writing ; calculated as (2*blocksize+1) mb.
```
vadlakondaswetha, line 35: add a comment here that buffered writes blockSize is set to 1MB for the tests

anushka567: Added

PR #3534, opened 2025-07-17
chore(ci): Don't run go specific test if PR doesn't contain any related file

raj-prince:
### Description
- Only run go tests when pr contains related changes.
- Used https://github.com/dorny/paths-filter to implement this. Can't use standard [paths or paths ignore](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) in workflow actions as it leads to stuck. This is because linux-test is required, and using path or paths-ignore at workflow level skip the execution of linux-test and PR will get stuck waiting for the status of required `linux-test`.

b/431911068

### Testing details
1. Manual - Tested.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,16 +7,30 @@ on:
   pull_request:
     branches:
       - '*'
-    paths-ignore:
-      - '.github/**'
-      - 'docs/**'
-      - '*.md'
-      - 'tools/**'
-      - 'perfmetrics/**'
-      - 'samples/**'
+permissions:
+  contents: read
 
 jobs:
+  filter:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      run_tests: ${{ steps.filter.outputs.run_tests }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v3.0.2
+        id: filter
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            run_tests:
+              - '!**/*.md'
+              - '!tools/**'
+              - '!perfmetrics/**'
+              - '!samples/**'
   format-test:
+    needs: filter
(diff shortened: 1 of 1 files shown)


kislaykishore (approved):

.github/workflows/ci.yml
```diff
@@ -7,9 +7,35 @@ on:
...
+    outputs:
+      run_tests: ${{ steps.filter.outputs.run_tests }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          # Test related paths are kept in the sorted order top to bottom.
+          filters: |
+            run_tests:
```
kislaykishore, line 26: How will we ensure that every new package gets added to this? Should we instead do something like "!**/.md" etc.?

raj-prince: Yeah, this is a problem.

I tried to the same thing but the problem with the !**/*.md is we can only use provide one predicate. If we provide such multiple predicates then it'll not work as intended.

Other simple options is to path-ignore at workflow level and make these tests optional.

raj-prince: Recently added feature `predicate-quantifier: 'every'` seems working.

.github/workflows/ci.yml
```diff
@@ -7,9 +7,30 @@ on:
...
+          predicate-quantifier: 'every'
+          filters: |
+            run_tests:
+              - '!**/*.md'
+              - '!tools/**'
+              - '!perfmetrics/**'
+              - '!samples/**'
   format-test:
+    needs: filter
+    if: ${{ needs.filter.outputs.run_tests == 'true' }}
```
kislaykishore: Formatting is cheap. I'd also like integration tests and other golang code to be correctly formatted when they are changed. So, let's remove this filtering from here.

Same comment for lint as well.

PR #3538, opened 2025-07-17
test(rapid appends): Appends to finalized object is not visible on GCS till Close()

anushka567:
### Description
The functional test introduced in this PR asserts that the appends made to finalized object will not be real time i.e. not immediately available for reads. This is so because real time appends is a feature of unfinalized object only. 

Write behavior for finalized object will be similar to objects in regional bucket.

[b/432141061](b/432141061)

### Testing details
1. Manual - Yes.
`--- PASS: TestRapidAppendsSuite/TestAppendsToFinalizedObjectNotVisibleUntilClose (1.36s)`
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/rapid_appends/appends_test.go
+++ b/tools/integration_tests/rapid_appends/appends_test.go
@@ -201,6 +201,39 @@ func (t *RapidAppendsSuite) TestContentAppendedInNonAppendModeNotVisibleTillClos
 	assert.Equal(t.T(), expectedContent, contentAfterClose)
 }
 
+func (t *RapidAppendsSuite) TestAppendsToFinalizedObjectNotVisibleUntilClose() {
+	t.fileName = fileNamePrefix + setup.GenerateRandomString(5)
+	// Create Finalized Object in the GCS bucket.
+	client.CreateObjectInGCSTestDir(
+		ctx, storageClient, testDirName, t.fileName, initialContent, t.T())
+	defer func() {
+		err := os.Remove(path.Join(primaryMntTestDirPath, t.fileName))
+		require.NoError(t.T(), err)
+	}()
+
+	// Append to the finalized object from the primary mount.
+	filePath := path.Join(primaryMntTestDirPath, t.fileName)
+	fh, err := os.OpenFile(filePath, os.O_APPEND|os.O_RDWR|syscall.O_DIRECT, operations.FilePermission_0600)
+	require.NoError(t.T(), err)
+	_, err = fh.Write([]byte(appendContent))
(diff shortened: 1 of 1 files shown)


tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -201,6 +201,39 @@ func (t *RapidAppendsSuite) TestContentAppendedInNonAppendModeNotVisibleTillClos
...
+	secondaryPath := path.Join(secondaryMntTestDirPath, t.fileName)
+	contentBeforeClose, err := operations.ReadFile(secondaryPath)
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), initialContent, string(contentBeforeClose))
+
+	// Close the file handle used for appending.
+	require.NoError(t.T(), fh.Close())
+
+	// Read the object from secondary mount to validate that appended content is now visible on GCS.
+	expectedContent := initialContent + appendContent
```
vadlakondaswetha: just an fyi, you dont need 2 mounts for these tests. you can directly read from GCS and verify too

anushka567: Updated to read from GCS directly.

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -201,6 +201,35 @@ func (t *RapidAppendsSuite) TestContentAppendedInNonAppendModeNotVisibleTillClos
 	assert.Equal(t.T(), expectedContent, contentAfterClose)
 }
 
+func (t *RapidAppendsSuite) TestAppendsToFinalizedObjectNotVisibleUntilClose() {
+	t.fileName = fileNamePrefix + setup.GenerateRandomString(5)
+	// Create Finalized Object in the GCS bucket.
+	client.CreateObjectInGCSTestDir(
+		ctx, storageClient, testDirName, t.fileName, initialContent, t.T())
+
+	// Append to the finalized object from the primary mount.
```
vadlakondaswetha, line 210: add a comment here that the size of the data ensures that first block is flushed when streaming writes are enabled

vadlakondaswetha (approved):

tools/integration_tests/rapid_appends/appends_test.go
anushka567: This is not going to be the streaming writes flow, as the object is finalized and non-zero size. 
I have still written enough data so that its an apple to apple comparison, that the same amount of data written in the following tests will lead to implicit flush , but here it is not going to be the case.

PR #3539, opened 2025-07-17
ci(read unfinalized objects): Extends (add md-cache and file-cache configs) and refactors rapid_appends e2e tests

gargnitingoogle:
### Description
This pull request significantly enhances the e2e test package rapid_appends functionality in GCSFuse. The primary focus is on expanding test coverage to include various metadata and file cache configurations ([b/433142062](http://b/433142062)), ensuring correct read behavior for unfinalized objects, especially around metadata cache expiry ([b/433143404](http://b/433143404)). Concurrently, the test suite structure has undergone a substantial refactoring ([b/433143407](http://b/433143407)) to improve maintainability and provide more granular control over test environments.

### Highlights

* **Expanded Test Coverage for Caching**: New test configurations have been added to the `rapid_appends` test package, specifically enabling and disabling metadata and file caches. This ensures comprehensive validation of read operations on unfinalized objects under different caching scenarios.
* **Metadata Cache Expiry Validation**: The tests now explicitly verify that reads return pre-append and post-append sizes correctly when metadata cache is enabled, by reading before and after the metadata TTL expiry.
* **Refactored Test Suite Architecture**: The `rapid_appends` test package has been refactored into three distinct test suites: `RapidAppendsSuite` (base), `SingleMountRapidAppendsSuite`, and `DualMountRapidAppendsSuite`. This new structure allows each test to control its own mount flag sets, improving test isolation and management.

<details>
<summary><b>Changelog</b></summary>

* **tools/integration_tests/rapid_appends/appends_test.go**
    * Removed boilerplate `RapidAppendsSuite` and its helper methods, along with the `TestAppendsAndRead` function.
    * Updated `TestAppendSessionInvalidatedByAnotherClientUponTakeover` and `TestContentAppendedInNonAppendModeNotVisibleTillClose` to run with various cache configurations, managing mount points within the test logic.
* **tools/integration_tests/rapid_appends/reads_after_appends_test.go**
    * **New file:** Introduced `TestAppendsAndReads` to specifically validate read consistency after appends across different metadata and file cache settings, including checks for metadata TTL expiry.
* **tools/integration_tests/rapid_appends/setup_test.go**
    * Removed global mount-related variables and simplified the `TestMain` function, delegating mount setup to the new test suites.
    * Added new constants for metadata cache, file cache, and rapid appends flags.
* **tools/integration_tests/rapid_appends/suites_test.go**
    * **New file:** Introduced a new test suite architecture with `RapidAppendsSuite` (base), `SingleMountRapidAppendsSuite`, and `DualMountRapidAppendsSuite` to manage test environments and mount configurations more effectively.
    * Added helper methods for mount point management, object creation/deletion, and file appending within the new suite structure.
</details>

#### What is planned in next PR(s):
- Add support for sequential/random read configurations ([b/433142769](http://b/433142769)). This is WIP in https://github.com/GoogleCloudPlatform/gcsfuse/pull/3543 .

[b/433142062](http://b/433142062)
[b/433143407](http://b/433143407)
[b/433143404](http://b/433143404)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Ran as part of presubmit: [run1](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/3061b785-5452-4ab2-8ded-e926b338bc02/details), [run2](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/08745b8c-67f7-4dcd-b25e-058fc41fe3e6), [run3](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/8a3570cd-e25f-4855-ae8f-0c27797d20b7/summary), [run4](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/ca3f85a4-21e4-4476-b7d1-720580c5c4e3/details), [run5](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/88030833-e9cd-4d53-bb00-a266d1e6243e), [run6](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/1f7a3969-1dd4-4410-b7ef-b20a0d706317) .

### DIFF
--- a/tools/integration_tests/rapid_appends/appends_test.go
+++ b/tools/integration_tests/rapid_appends/appends_test.go
@@ -41,8 +41,7 @@ func (t *DualMountAppendsSuite) TestAppendSessionInvalidatedByAnotherClientUponT
 	const initialContent = "dummy content"
 	const appendContent = "appended content"
 	for _, flags := range [][]string{
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1"},
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1", "--file-cache-max-size-mb=-1", "--cache-dir=" + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+		{"--write-experimental-enable-rapid-appends=true", "--write-block-size-mb=1"},
 	} {
 		var err error
 		func() {
@@ -99,8 +98,7 @@ func (t *SingleMountAppendsSuite) TestContentAppendedInNonAppendModeNotVisibleTi
 	t.T().Skip()
 
 	for _, flags := range [][]string{
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1"},
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1", "--file-cache-max-size-mb=-1", "--cache-dir=" + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+		{"--write-experimental-enable-rapid-appends=true", "--write-block-size-mb=1"},
 	} {
 		func() {
 			t.mountPrimaryMount(flags)
@@ -144,8 +142,7 @@ func (t *SingleMountAppendsSuite) TestContentAppendedInNonAppendModeNotVisibleTi
 func (t *SingleMountAppendsSuite) TestAppendsToFinalizedObjectNotVisibleUntilClose() {
 	const initialContent = "dummy content"
 	for _, flags := range [][]string{
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1"},
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1", "--file-cache-max-size-mb=-1", "--cache-dir=" + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+		{"--write-experimental-enable-rapid-appends=true", "--write-block-size-mb=1"},
 	} {
 		func() {
 			t.mountPrimaryMount(flags)
@@ -186,8 +183,7 @@ func (t *SingleMountAppendsSuite) TestAppendsToFinalizedObjectNotVisibleUntilClo
 func (t *SingleMountAppendsSuite) TestAppendsVisibleInRealTimeWithConcurrentRPlusHandle() {
 	const initialContent = "dummy content"
 	for _, flags := range [][]string{
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1"},
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1", "--file-cache-max-size-mb=-1", "--cache-dir=" + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+		{"--write-experimental-enable-rapid-appends=true", "--write-block-size-mb=1"},
 	} {
 		func() {
 			t.mountPrimaryMount(flags)
@@ -241,8 +237,7 @@ func (t *SingleMountAppendsSuite) TestRandomWritesVisibleAfterCloseWithConcurren
 
 	const initialContent = "dummy content"
 	for _, flags := range [][]string{
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1"},
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1", "--file-cache-max-size-mb=-1", "--cache-dir=" + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+		{"--write-experimental-enable-rapid-appends=true", "--write-block-size-mb=1"},
 	} {
 		func() {
 			t.mountPrimaryMount(flags)
@@ -295,8 +290,7 @@ func (t *SingleMountAppendsSuite) TestFallbackHappensWhenNonAppendHandleDoesFirs
 	t.T().Skip()
 
 	for _, flags := range [][]string{
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1"},
-		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-block-size-mb=1", "--file-cache-max-size-mb=-1", "--cache-dir=" + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+		{"--write-experimental-enable-rapid-appends=true", "--write-block-size-mb=1"},
 	} {
 		func() {
 			t.mountPrimaryMount(flags)


tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -56,8 +61,54 @@ var (
...
+			enableFileCache:     true,
+		},
+		{ // File cache enabled.
+			enableFileCache: true,
+		},
+	}
+}
+
+func flagsFromScenario(scenario scenarioConfig, rapidAppendsCacheDir string) []string {
+	// TODO: Make these constant literals somewhere.
```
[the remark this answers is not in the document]
gargnitingoogle: done

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -56,8 +66,53 @@ var (
...
+	if scenario.enableMetadataCache {
+		flags = append(flags, metadataCacheEnableFlags...)
+	} else {
+		flags = append(flags, metadataCacheDisableFlags...)
+	}
+	if scenario.enableFileCache {
+		flags = append(flags, fileCacheEnableFlags...)
+	} else {
+		flags = append(flags, fileCacheDisableFlags...)
+	}
```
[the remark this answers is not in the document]
gargnitingoogle: Sounds good. Will make the change.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -194,6 +194,15 @@ func RemoveFile(filePath string) {
 	}
 }
 
+// RemoveFileStrictly deletes the given files,
+// and logs fatally if the delete failed for any reason.
+func RemoveFileStrictly(filePath string) {
```
gargnitingoogle: This function is not used anywhere. Remove it.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -123,7 +129,13 @@ func (t *RapidAppendsSuite) TestAppendsAndRead() {
 				gotContent, err := operations.ReadFile(readPath)
 
 				require.NoError(t.T(), err)
-				assert.Equal(t.T(), t.fileContent, string(gotContent))
+				readContent := string(gotContent)
+				expectedContent := t.fileContent
```
gargnitingoogle: this looks like an unnecessary alias. This is neither shorter, not more readable. Might as well as remove it.

tools/integration_tests/util/operations/file_operations.go
gargnitingoogle: done

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: done

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -29,37 +30,55 @@ import (
...
+		// Undo secondary mount.
+		setup.UnmountGCSFuse(t.secondaryMount.rootDir)
+		// Clean up.
+		if t.T().Failed() {
+			setup.SetLogFile(t.secondaryMount.logFilePath)
+			log.Printf("Saving secondary mount log file ...")
+			setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+		}
+	}
 	if t.T().Failed() {
```
gargnitingoogle: nit: This should be moved to the caller code in TestMain() .

gargnitingoogle: done

tools/integration_tests/rapid_appends/setup_test.go
meet2mky, line 1: Since this is going to be merged with the `unfinalized_object` package shouldn't we make it similar to that package for setup_test.go file? How do you think we should plan for merging with minimum friction?

gargnitingoogle: @meet2mky  Yes, that's the plan. I am making changes step by step to move to that final state. In that direction, this PR moves the secondary mount completely to the RapidAppendsSuite. In another PR, I would move the primary mount related variables and its mount/unmount/configs also to RapidAppendsSuite . It felt like complicating it too much doing it all in a single PR. So, this is an intermediate step with also adding more test configurations.

gargnitingoogle:
@anushka567 This is the PR I Was talking about.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -29,37 +30,55 @@ import (
...
 // //////////////////////////////////////////////////////////////////////
 
-// TODO: Split the suite in two suites single mount and multi-mount.
 type RapidAppendsSuite struct {
 	suite.Suite
 	fileName    string
 	fileContent string
+
+	// Optional values for supporting secondary mount.
+	hasSecondaryMount bool
```
gargnitingoogle: nit

```suggestion
	enableSecondaryMount bool
```

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -91,39 +110,42 @@ func (t *RapidAppendsSuite) appendToFile(file *os.File, appendContent string) {
...
 
 				gotContent, err := operations.ReadFile(readPath)
 
 				require.NoError(t.T(), err)
-				assert.Equal(t.T(), t.fileContent, string(gotContent))
+				readContent := string(gotContent)
+				if !scenario.enableMetadataCache {
+					assert.Equal(t.T(), t.fileContent, readContent)
+				} else {
+					assert.Truef(t.T(), strings.HasPrefix(t.fileContent, readContent), "With metadata-enabled, read content expected to be a prefix of the written content, but failed. Written content = %q, Read content = %q", t.fileContent, readContent)
```
gargnitingoogle: nit: Add a comment to describe why we are matching readContent as a prefix of the full file content.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -91,39 +110,42 @@ func (t *RapidAppendsSuite) appendToFile(file *os.File, appendContent string) {
...
-			readMountPath: secondaryMntTestDirPath,
-			syncNeeded:    true, // Sync is required for writes to be visible on another mount.
+			name: "SequentialRead",
 		},
 	}
 
+	// Open the file for appending on the primary mount.
+	appendMountPath := primaryMount.testDirPath
+	syncNeeded := false
+	if t.hasSecondaryMount {
```
vadlakondaswetha: previous code was more cleaner than this. why dont we have a seperate test suite for 2 mounts vs single mount

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -85,47 +128,44 @@ func TestMain(m *testing.M) {
...
-		log.Printf("Running tests with flags: %v", flagsSet[i])
-		flags = flagsSet[i]
-		successCode = m.Run()
+	for _, scenario = range scenariosToBeRun() {
+		successCode = func() int {
+			// Create a cache-dir if needed.
+			var rapidAppendsCacheDir string
+			var err error
+			if scenario.enableFileCache {
+				rapidAppendsCacheDir, err = os.MkdirTemp("", "rapid_appends_cache_dir_*")
```
vadlakondaswetha: can you point me to the code on how the cleanup is done for read_cache_tests

gargnitingoogle: cache-dir is not explicitly deleted in the read_cache tests. It's created inside the test-dir at https://github.com/GoogleCloudPlatform/gcsfuse/blob/8b4e5ee56bdf8040975f61abc5a1d1a928b89508/tools/integration_tests/read_cache/setup_test.go#L116. The test-dir is deleted by the test set-up during the unmount I think. I'll do the same to simplify code.

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: > previous code was more cleaner than this.

Agree. But this is in transition to what we agreed upon in https://github.com/GoogleCloudPlatform/gcsfuse/pull/3473#discussion_r2213825072 where we committed to merging the tests in this packaged into the unfinalized_object package. So, this cluttering up the RapidAppends suite is expected. This is not the final state though.

> why dont we have a seperate test suite for 2 mounts vs single mount

I do have two separate test suite functions TestRapidAppendsSingleMountSuite and TestRapidAppendsDualMountSuite both of which use the same struct with an optional secondary-mount. 
The reason to keep a single struct was to avoid duplicating the logic of this test function TestAppendsAndReads itself. The best we could do to simplify this logic would be to abstract out some of this logic into two different structs one for single-mount and a dual-mount with some abstraction.

tools/integration_tests/rapid_appends/setup_test.go
gargnitingoogle: I have now made the cache-dir management similar to that in the read_cache package, except for the deletion and creation in-between to avoid any interference between the tests.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -37,48 +37,29 @@ const (
 )
 
 // //////////////////////////////////////////////////////////////////////
-// Boilerplate
+// Suite Setup and Teardown
 // //////////////////////////////////////////////////////////////////////
 
+// RapidAppendsSuite is the suite for rapid appends tests with a single mount,
+// i.e. primary mount for for both reading and writing/appending.
```
gargnitingoogle: nit: typo

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: done

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: Now I have split the single suite to two suites, RapidAppendsSuite which has a single mount, and DualMountRapidAppendsSuite which has a second mount for appending/mounting. The DualMountRapidAppendsSuite also does an fsync after every write/append through it.

> previous code was more cleaner than this

Please take a look again at the latest code and let me know if this is still the case. If yes, let's discuss this.

tools/integration_tests/rapid_appends/rapid_appends_test.go
```diff
@@ -142,17 +142,26 @@ func (t *RapidAppendsSuite) TestAppendsAndRead() {
...
 				readContent := string(gotContent)
-				if !scenario.enableMetadataCache {
-					assert.Equal(t.T(), t.fileContent, readContent)
+				if !scenario.enableMetadataCache || !t.isSyncNeededAfterAppend || (i == 0) {
+					assert.Equalf(t.T(), t.fileContent, readContent, "failed to match full content in non-metadata-cache/single-mount after %v appends", i+1)
 				} else {
-					assert.Truef(t.T(), strings.HasPrefix(t.fileContent, readContent), "With metadata-enabled, read content expected to be a prefix of the written content, but failed. Written content = %q, Read content = %q", t.fileContent, readContent)
+					assert.Equalf(t.T(), t.fileContent[:sizeBeforeAppend], readContent, "failed to match partial content in metadata-cache dual-mount after %v appends", i+1)
+
+					time.Sleep(time.Duration(metadataCacheTTLSecs) * time.Second) // Wait for metadata cache to be updated.
```
gargnitingoogle: s/updated/expired

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -85,47 +151,29 @@ func TestMain(m *testing.M) {
...
 	log.Println("Running static mounting tests...")
 	mountFunc = static_mounting.MountGcsfuseWithStaticMounting
 
 	var successCode int
-	for i := range flagsSet {
-		log.Printf("Running tests with flags: %v", flagsSet[i])
-		flags = flagsSet[i]
-		successCode = m.Run()
+	for _, scenario = range scenariosToBeRun() {
+		successCode = func() int {
```
meet2mky: I would suggest we shouldn't put this much logic in just generating the flagSet, From readers perspective we need to look at the  `scenariosToBeRun` and then `flagsFromScenario` to mentally generate the flagSet whereas we could simply put them hardcoded as we have done in many other test packages and from a single glance you could tell for what mount configurations a package is running.. Adding new flagSets become easier you could just add a new array to flagSet while in the current implementation we would need to add a scenario and verify that it does generate the correct flagSet. Wdyt?
go/unit-testing-practices?polyglot=go#logic

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -30,35 +31,100 @@ import (
...
+		flags = append(flags, metadataCacheDisableFlag)
+	}
+	if scenario.enableFileCache {
+		flags = append(flags, fileCacheMaxSizeFlag, cacheDirFlagPrefix+rapidAppendsCacheDir)
+	}
+	return flags
+}
+
+func getNewEmptyCacheDir() string {
+	// Clear previous test's cache directory to minimize
```
gargnitingoogle: We can simplify this code by removing the logic of clearing previous cache-dir, and only creating and returning a new one here. With this change, we also shouldn't store the cacheDirPath variable.

gargnitingoogle: done

tools/integration_tests/rapid_appends/setup_test.go
gargnitingoogle: This was needed at the time because I needed to know the values of enableMetadataCache (and possibly even enableFileCache) in the test code for the currently running test scenario. That wasn't possible if I directly stored a list of hardcoded flag-sets. So, I had to break flagSet down into an intermediate representation called scenarioConfig. 

I had a discussion with @vadlakondaswetha around refactoring needed in the test code. Let's revisit your question once I finish that refactoring.

gargnitingoogle:
Converted to draft, as I am refactoring this code based on my discussion with @vadlakondaswetha . Will reopen for review soon.

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: obsolete

tools/integration_tests/rapid_appends/rapid_appends_test.go
gargnitingoogle: Swetha and I had a discussion, where she shared her suggestion for how to refactor the test code to close on this discussion. I am working on that now. Marking this resolved.

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,193 +19,122 @@ import (
...
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks, "--write-block-size-mb=1", fileCacheMaxSizeFlag, cacheDirFlagPrefix + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+	} {
+		var err error
+		func() {
+			t.mountPrimaryMount(flags)
+			defer t.unmountPrimaryMount()
+
+			log.Printf("Running tests with flags: %v", flags)
+
+			const initialContent = "dummy content"
```
gargnitingoogle: nit: These two are probably not required as they might be already defined elsewhere in the package.

gargnitingoogle: No, these variables aren't defined elsewhere, so these are just okay.

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,193 +19,122 @@ import (
...
+			// append session.
+			err = newAppendFileHandle.Sync()
+			assert.NoError(t.T(), err)
+
+			// Read from primary mount to validate the contents which has persisted in GCS after
+			// takeover from the secondary mount.
+			// Close the open append handle before issuing read on the file as Sync() triggered on
+			// ReadFile() due to BWH still being initialized, is expected to error out with stale NFS file handle.
+			operations.CloseFileShouldThrowError(t.T(), appendFileHandle)
+			expectedContent := t.fileContent + appendContent
```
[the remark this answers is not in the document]
gargnitingoogle: No, the original code logic is correct, because the append from the primary mount never worked, so initialContent would not end up in the file reading it.

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,193 +19,122 @@
...
+			require.NoError(t.T(), err)
+			assert.Equal(t.T(), initialContent, contentBeforeClose)
+
+			// Close() from primary mount to ensure data persists in GCS.
+			err = wh.Close()
+			require.NoError(t.T(), err)
+
+			// Validate that appended content is visible in GCS.
+			expectedContent := initialContent + data
+			contentAfterClose, err := client.ReadObjectFromGCS(ctx, storageClient, path.Join(path.Join(testDirName, t.fileName)))
```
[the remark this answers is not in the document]
gargnitingoogle: good point. Will fix.

tools/integration_tests/rapid_appends/reads_after_appends_test.go
```diff
@@ -0,0 +1,110 @@
...
+	}, {
+		enableMetadataCache: true,
+		enableFileCache:     true,
+		flags:               []string{writeRapidAppendsEnableFlag, infiniteWriteGlobalMaxBlocks, fmt.Sprintf("%s%v", metadataCacheEnableFlagPrefix, metadataCacheTTLSecs), fileCacheMaxSizeFlag, cacheDirFlagPrefix + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+	}, {
+		enableMetadataCache: false,
+		enableFileCache:     true,
+		flags:               []string{writeRapidAppendsEnableFlag, infiniteWriteGlobalMaxBlocks, metadataCacheDisableFlag, fileCacheMaxSizeFlag, cacheDirFlagPrefix + getNewEmptyCacheDir(t.primaryMount.rootDir)},
+	}} {
+		flags := scenario.flags
```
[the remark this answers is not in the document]
gargnitingoogle: ok. Sounds good.

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -30,40 +30,31 @@ import (
 const (
 	testDirName    = "RapidAppendsTest"
 	fileNamePrefix = "rapid-append-file-"
-	initialContent = "dummy content"
-	appendContent  = "appended content"
 	// Minimum content size to write in order to trigger block upload while writing ; calculated as (2*blocksize+1) mb.
 	// Block size for buffered writes is set to 1MiB.
 	contentSizeForBW = 3
+	// metadataCacheTTLSecs is applicable only for those tests for which enableMetadataCache is true.
+	metadataCacheTTLSecs          int = 10
```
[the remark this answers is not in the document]
gargnitingoogle: done

tools/integration_tests/rapid_appends/reads_after_appends_test.go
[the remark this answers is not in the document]
gargnitingoogle: done

tools/integration_tests/rapid_appends/appends_test.go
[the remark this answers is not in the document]
gargnitingoogle: done

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -30,35 +30,22 @@ import (
 const (
 	testDirName    = "RapidAppendsTest"
 	fileNamePrefix = "rapid-append-file-"
-	initialContent = "dummy content"
-	appendContent  = "appended content"
 	// Minimum content size to write in order to trigger block upload while writing ; calculated as (2*blocksize+1) mb.
 	// Block size for buffered writes is set to 1MiB.
 	contentSizeForBW = 3
+	// metadataCacheTTLSecs is applicable only for those tests for which enableMetadataCache is true.
+	metadataCacheTTLSecs          = 10
```
meet2mky: Nit: Do you really need this? It's been used as local package constant which adds a layer of indirection, I would say we follow this pattern style which is more direct and readable https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/streaming_writes/buffer_size_test.go#L26

tools/integration_tests/rapid_appends/suites_test.go
```diff
@@ -0,0 +1,195 @@
...
+// Struct to store the details of a mount point
+type mountPoint struct {
+	rootDir     string // Root directory of the test folder, which contains mnt and gcsfuse.log.
+	mntDir      string // Directory where the GCS bucket is mounted. This is 'mnt' inside rootDir.
+	testDirPath string // Path to the 'RapidAppendsTest' directory inside mntDir.
+	logFilePath string // Path to the GCSFuse log file. This is gcsfuse.log inside rootDir.
+}
+
+const (
+	// TODO(b/432179045): `--write-global-max-blocks=-1` is needed right now because of a bug in global semaphore release.
```
meet2mky: Nit: We don't need this as well based on above comment.

tools/integration_tests/rapid_appends/suites_test.go
```diff
@@ -0,0 +1,195 @@
...
+func (t *RapidAppendsSuite) TearDownSuite() {
+	if t.T().Failed() {
+		setup.SetLogFile(t.primaryMount.logFilePath)
+		log.Printf("Saving primary mount log file %q ...", t.primaryMount.logFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+
+// SingleMountRapidAppendsSuite is the suite for rapid appends tests with a single mount,
+// i.e. primary mount for both reading and writing/appending.
```
meet2mky: Nit: appending

tools/integration_tests/rapid_appends/suites_test.go
```diff
@@ -0,0 +1,195 @@
...
+// RapidAppendsSuite is the base suite for rapid appends tests.
+type RapidAppendsSuite struct {
+	suite.Suite
+	primaryMount            mountPoint
+	fileName                string
+	fileContent             string
+	isSyncNeededAfterAppend bool
+	appendMountPath         string
+}
+
```
meet2mky, line 60: Nit: We can add a method similar to below.
```
func(mp *MountPoint) Setup(){
	mp.rootDir = setup.TestDir()
	mp.mntDir = setup.MntDir()
	mp.logFilePath = setup.LogFile()
	mp.testDirPath = path.Join(setup.MntDir(), testDirName)
}
```

Ant then in the SingleMountRapidAppensSuite: SetupSuite()

```
t.primaryMount.Setup()
```

In the dual mount setup suite()

```
t.primaryMount.Setup()
t.secondaryMount.Setup()
```

tools/integration_tests/rapid_appends/suites_test.go
```diff
@@ -0,0 +1,195 @@
...
+	if t.T().Failed() {
+		setup.SetLogFile(t.primaryMount.logFilePath)
+		log.Printf("Saving primary mount log file %q ...", t.primaryMount.logFilePath)
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+}
+
+// SingleMountRapidAppendsSuite is the suite for rapid appends tests with a single mount,
+// i.e. primary mount for both reading and writing/appending.
+type SingleMountRapidAppendsSuite struct {
```
meet2mky: This one is redundant as this is same as `RapidAppendsSuite`. In this the `RapidAppendsSuite` can work as single mount and `DualMountRapidAppendsSuite` can work as dual mount (which extends RapidAppendsSuite).. Setup or Teardown which are common can go to the `RapidAppendsSuite`. if they differ they can be added independently to both suites.

tools/integration_tests/rapid_appends/suites_test.go
```diff
@@ -0,0 +1,195 @@
...
+		setup.SaveGCSFuseLogFileInCaseOfFailure(t.T())
+	}
+	t.RapidAppendsSuite.TearDownSuite()
+}
+
+////////////////////////////////////////////////////////////////////////
+// Helpers
+////////////////////////////////////////////////////////////////////////
+
+func (t *RapidAppendsSuite) deleteUnfinalizedObject() {
```
meet2mky: Nit: This is already handled in TestMain and file names are random so it doesn't add any value, might as well remove this.

tools/integration_tests/rapid_appends/suites_test.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (t *RapidAppendsSuite) createUnfinalizedObject() {
+	t.fileName = fileNamePrefix + setup.GenerateRandomString(5)
+	// Create unfinalized object.
+	t.fileContent = setup.GenerateRandomString(unfinalizedObjectSize)
+	client.CreateUnfinalizedObject(ctx, t.T(), storageClient, path.Join(testDirName, t.fileName), t.fileContent)
+}
+
+func (t *RapidAppendsSuite) mountPrimaryMount(flags []string) {
```
meet2mky: Nit: Can we put these TearDown(Sub)Test?

meet2mky (commented):
Added few comments, PTAL.

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,222 +19,165 @@ import (
...
-			readMountPath: secondaryMntTestDirPath,
-			syncNeeded:    true, // Sync is required for writes to be visible on another mount.
-		},
+func (t *DualMountRapidAppendsSuite) TestAppendSessionInvalidatedByAnotherClientUponTakeover() {
+	const initialContent = "dummy content"
+	const appendContent = "appended content"
+	for _, flags := range [][]string{
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks},
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks, "--write-block-size-mb=1"},
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks, "--write-block-size-mb=1", fileCacheMaxSizeFlag, cacheDirFlagPrefix + getNewEmptyCacheDir(t.primaryMount.rootDir)},
```
vadlakondaswetha: why do we need all these combinations for this test

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,222 +19,165 @@ import (
...
-
-	// Read from back-door to validate that appended content is now visible on GCS.
-	expectedContent := initialContent + data
-	contentAfterClose, err := client.ReadObjectFromGCS(ctx, storageClient, path.Join(testDirName, t.fileName))
-	require.NoError(t.T(), err)
-	assert.Equal(t.T(), expectedContent, string(contentAfterClose))
+	for _, flags := range [][]string{
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks},
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks, "--write-block-size-mb=1"},
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks, "--write-block-size-mb=1", fileCacheMaxSizeFlag, cacheDirFlagPrefix + getNewEmptyCacheDir(t.primaryMount.rootDir)},
```
vadlakondaswetha: same question: why do we need to run these for these flag combinations

@anushka567 - Please check if you need to run these tests for these flag combinations?

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,222 +19,165 @@ import (
...
-////////////////////////////////////////////////////////////////////////
-
-func TestRapidAppendsSuite(t *testing.T) {
-	rapidAppendsSuite := new(RapidAppendsSuite)
-	suite.Run(t, rapidAppendsSuite)
+func (t *SingleMountRapidAppendsSuite) TestAppendsToFinalizedObjectNotVisibleUntilClose() {
+	const initialContent = "dummy content"
+	for _, flags := range [][]string{
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks},
+		{writeRapidAppendsEnableFlag, metadataCacheDisableFlag, infiniteWriteGlobalMaxBlocks, "--write-block-size-mb=1"},
```
vadlakondaswetha: same

tools/integration_tests/rapid_appends/setup_test.go
vadlakondaswetha: +1.

tools/integration_tests/rapid_appends/reads_after_appends_test.go
```diff
@@ -0,0 +1,109 @@
...
+		flags               []string
+	}{{
+		// all cache disabled
+		enableMetadataCache: false,
+		enableFileCache:     false,
+		flags:               []string{writeRapidAppendsEnableFlag, infiniteWriteGlobalMaxBlocks, metadataCacheDisableFlag},
+	}, {
+		enableMetadataCache: true,
+		enableFileCache:     false,
+		flags:               []string{writeRapidAppendsEnableFlag, infiniteWriteGlobalMaxBlocks, fmt.Sprintf("%s%v", metadataCacheEnableFlagPrefix, metadataCacheTTLSecs)},
```
vadlakondaswetha: nit: all this is not required. why to define constants for flags in the first place and then use fmt.sprintf?

tools/integration_tests/rapid_appends/suites_test.go
```diff
@@ -0,0 +1,195 @@
...
+var (
+	secondaryMountFlags = []string{writeRapidAppendsEnableFlag, infiniteWriteGlobalMaxBlocks}
+)
+
+// //////////////////////////////////////////////////////////////////////
+// Suite Setup and Teardown
+// //////////////////////////////////////////////////////////////////////
+
+// RapidAppendsSuite is the base suite for rapid appends tests.
+type RapidAppendsSuite struct {
```
vadlakondaswetha: remove rapid name in all suites

tools/integration_tests/rapid_appends/reads_after_appends_test.go
```diff
@@ -0,0 +1,109 @@
...
+////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////
+
+func (t *RapidAppendsSuite) TestAppendsAndReads() {
+	testCases := []struct {
+		name string
+	}{
+		{
+			name: "SequentialRead",
```
vadlakondaswetha, line 46: wont the read logic be different for random reads? why do you want to club everything in one testcase?

gargnitingoogle: The only difference for random read test will be in the replacement of the operations.ReadFile() call with a different function call with the same interface. This support is in the next PR (#3543 ). Sample at https://github.com/GoogleCloudPlatform/gcsfuse/blob/a500fedf0a7551cf2058337cdc6ff13056968566/tools/integration_tests/rapid_appends/reads_after_appends_test.go#L175 .

> why do you want to club everything in one testcase?

This is only to avoid code duplication.

tools/integration_tests/rapid_appends/suites_test.go
gargnitingoogle: Sure, will do.

tools/integration_tests/rapid_appends/reads_after_appends_test.go
gargnitingoogle: This is needed because metadataCacheTTLSecs is an integer. If I make metadataCacheTTLSecs a string or a part of metadataCacheEnableFlag, then I'd need to parse it again for the sleep to cause metadata cache entry to expire. So, I have to do some kind of sprintf or parse for it.

tools/integration_tests/rapid_appends/appends_test.go
gargnitingoogle: @anushka567 please respond.

tools/integration_tests/rapid_appends/appends_test.go
anushka567: We only require the last 2 flagsets. The first one should be skipped as test logic relies on blocksize being 1MiB.

tools/integration_tests/rapid_appends/suites_test.go
gargnitingoogle: outdated

tools/integration_tests/rapid_appends/suites_test.go
gargnitingoogle: done. Added the utilities for secondary mount. Not moving the secondary mount mount/unmount to the test functions right now because the tests aren't varying on secondaryMountFlags, so mount/unmount in individual tests will be unnecessary execution time cost. Will change it once tests need to run with different sets of secondary-mount-flags.

gargnitingoogle: Ok, I misunderstood the comment. Yeah, I myself had something in mind. Will do this.

tools/integration_tests/rapid_appends/suites_test.go
gargnitingoogle: No, this is not redundant. The three different suite classes serve different purposes. The suites determine which all tests will run. Any tests added to the base class, run for both single and dual mount configurations. The ones we add only to the single or dual-mount suite, run only for those configurations. Hence, we need all 3.

tools/integration_tests/rapid_appends/suites_test.go
gargnitingoogle: As discussed before, I am cleaning up the test objects created for a test at the end of that test, to facilitate easier debugging and less chances of interferences between the test (e.g. using different cache directories across tests). There is no harm in this, other than couple of lines of code.

tools/integration_tests/rapid_appends/appends_test.go
gargnitingoogle: @anushka567 suggested removing the first flag-set from each test, so I have removed them all.

tools/integration_tests/rapid_appends/appends_test.go
gargnitingoogle: @anushka567 suggested removing the first flag-set from each test, so I have removed them all.

tools/integration_tests/rapid_appends/appends_test.go
gargnitingoogle: @anushka567 suggested removing the first flag-set from each test, so I have removed them all.

tools/integration_tests/rapid_appends/reads_after_appends_test.go
gargnitingoogle: I've cleaned up this code.

tools/integration_tests/rapid_appends/setup_test.go
gargnitingoogle: moved.

tools/integration_tests/rapid_appends/suites_test.go
gargnitingoogle: @meet2mky I don't understand what that comment means. Are you saying this flag isn't needed? If so, why ? Or that that code comment TODO is not required ? If so, why ?

gargnitingoogle: never mind. Removed the next line of comment.

tools/integration_tests/rapid_appends/suites_test.go
gargnitingoogle: This I am doing explicitly in every test as not every test need to to unmount and mount in every sub-test, in which case this would cause unnecessary unmount and mo. We could do this with checks around whether primary mount already exists or not, but that's unnecessary complexity in test code. So I am keeping it explicit for the test writer for now. Let's optimize this later if needed.

tools/integration_tests/rapid_appends/suites_test.go
meet2mky: above comment in my review: This one https://github.com/GoogleCloudPlatform/gcsfuse/pull/3539#discussion_r2230104776

tools/integration_tests/rapid_appends/setup_test.go
meet2mky: What do you mean `moved.`?

gargnitingoogle:
@vadlakondaswetha @meet2mky please review. I have addressed all your comments.

tools/integration_tests/rapid_appends/setup_test.go
meet2mky: What I was saying is that we don't need all of these constants when we can simply use the literal string directly where it is required.

gargnitingoogle: > moved

Moved the definition to the exact test case where it's used.

> What I was saying is that we don't need all of these constants when we can simply use the literal string directly where it is required.

This particular variable `metadataCacheTTLSecs` is an integer, not a string, because I need it as an integer to pass to time.Sleep in the test TestAppendAndReads. This is why it has to be a variable.

gargnitingoogle: I have changed all the variables/constants used for gcsfuse flags to literals everywhere now.

tools/integration_tests/rapid_appends/reads_after_appends_test.go
```diff
@@ -0,0 +1,115 @@
...
+					for i := range numAppends {
+						sizeBeforeAppend := len(t.fileContent)
+						t.appendToFile(appendFileHandle, setup.GenerateRandomString(appendSize))
+						sizeAfterAppend := len(t.fileContent)
+
+						gotContent, err := operations.ReadFile(readPath)
+
+						require.NoError(t.T(), err)
+						readContent := string(gotContent)
+						if !scenario.enableMetadataCache || !t.isSyncNeededAfterAppend || (i == 0) {
```
vadlakondaswetha, line 101: can you add comments here on why are you validating full content in few cases and partially in other cases.

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,327 +19,322 @@ import (
...
-	err = wh.Close()
-	require.NoError(t.T(), err)
-
-	// Validate that appended content is visible in GCS.
-	expectedContent := initialContent + data
-	contentAfterClose, err := client.ReadObjectFromGCS(ctx, storageClient, path.Join(path.Join(testDirName, t.fileName)))
-	require.NoError(t.T(), err)
-	assert.Equal(t.T(), expectedContent, contentAfterClose)
+	for _, flags := range [][]string{
+		{"--write-experimental-enable-rapid-appends=true", "--metadata-cache-ttl-secs=0", "--write-global-max-blocks=-1", "--write-block-size-mb=1"},
```
vadlakondaswetha: Sorry, i am not abel to get why these write flows needs to be run against filecache on&off and metadata-cache-ttl=0

gargnitingoogle: @vadlakondaswetha this is how it was in the tests before my refactor also. @anushka567 needs to reply here.

tools/integration_tests/rapid_appends/reads_after_appends_test.go
gargnitingoogle: sure. will add.

gargnitingoogle: done

tools/integration_tests/rapid_appends/appends_test.go
gargnitingoogle: Removed metadata-cache, file-cache configs as discussed offline.

tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -19,327 +19,316 @@ import (
...
-
-				require.NoError(t.T(), err)
-				assert.Equal(t.T(), t.fileContent, string(gotContent))
-			}
-		})
+func (t *DualMountAppendsSuite) TestAppendSessionInvalidatedByAnotherClientUponTakeover() {
+	const initialContent = "dummy content"
+	const appendContent = "appended content"
+	for _, flags := range [][]string{
+		{"--write-experimental-enable-rapid-appends=true", "--write-block-size-mb=1"},
```
vadlakondaswetha, line 44: This is not looking correct. you shouldn't have this mount logic for every test. thats what the suite should do.

approving this PR for now. We can take up refactoring in the next PR.

vadlakondaswetha (approved):

tools/integration_tests/rapid_appends/appends_test.go
gargnitingoogle: > We can take up refactoring in the next PR.

Ok. Let's address this in next PR.

PR #3542, opened 2025-07-18
feat(buffered-read): Create a Block Queue Entry that can be used by Buffered Reader

vipnydav:
### Description
This PR introduces a `blockQueueEntry` for use in the buffered reader's Queue. This entry is designed to manage data blocks that are being prefetched and read.

The `blockQueueEntry` struct contains two main components:
- A `Block`, which is the destination for the downloaded data that will be read.
- A `cancelFunction`, which allows for the cancellation of a download task if the associated block is no longer needed

b/430755847

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/bufferedread/block_queue.go
+++ b/internal/bufferedread/block_queue.go
@@ -1,0 +1,43 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bufferedread
+
+import (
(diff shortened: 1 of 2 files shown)


internal/bufferedread/block_queue.go
```diff
@@ -0,0 +1,45 @@
...
+// newBlockQueueEntry creates a new entry for the block queue.
+func newBlockQueueEntry(b block.Block, t *DownloadTask) *blockQueueEntry {
+	return &blockQueueEntry{
+		Block: b,
+		Task:  t,
+	}
+}
+
+// blockQueue is a queue of blockQueueEntry instances.
+type blockQueue struct {
```
raj-prince: Why do we need this? we can directly use common.Queue[*blockQueueEntry].

vipnydav: The main idea for defining `blockqueue` struct was that we can add more specific methods to it later(e.g., for custom metrics) without being constrained to the methods of the `common.Queue` interface as `common.Queue[*blockQueueEntry]` would return an interface and we won't be able to add any new methods to it.

If you still want to remove the struct, I can remove it.

vipnydav: Removed the struct as per discussion

raj-prince (dismissed):

vipnydav:
Closing this PR as we are getting lint failure. Will merge these changes along with Buffered Reader.

PR #3555, opened 2025-07-21
feat(Notifier): Use notifier to invalidate stale dentry cache on clobber

aditimittal2003:
### Description

This change integrates `fuse.Notifier` to handle stale file entries when the experimental dentry cache is enabled.

When `experimental-enable-dentry-cache` is active, if a file is modified directly in GCS (clobbered), the kernel's dentry cache becomes stale. Subsequent read or write operations on that file would fail, typically with a "stale file handle" error.

This PR introduces logic to detect this specific failure (`FileClobberedError`) and uses the notifier to send an `InvalidateEntry` message to the kernel. This forces the kernel to evict the stale cache entry. As a result, the next operation on the same file will succeed, as gcsfuse will re-fetch the file's latest attributes.

This functionality is only enabled when the `experimental-enable-dentry-cache` flag is set to true.

b/423824806

### Testing details
1.  **Manual** - NA
2.  **Unit tests** - Added fs tests.
3.  **Integration tests** - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -1592,7 +1592,10 @@ func (fs *fileSystem) invalidateCachedEntry(childID fuseops.InodeID) error {
 		return fs.notifier.InvalidateEntry(fuseops.RootInodeID, path.Base(childInode.Name().LocalName()))
 	}
 
-	parentName := childName.ParentName()
+	parentName, err := childName.ParentName()
+	if err != nil {
+		return fmt.Errorf("invalidateCachedEntry: cannot find Parent name: %w", err)
+	}
 	childBase := path.Base(childName.LocalName())
 
 	var parentInodeID fuseops.InodeID
@@ -2832,9 +2835,9 @@ func (fs *fileSystem) ReadFile(
 	// making the kernel's dentry for this file stale. We use the notifier to
 	// invalidate this entry, providing feedback to the kernel about the dynamic
 	// content change and ensuring subsequent lookups fetch the correct metadata.
-	if err != nil && fs.newConfig.FileSystem.ExperimentalEnableDentryCache {
+	if fs.newConfig.FileSystem.ExperimentalEnableDentryCache {
 		var clobberedErr *gcsfuse_errors.FileClobberedError
-		if errors.As(err, &clobberedErr) {
+		if err != nil && errors.As(err, &clobberedErr) {
 			if invalidateErr := fs.invalidateCachedEntry(op.Inode); invalidateErr != nil {
 				err = fmt.Errorf("%w; additionally failed to invalidate entry: %w", err, invalidateErr)
 			}
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -133,6 +133,10 @@ type ServerConfig struct {
 
 	MetricHandle metrics.MetricHandle
 
+	// Notifier allows the file system to send invalidation messages to the FUSE
+	// kernel module. This enables proactive cache invalidation (e.g., for dentries)
+	// when underlying content changes, improving consistency while still leveraging
+	// kernel caching.
 	Notifier *fuse.Notifier
 }
 
@@ -501,6 +505,9 @@ type fileSystem struct {
 	// streaming writes are enabled.
 	globalMaxWriteBlocksSem *semaphore.Weighted
 
+	// notifier allows sending invalidation messages to the FUSE kernel module.
+	// It is used to invalidate the kernel's dentry cache,
+	// providing feedback to the kernel about dynamic content changes.
 	notifier *fuse.Notifier
 }
 
@@ -1562,12 +1569,17 @@ func (fs *fileSystem) lookupAndFetchAttributesForLocalFileEntriesPlus(parentName
 	return
 }
 
-// invalidateCachedEntry invalidates a specific directory entry in the kernel's cache,
-// corresponding to the child inode identified by childID.
+// invalidateCachedEntry sends a notification to the kernel to invalidate a stale
+// directory entry, ensuring consistency when file content changes dynamically.
+// It identifies the parent of the given child inode and sends a notification
+// to the kernel to remove the entry corresponding to the child's
+// name within that parent directory.
 //
 // LOCKS_EXCLUDED(fs.mu)
 func (fs *fileSystem) invalidateCachedEntry(childID fuseops.InodeID) error {
+	fs.mu.Lock()
 	childInode, ok := fs.inodes[childID]
+	fs.mu.Unlock()
 	if !ok {
 		return fmt.Errorf("inode with ID %d not found", childID)
 	}
@@ -2820,6 +2832,10 @@ func (fs *fileSystem) ReadFile(
 		op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb)
 	}
 
+	// A FileClobberedError indicates the underlying GCS object has changed,
+	// making the kernel's dentry for this file stale. We use the notifier to
+	// invalidate this entry, providing feedback to the kernel about the dynamic
+	// content change and ensuring subsequent lookups fetch the correct metadata.
 	if err != nil && fs.newConfig.FileSystem.ExperimentalEnableDentryCache {
 		var clobberedErr *gcsfuse_errors.FileClobberedError
 		if errors.As(err, &clobberedErr) {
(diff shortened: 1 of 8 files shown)


internal/fs/fs.go
```diff
@@ -1554,6 +1569,54 @@ func (fs *fileSystem) lookupAndFetchAttributesForLocalFileEntriesPlus(parentName
...
+// LOCKS_EXCLUDED(fs.mu)
+func (fs *fileSystem) invalidateCachedEntry(childID fuseops.InodeID) error {
+	fs.mu.Lock()
+	childInode, ok := fs.inodes[childID]
+	fs.mu.Unlock()
+	if !ok {
+		return fmt.Errorf("inode with ID %d not found", childID)
+	}
+
+	if childID == fuseops.RootInodeID {
```
raj-prince: nit: we can move this check on top. Also, good to mention about the RootInodeId in the function doc.

internal/fs/fs.go
```diff
@@ -1554,6 +1569,54 @@ func (fs *fileSystem) lookupAndFetchAttributesForLocalFileEntriesPlus(parentName
...
+// to the kernel to remove the entry corresponding to the child's
+// name within that parent directory.
+//
+// LOCKS_EXCLUDED(fs.mu)
+func (fs *fileSystem) invalidateCachedEntry(childID fuseops.InodeID) error {
+	fs.mu.Lock()
+	childInode, ok := fs.inodes[childID]
+	fs.mu.Unlock()
+	if !ok {
+		return fmt.Errorf("inode with ID %d not found", childID)
```
raj-prince:
```suggestion
		return fmt.Errorf("invalidateCachedEntry: inode with ID %d not found", childID)
```

raj-prince: similarly for other places.

internal/fs/fs.go
```diff
@@ -1554,6 +1569,54 @@ func (fs *fileSystem) lookupAndFetchAttributesForLocalFileEntriesPlus(parentName
...
+
+	if childID == fuseops.RootInodeID {
+		return fmt.Errorf("cannot invalidate root inode %d", childID)
+	}
+
+	childName := childInode.Name()
+	parentPath := path.Dir(childName.LocalName())
+	// If the parent path resolves to the current directory ".", it means the parent
+	// is the root of the file system.
+	if parentPath == "." || parentPath == "/" || parentPath == "" {
```
raj-prince: Shouldn't we return error for `parentPath == "/"` and `parentPath == ""`? When will this condition be triggered?

internal/fs/inode/name.go
```diff
@@ -123,3 +123,27 @@ func (name Name) IsDirectChildOf(parent Name) bool {
 	cleanDiff := strings.TrimSuffix(diff, "/")
 	return !strings.Contains(cleanDiff, "/")
 }
+
+// ParentName returns the Name of the parent directory of the current Name.
+func (name Name) ParentName() Name {
```
raj-prince: Please write unit test for this method.

internal/fs/notifier_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (t *NotifierTest) TearDownSuite() {
+	t.fsTest.TearDownTestSuite()
+}
+
+func (t *NotifierTest) TestWriteFileWithRootDirParent() {
+	filePath := path.Join(mntDir, "foo")
+	// Create a file in GCS.
+	_, err := storageutil.CreateObject(ctx, bucket, fileName, []byte("initial content"))
+	assert.NoError(t.T(), err)
```
raj-prince:
```suggestion
	require.NoError(t.T(), err)
```

internal/fs/notifier_test.go
```diff
@@ -0,0 +1,132 @@
...
+	// Stat file to cache its entry.
+	_, err = os.Stat(filePath)
+	assert.NoError(t.T(), err)
+	// Clobber the file in GCS. This changes the object's generation, making
+	// our file handle stale.
+	_, err = storageutil.CreateObject(ctx, bucket, fileName, []byte("modified"))
+	assert.NoError(t.T(), err)
+
+	// Attempt to write.
+	err = operations.WriteFile(filePath, "new data")
```
raj-prince: Could you please create this method in common/util? Actually, don't want this code dependent of integration_test package.

raj-prince (changes requested):
Overall looks good. Suggested couple of changes.

internal/fs/inode/name.go
aditimittal2003: Not needed. I removed this.

internal/fs/fs.go
aditimittal2003: Not needed. I removed this.

internal/fs/fs.go
```diff
@@ -2769,6 +2828,18 @@ func (fs *fileSystem) ReadFile(
 		op.Dst, op.BytesRead, err = fh.Read(ctx, op.Dst, op.Offset, fs.sequentialReadSizeMb)
 	}
 
+	// A FileClobberedError indicates the underlying GCS object has changed,
+	// making the kernel's dentry for this file stale. We use the notifier to
+	// invalidate this entry, providing feedback to the kernel about the dynamic
+	// content change and ensuring subsequent lookups fetch the correct metadata.
+	if err != nil && fs.newConfig.FileSystem.ExperimentalEnableDentryCache {
```
raj-prince: nit: can we move the error check inside 

if DentryCache {
    if err != nil && errors.Is(clobbered) {
       invalidate()
  }
}

internal/fs/fs.go
```diff
@@ -2769,6 +2828,18 @@ func (fs *fileSystem) ReadFile(
...
+		var clobberedErr *gcsfuse_errors.FileClobberedError
+		if errors.As(err, &clobberedErr) {
```
raj-prince: nit: see if we can use errors.Is?

internal/fs/fs.go
```diff
@@ -2815,7 +2886,19 @@ func (fs *fileSystem) WriteFile(
...
 	defer in.Unlock()
 	if err = fs.initBufferedWriteHandlerAndSyncFileIfEligible(ctx, in, fh.OpenMode()); err != nil {
-		return
+		// A FileClobberedError on write indicates the file was modified in GCS,
+		// making the kernel's dentry stale. By invalidating the cache
+		// entry, we ensure the filesystem corrects the inconsistency caused by this
+		// dynamic content change.
+		if fs.newConfig.FileSystem.ExperimentalEnableDentryCache {
+			var clobberedErr *gcsfuse_errors.FileClobberedError
+			if errors.As(err, &clobberedErr) {
```
raj-prince, line 2898: nit: see if we can use errors.Is()

internal/fs/inode/name.go
```diff
@@ -123,3 +123,27 @@ func (name Name) IsDirectChildOf(parent Name) bool {
...
+	if name.IsBucketRoot() {
+		panic("Root has no parent")
```
raj-prince: See if we can replace the panic with error. As we don't want to crash the whole app instead return input/output error for this particular operation.

internal/fs/fs.go
```diff
@@ -1554,6 +1569,53 @@ func (fs *fileSystem) lookupAndFetchAttributesForLocalFileEntriesPlus(parentName
...
+		return fs.notifier.InvalidateEntry(fuseops.RootInodeID, path.Base(childInode.Name().LocalName()))
+	}
+
+	parentName, err := childName.ParentName()
+	if err != nil {
+		return fmt.Errorf("invalidateCachedEntry: cannot find Parent name: %w", err)
+	}
+	childBase := path.Base(childName.LocalName())
+
+	var parentInodeID fuseops.InodeID
```
raj-prince:
```suggestion
	fs.mu.Lock()
	defer fs.mu.Unlock()
	
	var parentInodeID fuseops.InodeID
	// Check in all maps: implicit dirs → folders → generation-backed
```

internal/fs/inode/name.go
```diff
@@ -123,3 +124,27 @@ func (name Name) IsDirectChildOf(parent Name) bool {
...
+
+// ParentName returns the Name of the parent directory of the current Name.
+func (name Name) ParentName() (Name, error) {
+	if name.IsBucketRoot() {
+		return Name{}, errors.New("root has no parent")
+	}
+
+	objectName := strings.TrimSuffix(name.objectName, "/") // normalize for dir or file
+	lastSlash := strings.LastIndex(objectName, "/")
+
```
raj-prince:
```suggestion
	lastSlash := strings.LastIndex(objectName, "/")
	if lastSlash == -1 {
```

raj-prince (approved):

PR #3570, opened 2025-07-23
feat(Migrate Auth TPC): Add auth option while creating HTTP/gRPC client

Tulsishah:
### Description
- Add an authentication option for HTTP/gRPC clients, guarded by a feature flag. This new authentication method is compatible with both GCS and TPC.
- Code coverage increased
storage_handle.go 72% to 80%
client.go 80% to 95%

[b/432173154](https://b.corp.google.com/issues/432173154)

### Testing details
1. Manual - Tested on TPC
2. Unit tests - Added

### DIFF
--- a/internal/storage/storage_handle_test.go
+++ b/internal/storage/storage_handle_test.go
@@ -189,7 +189,7 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleWhenCustomEndpointIsNilA
 	assert.NotNil(testSuite.T(), handleCreated)
 }
 
-func (testSuite *StorageHandleTest) TestNewStorageHandleWhenKeyFileIsEmpty() {
+func (testSuite *StorageHandleTest) TestNewStorageHandleWhenAnonymousAccessTrue() {
 	sc := storageutil.GetDefaultStorageClientConfig(keyFile)
 	sc.KeyFile = ""
 	sc.AnonymousAccess = true


internal/storage/storage_handle_test.go
```diff
@@ -190,6 +190,7 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleWhenCustomEndpointIsNilA
 func (testSuite *StorageHandleTest) TestNewStorageHandleWhenKeyFileIsEmpty() {
 	sc := storageutil.GetDefaultStorageClientConfig()
 	sc.KeyFile = ""
+	sc.AnonymousAccess = true
```
Tulsishah, line 195: For reviewers: perviously it was default config due to that this test was working. 

Should I remove this test? I don't feel the worth as AnonymousAccess true will not even try to fetch credentials

vadlakondaswetha: Do we have a test case for anonymous access, if yes we can remove this one.

internal/storage/storage_handle_test.go
```diff
@@ -34,6 +34,8 @@ import (
 const invalidBucketName string = "will-not-be-present-in-fake-server"
 const projectID string = "valid-project-id"
 
+var keyFile = "storageutil/testdata/key.json"
```
vadlakondaswetha, line 37: if keyfile is same in all tests why did you add this a param?

Tulsishah: It's same file but path is different respective to file.
e.g. storage_handle_test.go file it is "storageutil/testdata/key.json" where as for client_test.go it is testdata/key.json

internal/storage/storage_handle_test.go
Tulsishah: Okay. Renamed the test.

vadlakondaswetha (approved):

PR #3577, opened 2025-07-23
feat(parallel random reads): Changes to enable parallel random read handling in random reader

abhishek10004:
This pull request introduces changes to enable parallel read handling in the randomReader. The core idea is to use a mutex to serialize access for RangeReader while allowing MultiRangeReader operations to proceed in parallel. 

The main approach is to determine read type without taking any locks and then allow parallel processing for MRD. For range reader, we would take a lock to synchronize access. Since reads wait on a lock for range reader, the reader type computation is done again before actually reading via range reader to ensure that the read should not be served by MRD.

### DIFF
--- a/internal/gcsx/multi_range_downloader_wrapper.go
+++ b/internal/gcsx/multi_range_downloader_wrapper.go
@@ -69,7 +69,7 @@ type MultiRangeDownloaderWrapper struct {
 	// Refcount is used to determine when to close the MultiRangeDownloader.
 	refCount int
 	// Mutex is used to synchronize access over refCount.
-	mu sync.Mutex
+	mu sync.RWMutex
 	// Holds the cancel function, which can be called to cancel the cleanup function.
 	cancelCleanup context.CancelFunc
 	// Used for waiting for timeout (helps us in mocking the functionality).
@@ -94,8 +94,8 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetMinObject() *gcs.MinObject {
 
 // GetRefCount returns current refcount.
 func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
-	mrdWrapper.mu.Lock()
-	defer mrdWrapper.mu.Unlock()
+	mrdWrapper.mu.RLock()
+	defer mrdWrapper.mu.RUnlock()
 	return mrdWrapper.refCount
 }
 
@@ -165,11 +165,15 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader() (err
 	if mrdWrapper.object == nil || mrdWrapper.bucket == nil {
 		return fmt.Errorf("ensureMultiRangeDownloader error: Missing minObject or bucket")
 	}
+
+	mrdWrapper.mu.RLock()
 	// Create the MRD if it does not exist.
 	// In case the existing MRD is unusable due to closed stream, recreate the MRD.
 	if mrdWrapper.Wrapped == nil || mrdWrapper.Wrapped.Error() != nil {
+		mrdWrapper.mu.RUnlock()
 		mrdWrapper.mu.Lock()
 		defer mrdWrapper.mu.Unlock()
+
 		// Checking if the mrdWrapper state is same after taking the lock.
 		if mrdWrapper.Wrapped == nil || mrdWrapper.Wrapped.Error() != nil {
 			var mrd gcs.MultiRangeDownloader
@@ -183,6 +187,8 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader() (err
 				mrdWrapper.Wrapped = mrd
 			}
 		}
+	} else {
+		mrdWrapper.mu.RUnlock()
 	}
 	return
 }


internal/gcsx/random_reader.go
```diff
@@ -329,60 +344,67 @@ func (rr *randomReader) ReadAt(
...
+	readInfo := rr.getReadInfo(offset)
+	reqReaderType := readerType(readInfo.readType, rr.bucket.BucketType())
+
+	if reqReaderType == RangeReader {
+		rr.mu.Lock()
+		expectedOffset := rr.expectedOffset.Load()
+		if readInfo.expectedOffset != expectedOffset {
+			readInfo = rr.getReadInfo(offset)
+			reqReaderType = readerType(readInfo.readType, rr.bucket.BucketType())
 		}
```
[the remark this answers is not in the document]
abhishek10004: 1. Made changes to ensure seek is not recorded twice for same read request
2. We can't do that because then random reads would need to wait till seq read call finishes.

internal/gcsx/random_reader.go
```diff
@@ -329,60 +345,72 @@ func (rr *randomReader) ReadAt(
...
-	// clean it up and throw it away.
-	// We will also clean up the existing reader if it can't serve the entire request.
-	dataToRead := math.Min(float64(offset+int64(len(p))), float64(rr.object.Size))
-	if rr.reader != nil && (rr.start != offset || int64(dataToRead) > rr.limit) {
-		rr.closeReader()
-		rr.reader = nil
-		rr.cancel = nil
-	}
+		if reqReaderType == MultiRangeReader {
+			rr.mu.Unlock()
```
vadlakondaswetha, line 365: you can add this line also under this if clause:
objectData.Size, err = rr.readFromMultiRangeReader(ctx, p, offset, offset+int64(len(p)), TimeoutForMultiRangeRead)

then extract the rest of the logic into seperate method - readFromRangeReader or something like that.

internal/gcsx/random_reader.go
```diff
@@ -516,26 +544,58 @@ func (rr *randomReader) startRead(start int64, end int64) (err error) {
...
-		return
+// isSeekNeeded determines if the current read at `offset` should be considered a
+// seek, given the previous read pattern & the expected offset.
+func isSeekNeeded(readType, offset, expectedOffset int64) bool {
+	if expectedOffset == 0 {
+		return false
+	} else if readType == metrics.ReadTypeRandom {
+		return offset != expectedOffset
+	} else if readType == metrics.ReadTypeSequential {
+		return offset < expectedOffset || offset > expectedOffset+maxReadSize
```
vadlakondaswetha, line 535: earlier we are only checking if the existing reader can be forwarded by 8MB to serve the current read. now we are saying any reads which falls within 8MB offset is sequential.

internal/gcsx/random_reader.go
```diff
@@ -177,6 +178,15 @@ type randomReader struct {
 	// Specifies the next expected offset for the reads. Used to distinguish between
 	// sequential and random reads.
 	expectedOffset atomic.Int64
+
+	// To synchronize reads served from range reader.
+	mu sync.Mutex
```
vadlakondaswetha, line 185: Are the changes to remove fileHandle will be in seperate PR?

internal/gcsx/random_reader.go
```diff
@@ -329,60 +345,72 @@ func (rr *randomReader) ReadAt(
...
+			objectData.Size, err = rr.readFromRangeReader(ctx, p, offset, end, readInfo.readType)
+			return
+		}
 	}
 
-	readerType := readerType(rr.readType.Load(), offset, end, rr.bucket.BucketType())
-	if readerType == RangeReader {
-		objectData.Size, err = rr.readFromRangeReader(ctx, p, offset, end, rr.readType.Load())
-		return
+	if reqReaderType == MultiRangeReader {
```
vadlakondaswetha, line 382: can you move this to the top before checking if reqReaderType == Rangereader. somethign like this

if reqReaderType == MultiRangeReader {
 // read from multirangeReader
 return
}

// add your new logic here
rr.mu.Lock()
expectedOffset := rr.expectedOffset.Load()

		// Calculating reader type again for zonal buckets in case another read has been served
		// since last computation. This is to ensure that we don't use range reader incorrectly
		// when MRD should've been used.
		if rr.bucket.BucketType().Zonal && readInfo.expectedOffset != expectedOffset {
			readInfo = rr.getReadInfo(offset, readInfo.seekRecorded)
			reqReaderType = readerType(readInfo.readType, rr.bucket.BucketType())
		}

internal/gcsx/random_reader.go
```diff
@@ -329,60 +345,72 @@ func (rr *randomReader) ReadAt(
...
-		rr.reader = nil
-		rr.cancel = nil
-	}
+		if reqReaderType == MultiRangeReader {
+			rr.mu.Unlock()
+		} else {
+			defer rr.mu.Unlock()
+
+			// Check first if we can read using existing reader. if not, determine which
+			// api to use and call gcs accordingly.
```
vadlakondaswetha: for readability, please extract this into new method.

internal/gcsx/random_reader.go
abhishek10004: yes, they will be in a separate PR

internal/gcsx/random_reader.go
abhishek10004: yes, if the previous read was seq, then we are just checking whether it falls in the next 8MB range to determine whether current read is sequential or not.

internal/gcsx/random_reader.go
abhishek10004: I kept range reader at top because I check the read type inside seq block (after taking a lock as well) and if the read type turns out to be random, I fall back to MRD (by updating read type and bailing out of this block).

internal/gcsx/random_reader.go
abhishek10004: I've refactored code for better readability

internal/gcsx/random_reader.go
```diff
@@ -53,6 +54,8 @@ const minSeeksForRandom = 2
 // TODO(b/385826024): Revert timeout to an appropriate value
 const TimeoutForMultiRangeRead = time.Hour
 
+var FallbackToNewRangeReader = errors.New("fallback to new rnage reader is required")
```
vadlakondaswetha: type: rnage -> range

internal/gcsx/random_reader.go
vadlakondaswetha: if we are doing random reads in a small file, it might be a problem/

internal/gcsx/random_reader.go
```diff
@@ -311,6 +323,12 @@ func (rr *randomReader) ReadAt(
 	if offset >= int64(rr.object.Size) {
 		err = io.EOF
 		return
+	} else if offset < 0 {
+		err = fmt.Errorf(
+			"illegal offset %d for %d-byte object",
```
meet2mky:
```
"illegal offset %d for %d-byte object",
```
is the `-`  is typo?

internal/gcsx/random_reader_stretchr_test.go
```diff
@@ -87,95 +86,174 @@ func (t *RandomReaderStretchrTest) TearDownTest() {
 	t.rr.Destroy()
 }
 
-func (t *RandomReaderStretchrTest) Test_ReadInfo() {
```
raj-prince, line 90: Should we add concurrent test (multiple go routines reading random offset, concurrently) to detect any race incase? Ref: https://github.com/GoogleCloudPlatform/gcsfuse/blob/72d85da9151cd23d1236a69918ab29494e7c3610/internal/cache/lru/lru_test.go#L316

abhishek10004: Agreed. Will add a test for concurrency.

internal/gcsx/random_reader.go
```diff
@@ -577,14 +608,60 @@ func (rr *randomReader) getReadInfo(
...
+func readerType(readType int64, bucketType gcs.BucketType) ReaderType {
+	if readType == metrics.ReadTypeRandom && bucketType.Zonal {
 		return MultiRangeReader
 	}
 	return RangeReader
 }
 
+// skipBytes attempts to advance the reader position to the given offset without
+// discarding the existing reader.
+func (rr *randomReader) skipBytes(offset int64) {
```
raj-prince, line 630: Can we explicitly add - calling this (and below) methods require mu.Lock()?

internal/gcsx/random_reader.go
```diff
@@ -516,26 +516,58 @@ func (rr *randomReader) startRead(start int64, end int64) (err error) {
...
-	if start < 0 || uint64(start) > rr.object.Size || size < 0 {
-		err = fmt.Errorf(
-			"range [%d, %d) is illegal for %d-byte object",
-			start,
-			start+size,
-			rr.object.Size)
-		return
+// isSeekNeeded determines if the current read at `offset` should be considered a
+// seek, given the previous read pattern & the expected offset.
+func isSeekNeeded(readType, offset, expectedOffset int64) bool {
```
raj-prince, line 525:
```suggestion
func isSeekNeeded(readType, offset, expectedOffset int64) bool {
	if expectedOffset == 0 {
		return false
	}
	
	if readType == metrics.ReadTypeRandom {
		return offset != expectedOffset
	}
	
	if readType == metrics.ReadTypeSequential {
		return offset < expectedOffset || offset > expectedOffset+maxReadSize
	}
```

Ref: https://google.github.io/styleguide/go/decisions#indent-error-flow

raj-prince, line 525: Should we write unit test for this?

internal/gcsx/random_reader.go
```diff
@@ -329,60 +347,42 @@ func (rr *randomReader) ReadAt(
...
+
+			// Check first if we can read using existing reader. if not, create a new range reader
+			objectData.Size, err = rr.readFromExistingRangeReader(ctx, p, offset)
+			if errors.Is(err, FallbackToNewRangeReader) {
+				// reader does not exist and need to be created, get the end offset.
+				end := rr.getEndOffset(offset)
+				objectData.Size, err = rr.readFromRangeReader(ctx, p, offset, end, readInfo.readType)
+			}
+			return
+		}
```
raj-prince, line 379:
```suggestion
		if reqReaderType == RangeReader {
			defer rr.mu.Unlock()

			// Check first if we can read using existing reader. if not, create a new range reader
			objectData.Size, err = rr.readFromExistingRangeReader(ctx, p, offset)
			if errors.Is(err, FallbackToNewRangeReader) {
				// reader does not exist and need to be created, get the end offset.
				end := rr.getEndOffset(offset)
				objectData.Size, err = rr.readFromRangeReader(ctx, p, offset, end, readInfo.readType)
			}
			return
		}
		
		rr.mu.Unlock()
```

internal/gcsx/random_reader.go
abhishek10004: I am not seeing any regression (checked 4K random reads in 1M & 10M file on regional & 4K random reads on 10M file on ZB).
Why do you expect issues in this particular scenario?

internal/gcsx/random_reader.go
abhishek10004: Done

internal/gcsx/random_reader.go
abhishek10004: Done

internal/gcsx/random_reader.go
vadlakondaswetha: For a file of size 8MB.
First read is always sequential and then every read will fall within 8MB range, so it becomes sequential always. The problem is not sequential or random you end up fetching more data from GCS always.

vadlakondaswetha (approved):

internal/gcsx/random_reader.go
abhishek10004: I wanted to keep the unlock close to the lock statement so that it's easy to reason about it.

internal/gcsx/random_reader.go
abhishek10004: Discussed offline with Swetha.

internal/gcsx/random_reader.go
abhishek10004: Done

vadlakondaswetha (approved):

PR #3579, opened 2025-07-24
style: removing unused code - part2

vadlakondaswetha:
Removing unused code as suggested by golangci-lint.

### Testing details
1. Manual - Manually verified mounting and few operations
2. Unit tests - github actions will take care of this
3. Integration tests - github actions will take care of this

### DIFF
--- a/internal/storage/caching/integration_test.go
+++ b/internal/storage/caching/integration_test.go
@@ -103,7 +103,7 @@ func (t *IntegrationTest) CreateInsertsIntoCache() {
 }
 
 func (t *IntegrationTest) StatInsertsIntoCache() {
-	const name = "taco"
+	const name = "foo"
 	var err error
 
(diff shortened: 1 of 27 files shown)


internal/storage/caching/integration_test.go
```diff
@@ -103,7 +103,7 @@ func (t *IntegrationTest) CreateInsertsIntoCache() {
 }
 
 func (t *IntegrationTest) StatInsertsIntoCache() {
-	const name = "taco"
+	const name = "foo"
```
raj-prince, line 106: Was this change required?

vadlakondaswetha: yeah, otherwise it is complaining of a function which gets foo as argument always. so i just changed one test to invoke with different name.

PR #3594, opened 2025-07-28
test(dentry_cache): add integration tests for delete operations

aditimittal2003:
### Description
This change introduces new integration tests for the experimental dentry cache feature, focusing on scenarios involving file deletion. These tests ensure that the cache behaves as expected when files are removed or modified externally.

The following test cases have been added:

- `TestDeleteFileWhenFileIsClobbered`: Verifies that deleting a file from the mount point succeeds even after the corresponding object has been clobbered (overwritten) in GCS.
- `TestDeleteFileWithDentryCacheEnabled`: Confirms that after a file is deleted directly from GCS, a subsequent read correctly triggers a notifier to invalidate the stale dentry cache entry.
- `TestStatWhenFileIsDeletedDirectlyFromGCS`: Checks that stat returns cached data for a file deleted directly from GCS until the cache TTL expires, after which it correctly returns error as entry does not exist.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - Added new integration tests and verified they pass. The run_tests_mounted_directory.sh script has been updated to execute these new tests.


tools/integration_tests/dentry_cache/stat_test.go
```diff
@@ -72,6 +72,29 @@ func (s *statWithDentryCacheEnabledTest) TestStatWithDentryCacheEnabled(t *testi
...
+	// Delete the object directly from GCS.
+	objectName := path.Join(testDirName, testFileName)
+	require.Nil(t, client.DeleteObjectOnGCS(ctx, storageClient, objectName))
+
+	// Stat again, it should give old cached attributes rather than giving not found error.
+	fileInfo, err := os.Stat(filePath)
+
+	assert.Nil(t, err)
+	assert.Equal(t, int64(initialContentSize), fileInfo.Size())
+	// Wait until entry expires in cache.
```
raj-prince: This assumes gcsfuse is mounted with 1 sec as metadata-cache-ttl. 
Option 1: please clearly mention in the comment.
Option 2: use a ttl variable and use that while mounting and waiting. I am fine with any of them.

raj-prince (approved):
LGTM, one minor comment.

PR #3599, opened 2025-07-28
perf(metrics): Auto-generate metrics code to be more efficient

kislaykishore:
### Description
Code-generate metrics implementation in a separate package as part of build. We'll eventually get rid of this package and move the implementation to use the metrics package itself. But using a different package in order to keep the auto-generated and hand-written code separate.

This implementation is supposed to be much faster and efficient that the current implementation. However, we aren't switching to this yet. In a subsequent PR, we will verify the correctness of this code-generation by running the existing tests against this implementation.

b/434616359
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

No


optimizedmetrics/metrics.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - "Parallel"
+    - "Random"
+    - "Sequential"
+
+- metric-name: "file_cache/read_latencies"
+  description: "The cumulative distribution of the file cache read latencies along with cache hit - true/false."
+  type: "int_histogram"
+  unit: "us"
+  boundaries:
+  - 1
```
charith87, line 18: how are these boundaries decided?

kislaykishore: This is currently static and the same for all distribution metrics. I am not changing it as part of this effort.  We will explore moving to exponential histograms and we already have an FR for that.

charith87: Yea exponential makes sense for future. Powers of 1.2 are used extensively as i can see. I am just checking if the below numbers are if based on some assumptions and needs this many or could trim something

optimizedmetrics/otel_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+		record = histogramRecord{ctx: ctx, instrument: o.fileCacheReadLatencies, value: latency.Microseconds(), attributes: fileCacheReadLatenciesCacheHitFalseAttrSet}
+	}
+
+	select {
+	case o.ch <- histogramRecord{instrument: record.instrument, value: record.value, attributes: record.attributes, ctx: ctx}: // Do nothing
+	default: // Unblock writes to channel if it's full.
+	}
+}
+
+func NewOTelMetrics(ctx context.Context, workers int, bufferSize int) (*otelMetrics, error) {
```
charith87, line 84: *nit = CreateOTelMetrics

optimizedmetrics/otel_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	ch := make(chan histogramRecord, bufferSize)
+	var wg sync.WaitGroup
+	for range workers {
```
charith87, line 87: where is the number of workers going to be set?

optimizedmetrics/otel_metrics.go
kislaykishore: go/go-style/decisions#repetition - Golang convention is to use New.

optimizedmetrics/otel_metrics.go
kislaykishore: This will be fetched from a flag and set [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/72d85da9151cd23d1236a69918ab29494e7c3610/cmd/legacy_main.go#L399).

optimizedmetrics/metrics.yaml
kislaykishore: Just maintaining status quo. Here is the FR: b/432227120

charith87 (approved):

charith87:
Approving. Taking into consideration that approach is already approved in ZB stream and Perf (CPU/Memory) will be measured before making a decision to use this. cc @vadlakondaswetha

PR #3602, opened 2025-07-28
feat(bufferedread): Implement ReadAt method of Buffered Reader

vipnydav:
### Description
This PR implements the `ReadAt` method on BufferedReader to support reads from GCS objects, satisfying the `gcsx.Reader` interface.

The method leverages the existing prefetching mechanism to read from in-memory blocks. Key features of the implementation include:

- Random Read Detection: Identifies random read patterns and falls back to a simpler GCS reader if the configured threshold for random seeks is exceeded.
- Efficient Seek Handling: Discards stale blocks from the queue on seeks to ensure the reader starts from the correct offset.
- Continuous Prefetching: Triggers new prefetch operations as data is consumed to keep the read pipeline full and minimize latency.

https://b.corp.google.com/issues/432169239

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -168,21 +168,21 @@ func (p *BufferedReader) prepareQueueForOffset(offset int64) {
 //
 // The read is satisfied by reading from in-memory blocks that are prefetched
 // in the background. The core logic is as follows:
-// 1. Detect if the read pattern is random. If so, and if the random read
-//    threshold is exceeded, it returns a FallbackToAnotherReader error.
-// 2. Prepare the internal block queue by discarding any stale blocks from the
-//    head of the queue that are before the requested offset.
-// 3. If the queue becomes empty (e.g., on a fresh read or a large seek), it
-//    initiates a "fresh start" to prefetch blocks starting from the current
-//    offset.
-// 4. It then enters a loop to fill the destination buffer:
-//    a. It waits for the block at the head of the queue to be downloaded.
-//    b. If the download failed or was cancelled, it returns an appropriate error.
-//    c. If successful, it copies data from the downloaded block into the buffer.
-//    d. If a block is fully consumed, it is removed from the queue, and a new
-//       prefetch operation is triggered to keep the pipeline full.
-// 5. The loop continues until the buffer is full, the end of the file is
-//    reached, or an error occurs.
+//  1. Detect if the read pattern is random. If so, and if the random read
+//     threshold is exceeded, it returns a FallbackToAnotherReader error.
+//  2. Prepare the internal block queue by discarding any stale blocks from the
+//     head of the queue that are before the requested offset.
+//  3. If the queue becomes empty (e.g., on a fresh read or a large seek), it
+//     initiates a "fresh start" to prefetch blocks starting from the current
+//     offset.
+//  4. It then enters a loop to fill the destination buffer:
+//     a. It waits for the block at the head of the queue to be downloaded.
+//     b. If the download failed or was cancelled, it returns an appropriate error.
+//     c. If successful, it copies data from the downloaded block into the buffer.
+//     d. If a block is fully consumed, it is removed from the queue, and a new
+//     prefetch operation is triggered to keep the pipeline full.
+//  5. The loop continues until the buffer is full, the end of the file is
+//     reached, or an error occurs.
 func (p *BufferedReader) ReadAt(ctx context.Context, inputBuf []byte, off int64) (gcsx.ReaderResponse, error) {
 	resp := gcsx.ReaderResponse{DataBuf: inputBuf}
 	reqID := uuid.New()
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -121,13 +121,15 @@ func (p *BufferedReader) handleRandomRead(offset int64) error {
 		return gcsx.FallbackToAnotherReader
 	}
 
-	isRandom := p.isRandomRead(offset)
-	if !isRandom {
+	if !p.isRandomSeek(offset) {
 		return nil
 	}
 
 	p.randomSeekCount++
 
+	// When a random seek is detected, the prefetched blocks in the queue become
+	// irrelevant. We must clear the queue, cancel any ongoing downloads, and
+	// release the blocks back to the pool.
 	for !p.blockQueue.IsEmpty() {
 		entry := p.blockQueue.Pop()
 		entry.cancel()
@@ -146,8 +148,8 @@ func (p *BufferedReader) handleRandomRead(offset int64) error {
 	return nil
 }
 
-// isRandomRead checks if the read for the given offset is random or not.
-func (p *BufferedReader) isRandomRead(offset int64) bool {
+// isRandomSeek checks if the read for the given offset is random or not.
+func (p *BufferedReader) isRandomSeek(offset int64) bool {
 	if p.blockQueue.IsEmpty() {
 		return offset != 0
 	}
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -18,7 +18,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"time"
 
+	"github.com/google/uuid"
 	"github.com/googlecloudplatform/gcsfuse/v3/common"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/block"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/gcsx"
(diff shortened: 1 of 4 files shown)


internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,191 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+	var bytesRead int
+	var err error
+
+	logger.Tracef("%.13v <- ReadAt(offset=%d, len=%d, blockIdx=%d)", reqID, off, len(inputBuf), blockIdx)
+
+	if off >= int64(p.object.Size) {
+		err = io.EOF
+		return resp, err
+	}
+	if len(inputBuf) == 0 {
```
raj-prince, line 231: nit: empty line before.

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,191 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+		if offset < start || offset >= end {
+			isRandom = true
+		}
+	}
+
+	if isRandom {
+		p.randomSeekCount++
+	}
+
+	if p.randomSeekCount > p.config.RandomReadsThreshold {
```
raj-prince, line 142: Once random is always random at least as part of phase 1.  Can we release the block (in case of non-empty queue)?

Also, can we keep this check on top?

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,191 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+		block := entry.block
+		blockStart := block.AbsStartOff()
+		blockEnd := blockStart + block.Cap()
+
+		if offset < blockStart || offset >= blockEnd {
+			// Offset is either before or beyond this block – discard.
+			p.blockQueue.Pop()
+			entry.cancel()
+
+			if _, waitErr := block.AwaitReady(context.Background()); waitErr != nil && !errors.Is(waitErr, context.Canceled) {
```
raj-prince: Why are we filtering context.Canceled error? Shouldn't we log the warning for any error coming from AwaitReady?

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,191 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+//     offset.
+//  4. It then enters a loop to fill the destination buffer:
+//     a. It waits for the block at the head of the queue to be downloaded.
+//     b. If the download failed or was cancelled, it returns an appropriate error.
+//     c. If successful, it copies data from the downloaded block into the buffer.
+//     d. If a block is fully consumed, it is removed from the queue, and a new
+//     prefetch operation is triggered to keep the pipeline full.
+//  5. The loop continues until the buffer is full, the end of the file is
+//     reached, or an error occurs.
+func (p *BufferedReader) ReadAt(ctx context.Context, inputBuf []byte, off int64) (gcsx.ReaderResponse, error) {
```
raj-prince, line 214: Do we know what happens, if len of inputBuf[] is more than the object size while reading tail content?

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,191 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+		n, readErr := blk.ReadAt(inputBuf[bytesRead:], relOff)
+		bytesRead += n
+		off += int64(n)
+
+		if readErr != nil && !errors.Is(readErr, io.EOF) {
+			err = fmt.Errorf("ReadAt: block.ReadAt failed: %w", readErr)
+			break
+		}
+
+		if off >= int64(p.object.Size) {
```
raj-prince, line 295: With this, it seems we will return io.EOF error for the tail content?

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,191 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+			err = io.EOF
+			break
+		}
+
+		if off >= blk.AbsStartOff()+blk.Cap() {
```
raj-prince: Shouldn't we use blk.Size() here? Although both will be same here.

internal/bufferedread/buffered_reader_test.go
```diff
@@ -60,6 +61,7 @@ type BufferedReaderTest struct {
 ////////////////////////////////////////////////////////////////////////
 
 // createFakeReader returns a FakeReader with deterministic, non-zero content.
+// It's used for mocking GCS object reads.
```
raj-prince: Can we re-use FakeReaderWithOffset for this?

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,191 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+			logger.Tracef("%.13v -> ReadAt OK (offset=%d, len=%d, read=%d) (%v)", reqID, initOff, len(inputBuf), bytesRead, dur)
+		}
+	}()
+
+	if err = p.handleRandomRead(off); err != nil {
+		err = fmt.Errorf("ReadAt: handleRandomRead failed: %w", err)
+		return resp, err
+	}
+
+	prefetchTriggered := false
```
raj-prince, line 249: Just thinking out loud here. Currently, we will be triggering only once for a given ReadAt call, this assumes size of ReadAt() < block-size.

What if size of ReadAt() > block-size: in this case one ReadAt() will basically evict many small blocks? Although this configuration is not recommended, but good to handle if we can. Ideally, we should trigger prefetch for every block eviction. Or may be we can ensure block-size can't be less than 1 MiB.

internal/bufferedread/buffered_reader_test.go
```diff
@@ -634,5 +678,352 @@ func (t *BufferedReaderTest) TestPrefetchLimitedByAvailableSlots() {
...
+
+	// Read from a reader where the next block download was cancelled.
+	_, err = reader.ReadAt(t.ctx, make([]byte, 10), 0)
+
+	assert.ErrorIs(t.T(), err, context.Canceled)
+	status, err := b.AwaitReady(t.ctx)
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), block.BlockStateDownloadFailed, status.State)
+	assert.ErrorIs(t.T(), status.Err, context.Canceled)
+
```
raj-prince: nit: remove empty line.

internal/bufferedread/buffered_reader_test.go
```diff
@@ -634,5 +678,352 @@ func (t *BufferedReaderTest) TestPrefetchLimitedByAvailableSlots() {
...
+	// Read from a reader where the next block has failed to download.
+	_, err = reader.ReadAt(t.ctx, make([]byte, 10), 0)
+
+	assert.ErrorIs(t.T(), err, downloadError)
+	status, err := b.AwaitReady(t.ctx)
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), block.BlockStateDownloadFailed, status.State)
+	assert.ErrorIs(t.T(), status.Err, downloadError)
+
```
raj-prince: nit: remove empty line.

internal/bufferedread/buffered_reader_test.go
```diff
@@ -634,5 +678,352 @@ func (t *BufferedReaderTest) TestPrefetchLimitedByAvailableSlots() {
...
+	// Read from a reader where the next block is in an unexpected state.
+	_, err = reader.ReadAt(t.ctx, make([]byte, 10), 0)
+
+	assert.ErrorContains(t.T(), err, "unexpected block state")
+	status, err := b.AwaitReady(t.ctx)
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), block.BlockStateInProgress, status.State)
+	assert.Nil(t.T(), status.Err)
+
```
raj-prince: nit: no empty line.

internal/bufferedread/buffered_reader_test.go
```diff
@@ -634,5 +678,352 @@ func (t *BufferedReaderTest) TestPrefetchLimitedByAvailableSlots() {
...
+	// Read from a block that is already downloaded and in the queue.
+	resp, err := reader.ReadAt(t.ctx, buf, 0)
+
+	assert.NoError(t.T(), err)
+	assert.Equal(t.T(), 5, resp.Size)
+	assert.Equal(t.T(), "abcde", string(buf))
+	assert.False(t.T(), reader.blockQueue.IsEmpty())
+}
+
+func (t *BufferedReaderTest) TestReadAtExactlyToEndOfFile() {
```
raj-prince, line 913: Can we also test the case - where input buffer exceeds the end of file?

internal/bufferedread/buffered_reader.go
vipnydav: Agreed, we should always clear the queue on a random read, even if we’re past the threshold.

But moving the threshold check to the top could cause inconsistent behavior: same `randomSeekCount` might allow one read and fail another, depending on when the counter was incremented. Keeping the check after isRandom keeps it predictable.

internal/bufferedread/buffered_reader.go
vipnydav: I was not logging for expected error where the context is explicitly canceled by the buffered reader due to random reads or other reasons as done in `Destroy` method.
Do you still want me to log error for it?

internal/bufferedread/buffered_reader.go
vipnydav: Yes

internal/bufferedread/buffered_reader.go
vipnydav: The check in https://github.com/GoogleCloudPlatform/gcsfuse/pull/3602/files#diff-42569fce5300eff209ea0f3673e02549844ac3a590ec8a46193c40f26ed98d53R272
will return `io.EOF` if inputBuf[] size is more thn object size while reading tail content.

I'll add a unit test for it.

internal/bufferedread/buffered_reader.go
vipnydav: I don't think keeping block size less than 1MB would make sense. So I'll add in checkInvariant to check for prefetchBlockSize > 1MB.

internal/bufferedread/buffered_reader.go
vipnydav: Yeah, I think blk.Size() would be better here. I'll update it.

internal/bufferedread/buffered_reader_test.go
vipnydav: Sure, removed FakeReader and used FakeReaderWithOffset everywhere.

internal/bufferedread/buffered_reader.go
vipnydav: As discussed offline, this error can only come from timeout exceeded or other errors which are not due to random reads. So, we should log all the errors.

internal/bufferedread/buffered_reader_test.go
vipnydav: Yeah, added a unit test for it.

internal/bufferedread/buffered_reader.go
```diff
@@ -418,6 +434,11 @@ func (p *BufferedReader) CheckInvariants() {
 		panic(fmt.Sprintf("BufferedReader: PrefetchBlockSizeBytes must be positive, but is %d", p.config.PrefetchBlockSizeBytes))
 	}
 
+	// The prefetch block size must be at least 1 MiB.
+	if p.config.PrefetchBlockSizeBytes < 1<<20 {
```
raj-prince: nit: instead create a const MiB if not already.

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +106,207 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+// reader should be used.
+func (p *BufferedReader) handleRandomRead(offset int64) error {
+	// Exit early if we have already decided to fall back to a GCS reader. This
+	// avoids re-evaluating the read pattern on every call when the random read
+	// threshold has been met.
+	if p.randomSeekCount > p.config.RandomReadsThreshold {
+		return gcsx.FallbackToAnotherReader
+	}
+
+	var isRandom bool
```
raj-prince: Extract a method to check if the read for a given offset is random or not.

raj-prince (changes requested):
Suggest a minor refactoring.

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +108,210 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+// reader should be used.
+func (p *BufferedReader) handleRandomRead(offset int64) error {
+	// Exit early if we have already decided to fall back to a GCS reader. This
+	// avoids re-evaluating the read pattern on every call when the random read
+	// threshold has been met.
+	if p.randomSeekCount > p.config.RandomReadsThreshold {
+		return gcsx.FallbackToAnotherReader
+	}
+
+	isRandom := p.isRandomRead(offset)
```
raj-prince:
```suggestion
	if !p.isRandomRead(offset) {
		return nil
	}
```

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +108,210 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+		logger.Tracef("Too many random reads for object %q (count: %d, threshold: %d); falling back to GCS reader.",
+			p.object.Name, p.randomSeekCount, p.config.RandomReadsThreshold)
+		return gcsx.FallbackToAnotherReader
+	}
+
+	return nil
+}
+
+// isRandomRead checks if the read for the given offset is random or not.
+func (p *BufferedReader) isRandomRead(offset int64) bool {
```
raj-prince: nit: isRandomSeek makes more sense.

internal/bufferedread/buffered_reader.go
```diff
@@ -103,6 +108,210 @@ func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *Buffere
...
+	}
+
+	isRandom := p.isRandomRead(offset)
+	if !isRandom {
+		return nil
+	}
+
+	p.randomSeekCount++
+
+	for !p.blockQueue.IsEmpty() {
```
raj-prince, line 133: nit: please add comment, why is this required?

raj-prince (approved):
few nits.

raj-prince (approved):

PR #3604, opened 2025-07-29
chore(flags): Add flag-name validations

kislaykishore:
### Description
Ensure that flag-names conform to the following rules:
* It's not empty
* It only consists of lower-case characters
* It starts and ends with alphabets.
* It only has hyphens or underscores as separators.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/tools/config-gen/parser_test.go
+++ b/tools/config-gen/parser_test.go
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


tools/config-gen/parser_test.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func TestCheckFlagName_Valid(t *testing.T) {
+	validNames := []string{"a", "abc", "ab-c", "ab-c-d", "a_b"}
+
+	for _, name := range validNames {
+		t.Run(name, func(t *testing.T) {
+			assert.NoError(t, checkFlagName(name))
+		})
+
```
raj-prince:
```suggestion
```

raj-prince (approved):

PR #3608, opened 2025-07-30
test(rapid appends): test size visible to kernel on appends

anushka567:
### Description
This PR covers the functional test for rapid appends , majorly concerned with the size which is visible to kernel as appends to an unfinalized object is done. Scenarios include: 
- When a file is appended and the stat cache is valid, the kernel must see the correct updated file size
- When a file is appended and the stat cache is expired (new inode for the file), the kernel must see the correct file size.
[b/433375953](b/433375953)
[b/433375791](b/433375791)

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - Part of the PR.

### DIFF
--- a/tools/integration_tests/rapid_appends/appends_test.go
+++ b/tools/integration_tests/rapid_appends/appends_test.go
@@ -15,7 +15,6 @@
 package rapid_appends
 
 import (
-	"fmt"
 	"log"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


tools/integration_tests/rapid_appends/appends_test.go
```diff
@@ -332,3 +333,70 @@ func (t *SingleMountAppendsSuite) TestFallbackHappensWhenNonAppendHandleDoesFirs
...
+			expectedFileSize := int64(unfinalizedObjectSize + len(initialContent))
+			fileInfo, err := operations.StatFile(filePath)
+			assert.NoError(t.T(), err)
+			assert.Equal(t.T(), expectedFileSize, (*fileInfo).Size())
+		}()
+	}
+}
+
+func (t *SingleMountAppendsSuite) TestKernelShouldSeeUpdatedSizeOnAppendWithExpiredStatCache() {
+	const initialContent = "dummy content"
```
vadlakondaswetha: can you combine these 2 using testcases mode. the only code diff is sleep

anushka567: Done.

PR #3611, opened 2025-07-30
perf(metrics): Add support for all metrics

kislaykishore:
### Description
*  Add generated code for all metrics
* Made some minor formatting fixes.

We haven't added any tests here since when we actually integrate this code, we will reuse the tests that we already have in order to ensure that the two implementations are equivalent.
We'll also ensure via a subsequent PR that we verify that none of the existing attributes are missed.

b/434616359

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/optimizedmetrics/metrics.yaml
+++ b/optimizedmetrics/metrics.yaml
@@ -170,7 +170,6 @@
     - "closed"
     - "opened"
 
-
 - metric-name: "gcs/request_count"
   description: "The cumulative number of GCS requests processed along with the GCS method."
   type: "int_counter"


optimizedmetrics/metrics.yaml
```diff
@@ -52,3 +62,154 @@
...
+
+- metric-name: "fs/ops_latency"
+  description: "The cumulative distribution of file system operation latencies"
+  type: "int_histogram"
+  unit: "us"
+  boundaries: *common_boundaries
+  attributes:
+  - attribute-name: fs_op
+    attribute-type: string
+    values: *fs_ops_list
```
raj-prince, line 139: I see, values somewhere `*` and somewhere `&`. Is it expected?

optimizedmetrics/metrics.yaml
```diff
@@ -5,16 +5,26 @@
...
 
+- metric-name: "file_cache/read_count"
+  description: "Specifies the number of read requests made via file cache along with type - Sequential/Random and cache hit - true/false"
+  type: "int_counter"
+  attributes:
+  - attribute-name: cache_hit
+    attribute-type: bool
+  - attribute-name: read_type
+    attribute-type: string
+    values: *read_types_list
```
raj-prince, line 21: Does this use the value defined at line:8?

kislaykishore: Yes

optimizedmetrics/metrics.yaml
```diff
@@ -52,3 +62,154 @@
...
+  description: "The cumulative number of GCS object readers opened or closed."
+  type: "int_counter"
+  attributes:
+  - attribute-name: io_method
+    attribute-type: string
+    values:
+    - "closed"
+    - "opened"
+
+
```
raj-prince:
```suggestion
```

raj-prince (approved):

PR #3617, opened 2025-07-31
chore(metrics): adding flags for metrics workers and buffer size

anushka567:
[b/435202471](b/435202471)
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


cfg/validate.go
```diff
@@ -205,6 +205,9 @@ func isValidMetricsConfig(m *MetricsConfig) error {
 	if m.PrometheusPort > maxPortNumber {
 		return fmt.Errorf("prometheus-port must not be higher than the maximum allowed port number: %d but received: %d instead", maxPortNumber, m.PrometheusPort)
 	}
+	if m.Workers < 1 || m.BufferSize < 1 {
+		return fmt.Errorf("number of metrics workers or metrics buffer size cannot be less than 1")
```
kislaykishore: Let's give a more detailed message than this - the error should indicate which of metric-workers and metrics-buffer-size is invalid.

anushka567: Split the condition here.

PR #3619, opened 2025-07-31
feat(parallel random reads): Changes to enable parallel random read handling in gcs reader

abhishek10004:
### Description
This pull request introduces changes to enable parallel random reads in the GCS reader. This is achieved by refactoring the read logic in gcs_reader.go, introducing a mutex to manage concurrent access for range reader while allowing reads to go in parallel for MRD. The changes also include updates to the test suite to cover the new logic and parallel read scenarios.

Since, we're removed the high level fh lock for read operation (in fs.go), we've added extra synchronization logic in file handle operations.

b/427137532

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Done


internal/fs/fs.go
```diff
@@ -2803,9 +2803,7 @@ func (fs *fileSystem) ReadFile(
 	fh := fs.handles[op.Handle].(*handle.FileHandle)
 	fs.mu.Unlock()
 
-	fh.Lock()
 	fh.Inode().Lock()
-	defer fh.Unlock()
```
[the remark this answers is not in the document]
abhishek10004: This is okay since we are now synchronizing at the reader level (for range readers)

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,120 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
-	if err != nil {
-		err = fmt.Errorf("ReadAt: getReaderInfo: %w", err)
-		return readerResponse, err
+	if reqReaderType == RangeReaderType {
+		gr.mu.Lock()
+		if gr.bucket.BucketType().Zonal && readInfo.expectedOffset != gr.expectedOffset.Load() {
+			readInfo = gr.getReadInfo(offset, readInfo.seekRecorded)
+			reqReaderType = gr.readerType(readInfo.readType, gr.bucket.BucketType())
+		}
+		if reqReaderType == RangeReaderType {
```
vadlakondaswetha, line 150: Please add code comments here

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,120 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
-		readerResponse, err = gr.rangeReader.ReadAt(ctx, readReq)
-		return readerResponse, err
+	if reqReaderType == MultiRangeReaderType {
+		readerResponse, err = gr.mrr.ReadAt(ctx, readReq)
 	}
 
-	readerResponse, err = gr.mrr.ReadAt(ctx, readReq)
-
-	return readerResponse, err
+	return
```
Tulsishah: nit: return readerResponse, err

internal/gcsx/client_readers/range_reader.go
```diff
@@ -335,13 +330,12 @@ func (rr *RangeReader) invalidateReaderIfMisalignedOrTooSmall(offset int64, p []
 // If a reader exists and the read is successful, the data is returned.
 // Otherwise, it returns an error indicating that a fallback to another reader is needed.
 // Make sure to call invalidateReaderIfMisalignedOrTooSmall before using this method.
-func (rr *RangeReader) readFromExistingReader(ctx context.Context, req *gcsx.GCSReaderRequest) (gcsx.ReaderResponse, error) {
+func (rr *RangeReader) readFromExistingReader(ctx context.Context, req *gcsx.GCSReaderRequest) (int, error) {
+	rr.skipBytes(req.Offset)
+	rr.invalidateReaderIfMisalignedOrTooSmall(req.Offset, req.Offset+int64(len(req.Buffer)))
```
vadlakondaswetha: please compute and put the value in a variable so it can be reused. req.Offset+int64(len(req.Buffer))

internal/gcsx/client_readers/gcs_reader.go
Tulsishah: nit: return readerResponse, err

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,120 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
+
+	return false
+}
+
+func (gr *GCSReader) getEndOffset(
+	start int64) (end int64) {
+
+	end = gr.determineEnd(start)
 	end = gr.limitEnd(start, end)
+	return
```
Tulsishah: return end

internal/gcsx/client_readers/range_reader.go
vadlakondaswetha: also what is the difference between endOffset in GCSReaderRequest vs this one?

internal/gcsx/client_readers/gcs_reader_test.go
```diff
@@ -407,129 +410,341 @@ func (t *gcsReaderTest) Test_ReadAt_PropagatesCancellation() {
 	}
 }
 
-func (t *gcsReaderTest) Test_ReadInfo_WithInvalidInput() {
-	t.object.Size = 10 * MiB
+func (t *gcsReaderTest) Test_IsSeekNeeded() {
```
vadlakondaswetha, line 413: not reviewing the tests assuming you have written the same tests as random_reader

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -52,6 +52,12 @@ const (
 	MultiRangeReaderType
 )
 
+type readInfo struct {
```
Tulsishah, line 56: Why are we not adding these fields in GCSReader struct?

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,120 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
+		gr.mu.Lock()
+		if gr.bucket.BucketType().Zonal && readInfo.expectedOffset != gr.expectedOffset.Load() {
+			readInfo = gr.getReadInfo(offset, readInfo.seekRecorded)
+			reqReaderType = gr.readerType(readInfo.readType, gr.bucket.BucketType())
+		}
+		if reqReaderType == RangeReaderType {
+			defer gr.mu.Unlock()
+			readReq.EndOffset = gr.getEndOffset(readReq.Offset)
+			readerResponse, err = gr.rangeReader.ReadAt(ctx, readReq)
+			return
```
Tulsishah: nit: return readerResponse, err

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,120 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
+			gr.object.Size)
+		return readerResponse, err
 	}
 
-	gr.rangeReader.invalidateReaderIfMisalignedOrTooSmall(offset, p)
-
 	readReq := &gcsx.GCSReaderRequest{
 		Buffer:    p,
 		Offset:    offset,
-		EndOffset: -1,
```
Tulsishah, line 108: Sorry why did we change this, I think you mentioned in the meeting but I forget.

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,120 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
 
-	// If we don't have a reader, determine whether to read from RangeReader or MultiRangeReader.
-	end, err := gr.getReadInfo(offset, int64(len(p)))
-	if err != nil {
-		err = fmt.Errorf("ReadAt: getReaderInfo: %w", err)
-		return readerResponse, err
+	if reqReaderType == RangeReaderType {
+		gr.mu.Lock()
+		if gr.bucket.BucketType().Zonal && readInfo.expectedOffset != gr.expectedOffset.Load() {
+			readInfo = gr.getReadInfo(offset, readInfo.seekRecorded)
```
Tulsishah, line 145: Can we write comment here on why we are calculating it again?

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,120 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
+
+	end = gr.determineEnd(start)
 	end = gr.limitEnd(start, end)
+	return
+}
+
+// getReadInfo determines the read strategy (sequential or random) for a read
+// request at a given offset and returns read metadata. It also updates the
+// reader's internal state based on the read pattern.
+func (gr *GCSReader) getReadInfo(offset int64, seekRecorded bool) readInfo {
```
Tulsishah, line 206: Can we clarify as a comment on what seekRecorded does refers?

internal/gcsx/client_readers/range_reader.go
```diff
@@ -317,14 +314,12 @@ func (rr *RangeReader) skipBytes(offset int64) {
 // Parameters:
 //   - offset: the starting byte position of the requested read.
 //   - p: the buffer representing the size of the requested read.
```
Tulsishah: Please change these comments as you changed method signature

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: Done

internal/gcsx/client_readers/range_reader.go
abhishek10004: GCSReaderRequest has the computed endOffset which would be used while creating the reader. Here, we're reading from the existing reader, so I'm reading only what the user has requested.

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: Done

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: Because we are not calculating the end offset for MRD separately. So, setting it to proper value here. For Range reader, we will call getEndOffset to get the end offset which would be used in range_reader in case a new reader has to be created.

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: This data is read request specific and not reader specific. The same GCSReader instance might be serving multiple requests in parallel. Hence, created another struct to store this data per read request.

raj-prince, line 56: Should we rename it to more specific and clear? Something like - `RequestState`?

Also, I feel good to document the purpose of this struct, as not very straightforward to understand.

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -52,6 +52,12 @@ const (
 	MultiRangeReaderType
 )
 
+type readInfo struct {
+	readType       int64
+	expectedOffset int64
+	seekRecorded   bool
```
raj-prince: Could you add document the purpose of attributes?

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -93,84 +102,131 @@ func NewGCSReader(obj *gcs.MinObject, bucket gcs.Bucket, config *GCSReaderConfig
...
+	reqReaderType := gr.readerType(readInfo.readType, gr.bucket.BucketType())
+
+	if reqReaderType == RangeReaderType {
+		gr.mu.Lock()
+
+		// In case of multiple threads reading parallely, it is possible that many of them might be waiting
+		// at this lock and hence the earlier calculated value of readerType might not be valid once they
+		// acquire the lock. Hence, needs to be calculated again.
+		// Recalculating only for ZB and only when another read had been performed between now and
+		// the time when readerType was calculated for this request
```
raj-prince:
```suggestion
		// the time when readerType was calculated for this request.
```

internal/fs/handle/file.go
```diff
@@ -86,7 +86,6 @@ func NewFileHandle(inode *inode.FileInode, fileCacheHandler *file.CacheHandler,
 
 // Destroy any resources associated with the handle, which must not be used
 // again.
-// LOCKS_REQUIRED(fh.mu)
```
Tulsishah: LOCK_EXCLUDED(fh.mu) ?

abhishek10004: Fixed.

internal/gcsx/client_readers/gcs_reader_test.go
abhishek10004: yeah, they are same as random_reader

internal/fs/handle/file.go
```diff
@@ -257,6 +261,8 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
 	// If the inode is dirty, there's nothing we can do. Throw away our reader if
 	// we have one.
 	if !fh.inode.SourceGenerationIsAuthoritative() {
+		fh.mu.Lock()
+		defer fh.mu.Unlock()
 		if fh.reader != nil {
 			fh.reader.Destroy()
```
vadlakondaswetha: Too many locks/unlocks in this code
can we avoid destroying the reader incase we need to read it from inode. that way we simplify the lock/unlock logic
Anyways reader will get destroyed when fileHandle is closed. 

@raj-prince - thoughts?

raj-prince: I am in favor of closing the reader. More open connections leads to performance issue, specially when high throughput workload (sequential read) with gRPC protocol is being performed. Ref: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3585

Although such workload is very rare, I feel destroying the reader seems important than simplifying lock/unlock.

internal/fs/handle/file.go
```diff
@@ -257,6 +259,8 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
 	// If the inode is dirty, there's nothing we can do. Throw away our reader if
 	// we have one.
 	if !fh.inode.SourceGenerationIsAuthoritative() {
+		fh.mu.Lock()
```
raj-prince: Seems we are violating the lock ordering here.

Handle lock should be acquired before the inode lock.

internal/fs/handle/file.go
```diff
@@ -273,11 +277,15 @@ func (fh *FileHandle) tryEnsureReader(ctx context.Context, sequentialReadSizeMb
 			fh.reader.Object().Size = fh.inode.SourceGeneration().Size
 			return
 		}
+		fh.mu.Lock()
```
raj-prince: I don't see much benefit here in splitting the lock at multiple places. Instead, we should take for the complete `tryEnsureReader`.

internal/fs/handle/file.go
```diff
@@ -316,6 +324,8 @@ func (fh *FileHandle) tryEnsureReadManager(ctx context.Context, sequentialReadSi
 	fh.destroyReadManager()
 
 	// Create a new read manager for the current inode state.
+	fh.mu.Lock()
```
raj-prince: Seem we are violating the lock ordering - duplicating the comment as it's important.
Ref(lock-ordering): https://github.com/GoogleCloudPlatform/gcsfuse/blob/0c18cd143524ab58dfaee6cfb7f9fab0be05029c/internal/fs/fs.go#L329

internal/fs/handle/file.go
```diff
@@ -329,11 +339,13 @@ func (fh *FileHandle) tryEnsureReadManager(ctx context.Context, sequentialReadSi
 }
 
 // destroyReadManager is a helper function to safely destroy and nil the readManager.
-// This assumes the necessary locks (fh.mu, fh.inode.mu) are already held by the caller.
+// This assumes the necessary locks (fh.inode.mu) are already held by the caller.
 func (fh *FileHandle) destroyReadManager() {
 	if fh.readManager == nil {
 		return
 	}
+	fh.mu.Lock()
```
raj-prince: lock ordering.

internal/gcsx/client_readers/range_reader.go
```diff
@@ -315,16 +312,14 @@ func (rr *RangeReader) skipBytes(offset int64) {
 // a new reader unnecessarily.
 //
 // Parameters:
-//   - offset: the starting byte position of the requested read.
-//   - p: the buffer representing the size of the requested read.
-func (rr *RangeReader) invalidateReaderIfMisalignedOrTooSmall(offset int64, p []byte) {
-	rr.skipBytes(offset)
-
+//   - startOffset: the starting byte position of the requested read.
+//   - endOffset: the ending byte positino of the requested read.
```
raj-prince: nit:
```suggestion
//   - endOffset: the ending byte position of the requested read.
```

internal/gcsx/client_readers/range_reader.go
```diff
@@ -335,13 +330,15 @@ func (rr *RangeReader) invalidateReaderIfMisalignedOrTooSmall(offset int64, p []
 // If a reader exists and the read is successful, the data is returned.
 // Otherwise, it returns an error indicating that a fallback to another reader is needed.
 // Make sure to call invalidateReaderIfMisalignedOrTooSmall before using this method.
```
raj-prince: nit: we can remove this comment, as we are making sure internally.

raj-prince (changes requested):
Overall LGTM, suggested few changes.

internal/fs/handle/file.go
abhishek10004: Thanks. I was not aware of this. Have fixed it now.

internal/fs/handle/file.go
abhishek10004: Without splitting, we are not seeing the perf improvement.

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: Keeping it as readInfo as it stores information about a read request. Let me know if you feel strongly about it.

internal/fs/handle/file.go
abhishek10004: Agreed that the code is getting complicated but I'm also in favor of closing the reader.

raj-prince (approved):
LGTM.

We can add UTs for newly created methods in file-handles, we can take this as fast-follow up.

internal/fs/handle/file.go
```diff
@@ -115,10 +115,25 @@ func (fh *FileHandle) Unlock() {
 	fh.mu.Unlock()
 }
 
+// lockHandleAndRelockInode is a helper function which locks fh.mu maintaing the locking
+// order, i.e. it first unlocks inode lock, then locks fh.mu (RLock or RWlock) and then
+// relocks inode lock.
+// This assumes the necessary locks (fh.inode.mu) are already held by the caller.
```
vadlakondaswetha: we use LOCKS_REQUIRED(fh.inode.mu) annotations to specify the state of locks when the function is invoked. use the same here.

internal/fs/handle/file.go
vadlakondaswetha: we are talking only about a scenario where read happens when teh write is in progress which is rare.

The number of locks/unlocks have increased in the fh.Read() method which also which cause congestion. For eg: 

func (fh *FileHandle) destroyReadManager() {
	fh.lockHandleAndRelockInode(false)
	defer fh.mu.Unlock()
	if fh.readManager == nil {
		return
	}
	fh.readManager.Destroy()
	fh.readManager = nil
}

you dont need to lock/unlock when fh.ReadManager is nil. also why do you need to lock inode for updating fh state?

abhishek10004: Since we are accessing fh.ReadManager (even to check if it's nil), we need a RLock. Since we'll be taking a RWlock after this (in case it's not nil), we took the RWlock at the start itself.

This is due to the lock ordering that is being followed in GCSFuse (https://github.com/GoogleCloudPlatform/gcsfuse/blob/0c18cd143524ab58dfaee6cfb7f9fab0be05029c/internal/fs/fs.go#L329)

internal/fs/handle/file.go
abhishek10004: Done

internal/fs/handle/file.go
```diff
@@ -115,20 +115,65 @@ func (fh *FileHandle) Unlock() {
...
+		n, err := fh.inode.Read(ctx, dst, offset)
+		return dst, n, err
+	}
+
+	fh.lockHandleAndRelockInode(true)
+	defer fh.mu.RUnlock()
+
+	// If the inode is dirty, there's nothing we can do. Throw away our readManager if
+	// we have one & create a new readManager
+	if !fh.isValidReadManager() {
```
vadlakondaswetha: do we have test cases for all these flows?

internal/fs/handle/file.go
```diff
@@ -166,18 +212,39 @@ func (fh *FileHandle) ReadWithReadManager(ctx context.Context, dst []byte, offse
...
+		fh.mu.RUnlock()
+		fh.lockHandleAndRelockInode(false)
+
+		fh.destroyReader()
+		// Attempt to create an appropriate reader.
+		fh.reader = gcsx.NewRandomReader(fh.inode.Source(), fh.inode.Bucket(), sequentialReadSizeMb, fh.fileCacheHandler, fh.cacheFileForRangeRead, fh.metricHandle, &fh.inode.MRDWrapper, fh.config)
+
+		// Release RWLock and take RLock on file handle again
+		fh.mu.Unlock()
+		fh.lockHandleAndRelockInode(true)
```
vadlakondaswetha: you dont need to take inode lock here which is unlocked below.

if fh.isValidReader {
  fh.inode.unlock
\} else {
}

// do the read.

vadlakondaswetha: why do we still need fh.reader != nil check in line no: 254?

internal/fs/handle/file.go
abhishek10004: My plan is to check and add new UTs (to ensure all flows are covered) as part of a follow up PR.

internal/fs/handle/file.go
abhishek10004: It is an extra check currently. It's not needed as NewRandomReader would never return a nil pointer. Will remove it.

abhishek10004: Done

PR #3625, opened 2025-08-01
feat(block-pool): adding tryGet similar to Get() but doesn't block execution when block-pool is full

raj-prince:
### Description
- Adding a new method TryGet in the block-pool, which doesn't block the call in case block-pool limit is reached. Instead returns an error.
- Existing `Get` method is blocks the execution in case block-pool limit is reached.

b/435631891

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/block/block_pool_test.go
+++ b/internal/block/block_pool_test.go
@@ -16,7 +16,6 @@ package block
 
 import (
 	"fmt"
-	"io"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/block/block_pool_test.go
```diff
@@ -74,6 +74,56 @@ func (t *BlockPoolTest) TestInitBlockPoolForNegativeMaxBlocks() {
...
+	require.Nil(t.T(), err)
+	content := []byte("hi")
+	n, err := b.Write(content)
+	require.Equal(t.T(), 2, n)
+	require.Nil(t.T(), err)
+	// Validating the content of the block
+	require.Equal(t.T(), int64(0), b.(*memoryBlock).readSeek)
+	output, err := io.ReadAll(b)
+	require.Nil(t.T(), err)
+	require.Equal(t.T(), content, output)
```
ashmeenkaur:
```suggestion
```
Nit: can be removed

internal/block/block_pool_test.go
```diff
@@ -244,6 +294,37 @@ func (t *BlockPoolTest) TestBlockPoolCreationFailsWhenGlobalMaxBlocksIsZero() {
...
+	b3, err3 := bp.TryGet()
+
+	require.Nil(t.T(), b3)
+	require.NotNil(t.T(), err3)
+	require.ErrorIs(t.T(), err3, CantAllocateAnyBlockError)
+	require.Equal(t.T(), int64(2), bp.totalBlocks)
+}
+
+func (t *BlockPoolTest) TestTryGetWhenTotalBlocksEqualToMaxBlocks() {
+	bp, err := NewGenBlockPool(1024, 10, semaphore.NewWeighted(2), createBlock)
```
ashmeenkaur:
```suggestion
	bp, err := NewGenBlockPool(1024, 10, semaphore.NewWeighted(20), createBlock)
```
it will be limited by global blocks otherwise

ashmeenkaur (approved):

PR #3629, opened 2025-08-04
test(buffered-read): adding log parser utility required for buffered-read e2e test

raj-prince:
### Description
- Implementing parser for buffered read logs. This will be required to write e2e test for buffered-read feature.

b/436123285

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/util/log_parser/json_parser/read_logs/buffered_read_log_parser_test.go
+++ b/tools/integration_tests/util/log_parser/json_parser/read_logs/buffered_read_log_parser_test.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Google LLC
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/tools/integration_tests/util/log_parser/json_parser/read_logs/buffered_read_log_parser_test.go
+++ b/tools/integration_tests/util/log_parser/json_parser/read_logs/buffered_read_log_parser_test.go
@@ -1,0 +1,203 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


tools/integration_tests/util/log_parser/json_parser/read_logs/buffered_read_log_parser_test.go
```diff
@@ -0,0 +1,203 @@
...
+							ExecutionTime:    "223.643µs",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "Test buffered read logs with multiple chunks",
+			reader: bytes.NewReader([]byte(`{"timestamp":{"seconds":1754207548,"nanos":733110719},"severity":"TRACE","message":"fuse_debug: Op 0x0000004e        connection.go:453] <- ReadFile (inode 2, PID 564246, handle 0, offset 34603008, 1048576 bytes)"}
+{"timestamp":{"seconds":1754207548,"nanos":733199657},"severity":"TRACE","message":"2e4645d9-19a8 <- ReadAt(princer-working-dirs:/10G_file, 0, 34603008, 1048576, 2)"}
```
ashmeenkaur, line 73: the log is not readable. Do you want to add keys to it? like what is 0, 34603008, 1048576, etc?

ashmeenkaur (approved):
Don't think this PR requires an explicit review as it is just log parsing logic. 
Discussed offline with prince that we will merge the log parsing logiceventually in follow up PRs. And we will also take a call on adding keys for better log readability. This PR is required to unblock e2e tests so approving.

tools/integration_tests/util/log_parser/json_parser/read_logs/buffered_read_log_parser_test.go
raj-prince: I'll add keys for the important one, offset, len and block-index, and will take as a follow up PR.

PR #3638, opened 2025-08-04
refactor(metrics): use optimized metrics implementation

anushka567:
### Description
This PR contains code refactoring to make use of the optimized metrics implementation and cleanup of obsolete code.

[b/435204527](b/435204527)
[b/435204267](b/435204267)

### Testing details
1. Manual - Yes, basic flows
2. Unit tests - NA

### DIFF
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -23,12 +23,16 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 )
 
+const logInterval = 5 * time.Minute
+
 var (
+	unrecognizedAttr                                                                    atomic.Value
 	fileCacheReadBytesCountReadTypeParallelAttrSet                                      = metric.WithAttributeSet(attribute.NewSet(attribute.String("read_type", "Parallel")))
 	fileCacheReadBytesCountReadTypeRandomAttrSet                                        = metric.WithAttributeSet(attribute.NewSet(attribute.String("read_type", "Random")))
 	fileCacheReadBytesCountReadTypeSequentialAttrSet                                    = metric.WithAttributeSet(attribute.NewSet(attribute.String("read_type", "Sequential")))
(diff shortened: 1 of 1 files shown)


internal/gcsx/random_reader.go
```diff
@@ -516,7 +516,7 @@ func (rr *randomReader) startRead(start int64, end int64) (err error) {
 	rr.limit = end
 
 	requestedDataSize := end - start
-	metrics.CaptureGCSReadMetrics(ctx, rr.metricHandle, metrics.ReadTypeNames[metrics.ReadTypeSequential], requestedDataSize)
+metrics.CaptureGCSReadMetrics(rr.metricHandle, metrics.ReadTypeNames[metrics.ReadTypeSequential], requestedDataSize)
```
kislaykishore: Please run `make build` to format this file.

tools/metrics-gen/main.go
kislaykishore, line 1: Please do this change in a separate PR.

tools/metrics-gen/metric_handle.tpl
```diff
@@ -39,3 +39,8 @@ type MetricHandle interface {
 
 {{end}}
 }
+
+func CaptureGCSReadMetrics(mh MetricHandle, readType string, downloadBytes int64) {
+	mh.GcsReadCount(1, readType)
+	mh.GcsDownloadBytesCount(downloadBytes, readType)
+}
```
kislaykishore: Let's not have this in the template; instead have it hand-written. The generated code depends on metrics.yaml, but this change will cause metrics.yaml too to depend on generated code which is cyclic in a way. How? Try deleting gcs/read_count metric from metrics.yaml. The generated code won't compile. So, you can't delete gcs/read_count metric without changing the code-generation logic.

tools/metrics-gen/otel_metrics.tpl
```diff
@@ -93,6 +97,7 @@ func (o *otelMetrics) {{toPascal .Name}}(
 func NewOTelMetrics(ctx context.Context, workers int, bufferSize int) (*otelMetrics, error) {
   ch := make(chan histogramRecord, bufferSize)
   var wg sync.WaitGroup
+  startSampledLogging(ctx)
```
kislaykishore: Please raise a separate PR for logging

tools/metrics-gen/metric_handle.tpl
anushka567: Added it to a helper file. 

From what I've understood, if the corresponding metric is removed from the metrics.yaml, then the generated code will still fail in both the case due to undefined method being referenced (either from helper or auto-generated file). However, in the handwritten case, we keep code generation aspect independant of correctness making it easier to debug in case of failure.

kislaykishore: Yes, that's correct.

PR #3645, opened 2025-08-06
feat(block-pool): adding reserved block support in block-pool

raj-prince:
### Description
Adding support to reserve x number of blocks while creation, returns error in case this number of blocks is not available in the globalMaxPool.

TODO

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/block/block_pool.go
+++ b/internal/block/block_pool.go
@@ -45,6 +45,9 @@ type GenBlockPool[T GenBlock] struct {
 	// Total number of blocks created so far.
 	totalBlocks int64
 
+	// Number of blocks reserved at the time of block pool creation.
+	reservedBlocks int64
+
 	// Semaphore used to limit the total number of blocks created across
 	// different files.
 	globalMaxBlocksSem *semaphore.Weighted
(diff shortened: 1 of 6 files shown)


internal/block/block_pool.go
```diff
@@ -172,15 +188,16 @@ func (bp *GenBlockPool[T]) ClearFreeBlockChannel(releaseLastBlock bool) error {
 				return fmt.Errorf("munmap error: %v", err)
 			}
 			bp.totalBlocks--
-			// Release semaphore for all but the last block.
-			if bp.totalBlocks != 0 {
+			// Release semaphore for all but the reserved blocks.
```
vadlakondaswetha, line 190: why not move the decrement after the if clause?

internal/block/block_pool_test.go
```diff
@@ -443,3 +553,78 @@ func (t *BlockPoolTest) TestCanAllocateBlock() {
...
+
+func (t *BlockPoolTest) TestBlockPoolCreationWithReservedBlocksFailure() {
+	tests := []struct {
+		name            string
+		reservedBlocks  int64
+		maxBlocks       int64
+		globalMaxBlocks int64
+	}{
+		{
+			name:            "reserved_blocks_greater_than_max_blocks",
```
vadlakondaswetha, line 604: should we validate the error type also since its different for each case

PR #3646, opened 2025-08-06
chore: clean up timeout from MRD

ashmeenkaur:
### Description
timeout is currently unused in MRD flow. Cleanup timeout value from the code while we decide how to implement it properly.

b/435356759

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

NA

### DIFF
--- a/internal/gcsx/client_readers/multi_range_reader.go
+++ b/internal/gcsx/client_readers/multi_range_reader.go
@@ -19,18 +19,12 @@ import (
 	"fmt"
 	"io"
 	"sync/atomic"
-	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs"
 	"github.com/googlecloudplatform/gcsfuse/v3/metrics"
 )
-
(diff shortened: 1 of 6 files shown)


charith87 (approved):

internal/gcsx/client_readers/multi_range_reader.go
```diff
@@ -78,7 +78,7 @@ func (mrd *MultiRangeReader) readFromMultiRangeReader(ctx context.Context, p []b
 		mrd.mrdWrapper.IncrementRefCount()
 	}
```
abhishek10004, line 74: Let's remove timeout from readFromMultiRangeReader function as well.

internal/gcsx/client_readers/multi_range_reader_test.go
```diff
@@ -114,7 +114,7 @@ func (t *multiRangeReaderTest) Test_ReadFromMultiRangeReader_ReadFull() {
 	}
 }
 
-func (t *multiRangeReaderTest) Test_ReadFromMultiRangeReader_TimeoutExceeded() {
+func (t *multiRangeReaderTest) TestReadFromMultiRangeReaderTimeoutUnimplemented() {
```
abhishek10004: I would suggest removing timeout parameter from readFromMultiRangeReader as well and then we should remove this test as well

PR #3648, opened 2025-08-06
fix: Handle Symlink scenario in rename

vipnydav:
### Description
A recent refactoring of the `Rename` operation in `internal/fs/fs.go` introduced an issue that prevented sym links from being renamed.

The new implementation incorrectly assumed that all inodes being renamed would implement the `inode.BucketOwnedInode` interface. Since `SymlinkInode` did not, any attempt to rename a symlink would fail with the error: `Rename: input/output error, child inode (id <X>) is not owned by any bucket`.

This change resolves the issue by making `SymlinkInode` implement the `BucketOwnedInode` interface. This involves:
1.  Adding a `bucket` field to `SymlinkInode`.
2.  Updating the `NewSymlinkInode` function and its call sites to correctly propagate bucket information.
3.  Adding logic within the `Rename` function to correctly handle symlinks as bucket-owned inodes.

With this fix, renaming symbolic links now works as expected.

https://b.corp.google.com/issues/436756784

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/internal/fs/rename_dir_test.go
+++ b/internal/fs/rename_dir_test.go
@@ -28,6 +28,7 @@ import (
 
 type RenameDirTests struct {
 	suite.Suite
+	fsTest
 }
 
 ////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


internal/fs/rename_dir_test.go
```diff
@@ -83,6 +84,29 @@ func (t *RenameDirTests) TestRenameFolderWithEmptySourceDirectory() {
...
+	err := os.Mkdir(path.Join(mntDir, targetDirName), dirPerms)
+	require.NoError(t.T(), err)
+	oldPath := path.Join(mntDir, "symlink_old")
+	err = os.Symlink(targetDirName, oldPath)
+	require.NoError(t.T(), err)
+	newPath := path.Join(mntDir, "symlink_new")
+
+	err = os.Rename(oldPath, newPath)
+
+	assert.NoError(t.T(), err)
```
ashmeenkaur, line 98: nit:.
```suggestion
	require.NoError(t.T(), err)
```
at other places too

ashmeenkaur (approved):

internal/fs/rename_dir_test.go
vipnydav: Will update it in next PR.

PR #3659, opened 2025-08-08
test: Additional UTs for file handle

abhishek10004:
### Description
Additional UTs for file handle

### Testing details
1. Manual - NA
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/fs/handle/file_test.go
+++ b/internal/fs/handle/file_test.go
@@ -214,14 +214,13 @@ func (t *fileTest) Test_IsValidReadManager_GenerationValidation() {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func() {
-			mockReadManager := new(read_manager.MockReadManager)
-			mockReadManager.On("Object").Return(&gcs.MinObject{Generation: tc.readerGeneration})
-			fh.readManager = mockReadManager
+			minObj := in.Source()
+			minObj.Generation = tc.readerGeneration
+			fh.readManager = read_manager.NewReadManager(minObj, &t.bucket, &read_manager.ReadManagerConfig{Config: config})
 
 			result := fh.isValidReadManager()
 
 			assert.Equal(t.T(), tc.expectedIsValid, result)
-			mockReadManager.AssertExpectations(t.T())
 		})
 	}
 }
@@ -271,14 +270,13 @@ func (t *fileTest) Test_IsValidReader_GenerationValidation() {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func() {
-			mockReader := new(gcsx.MockRandomReader)
-			mockReader.On("Object").Return(&gcs.MinObject{Generation: tc.readerGeneration})
-			fh.reader = mockReader
+			minObj := in.Source()
+			minObj.Generation = tc.readerGeneration
+			fh.reader = gcsx.NewRandomReader(minObj, &t.bucket, 200, nil, false, metrics.NewNoopMetrics(), &in.MRDWrapper, config)
 
 			result := fh.isValidReader()
 
 			assert.Equal(t.T(), tc.expectedIsValid, result)
-			mockReader.AssertExpectations(t.T())
 		})
 	}
 }


internal/fs/handle/file_test.go
```diff
@@ -163,6 +167,129 @@ func (t *fileTest) TestFileHandleWrite() {
...
+	mockRMForMismatch := new(read_manager.MockReadManager)
+	// Inode has generation 1. Let's set readManager's object generation to 2.
+	mockRMForMismatch.On("Object").Return(&gcs.MinObject{Generation: 2})
+
+	rmObjectForMatch := &gcs.MinObject{Generation: 1, Size: 0}
+	mockRMForMatch := new(read_manager.MockReadManager)
+	// Inode has generation 1.
+	mockRMForMatch.On("Object").Return(rmObjectForMatch)
+
+	testCases := []struct {
```
vadlakondaswetha, line 198: we shouldnt force fit everything under testcases. why can't we write seperate test case of each flow, improves readability

internal/fs/handle/file_test.go
```diff
@@ -163,6 +167,129 @@ func (t *fileTest) TestFileHandleWrite() {
 	assert.Equal(t.T(), data, buf)
 }
 
+func (t *fileTest) TestIsValidReadManager() {
```
vadlakondaswetha: do we have a scenario where we test the ReadWithReadManager when we have to destroy the reader?

abhishek10004: destroyReadManager is called internally from ReadWithReadManager only (similar for reader).
The fileHandle's Destroy is called from ReleaseFileHandle (in fs.go) after taking a RWLock on file handle. 

> do we have a scenario where we test the ReadWithReadManager when we have to destroy the reader?
Currently we're not testing this.

internal/fs/handle/file_test.go
abhishek10004: Done

internal/fs/handle/file_test.go
abhishek10004: Added a test case for this - Test_ReadWithReadManager_ReadManagerInvalidatedByGenerationChange

Also fixed an issue in CreateFileInode where the object name was not matching in the inode & the minObject.

internal/fs/handle/file_test.go
```diff
@@ -166,6 +169,118 @@ func (t *fileTest) TestFileHandleWrite() {
...
+	parent := createDirInode(&t.bucket, &t.clock)
+	config := &cfg.Config{}
+	const objectName = "test_obj"
+	const objectContent = "some data"
+	in := createFileInode(t.T(), &t.bucket, &t.clock, config, parent, objectName, []byte(objectContent), false)
+	fh := NewFileHandle(in, nil, false, nil, util.Read, config, nil, nil)
+	fh.inode.Lock()
+	defer fh.inode.Unlock()
+	mockRMForMismatch := new(read_manager.MockReadManager)
+	mockRMForMismatch.On("Object").Return(&gcs.MinObject{Generation: 2})
```
Tulsishah: nit: remove new line.

internal/fs/handle/file_test.go
```diff
@@ -166,6 +169,118 @@ func (t *fileTest) TestFileHandleWrite() {
 	assert.Equal(t.T(), data, buf)
 }
```
Tulsishah, line 171: Can we write one test on concurrent writing and reading(O_DIRECT) from the same file with multiple go routines?

Asking because I was able to [catch crash](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3694) scenario for buffered reader with this scenario.

abhishek10004: Discussed offline with Tulsi. Will cover this in the e2e tests.

internal/fs/handle/file_test.go
```diff
@@ -166,6 +169,118 @@ func (t *fileTest) TestFileHandleWrite() {
...
+func (t *fileTest) Test_IsValidReadManager_GenerationMisMatch() {
+	parent := createDirInode(&t.bucket, &t.clock)
+	config := &cfg.Config{}
+	const objectName = "test_obj"
+	const objectContent = "some data"
+	in := createFileInode(t.T(), &t.bucket, &t.clock, config, parent, objectName, []byte(objectContent), false)
+	fh := NewFileHandle(in, nil, false, nil, util.Read, config, nil, nil)
+	fh.inode.Lock()
+	defer fh.inode.Unlock()
+	mockRMForMismatch := new(read_manager.MockReadManager)
```
vadlakondaswetha: why do we need mocks for readManager and RandomREader in these tests. cant we use the real implemenation?

internal/fs/handle/file_test.go
```diff
@@ -166,6 +169,118 @@ func (t *fileTest) TestFileHandleWrite() {
...
+	mockReaderForMismatch.On("Object").Return(&gcs.MinObject{Generation: 2})
+	fh.reader = mockReaderForMismatch
+
+	result := fh.isValidReader()
+
+	assert.False(t.T(), result)
+	mockReaderForMismatch.AssertExpectations(t.T())
+}
+
+func (t *fileTest) Test_IsValidReader_GenerationMatch() {
```
vadlakondaswetha: cant we club generationmismatch and match into 1 using testcases??

internal/fs/handle/file_test.go
abhishek10004: We can use the real implementation, but the goal of this test is to test the implementation of isValidReader/isValidReadManager and we just need to get the minObject for generation matching from the readManager/reader. Hence, it makes sense to use mock readManager/reader.

internal/fs/handle/file_test.go
abhishek10004: yeah, will do that

vadlakondaswetha (approved):

PR #3664, opened 2025-08-10
test: Stabilize mount_timeout test by waiting for unmount

vipnydav:
### Description
This PR addresses flaky `mount_timeout` integration test that fails with a `fusermount: exit status 1` error. The root cause seems to be a race condition where a new mount is attempted before the kernel has fully released the resources from a previous unmount. To fix this, a new `unmountAndWait` function is introduced, which polls `/proc/mounts` to ensure the directory is fully unmounted before proceeding. This is a more robust solution than a static sleep.

https://b.corp.google.com/issues/435312363

### Testing details
1. Manual - Done
2. Unit tests - NA


tools/integration_tests/mount_timeout/gcsfuse_mount_timeout_test.go
```diff
@@ -195,11 +197,58 @@ func (testSuite *MountTimeoutTest) mountOrTimeout(bucketName, mountDir, clientPr
...
+		}
+	}
+
+	// Poll for up to 5 seconds for the unmount to complete.
+	for i := 0; i < 500; i++ {
+		mounted, err := isMounted(mountDir)
+		if err != nil || !mounted {
+			return err // Success or error checking mount status.
+		}
+		time.Sleep(10 * time.Millisecond)
```
Tulsishah: Shouldn't we try to unmount in else condition?

tools/integration_tests/mount_timeout/gcsfuse_mount_timeout_test.go
```diff
@@ -195,11 +197,58 @@ func (testSuite *MountTimeoutTest) mountOrTimeout(bucketName, mountDir, clientPr
...
+
+	if err := util.Unmount(mountDir); err != nil {
+		// It might already be unmounted or in a weird state. If the error
+		// indicates it's not mounted, we can proceed to verify.
+		if !strings.Contains(err.Error(), "not mounted") && !strings.Contains(err.Error(), "no such file or directory") {
+			return fmt.Errorf("unmount call failed: %w", err)
+		}
+	}
+
+	// Poll for up to 5 seconds for the unmount to complete.
```
Tulsishah, line 235: nit: 10s?

tools/integration_tests/mount_timeout/gcsfuse_mount_timeout_test.go
```diff
@@ -195,11 +197,58 @@ func (testSuite *MountTimeoutTest) mountOrTimeout(bucketName, mountDir, clientPr
...
+		if err != nil || !mounted {
+			return err // Success or error checking mount status.
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+
+	return fmt.Errorf("timed out waiting for %q to unmount", mountDir)
+}
+
+func (testSuite *NonZBMountTimeoutTest) mountOrTimeout(bucketName, clientProtocol string, expectedMountTime time.Duration) error {
```
gargnitingoogle: You don't need to explicitly override this function in these ZBMountTimeoutTest and NonZBMountTimeoutTest if the implementation is same as in their encapsulated class MountTimeoutTest, unless you want to change the implementation somehow. 

Take a look at https://go.dev/play/p/G7dYb2eZQjD as reference.

tools/integration_tests/mount_timeout/gcsfuse_mount_timeout_test.go
```diff
@@ -195,11 +197,58 @@ func (testSuite *MountTimeoutTest) mountOrTimeout(bucketName, mountDir, clientPr
...
+	if err := util.Unmount(mountDir); err != nil {
+		// It might already be unmounted or in a weird state. If the error
+		// indicates it's not mounted, we can proceed to verify.
+		if !strings.Contains(err.Error(), "not mounted") && !strings.Contains(err.Error(), "no such file or directory") {
+			return fmt.Errorf("unmount call failed: %w", err)
+		}
+	}
+
+	// Poll for up to 5 seconds for the unmount to complete.
+	for i := 0; i < 500; i++ {
```
gargnitingoogle:
```suggestion
	for range 500; i++ {
```

tools/integration_tests/mount_timeout/gcsfuse_mount_timeout_test.go
```diff
@@ -195,11 +197,58 @@ func (testSuite *MountTimeoutTest) mountOrTimeout(bucketName, mountDir, clientPr
...
+		mounted, err := isMounted(mountDir)
+		if err != nil || !mounted {
+			return err // Success or error checking mount status.
```
gargnitingoogle: If error is nil, which means mounted is also false, then the error returned will be nil, which is misleading. Shouldn't we then still return an error ?

gargnitingoogle: never mind. This function should not return error if unmount happened or mount wasn't found.

gargnitingoogle: This can be marked resolved.

tools/integration_tests/mount_timeout/gcsfuse_mount_timeout_test.go
Tulsishah: Discussed offline.

The unmount command succeeded, but we're still waiting for the resources to be cleaned up from the mount stack.

tools/integration_tests/mount_timeout/gcsfuse_mount_timeout_test.go
vipnydav: The methods was pre existing. I just removed the mountDir parameter from it to remove redundancy.
But will try to remove the method itself.

PR #3666, opened 2025-08-11
fix(metrics): Prevent panic from unrecognized histogram attributes

anushka567:
### Description
This change fixes a `panic: runtime error: invalid memory address or nil pointer dereference` that occurs when GCSfuse is run with metrics enabled and encounters an operation with an unrecognized attribute for a histogram-based metric .

When an attribute for one of these metrics is not recognized, the code proceeds with a nil histogramRecord . This nil record is then passed to a background goroutine for processing, causing a nil pointer dereference.

The fix is to add a return statement immediately after logging the unrecognized attribute.

Follow-up in seperate [PR](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3667) : Run all integration test packages with metrics enabled .

### Testing details
1. Manual - Since the issue raised was related to readdirplus experimental flag, i have tested manually by enabling the feature flags and metrics , and performed long listing. The code did not panic. 
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/metrics-gen/main.go
+++ b/tools/metrics-gen/main.go
@@ -253,6 +253,7 @@ func generateCombinations(attributes []Attribute) []AttrCombination {
 func handleDefaultInSwitchCase(level int, attrName string, builder *strings.Builder) {
 	builder.WriteString(fmt.Sprintf("%sdefault:\n", strings.Repeat("\t", level+2)))
 	builder.WriteString(fmt.Sprintf("%supdateUnrecognizedAttribute(%s)\n", strings.Repeat("\t", level+3), toCamel(attrName)))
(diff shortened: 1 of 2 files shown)


tools/metrics-gen/main.go
```diff
@@ -388,6 +388,9 @@ func buildSwitches(metric Metric) string {
 		}
 		if attr.Type == "string" {
 			handleDefaultInSwitchCase(level, attr.Name, &builder)
+			if metric.Type == "int_histogram" {
```
kislaykishore: I think this logic should be embedded within the handleDefaultInSwitchCase method.

anushka567: Initially excluded due to redundancy. I have added it back for better readability.

PR #3668, opened 2025-08-11
feat(buffered-read): fallback to another reader if sufficient blocks are not available in global-block-pool

raj-prince:
### Description
- Taking a data driven approach to compute the minimum number of blocks to not regress the performance via buffered-read.
- Why are we not reserving the block for best performance?
   - Keeping the smaller min-block-cnt help enabling buffered reader for low-end VMs too.
   - Throughput is measured for the complete workload. So, should be okay to look at holistically.

b/437878537

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/bufferedread/min_block_predictor.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (d *defaultMinBlockPredictor) PredictMinBlockCount(blockSize int64, objectSize uint64) (uint, error) {
+	if blockSize <= 0 {
+		return 0, fmt.Errorf("invalid block-size: %d", blockSize)
+	}
+
+	// Cap the block count based on the object size.
+	maxBlockCount := (objectSize + uint64(blockSize) - 1) / uint64(blockSize)
+
+	if blockSize <= 4*util.MiB {
```
Tulsishah: Can we create const for 4 and 8MIB? and also mentioned as a comment why did we choose it?

internal/bufferedread/min_block_predictor_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			predictor := &staticMinBlockPredictor{minBlockCount: tt.staticBlockCnt}
+
+			_, err := predictor.PredictMinBlockCount(tt.blockSize, tt.objectSize)
+
+			assert.Error(t, err)
+
```
Tulsishah: nit: remove empty line

internal/bufferedread/buffered_reader_test.go
```diff
@@ -126,7 +126,7 @@ func (t *BufferedReaderTest) TearDownTest() {
 }
 
 func (t *BufferedReaderTest) TestNewBufferedReader() {
-	reader, err := NewBufferedReader(t.object, t.bucket, t.config, t.globalMaxBlocksSem, t.workerPool, t.metricHandle)
+	reader, err := newBufferedReaderWithMinBlockPredictor(t.object, t.bucket, t.config, &staticMinBlockPredictor{minBlockCount: 2}, t.globalMaxBlocksSem, t.workerPool, t.metricHandle)
```
Tulsishah, line 129: Why are we not calling `NewBufferedReader` here? Is it due to you want to choose static blocks?

internal/bufferedread/buffered_reader.go
```diff
@@ -94,8 +94,17 @@ type BufferedReader struct {
 
 // NewBufferedReader returns a new bufferedReader instance.
 func NewBufferedReader(object *gcs.MinObject, bucket gcs.Bucket, config *BufferedReadConfig, globalMaxBlocksSem *semaphore.Weighted, workerPool workerpool.WorkerPool, metricHandle metrics.MetricHandle) (*BufferedReader, error) {
-	// TODO: To pass the minimum required block-count based on the block-size.
-	blockpool, err := block.NewPrefetchBlockPool(config.PrefetchBlockSizeBytes, config.MaxPrefetchBlockCnt, 2, globalMaxBlocksSem)
+	return newBufferedReaderWithMinBlockPredictor(object, bucket, config, &defaultMinBlockPredictor{}, globalMaxBlocksSem, workerPool, metricHandle)
```
Tulsishah, line 97: Shouldn't we have an option to choose static blocks or if not passed then default one?

raj-prince: You are saying option in the NewBufferedReader() method.

I didn't want to confuse the client with this option, as mostly one would be active. so thought to hide this. And for internal testing created a private method which can take different MinBlockPredictor.

internal/bufferedread/buffered_reader_test.go
raj-prince: Yes.

raj-prince:
Closing in favor of this PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3679

PR #3669, opened 2025-08-11
refactor: explicit_dir test package migration [GKE-GCSFuse Test migration]

ashmeenkaur:
### Description
This PR includes changes to migrate explicit_dir package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.
 
**Changes include:**
1. config file flag addition.
2. refactoring to use config file by test methods.
3. changes are made in a backward compatible way so tests can run with both config file and flags. This will be cleaned up in future PRs after the migration is complete.
4. Migration of explicit dir package so it can use config file.

b/430184649

### Testing details
5. Unit tests - NA
6. Integration tests - via KOKORO

NA

### DIFF
--- a/tools/integration_tests/util/mounting/persistent_mounting/perisistent_mounting.go
+++ b/tools/integration_tests/util/mounting/persistent_mounting/perisistent_mounting.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/test_suite"
 )
 
 // Change e.g --log_severity=trace to log_severity=trace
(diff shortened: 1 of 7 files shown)


tools/integration_tests/util/mounting/persistent_mounting/perisistent_mounting.go
```diff
@@ -76,12 +77,19 @@ func executeTestsForPersistentMounting(flagsSet [][]string, m *testing.M) (succe
 	return
 }
 
+// Deprecated: Use RunTestsWithConfigFile instead.
 func RunTests(flagsSet [][]string, m *testing.M) (successCode int) {
-	log.Println("Running persistent mounting tests...")
-
-	successCode = executeTestsForPersistentMounting(flagsSet, m)
+	config := &test_suite.TestConfig{
```
Tulsishah, line 83: Should we add TODO that we will slowly deprecate existing and pass config directly?

tools/integration_tests/util/setup/implicit_and_explicit_dir_setup/implicit_and_explicit_dir_setup.go
```diff
@@ -46,6 +47,32 @@ const SecondFileInExplicitDirectory = "fileInExplicitDir2"
 const FileInImplicitDirectory = "fileInImplicitDir1"
 const FileInImplicitSubDirectory = "fileInImplicitDir2"
 
+func RunTestsForExplicitDir(config *test_suite.TestConfig, flags [][]string, m *testing.M) int {
+	if config == nil {
+		log.Println("config is nil")
+		return 1
+	}
+
```
Tulsishah, line 55: Shouldn't we have a check here for bucket type?

ashmeenkaur: the check is already there in TestMain - tools/integration_tests/explicit_dir/explicit_dir_test.go line 88

PR #3674, opened 2025-08-11
fix(Buffered Read):  Unexpected EOF from buffered reader causes reads to fail before completion

Tulsishah:
### Description
When reading a file using the buffered reader, an io.EOF is returned prematurely from [buffered_reader.go](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/handle/file.go#L216). This causes read operations to terminate before reading the expected number of bytes.

A test case attempting to read a 5MB file fails because it receives 0 bytes instead of the expected 5MB. The issue was traced back to the following line where EOF is set.

Repro test - https://paste.googleplex.com/5990119374585856

[b/437844604](https://b.corp.google.com/issues/437844604)

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA


internal/fs/handle/file_test.go
```diff
@@ -414,3 +416,38 @@ func (t *fileTest) TestFileHandle_CheckInvariants_WithNilReaderAndManager() {
 		fh.checkInvariants()
 	})
 }
+
+func (t *fileTest) Test_ReadWithReadManager_FullReadSuccessWithBufferedRead() {
+	const (
+		fileSize = 5 * 1024 * 1024 // 5 MiB
```
raj-prince: Can we simplify the test by keeping fileSize as 1 MiB, which is realistic too? As we get max 1MiB call from kernel.

Tulsishah: Done

internal/fs/handle/file_test.go
```diff
@@ -414,3 +416,38 @@ func (t *fileTest) TestFileHandle_CheckInvariants_WithNilReaderAndManager() {
 		fh.checkInvariants()
 	})
 }
+
+func (t *fileTest) Test_ReadWithReadManager_FullReadSuccessWithBufferedRead() {
+	const (
+		fileSize = 1 * 1024 * 1024 // 5 MiB
```
raj-prince, line 422: nit: 1 MiB

PR #3677, opened 2025-08-12
fix(Buffered read): Make buffered read path thread safe

Tulsishah:
### Description
Following the recent removal of file-handle-level locks, the buffered read path is no longer thread-safe. Consequently, running multiple goroutines on the same file handle can lead to issues due to shared resources. Adding locks in buffered Read readAt method to make it thread safe. Also adding unit test to validate the same.

[b/438091542](https://b.corp.google.com/issues/438091542)

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -93,7 +93,7 @@ type BufferedReader struct {
 
 	randomReadsThreshold int64 // Number of random reads after which the reader falls back to another reader.
 
-	// mu synchronizes reads through buffered reader.
+	// mu synchronizes access to the buffered reader's shared state.
 	mu sync.Mutex
 }
 
@@ -440,6 +440,9 @@ func (p *BufferedReader) scheduleBlockWithIndex(b block.PrefetchBlock, blockInde
 }
 
 func (p *BufferedReader) Destroy() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
 	if p.cancelFunc != nil {
 		p.cancelFunc()
 		p.cancelFunc = nil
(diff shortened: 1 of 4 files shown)


internal/bufferedread/buffered_reader.go
```diff
@@ -77,20 +78,35 @@ type BufferedReader struct {
...
+	// All shared variables, such as the block pool and queue, require this lock before any operation.
+	mu sync.Mutex
+
+	// GUARDED by (mu)
+	workerPool workerpool.WorkerPool
+
+	// blockQueue is the core of the prefetching pipeline, holding blocks that are
+	// either downloaded or in the process of being downloaded. The head of the
+	// queue always contains the next block required to satisfy a read request.
+	// GUARDED by (mu)
```
raj-prince:
```suggestion
	// blockQueue is the core of the prefetching pipeline, holding blocks that are
	// either downloaded or in the process of being downloaded.
	// GUARDED by (mu)
```

internal/bufferedread/buffered_reader.go
```diff
@@ -77,20 +78,35 @@ type BufferedReader struct {
...
+	blockQueue common.Queue[*blockQueueEntry]
+
+	// blockPool manages a pool of reusable PrefetchBlock buffers. Instead of
+	// allocating a new buffer for each download, the reader acquires one from
+	// this pool and releases it back when it's no longer needed.
```
raj-prince:
```suggestion
	// blockPool is a pool of blocks that can be reused for prefetching.
	// It is used to avoid allocating new blocks for each prefetch operation.
	// The pool is initialized with a maximum number of blocks that can be
	// prefetched at a time, and it allows for efficient reuse of blocks.
	// The pool is also responsible for managing the global limit on the number
	// of blocks that can be allocated across all BufferedReader instances.
```

PR #3680, opened 2025-08-12
feat(bufferedread): Handle edge cases for non-blocking TryGet in Buffered Reader

vipnydav:
### Description
This PR handles the edge cases involved while using blockPool.TryGet() in Buffered Reader. This prevents read operations from stalling when the block pool is exhausted under high load, making the reader more resilient.

The new error handling logic is as follows:

For urgent reads (e.g., on a cache miss), a failure to acquire a block now triggers a fallback to the next available reader, ensuring the read can continue without delay.
For background prefetching, if a block cannot be acquired, the failure is logged, and the prefetch cycle is gracefully skipped without impacting the ongoing read operation.

https://b.corp.google.com/issues/438120182

### Testing details
1. Manual - Done
2. Unit tests - Added


internal/bufferedread/buffered_reader.go
```diff
@@ -330,32 +329,45 @@ func (p *BufferedReader) ReadAt(ctx context.Context, inputBuf []byte, off int64)
...
 		return nil
 	}
 
-	totalBlockCount := (int64(p.object.Size) + p.config.PrefetchBlockSizeBytes - 1) / p.config.PrefetchBlockSizeBytes
-	for i := int64(0); i < blockCountToPrefetch; i++ {
-		if p.nextBlockIndexToPrefetch >= totalBlockCount {
-			break
-		}
+	var blocksScheduled int64
+	for blocksScheduled < blockCountToPrefetch {
```
raj-prince: nit: why not complete for loop? 

for blocksScheduled := 0; blocksScheduled < blockCountToPrefetch; blocksSchedules++ {
 ....
}

raj-prince: Ohh, got it. 

Keeping a boolean would be more readable, and make that false if hits break condition.

internal/bufferedread/buffered_reader_test.go
```diff
@@ -1088,3 +1169,67 @@ func (t *BufferedReaderTest) TestReadAtExceedsObjectSize() {
 	assertBufferContent(t.T(), buf[:resp.Size], readOffset)
 	t.bucket.AssertExpectations(t.T())
 }
+
+func (t *BufferedReaderTest) TestReadAtSucceedsWhenBackgroundPrefetchFails() {
+	t.config.MaxPrefetchBlockCnt = 2
+	t.config.InitialPrefetchBlockCnt = 2
+	t.globalMaxBlocksSem = semaphore.NewWeighted(2)
```
raj-prince: nit: lets make it obvious by writing a comment. MaxPrefetchBlockCnt should be higher here? Shouldn't?

vipnydav: MaxPrefetchBlockCnt is intentionally set to 2 so that the second block that needs to be prefetched doesn't get a block and prefetching fails.

vipnydav: Removed this test as it was similar to the next test and seemed redundant.

PR #3683, opened 2025-08-12
validate: add logging format validation with default to "json"

alleaditya:
### Description
This pull request refactors the logging configuration validation by introducing constants for logging formats, thereby improving code readability. The validation logic is updated to use a more efficient, and the warning for an unsupported format is now printed to `stdout`.

b/328992990

### Testing details
1. Manual - NA
2. Unit tests - Added test cases
3. Integration tests - NA

### DIFF
--- a/cfg/validate.go
+++ b/cfg/validate.go
@@ -17,8 +17,10 @@ package cfg
 import (
 	"errors"
 	"fmt"
+	"log"
 	"math"
 	"regexp"
+	"slices"
 
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
 )
@@ -245,6 +247,20 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
 	return nil
 }
 
+func isValidLoggingConfig(loggingConfig *LoggingConfig) error {
+	if loggingConfig == nil {
+		return nil
+	}
+
+	SupportedLogFormats := []string{"json", "text"}
+	if loggingConfig.Format != "" && !slices.Contains(SupportedLogFormats, loggingConfig.Format) {
+		loggingConfig.Format = "json" // default to text format if unsupported format is provided
+		log.Printf("WARN: Unsupported logging format provided: %s, defaulting to json format", loggingConfig.Format)
+	}
+
+	return nil
+}
+
 // ValidateConfig returns a non-nil error if the config is invalid.
 func ValidateConfig(v isSet, config *Config) error {
 	var err error
@@ -303,6 +319,10 @@ func ValidateConfig(v isSet, config *Config) error {
 
 	if err = isValidBufferedReadConfig(&config.Read); err != nil {
 		return fmt.Errorf("error parsing buffered read config: %w", err)
+	}
+
+	if err = isValidLoggingConfig(&config.Logging); err != nil {
+		return fmt.Errorf("error parsing logging config: %w", err)
 	}
 
 	return nil
--- a/cfg/validate_test.go
+++ b/cfg/validate_test.go
@@ -918,3 +918,56 @@ func TestValidateLogSeverityRanks(t *testing.T) {
 		})
 	}
 }
+
+func TestIsValidLoggingConfig(t *testing.T) {
+	testCases := []struct {
+		name           string
+		loggingConfig  *LoggingConfig
+		expectedFormat string
+	}{
+		{
+			name: "valid format json",
+			loggingConfig: &LoggingConfig{
+				Format: "json",
+			},
+			expectedFormat: "json",
+		},
+		{
+			name: "valid format text",
+			loggingConfig: &LoggingConfig{
+				Format: "text",
+			},
+			expectedFormat: "text",
+		},
+		{
+			name: "empty format",
+			loggingConfig: &LoggingConfig{
+				Format: "",
+			},
+			expectedFormat: "",
+		},
+		{
+			name: "invalid format",
+			loggingConfig: &LoggingConfig{
+				Format: "invalid",
+			},
+			expectedFormat: "json",
+		},
+		{
+			name:           "nil config",
+			loggingConfig:  nil,
+			expectedFormat: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := isValidLoggingConfig(tc.loggingConfig)
+
+			assert.NoError(t, err)
+			if tc.loggingConfig != nil {
+				assert.Equal(t, tc.expectedFormat, tc.loggingConfig.Format)
+			}
+		})
+	}
+}


cfg/validate.go
```diff
@@ -245,6 +247,20 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
 	return nil
 }
 
+func isValidLoggingConfig(loggingConfig *LoggingConfig) error {
+	if loggingConfig == nil {
+		return nil
+	}
+
+	SupportedLogFormats := []string{"json", "text"}
```
gargnitingoogle: Move this definition into a var() section at the top of this file, e.g.

```go
var(
  SupportedLogFormats = []string{"json", "text"}
)
```

gargnitingoogle: nit: Make the values "json" and "text" as constants.

e.g.
```go
const (
  LOG_FORMAT_JSON = "json"
  LOG_FORMAT_TEXT = "text"
)
```

This will help with reuse.

cfg/validate_test.go
```diff
@@ -918,3 +918,56 @@ func TestValidateLogSeverityRanks(t *testing.T) {
 		})
 	}
 }
+
+func TestIsValidLoggingConfig(t *testing.T) {
```
gargnitingoogle, line 923: Unit tests are needed in this file also: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/cmd/root_test.go  .

cfg/validate.go
```diff
@@ -245,6 +247,20 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
...
+	if loggingConfig.Format != "" && !slices.Contains(SupportedLogFormats, loggingConfig.Format) {
+		loggingConfig.Format = "json" // default to text format if unsupported format is provided
+		log.Printf("WARN: Unsupported logging format provided: %s, defaulting to json format", loggingConfig.Format)
```
gargnitingoogle: This log won't go into the gcsfuse log-file, so this needs to be improved.

cfg/validate.go
```diff
@@ -245,6 +247,20 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
...
 }
 
+func isValidLoggingConfig(loggingConfig *LoggingConfig) error {
+	if loggingConfig == nil {
+		return nil
+	}
+
+	SupportedLogFormats := []string{"json", "text"}
+	if loggingConfig.Format != "" && !slices.Contains(SupportedLogFormats, loggingConfig.Format) {
+		loggingConfig.Format = "json" // default to text format if unsupported format is provided
```
gargnitingoogle:
```suggestion
		loggingConfig.Format = "json" // default to json format if unsupported format is provided
```

cfg/validate_test.go
```diff
@@ -918,3 +918,56 @@ func TestValidateLogSeverityRanks(t *testing.T) {
...
+				Format: "text",
+			},
+			expectedFormat: "text",
+		},
+		{
+			name: "empty format",
+			loggingConfig: &LoggingConfig{
+				Format: "",
+			},
+			expectedFormat: "",
```
gargnitingoogle: Shouldn't this be "json" ?


```suggestion
			expectedFormat: "json",
```

cfg/validate_test.go
```diff
@@ -918,3 +918,56 @@ func TestValidateLogSeverityRanks(t *testing.T) {
...
+			name: "invalid format",
+			loggingConfig: &LoggingConfig{
+				Format: "invalid",
+			},
+			expectedFormat: "json",
+		},
+		{
+			name:           "nil config",
+			loggingConfig:  nil,
+			expectedFormat: "",
```
gargnitingoogle: Shouldn't this return "json"?

```suggestion
			expectedFormat: "json",
```

gargnitingoogle (changes requested):
Added some comments.

cfg/validate.go
alleaditya: Used constants to avoid recreating the values here.

alleaditya: Moved these definition to constants files

cfg/validate_test.go
alleaditya: Updated

cfg/constants.go
```diff
@@ -32,6 +32,13 @@ const (
 	OFF     string = "OFF"
 )
 
+const (
+	// Logging-format constants
+
```
gargnitingoogle: nit: remove the blank line.

cfg/constants.go
```diff
@@ -32,6 +32,13 @@ const (
...
+	LoggingFormatText string = "text"
```
gargnitingoogle: nit: the type string specification is unnecessary in const.

```suggestion
	LoggingFormatText = "text"
```

cfg/validate.go
```diff
@@ -245,6 +246,22 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
...
+func isValidLoggingConfig(loggingConfig *LoggingConfig) error {
+	if loggingConfig == nil {
+		return nil
+	}
+
+	switch strings.TrimSpace(loggingConfig.Format) {
+	case LoggingFormatText, LoggingFormatJson:
+		break
+	default:
+		loggingConfig.Format = LoggingFormatJson // default to JSON format if unsupported format is provided
```
gargnitingoogle: nit: I wonder if we should make another constant like

```go
const DefaultLoggingFormat = LoggingFormatJson
```

Then we can use DefaultLoggingFormat in below log as well as in the tests as expected value in case of improper/empty value.

cfg/validate_test.go
```diff
@@ -18,8 +18,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
```
gargnitingoogle, line 23: is this sorted correctly ? Can you try running `go fmt ./...` or `make build` in gcsfuse folder and see what it does ?

cfg/validate_test.go
```diff
@@ -918,3 +919,55 @@ func TestValidateLogSeverityRanks(t *testing.T) {
...
+		{
+			name: "invalid format",
+			loggingConfig: &LoggingConfig{
+				Format: "invalid",
+			},
+			expectedFormat: "json",
+		},
+		{
+			name:          "nil config",
+			loggingConfig: nil,
```
gargnitingoogle, line 960: nil: let's  add a comment below to indicate that there is no expectedFormat value when loggingConfig is nil.


```suggestion
			loggingConfig: nil,
			// expectedFormat is not relevant if loggingConfig is nil.
```

cmd/root_test.go
```diff
@@ -1023,6 +1023,69 @@ func TestArgsParsing_FileSystemFlagsThrowsError(t *testing.T) {
...
+			name: "default format",
+			args: []string{"gcsfuse", "abc", "pqr"},
+			expectedConfig: &cfg.Config{
+				Logging: cfg.LoggingConfig{
+					Format: "json",
+				},
+			},
+		},
+		{
+			name: "default format",
```
gargnitingoogle: nit: The value `txt` is really sneaky. Took me a few seconds to realize why the default value. Can  we change the testcse name here to something like `unsupported format value`.

cfg/constants.go
gargnitingoogle: I think if you run `go fmt ./...` or `make build` in the gcsfuse folder just before creating the commit, that should fix formatting issues like this I think.

alleaditya: I inferred the style from the previous block of Logging-level constants definitions.

It looks like the link breaks are avoidable when it's a line comments.

Also the comments suggested made no difference on code format before the change.

cmd/root_test.go
alleaditya: I was playing with the test cases to replicate typo errors and hence considered it. Resolved

cfg/validate_test.go
alleaditya: I guess side affect of one of the extensions I installed earlier.

cfg/validate.go
```diff
@@ -245,6 +246,23 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
 	return nil
 }
 
+func isValidLoggingConfig(loggingConfig *LoggingConfig) error {
+	if loggingConfig == nil {
+		return nil
+	}
+
+	const DefaultLoggingFormat = LoggingFormatJson
```
gargnitingoogle: nit: Could you move this also to the `const()` section at the top ?

```suggestion
	const DefaultLoggingFormat = LoggingFormatJson
```

alleaditya: By "Top", I hope you meant to the top of the method.
Not really sure if that's helpful but addressed it.

gargnitingoogle:
My bad. I should have been more specific. I think top of this file.

Thanks and regards,
Nitin Garg

cfg/validate.go
```diff
@@ -245,6 +246,23 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
 	return nil
 }
 
+func isValidLoggingConfig(loggingConfig *LoggingConfig) error {
+	const DefaultLoggingFormat = LoggingFormatJson
```
abhishek10004, line 250: Can we move this out of the function to file scope (in constants.go)? This will allow us to use this constant in tests as well.

cfg/validate_test.go
```diff
@@ -918,3 +918,56 @@ func TestValidateLogSeverityRanks(t *testing.T) {
...
+		name           string
+		loggingConfig  *LoggingConfig
+		expectedFormat string
+	}{
+		{
+			name: "valid format json",
+			loggingConfig: &LoggingConfig{
+				Format: "json",
+			},
+			expectedFormat: "json",
```
abhishek10004, line 934: Can we use the constants wherever applicable (instead of "json", "text")? This will allow the tests to remain unchanged (let's say when default format is changed).

cfg/validate_test.go
```diff
@@ -918,3 +918,56 @@ func TestValidateLogSeverityRanks(t *testing.T) {
...
+		},
+		{
+			name: "invalid format",
+			loggingConfig: &LoggingConfig{
+				Format: "invalid",
+			},
+			expectedFormat: "json",
+			// expectedFormat is not relevant if loggingConfig is nil.
```
abhishek10004: This seems to be relevant for the next test case, not this one.

cfg/validate.go
alleaditya: The constants for each type is defined in constants.go.

I don't think it would serve much use by moving a redundant constant referring to another exact same value in the constants file.

cfg/validate_test.go
alleaditya: I am unable to find a link however this was suggested in one of the tott's to inline Values in test cases.

* Test cases should preferably be clear with exact values to improve readability. 
* This makes the tests more robust against any accidental changes leading that could define a different unintentional behavior.

cfg/validate_test.go
alleaditya: Moved the comment to the start of the test case.

cfg/constants.go
```diff
@@ -32,6 +32,12 @@ const (
 	OFF     string = "OFF"
 )
 
+const (
+	// Logging-format constants
+	LoggingFormatText = "text"
```
kislaykishore, line 37: This doesn't need to be an exported field. Same for json.

cfg/constants.go
```diff
@@ -32,6 +32,12 @@ const (
...
+	LoggingFormatJson = "json"
```
kislaykishore, line 38: The convention in Golang is to keep the abbreviation in the same case.
```suggestion
	LoggingFormatJSON = "json"
```

cfg/validate.go
```diff
@@ -245,6 +246,23 @@ func isValidBufferedReadConfig(rc *ReadConfig) error {
...
+	if loggingConfig == nil {
+		return nil
+	}
+
+	switch strings.TrimSpace(loggingConfig.Format) {
+	case LoggingFormatText, LoggingFormatJson:
+		break
+	default:
+		loggingConfig.Format = DefaultLoggingFormat // default to JSON format if unsupported format is provided
```
kislaykishore, line 259: We should not modify the config in validate.go; it's the job of rationalize.go.

cmd/root_test.go
```diff
@@ -1023,6 +1024,69 @@ func TestArgsParsing_FileSystemFlagsThrowsError(t *testing.T) {
...
+				return nil
+			})
+			require.Nil(t, err)
+			cmd.SetArgs(convertToPosixArgs(tc.args, cmd))
+
+			err = cmd.Execute()
+
+			if assert.NoError(t, err) {
+				assert.Equal(t, tc.expectedConfig.Logging.Format, gotConfig.Logging.Format)
+			}
```
kislaykishore, line 1085: It's recommended to avoid if-else in tests. Also, require.NoError ensures that the test doesn't proceed if an error is thrown.
```suggestion
			err = cmd.Execute()
			require.NoError(t, err)
			
			assert.Equal(t, tc.expectedConfig.Logging.Format, gotConfig.Logging.Format)
```

kislaykishore (changes requested):

alleaditya:
Redoing this in a different changeset

PR #3685, opened 2025-08-12
test(bufferedread): Add setup and helper methods for Buffered Read integration tests

vipnydav:
### Description
This PR adds setup and helper methods for Buffered Read integration tests.
It also adds integration tests to test reading small and large files using buffered reader.

b/438234047
b/438411799

### Testing details
1. Manual - NA
2. Unit tests - NA


tools/integration_tests/buffered_read/helpers_test.go
```diff
@@ -0,0 +1,91 @@
...
+		ObjectName:            path.Join(path.Base(testDir), fileName),
+	}
+	if setup.DynamicBucketMounted() != "" {
+		expected.BucketName = setup.DynamicBucketMounted()
+	}
+
+	var content []byte
+	var err error
+
+	if readFullFile {
```
Tulsishah, line 52: Can we utilize [this](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/read_cache/helpers_test.go#L46) only by putting it in common client.go file?

tools/integration_tests/buffered_read/helpers_test.go
```diff
@@ -0,0 +1,91 @@
...
+	assert.Equal(t, expected.ObjectName, logEntry.ObjectName, "Object names don't match! Expected: %s, Got from logs: %s",
+		expected.ObjectName, logEntry.ObjectName)
+
+	assert.Equal(t, fallback, logEntry.Fallback, "Expected Fallback: %t, Got from logs: %t", fallback, logEntry.Fallback)
+}
+
+func setupFileInTestDir(ctx context.Context, storageClient *storage.Client, testDir string, fileSize int64, t *testing.T) (fileName string) {
+	fileName = testFileName + setup.GenerateRandomString(4)
+	client.SetupFileInTestDirectory(ctx, storageClient, path.Base(testDir), fileName, fileSize, t)
+	return
```
Tulsishah: nit: return fileName

tools/integration_tests/buffered_read/sequential_read_test.go
```diff
@@ -0,0 +1,126 @@
...
+	setup.SaveGCSFuseLogFileInCaseOfFailure(s.T())
+}
+
+// //////////////////////////////////////////////////////////////////////
+// Test Cases
+// //////////////////////////////////////////////////////////////////////
+func (s *SequentialReadSuite) TestSequentialRead() {
+	testCases := []struct {
+		name       string
+		sizeFactor float64
```
Tulsishah: What is this sizeFactor for?
I think we should have fields `file-size`,  `block-size`

tools/integration_tests/buffered_read/setup_test.go
```diff
@@ -0,0 +1,114 @@
...
+	minBlocksPerHandle   int64
+}
+
+////////////////////////////////////////////////////////////////////////
+// Helpers
+////////////////////////////////////////////////////////////////////////
+
+func createConfigFile(flags *gcsfuseTestFlags) string {
+	mountConfig := map[string]interface{}{
+		"read": map[string]interface{}{
```
Tulsishah: I think we should make a check before assigning value. For example if someone missed to pass protocol then it will assign `client-protocol: ""` which  can cause crash

tools/integration_tests/util/log_parser/json_parser/read_logs/buffered_read_log_parser_test.go
```diff
@@ -134,6 +134,76 @@ func TestParseBufferedReadLogsFromLogReaderSuccessful(t *testing.T) {
...
+							StartTimeSeconds: 1754207548,
+							StartTimeNanos:   733199657,
+							RequestID:        "2e4645d9-19a8",
+							Offset:           34603008,
+							Size:             1048576,
+							BlockIndex:       2,
+							ExecutionTime:    "223.643µs",
+						},
+					},
+					Fallback:        true,
```
Tulsishah, line 236: I think let's add failure scenario as well. e.g. No fall back

tools/integration_tests/buffered_read/sequential_read_test.go
vipnydav: sizeFactor was used to calculate the fileSize in terms of blockSize but now I have updated the field to fileSize and calculated it there itself.

tools/integration_tests/util/client/gcs_helper.go
```diff
@@ -183,3 +195,44 @@ func CreateUnfinalizedObject(ctx context.Context, t *testing.T, client *storage.
...
 	return writer
 }
+
+// ReadAndValidateGCSObject reads a file from the mount point and validates its
+// content against the corresponding object in GCS. It can read the full file
+// or a specific chunk. It returns the content read.
+func ReadAndValidateGCSObject(ctx context.Context, storageClient *storage.Client, objectDir, fileName, mountedFilePath string, readFullFile bool, offset, chunkSizeToRead int64, t *testing.T) (content []byte) {
+	var err error
+	if readFullFile {
+		content, err = operations.ReadFileSequentially(mountedFilePath, chunkSizeToRead)
```
raj-prince: So, util packages should be as independent as possible. Why are we making dependency over operations package?

Logically also operations calls doesn't belong to client. So, good to avoid.

I don't mind duplicating the complete content in the test, instead that will increase the readability of the test.

tools/integration_tests/buffered_read/sequential_read_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	for _, tc := range testCases {
+		s.T().Run(tc.name, func(t *testing.T) {
+			err := os.Truncate(setup.LogFile(), 0)
+			require.NoError(t, err, "Failed to truncate log file")
+			testDir := setup.SetupTestDirectory(testDirName)
+			fileName := setupFileInTestDir(ctx, storageClient, testDir, tc.fileSize, t)
+
+			expected := client.ReadFileAndValidate(ctx, storageClient, testDir, fileName, true, 0, chunkSizeToRead, t)
```
raj-prince: Not very readable, I was looking for where are we reading from the file in the test. So, good to duplicate the function execution.

raj-prince: Would it be easy to extend this test for different chunkSize? - we can take that as a separate PR too.

tools/integration_tests/buffered_read/sequential_read_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+			f, err := os.Open(setup.LogFile())
+			require.NoError(t, err, "Failed to open log file")
+			defer f.Close()
+			logEntries, err := read_logs.ParseBufferedReadLogsFromLogReader(f)
```
raj-prince: We can create a helper for this.

raj-prince (commented):
LGTM, few nits.

tools/integration_tests/buffered_read/sequential_read_test.go
vipnydav: Yeah, it is quite easy to incorporate it. Will include in this PR itself.

Tulsishah (dismissed):

Tulsishah (approved):

raj-prince (approved):

PR #3687, opened 2025-08-13
refactor: operations test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate operations package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

[b/438378548](https://b.corp.google.com/issues/438378548)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/operations/operations_test.go
+++ b/tools/integration_tests/operations/operations_test.go
@@ -146,29 +146,28 @@ func TestMain(m *testing.M) {
 		os.Exit(successCodeTPC)
 	}
 
-	var successCode int
-
 	if len(cfg.Operations) == 0 {
 		log.Println("No configuration found for operations tests in config. Using flags instead.")
 		// Populate the config manually.
 		cfg.Operations = make([]test_suite.TestConfig, 1)
 		cfg.Operations[0].TestBucket = setup.TestBucket()
 		cfg.Operations[0].GKEMountedDirectory = setup.MountedDirectory()
-		cfg.Operations[0].Configs = make([]test_suite.ConfigItem, 2)
+		cfg.Operations[0].Configs = make([]test_suite.ConfigItem, 3)
 		cfg.Operations[0].Configs[0].Flags = []string{
+			"--metadata-cache-ttl-secs=0 --enable-streaming-writes=false",
+			"--kernel-list-cache-ttl-secs=-1 --implicit-dirs=true",
+		}
+		cfg.Operations[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": true}
+		cfg.Operations[0].Configs[1].Flags = []string{
 			"--enable-atomic-rename-object=true",
 			"--experimental-enable-json-read=true",
 			"--client-protocol=grpc --implicit-dirs=true --enable-atomic-rename-object=true",
-			"--metadata-cache-ttl-secs=0 --enable-streaming-writes=false",
-			"--kernel-list-cache-ttl-secs=-1 --implicit-dirs=true",
 		}
-		cfg.Operations[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": false, "zonal": false}
-		cfg.Operations[0].Configs[1].Flags = []string{
+		cfg.Operations[0].Configs[1].Compatible = map[string]bool{"flat": true, "hns": false, "zonal": false}
+		cfg.Operations[0].Configs[2].Flags = []string{
 			"--experimental-enable-json-read=true --enable-atomic-rename-object=true",
-			"--metadata-cache-ttl-secs=0 --enable-streaming-writes=false",
-			"--kernel-list-cache-ttl-secs=-1 --implicit-dirs=true",
 		}
-		cfg.Operations[0].Configs[1].Compatible = map[string]bool{"flat": false, "hns": true, "zonal": true}
+		cfg.Operations[0].Configs[2].Compatible = map[string]bool{"flat": false, "hns": true, "zonal": true}
 	}
 
 	ctx = context.Background()
@@ -196,9 +195,7 @@ func TestMain(m *testing.M) {
 	flags := setup.BuildFlagSets(cfg.Operations[0], bucketType, "")
 	setup.SetUpTestDirForTestBucket(&cfg.Operations[0])
 
-	if successCode == 0 {
-		successCode = static_mounting.RunTestsWithConfigFile(&cfg.Operations[0], flags, m)
-	}
+	successCode :=static_mounting.RunTestsWithConfigFile(&cfg.Operations[0], flags, m)
 
 	if successCode == 0 {
 		successCode = only_dir_mounting.RunTestsWithConfigFile(&cfg.Operations[0], flags, onlyDirMounted, m)
--- a/tools/integration_tests/operations/operations_test.go
+++ b/tools/integration_tests/operations/operations_test.go
@@ -17,7 +17,6 @@ package operations_test
 
 import (
 	"context"
-	"fmt"
 	"log"
 	"os"
 	"strings"
@@ -25,6 +24,7 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/client"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/creds_tests"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting/dynamic_mounting"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting/only_dir_mounting"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting/persistent_mounting"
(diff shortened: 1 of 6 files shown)


tools/integration_tests/util/setup/setup.go
```diff
@@ -131,6 +131,10 @@ func TestDir() string {
 	return testDir
 }
 
+func SetBucketFromConfigFile(testBucketFromConfigFile string) {
+	testBucket = &testBucketFromConfigFile
```
ashmeenkaur: I think we should clean up the testBucket flag eventually. Is it possible to use config file in the functions instead?

tools/integration_tests/operations/operations_test.go
```diff
@@ -157,67 +148,45 @@ func TestMain(m *testing.M) {
...
+		successCodeTPC := static_mounting.RunTestsWithConfigFile(&cfg.Operations[0], flags, m)
 		os.Exit(successCodeTPC)
 	}
 
-	successCode := static_mounting.RunTests(flagsSet, m)
+	successCode := static_mounting.RunTestsWithConfigFile(&cfg.Operations[0], flags, m)
 
 	if successCode == 0 {
-		successCode = only_dir_mounting.RunTests(flagsSet, onlyDirMounted, m)
+		successCode = only_dir_mounting.RunTests(flags, onlyDirMounted, m)
```
ashmeenkaur: We can create functions for `only_dir_mounting`, `dynamic_mounting`, and` creds_tests` similar to the ones we've created for `static_mounting` and `persistent_mounting` - `RunTestsWithConfigFile()` function.

tools/integration_tests/operations/operations_test.go
```diff
@@ -95,60 +97,49 @@ var (
...
+			"--enable-atomic-rename-object=true",
+			"--experimental-enable-json-read=true",
+			"--client-protocol=grpc --implicit-dirs=true --enable-atomic-rename-object=true",
+			"--experimental-enable-json-read=true --enable-atomic-rename-object=true",
+			"--create-empty-file=true --enable-atomic-rename-object=true",
+			"--metadata-cache-ttl-secs=0 --enable-streaming-writes=false",
+			"--kernel-list-cache-ttl-secs=-1 --implicit-dirs=true",
+		}
+		cacheDirflag := fmt.Sprintf("--file-cache-max-size-mb=2 --cache-dir=%s/cache-dir-operations-hns", TempDirectory)
+		cfg.Operations[0].Flags = append(cfg.Operations[0].Flags, cacheDirflag)
```
ashmeenkaur: we can hardcode the value to what it was there earlier. But for cases where config file is provided can we dynamically parse the flag to get this value?

PranjalC100: the value of **cacheDirFlag** is being hardcoded only in the block where no configurations is found for operations tests in config. 
If it is provided, it shall take it from the test_config.yaml (commented it out for now, still needs to be done)

tools/integration_tests/test_config.yaml
```diff
@@ -24,3 +24,22 @@ explicit_dir:
...
+    log_file: # Optional flag required by some tests where log parsing is done to validate end behavior
+    flags:
+      - "--enable-atomic-rename-object=true"
+      - "--experimental-enable-json-read=true"
+      - "--client-protocol=grpc --implicit-dirs=true --enable-atomic-rename-object=true"
+      - "--experimental-enable-json-read=true --enable-atomic-rename-object=true"
+      - "--create-empty-file=true --enable-atomic-rename-object=true"
+      - "--metadata-cache-ttl-secs=0 --enable-streaming-writes=false"
+      - "--kernel-list-cache-ttl-secs=-1 --implicit-dirs=true"
+      # - "--file-cache-max-size-mb=2 --cache-dir=${TMPDIR}/cache-dir-operations-hns"
```
PranjalC100: this wasn't working, so had it commented for now. or we will have to setup this up before running the tests.

tools/integration_tests/util/setup/setup.go
ashmeenkaur: discussed offline that we will clean this up later

tools/integration_tests/util/setup/setup.go
```diff
@@ -131,6 +131,10 @@ func TestDir() string {
 	return testDir
 }
 
+func SetBucketFromConfigFile(testBucketFromConfigFile string) {
```
ashmeenkaur: nit: can we add documentation (code comment on top) and rename function to `SetTestBucket()` instead

tools/integration_tests/util/mounting/only_dir_mounting/only_dir_mounting.go
```diff
@@ -22,15 +22,27 @@ import (
...
 func MountGcsfuseWithOnlyDir(flags []string) (err error) {
+	config := &test_suite.TestConfig{
+		TestBucket:       setup.TestBucket(),
+		MountedDirectory: setup.MountedDirectory(),
+		LogFile:          setup.LogFile(),
+	}
+	return MountGcsfuseWithOnlyDirMountingWithConfigFile(config, flags)
+}
+
+func MountGcsfuseWithOnlyDirMountingWithConfigFile(config *test_suite.TestConfig, flags []string) (err error) {
```
ashmeenkaur: Nit:
```suggestion
func MountGcsfuseWithOnlyDirWithConfigFile(config *test_suite.TestConfig, flags []string) (err error) {
```

tools/integration_tests/test_config.yaml
```diff
@@ -24,3 +24,22 @@ explicit_dir:
...
+
+operations:
+  - mounted_directory: "${MOUNTED_DIR}" # To be passed by GKE after mounting
+    test_bucket: "${BUCKET_NAME}" # To be passed by both gcsfuse and gke tests
+    log_file: # Optional flag required by some tests where log parsing is done to validate end behavior
+    flags:
+      - "--enable-atomic-rename-object=true"
+      - "--experimental-enable-json-read=true"
+      - "--client-protocol=grpc --implicit-dirs=true --enable-atomic-rename-object=true"
+      - "--experimental-enable-json-read=true --enable-atomic-rename-object=true"
```
ashmeenkaur: this may not work with TPC based on the previous test setup. Let's validate if it is still not supported.

tools/integration_tests/test_config.yaml
PranjalC100: resolved

tools/integration_tests/test_config.yaml
ashmeenkaur: There is scope to remove a few tests here but we can take that up separately.

tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -124,6 +125,8 @@ func RevokePermission(ctx context.Context, storageClient *storage.Client, servic
 	}
 }
 
+// Deprecated: Use RunTestsForKeyFileAndGoogleApplicationCredentialsEnvVarSetWithConfigFile instead.
+// TODO(b/438068132): cleanup deprecated methods after migration is complete.
 func RunTestsForKeyFileAndGoogleApplicationCredentialsEnvVarSet(ctx context.Context, storageClient *storage.Client, testFlagSet [][]string, permission string, m *testing.M) (successCode int) {
```
ashmeenkaur: Nit: can we generate config file and call RunTestsForKeyFileAndGoogleApplicationCredentialsEnvVarSetWithConfigFile function here instead of copying everything?

tools/integration_tests/util/setup/setup.go
PranjalC100: added

tools/integration_tests/util/creds_tests/creds.go
PranjalC100: got it!

tools/integration_tests/operations/operations_test.go
```diff
@@ -89,109 +91,131 @@ const Content = "line 1\nline 2\n"
...
 
-func createMountConfigsAndEquivalentFlags() (flags [][]string) {
-	mountConfig1 := map[string]any{
-		"metadata-cache": map[string]any{
-			"ttl-secs": 0,
-		},
-		"write": map[string]any{
-			"enable-streaming-writes": false,
-		},
+func overrideFilePathsInFlagSet(t *test_suite.TestConfig, GCSFuseTempDirPath string) {
```
ashmeenkaur, line 99: can we move this to setup and use everywhere?

tools/integration_tests/operations/operations_test.go
```diff
@@ -89,109 +91,131 @@ const Content = "line 1\nline 2\n"
...
+	cfg := test_suite.ReadConfigFile(setup.ConfigFile())
+
+	// TODO: b/469970353 : Update tpc_build.sh to run using test_config.yaml file.
+	if setup.TestOnTPCEndPoint() {
+		log.Println("Running TPC tests without config file.")
+		successCodeTPC := RunTestOnTPCEndPoint(cfg, m)
+		os.Exit(successCodeTPC)
+	}
+
+	var successCode int
```
ashmeenkaur:
```suggestion
```

tools/integration_tests/operations/operations_test.go
```diff
@@ -89,109 +91,131 @@ const Content = "line 1\nline 2\n"
...
+			"--experimental-enable-json-read=true",
+			"--client-protocol=grpc --implicit-dirs=true --enable-atomic-rename-object=true",
+			"--metadata-cache-ttl-secs=0 --enable-streaming-writes=false",
+			"--kernel-list-cache-ttl-secs=-1 --implicit-dirs=true",
+		}
+		cfg.Operations[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": false, "zonal": false}
+		cfg.Operations[0].Configs[1].Flags = []string{
+			"--experimental-enable-json-read=true --enable-atomic-rename-object=true",
+			"--metadata-cache-ttl-secs=0 --enable-streaming-writes=false",
+			"--kernel-list-cache-ttl-secs=-1 --implicit-dirs=true",
```
ashmeenkaur: Can you point me to the logic which was handling these flags for hns/zb separately earlier? Can we also club same flagset together?

tools/integration_tests/util/test_suite/config.go
```diff
@@ -45,6 +45,7 @@ type ConfigItem struct {
 	Compatible map[string]bool `yaml:"compatible"`
 	Run        string          `yaml:"run,omitempty"`
 	RunOnGKE   bool            `yaml:"run_on_gke"`
+	Tpc        bool            `yaml:"tpc,omitempty"`
```
ashmeenkaur: how do we plan to use this flag?

tools/integration_tests/operations/operations_test.go
PranjalC100: I'll do this in a separate PR, making changes in all other packages as well.

tools/integration_tests/operations/operations_test.go
PranjalC100: removed this and initialized this variable when running static mounting test for the first time.

tools/integration_tests/operations/operations_test.go
PranjalC100: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3687/changes#diff-d60121ff47afc8c00b1933c85d0209792a693ec0ff0fd9560360ae84b06a50e9L164-L170

The first flag is being assigned, and the other two were from createMountConfigsAndEquivalentFlags() function

tools/integration_tests/util/test_suite/config.go
PranjalC100: discussed offline

ashmeenkaur (approved):

PR #3699, opened 2025-08-18
test(write_large_files): Add slow writer integration tests.

meet2mky:
### Description
Add slow writer integration tests to verify the resumable api fixes and prevent future regression.

b/436756387

### Testing details
1. Manual - Tested
2. Unit tests - NA
3. Integration tests - Yes

NA

### DIFF
--- a/tools/integration_tests/write_large_files/slow_file_write_test.go
+++ b/tools/integration_tests/write_large_files/slow_file_write_test.go
@@ -45,6 +45,6 @@ func TestSlowWriteToFile(t *testing.T) {
 		operations.SyncFile(fh, t)
 		// Add a delay of 40 sec between each block to ensure 32 sec
 		// ChunkRetryDeadline is completely used while reading the second Chunk.
-		time.Sleep(40 * time.second)
+		time.Sleep(40 * time.Second)
 	}
 }


tools/integration_tests/write_large_files/slow_file_write_test.go
```diff
@@ -0,0 +1,48 @@
...
+	data := setup.GenerateRandomString(32 * operations.MiB)
+
+	// Write 4 blocks of 33MiB to file each with 1 min delay.
+	for range 4 {
+		n, err := fh.Write([]byte(data))
+
+		assert.NoError(t, err)
+		assert.Equal(t, len(data), n)
+		operations.SyncFile(fh, t)
+		time.Sleep(1 * time.Minute)
```
Tulsishah: Do we require 1 minute delay here? can't we reduce?

meet2mky: We are using a value more than 32 sec for the bug to manifest. I think a 40 - 50 sec will also work but 1 min seems fine as long as we are not increasing the e2e script run times.

meet2mky: Reduced blocks to 2 and delay to 40s. 32s (ChunkRetryDeadline) + 8s (Buffer).

PR #3700, opened 2025-08-18
test(buffered read): E2e test parallel read write

Tulsishah:
### Description
This change adds end-to-end tests for parallel reads and writes using 10 goroutines. These tests cover a scenario that previously caused a crash, which has now been [fixed](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3694). The new tests will help prevent future regression.

[b/438679986](https://b.corp.google.com/issues/438679986)
[b/439504167](https://b.corp.google.com/issues/439504167)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


Tulsishah:
@raj-prince will add this as part of concurrent e2e tests in his PR

PR #3701, opened 2025-08-18
refactor: implicit_dir test package migration [GKE-GCSFuse Test migration]

ashmeenkaur:
### Description
This PR includes changes to migrate implicit_dir package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

Changes include:
- changes are made in a backward compatible way so tests can run with both config file and flags. This will be cleaned up in future PRs after the migration is complete.
- Migration of implicit dir package so it can use config file.
- Created common method to fetch bucket type.
- Changes to the config structure so different flagset can run with different bucket types.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

implicir_dir and explicit_dir test package stopped working after this change without config file. This was fixed in PR #3706

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -32,7 +32,9 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+	"cloud.google.com/go/storage/experimental"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/test_suite"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/util"
 	"go.opentelemetry.io/contrib/detectors/gcp"
 	"go.opentelemetry.io/otel/sdk/resource"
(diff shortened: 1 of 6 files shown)


tools/integration_tests/util/setup/setup.go
```diff
@@ -551,6 +553,53 @@ func ResolveIsHierarchicalBucket(ctx context.Context, testBucket string, storage
...
+			// 3. If compatible, process its flags and add them to the result.
+			for _, flagString := range testCase.Flags {
+				dynamicFlags = append(dynamicFlags, strings.Fields(flagString))
+			}
+		}
+	}
+	return dynamicFlags
+}
+
+func SetBucketFromConfigFile(testBucketFromConfigFile string) {
```
ashmeenkaur, line 599: Will be removed after rebasing on top of operation package migration PR

tools/integration_tests/util/setup/setup.go
```diff
@@ -551,6 +553,53 @@ func ResolveIsHierarchicalBucket(ctx context.Context, testBucket string, storage
...
 
+const FlatBucket = "flat"
+const HNSBucket = "hns"
+const ZonalBucket = "zonal"
+
+func BucketType(ctx context.Context, testBucket string) (bucketType string, err error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	storageClient, err := storage.NewGRPCClient(ctx, experimental.WithGRPCBidiReads())
+	if err != nil {
```
Tulsishah, line 564: Should we add `defer storageClient.Close()` ?

ashmeenkaur: I'll do it in the next PR to avoid running KOKORO again

PR #3708, opened 2025-08-19
refactor: list large dir test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate list large dir package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

[b/439727980](https://b.corp.google.com/issues/439727980)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/list_large_dir/list_large_dir_test.go
+++ b/tools/integration_tests/list_large_dir/list_large_dir_test.go
@@ -72,7 +72,6 @@ func TestMain(m *testing.M) {
 		cfg.ListLargeDir[0].Configs[0].Flags = []string{"--implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1"}
 		cfg.ListLargeDir[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": true}
 		cfg.ListLargeDir[0].Configs[1].Flags = []string{
-			"--implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1",
 			"--client-protocol=grpc --implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1",
 		}
 		cfg.ListLargeDir[0].Configs[1].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": false}
--- a/tools/integration_tests/list_large_dir/list_large_dir_test.go
+++ b/tools/integration_tests/list_large_dir/list_large_dir_test.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


tools/integration_tests/list_large_dir/list_large_dir_test.go
```diff
@@ -40,10 +42,44 @@ var (
...
+		log.Println("No configuration found for list large dir tests in config. Using flags instead.")
+		// Populate the config manually.
+		cfg.ListLargeDir = make([]test_suite.TestConfig, 1)
+		cfg.ListLargeDir[0].TestBucket = setup.TestBucket()
+		cfg.ListLargeDir[0].MountedDirectory = setup.MountedDirectory()
+		cfg.ListLargeDir[0].Configs = make([]test_suite.ConfigItem, 2)
+		cfg.ListLargeDir[0].Configs[0].Flags = []string{"--implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1"}
+		cfg.ListLargeDir[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": true}
+		cfg.ListLargeDir[0].Configs[1].Flags = []string{
+			"--implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1",
```
ashmeenkaur: IIUC, this is duplicate and will run twice with same config.

tools/integration_tests/test_config.yaml
```diff
@@ -44,3 +44,23 @@ implicit_dir:
...
+          flat: true
+          hns: true
+          zonal: true
+      - flags:
+        - "--implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1"
+        - "--client-protocol=grpc --implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1"
+        compatible:
+          flat: true
+          hns: true
+          zonal: false
```
ashmeenkaur: Nit: leave a line

tools/integration_tests/test_config.yaml
```diff
@@ -44,3 +44,23 @@ implicit_dir:
...
+    run_on_gke: false
+    configs:
+      - flags:
+        - "--implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1"
+        compatible:
+          flat: true
+          hns: true
+          zonal: true
+      - flags:
+        - "--implicit-dirs=true --stat-cache-ttl=0 --kernel-list-cache-ttl-secs=-1"
```
ashmeenkaur: same here as well. This config will run twice

PranjalC100: removed this line since config runs twice

tools/integration_tests/list_large_dir/list_large_dir_test.go
PranjalC100: removed this line since config runs twice

ashmeenkaur (approved):

PR #3715, opened 2025-08-20
chore(bufferedread): Add required metrics for buffered reader

vipnydav:
### Description
This PR adds required metrics skeleton for Buffered Reader observability.

https://b.corp.google.com/issues/439762637

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/metrics/metrics.yaml
+++ b/metrics/metrics.yaml
@@ -1,29 +1,7 @@
-- metric-name: "file_cache/read_bytes_count"
-  description: "The cumulative number of bytes read from file cache along with read type - Sequential/Random"
-  unit: "By"
-  type: "int_counter"
-  attributes:
-  - attribute-name: read_type
-    attribute-type: string
-    values: &read_types_list
-    - "Parallel"
-    - "Random"
-    - "Sequential"
-
-- metric-name: "file_cache/read_count"
-  description: "Specifies the number of read requests made via file cache along with type - Sequential/Random and cache hit - true/false"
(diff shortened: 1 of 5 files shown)


metrics/metrics.yaml
```diff
@@ -1,29 +1,7 @@
...
-    values: *read_types_list
-
-- metric-name: "file_cache/read_latencies"
-  description: "The cumulative distribution of the file cache read latencies along with cache hit - true/false."
-  type: "int_histogram"
+- metric-name: "buffered_read/download_block_latency"
+  description: "The cumulative distribution of block download latencies, along with status: successful or cancelled."
   unit: "us"
+  type: "int_histogram"
   boundaries: &common_boundaries
```
kislaykishore, line 5: Could you think about whether the boundaries make sense for your use-case or could we be more specific? Same comment for buffered_read/read_latency

metrics/metrics.yaml
kislaykishore, line 1: Please update the https://cloud.google.com/storage/docs/cloud-storage-fuse/metrics page once this change is released.

metrics/metrics.yaml
vipnydav: Discussed with @raj-prince offline that these boundaries look fine for now. We'll reiterate on it once we start getting metrics and decide if we need to update these.

kislaykishore (approved):

PR #3724, opened 2025-08-21
feat(dnscaching): Implement DNS caching

kislaykishore:
### Description
Implement DNS caching in GCSFuse

b/446170703

### Testing details
1. Manual - Yes
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -144,6 +144,7 @@ func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle m
 		ReadStallRetryConfig:       newConfig.GcsRetries.ReadStall,
 		MetricHandle:               metricHandle,
 		TracingEnabled:             cfg.IsTracingEnabled(newConfig),
+		EnableHTTPDNSCache:         newConfig.GcsConnection.EnableHttpDnsCache,
 	}
 	logger.Infof("UserAgent = %s\n", storageClientConfig.UserAgent)
 	storageHandle, err = storage.NewStorageHandle(context.Background(), storageClientConfig, newConfig.GcsConnection.BillingProject)
--- a/internal/storage/storageutil/client.go
+++ b/internal/storage/storageutil/client.go
@@ -17,6 +17,7 @@ package storageutil
 import (
 	"crypto/tls"
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/httptrace"
 	"strings"
@@ -25,6 +26,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/cfg"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/auth"
 	"github.com/googlecloudplatform/gcsfuse/v3/metrics"
+	dns "github.com/ncruces/go-dns"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.opentelemetry.io/otel"
@@ -68,6 +70,18 @@ type StorageClientConfig struct {
 	MetricHandle metrics.MetricHandle
 
 	TracingEnabled bool
+
+	EnableHTTPDNSCache bool
+}
+
+func getDialerContext(enableHTTPDNSCache bool) func(ctx context.Context, network, address string) (net.Conn, error) {
+	if !enableHTTPDNSCache {
+		return nil
+	}
+	dialer := net.Dialer{
+		Resolver: dns.NewCachingResolver(nil),
+	}
+	return dialer.DialContext
 }
 
 func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.TokenSource) (httpClient *http.Client, err error) {
@@ -75,6 +89,7 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
 	// Using http1 makes the client more performant.
 	if storageClientConfig.ClientProtocol == cfg.HTTP1 {
 		transport = &http.Transport{
+			DialContext:         getDialerContext(storageClientConfig.EnableHTTPDNSCache),
 			Proxy:               http.ProxyFromEnvironment,
 			MaxConnsPerHost:     storageClientConfig.MaxConnsPerHost,
 			MaxIdleConnsPerHost: storageClientConfig.MaxIdleConnsPerHost,
--- a/internal/storage/storageutil/client.go
+++ b/internal/storage/storageutil/client.go
@@ -79,7 +79,7 @@ func getDialerContext(enableHTTPDNSCache bool) func(ctx context.Context, network
 		return nil
 	}
 	dialer := net.Dialer{
-		Resolver: dns.NewCachingResolver(nil),
+		Resolver: dns.NewCachingResolver(nil, dns.MinCacheTTL(1*time.Minute)),
 	}
 	return dialer.DialContext
 }
--- a/internal/storage/storageutil/client_test.go
+++ b/internal/storage/storageutil/client_test.go
@@ -177,3 +177,10 @@ func (t *clientTest) TestCreateHttpClientWithHttpTracing() {
 		return slices.ContainsFunc(ss, func(s tracetest.SpanStub) bool { return s.Name == "http.send" })
 	})
 }
+
+func TestGetDialerContext_nilWhenDNSCacheDisabled(t *testing.T) {
+	assert.Nil(t, getDialerContext(false))
+}
+func TestGetDialerContext_notNilWhenDNSCacheEnabled(t *testing.T) {
+	assert.NotNil(t, getDialerContext(true))
+}
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/jacobsa/syncutil v0.0.0-20180201203307-228ac8e5a6c3
 	github.com/jacobsa/timeutil v0.0.0-20170205232429-577e5acbbcf6
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0
+	github.com/ncruces/go-dns v1.2.7
 	github.com/prometheus/client_golang v1.22.0
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.65.0
--- a/go.sum
+++ b/go.sum
@@ -174,6 +174,8 @@ github.com/minio/minio-go/v7 v7.0.86 h1:DcgQ0AUjLJzRH6y/HrxiZ8CXarA70PAIufXHodP4
 github.com/minio/minio-go/v7 v7.0.86/go.mod h1:VbfO4hYwUu3Of9WqGLBZ8vl3Hxnxo4ngxK4hzQDf4x4=
 github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 h1:C3w9PqII01/Oq1c1nUAm88MOHcQC9l5mIlSMApZMrHA=
 github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822/go.mod h1:+n7T8mK8HuQTcFwEeznm/DIxMOiR9yIdICNftLE1DvQ=
+github.com/ncruces/go-dns v1.2.7 h1:NMA7vFqXUl+nBhGFlleLyo2ni3Lqv3v+qFWZidzRemI=
+github.com/ncruces/go-dns v1.2.7/go.mod h1:SqmhVMBd8Wr7hsu3q6yTt6/Jno/xLMrbse/JLOMBo1Y=
 github.com/pelletier/go-toml/v2 v2.2.4 h1:mye9XuhQ6gvn5h28+VilKrrPoQVanw5PMw/TB0t5Ec4=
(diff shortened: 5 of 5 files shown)


internal/storage/storageutil/client.go
```diff
@@ -70,11 +72,69 @@ type StorageClientConfig struct {
...
+		if err != nil {
+			return nil, fmt.Errorf("dns lookup failed for host %s: %w", host, err)
+		}
+		entry = dnsCacheEntry{host: host, addrs: addrs, expiration: time.Now().Add(dnsCacheTTL)}
+		dnsCache.Store(entry)
+	}
+
+	// Dial any of the resolved addresses.
+	for _, address := range entry.addrs {
+		conn, err := defaultDialer.DialContext(ctx, network, net.JoinHostPort(address, port))
```
ameoghosh: As per discussion today with Yuan, this need change how to give preference to IPv6 over IPv4 in case of dual stack interfaces.

internal/storage/storageutil/client.go
```diff
@@ -70,11 +72,69 @@ type StorageClientConfig struct {
...
+		if entry.host == host && time.Now().Before(entry.expiration) {
+			// Cache hit and is valid.
+		} else {
+			// Cache is for a different host or is expired.
+			cached = nil
+		}
+	}
+
+	// If the cache is empty or invalid, perform a DNS lookup and update the cache.
+	if cached == nil {
```
ameoghosh: this need guarding in case of parallel connections creation called

internal/storage/storageutil/client.go
kislaykishore: Do you know if this is something that already happens in the existing Golang client? I was hoping that since I'm using the DefaultResolver, I should be mostly set. But I haven't gone through the code.

ameoghosh: IIRC it scan all type of addresses and then define precedence rule use IPv6 first over IPv4 or vice versa with an option to override it for add-type.

ameoghosh: based on dialer.FallbackDelay it scans if IPv4 address fails then move to next one could be IPv6

ameoghosh: as we are passing list of address so default selection of IPv4 / IPv6 based on Happy Eyeballs (RFC 8305) will be done by defaultDialer. We can verify if all 3 scenario works with IPv4 Only, IPv6 Only & Dual stack compute where gcsfuse is mounted

internal/storage/storageutil/client.go
```diff
@@ -86,6 +146,7 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
 	} else {
 		// For http2, change in MaxConnsPerHost doesn't affect the performance.
 		transport = &http.Transport{
+			DialContext:       dialContextWithDNSCache,
```
liuyuan10: no need to do it for http2. http2 is able to reuse connections without draining

internal/storage/storageutil/client.go
```diff
@@ -86,6 +146,7 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
...
 			Proxy:             http.ProxyFromEnvironment,
 			DisableKeepAlives: true,
 			MaxConnsPerHost:   storageClientConfig.MaxConnsPerHost,
```
liuyuan10, line 106: does the httpclient on line 165 need the cache?

internal/storage/storageutil/client.go
```diff
@@ -70,11 +72,69 @@ type StorageClientConfig struct {
...
+	return nil, fmt.Errorf("failed to dial any of a host's addresses for %s", host)
+}
+
 func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.TokenSource) (httpClient *http.Client, err error) {
 	var transport *http.Transport
 	// Using http1 makes the client more performant.
 	if storageClientConfig.ClientProtocol == cfg.HTTP1 {
 		transport = &http.Transport{
+			DialContext: func() func(context.Context, string, string) (net.Conn, error) {
+				storageClientConfig.MetricHandle.GcsDnsCount(1)
```
liuyuan10: will this metric be visable in internal viceroy dashboard? or cloud monitoring dashboard?

internal/storage/storageutil/client.go
```diff
@@ -70,11 +72,69 @@ type StorageClientConfig struct {
...
+	}
+
+	// Atomically load the cached entry.
+	cached := dnsCache.Load()
+	var entry dnsCacheEntry
+
+	// Check if the cache is valid for the current host and not expired.
+	if cached != nil {
+		entry = cached.(dnsCacheEntry)
+		if entry.host == host && time.Now().Before(entry.expiration) {
```
liuyuan10: I know in GCS probably it's always the same host. but coud you add a log if the cached entry is not the same as host? So we know something unexpected is happening.

internal/storage/storageutil/client.go
liuyuan10: this is a local variable, already safe?

internal/storage/storageutil/client.go
```diff
@@ -68,13 +70,37 @@ type StorageClientConfig struct {
...
+		KeepAlive: 30 * time.Second,
+	}
+	return getDialerContextWithDialer(dnsCacheTTLSecs, &dialer)
+}
+
+func getDialerContextWithDialer(dnsCacheTTLSecs int64, dialer *net.Dialer) func(ctx context.Context, network, address string) (net.Conn, error) {
+	dnsTTLSecOption := dns.MaxCacheTTL(time.Second * time.Duration(dnsCacheTTLSecs))
+	dialer.Resolver = &net.Resolver{
+		PreferGo: true,
+		Dial:     dns.NewCachingDialer(dialer.Resolver.Dial, dnsTTLSecOption),
```
liuyuan10: isn't this "dialer.Resolver" nil by default?

internal/storage/storageutil/client.go
```diff
@@ -68,13 +70,37 @@ type StorageClientConfig struct {
...
+
+	HTTPDNSCacheTTLSecs int64
+}
+
+func getDialerContext(dnsCacheTTLSecs int64) func(ctx context.Context, network, address string) (net.Conn, error) {
+	if dnsCacheTTLSecs == 0 {
+		return nil
+	}
+	dialer := net.Dialer{
+		Timeout:   30 * time.Second,
```
liuyuan10: these are not set originally. why introducing them as part of this change?

internal/storage/storageutil/client.go
```diff
@@ -68,13 +70,37 @@ type StorageClientConfig struct {
...
+	return dialer.DialContext
 }
 
 func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.TokenSource) (httpClient *http.Client, err error) {
 	var transport *http.Transport
+
 	// Using http1 makes the client more performant.
 	if storageClientConfig.ClientProtocol == cfg.HTTP1 {
 		transport = &http.Transport{
+			DialContext:         getDialerContext(storageClientConfig.HTTPDNSCacheTTLSecs),
```
liuyuan10: does the httpclient on line 130 need this cache?

cfg/config.go
```diff
@@ -563,6 +565,12 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.DurationP("http-client-timeout", "", 0*time.Nanosecond, "The time duration that http client will wait to get response from the server. A value of 0 indicates no timeout.")
 
+	flagSet.IntP("http-dns-cache-ttl-secs", "", 0, "Sets the DNS cache TTL to the specified value. Special values: -1 means infinite TTL and 0 disables caching")
```
liuyuan10: can you make it clear that it only applies to http1? maybe just name it http1-dns-cahce-ttl-secs

liuyuan10: so the default is "0" (cache disabled)?

liuyuan10: why "-1" means infinite TTL? time.Duration(-1) means -1ns, and will set all ttl to this if I understand the code correctly

internal/storage/storageutil/client.go
kislaykishore: Instead of using a caching dialer, I'm now using a CachingResolver.

internal/storage/storageutil/client.go
kislaykishore: It might - I'll still want to keep it out of the scope for this PR and handle it separately since it's not a common use-case.

cfg/config.go
kislaykishore: I've kept the name as http-* for consistency reasons since that's how the other HTTP/1 flags have been named. HTTP/2, while supported is not a common use-case in GCSFuse.

kislaykishore: Obsolete.

internal/storage/storageutil/client.go
kislaykishore: obsolete

internal/storage/storageutil/client.go
kislaykishore: I had introduced this metric exclusively for testing. I've now removed it.

internal/storage/storageutil/client.go
kislaykishore: That's not a common use-case so I'll shelve that for later.

internal/storage/storageutil/client.go
kislaykishore: Obsolete.

internal/storage/storageutil/client_test.go
```diff
@@ -177,3 +177,10 @@ func (t *clientTest) TestCreateHttpClientWithHttpTracing() {
 		return slices.ContainsFunc(ss, func(s tracetest.SpanStub) bool { return s.Name == "http.send" })
 	})
 }
+
+func TestGetDialerContext_nilWhenDNSCacheDisabled(t *testing.T) {
+	assert.Nil(t, getDialerContext(false))
+}
+func TestGetDialerContext_notNilWhenDNSCacheEnabled(t *testing.T) {
+	assert.NotNil(t, getDialerContext(true))
+}
```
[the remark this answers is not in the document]
kislaykishore: I prefer not having if-else in my tests.

internal/storage/storageutil/client.go
```diff
@@ -68,13 +70,26 @@ type StorageClientConfig struct {
...
+	}
+	return dialer.DialContext
 }
 
 func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.TokenSource) (httpClient *http.Client, err error) {
 	var transport *http.Transport
 	// Using http1 makes the client more performant.
 	if storageClientConfig.ClientProtocol == cfg.HTTP1 {
 		transport = &http.Transport{
+			DialContext:         getDialerContext(storageClientConfig.EnableHTTPDNSCache),
```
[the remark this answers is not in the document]
kislaykishore: No, it's only relevant for HTTP/1

cmd/legacy_main.go
```diff
@@ -144,6 +144,7 @@ func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle m
 		ReadStallRetryConfig:       newConfig.GcsRetries.ReadStall,
 		MetricHandle:               metricHandle,
 		TracingEnabled:             cfg.IsTracingEnabled(newConfig),
+		EnableHTTPDNSCache:         newConfig.GcsConnection.EnableHttpDnsCache,
```
ameoghosh, line 147: Will this flag be passed by GKE when host networking is used ?

kislaykishore: No, we will turn DNS caching on by default.

internal/storage/storageutil/client.go
```diff
@@ -68,13 +70,26 @@ type StorageClientConfig struct {
...
+
+	EnableHTTPDNSCache bool
+}
+
+func getDialerContext(enableHTTPDNSCache bool) func(ctx context.Context, network, address string) (net.Conn, error) {
+	if !enableHTTPDNSCache {
+		return nil
+	}
+	dialer := net.Dialer{
+		Resolver: dns.NewCachingResolver(nil),
```
ameoghosh: Do we need explicit setting of TTL here ?

kislaykishore: Nope, the library takes care of it: https://github.com/ncruces/go-dns/blob/97163c1ea04a9a5fca3b54e97d722e79042e35c9/cache.go#L99

cfg/config.go
kislaykishore: Obsolete.

internal/storage/storageutil/client.go
ameoghosh: I am wondering if TTL need to explicitly set as part of option, if default behaviour seams to be 0 (unlimited)

below options i am referring to:
func (o maxTTLOption) apply(c *cache)        { c.maxTTL = time.Duration(o) }
func (o minTTLOption) apply(c *cache)        { c.minTTL = time.Duration(o) }

kislaykishore: Discussed offline - in some cases the DNS resolution might not have TTLs. In such cases, let's keep a min-ttl to avoid frequent DNS lookups.

cmd/legacy_main.go
charith87: This is a good suggestion and reduces blast radius in case we see any issue in the perf tests that are planned before making it default.

PR #3728, opened 2025-08-21
refactor: cloud profiler test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate cloud profiler package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

[b/445951569](https://buganizer.corp.google.com/issues/445951569)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/cloud_profiler/cloud_profiler_test.go
+++ b/tools/integration_tests/cloud_profiler/cloud_profiler_test.go
@@ -58,16 +58,10 @@ func TestMain(m *testing.M) {
 	if len(cfg.CloudProfiler) == 0 {
 		log.Println("No configuration found for cloud profiler tests in config. Using flags instead.")
 
-		if setup.MountedDirectory() != "" {
-			if setup.ProfileLabelForMountedDirTest() == "" {
-				log.Fatal("Profile label should have been provided for mounted directory test.")
-			}
-			testServiceVersion = setup.ProfileLabelForMountedDirTest()
-		}
 		// Populate the config manually.
 		cfg.CloudProfiler = make([]test_suite.TestConfig, 1)
 		cfg.CloudProfiler[0].TestBucket = setup.TestBucket()
-		cfg.CloudProfiler[0].MountedDirectory = setup.MountedDirectory()
+		cfg.CloudProfiler[0].GKEMountedDirectory = setup.MountedDirectory()
 		cfg.CloudProfiler[0].Configs = make([]test_suite.ConfigItem, 1)
 		cfg.CloudProfiler[0].Configs[0].Flags = []string{
 			"--enable-cloud-profiler --cloud-profiler-cpu --cloud-profiler-heap --cloud-profiler-goroutines --cloud-profiler-mutex --cloud-profiler-allocated-heap",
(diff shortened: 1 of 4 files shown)


tools/integration_tests/cloud_profiler/cloud_profiler_test.go
```diff
@@ -50,8 +53,38 @@ var (
...
 
-	var storageClient *storage.Client
-	ctx := context.Background()
+	// 1. Load and parse the common configuration.
+	cfg := test_suite.ReadConfigFile(setup.ConfigFile())
+	if len(cfg.CloudProfiler) == 0 {
+		log.Println("No configuration found for write large files tests in config. Using flags instead.")
+		if setup.ProfileLabelForMountedDirTest() == "" {
+			log.Fatal("Profile label should have been provided for mounted directory test.")
+		}
```
ashmeenkaur: can we move this to the config file too since this is something that GKE team needs to pass. This will be an optional config similar to log file

tools/integration_tests/cloud_profiler/cloud_profiler_test.go
```diff
@@ -50,8 +53,38 @@ var (
 func TestMain(m *testing.M) {
 	setup.ParseSetUpFlags()
 
-	var storageClient *storage.Client
-	ctx := context.Background()
+	// 1. Load and parse the common configuration.
+	cfg := test_suite.ReadConfigFile(setup.ConfigFile())
+	if len(cfg.CloudProfiler) == 0 {
+		log.Println("No configuration found for write large files tests in config. Using flags instead.")
```
ashmeenkaur:
```suggestion
		log.Println("No configuration found for cloud profiler tests in config. Using flags instead.")
```

tools/integration_tests/cloud_profiler/cloud_profiler_test.go
ashmeenkaur: also this check is incorrect. This should only be set when mounted directory is set.

tools/integration_tests/cloud_profiler/cloud_profiler_test.go
```diff
@@ -50,8 +53,42 @@ var (
...
-	ctx := context.Background()
+	// 1. Load and parse the common configuration.
+	cfg := test_suite.ReadConfigFile(setup.ConfigFile())
+	if len(cfg.CloudProfiler) == 0 {
+		log.Println("No configuration found for cloud profiler tests in config. Using flags instead.")
+
+		if setup.MountedDirectory() != "" {
+			if setup.ProfileLabelForMountedDirTest() == "" {
+				log.Fatal("Profile label should have been provided for mounted directory test.")
+			}
```
ashmeenkaur: Is this check redundant, given that we are checking before running mounted directory tests?

tools/integration_tests/cloud_profiler/cloud_profiler_test.go
```diff
@@ -60,24 +97,33 @@ func TestMain(m *testing.M) {
...
 	testServiceVersion = fmt.Sprintf("ve2e0.0.0-%s", strings.ReplaceAll(uuid.New().String(), "-", "")[:8])
+	logger.Infof("Enabling cloud profiler with version tag: %s", testServiceVersion)
+
+	// Run tests for testBucket
+	// 4. Build the flag sets dynamically from the config.
+	flags := setup.BuildFlagSets(cfg.CloudProfiler[0], bucketType)
 
-	flags := [][]string{
-		{"--enable-cloud-profiling", "--profiling-cpu", "--profiling-heap", "--profiling-goroutines", "--profiling-mutex", "--profiling-allocated-heap", fmt.Sprintf("--profiling-label=%s", testServiceVersion)},
+	// Iterating over flagSets and updating any empty "--profiling-label=" flags.
```
ashmeenkaur: Instead of iterating, can we just do a string.Replace before building flagset?

tools/integration_tests/cloud_profiler/cloud_profiler_test.go
```diff
@@ -60,24 +97,33 @@ func TestMain(m *testing.M) {
 		}
 	}()
 
-	if setup.MountedDirectory() != "" {
+	// 3. To run mountedDirectory tests, we need both testBucket and mountedDirectory
+	if cfg.CloudProfiler[0].MountedDirectory != "" {
 		if setup.ProfileLabelForMountedDirTest() == "" {
```
ashmeenkaur: Can we have a way to pick this up from config file instead of flag?

PranjalC100: yes i've added a function that does the same.

ashmeenkaur (approved):

PR #3730, opened 2025-08-22
refactor: rename dir limit test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate read dir limit package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of rename dir limit package so it can use config file.

[b/439724878](https://b.corp.google.com/issues/439724878)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/rename_dir_limit/rename_dir_limit_test.go
+++ b/tools/integration_tests/rename_dir_limit/rename_dir_limit_test.go
@@ -62,16 +62,15 @@ func TestMain(m *testing.M) {
 		cfg.RenameDirLimit[0].MountedDirectory = setup.MountedDirectory()
 		cfg.RenameDirLimit[0].Configs = make([]test_suite.ConfigItem, 2)
 		cfg.RenameDirLimit[0].Configs[0].Flags = []string{
-			"--rename-dir-limit=3 --implicit-dirs",
 			"--rename-dir-limit=3 --implicit-dirs --client-protocol=grpc",
 			"--rename-dir-limit=3",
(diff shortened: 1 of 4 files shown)


ashmeenkaur (commented):
tools/integration_tests/test_config.yaml change missing

PranjalC100:
> tools/integration_tests/test_config.yaml change missing

added! didn't realize hadn't added it...

tools/integration_tests/rename_dir_limit/rename_dir_limit_test.go
```diff
@@ -51,42 +52,69 @@ var (
...
-	setup.AppendFlagsToAllFlagsInTheFlagsSet(&flags, "", "--client-protocol=grpc")
-	if hnsFlagSet, err := setup.AddHNSFlagForHierarchicalBucket(ctx, storageClient); err == nil {
-		flags = [][]string{hnsFlagSet}
-	}
-	setup.ExitWithFailureIfBothTestBucketAndMountedDirectoryFlagsAreNotSet()
-
 	if setup.TestBucket() == "" && setup.MountedDirectory() != "" {
 		log.Print("Please pass the name of bucket mounted at mountedDirectory to --testBucket flag.")
 		os.Exit(1)
 	}
```
ashmeenkaur: We can clean up this check as tests now run with config file

ashmeenkaur (dismissed):

ashmeenkaur (approved):

Issue #3705, opened 2025-08-19
Custom Endpoints prevent mount with HNS enabled


ivopatty:
**Describe the issue**
The hidden flag `--enable-hns` for Hierarchical Namespace support in combination with the specification of a Custom Endpoint for PSC usage prevents bucket mounting 

**System & Version (please complete the following information):**
 - OS: RHEL8/Fedora
 - Platform: vSphere VM & Fedora Docker Image
 - Version 3.2.0

**Steps to reproduce the behavior with following information:**
**Note: ** This describes the full use-case tested, in validation of the issue, specifying the default url of https://storage.googleapis.com/storage/v1 also reproduces the issue through public internet
1. Have a GCS Bucket with HNS **disabled**
2. Enable Private Service Connect to retrieve a PSC URL and have it linked through CloudVPN
3. Get a Service Account with JSON key for authentication
4. Configure GCS Fuse with the flags: `gcsfuse --implicit-dirs --allow_other --key-file=/path/to/key --custom-endpoint=storage.p.googleapis.com/storage/v1/`
5. FUSE mount fails
 
**Additional context**
Setting the hidden flag --enable-hns explicitly to `false` resolves the issue.

Error logging during the mount process

```
{"timestamp":{"seconds":1755588312,"nanos":565522854},"severity":"INFO","message":"Error occurred during command execution on gcsfuse/3.2.0 (Go version go1.24.5): mountWithStorageHandle: fs.NewServer: create file system: SetUpBucket: BucketHandle: storageLayout call failed: context deadline exceeded"}
{"timestamp":{"seconds":1755588361,"nanos":830359596},"severity":"ERROR","message":"Error while mounting gcsfuse: mountWithStorageHandle: fs.NewServer: create file system: SetUpBucket: BucketHandle: storageLayout call failed: context deadline exceeded\n"}
{"timestamp":{"seconds":1755588361,"nanos":830620850},"severity":"ERROR","message":"Failed to signal error to parent-process from daemon: write DAEMONIZE_STATUS_FD: broken pipe"}
{"timestamp":{"seconds":1755588361,"nanos":830701166},"severity":"INFO","message":"Error occurred during command execution on gcsfuse/3.2.0 (Go version go1.24.5): mountWithStorageHandle: fs.NewServer: create file system: SetUpBucket: BucketHandle: storageLayout call failed: context deadline exceeded"}
```

abhishek10004:
Hi Ivo,

[...]

PR #3735, opened 2025-08-25
fix: Fix custom-endpoint format in help doc

Tulsishah:
### Description
The help documentation for the GCS Fuse `--custom-endpoint` flag is causing user confusion and leading to mount failures because it does not specify the correct endpoint format.
Ref issue - https://github.com/GoogleCloudPlatform/gcsfuse/issues/3705

I recently found(b/429928623) that, standard URL formats results in errors. The only format that has been found to work reliably is `storage.<UNIVERSE_DOMAIN>:443`. Update the documentation with this URL format.

[b/440981458](https://b.corp.google.com/issues/440981458)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -670,7 +670,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data. The custom endpoint must support the equivalent resources and operations as the GCS JSON endpoint, https://storage.googleapis.com/storage/v1. If a custom endpoint is not specified, GCSFuse uses the global GCS JSON API endpoint, https://storage.googleapis.com/storage/v1.")
+	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data.  The custom endpoint must support resources equivalent to storage.googleapis.com:443.  If a custom endpoint is not specified, GCSFuse uses the default endpoint, which is storage.googleapis.com:443.")
 
 	flagSet.BoolP("debug_fs", "", false, "This flag is unused.")
 
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -670,7 +670,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data.  The custom endpoint must support resources equivalent to storage.googleapis.com:443.  If a custom endpoint is not specified, GCSFuse uses the default endpoint, which is storage.googleapis.com:443.")
+	flagSet.StringP("custom-endpoint", "", "", "To specify a custom storage endpoint, ensure it supports the same resources as the default storage.googleapis.com:443 and includes the port number.")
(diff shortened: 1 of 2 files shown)


cfg/params.yaml
```diff
@@ -326,9 +326,9 @@
   usage: >-
     Specifies an alternative custom endpoint for fetching data. The custom endpoint
     must support the equivalent resources and operations as the GCS JSON endpoint,
-    https://storage.googleapis.com/storage/v1. If a custom endpoint is not
+    storage.<UNIVERSE_DOMAIN>:443. If a custom endpoint is not
```
raj-prince: Do we have any public doc around universe domain? Linking here would be good for customer.

Tulsishah: No we have internal documentation only that too universe specific like GCS, TPC.

kislaykishore (dismissed):

cfg/config.go
```diff
@@ -670,7 +670,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data. The custom endpoint must support the equivalent resources and operations as the GCS JSON endpoint, https://storage.googleapis.com/storage/v1. If a custom endpoint is not specified, GCSFuse uses the global GCS JSON API endpoint, https://storage.googleapis.com/storage/v1.")
+	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data.  The custom endpoint must support resources equivalent to storage.googleapis.com:443.  If a custom endpoint is not specified, GCSFuse uses the default endpoint, which is storage.googleapis.com:443.")
```
raj-prince: Lets add: The port number is required for the gRPC-based control client used at mount to determine the bucket type (HNS or Flat). Ref: https://pkg.go.dev/google.golang.org/api/option#WithEndpoint

cfg/config.go
```diff
@@ -670,7 +670,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data. The custom endpoint must support the equivalent resources and operations as the GCS JSON endpoint, https://storage.googleapis.com/storage/v1. If a custom endpoint is not specified, GCSFuse uses the global GCS JSON API endpoint, https://storage.googleapis.com/storage/v1.")
+	flagSet.StringP("custom-endpoint", "", "", "Specifies an alternative custom endpoint for fetching data.  The custom endpoint must support resources equivalent to storage.googleapis.com:443.  The port number is required for the gRPC-based control client used at mount time to determine the bucket type (HNS or Flat).  If a custom endpoint is not specified, GCSFuse uses the default endpoint, which is storage.googleapis.com:443. Ref: https://pkg.go.dev/google.golang.org/api/option#WithEndpoint")
```
raj-prince: How about this?

To specify a custom storage endpoint, ensure it supports the same resources as the default `storage.googleapis.com:443` and includes the port number.

I think, just explicitly saying to include port number should be sufficient.

raj-prince (approved):

PR #3736, opened 2025-08-25
chore(zonal buckets): Revert zero byte reader logic to get size of unfinalized object

anushka567:
### Description
This PR reverts the logic for creating a zero byte reader for getting the size of unfinalized objects. This is done as the size of unfinalized object being returned by the stat calls is correct, hence eliminating the need for such a logic.

This PR reverts: 
- PR#3387
- PR#3632

[b/440226859](b/440226859)

### Testing details
1. Manual - Yes. For objects that were created after the support rolled out, correct size is being returned. 
2. Unit tests - NA


tools/integration_tests/rapid_appends/reads_after_appends_test.go
```diff
@@ -80,7 +80,7 @@ func readRandomlyAndVerify(t *testing.T, filePath string, expectedContent []byte
 ////////////////////////////////////////////////////////////////////////
 
 func (t *CommonAppendsSuite) TestAppendsAndReads() {
-	const metadataCacheTTLSecs = 10
+	const metadataCacheTTLSecs = 180
```
anushka567: For dual mount scenario, the metacache ttl has to be increased to stabilize the test (prevent metadata cache from expiring before expected while we wait for updated size to be returned from stat calls)

anushka567: For context- with every update to the unfinalized object, we wait for 1 minute to make sure that the stat calls return the updated size.

anushka567:
Since the tests were timing out when common high value of metadata cache ttl was used , I have split the tests into single mount and dual mount to prevent the same . Along with that, ensured minimum possible sleeps.

anushka567:
All tests are stable and passing now. There were some necessary changes made to stabilize the tests, refactor for those will be collectively taken care of in PR #3723

tools/integration_tests/rapid_appends/reads_after_appends_test.go
```diff
@@ -126,7 +127,88 @@ func (t *CommonAppendsSuite) TestAppendsAndReads() {
 			t.mountPrimaryMount(scenario.flags)
 			defer t.unmountPrimaryMount()
 
-			log.Printf("Running tests with flags: %v", scenario.flags)
+			log.Printf("Running single-mount tests with flags: %v", scenario.flags)
+
+			for _, tc := range testCases {
```
vadlakondaswetha, line 132: giving LGTM assuming the refactoring for all these tests will be done properly in the next PR

vadlakondaswetha (approved):

PR #3739, opened 2025-08-25
chore(logging): Add warning log for unsupported log format

alleaditya:
### Description
Introduce a warning log when an unsupported log format is specified, defaulting to the existing format. This enhances logging robustness by providing feedback on invalid configurations.

b/328992990

### Testing details
1. Manual - Tested with GKE logs ([Screenshot](https://screenshot.googleplex.com/AaPssUb9LD8ghUk))
   * No changes observed for valid formats supported currently (`text`, `json`)

None

### DIFF
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -15,6 +15,7 @@
 package cfg
 
 import (
+	"log"
 	"math"
 	"net/url"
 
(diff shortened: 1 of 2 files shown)


cfg/constants.go
```diff
@@ -32,6 +32,12 @@ const (
 	OFF     string = "OFF"
 )
 
+const (
+	// Logging-format constants
+	LoggingFormatText = "text"
+	LoggingFormatJson = "json"
```
kislaykishore: Golang recommends abbreviations to have the same casing.
```suggestion
	LoggingFormatJSON = "json"
```

internal/logger/logger.go
```diff
@@ -114,6 +114,10 @@ func SetLogFormat(format string) {
 	if format == defaultLoggerFactory.format {
 		return
 	}
+
+	if format != cfg.LoggingFormatText && format != cfg.LoggingFormatJSON {
+		Warnf("Unsupported log format: %s. Defaulting to %s log format.", format, defaultLoggerFactory.format)
+	}
```
ashmeenkaur:
```suggestion
if format != cfg.LoggingFormatText && format != cfg.LoggingFormatJSON {
    Warnf("Unsupported log format: %s. Defaulting to %s log format.", format, defaultLoggerFactory.format)
    format = defaultLoggerFactory.format
}

 if format == defaultLoggerFactory.format {
    // If the format is already set to the desired value, do nothing and return early. This avoid creating a new logger.
    return
}
```

alleaditya: Moving this to rationalize.go to remain in sync with other logging configurations.

PR #3745, opened 2025-08-26
refactor: read large files test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate read large files test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of read large files package so it can use config file.

[b/445948730](https://b.corp.google.com/issues/445948730)

### Testing details
2. Unit tests - NA


tools/integration_tests/read_large_files/read_large_files_test.go
```diff
@@ -41,70 +41,60 @@ var (
...
+		// Populate the config manually.
+		cfg.ReadLargeFiles = make([]test_suite.TestConfig, 1)
+		cfg.ReadLargeFiles[0].TestBucket = setup.TestBucket()
+		cfg.ReadLargeFiles[0].MountedDirectory = setup.MountedDirectory()
+		cfg.ReadLargeFiles[0].Configs = make([]test_suite.ConfigItem, 1)
+		cfg.ReadLargeFiles[0].Configs[0].Flags = []string{
+			"--implicit-dirs",
+			"--implicit-dirs --client-protocol=grpc",
+		}
+		cacheDirFlag1 := fmt.Sprintf("--implicit-dirs=true --file-cache-max-size-mb=700 --cache-file-for-range-read=true --cache-dir=%s/cache-dir-read-large-files-hns", os.TempDir())
```
PranjalC100: cache-dir path seems fine?

ashmeenkaur: We don't want cache dir to collide between hns and regional bucket runs. Can we set it to a random string in these tests similar to: https://github.com/GoogleCloudPlatform/gcsfuse/blob/21bfce33c73e9616707089096612d7484b0d1f99/tools/integration_tests/readonly/readonly_test.go#L115

tools/integration_tests/read_large_files/read_large_files_test.go
```diff
@@ -41,70 +41,60 @@ var (
...
+
+	// 3. To run mountedDirectory tests, we need both testBucket and mountedDirectory
+	// flags to be set, as ReadLargeFiles tests validates content from the bucket.
+	if cfg.ReadLargeFiles[0].MountedDirectory != "" && cfg.ReadLargeFiles[0].TestBucket != "" {
+		os.Exit(setup.RunTestsForMountedDirectory(cfg.ReadLargeFiles[0].MountedDirectory, m))
 	}
 
-	// Run tests for mountedDirectory only if --mountedDirectory flag is set.
-	setup.RunTestsForMountedDirectoryFlag(m)
+	// Run tests for testBucket// Run tests for testBucket
```
ashmeenkaur:
```suggestion
	// Run tests for testBucket.
```

tools/integration_tests/test_config.yaml
```diff
@@ -95,6 +95,25 @@ gzip:
           hns: true
           zonal: true
 
+read_large_files:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "${BUCKET_NAME}"
+    log_file: # Optional
+    run_on_gke: false # Not required by read large files tests.
```
```suggestion
    log_file:  # Not required by read large files tests.
    run_on_gke: false
```

tools/integration_tests/test_config.yaml
```diff
@@ -95,6 +95,25 @@ gzip:
...
+    configs:
+      - flags:
+          - "--implicit-dirs"
+          - "--implicit-dirs --client-protocol=grpc"
+          # Set the 'TMPDIR' environment variable to an absolute path, for example, 'export TMPDIR=/home/vm_name/gcsfuse/tempdir', before running the test.
+          - "--implicit-dirs=true --file-cache-max-size-mb=700 --cache-file-for-range-read=true --cache-dir={CACHE_DIR_PATH}"
+          - "--implicit-dirs=true --file-cache-max-size-mb=700 --cache-file-for-range-read=true --cache-dir={CACHE_DIR_PATH} --client-protocol=grpc"
+          - "--implicit-dirs=true --file-cache-max-size-mb=-1 --cache-file-for-range-read=false --cache-dir={CACHE_DIR_PATH}"
+          - "--implicit-dirs=true --file-cache-max-size-mb=-1 --cache-file-for-range-read=false --cache-dir={CACHE_DIR_PATH} --client-protocol=grpc"
```
ashmeenkaur: Missing `$` prefix before environment variable.
Also thinking a bit more about this, I think there is no need for GCSFuse team to pass this path.
Can we add a manual check in the setup file for both read_large_files and readonly package that if this is empty, we populate the value ourselves. 
For mounted directory tests we throw a fatal log.

PranjalC100: okay got it!
will implement the same

tools/integration_tests/read_large_files/read_large_files_test.go
PranjalC100: got it!

tools/integration_tests/read_large_files/read_large_files_test.go
```diff
@@ -43,6 +44,35 @@ var (
...
+	// The regex explicitly looks for "--cache-dir=" followed by optional whitespace (\s*)
+	// and then either another space or the end of the line ($).
+	// This ensures that it will not match flags that already have a value.
+	re := regexp.MustCompile(`--cache-dir=\s*(\s|$)`)
+
+	for _, flags := range flagSets {
+		var updatedFlags []string
+		for _, flag := range flags {
+			// Check if the flag string contains the specific pattern.
+			if re.MatchString(flag) {
```
ashmeenkaur: couldn't we have done an exact comparison here with "--cache-dir=" string?
Can we also move this function to setup.go so other packages can use it too.

tools/integration_tests/test_config.yaml
```diff
@@ -95,6 +95,25 @@ gzip:
...
+read_large_files:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "${BUCKET_NAME}"
+    log_file:  # Not required by read large files tests.
+    run_on_gke: false
+    configs:
+      - flags:
+          - "--implicit-dirs"
+          - "--implicit-dirs --client-protocol=grpc"
+          # Set the 'TMPDIR' environment variable to an absolute path, for example, 'export TMPDIR=/home/vm_name/gcsfuse/tempdir', before running the test.
```
```suggestion
```

tools/integration_tests/read_large_files/read_large_files_test.go
```diff
@@ -41,70 +41,61 @@ var (
...
+		// Populate the config manually.
+		cfg.ReadLargeFiles = make([]test_suite.TestConfig, 1)
+		cfg.ReadLargeFiles[0].TestBucket = setup.TestBucket()
+		cfg.ReadLargeFiles[0].MountedDirectory = setup.MountedDirectory()
+		cfg.ReadLargeFiles[0].Configs = make([]test_suite.ConfigItem, 1)
+		cfg.ReadLargeFiles[0].Configs[0].Flags = []string{
+			"--implicit-dirs",
+			"--implicit-dirs --client-protocol=grpc",
+		}
+		cacheDirFlag1 := fmt.Sprintf("--implicit-dirs=true --file-cache-max-size-mb=700 --file-cache-cache-file-for-range-read=true --cache-dir=%s/cache-dir-read-large-files-%s", os.TempDir(), setup.GenerateRandomString(4))
```
ashmeenkaur: are cacheDirFlag1 & cacheDirFlag2 still needed now that we are populating cache dir if it's not already there?

PranjalC100: yes, so  --file-cache-max-size-mb=700 --file-cache-cache-file-for-range-read=true values are being changed in the 2 flags.
I should have used a better name for the flag instead of cacheDirFlag1/2 to avoid confusion.

ashmeenkaur (approved):

PR #3747, opened 2025-08-26
chore(logging): Add log for unsupported log formats

alleaditya:
### Description
Introduce a log message when an unsupported log format is specified, defaulting to the existing `JSON` format. This enhances logging robustness by providing feedback on invalid configurations.

[b/328992990](https://b.corp.google.com/issues/328992990)

### Testing details
1. Manual - Tested with GKE logs ([Screenshot](https://screenshot.googleplex.com/DJTZFTrAj3xGjrh))
   * No changes observed for valid formats supported currently (`text`, `json`)
 2. Unit Tests - ✅ 

None

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -443,7 +443,7 @@ func (p *BufferedReader) scheduleNextBlock(urgent bool) error {
 		// can't get a block. For the buffered reader, this is a recoverable
 		// condition that should either trigger a fallback to another reader (for
 		// urgent reads) or be ignored (for background prefetches).
-		logger.Warnf("scheduleNextBlock: could not get block from pool (urgent=%t): %v", urgent, err)
+		logger.Tracef("scheduleNextBlock: could not get block from pool (urgent=%t): %v", urgent, err)
 		return ErrPrefetchBlockNotAvailable
 	}
 


cfg/constants.go
```diff
@@ -32,6 +32,13 @@ const (
 	OFF     string = "OFF"
 )
 
+const (
+	// Logging-format constants
+	LogFormatText    = "text"
```
kislaykishore: Can we make them not-exported constants?

cfg/rationalize_test.go
```diff
@@ -664,3 +665,63 @@ func TestRationalize_FileCacheAndBufferedReadConflict(t *testing.T) {
...
+
+func TestResolveLoggingConfig(t *testing.T) {
+	testCases := []struct {
+		name              string
+		config            *Config
+		expectedLogFormat string
+		expectWarning     bool
+	}{
+		{
+			name: "valid log format (json)",
```
kislaykishore: Please set names without spaces and special characters e.g. valid_log_format_json. This makes it easier to run specific tests from the command line.

cfg/rationalize_test.go
```diff
@@ -664,3 +665,63 @@ func TestRationalize_FileCacheAndBufferedReadConflict(t *testing.T) {
 		})
 	}
 }
+
+func TestResolveLoggingConfig(t *testing.T) {
```
kislaykishore: Seems excessive to me to test for the logged string. Checking the format should be good enough. Looking for the log string makes it a change detector test.

cfg/constants.go
alleaditya: Done

cfg/rationalize_test.go
alleaditya: Removed the logging check.

cfg/rationalize_test.go
alleaditya: Renamed.

Wondering if we are planning to do a refactor to get this uniform across test case naming?

cfg/rationalize_test.go
```diff
@@ -664,3 +664,48 @@ func TestRationalize_FileCacheAndBufferedReadConflict(t *testing.T) {
...
+	testCases := []struct {
+		name              string
+		config            *Config
+		expectedLogFormat string
+		expectWarning     bool
```
kislaykishore: Remove this.

PR #3754, opened 2025-08-26
refactor: gzip test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate gzip test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of gzip package so it can use config file.

[b/445942883](https://b.corp.google.com/issues/445942883)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/gzip/gzip_test.go
+++ b/tools/integration_tests/gzip/gzip_test.go
@@ -249,18 +249,18 @@ func TestMain(m *testing.M) {
 	}()
 
 	// 3. To run mountedDirectory tests, we need both testBucket and mountedDirectory
-	// flags to be set, as WriteLargeFiles tests validates content from the bucket.
-	if cfg.WriteLargeFiles[0].MountedDirectory != "" && cfg.WriteLargeFiles[0].TestBucket != "" {
-		os.Exit(setup.RunTestsForMountedDirectory(cfg.WriteLargeFiles[0].MountedDirectory, m))
+	// flags to be set, as Gzip tests validates content from the bucket.
+	if cfg.Gzip[0].MountedDirectory != "" && cfg.Gzip[0].TestBucket != "" {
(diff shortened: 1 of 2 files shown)


tools/integration_tests/gzip/gzip_test.go
```diff
@@ -193,25 +194,42 @@ func createContentOfSize(contentSize int) (string, error) {
...
-	flags := [][]string{commonFlags}
-
-	if !testing.Short() {
-		gRPCFlags := append(commonFlags, "--client-protocol=grpc")
-		flags = append(flags, gRPCFlags)
-	}
-
 	setup.ExitWithFailureIfBothTestBucketAndMountedDirectoryFlagsAreNotSet()
 
 	if setup.TestBucket() == "" && setup.MountedDirectory() != "" {
```
ashmeenkaur: we can remove these checks as now we will be using configs.

tools/integration_tests/test_config.yaml
```diff
@@ -80,3 +80,17 @@ write_large_files:
           flat: true
           hns: true
           zonal: true
+
+gzip:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "${BUCKET_NAME}"
+    log_file: # Optional
```
ashmeenkaur: Nit: Instead of writing Optional, can we write "Not Required by gzip tests" instead

tools/integration_tests/gzip/gzip_test.go
PranjalC100: removed!

tools/integration_tests/test_config.yaml
PranjalC100: changed, will do the same if not required for other packages later

ashmeenkaur (approved):

PR #3765, opened 2025-08-29
fix(do not review): make universe domain api retry able

Tulsishah:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+		// if the context is already cancelled.
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			return nil
+		}
+	}
+
+	jitteryBackoffDuration := time.Duration(1 + rand.Int63n(int64(nextDuration)))
```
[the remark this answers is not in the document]
Tulsishah: It is with go > 1.24 so not required

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,141 @@
...
+// This is similar to how gax-retries backoff after each failed retry.
+func (b *ExponentialBackoff) WaitWithJitter(ctx context.Context) error {
+	nextDuration := b.nextDuration()
+	jitteryBackoffDuration := time.Duration(1 + rand.Int63n(int64(nextDuration)))
+	select {
+	case <-time.After(jitteryBackoffDuration):
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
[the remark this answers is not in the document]
Tulsishah: I don't think it will ever be negative, but I've added a to-do to ensure the values are not initialized negatively.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+func (b *ExponentialBackoff) nextDuration() time.Duration {
+	next := b.next
+	b.next = min(b.config.Max, time.Duration(float64(b.next)*b.config.Multiplier))
+	return next
+}
+
+// WaitWithJitter waits for the next backoff duration with added jitter.
+// The jitter adds randomness to the backoff duration to prevent the thundering herd problem.
+// This is similar to how gax-retries backoff after each failed retry.
+func (b *ExponentialBackoff) WaitWithJitter(ctx context.Context) error {
```
gargnitingoogle: Does this need to be public ?


```suggestion
func (b *ExponentialBackoff) waitWithJitter(ctx context.Context) error {
```

internal/storage/storageutil/auth_client_option.go
```diff
@@ -45,9 +53,22 @@ func GetClientAuthOptionsAndToken(ctx context.Context, config *StorageClientConf
...
+		},
+	}
+	apiCall := func(attemptCtx context.Context) (string, error) {
+		return cred.UniverseDomain(attemptCtx)
+	}
+
+	domain, err := ExecuteWithRetry(ctx, retryConfig, "cred.UniverseDomain", "credentials", apiCall)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get UniverseDomain: %w", err)
+		return nil, nil, fmt.Errorf("failed to get UniverseDomain after retries: %w", err)
```
gargnitingoogle: No need to say `after retries`. The retry logs will themselves appear, if retry is made.

```suggestion
		return nil, nil, fmt.Errorf("failed to get UniverseDomain: %w", err)
```

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	Initial time.Duration
+	// Max duration for next backoff.
+	Max time.Duration
+	// The rate at which the backoff duration should grow
+	// over subsequent calls to next().
+	Multiplier float64
+}
+
+// ExponentialBackoff holds the duration parameters for exponential backoff.
+type ExponentialBackoff struct {
```
gargnitingoogle: Does this need to be public ?

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
+)
+
+// ExponentialBackoffConfig is config parameters
+// needed to create an ExponentialBackoff.
+type ExponentialBackoffConfig struct {
```
gargnitingoogle: Instead of making this public, we should take the preexisting `clientConfig *storageutil.StorageClientConfig` in the creation of RetryConfig i.e. don't change the structure of  RetryConfig, but provide a function e.g. NewRetryConfig, which takes `*storageutil.StorageClientConfig` instead of a `ExponentialBackoffConfig ` .

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	next := b.next
+	b.next = min(b.config.Max, time.Duration(float64(b.next)*b.config.Multiplier))
+	return next
+}
+
+// WaitWithJitter waits for the next backoff duration with added jitter.
+// The jitter adds randomness to the backoff duration to prevent the thundering herd problem.
+// This is similar to how gax-retries backoff after each failed retry.
+func (b *ExponentialBackoff) WaitWithJitter(ctx context.Context) error {
+	nextDuration := b.nextDuration()
```
gargnitingoogle, line 73: This is different from what was in the original waitWithJitter function. This one has no randomness in it.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	if nextDuration <= 0 {
```
gargnitingoogle: This duration can never be negative. Can it?

internal/storage/storageutil/retry.go
gargnitingoogle: My bad. This code is structured different from the original. Please don't change the structure, it can cause unnecessary issues.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	BackoffConfig ExponentialBackoffConfig
+}
+
+// ExecuteWithRetry encapsulates the retry logic for control client operations.
+// It performs time-bound, exponential backoff retries for a given API call.
+// It is expected that the given apiCall returns a structure, and not an HTTP response,
+// so that it does not leave behind any trace of a pending operation on server.
+func ExecuteWithRetry[T any](
+	ctx context.Context,
+	config RetryConfig,
```
gargnitingoogle: Take this as pointer, to keep it simple and more efficient.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	operationName string,
+	reqDescription string,
+	apiCall func(attemptCtx context.Context) (T, error),
+) (T, error) {
+	var zero T
+
```
gargnitingoogle, line 123: Unnecessary blank line.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	defer cancel()
+
+	// Create a new backoff controller specific to this api call.
+	backoff := NewExponentialBackoff(&config.BackoffConfig)
+	for {
+		attemptCtx, attemptCancel := context.WithTimeout(parentCtx, config.RetryDeadline)
+
+		logger.Tracef("Calling %s for %q with deadline=%v ...", operationName, reqDescription, config.RetryDeadline)
+		result, err := apiCall(attemptCtx)
+		// Cancel attemptCtx after it is no longer needed to free up its resources.
```
gargnitingoogle: nit

```suggestion
		// Cancel attemptCtx after it is no longer needed, to free up its resources.
```

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type ExponentialBackoffTest struct {
```
gargnitingoogle:
```suggestion
type ExponentialBackoffTestSuite struct {
```

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	start := time.Now()
+	err := b.WaitWithJitter(ctx)
+	elapsed := time.Since(start)
+
+	assert.ErrorIs(t.T(), err, context.Canceled)
+	// The function should return almost immediately.
+	assert.Less(t.T(), elapsed, initial, "waitWithJitter should return quickly when context is cancelled")
+}
+
```
gargnitingoogle, line 106: We need some UTs for RetryConfig also.

gargnitingoogle (changes requested):

gargnitingoogle:
@Tulsishah please also run the perf test with the presubmit label, before merge once for confirmation.

gargnitingoogle:
It is a bit weird that gemini has not provided summary/review despite being asked explicitly.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+
+	parentCtx, cancel := context.WithTimeout(ctx, config.TotalRetryBudget)
+	defer cancel()
+
+	// Create a new backoff controller specific to this api call.
+	backoff := NewExponentialBackoff(&config.BackoffConfig)
+	for {
+		attemptCtx, attemptCancel := context.WithTimeout(parentCtx, config.RetryDeadline)
+
+		logger.Tracef("Calling %s for %q with deadline=%v ...", operationName, reqDescription, config.RetryDeadline)
```
gargnitingoogle: I know it's different from the original code, but this is an improvement thought I had:


```suggestion
        if i == 0 {
		  logger.Tracef("Calling %s for %q with deadline=%v ...", operationName, reqDescription, config.RetryDeadline)
		} else {
		  logger.Tracef("Retrying %s for %q with deadline=%v ...", operationName, reqDescription, config.RetryDeadline)
		}
```

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,144 @@
...
+	TotalRetryBudget time.Duration
+	// Config for managing backoff durations in-between retry attempts.
+	BackoffConfig ExponentialBackoffConfig
+}
+
+// ExecuteWithRetry encapsulates the retry logic for control client operations.
+// It performs time-bound, exponential backoff retries for a given API call.
+// It is expected that the given apiCall returns a structure, and not an HTTP response,
+// so that it does not leave behind any trace of a pending operation on server.
+func ExecuteWithRetry[T any](
```
gargnitingoogle, line 111: Now that it's an independent utility, we should add independent unit tests for this, similar to the ones I added in control_client_wrapper_test.go e.g. define a function which takes a defined action, and also a customizable stall duration. Then call this function with ExecuteWithRetry with different stall durations and different types of failures (some tests with retryable errors, some with non-retryable) etc. to see that it does the retries etc. as expected (e.g. keep a count of how many times the function got called and what it returned eventually and after how long etc).

gargnitingoogle: Add another test here to confirm if the parent context has a timeout < retryBudget, then the retries should fail based on the parent context timeout and not retryBudget .

internal/storage/storageutil/retry.go
Tulsishah: I also thought about it but gemini made a review with high priority so I made a fix for safer side.
https://github.com/GoogleCloudPlatform/gcsfuse/pull/3765#discussion_r2313168761

internal/storage/storageutil/retry.go
Tulsishah: I think it is same only, just added gemini suggested review comment

internal/storage/storageutil/auth_client_option.go
```diff
@@ -17,13 +17,21 @@ package storageutil
...
 
 	"cloud.google.com/go/auth/oauth2adapt"
 	auth2 "github.com/googlecloudplatform/gcsfuse/v3/internal/auth"
 	"golang.org/x/oauth2"
 	"google.golang.org/api/option"
 )
 
+const (
+	// Default retry parameters for auth client calls.
+	defaultAuthClientRetryDeadline    = 30 * time.Second
```
vadlakondaswetha: isn't this config same as what is used for storageControlClient, if yes should we define a defaultREtryConfig and use it at both the places?

internal/storage/storageutil/retry.go
Tulsishah: As discussed offline added a to-do to ensure the values are not initialized negatively.

internal/storage/storageutil/retry.go
Tulsishah: As discussed offline added a to-do to ensure the values are not initialized negatively.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,162 @@
...
+	// Time-limit on every individual retry attempt.
+	RetryDeadline time.Duration
+	// Total duration allowed across all the attempts.
+	TotalRetryBudget time.Duration
+	// Config for managing backoff durations in-between retry attempts.
+	BackoffConfig exponentialBackoffConfig
+}
+
+// NewRetryConfig creates a new RetryConfig.
+func NewRetryConfig(clientConfig *StorageClientConfig, retryDeadline, totalRetryBudget, initialBackoff time.Duration) *RetryConfig {
```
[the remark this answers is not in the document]
Tulsishah: Will add in separate PR. For now we set only correct values.

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,162 @@
...
+func ExecuteWithRetry[T any](
+	ctx context.Context,
+	config *RetryConfig,
+	operationName string,
+	reqDescription string,
+	apiCall func(attemptCtx context.Context) (T, error),
+) (T, error) {
+	// If the context is already cancelled, return immediately.
+	if err := ctx.Err(); err != nil {
+		return *new(T), err
```
gargnitingoogle: Use zero here.

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+	start := time.Now()
+	err := b.waitWithJitter(ctx)
+	elapsed := time.Since(start)
+
+	assert.NoError(t.T(), err)
+	// The function should wait for a duration close to initial.
+	assert.LessOrEqual(t.T(), elapsed, initial*2, "waitWithJitter should not wait excessively long")
+}
+
+func (t *ExponentialBackoffTestSuite) TestNewRetryConfig() {
```
gargnitingoogle: let's split this off as RetryConfigTestSuite, or rename ExponentialBackoffTestSuite to RetryConfigTestSuite .

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+type ExecuteWithRetryTestSuite struct {
+	suite.Suite
+	retryConfig *RetryConfig
+}
+
+func (t *ExecuteWithRetryTestSuite) SetupTest() {
+	t.retryConfig = &RetryConfig{
+		RetryDeadline:    50 * time.Millisecond,
```
gargnitingoogle: let's reduce these durations to `x*100*time.Microsecond` .

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+	result, err := ExecuteWithRetry(context.Background(), t.retryConfig, "testOp", "testReq", apiCall)
+
+	// Assert
+	assert.Error(t.T(), err)
+	assert.Empty(t.T(), result)
+	assert.ErrorIs(t.T(), err, nonRetryableErr)
+	assert.Equal(t.T(), 1, callCount)
+}
+
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_Timeout() {
```
gargnitingoogle, line 249: Let's add another test where first attempt returns a retryable error and then the next attempt returns a non-retryable error.

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+		return "", ctx.Err()
+	}
+
+	// Act
+	result, err := ExecuteWithRetry(context.Background(), t.retryConfig, "testOp", "testReq", apiCall)
+
+	// Assert
+	assert.Error(t.T(), err)
+	assert.Empty(t.T(), result)
+	assert.ErrorIs(t.T(), err, context.DeadlineExceeded)
```
gargnitingoogle: I don't think you need the check for `assert.Error(t.T(), err)` . It's redundant if we have a check for assert.ErrorIs already.

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_TotalRetryBudgetExceeded() {
+	// Arrange
+	var callCount int
+	// Set a short total retry budget.
+	t.retryConfig.TotalRetryBudget = 30 * time.Millisecond
```
gargnitingoogle: lower duration.

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+		callCount++
+		return "", status.Error(codes.Unavailable, "server unavailable")
+	}
+
+	// Act
+	_, err := ExecuteWithRetry(context.Background(), t.retryConfig, "testOp", "testReq", apiCall)
+
+	// Assert
+	assert.Error(t.T(), err)
+	assert.ErrorIs(t.T(), err, context.DeadlineExceeded, "The error should be from the total retry budget timeout")
```
gargnitingoogle, line 287: Also add a check to confirm that the callCount > 1.

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_ParentContextTimeout() {
+	// Arrange
+	var callCount int
+	// Set a parent context timeout that is shorter than the total retry budget.
+	parentCtx, cancel := context.WithTimeout(context.Background(), t.retryConfig.TotalRetryBudget-100*time.Millisecond)
+	defer cancel()
+	apiCall := func(ctx context.Context) (string, error) {
+		callCount++
```
gargnitingoogle, line 301: add a stall here.

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// Act
+	_, err := ExecuteWithRetry(context.Background(), t.retryConfig, "testOp", "testReq", apiCall)
+
+	// Assert
+	assert.Error(t.T(), err)
+	assert.ErrorIs(t.T(), err, context.DeadlineExceeded, "The error should be from the total retry budget timeout")
+}
+
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_ParentContextTimeout() {
```
gargnitingoogle: We need two more variants of this function.

1. Parent context timeout < retryDeadline.
2. Parent context `timeout > retryDeadline`, but `timeout < totalRetryBudget`.
4. Parent context `timeout > totalRetryBudget`.

internal/storage/control_client_wrapper.go
```diff
@@ -96,125 +87,20 @@ func withBillingProject(controlClient StorageControlClient, billingProject strin
...
-
-	// Time-limit on every individual retry attempt.
-	retryDeadline time.Duration
-	// Total duration allowed across all the attempts.
-	totalRetryBudget time.Duration
-
-	// Config for managing backoff durations in-between retry attempts.
-	backoffConfig exponentialBackoffConfig
+	raw         StorageControlClient
+	retryConfig *storageutil.RetryConfig
```
gargnitingoogle, line 96: Switch this to object.


```suggestion
	retryConfig storageutil.RetryConfig
```

internal/storage/control_client_wrapper_test.go
```diff
@@ -151,7 +151,11 @@ func (t *AllApiRetryWrapperTest) SetupTest() {
 
 func (t *StorageLayoutRetryWrapperTest) TestGetStorageLayout_SuccessOnFirstAttempt() {
 	// Arrange
-	client := newRetryWrapper(t.stallingClient, 100*time.Microsecond, 1000*time.Microsecond, 1*time.Microsecond, 10*time.Microsecond, 2, false)
+	clientConfig := &storageutil.StorageClientConfig{
```
gargnitingoogle: Can you wrap both these lines in a helper function which takes the same set of arguments as newRetryWrapper to reduce the diff-size and to make the code shorter ?

Tulsishah: I felt what I did was more readable but as dicussed offline I made change for less diff

internal/storage/control_client_wrapper.go
Tulsishah: Generally, we should use a pointer to a struct *storageutil.RetryConfig for retryConfig within the storageControlClientWithRetry struct. This is because pointer initialization is faster and uses less memory compared to initializing a full struct.

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,370 @@
...
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_ParentContextTimeoutShorterThanRetryDeadline() {
+	// Arrange
+	var callCount int
+	stallDuration := t.retryConfig.RetryDeadline + 5*time.Microsecond
+	// Set a parent context timeout that is shorter than the total retry budget.
+	parentCtx, cancel := context.WithTimeout(context.Background(), t.retryConfig.RetryDeadline-50*time.Microsecond)
+	defer cancel()
+	apiCall := func(ctx context.Context) (string, error) {
+		callCount++
+		time.Sleep(stallDuration)
```
gargnitingoogle: time.Sleep won't work here. This function should return context-deadline error if ctx's remaining time-out < stallDuration, but that won't happen here, and it'll always return Unavailable error. This logic doesn't care about what ctx's remaining time-out value is, which means the retryDeadline is also ignored.

We should switch this to:


```go
select {
  case <-time.After(stallDuration):
  case <-ctx.Done():
    return "", ctx.Error()
}
```

Same goes for all the other tests which have a time.Sleep() in their `apiCall` definition .

internal/storage/control_client_wrapper_test.go
```diff
@@ -149,9 +149,18 @@ func (t *AllApiRetryWrapperTest) SetupTest() {
 	}
 }
 
+func (t *StorageLayoutRetryWrapperTest) newHelperRetryWrapper(controlClient StorageControlClient, retryDeadline, totalRetryBudget, initialBackoff, maxRetrySleep time.Duration, backoffMultiplier float64, retryFolderAPIs bool) StorageControlClient {
```
gargnitingoogle: This function is exactly the same in this class and in AllAPIStorageLayoutRetryWrapperTest. These can be merged into a single method in the parent class i.e. ControlClientRetryWrapperTest .

```go
func (t *ControlClientRetryWrapperTest) newHelperRetryWrapper(controlClient StorageControlClient, retryDeadline, totalRetryBudget, initialBackoff, maxRetrySleep time.Duration, backoffMultiplier float64, retryFolderAPIs bool) StorageControlClient {
	t.T().Helper()
	clientConfig := &storageutil.StorageClientConfig{
		MaxRetrySleep:   maxRetrySleep,
		RetryMultiplier: backoffMultiplier,
	}
	return newRetryWrapper(t.stallingClient, clientConfig, retryDeadline, totalRetryBudget, initialBackoff, retryFolderAPIs)
}
```

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_Timeout() {
+	// Arrange
+	stallDuration := t.retryConfig.RetryDeadline + 5*time.Microsecond
+	var callCount int
+	apiCall := func(ctx context.Context) (string, error) {
+		callCount++
+		// Simulate a call that always takes longer than the per-attempt deadline.
+		time.Sleep(stallDuration)
```
gargnitingoogle:
```suggestion
select {
  case <-time.After(stallDuration):
  case <-ctx.Done():
    return "", ctx.Error()
}
```

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,370 @@
...
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_ParentContextTimeoutBetweenDeadlines() {
+	// Arrange
+	var callCount int
+	stallDuration := t.retryConfig.RetryDeadline + 5*time.Microsecond
+	// Set a parent context timeout that is longer than one attempt but shorter than the total budget.
+	parentCtx, cancel := context.WithTimeout(context.Background(), t.retryConfig.RetryDeadline+50*time.Microsecond)
+	defer cancel()
+	apiCall := func(ctx context.Context) (string, error) {
+		callCount++
+		time.Sleep(stallDuration)
```
gargnitingoogle:
```suggestion
select {
  case <-time.After(stallDuration):
  case <-ctx.Done():
    return "", ctx.Error()
}
```

internal/storage/storageutil/retry_test.go
```diff
@@ -0,0 +1,370 @@
...
+	assert.Greater(t.T(), callCount, 0, "apiCall should have been called at least once")
+}
+
+func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_ParentContextTimeoutLongerThanBudget() {
+	// Arrange
+	stallDuration := t.retryConfig.RetryDeadline + 5*time.Microsecond
+	parentCtx, cancel := context.WithTimeout(context.Background(), t.retryConfig.TotalRetryBudget+100*time.Microsecond)
+	defer cancel()
+	apiCall := func(ctx context.Context) (string, error) {
+		time.Sleep(stallDuration)
```
gargnitingoogle:
```suggestion
select {
  case <-time.After(stallDuration):
  case <-ctx.Done():
    return "", ctx.Error()
}
```

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,162 @@
...
+		TotalRetryBudget: totalRetryBudget,
+		BackoffConfig: exponentialBackoffConfig{
+			initial:    initialBackoff,
+			max:        clientConfig.MaxRetrySleep,
+			multiplier: clientConfig.RetryMultiplier,
+		},
+	}
+}
+
+// ExecuteWithRetry encapsulates the retry logic for control client operations.
```
gargnitingoogle: nit:

```suggestion
// ExecuteWithRetry encapsulates the retry logic over a given operation.
```

internal/storage/storageutil/retry.go
```diff
@@ -0,0 +1,162 @@
...
+		result, err := apiCall(attemptCtx)
+		// Cancel attemptCtx after it is no longer needed, to free up its resources.
+		attemptCancel()
+
+		if err == nil {
+			return result, nil
+		}
+
+		// If the error is not retryable, return it immediately.
+		if !ShouldRetry(err) {
```
gargnitingoogle, line 147: Improvement for later: ShouldRetry should be a function value stored in the RetryConfig, rather than using the fixed ShouldRetry function. This will be needed for better control over this function. Please add a comment/task/buganizer for capturing this.

internal/storage/control_client_wrapper.go
gargnitingoogle: > we should use a pointer to a struct *storageutil.RetryConfig for retryConfig within the storageControlClientWithRetry struct

Why is this ? On the contrary, if the struct is a fixed-size struct without any slice or pointer inside, and if it is always needed, then it is better not being a pointer.

gargnitingoogle: > This is because pointer initialization is faster and uses less memory compared to initializing a full struct.

Can you please share the guidance on this ?

`pointer initialization is faster` only when it's initialized without any data AFAIK. e.g. `var s *Config` is faster and less memory than `var s Config`. But I don't know why `s := &Config{...}` would be faster and less memory consuming than `s := Config{...}` . In fact the pointer semantic will invoke GC unnecessarily for this simple structure and will hold the extra memory for the pointer variable.

gargnitingoogle: @Tulsishah ^

gargnitingoogle (changes requested):
One major comment to be addressed (https://github.com/GoogleCloudPlatform/gcsfuse/pull/3765#discussion_r2318044980).

internal/storage/storageutil/retry.go
Tulsishah: I think people can see this function and extend it as necessary when ever require.  Don't think so require to add too many TODOs until it's really required.

gargnitingoogle:
> One major comment to be addressed.

As discussed offline, @Tulsishah let's capture the issue in a buganizer ticket. We'll handle it separately.

gargnitingoogle (dismissed):
@Tulsishah @raj-prince as discussed offline, I am approving this PR with couple of TODOs for later.
1. Change retryConfig from pointer to value in storageControlClientWithRetry. This is a design inefficiency (https://github.com/GoogleCloudPlatform/gcsfuse/pull/3765#discussion_r2318044980).
2. Make `shouldRetry` configurable in `RetryConfig`. This is an improvement for usability.

internal/storage/storageutil/retry_test.go
```diff
@@ -158,8 +158,8 @@ type ExecuteWithRetryTestSuite struct {
 
 func (t *ExecuteWithRetryTestSuite) SetupTest() {
 	t.retryConfig = &RetryConfig{
-		RetryDeadline:    100 * time.Microsecond,
-		TotalRetryBudget: 500 * time.Microsecond,
+		RetryDeadline:    500 * time.Microsecond,
```
gargnitingoogle, line 161: What is the total runtime of ExecuteWithRetryTestSuite after all these changes ? If it's above 100ms, then I'd worry about impacting the total runtime of the unit test runs.

Tulsishah: It's 10ms only

internal/storage/storageutil/retry.go
Tulsishah: Existing blank line from original function (just moved it from private to util), removed anyways.

internal/storage/storageutil/retry_test.go
Tulsishah: I've already included:
1. A test for a retryable error.
2. A test for a non-retryable error.

I believe these two tests adequately cover the scenario, as the logic depends solely on the shouldRetry function. While I don't see the need for additional tests beyond these, I've added them to address the concerns you have.

internal/storage/storageutil/retry_test.go
Tulsishah: I think there is no harm in doing so(asserting error) but we can remove it.

gargnitingoogle (approved):

raj-prince (approved):
Please also create a bug at metadata layer: that the default retry value is not sufficient, either they should expose a flag to change the default retry value. Or consider increasing the metadata layer default retry value.

PR #3773, opened 2025-09-02
fix(config): Use BlockSizeMb value in form of MB instead of converting it to bytes during rationalization.

meet2mky:
### Description
Fixing streaming writes max-block-size-mb to be used as MB instead of bytes as it would confuse the developer/customer looking at config which shows flag name in MB but value in bytes.
Ex: 
7GhmDBVsuaBFrhE

Note: checked that flag is not used anywhere else apart from this.
b/442479563
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/fs/inode/file_test.go
```diff
@@ -1891,15 +1891,15 @@ func (t *FileTest) TestRegisterFileHandle() {
 func getWriteConfig() *cfg.WriteConfig {
 	return &cfg.WriteConfig{
 		MaxBlocksPerFile:      10,
-		BlockSizeMb:           10,
```
ashmeenkaur, line 1894: is this an unintended change?

ashmeenkaur: same in internal/fs/streaming_writes_local_file_test.go and internal/fs/streaming_writes_empty_gcs_object_test.go

meet2mky: Yes, Before that it would have created buffers of size 10 Bytes (Because we are directly injecting the values to config). Now it would created buffers of size 10MB which is a lot for composite tests so we changed this to 1MB to not use a lot of memory during composite tests.

PR #3783, opened 2025-09-08
chore(metrics): Ignore update and log warning if counter metrics receive a negative increment

kislaykishore:
b/443638827
### Testing details
1. Manual - Verified manually
2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -1273,6 +1273,10 @@ func (o *otelMetrics) BufferedReadDownloadBlockLatency(
 
 func (o *otelMetrics) BufferedReadFallbackTriggerCount(
 	inc int64, reason string) {
+	if inc < 0 {
+		logger.Warnf("Counter metric buffered_read/fallback_trigger_count received a negative increment: %d", inc)
+		return
+	}
 	switch reason {
 	case "insufficient_memory":
 		o.bufferedReadFallbackTriggerCountReasonInsufficientMemoryAtomic.Add(inc)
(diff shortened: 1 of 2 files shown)


metrics/otel_metrics.go
```diff
@@ -1273,6 +1273,10 @@ func (o *otelMetrics) BufferedReadDownloadBlockLatency(
 
 func (o *otelMetrics) BufferedReadFallbackTriggerCount(
 	inc int64, reason string) {
+	if inc < 0 {
```
charith87, line 1276: Should this be <= 0

metrics/otel_metrics.go
```diff
@@ -1314,6 +1322,10 @@ func (o *otelMetrics) BufferedReadScheduledBlockCount(
 
 func (o *otelMetrics) FileCacheReadBytesCount(
 	inc int64, readType string) {
+	if inc < 0 {
```
charith87, line 1325: Similar here as well. <= 0

metrics/otel_metrics.go
kislaykishore: Discussed offline that we'd like this check to be "< 0". This is to cater to cases where say we have function in which there is a variable `inc` tracking the GCS request bytes fetched which is initialized with 0. Now, at the end of the function, we invoke `GcsReadBytesCount(inc)` so that it's incremented with the amount of data read. It could however happen that within the method, it's determined that the value is already present in file-cache and the data is fetched from there. The `inc` will remain at 0. We don't want such cases to start throwing a warning.

charith87 (approved):

PR #3795, opened 2025-09-10
feat: Retry for short reads & MRD Read Handle Support

abhishek10004:
### Description
1. Adds read handle support with MRD.
2. Adds retry logic for short reads.

### Testing details
1. Manual - Done
3. Unit tests - Done
4. Integration tests - NA


internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -120,18 +122,41 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (reader
...
 	defer func() {
 		gr.updateExpectedOffset(offset + int64(readerResponse.Size))
 		gr.totalReadBytes.Add(uint64(readerResponse.Size))
 	}()
 
+	bytesRead, err := gr.read(ctx, readReq)
+	readerResponse.Size = bytesRead
+
+	// Retry reading in case of short read.
+	// This would be used for MRD only as range reader would have created a new reader before trying to read
```
vadlakondaswetha: if range reaqder is doing internally why are we bringing it out for mrd?

internal/gcsx/client_readers/range_reader.go
```diff
@@ -171,7 +180,7 @@ func (rr *RangeReader) readFromRangeReader(ctx context.Context, p []byte, offset
 		// if the reader peters out early. That's fine, but it means we should
 		// have hit the limit above.
 		if rr.reader != nil {
-			err = fmt.Errorf("range reader returned early by skipping %d bytes", rr.limit-rr.start)
+			err = fmt.Errorf("range reader returned early by skipping %d bytes: %w", rr.limit-rr.start, util.ErrShortRead)
```
vadlakondaswetha, line 183: same question. why do we need these changes if range_reader is already retyring the request

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -161,49 +163,60 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) cleanupMultiRangeDownloader() {
...
 			mrd, err = mrdWrapper.bucket.NewMultiRangeDownloader(context.Background(), &gcs.MultiRangeDownloaderRequest{
 				Name:           mrdWrapper.object.Name,
 				Generation:     mrdWrapper.object.Generation,
 				ReadCompressed: mrdWrapper.object.HasContentEncodingGzip(),
+				ReadHandle:     mrdWrapper.handle,
 			})
 			if err == nil {
 				// Updating mrdWrapper.Wrapped only when MRD creation was successful.
 				mrdWrapper.Wrapped = mrd
+				mrdWrapper.handle = mrd.GetHandle()
```
vadlakondaswetha: if you store the handle into a variable, how will it get refreshed?

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -120,18 +122,41 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (reader
...
+	// if the existing reader's limit did not cover the object size.
+	if (err == nil || errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, util.ErrShortRead)) &&
```
vadlakondaswetha: please split these conditions in new method.

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -120,18 +122,41 @@ func (gr *GCSReader) ReadAt(ctx context.Context, p []byte, offset int64) (reader
...
 		gr.updateExpectedOffset(offset + int64(readerResponse.Size))
 		gr.totalReadBytes.Add(uint64(readerResponse.Size))
 	}()
 
+	bytesRead, err := gr.read(ctx, readReq)
+	readerResponse.Size = bytesRead
+
+	// Retry reading in case of short read.
+	// This would be used for MRD only as range reader would have created a new reader before trying to read
+	// if the existing reader's limit did not cover the object size.
```
vadlakondaswetha: you can add a condition here itself to reduce the if conditions
if !gr.bucket.BucketType().Zonal {
return}

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: I should have rephrased it better. Will update the comment.

What range reader is doing is that it creates a new reader before the actual read if the read request is beyond what the current reader can support. So, in that sense, range reader would try with a range reader created till the end of the object size.
It will not retry in case of short reads which is rare but still possible even when the reader was created earlier with proper limit. Hence, this retry would mostly benefit MRD but will be useful for range reader as well.

internal/gcsx/multi_range_downloader_wrapper.go
abhishek10004: Does the handle get refreshed? Even go sdk is storing it in a variable at the time of gRPCBidiReader creation and then just returns the stored value as part of getHandle

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: will create a new method for this

internal/gcsx/client_readers/range_reader.go
abhishek10004: Answered as part of another comment.

internal/gcsx/client_readers/gcs_reader.go
abhishek10004: created a new method to handle it.

internal/gcsx/multi_range_downloader_wrapper.go
vadlakondaswetha: For MRD  it should get refreshed periodically. FWIK, they periodically send a new handle as part of the read responses.

internal/gcsx/client_readers/range_reader.go
```diff
@@ -118,6 +119,14 @@ func (rr *RangeReader) ReadAt(ctx context.Context, req *gcsx.GCSReaderRequest) (
 	}
 	var err error
 
+	if req.ForceCreateReader && rr.reader != nil {
```
vadlakondaswetha, line 122: unit test for this?

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -161,49 +163,60 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) cleanupMultiRangeDownloader() {
...
+func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader(forceRecreateMRD bool) (err error) {
 	if mrdWrapper.object == nil || mrdWrapper.bucket == nil {
 		return fmt.Errorf("ensureMultiRangeDownloader error: Missing minObject or bucket")
 	}
 
-	mrdWrapper.mu.RLock()
 	// Create the MRD if it does not exist.
 	// In case the existing MRD is unusable due to closed stream, recreate the MRD.
-	if mrdWrapper.Wrapped == nil || mrdWrapper.Wrapped.Error() != nil {
+	if forceRecreateMRD || mrdWrapper.Wrapped == nil || mrdWrapper.Wrapped.Error() != nil {
```
vadlakondaswetha, line 150: unit test to check a new MRD is created when forceRecreateMRD is true

internal/gcsx/multi_range_downloader_wrapper.go
abhishek10004: Thanks. I just checked the code and they are updating the read handle during read responses. I've now made changes to cache the handle only when we're closing the MRD, otherwise we'll fetch the latest handle from the MRD and use that.

internal/gcsx/multi_range_downloader_wrapper.go
abhishek10004: Done

internal/gcsx/client_readers/range_reader.go
abhishek10004: Done

internal/gcsx/client_readers/gcs_reader_test.go
```diff
@@ -863,6 +863,31 @@ func (t *gcsReaderTest) Test_ReadAt_ValidateZonalRandomReads() {
...
 
+func (t *gcsReaderTest) Test_ReadAt_MRDShortReadOnZonal() {
+	t.object.Size = 200
+	t.mockBucket.On("BucketType", mock.Anything).Return(gcs.BucketType{Zonal: true})
+	testContent := testUtil.GenerateRandomBytes(int(t.object.Size))
+	fakeMRDWrapper, err := gcsx.NewMultiRangeDownloaderWrapper(t.mockBucket, t.object, &cfg.Config{})
+	require.NoError(t.T(), err)
+	t.gcsReader.mrr.mrdWrapper = &fakeMRDWrapper
+	// First call to NewMultiRangeDownloader will return a short read.
+	t.mockBucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(fake.NewFakeMultiRangeDownloaderWithShortRead(t.object, testContent), nil).Once()
```
vadlakondaswetha, line 874: can we make this mock return a different data instead of testContent to ensure that it is indeed returning a short content

abhishek10004: We are asserting expectations which means NewMultiRangeDownloader has to be called twice otherwise test would fail. This guarantees that the first read would have been short.
This test is to check the retry logic works properly in case of short reads and we are able to read the whole content. multi_range_downloader_wrapper_test.go has Test_Read_ShortRead to test that actual data returned is less than asked for short read.

PR #3801, opened 2025-09-11
feat: Changes for Cache Include Regex Feature

ameoghosh:
### Description
Changes for Cache Include Regex Feature to support:
  * Add Include Feature
  * Remove Experimental on Exclude Feature.
  * Both Include and Exclude to work together as Customer Guidelines, etc.
 
### Testing details
Integration tests - Run manually all test matching TestCacheFileFor*

No

### DIFF
--- a/internal/cache/file/cache_handler.go
+++ b/internal/cache/file/cache_handler.go
@@ -82,7 +82,6 @@ func NewCacheHandler(fileInfoCache *lru.Cache, jobManager *downloader.JobManager
 
 func compileRegex(regexString string) *regexp.Regexp {
 	var compiledRegex *regexp.Regexp
-	compiledRegex = nil
 
 	if regexString != "" {
 		var err error
--- a/internal/cache/file/cache_handler_test.go
+++ b/internal/cache/file/cache_handler_test.go
@@ -87,12 +87,6 @@ func initializeCacheHandlerTestArgs(t *testing.T, fileCacheConfig *cfg.FileCache
 	jobManager := downloader.NewJobManager(cache, util.DefaultFilePerm,
 		util.DefaultDirPerm, cacheDir, DefaultSequentialReadSizeMb, fileCacheConfig, metrics.NewNoopMetrics())
 
-	// Only added for code coverage (not used in real tests) Mocked cached handler object (invalid include regex).
-	NewCacheHandler(cache, jobManager, cacheDir, util.DefaultFilePerm, util.DefaultDirPerm, fileCacheConfig.ExcludeRegex, fileCacheConfig.IncludeRegex+"[")
-
-	// Only added for code coverage (not used in real tests) Mocked cached handler object (invalid exclude regex).
-	NewCacheHandler(cache, jobManager, cacheDir, util.DefaultFilePerm, util.DefaultDirPerm, fileCacheConfig.ExcludeRegex+"[", fileCacheConfig.IncludeRegex)
-
 	// Mocked cached handler object.
 	cacheHandler := NewCacheHandler(cache, jobManager, cacheDir, util.DefaultFilePerm, util.DefaultDirPerm, fileCacheConfig.ExcludeRegex, fileCacheConfig.IncludeRegex)
 
@@ -598,6 +592,20 @@ func Test_GetCacheHandle_IncludeAndExclude(t *testing.T) {
 	// Check cache handle is not created for excluded file even if it matches include pattern.
 	chTestArgs.object.Name = "some_file_internal.txt"
 	cacheHandle, err = chTestArgs.cacheHandler.GetCacheHandle(chTestArgs.object, chTestArgs.bucket, false, 0)
+	assert.True(t, errors.Is(err, util.ErrFileExcludedFromCacheByRegex))
+	assert.Nil(t, cacheHandle)
+}
+
+func Test_GetCacheHandle_SameIncludeAndExcludeRegex(t *testing.T) {
+	regex := ".*\\.txt"
+	cacheDir := path.Join(os.Getenv("HOME"), "CacheHandlerTest/dir")
+	chTestArgs := initializeCacheHandlerTestArgs(t, &cfg.FileCacheConfig{EnableCrc: true, IncludeRegex: regex, ExcludeRegex: regex}, cacheDir)
+
+	// Check cache handle is not created for a file that matches both include and
+	// exclude regex, as exclude takes precedence.
+	chTestArgs.object.Name = "some_file.txt"
+	cacheHandle, err := chTestArgs.cacheHandler.GetCacheHandle(chTestArgs.object, chTestArgs.bucket, false, 0)
+
 	assert.True(t, errors.Is(err, util.ErrFileExcludedFromCacheByRegex))
 	assert.Nil(t, cacheHandle)
 }


coverage.out
```diff
@@ -0,0 +1,140 @@
+mode: set
```
Tulsishah: I think it is added mistakenly. Please remove it.

internal/cache/file/cache_handler.go
```diff
@@ -297,13 +312,23 @@ func (chr *CacheHandler) Destroy() (err error) {
...
+	if chr.includeRegex == nil && chr.excludeRegex == nil {
 		return false
 	}
 
-	// Get the GCS name of the object and create the cloud path in the format bucket/object.
-	cloudPath := path.Join(bucket.Name(), bucket.GCSName(object))
-	return chr.excludeRegex.MatchString(cloudPath)
+	objectName := path.Join(bucket.Name(), bucket.GCSName(object))
+
+	// Exclude if it matches the exclude pattern.
```
Tulsishah, line 322: Should the flag description mention that the exclude flag takes precedence if both flags are set with the same data?

internal/cache/file/cache_handler_test.go
```diff
@@ -87,8 +87,14 @@ func initializeCacheHandlerTestArgs(t *testing.T, fileCacheConfig *cfg.FileCache
 	jobManager := downloader.NewJobManager(cache, util.DefaultFilePerm,
 		util.DefaultDirPerm, cacheDir, DefaultSequentialReadSizeMb, fileCacheConfig, metrics.NewNoopMetrics())
 
+	// Mocked cached handler object (invalid include regex).
+	NewCacheHandler(cache, jobManager, cacheDir, util.DefaultFilePerm, util.DefaultDirPerm, fileCacheConfig.ExcludeRegex, fileCacheConfig.IncludeRegex+"[")
```
Tulsishah: We are avoiding mocks and prefering fakes.

Tulsishah: Also we are not catching the response for these calls.

internal/cache/file/cache_handler_test.go
```diff
@@ -559,6 +565,43 @@ func Test_GetCacheHandle_ExcludeFromCache(t *testing.T) {
...
+	chTestArgs := initializeCacheHandlerTestArgs(t, &cfg.FileCacheConfig{EnableCrc: true, IncludeRegex: regex}, cacheDir)
+
+	// Check cache handle is created for included file.
+	chTestArgs.object.Name = "object_1"
+	cacheHandle, err := chTestArgs.cacheHandler.GetCacheHandle(chTestArgs.object, chTestArgs.bucket, false, 0)
+	assert.NoError(t, err)
+	assert.Nil(t, cacheHandle.validateCacheHandle())
+
+	// Check cache handle is not created for file not included.
+	chTestArgs.object.Name = "object_2"
```
Tulsishah, line 574: Why are we validating object_2? we should validate object_1 is included right?

internal/fs/read_cache_test.go
```diff
@@ -69,10 +69,10 @@ func (t *FileCacheTest) SetUpTestSuite() {
...
 	t.serverCfg.NewConfig = &cfg.Config{
 		FileCache: cfg.FileCacheConfig{
-			MaxSizeMb:                FileCacheSizeInMb,
-			CacheFileForRangeRead:    false,
-			EnableCrc:                true,
-			ExperimentalExcludeRegex: CacheExcludeName,
+			MaxSizeMb:             FileCacheSizeInMb,
+			CacheFileForRangeRead: false,
+			EnableCrc:             true,
+			ExcludeRegex:          CacheExcludeName,
```
Tulsishah, line 75: Should we include IncludeRegex parameter also?

tools/integration_tests/read_cache/cache_file_for_include_regex_test.go
```diff
@@ -0,0 +1,151 @@
...
+	expectedOutcome1 := readChunkAndValidateObjectContentsFromGCS(s.ctx, s.storageClient, testFileName, zeroOffset, t)
+	expectedOutcome2 := readChunkAndValidateObjectContentsFromGCS(s.ctx, s.storageClient, testFileName, offset1000, t)
+
+	structuredReadLogs := read_logs.GetStructuredLogsSortedByTimestamp(setup.LogFile(), t)
+	validate(expectedOutcome1, structuredReadLogs[0], true, false, 1, t)
+	validate(expectedOutcome2, structuredReadLogs[1], false, true, 1, t)
+	validateFileIsCached(testFileName, t)
+}
+
+func (s *cacheFileForIncludeRegexTest) testReadsForNonIncludedFile(t *testing.T) {
```
Tulsishah: nit: We generally write Test in captital.

tools/integration_tests/read_cache/cache_file_for_include_regex_test.go
```diff
@@ -0,0 +1,151 @@
...
+	testFileName := setupFileInTestDir(s.ctx, s.storageClient, fileSizeForRangeRead, t)
+
+	// Read the file and validate that it is not cached.
+	expectedOutcome1 := readChunkAndValidateObjectContentsFromGCS(s.ctx, s.storageClient, testFileName, zeroOffset, t)
+	expectedOutcome2 := readChunkAndValidateObjectContentsFromGCS(s.ctx, s.storageClient, testFileName, offset1000, t)
+
+	structuredReadLogs := read_logs.GetStructuredLogsSortedByTimestamp(setup.LogFile(), t)
+	validate(expectedOutcome1, structuredReadLogs[0], true, false, 1, t)
+	validate(expectedOutcome2, structuredReadLogs[1], false, true, 1, t)
```
Tulsishah: How can we get cache hit true for non included files?

coverage.out
ameoghosh: Yes removed.

internal/cache/file/cache_handler.go
ameoghosh: Noted

tools/integration_tests/read_cache/cache_file_for_include_regex_test.go
ameoghosh: > nit: We generally write Test in captital.

Purposefully it has to be written in small, otherwise test method will get called twice with incorrect flow of config, that was causing issue earlier during test case failure.

Init Cap method makes it Public where as we need private test method for this.

internal/cache/file/cache_handler_test.go
ameoghosh: This call is added only for code coverage and New method does not return error to be checked as it continues with next command.

internal/cache/file/cache_handler_test.go
ameoghosh: Second check is for exclude regex not provided and includeRegex does not match in this case object_2 should not have cacheHandle.

tools/integration_tests/read_cache/cache_file_for_include_regex_test.go
ameoghosh: Updated

tools/integration_tests/read_cache/cache_file_for_include_regex_test.go
ameoghosh: Fixed

internal/fs/read_cache_test.go
ameoghosh: Not needed for cache functionality, either Include / exclude is sufficient. Explciit test for Include regex is added in seperate test for +ve / -ve scenario.

internal/cache/file/cache_handler.go
```diff
@@ -56,16 +56,30 @@ type CacheHandler struct {
...
-			compiledRegex = nil
+			compiledExcludeRegex = nil
+		}
+	}
+
+	if includeRegex != "" {
+		var err error
+		compiledIncludeRegex, err = regexp.Compile(includeRegex)
+		if err != nil {
+			logger.Warnf("Failed to compile include regex %q: %v", includeRegex, err)
```
vadlakondaswetha: we should exact this logic into a seperate method and use directly in line 93 & 94.

internal/cache/file/cache_handler_test.go
vadlakondaswetha: I would recommend not adding code like this just for coverage. We are not blocking the PR for coverage right?

internal/cache/file/cache_handler_test.go
```diff
@@ -559,6 +565,43 @@ func Test_GetCacheHandle_ExcludeFromCache(t *testing.T) {
...
+	cacheHandle, err = chTestArgs.cacheHandler.GetCacheHandle(chTestArgs.object, chTestArgs.bucket, false, 0)
+	assert.True(t, errors.Is(err, util.ErrFileExcludedFromCacheByRegex))
+	assert.Nil(t, cacheHandle)
+}
+
+func Test_GetCacheHandle_IncludeAndExclude(t *testing.T) {
```
vadlakondaswetha, line 580: another test case when include and exclude pattern are exactly same

internal/fs/fs.go
```diff
@@ -2289,11 +2289,6 @@ func (fs *fileSystem) renameFile(ctx context.Context, op *fuseops.RenameOp, chil
 
 	switch c := child.(type) {
 	case *inode.FileInode:
-		// TODO(b/402335988): Fix rename flow for local files when streaming writes is disabled.
```
vadlakondaswetha: you might have to rebase correctly for these changes to not show up

tools/integration_tests/read_cache/cache_file_for_include_regex_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	//cacheEnable=true - Test included file is cached
+	//cacheEnable=false - Test non-included file is not cached
+	tests := []struct {
+		name        string
+		flags       gcsfuseTestFlags
+		cacheEnable bool
+	}{
+		{
+			name: "Test included file is cached",
```
vadlakondaswetha: We use testcases when we want to run the same test for multiple scenarios. in your case you have 2 tests and each one is run only for one testcase.

either combine those testcases and take the output as test case parameter or
do the test setup and do suite.run which should run both tests. no need of testcases.

vadlakondaswetha (commented):
We follow the practice of not putting multiple changes in single PR to reduce review time. In this case exclude-Regex name change could be in seperate PR.

internal/cache/file/cache_handler_test.go
ameoghosh: What is the recommended way of improving coverage for such constructor or its not required ?

cfg/config.go
```diff
@@ -531,9 +533,9 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.BoolP("file-cache-enable-parallel-downloads", "", false, "Enable parallel downloads.")
 
-	flagSet.StringP("file-cache-experimental-exclude-regex", "", "", "Exclude file paths (in the format bucket_name/object_key) specified by this regex from file caching.")
+	flagSet.StringP("file-cache-exclude-regex", "", "", "Exclude file paths (in the format bucket_name/object_key) specified by this regex from file caching.")
 
-	if err := flagSet.MarkHidden("file-cache-experimental-exclude-regex"); err != nil {
+	if err := flagSet.MarkHidden("file-cache-exclude-regex"); err != nil {
```
vadlakondaswetha, line 538: is this intentional that we want to hide these flags?

internal/cache/file/cache_handler.go
```diff
@@ -76,10 +75,25 @@ func NewCacheHandler(fileInfoCache *lru.Cache, jobManager *downloader.JobManager
...
 		mu:            locker.New("FileCacheHandler", func() {}),
-		excludeRegex:  compiledRegex,
+		excludeRegex:  compiledExcludeRegex,
+		includeRegex:  compiledIncludeRegex,
 	}
 }
 
+func compileRegex(regexString string) *regexp.Regexp {
+	var compiledRegex *regexp.Regexp
+	compiledRegex = nil
```
vadlakondaswetha: default value of pointer is nil, we dont need this line

internal/cache/file/cache_handler_test.go
vadlakondaswetha: its not required. can you remove those lines.

internal/cache/file/cache_handler_test.go
vadlakondaswetha: sorry, i didnt find this test?

cfg/config.go
ameoghosh: @charith87 can you confirm, if MarkHidden is what the intention here and only "experimental" was removed?

internal/cache/file/cache_handler_test.go
ameoghosh: > its not required. can you remove those lines.

Okay

internal/cache/file/cache_handler_test.go
ameoghosh: Added one additional unit test for same file matching exclude and include both.

cfg/config.go
charith87: It was already existing with this particular field. That being said, We'll un-hide the feature in a subsequent PR after this is submitted like typical features with switches before release. The same is the case with the include flag as well.

PR #3813, opened 2025-09-17
(do not review) never finalize objects with updated time delays

anushka567:
**Please ensure your PR title follows the format:**
```
type(scope): subject
```
Example:
`feat(api): add user login endpoint`

Available types:
- `feat`: A new feature
- `fix`: A bug fix
- `docs`: Documentation only changes
- `style`: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
- `refactor`: A code change that neither fixes a bug nor adds a feature
- `perf`: A code change that improves performance
- `test`: Adding missing tests or correcting existing tests
- `build`: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
- `ci`: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
- `chore`: Other changes that don't modify src or test files
- `revert`: Reverts a previous commit

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -197,8 +197,12 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 	wc.ProgressFunc = req.CallBack
 	// All objects in zonal buckets must be appendable.
 	wc.Append = bh.BucketType().Zonal
-	// FinalizeOnClose should be true for all writes for now.
-	wc.FinalizeOnClose = true
+	// Objects in zonal buckets should not finalized.
+	if bh.BucketType().Zonal {
+		wc.FinalizeOnClose = false
+	} else {
+		wc.FinalizeOnClose = true
+	}
 
 	// Copy the contents to the writer.
 	if _, err = io.Copy(wc, req.Contents); err != nil {
@@ -229,8 +233,12 @@ func (bh *bucketHandle) CreateObjectChunkWriter(ctx context.Context, req *gcs.Cr
 	wc.ProgressFunc = callBack
 	// All objects in zonal buckets must be appendable.
 	wc.Append = bh.BucketType().Zonal
-	// FinalizeOnClose should be true for all writes for now.
-	wc.FinalizeOnClose = true
+	// Objects in zonal buckets should not finalized.
+	if bh.BucketType().Zonal {
+		wc.FinalizeOnClose = false
+	} else {
+		wc.FinalizeOnClose = true
+	}
 
 	return wc, nil
 }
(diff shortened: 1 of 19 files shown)


internal/storage/bucket_handle.go
```diff
@@ -197,8 +197,12 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
...
 	// All objects in zonal buckets must be appendable.
 	wc.Append = bh.BucketType().Zonal
-	// FinalizeOnClose should be true for all writes for now.
-	wc.FinalizeOnClose = true
+	// Objects in zonal buckets should not finalized.
+	if bh.BucketType().Zonal {
+		wc.FinalizeOnClose = false
+	} else {
+		wc.FinalizeOnClose = true
+	}
```
vadlakondaswetha, line 205: why do we need a else check? this parameter is not valid for regional buckets right?

tools/integration_tests/util/operations/string_operations.go
```diff
@@ -55,3 +56,11 @@ func SplitBucketNameAndDirPath(t *testing.T, bucketNameWithDirPath string) (buck
 	}
 	return
 }
+
+// WaitForSizeUpdate waits for a specified time duration to ensure that stat()
+// call returns correct size for unfinalized object.
+func WaitForSizeUpdate(isZonal bool, duration time.Duration) {
```
vadlakondaswetha: shouldn't this be in file_operatons.go instead of string_operations

tools/integration_tests/write_large_files/concurrent_write_to_same_file_test.go
```diff
@@ -28,6 +28,9 @@ import (
 )
 
 func TestWriteToSameFileConcurrently(t *testing.T) {
+	if setup.IsZonalBucketRun() {
+		t.Skip("Skipping tests related to concurrent mutation of same object for ZB")
```
vadlakondaswetha, line 32: why are we skipping?

anushka567: This is just temporarily skipped.

tools/integration_tests/util/operations/string_operations.go
anushka567: Moving to operations.go since this isn't a file operation.

internal/storage/bucket_handle.go
anushka567: Not needed. If writer.Append = false (as in regional) then this field is ignored. WIll remove this .

vadlakondaswetha (approved):

PR #3818, opened 2025-09-17
feat (logging): Enhance GCSFuse logging with custom attribute in logs and printing CLI, CONFIG-FILE and complete GCSFuse Config

meet2mky:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/logger/logger.go
```diff
@@ -100,6 +107,25 @@ func InitLogFile(newLogConfig cfg.LoggingConfig) error {
...
+			log.Printf("Could not retrieve %s env variable. Using default: %v", GCSFuseMountInstanceIdEnvKey, mountDefaultInstanceId)
+		}
+	} else {
+		uuid, err := uuid.NewRandom()
+		if err == nil && uuid.String() != "" {
+			MountInstanceId = uuid.String()[:8]
+		} else {
+			MountInstanceId = mountDefaultInstanceId
+			log.Printf("Could not generate random UUID for logger, err %v. using default: %v", err, mountDefaultInstanceId)
+		}
```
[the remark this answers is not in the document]
meet2mky: Done!

internal/logger/logger.go
```diff
@@ -32,14 +34,19 @@ import (
...
+	GCSFuseMountInstanceIdEnvKey string = "GCSFUSE_MOUNT_INSTANCE_ID"
+	textFormat                   string = "text"
+	mountLoggerIdKey             string = "mount_id"
+	defaultMountInstanceId       string = "00000000"
 )
 
 var (
 	defaultLoggerFactory *loggerFactory
 	defaultLogger        *slog.Logger
+	MountInstanceId      string
```
vadlakondaswetha: we generally dont shares variables like this. what is the guarantee that the value is set before accessed. is there any other way to do this?

meet2mky: The value initialization part of `MountInstanceId` is driven by `init`: https://go.dev/doc/effective_go#init. In the same way we are initializing the `defaultLoggerFactory` or `defaultLogger`.

So the `MountInstanceId` is never changed once initialized and the value will always persist. There are two cases where this value might be set to defaultMountInstanceId(00000000). There aren't any cases for unset variable in this.

A) `uuid.NewRandom()` fails to return UUID. This is highly unlikely as `github.com/google/uuid` is the canonical way to generate UUIDs in go.
B) GCSFuse started the daemon process but was not able to pass the MountInstanceId environment variable. Since there are other variables which are passed in the same way, I do not expect the variable to be not present when it was passed to daemon process by parent process (If in any case the variable is not present, as a failsafe we use the default value of `defaultMountInstanceId(00000000)`).

Both of these scenarios seems highly unlikely to me.

Another way to do this could be that we generate the UUID during Mount operation and pass it to daemon in command line flag. This might increase more changes as we would need to add a hidden flag `--mount-instance-id` because this information needs to be passed to daemon process so that initial process and daemon process both can have same 
`MountInstanceId`

gargnitingoogle: If the issue is that we're sharing a non-constant integer to outside the package, and its value is indeed needed from another package, then we can make this variable unexposed, i.e. `mountInstanceID`, and provide a getter to it which checks if it's properly assigned, or not, and accordingly acts/panics. E.g.

```go
var (
  mountInstanceID string
)
...
func GetMountInstanceID() string, error {
    if mountInstanceID == "" {
       // create mount instance id
       // return error if failed to create
    }
    return mountInstanceID, nil
}
```

cmd/config_validation_test.go
```diff
@@ -30,7 +30,7 @@ import (
 func getConfigObject(t *testing.T, args []string) (*cfg.Config, error) {
 	t.Helper()
 	var c *cfg.Config
-	cmd, err := newRootCmd(func(config *cfg.Config, _, _ string) error {
+	cmd, err := newRootCmd(func(config *cfg.Config, _ map[string]interface{}, _, _ string) error {
```
gargnitingoogle: Use `any` instead of `interface{}`, that's shorter and much better for readability. go/go-style/decisions#use-any

cmd/legacy_main.go
```diff
@@ -52,6 +52,7 @@ import (
 const (
 	SuccessfulMountMessage         = "File system has been successfully mounted."
 	UnsuccessfulMountMessagePrefix = "Error while mounting gcsfuse"
+	DynamicMountFsName             = "gcsfuse"
```
gargnitingoogle: go/go-style/decisions#initialisms

```suggestion
	DynamicMountFSName             = "gcsfuse"
```

internal/logger/logger.go
gargnitingoogle: go/go-style/decisions#initialisms

```suggestion
	MountInstanceID      string
```

internal/logger/logger.go
```diff
@@ -32,14 +34,19 @@ import (
 // used to filter the logs from syslog and write it to respective log files -
 // gcsfuse.log in case of GCSFuse.
 const (
-	ProgrammeName           string = "gcsfuse"
-	GCSFuseInBackgroundMode string = "GCSFUSE_IN_BACKGROUND_MODE"
-	textFormat              string = "text"
+	ProgrammeName                string = "gcsfuse"
```
gargnitingoogle: Don't need to put type name in constant declarations in golang unless it's non-trivial. go/go-style/decisions#constant-names .

```suggestion
	ProgrammeName                = "gcsfuse"
```

internal/logger/logger.go
```diff
@@ -97,8 +104,37 @@ func InitLogFile(newLogConfig cfg.LoggingConfig) error {
...
+		// If GCSFuse is in background mode then look for the MountInstanceId in env which was set by the caller of demonize run.
+		if parentMountInstanceId, ok := os.LookupEnv(GCSFuseMountInstanceIdEnvKey); ok && parentMountInstanceId != "" {
+			MountInstanceId = parentMountInstanceId
+		} else {
+			MountInstanceId = defaultMountInstanceId
+			log.Printf("Could not retrieve %s env variable. Using default: %v", GCSFuseMountInstanceIdEnvKey, defaultMountInstanceId)
+		}
+		return
+	}
+	// If GCSFuse is not running in the background mode then generate a random UUID and store the trimmed UUID as MountInstanceId.
```
gargnitingoogle: Can we separate out the logic for the uuid generation into a separate function? It'll be useful for testing it independently.

internal/logger/logger.go
```diff
@@ -97,8 +104,37 @@ func InitLogFile(newLogConfig cfg.LoggingConfig) error {
...
-// init initializes the logger factory to use stdout and stderr.
+// init runs only once at package initialization.
 func init() {
+	setupMountInstanceID()
+	initializeDefaultLogger()
+}
+
+// setupMountInstanceID handles the retrieval of MountInstanceId if GCSFuse is in
+// background mode or generates one if running in foreground mode.
+func setupMountInstanceID() {
```
gargnitingoogle, line 134: Instead of setting the MountInstanceID inside the function, it might be better to return it as a string from this function. That'll be better for testing and readability at the caller end.

internal/logger/logger.go
```diff
@@ -109,12 +145,11 @@ func init() {
...
-// SetLogFormat updates the log format of default logger.
-func SetLogFormat(format string) {
-	if format == defaultLoggerFactory.format {
-		return
-	}
+// SetLogFormat updates the log format of default logger to given format
+// and initializes the default logger with fsName and mountInstanceId
+func SetLogFormatAndFsName(format, fsName string) {
 	defaultLoggerFactory.format = format
+	MountFsName = fsName
```
gargnitingoogle: Why is this tied to the function SetLogFormatAndFsName ? Can't these be two different functions SetLogFormat and setFSName ?

internal/logger/logger_test.go
```diff
@@ -27,17 +29,17 @@ import (
...
-	textWarningString = "^time=\"[a-zA-Z0-9/:. ]{26}\" severity=WARNING message=\"TestLogs: www.warningExample.com\""
-	textErrorString   = "^time=\"[a-zA-Z0-9/:. ]{26}\" severity=ERROR message=\"TestLogs: www.errorExample.com\""
-
-	jsonTraceString   = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"TRACE\",\"message\":\"TestLogs: www.traceExample.com\"}"
-	jsonDebugString   = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"DEBUG\",\"message\":\"TestLogs: www.debugExample.com\"}"
-	jsonInfoString    = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"INFO\",\"message\":\"TestLogs: www.infoExample.com\"}"
-	jsonWarningString = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"WARNING\",\"message\":\"TestLogs: www.warningExample.com\"}"
-	jsonErrorString   = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"ERROR\",\"message\":\"TestLogs: www.errorExample.com\"}"
+	textTraceString   = "^time=\"[a-zA-Z0-9/:. ]{26}\" severity=TRACE message=\"TestLogs: www.traceExample.com\" mount_id=[a-zA-Z0-9-]+\\s*$"
```
gargnitingoogle: Pull out the fixed pattern `[a-zA-Z0-9-]+\\s*` as a constant or at least as a var ? That'll be good for readability and less risky for typos.

internal/logger/logger_test.go
```diff
@@ -54,9 +56,10 @@ func TestLoggerSuite(t *testing.T) {
 
 func redirectLogsToGivenBuffer(buf *bytes.Buffer, level string) {
 	var programLevel = new(slog.LevelVar)
-	defaultLogger = slog.New(
-		defaultLoggerFactory.createJsonOrTextHandler(buf, programLevel, "TestLogs: "),
-	)
+	handler := defaultLoggerFactory.createJsonOrTextHandler(buf, programLevel, "TestLogs: ")
```
gargnitingoogle, line 61: Can this call fail ? If yes, how are we handling it ?

cmd/legacy_main.go
```diff
@@ -263,6 +264,13 @@ func isDynamicMount(bucketName string) bool {
 	return bucketName == "" || bucketName == "_"
 }
 
+func fsName(bucketName string) string {
```
gargnitingoogle, line 267: Please add unit tests for this.

cmd/legacy_main.go
```diff
@@ -324,15 +332,19 @@ func forwardedEnvVars() []string {
 	// process and daemon process. If this environment variable set that means
 	// programme is running as daemon process.
 	env = append(env, fmt.Sprintf("%s=true", logger.GCSFuseInBackgroundMode))
+
+	// This environment variable is used to enhance gcsfuse logging by using unique
+	// mount instance Id to identify logs from different mounts.
+	env = append(env, fmt.Sprintf("%s=%s", logger.GCSFuseMountInstanceIdEnvKey, logger.MountInstanceId))
```
gargnitingoogle: Please add unit tests for this change at https://github.com/GoogleCloudPlatform/gcsfuse/blob/444a43f069e22a4c4bcc5ee17346b7c60f9a0921/cmd/legacy_main_test.go#L314.

cmd/root.go
```diff
@@ -23,21 +23,53 @@ import (
...
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
 
-type mountFn func(c *cfg.Config, bucketName, mountPoint string) error
+// gcsfuseConfigs provides the configuration values provided by the user in CLI flags,
+// config file and full gcsfuse config in a wrapper config.
+func gcsfuseConfigs(v *viper.Viper, cmd *cobra.Command, finalConfig cfg.Config) map[string]interface{} {
+	wrapperConfig := make(map[string]interface{})
```
gargnitingoogle: nit: Is it possible to create an alias for `map[string]interface{}` somewhere? This expression is repeating at too many places.

```go
type FlagToAnyMap = map[string]any
```

cmd/root.go
```diff
@@ -23,21 +23,53 @@ import (
...
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
 
-type mountFn func(c *cfg.Config, bucketName, mountPoint string) error
+// gcsfuseConfigs provides the configuration values provided by the user in CLI flags,
+// config file and full gcsfuse config in a wrapper config.
+func gcsfuseConfigs(v *viper.Viper, cmd *cobra.Command, finalConfig cfg.Config) map[string]interface{} {
```
gargnitingoogle: Please provide some unit tests for this function, if possible.

internal/logger/logger.go
gargnitingoogle: nit: shorter name. Means the same thing.

```suggestion
	ProgramName                = "gcsfuse"
```

internal/logger/logger_test.go
gargnitingoogle: Never mind. It'll reduce readability and will not be const anymore. Can we create and call a function to set all these constants rather than these long unreadable strings which are easy to get wrong?

internal/logger/logger_test.go
```diff
@@ -337,5 +343,74 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+			defaultMountInstanceId,
+			"gcsfuse",
+			"text",
+			textInfoString,
+		},
+	}
+
+	for _, test := range testData {
+		t.T().Run(test.name, func(t *testing.T) {
+			os.Setenv(GCSFuseInBackgroundMode, "true")
```
gargnitingoogle: This is common to all tests. Shouldn't this be moved out of the for loop ?

internal/logger/logger_test.go
```diff
@@ -337,5 +343,74 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+			assert.NotNil(t, defaultLogger)
+
+			// Create a logger using defaultLoggerFactory that writes to buffer.
+			var buf bytes.Buffer
+			redirectLogsToGivenBuffer(&buf, defaultLoggerFactory.level)
+			Info("www.infoExample.com")
+			output := buf.String()
+
+			// Compare expected and actual log.
+			expectedRegexp := regexp.MustCompile(test.expectedOutput)
```
gargnitingoogle: It'll be more runtime-efficient to create `jsonInfoStringRegex := regexp.MustCompile(jsonInfoString)` and same for textInfoString, and use these in the struct, e.g.

```go
testData := []struct {
        ...
		expectedOutputRegex     *regexp.RegexP
	}{{
    ...
    expectedOutputRegex: jsonInfoStringRegex
    }, ...
```

If you find that too complicated, you can even directly use assert.Regexp for redability

```go
assert.Regexp(t, test.expectedOutput, output)
```

internal/logger/logger_test.go
```diff
@@ -337,5 +343,74 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+				err := os.Setenv(GCSFuseMountInstanceIdEnvKey, test.expectedInstanceId)
+				if err != nil {
+					t.Fatalf("Unable to set env, err: %v", err)
+				}
+			} else {
+				err := os.Unsetenv(GCSFuseMountInstanceIdEnvKey)
+				if err != nil {
+					t.Fatalf("Unable to unset env, err: %v", err)
+				}
+			}
```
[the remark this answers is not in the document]
gargnitingoogle: +1

internal/logger/logger.go
```diff
@@ -173,7 +208,8 @@ type loggerFactory struct {
 func (f *loggerFactory) newLogger(level string) *slog.Logger {
 	// create a new logger
 	var programLevel = new(slog.LevelVar)
-	logger := slog.New(f.handler(programLevel, ""))
+	logger := slog.New(f.handler(programLevel, "").WithAttrs([]slog.Attr{slog.String(mountLoggerIdKey,
+		fmt.Sprintf("%s-%s", MountFsName, MountInstanceId))}))
```
kislaykishore: I'm not sure about the performance characteristics of using Sprintf here. Should we do something like:
```suggestion
	logger := slog.New(f.handler(programLevel, "").WithAttrs([]slog.Attr{slog.String(mountLoggerIdKey, MountInstanceId), slog.String("fs", MountFsName))}))
```

cmd/root.go
```diff
@@ -23,21 +23,53 @@ import (
...
+	wrapperConfig := make(map[string]interface{})
+	cliFlags := make(map[string]interface{})
+	cmd.PersistentFlags().VisitAll(func(f *pflag.Flag) {
```
kislaykishore, line 38: How about we print the os.Args[1:] alone? This will print the flag details in a single string so it will just contain only those flags that user actually passed.

cmd/root.go
kislaykishore: What's the requirement of printing the CLI flags and config-file details when we are printing the resolved configuration at the end?

kislaykishore (commented):
Could we also split this PR into two - one for the config-cli-logging related changes and the other for mount id?

cmd/root.go
meet2mky: Requirement was outlined in https://b.corp.google.com/issues/340798905#comment5. This is done as per previous discussion and it brings more observability into what customer provided/what optimizations were made and what was the final config was used in GCSFuse. These changes would ensure we get everything from logs at a single place.

internal/logger/logger.go
meet2mky: There is a performance impact if this `fmt.Sprintf` is being called repeatedly. However this is not the case here as it's done during the mounting process and initialization is done only once during startup.

kislaykishore: Makes sense. Thanks for explaining.

I'd also like to understand why did you choose to log one attribute instead of two: mount-id and bucket separately?

Secondly, how do you plan to assess the performance characteristics of this change?

internal/logger/logger.go
meet2mky: Added getter for both `MountFSName()` and `MountInstanceID()`.

meet2mky: Done

internal/logger/logger_test.go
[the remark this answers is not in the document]
meet2mky: Done

internal/logger/logger_test.go
meet2mky: moot as per new review fixes.

internal/logger/logger.go
meet2mky: Done

cmd/root.go
meet2mky: changed to this `map[string]any`. It's more readable and good enough. keys in map[string] aren't actually Flag so we can't use `FlagToAnyMap`.

internal/logger/logger_test.go
meet2mky: It calls this https://pkg.go.dev/log/slog#NewJSONHandler internally so it shouldn't fail as the method signature doesn't return any error. Any reason you think this might fail?

cmd/config_validation_test.go
meet2mky: Done

cmd/legacy_main.go
meet2mky: Done

internal/logger/logger.go
meet2mky: Done!

cmd/legacy_main.go
meet2mky: Done!

cmd/legacy_main.go
meet2mky: Done

internal/logger/logger_test.go
meet2mky: Done.

internal/logger/logger_test.go
meet2mky: Done

internal/logger/logger.go
meet2mky: Done

internal/logger/logger.go
meet2mky: I have separated the method into 2 parts:

A) One to generate the instance ID
B) to setup the instance ID

internal/logger/logger.go
meet2mky: This is done to avoid initialisation of logger twice.

cmd/root.go
meet2mky: Added.

internal/logger/logger.go
meet2mky: Since this attribute's core requirement is to distinguish logs when multiple mount instance are used in the same pod. We can use two attributes as well but that would cause inconvenience to filter logs. Please refer [go/gcsfuse-mount-logging-improvements](http://go/gcsfuse-mount-logging-improvements). We chose a single attribute instead of choosing two as latter would make you add the filter twice to get the logs from the same mount in LogExplorer.

meet2mky: As per this https://go.dev/blog/slog. Adding custom attribute shouldn't increase a lot of cpu as formatting is pre-computed and also we don't have large size attribute so we shouldn't see any increase in memory usage. But we can perform due diligence for this. Do you know of any micro benchmarks or some setup to evaluate the performance characteristics of this?

cmd/root.go
meet2mky: Will there be any difference? Single line in my opinion would not be readable in logExplorer. And we would have to manually remove the --foreground flag from this also because when user starts gcsfuse in background mode then we modify the flags.

cmd/root.go
meet2mky: How logs will display in logExplorer can be found [here]( https://docs.google.com/document/d/1Xc80Njw43z3xzttZQEtfT6rD-sdayZ5xTftRUYOMNUw/edit?tab=t.0#bookmark=id.uaytdntlid21)

meet2mky:
Hi @kislaykishore, I will split this into two parts. Since logging changes depend on the mountInstanceID i have kept both of them together to verify end-to-end logging enhancement.

internal/logger/logger.go
meet2mky: PS: I did perform the perf testing which we have on the PR presubmit. Results are attached in the PR description.

cmd/legacy_main_test.go
```diff
@@ -315,6 +358,9 @@ func (t *MainTest) TestForwardedEnvVars() {
 	}, {
 		inputEnvVars:             map[string]string{"GRPC_GO_LOG_VERBOSITY_LEVEL": "99", "GRPC_GO_LOG_SEVERITY_LEVEL": "INFO"},
 		expectedForwardedEnvVars: []string{"GRPC_GO_LOG_VERBOSITY_LEVEL=99", "GRPC_GO_LOG_SEVERITY_LEVEL=INFO"},
+	}, {
+		inputEnvVars:             map[string]string{"GCSFUSE_IN_BACKGROUND_MODE": "true", "GCSFUSE_MOUNT_INSTANCE_ID": logger.MountInstanceID()},
```
gargnitingoogle, line 362: Let's simplify this by using an integer literal instead of dependency on the call or output of `logger.MountInstanceID()`.


```suggestion
		inputEnvVars:             map[string]string{"GCSFUSE_IN_BACKGROUND_MODE": "true", "GCSFUSE_MOUNT_INSTANCE_ID": <some-integer-literal>},
		expectedForwardedEnvVars: []string{"GCSFUSE_IN_BACKGROUND_MODE=true", "GCSFUSE_MOUNT_INSTANCE_ID=" + <same-integer-literal>},
```

internal/logger/logger.go
```diff
@@ -97,8 +109,44 @@ func InitLogFile(newLogConfig cfg.LoggingConfig) error {
...
+}
+
+// setupMountInstanceID handles the retrieval of mountInstanceId if GCSFuse is in
+// background mode or generates one if running in foreground mode.
+func setupMountInstanceID() {
+	if _, ok := os.LookupEnv(GCSFuseInBackgroundMode); ok {
+		// If GCSFuse is in background mode then look for the MountInstanceId in env which was set by the caller of demonize run.
+		mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey)
+		if !ok || mountInstanceID == "" {
+			mountInstanceID = defaultMountInstanceID
```
gargnitingoogle, line 139: In what situation can we expect to land here ? If there are valid cases, then let's document those here. If not a valid scenario, then this must be considered an error.

cmd/legacy_main_test.go
meet2mky: This won't work as the forwardedEnvVars forwards the current logger.MountInstanceID() and we can't mock that implementation without increase more complexity. So if we were to do this we need to be able to control what logger.MountInstanceID() returns. This would increase more complexity for this test.

internal/logger/logger.go
meet2mky: I don't think there are any valid cases. We are passing multiple variables this way. This is only present for safe guarding if anything happens which is not expected in normal path. I don't see any reason why this would fail. Do you have anything in mind when we pass the environment variable and it doesn't get passed to the process?

meet2mky: The reason we aren't treating this as an error is that it's not a good user experience to fail mount (even if this not possible) just because you couldn't get the mountID.

cmd/root_test.go
```diff
@@ -126,7 +244,7 @@ func TestArgsParsing_MountPoint(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			var bucketName, mountPoint string
-			cmd, err := newRootCmd(func(_ *cfg.Config, b string, m string) error {
+			cmd, err := newRootCmd(func(_ *cfg.Config, _ map[string]interface{}, b string, m string) error {
```
gargnitingoogle, line 247: Use `any` instead of `interface{}` everywhere for consistency.

internal/logger/logger.go
```diff
@@ -97,8 +109,44 @@ func InitLogFile(newLogConfig cfg.LoggingConfig) error {
...
+	setupMountInstanceID()
+	initializeDefaultLogger()
+}
+
+// generateMountInstanceID generates a random string of mountInstanceIDLength characters
+// from a new UUID. It falls back to defaultMountInstanceID on failure.
+func generateMountInstanceID() string {
+	uuid, err := newRandomUUID()
+	if err != nil {
+		log.Printf("Could not generate random UUID for logger, err %v. using default: %v", err, defaultMountInstanceID)
```
gargnitingoogle, line 124: This should be a fatal error, not an ignored error. When can this happen at all ?

internal/logger/logger.go
```diff
@@ -32,14 +35,23 @@ import (
...
-	ProgrammeName           string = "gcsfuse"
-	GCSFuseInBackgroundMode string = "GCSFUSE_IN_BACKGROUND_MODE"
-	textFormat              string = "text"
+	ProgrammeName                = "gcsfuse"
+	GCSFuseInBackgroundMode      = "GCSFUSE_IN_BACKGROUND_MODE"
+	GCSFuseMountInstanceIDEnvKey = "GCSFUSE_MOUNT_INSTANCE_ID"
+	textFormat                   = "text"
+	mountLoggerIDKey             = "mount_id"
+	// Max possible length can be 32 as UUID has 32 characters excluding 4 hyphens.
+	mountInstanceIDLength = 8
```
gargnitingoogle, line 44: What is the harm in making this full 32 characters ?

cmd/legacy_main_test.go
gargnitingoogle: Then there is no point in passing GCSFUSE_MOUNT_INSTANCE_ID in the input env vars, right? If so, let's remove it from inputEnvVars because otherwise it's misleading.

internal/logger/logger.go
gargnitingoogle: > Do you have anything in mind when we pass the environment variable and it doesn't get passed to the process?

it was possible if it was not handled correctly at forwardedEnvVars in cmd/legacy_main.go , but it is handled there, so there is no valid case for this failure. This is why I think this should be a failure, because it tells me that something has gone wrong somewhere.

gargnitingoogle (changes requested):
Added some comments

internal/logger/logger.go
kislaykishore: That's fine - I don't see much of a problem either way.

cmd/root.go
kislaykishore: Sounds good.

cmd/root.go
```diff
@@ -23,21 +23,53 @@ import (
...
+		if f.Changed {
+			cliFlags[f.Name] = f.Value.String()
+		}
+	})
+	if _, ok := os.LookupEnv(logger.GCSFuseInBackgroundMode); ok {
+		// Do not display --foreground flag to the user in logs if user
+		// hasn't passed this flag and was added by gcsufse during demonized run.
```
kislaykishore, line 45:
```suggestion
		// hasn't passed this flag and was added by gcsfuse during demonized run.
```

cmd/root.go
```diff
@@ -23,21 +23,53 @@ import (
...
+		delete(cliFlags, "foreground")
+	}
+	wrapperConfig["CliFlags"] = cliFlags
```
kislaykishore, line 48:
```suggestion
	wrapperConfig["CLIFlags"] = cliFlags
```

internal/logger/logger.go
kislaykishore: Where are the getters added? Also, could you un-export the variables.

internal/logger/logger.go
```diff
@@ -32,14 +35,23 @@ import (
 // used to filter the logs from syslog and write it to respective log files -
 // gcsfuse.log in case of GCSFuse.
 const (
-	ProgrammeName           string = "gcsfuse"
-	GCSFuseInBackgroundMode string = "GCSFUSE_IN_BACKGROUND_MODE"
-	textFormat              string = "text"
+	ProgrammeName                = "gcsfuse"
```
kislaykishore, line 38:
```suggestion
	ProgramName                = "gcsfuse"
```

internal/logger/logger.go
```diff
@@ -109,12 +157,25 @@ func init() {
 	defaultLogger = defaultLoggerFactory.newLogger(cfg.INFO)
 }
 
-// SetLogFormat updates the log format of default logger.
-func SetLogFormat(format string) {
-	if format == defaultLoggerFactory.format {
-		return
+func MountInstanceID() string {
```
kislaykishore, line 160: Using init can lead to complex code since it's invoked implicitly. Consider using `sync.Once` here to generate the mount instance id once instead of using the init method.

internal/logger/logger_test.go
```diff
@@ -16,34 +16,38 @@ package logger
...
 	suite.Suite
 }
 
+func expectedLogRegex(format, severity, message string) string {
+	if format == "text" {
+		return fmt.Sprintf(textLogPattern, severity, message)
+	}
+	if format == "json" {
+		return fmt.Sprintf(jsonLogPattern, severity, message)
+	}
```
kislaykishore, line 47: Use switch-case

internal/logger/logger_test.go
```diff
@@ -335,7 +342,142 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+			setupMountInstanceID()
+
+			if tc.expectDefaultID {
+				assert.Equal(t, defaultMountInstanceID, mountInstanceID)
+			} else if tc.expectedID != "" {
+				assert.Equal(t, tc.expectedID, mountInstanceID)
+			} else {
+				assert.NotEqual(t, defaultMountInstanceID, mountInstanceID)
+				assert.Len(t, mountInstanceID, mountInstanceIDLength)
+				assert.Regexp(t, fmt.Sprintf("^[0-9a-f]{%d}$", mountInstanceIDLength), mountInstanceID)
```
kislaykishore, line 479: It's recommended not to have if-else in test cases because they harm readability. Please split them into three separate tests.

internal/logger/logger_test.go
```diff
@@ -141,7 +144,7 @@ func (t *LoggerTest) TestTextFormatLogs_LogLevelERROR() {
 
 func (t *LoggerTest) TestTextFormatLogs_LogLevelWARNING() {
 	var expected = []string{
-		"", "", "", textWarningString, textErrorString,
+		"", "", "", expectedLogRegex("text", "WARNING", "www.warningExample.com"), expectedLogRegex("text", "ERROR", "www.errorExample.com"),
```
kislaykishore, line 147: Thanks for doing this.

internal/logger/logger_test.go
```diff
@@ -16,34 +16,38 @@ package logger
...
+	if format == "text" {
+		return fmt.Sprintf(textLogPattern, severity, message)
+	}
+	if format == "json" {
+		return fmt.Sprintf(jsonLogPattern, severity, message)
+	}
+	return ""
+}
+
 func TestLoggerSuite(t *testing.T) {
```
kislaykishore, line 51: Seems like using suites is redundant here. I know that's beyond the scope of this PR, so feel free to keep it this way.

internal/logger/logger_test.go
```diff
@@ -335,7 +342,142 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+			}
+			setupMountInstanceID()
+			initializeDefaultLogger()
+			SetLogFormatAndFsName(test.format, test.fsName)
+			assert.NotNil(t, defaultLoggerFactory)
+			assert.NotNil(t, defaultLogger)
+
+			// Create a logger using defaultLoggerFactory that writes to buffer.
+			var buf bytes.Buffer
+			redirectLogsToGivenBuffer(&buf, defaultLoggerFactory.level)
```
kislaykishore, line 398: This too should be in the Arrange block.

internal/logger/logger_test.go
```diff
@@ -335,7 +342,142 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+		t.T().Run(test.name, func(t *testing.T) {
+			t.Setenv(GCSFuseInBackgroundMode, "true")
+			if test.instanceIDSet {
+				t.Setenv(GCSFuseMountInstanceIDEnvKey, test.expectedInstanceID)
+			}
+			setupMountInstanceID()
+			initializeDefaultLogger()
+			SetLogFormatAndFsName(test.format, test.fsName)
+			assert.NotNil(t, defaultLoggerFactory)
+			assert.NotNil(t, defaultLogger)
```
kislaykishore, line 394: Use require instead of assert to force immediate test failure.

internal/logger/logger_test.go
```diff
@@ -335,7 +342,142 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+	id := generateMountInstanceID()
+
+	assert.Len(t.T(), id, mountInstanceIDLength)
+	assert.NotEqual(t.T(), defaultMountInstanceID, id)
+	assert.Regexp(t.T(), fmt.Sprintf("^[0-9a-f]{%d}$", mountInstanceIDLength), id)
+}
+
+func (t *LoggerTest) TestGenerateMountInstanceID_Failure() {
+	originalNewRandom := newRandomUUID
+	defer func() { newRandomUUID = originalNewRandom }()
```
kislaykishore, line 419: Instead of using  newRandomUUID member, consider passing an argument to the generateMountInstanceID function or create a separate function that generates error. That'll make the tests more robust and isolated from each other.

meet2mky:
Removing the `review-reminder` and splitting this PR as it has become huge.

cmd/legacy_main_test.go
meet2mky: Done! in the split PR, PTAL https://github.com/GoogleCloudPlatform/gcsfuse/pull/3872

internal/logger/logger.go
meet2mky: I think, the main question here is: What do we do if we couldn't get the `MountInstaceID` if it was running in whatever mode background or foreground.

There are two ways to think about it.
A) Fail the mount -- This makes a part of logging feature which is having custom attribute on the logger a critical path during the mount.

B) Let the mount continue -- This ensures whatever the reason may be for error (error will be logged of course) and we couldn't get the `MountInstanceID` so we may not have a custom attribute with Random MountInstanceID on the logger attribute but we continue the mount to succeed.

I think option B, is better as this separates logging failure from actual mount operations. Let me know what your opinion is?

internal/logger/logger.go
meet2mky: Please see: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3818#discussion_r2415594839

internal/logger/logger.go
meet2mky: It's not about the harm but I think, unless we require it and we find the practical need of having a higher length `MountInstaceID`, we shouldn't increase it just because we can. Having higher length UUID would just increase the logging size without any practical benefits.

internal/logger/logger_test.go
meet2mky: Refactored in the split PR.

internal/logger/logger_test.go
meet2mky: Refactored in the split PR:

meet2mky:
Splitting the PR into multiple smaller PR(s):

First Split: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3872
Second Split: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3889
Third Split: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3894
fourth split: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3914
Fifth Split: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3926
Sixth split: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3933

meet2mky:
Closing as the changes are all merged in split PR(s).

PR #3821, opened 2025-09-18
refactor: migrate streaming writes test package to configs

ashmeenkaur:
### Description
Refactoring of the streaming writes integration test suite. The primary goal is to transition these tests from a flag-based setup to a common config based testing. This change standardizes how test environments are initialized and managedin GCE and GKE test environment. It involves updating the core test setup logic and adapting individual test cases to leverage the new centralized configuration.

b/445951464

### Testing details
1. Manual - Manually verified that the tests work with config file.
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/streaming_writes/buffer_size_test.go
+++ b/tools/integration_tests/streaming_writes/buffer_size_test.go
@@ -41,17 +41,6 @@ func TestWritesWithDifferentConfig(t *testing.T) {
 	defer func() {
 		testEnv.cfg.GCSFuseMountedDirectory = oldMntDir
 	}()
-	//setup.SetGlobalVars(&test_suite.TestConfig{
-	//	TestBucket:          setup.TestBucket(),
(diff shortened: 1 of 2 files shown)


tools/integration_tests/local_file/setup_test.go
```diff
@@ -69,6 +69,8 @@ func TestMain(m *testing.M) {
 		{"--implicit-dirs=true", "--rename-dir-limit=3", "--enable-streaming-writes=false"},
 		{"--implicit-dirs=false", "--rename-dir-limit=3", "--enable-streaming-writes=false"},
 		{"--implicit-dirs=false", "--rename-dir-limit=3", "--enable-streaming-writes=false", "--client-protocol=grpc"},
+		// Running with write-global-max-blocks=0 falls back to disk (local file tests behavior)
```
ashmeenkaur: @PranjalC100 can you add this case to local file test migration?

PR #3827, opened 2025-09-19
fix: Disabling CRC validation for zonal buckets & fixing read cache tests

abhishek10004:
1. Making crc validation a no op for file cache in zonal buckets 
2. Fixing a test in parallel Downloader UTs.

b/446066150
b/446447164

### Testing details
1. Manual - Done
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -517,7 +517,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.IntP("file-cache-download-chunk-size-mb", "", 200, "Size of chunks in MiB that each concurrent request downloads.")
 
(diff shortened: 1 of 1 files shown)


internal/cache/file/downloader/job.go
```diff
@@ -525,7 +525,8 @@ func (job *Job) GetStatus() JobStatus {
 // Compares CRC32 of the downloaded file with the CRC32 from GCS object metadata.
 // In case of mismatch deletes the file and corresponding entry from file cache.
 func (job *Job) validateCRC() (err error) {
-	if !job.fileCacheConfig.EnableCrc {
+	// Skipping crc check for zonal buckets
```
vadlakondaswetha: please create a bug and link it here with appropriate comment - disabling it till we solve it for unfinalized objects

vadlakondaswetha (approved):

PR #3831, opened 2025-09-22
refactor: local file test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate local file test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of local file package so it can use config file.

[b/445952157](https://buganizer.corp.google.com/issues/445952157)

### Testing details
2. Unit tests - NA


tools/integration_tests/test_config.yaml
```diff
@@ -149,3 +149,25 @@ rename_dir_limit:
...
+
+local_file:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "${BUCKET_NAME}"
+    log_file: # Optional
+    run_on_gke: false
+    configs:
+      - flags:
+          - "--implicit-dirs=true --rename-dir-limit=3 --enable-streaming-writes=false"
+          - "--implicit-dirs=false --rename-dir-limit=3 --enable-streaming-writes=false"
```
ashmeenkaur: Nit: let's remove this flag as the test set is not dependent on implicit-dirs flag and we have implicit-dirs=false in grpc combination

tools/integration_tests/local_file/setup_test.go
```diff
@@ -38,55 +39,65 @@ import (
...
+		cfg.LocalFile[0].Configs[1].Flags = []string{
+			"--rename-dir-limit=3 --write-block-size-mb=1 --write-max-blocks-per-file=2 --write-global-max-blocks=-1 --client-protocol=grpc",
+			"--rename-dir-limit=3 --write-block-size-mb=1 --write-max-blocks-per-file=2 --write-global-max-blocks=-1",
+		}
+		cfg.LocalFile[0].Configs[1].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": false}
+	}
+
+	// 2. Create storage client before running tests.
 	ctx = context.Background()
+	bucketType := setup.BucketTestEnvironment(ctx, cfg.LocalFile[0].TestBucket)
```
ashmeenkaur: Can we instead use TestEnvironment function

tools/integration_tests/local_file/setup_test.go
```diff
@@ -38,55 +39,65 @@ import (
...
 
-	successCode := static_mounting.RunTests(flagsSet, m)
+	successCode := static_mounting.RunTestsWithConfigFile(&cfg.LocalFile[0], flags, m)
 
 	if successCode == 0 {
-		successCode = only_dir_mounting.RunTests(flagsSet, onlyDirMounted, m)
+		successCode = only_dir_mounting.RunTestsWithConfigFile(&cfg.LocalFile[0], flags, onlyDirMounted, m)
 	}
 
 	// Dynamic mounting tests create a bucket and perform tests on that bucket,
```
ashmeenkaur: update comment as it no longer creates a bucket. This also means that it can now run with HNS bucket too.

tools/integration_tests/local_file/setup_test.go
```diff
@@ -38,55 +39,65 @@ import (
...
 	// which is not a hierarchical bucket. So we are not running those tests with
 	// hierarchical bucket.
-	if successCode == 0 && !setup.IsHierarchicalBucket(ctx, storageClient) {
-		successCode = dynamic_mounting.RunTests(ctx, storageClient, flagsSet, m)
+	if successCode == 0 && !setup.ResolveIsHierarchicalBucket(ctx, setup.TestBucket(), storageClient) {
+		successCode = dynamic_mounting.RunTestsWithConfigFile(&cfg.LocalFile[0], flags, m)
 	}
 
 	// Clean up test directory created.
```
ashmeenkaur: Is this required? I think we do clean up the test directory before running tests.

tools/integration_tests/test_config.yaml
PranjalC100: removed!

tools/integration_tests/local_file/setup_test.go
PranjalC100: updated comment, but it still doesnt perform tests on hns buckets..

tools/integration_tests/local_file/setup_test.go
```diff
@@ -38,59 +38,65 @@ import (
...
+		successCode = only_dir_mounting.RunTestsWithConfigFile(&cfg.LocalFile[0], flags, onlyDirMounted, m)
 	}
 
-	// Dynamic mounting tests create a bucket and perform tests on that bucket,
-	// which is not a hierarchical bucket. So we are not running those tests with
-	// hierarchical bucket.
-	if successCode == 0 && !setup.IsHierarchicalBucket(ctx, storageClient) {
-		successCode = dynamic_mounting.RunTests(ctx, storageClient, flagsSet, m)
+	// Dynamic mounting tests perform tests on non-HNS buckets.
+	if successCode == 0 && !setup.ResolveIsHierarchicalBucket(ctx, setup.TestBucket(), storageClient) {
```
PranjalC100: remove hns condition, run with it as well.

tools/integration_tests/local_file/setup_test.go
PranjalC100: running for hns as well now

ashmeenkaur (dismissed):

ashmeenkaur (approved):

PR #3832, opened 2025-09-22
fix(streaming): Enhance streaming writes warning log messages and remove duplication of warning logs.

meet2mky:
### Description
Enhance streaming writes warning/info log messages more customer friendly and point a link to semantic for more details on understanding these warnings/info logs. This would help customers and CE's can understand the warning logs themself more easily by following the semantic docs link to read the GCSFuse write paths supported operations and details.

**Note: Fixed a bug where when a sequential write goes in to truncate to lower then duplicate logs were getting printed as warning.**
b/438404050

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - Part of presubmit


internal/fs/inode/file.go
```diff
@@ -632,7 +629,9 @@ func (f *FileInode) writeUsingBufferedWrites(ctx context.Context, data []byte, o
 	}
 	// Fall back to temp file for Out-Of-Order Writes.
 	if errors.Is(err, bufferedwrites.ErrOutOfOrderWrite) {
-		logger.Infof("Falling back to staged writes on disk for file %s (inode %d) due to err: %v.", f.Name(), f.ID(), err.Error())
+		logger.Warnf("Out of order write detected. File %s will now use legacy staged writes because "+
+			"streaming writes is supported for sequential writes to new/empty files. "+
+			"For more details, see: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#writes", f.name.String())
```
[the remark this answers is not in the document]
meet2mky: Done, Changed the streaming semantics but kept the out of order write in place.

internal/fs/inode/file.go
```diff
@@ -632,7 +633,7 @@ func (f *FileInode) writeUsingBufferedWrites(ctx context.Context, data []byte, o
 	}
 	// Fall back to temp file for Out-Of-Order Writes.
 	if errors.Is(err, bufferedwrites.ErrOutOfOrderWrite) {
-		logger.Infof("Falling back to staged writes on disk for file %s (inode %d) due to err: %v.", f.Name(), f.ID(), err.Error())
+		logger.Warnf("Out of order write detected. File %s will now use legacy staged writes. "+StreamingWritesSemantics, f.name.String())
```
meet2mky: Made the LogLevel consistent as similar OutOfOrder write had `warn` level in same file at other place but here it had `info` level

meet2mky: resolved as per offline discussion.

internal/fs/inode/file.go
```diff
@@ -41,7 +41,11 @@ import (
 
 // A GCS object metadata key for file mtimes. mtimes are UTC, and are stored in
 // the format defined by time.RFC3339Nano.
-const FileMtimeMetadataKey = gcs.MtimeMetadataKey
+const (
+	FileMtimeMetadataKey     = gcs.MtimeMetadataKey
+	StreamingWritesSemantics = "Streaming writes is supported for sequential writes to new/empty files. " +
```
ashmeenkaur, line 47: StreamingWritesSemantics are incorrect for zb where appends are also supported via streaming writes. Shall we use different message for zb and non zb?

meet2mky: That's correct, ZB is not yet out on the semantics doc so we can't include in the error message which points to semantic doc saying about zonal buckets. We can update the error message with relevant link once the semantic doc is updated with the ZB related streaming writes functionality. Wdyt?

ashmeenkaur: Let's create a backlog bug to update this along with semantics.

charith87 (approved):

internal/fs/inode/file.go
meet2mky: Done!

meet2mky:
Added a TODO comment to update it again once ZB release and semantic doc updates are complete. @ashmeenkaur @charith87 Please approve again.

ashmeenkaur (approved):

charith87 (approved):

PR #3836, opened 2025-09-23
feat(metrics): Add support for up-down counter

kislaykishore:
### Description
Things like number of open readers can be represented well via an updown counter. Let's add support for that.

b/446663088

### Testing details
1. Manual - Yes
2. Unit tests - Added
3. Integration tests - NA

No


metrics/otel_metrics.go
```diff
@@ -2839,6 +2844,26 @@ func (o *otelMetrics) GcsRetryCount(
 
 }
 
+func (o *otelMetrics) TestUpdownCounter(
+	inc int64) {
+	o.testUpdownCounterAtomic.Add(inc)
+
```
raj-prince, line 2850:
```suggestion
```

metrics/otel_metrics.go
```diff
@@ -2839,6 +2844,26 @@ func (o *otelMetrics) GcsRetryCount(
...
+	switch requestType {
+	case "attr1":
+		o.testUpdownCounterWithAttrsRequestTypeAttr1Atomic.Add(inc)
+	case "attr2":
+		o.testUpdownCounterWithAttrsRequestTypeAttr2Atomic.Add(inc)
+	default:
+		updateUnrecognizedAttribute(requestType)
+		return
+	}
+
```
raj-prince, line 2864:
```suggestion
```

PR #3839, opened 2025-09-23
chore(rapid): Enable e2e test packages for ZB in louhi

anushka567:
### Description
Same as title
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -239,30 +239,31 @@ TEST_DIR_NON_PARALLEL=(
 
 # For Zonal buckets : Test directory arrays
 TEST_DIR_PARALLEL_ZONAL=(
+  buffered_read
+  concurrent_operations
+  dentry_cache
+  explicit_dir
   gzip
+  implicit_dir
   interrupt
   kernel_list_cache
   local_file
+  log_content
   log_rotation
   monitoring
-  mounting
   mount_timeout
+  mounting
   negative_stat_cache
+  operations
+  #rapid_appends
   read_large_files
+  readdirplus
+  release_version
   rename_dir_limit
   stale_handle
-  write_large_files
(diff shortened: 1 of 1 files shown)


tools/cd_scripts/e2e_test.sh
```diff
@@ -244,27 +244,33 @@ TEST_DIR_PARALLEL_ZONAL=(
...
+  explicit_dir
+  implicit_dir
+  log_content
+  operations
+  streaming_writes
+  release_version
+  readdirplus
+  dentry_cache
+  buffered_read
+  unfinalized_object
```
gargnitingoogle, line 265: Can you sort these to make it easy to look at what's missing/disabled ?

tools/cd_scripts/e2e_test.sh
```diff
@@ -244,27 +244,33 @@ TEST_DIR_PARALLEL_ZONAL=(
...
   stale_handle
   write_large_files
-  #concurrent_operations
-  #explicit_dir
-  #implicit_dir
-  #list_large_dir
-  #log_content
-  #operations
-  #streaming_writes
+  concurrent_operations
```
gargnitingoogle: I still see the following packages missing in it.
- benchmarking
- read_cache
Please add them too.

As for package `rapid_appends`, even though we aren't enabling, please add it here as commented out, so that it's easy to spot in the future as it needs to be enabled in the future.

anushka567: benchmarking package is not run in louhi. read_cache is run seperately so it is not in the common list , you can find the logic in the script itself . As for rapid_appends, I will add that commented out.

tools/cd_scripts/e2e_test.sh
anushka567: Done.

PR #3842, opened 2025-09-24
feat(dns_cache): Guard HTTP/1 DNS cache enablement with a boolean flag

kislaykishore:
### Description
Rename the erstwhile http-dns-cache-ttl-secs flag to enable-http-dns-cache which is a boolean flag. The DNS caching infrastructure will take care of handling the TTL.

b/446170703

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

No

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -467,13 +467,13 @@ type GcsConnectionConfig struct {
 
 	CustomEndpoint string `yaml:"custom-endpoint"`
 
+	EnableHttpDnsCache bool `yaml:"enable-http-dns-cache"`
+
 	ExperimentalEnableJsonRead bool `yaml:"experimental-enable-json-read"`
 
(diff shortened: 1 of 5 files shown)


cfg/config.go
```diff
@@ -750,6 +750,12 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	flagSet.BoolP("enable-http-dns-cache", "", false, "Enables DNS cache for HTTP/1 connections")
```
liuyuan10, line 753: so it will be false by default and each users need to set it to opt-in?

kislaykishore: No, we will set it to true in a subsequent pr.

charith87 (approved):

PR #3845, opened 2025-09-24
refactor(cd): fixes in cd script for robustness

anushka567:
### Description
Fixes for 
- b/443039033 targeting rhel/rocky-linux-8
- b/447064201 targeting rhel/centos/rocky-linux-9s
- b/447059258 targeting rhel/centos/rocky-linux-10s
### Testing details
1. Manual - yes
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -68,14 +68,75 @@ gcloud storage cp gs://gcsfuse-release-packages/version-detail/details.txt .
 # Writing VM instance name to details.txt (Format: release-test-<os-name>)
 curl http://metadata.google.internal/computeMetadata/v1/instance/name -H "Metadata-Flavor: Google" >>details.txt
 
-# Based on the os type(from vm instance name) in detail.txt, run the following commands to add starterscriptuser
-if grep -q ubuntu details.txt || grep -q debian details.txt; then
-	#  For ubuntu and debian os
-	sudo adduser --ingroup google-sudoers --disabled-password --home=/home/starterscriptuser --gecos "" starterscriptuser
-else
-	#  For rhel and centos
-	sudo adduser -g google-sudoers --home-dir=/home/starterscriptuser starterscriptuser
+# Function to create the local user
+create_user() {
+  if id "${USERNAME}" &>/dev/null; then
+    echo "User ${USERNAME} already exists."
+    return 0
+  fi
+
+  echo "Creating user ${USERNAME}..."
+  if grep -qi -E 'ubuntu|debian' details.txt; then
+    # For Ubuntu and Debian
+    sudo adduser --disabled-password --home "${HOMEDIR}" --gecos "" "${USERNAME}"
+  elif grep -qi -E 'rhel|centos|rocky' details.txt; then
+    # For RHEL, CentOS, Rocky Linux
+    sudo adduser --home-dir "${HOMEDIR}" "${USERNAME}"
+    # Optionally disable password login for the local user, assuming SSH key access
+    sudo passwd -d "${USERNAME}"
+  else
+    echo "Unsupported OS type in details.txt"
+    return 1
+  fi
+
+  if [ $? -eq 0 ]; then
+    echo "User ${USERNAME} created successfully."
+    return 0
+  else
+    echo "Failed to create user ${USERNAME}."
+    return 1
+  fi
+}
+
+# Function to grant sudo access by creating a file in /etc/sudoers.d/
+grant_sudo() {
+  if ! id "${USERNAME}" &>/dev/null; then
+    echo "User ${USERNAME} does not exist. Cannot grant sudo."
+    return 1
+  fi
+
+  SUDOERS_FILE="/etc/sudoers.d/${USERNAME}"
+
+  if sudo test -f "${SUDOERS_FILE}"; then
+    echo "Sudoers file ${SUDOERS_FILE} already exists."
+  else
+    echo "Granting ${USERNAME} NOPASSWD sudo access..."
+    # Create the sudoers file with the correct content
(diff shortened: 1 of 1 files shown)


tools/cd_scripts/e2e_test.sh
```diff
@@ -68,15 +68,77 @@ gcloud storage cp gs://gcsfuse-release-packages/version-detail/details.txt .
...
+  fi
+  return 0
+}
+################################################################################
+# Main script execution flow starts here.
+# The script will first attempt to create the user specified by $USERNAME.
+# If the user creation is successful, it will then proceed to grant sudo
+# privileges to the newly created user.
+################################################################################
+USERNAME=starterscriptuser
```
gargnitingoogle, line 151: It's not a good idea to declare variables USERNAME, HOMEDIR later and use them in a function defined much earlier. It may work, but it's easy to get it wrong. I'd suggest one of the following:
1. Pass these as variables to the function, [PREFERRED]
```sh
function create_user() {
  local USER=$1
  local HOMEDIR=$2
  ...
}
...
if $(create_user ${USER} ${HOMEDIR}); then
```
2. Declare/assign these constants at the top of the script (as constants as possible).

gargnitingoogle: Same goes for the function `grant_sudo` .

tools/cd_scripts/e2e_test.sh
```diff
@@ -68,15 +68,77 @@ gcloud storage cp gs://gcsfuse-release-packages/version-detail/details.txt .
...
+  fi
+
+  SUDOERS_FILE="/etc/sudoers.d/${USERNAME}"
+
+  if sudo test -f "${SUDOERS_FILE}"; then
+    echo "Sudoers file ${SUDOERS_FILE} already exists."
+  else
+    echo "Granting ${USERNAME} NOPASSWD sudo access..."
+    # Create the sudoers file with the correct content
+    if ! echo "${USERNAME} ALL=(ALL:ALL) NOPASSWD:ALL" | sudo tee "${SUDOERS_FILE}" > /dev/null; then
```
gargnitingoogle, line 129: Is the folder `/etc/sudoers.d/` guaranteed to exist beforehand ? If not, then this will fail.

gargnitingoogle, line 129: To make it a bit less confusing to read

```suggestion
    if ! ( echo "${USERNAME} ALL=(ALL:ALL) NOPASSWD:ALL" | ( sudo tee "${SUDOERS_FILE}" > /dev/null ) ); then
```

tools/cd_scripts/e2e_test.sh
```diff
@@ -68,15 +68,77 @@ gcloud storage cp gs://gcsfuse-release-packages/version-detail/details.txt .
...
+      echo "Failed to create sudoers file." >&2
```
gargnitingoogle, line 130:
```suggestion
      echo "Failed to create sudoers file ${SUDOERS_FILE}" >&2
```

tools/cd_scripts/e2e_test.sh
```diff
@@ -68,15 +68,77 @@ gcloud storage cp gs://gcsfuse-release-packages/version-detail/details.txt .
...
+}
+################################################################################
+# Main script execution flow starts here.
+# The script will first attempt to create the user specified by $USERNAME.
+# If the user creation is successful, it will then proceed to grant sudo
+# privileges to the newly created user.
+################################################################################
+USERNAME=starterscriptuser
+HOMEDIR="/home/${USERNAME}"
+if create_user; then
```
gargnitingoogle: We have a `set -e` at the top, so this check is unnecessary. Anyway we proceed to the next line only if the previous line passes. Also It'll be more readable if
```sh
create_user
grant_sudo
```

Even better to just invoke `grant_sudo` directly inside `create_user` when the user is being created, and rename `create_user` to `create_sudoer_user` as that's the only use of `grant_sudo` .

anushka567:
Taking this up as an improvement for now.

tools/cd_scripts/e2e_test.sh
```diff
@@ -31,10 +31,20 @@ fi
 
 # Upgrade gcloud
 echo "Upgrade gcloud version"
-gcloud version
```
anushka567, line 34: This is causing failures in some machine, Skipping as anyways the upgrade is done shortly after.

tools/cd_scripts/e2e_test.sh
```diff
@@ -31,10 +31,20 @@ fi
...
 sudo tar xzf gcloud.tar.gz && sudo cp -r google-cloud-sdk /usr/local && sudo rm -r google-cloud-sdk
-sudo /usr/local/google-cloud-sdk/install.sh
+
+# Conditionally install python3.9 and run gcloud installer with it for RHEL 8 and Rocky 8
+INSTALL_COMMAND="sudo /usr/local/google-cloud-sdk/install.sh --quiet"
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
+    if [[ ($ID == "rhel" || $ID == "rocky") && $VERSION_ID == 8* ]]; then
+        sudo yum install -y python39
+        INSTALL_COMMAND="sudo env CLOUDSDK_PYTHON=/usr/bin/python3.9 /usr/local/google-cloud-sdk/install.sh --quiet"
```
anushka567, line 43: Dependency itself relies on hardcoded python versions , so reflecting the same here (only for rhel/rocky-linux-8) Rest is compatible.

Tulsishah: We can upgrade this for all the OSs but as discussed offline, it requires to get tested on all the machine. So we can avoid this for now as not getting fit in release timeframe.

Tulsishah (approved):

PR #3855, opened 2025-09-30
feat(bufferedread):  Handle -1 for read.global-max-blocks as infinite

vipnydav:
### Description
This PR handles the -1 value for `read.global-max-blocks`. When this parameter is set to -1, it will be interpreted as an infinite value and set to `math.MaxInt32`.

https://b.corp.google.com/issues/448271034

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -134,6 +134,12 @@ func resolveFileCacheAndBufferedReadConflict(v isSet, c *Config) {
 	}
 }
 
+func resolveReadConfig(r *ReadConfig) {
+	if r.GlobalMaxBlocks == -1 {
+		r.GlobalMaxBlocks = math.MaxInt32
+	}
+}
+
 func resolveLoggingConfig(config *Config) {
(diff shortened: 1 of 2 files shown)


cfg/rationalize.go
```diff
@@ -134,6 +134,12 @@ func resolveFileCacheAndBufferedReadConflict(v isSet, c *Config) {
 	}
 }
 
+func resolveReadConfig(r *ReadConfig) {
+	if r.GlobalMaxBlocks == -1 {
+		r.GlobalMaxBlocks = math.MaxInt32
```
raj-prince, line 139: GlobalMaxBlocks is int64 right? - lets do it MaxInt64.

vipnydav: I reduced it because of integer overflow here
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/workerpool/static_worker_pool.go#L59

raj-prince: Okay, then lets go ahead with this only. As 2PB memory is more than enough.

PR #3867, opened 2025-10-08
feat(metrics-landing): Group unused/unimplemented metric labels together.

kislaykishore:
### Description

This will help reduce cardinality of the metrics.

b/450162887

### Testing details
1. Manual - Yes
2. Unit tests - Automated
3. Integration tests - Added

No - metrics changes aren't considered backward incompatible.

### DIFF
--- a/tools/integration_tests/monitoring/prom_test.go
+++ b/tools/integration_tests/monitoring/prom_test.go
@@ -232,7 +232,7 @@ func (testSuite *PromTest) TestSetXAttrMetrics() {
 	err := xattr.Set(path.Join(testSuite.mountPoint, "hello/hello.txt"), "alpha", []byte("beta"))
 
 	assert.Error(testSuite.T(), err)
-	assertNonZeroCountMetric(testSuite, "fs_ops_count", "fs_op", "NotImplemented")
+	assertNonZeroCountMetric(testSuite, "fs_ops_count", "fs_op", "Others")
 }
 
 func (testSuite *PromTest) TestReadMetrics() {


metrics/metric_handle.go
```diff
@@ -50,15 +50,13 @@ const (
...
-	FsOpFallocateAttr          FsOp = "Fallocate"
 	FsOpFlushFileAttr          FsOp = "FlushFile"
 	FsOpForgetInodeAttr        FsOp = "ForgetInode"
 	FsOpGetInodeAttributesAttr FsOp = "GetInodeAttributes"
-	FsOpGetXattrAttr           FsOp = "GetXattr"
-	FsOpListXattrAttr          FsOp = "ListXattr"
 	FsOpLookUpInodeAttr        FsOp = "LookUpInode"
 	FsOpMkDirAttr              FsOp = "MkDir"
 	FsOpMkNodeAttr             FsOp = "MkNode"
+	FsOpNotImplementedAttr     FsOp = "NotImplemented"
```
charith87: I would call this Other instead of NotImplemented. This is also visible to customers and so having it NotImplemented gives the illusion to the customer that something is wrong with their setup.

internal/fs/wrappers/monitoring.go
```diff
@@ -364,25 +364,25 @@ func (fs *monitoring) ReadSymlink(ctx context.Context, op *fuseops.ReadSymlinkOp
 }
 
 func (fs *monitoring) RemoveXattr(ctx context.Context, op *fuseops.RemoveXattrOp) error {
-	return fs.invokeWrapped(ctx, metrics.FsOpRemoveXattrAttr, func(ctx context.Context) error { return fs.wrapped.RemoveXattr(ctx, op) })
+	return fs.invokeWrapped(ctx, metrics.FsOpNotImplementedAttr, func(ctx context.Context) error { return fs.wrapped.RemoveXattr(ctx, op) })
```
charith87: Just trying to understand. Cardinality of say 25 to 19 does not look like it actually benefit much. Was this causing any bottleneck because we are also losing the fs call details as well along with it. Reducing dimensions or reducing the number by exponential ways may help more in this regard or does it help in that?

kislaykishore: This has a multiplicative effect from the cardinality perspective. Cardinality reduction:
fs/ops_count: From 25 to 18
fs/ops_error_count: 16  error categories * 6 = 96
fs/ops_latency: 34 buckets * 6 = 204

metrics/metric_handle.go
kislaykishore: Done.

charith87 (approved):

PR #3868, opened 2025-10-08
refactor: Read cache test migration

ashmeenkaur:
### Description
This pull request refactors the read cache integration tests. The primary goal is to have a common configuration that will be shared by both GCSFuse and GKE test suite. This PR also moves the test setup and teardown for GKE tests to the test suite by storing the artifacts in `/gcsfuse-tmp` directory.

b/439725014
b/432148276

### Testing details
1. Manual - Manually verified that the tests are backward compatible both GCSFuse and GKE and they also work with config file.
2. Unit tests - NA

NA

Note: This PR disables Ram cache tests. These tests will be enabled in follow up PR due to bug b/383682524

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -612,6 +612,9 @@ func AddCacheDirToFlags(flagSets [][]string, testname string) [][]string {
 
 // BuildFlagSets dynamically builds a list of flag sets based on bucket compatibility.
 // bucketType should be "flat", "hns", or "zonal".
+// The run parameter filters flag sets based on the 'Run' field in the test
+// configuration, which typically corresponds to a specific test name. If run is
+// an empty string, all flag sets for the package are returned.
 func BuildFlagSets(cfg test_suite.TestConfig, bucketType string, run string) [][]string {
 	var dynamicFlags [][]string
 
--- a/tools/integration_tests/read_cache/setup_test.go
+++ b/tools/integration_tests/read_cache/setup_test.go
@@ -294,10 +294,10 @@ func TestMain(m *testing.M) {
 		cfg.ReadCache[0].Configs[15].Run = "TestRemountTest"
 
 		cfg.ReadCache[0].Configs[16].Flags = []string{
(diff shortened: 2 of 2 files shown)


tools/integration_tests/util/mounting/mounting.go
```diff
@@ -31,16 +32,21 @@ func MountGcsfuse(binaryFile string, flags []string) error {
...
+		fmt.Println("Could not open logfile: ", err.Error())
 	}
 	// Closing file at the end.
 	defer operations.CloseFile(file)
 
 	_, err = file.WriteString(mountCmd.String() + "\n")
 	if err != nil {
-		fmt.Println("Could not write cmd to logFile")
+		fmt.Println("Could not write cmd to logFile: ", err.Error())
 	}
```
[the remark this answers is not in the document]
ashmeenkaur: The code exists from the previous PR, not making this change as part of this PR as the PR is already pretty large.

tools/integration_tests/util/setup/setup.go
```diff
@@ -612,14 +612,15 @@ func AddCacheDirToFlags(flagSets [][]string, testname string) [][]string {
 
 // BuildFlagSets dynamically builds a list of flag sets based on bucket compatibility.
 // bucketType should be "flat", "hns", or "zonal".
-func BuildFlagSets(cfg test_suite.TestConfig, bucketType string) [][]string {
+func BuildFlagSets(cfg test_suite.TestConfig, bucketType string, run string) [][]string {
```
Tulsishah, line 618: Can you add comment what is run for?

tools/integration_tests/read_cache/cache_file_for_exclude_regex_test.go
```diff
@@ -37,19 +37,30 @@ type cacheFileForExcludeRegexTest struct {
 	flags         []string
 	storageClient *storage.Client
 	ctx           context.Context
+	baseTestName  string
 	suite.Suite
 }
 
+func (s *cacheFileForExcludeRegexTest) SetupSuite() {
+	setupLogFileAndCacheDir(s.baseTestName)
+	mountGCSFuseAndSetupTestDir(s.flags, s.ctx, s.storageClient)
```
Tulsishah, line 46: Don't we require to mount and unmount for every test?

Tulsishah (approved):

tools/integration_tests/read_cache/cache_file_for_exclude_regex_test.go
ashmeenkaur: No it is no longer required! Was required earlier because GKE was managing the log file. The tests have now been refactored so GCSFuse test suite manages that part.

PR #3872, opened 2025-10-09
feat(logging): Logging Improvement Part 1 - Add MountInstaceID logic in GCSFuse

meet2mky:
### Description
Adding MountInstaceID generation in GCSFuse for logging improvement. This MountInstaceID would be used in follow up PRs to add a custom attributes in logger and user agent.
b/450423708

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -138,12 +138,12 @@ func setupMountInstanceID() {
 		if mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey); !ok || mountInstanceID == "" {
 			Fatal("Could not retrieve %s env variable or it's empty.", GCSFuseMountInstanceIDEnvKey)
 		}
-	} else {
-		// If GCSFuse is not running in the background mode then generate a random UUID.
-		var err error
-		if mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength); err != nil {
-			Fatal("Could not generate MountInstanceID of length %d, err: %v", mountInstanceIDLength, err)
-		}
+		return
+	}
+	// If GCSFuse is not running in the background mode then generate a random UUID.
+	var err error
+	if mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength); err != nil {
+		Fatal("Could not generate MountInstanceID of length %d, err: %v", mountInstanceIDLength, err)
 	}
 }
 


internal/logger/logger.go
```diff
@@ -109,6 +118,48 @@ func init() {
...
+		if !ok || mountInstanceID == "" {
+			log.Printf("Could not retrieve %s env variable. Using default: %s", GCSFuseMountInstanceIDEnvKey, defaultMountInstanceID)
+			mountInstanceID = defaultMountInstanceID
+		}
+	} else {
+		// If GCSFuse is not running in the background mode then generate a random UUID.
+		var err error
+		mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength, uuid.NewRandom)
+		if err != nil {
+			log.Printf("Could not generate MountInstanceID, Using default: %s, err: %v", err, defaultMountInstanceID)
```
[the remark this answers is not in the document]
meet2mky: Done

internal/logger/logger.go
```diff
@@ -109,6 +118,48 @@ func init() {
...
+	uuid, err := uuidGenerator()
+	if err != nil {
+		return "", err
+	}
+	uuidStr := strings.ReplaceAll(uuid.String(), "-", "")
+	if size > len(uuidStr) {
+		return "", fmt.Errorf("UUID is smaller than requested size %d for MountInstanceID, UUID: %s", size, uuidStr)
+	}
+	return uuidStr[:size], nil
+}
```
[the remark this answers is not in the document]
meet2mky: Done!

internal/logger/logger.go
```diff
@@ -109,6 +118,48 @@ func init() {
...
+// background mode or generates one if running in foreground mode.
+func setupMountInstanceID() {
+	defaultMountInstanceID := strings.Repeat("0", mountInstanceIDLength)
+	if _, ok := os.LookupEnv(GCSFuseInBackgroundMode); ok {
+		// If GCSFuse is in background mode then look for the MountInstanceId in env which was set by the caller of demonize run.
+		mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey)
+		if !ok || mountInstanceID == "" {
+			log.Printf("Could not retrieve %s env variable. Using default: %s", GCSFuseMountInstanceIDEnvKey, defaultMountInstanceID)
+			mountInstanceID = defaultMountInstanceID
+		}
```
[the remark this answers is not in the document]
meet2mky: Not required: It's going to be same within the mount and won't be modified.

cmd/legacy_main_test.go
```diff
@@ -336,6 +337,9 @@ func (t *MainTest) TestForwardedEnvVars() {
 	}, {
 		inputEnvVars:             map[string]string{"GRPC_GO_LOG_VERBOSITY_LEVEL": "99", "GRPC_GO_LOG_SEVERITY_LEVEL": "INFO"},
 		expectedForwardedEnvVars: []string{"GRPC_GO_LOG_VERBOSITY_LEVEL=99", "GRPC_GO_LOG_SEVERITY_LEVEL=INFO"},
+	}, {
+		inputEnvVars:             map[string]string{"GCSFUSE_IN_BACKGROUND_MODE": "true"},
+		expectedForwardedEnvVars: []string{"GCSFUSE_IN_BACKGROUND_MODE=true", "GCSFUSE_MOUNT_INSTANCE_ID=" + logger.MountInstanceID()},
 	},
```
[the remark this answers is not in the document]
meet2mky: Incorrect understanding 👎🏻

internal/logger/logger.go
```diff
@@ -22,24 +22,32 @@ import (
...
+	mountInstanceIDLength = 8
 )
 
 var (
-	defaultLoggerFactory *loggerFactory
-	defaultLogger        *slog.Logger
+	defaultLoggerFactory     *loggerFactory
+	defaultLogger            *slog.Logger
+	mountInstanceID          string
+	setupMountInstanceIDOnce sync.Once
```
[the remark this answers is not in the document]
meet2mky: Not required.

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
 	defaultLogger = defaultLoggerFactory.newLogger(cfg.INFO)
 }
 
+// generateMountInstanceID generates a random string of size
+// from UUID returned from uuidGenerator.
+func generateMountInstanceID(size int, uuidGenerator func() (uuid.UUID, error)) (string, error) {
+	if size <= 0 {
```
gargnitingoogle, line 122: Also fail if size > 32 as UUID can't be more 32 chars.

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
...
+
+// setupMountInstanceID handles the retrieval of mountInstanceId if GCSFuse is in
+// background mode or generates one if running in foreground mode.
+func setupMountInstanceID() {
+	defaultMountInstanceID := strings.Repeat("0", mountInstanceIDLength)
+	if _, ok := os.LookupEnv(GCSFuseInBackgroundMode); ok {
+		// If GCSFuse is in background mode then look for the MountInstanceId in env which was set by the caller of demonize run.
+		mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey)
+		if !ok || mountInstanceID == "" {
+			Warnf("Could not retrieve %s env variable. Using default: %s", GCSFuseMountInstanceIDEnvKey, defaultMountInstanceID)
```
gargnitingoogle: This should simply be an error/panic, not a warning. This shouldn't be allowed because we are explicitly passing the GCSFuseMountInstanceIDEnvKey env var in background daemon process.

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
...
+			mountInstanceID = defaultMountInstanceID
+		}
+	} else {
+		// If GCSFuse is not running in the background mode then generate a random UUID.
+		var err error
+		mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength, uuid.NewRandom)
+		if err != nil {
+			Warnf("Could not generate MountInstanceID, Using default: %s, err: %v", defaultMountInstanceID, err)
```
gargnitingoogle: This should be an error/panic, not a warning.

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
...
+		// If GCSFuse is in background mode then look for the MountInstanceId in env which was set by the caller of demonize run.
+		mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey)
+		if !ok || mountInstanceID == "" {
+			Warnf("Could not retrieve %s env variable. Using default: %s", GCSFuseMountInstanceIDEnvKey, defaultMountInstanceID)
+			mountInstanceID = defaultMountInstanceID
+		}
+	} else {
+		// If GCSFuse is not running in the background mode then generate a random UUID.
+		var err error
+		mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength, uuid.NewRandom)
```
gargnitingoogle: What is stopping this from getting a new mountInstanceID being generated again even if it was assigned already by a previous call. Please add a check here to see that mountInstanceID is defaultMountInstanceID.

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
...
+	if size > len(uuidStr) {
+		return "", fmt.Errorf("UUID is smaller than requested size %d for MountInstanceID, UUID: %s", size, uuidStr)
+	}
+	return uuidStr[:size], nil
+}
+
+// setupMountInstanceID handles the retrieval of mountInstanceId if GCSFuse is in
+// background mode or generates one if running in foreground mode.
+func setupMountInstanceID() {
+	defaultMountInstanceID := strings.Repeat("0", mountInstanceIDLength)
```
gargnitingoogle: Move this into vars or consts and initialize mountInstanceID with it.

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
...
+		var err error
+		mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength, uuid.NewRandom)
+		if err != nil {
+			Warnf("Could not generate MountInstanceID, Using default: %s, err: %v", defaultMountInstanceID, err)
+			mountInstanceID = defaultMountInstanceID
+		}
+	}
+}
+
+func MountInstanceID() string {
```
gargnitingoogle, line 155:
```suggestion
func AssignMountInstanceID() string {
```

internal/logger/logger_test.go
```diff
@@ -339,3 +344,134 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+			name: "ThirtyTwoChars",
+			size: 32,
+		},
+	}
+	for _, tc := range testCases {
+		t.T().Run(tc.name, func(t *testing.T) {
+			mountInstanceID, err := generateMountInstanceID(tc.size, uuid.NewRandom)
+
+			require.NoError(t, err)
+			assert.Regexp(t, fmt.Sprintf("^[0-9a-f]{%d}$", tc.size), mountInstanceID)
```
gargnitingoogle: Instead of generating regex pattern using `fmt.Sprintf`, define it as `expectedMountInstanceIDRegex string` in the testcase structure. That'll avoid unnecessary use of fmt.Sprintf and also be more readable.

internal/logger/logger_test.go
```diff
@@ -339,3 +344,134 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+		mountInstanceIDEnv string
+		expectedID         string
+	}{
+		{
+			name:             "ForegroundMode",
+			inBackgroundMode: false,
+			expectedID:       "",
+		},
+		{
+			name:               "BackgroundModeWithInstanceID",
```
gargnitingoogle, line 403:
```suggestion
			name:               "BackgroundMode",
```

internal/logger/logger_test.go
gargnitingoogle: Same goes in the other tests which have this same problem.

internal/logger/logger.go
meet2mky: Please see line 130 to 133.

internal/logger/logger.go
meet2mky: Can you tell why do you think this can be const or should be moved into vars?

internal/logger/logger.go
meet2mky: I think, the main question here is: What do we do if we couldn't get the `MountInstaceID` if it was running in whatever mode background or foreground.

There are two ways to think about it.
A) Fail the mount -- This makes a part of logging feature which is having custom attribute on the logger a critical path during the mount.

B) Let the mount continue -- This ensures whatever the reason may be for error (error will be logged of course) and we couldn't get the `MountInstanceID` so we may not have a custom attribute with Random MountInstanceID on the logger attribute but we continue the mount to succeed.

I think option B, is better as this separates logging failure from actual mount operations. If you think this is something non negotiable and we should always fail the mount on logging error?

@kislaykishore @vadlakondaswetha Wdyt?

internal/logger/logger.go
meet2mky: See line 161, setupMountInstanceIDOnce.Do(setupMountInstanceID). This method is run with Sync.Once.

You can read about it here: https://pkg.go.dev/sync#Once

internal/logger/logger.go
meet2mky: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3872/files/db10835121e137b4517b191db93749367c742ee1#r2418502449

internal/logger/logger.go
meet2mky: This is a getter. That's why it's named as per variable.

internal/logger/logger_test.go
meet2mky: Can you tell your reasoning?

internal/logger/logger.go
gargnitingoogle: Because it's a single fixed value. No value in creating it again and again.

meet2mky: * It's a function local variable (used in this method nowhere else). Using at the package level wastes the memory as it's not going to be out of scope ever.
* It's not created again and again because it's run only once when MountInstanceID is called.

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
...
+// background mode or generates one if running in foreground mode.
+func setupMountInstanceID() {
+	defaultMountInstanceID := strings.Repeat("0", mountInstanceIDLength)
+	if _, ok := os.LookupEnv(GCSFuseInBackgroundMode); ok {
+		// If GCSFuse is in background mode then look for the MountInstanceId in env which was set by the caller of demonize run.
+		mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey)
+		if !ok || mountInstanceID == "" {
+			Warnf("Could not retrieve %s env variable. Using default: %s", GCSFuseMountInstanceIDEnvKey, defaultMountInstanceID)
+			mountInstanceID = defaultMountInstanceID
+		}
```
kislaykishore:
```suggestion
		if mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey); !ok {
			Warnf("Could not retrieve %s env variable. Using default: %s", GCSFuseMountInstanceIDEnvKey, defaultMountInstanceID)
			mountInstanceID = defaultMountInstanceID
		}
```

internal/logger/logger_test.go
gargnitingoogle: Simplification. This has no other tests like BackgroundModeWithoutInstanceID

internal/logger/logger.go
```diff
@@ -109,6 +117,51 @@ func init() {
...
+	} else {
+		// If GCSFuse is not running in the background mode then generate a random UUID.
+		var err error
+		mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength, uuid.NewRandom)
+		if err != nil {
+			Warnf("Could not generate MountInstanceID, Using default: %s, err: %v", defaultMountInstanceID, err)
+			mountInstanceID = defaultMountInstanceID
+		}
+	}
```
kislaykishore:
```suggestion
    return
	}
		// If GCSFuse is not running in the background mode then generate a random UUID.
		var err error
		if mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength, uuid.NewRandom); err != nil {
			Warnf("Could not generate MountInstanceID, Using default: %s, err: %v", defaultMountInstanceID, err)
			mountInstanceID = defaultMountInstanceID
	  }
```

internal/logger/logger_test.go
```diff
@@ -339,3 +344,134 @@ func (t *LoggerTest) TestSetLogFormatToText() {
...
+	for _, tc := range testCases {
+		t.T().Run(tc.name, func(t *testing.T) {
+			t.Cleanup(func() {
+				mountInstanceID = ""
+				setupMountInstanceIDOnce = sync.Once{}
+			})
+			if tc.inBackgroundMode {
+				t.Setenv(GCSFuseInBackgroundMode, "true")
+				t.Setenv(GCSFuseMountInstanceIDEnvKey, tc.mountInstanceIDEnv)
+			}
```
kislaykishore:
```suggestion
			t.Setenv(GCSFuseInBackgroundMode, "true")
			t.Setenv(GCSFuseMountInstanceIDEnvKey, tc.mountInstanceIDEnv)
```

internal/logger/logger.go
meet2mky: Done!

internal/logger/logger.go
meet2mky: Done!

internal/logger/logger_test.go
meet2mky: Done!

internal/logger/logger_test.go
meet2mky: So other tests have this type of Names. SetupMountInstanceID behaviour depends on the Background mode and if MountInstaceID env is set or not. So the test is named accordingly.

internal/logger/logger_test.go
meet2mky: Done!

meet2mky: Thanks, I noticed there was logic error in the test. Fixed that as well.

internal/logger/logger.go
gargnitingoogle: > What do we do if we couldn't get the MountInstaceID if it was running in whatever mode background or foreground.

Foreground mode is a different question. But in the background mode, you are explicitly forwarding the env var GCSFuseMountInstanceIDEnvKey down to the daemon process, in this same PR. Why is it then okay, if the daemon process fails to find it in lookup?

meet2mky: Discussed offline, going with the Fatal logs.

internal/logger/logger.go
meet2mky: Removed entirely as we don't need this now.

internal/logger/logger.go
```diff
@@ -109,6 +117,42 @@ func init() {
...
+	} else {
+		// If GCSFuse is not running in the background mode then generate a random UUID.
+		var err error
+		if mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength); err != nil {
+			Fatal("Could not generate MountInstanceID of length %d, err: %v", mountInstanceIDLength, err)
+		}
+	}
+}
+
+func MountInstanceID() string {
```
gargnitingoogle, line 155: nit: Add a function header as is the norm for exposed functions.

meet2mky: Done!

internal/logger/logger.go
```diff
@@ -109,6 +117,46 @@ func init() {
...
+		if mountInstanceID, ok = os.LookupEnv(GCSFuseMountInstanceIDEnvKey); !ok || mountInstanceID == "" {
+			Fatal("Could not retrieve %s env variable or it's empty.", GCSFuseMountInstanceIDEnvKey)
+		}
+	} else {
+		// If GCSFuse is not running in the background mode then generate a random UUID.
+		var err error
+		if mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength); err != nil {
+			Fatal("Could not generate MountInstanceID of length %d, err: %v", mountInstanceIDLength, err)
+		}
+	}
```
kislaykishore, line 147:
```suggestion
		}
		return
	} 
		// If GCSFuse is not running in the background mode then generate a random UUID.
		var err error
		if mountInstanceID, err = generateMountInstanceID(mountInstanceIDLength); err != nil {
			Fatal("Could not generate MountInstanceID of length %d, err: %v", mountInstanceIDLength, err)
		}
```

meet2mky: Done!

PR #3878, opened 2025-10-13
refactor(read-type-heuristic): Implement read-type classifier

raj-prince:
### Description
This change isolates the read-type-heuristic logic as a separate entity named `ReadTypeClassifier`.
- Unit test for `IsSeekNeeded`, `GetReadInfo` has been copied directly from the gcs_reader_test.go
- Did few minor refactoring like, `avgReadBytes` to keep the zero seek check at one place.
- Overall integration of read-type-classifier and removal of duplicate heuristic logic in gcs-reader will be done in the next next PR.

b/451434114

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/read_type_classifier.go
+++ b/internal/gcsx/read_type_classifier.go
@@ -151,6 +151,8 @@ func (rtc *ReadTypeClassifier) GetReadInfo(offset int64, seekRecorded bool) Read
 // If the read pattern is classified as random, it calculates an appropriate
 // read size based on the average read size per seek, bounded by min and max read sizes.
 // If the read pattern is sequential, it returns the configured sequential read size.
+// Note: The returned prefetch window size is not limited by the object size, caller should
+// handle that separately.
 func (rtc *ReadTypeClassifier) ComputeSeqPrefetchWindowAndAdjustType() int64 {
 	if seeks := rtc.seeks.Load(); seeks >= minSeeksForRandom {
 		averageReadBytes := avgReadBytes(rtc.totalReadBytes.Load(), seeks)


internal/gcsx/read_type_classifier_test.go
```diff
@@ -0,0 +1,603 @@
...
+		initialTotalReadBytes uint64
+		sequentialReadSizeMb  int64
+		expectedSeqIO         int64
+	}{
+		{
+			name:                  "Sequential Read, No seek",
+			initialNumSeeks:       0,
+			initialTotalReadBytes: 0,
+			sequentialReadSizeMb:  22,
+			expectedSeqIO:         22 * MiB,
```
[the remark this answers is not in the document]
raj-prince: MiB is defined in random_reader.go, but changing to MB for coherency.

internal/gcsx/read_type_classifier.go
```diff
@@ -0,0 +1,187 @@
...
+		SeekRecorded:   seekRecorded,
+	}
+}
+
+// SeqReadIO returns the read size to be used for the next read operation based on
+// the current read pattern. It also updates the readType if needed.
+// If the read pattern is classified as random, it calculates an appropriate
+// read size based on the average read size per seek, bounded by min and max read sizes.
+// If the read pattern is sequential, it returns the configured sequential read size.
+func (rpt *ReadTypeClassifier) SeqReadIO() int64 {
```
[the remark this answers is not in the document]
raj-prince: Renamed to `omputeSeqReadIOAndAdjustType`

internal/gcsx/read_type_classifier.go
```diff
@@ -0,0 +1,187 @@
...
+		averageReadBytes := avgReadBytes(rtc.totalReadBytes.Load(), seeks)
+		if averageReadBytes < maxReadSize {
+			randomReadSize := int64(((averageReadBytes / MB) + 1) * MB)
+			if randomReadSize < minReadSize {
+				randomReadSize = minReadSize
+			}
+			if randomReadSize > maxReadSize {
+				randomReadSize = maxReadSize
+			}
+			rtc.readType.Store(metrics.ReadTypeRandom)
```
abhishek10004, line 163: The size returned from here has to be bound by object size for both seq as well as random read pattern.

internal/gcsx/read_type_classifier.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// RecordSeek checks if the read at the given offset is a seek and updates the internal state accordingly.
+// Call it before starting the read operation.
+func (rtc *ReadTypeClassifier) RecordSeek(offset int64) {
+	rtc.GetReadInfo(offset, false)
+}
+
+// RecordRead records a read operation of the given size at the given offset.
+// This must be called before the read operation.
```
abhishek10004: would this be called before or after the read operation?

internal/gcsx/read_type_classifier.go
```diff
@@ -0,0 +1,187 @@
...
+// ComputeSeqReadIOAndAdjustType computes the sequential IO size heuristically based on
+// the current read pattern. It also updates the readType if needed.
+// If the read pattern is classified as random, it calculates an appropriate
+// read size based on the average read size per seek, bounded by min and max read sizes.
+// If the read pattern is sequential, it returns the configured sequential read size.
+func (rtc *ReadTypeClassifier) ComputeSeqReadIOAndAdjustType() int64 {
+	if seeks := rtc.seeks.Load(); seeks >= minSeeksForRandom {
+		averageReadBytes := avgReadBytes(rtc.totalReadBytes.Load(), seeks)
+		if averageReadBytes < maxReadSize {
+			randomReadSize := int64(((averageReadBytes / MB) + 1) * MB)
```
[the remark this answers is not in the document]
abhishek10004: I know what we have in the PR is same as what we have currently but IMO it makes sense to update it to ensure we round up correctly.

internal/gcsx/read_type_classifier.go
raj-prince: Yeah, after the read. Thanks!

internal/gcsx/read_type_classifier.go
raj-prince: I was thinking to keep this independent of object size, and only predicts the SeqPrefetchWindow, and client can make sure object size limit and all. WDUT?

To support this we need to start keeping object information, which I am not keeping right now.

abhishek10004: Okay. Should we add this detail at the top comment to make it explicit for the users of this function that this does not bound the range to object size?

raj-prince: Sure.

abhishek10004 (approved):
LGTM

PR #3883, opened 2025-10-13
refactor: readdirplus test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate readdirplus test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of readdirplus package so it can use config file.

[b/445951883](https://buganizer.corp.google.com/issues/445951883)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/test_config.yaml
+++ b/tools/integration_tests/test_config.yaml
@@ -284,3 +284,24 @@ read_cache:
           hns: true
           zonal: true
         run: TestCacheFileForRangeReadTrueTest
+
+readdirplus:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "pranjal-bucket-1"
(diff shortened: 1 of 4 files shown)


tools/integration_tests/test_config.yaml
```diff
@@ -283,4 +283,25 @@ read_cache:
           flat: true
           hns: true
           zonal: true
-        run: TestCacheFileForRangeReadTrueTest
\ No newline at end of file
+        run: TestCacheFileForRangeReadTrueTest
+
+readdirplus:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "pranjal-bucket-1"
```
ashmeenkaur, line 290: Nit: hardcoded bucket

PR #3889, opened 2025-10-14
feat(logging): Logging Improvement Part 2 - Add Helper method with unit tests and refactor complex forwardEnv test.

meet2mky:
### Description
Add `fsName` helper method with unit tests and refactor unit tests for easier understanding.

Note: This method would be used in later PR(s) for logging improvements.

b/451496620

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Part of Presubmit tests.

### DIFF
--- a/cmd/legacy_main_test.go
+++ b/cmd/legacy_main_test.go
@@ -420,24 +420,29 @@ func (t *MainTest) TestForwardedEnvVars_Precedence() {
 func (t *MainTest) TestForwardedEnvVars_PassedWhenSet() {
 	// These variables are only forwarded if they are set in the environment.
 	testCases := []struct {
-		name         string
-		inputEnvVars map[string]string
+		name                     string
+		inputEnvVars             map[string]string
+		expectedForwardedEnvVars []string
 	}{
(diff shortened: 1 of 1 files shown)


cmd/legacy_main_test.go
```diff
@@ -293,70 +294,188 @@ func (t *MainTest) TestCallListRecursiveOnNonExistingDirectory() {
...
+			inputEnvVars: map[string]string{"no_proxy": "no-proxy-123"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.T().Run(tc.name, func(t *testing.T) {
+			for k, v := range tc.inputEnvVars {
+				t.Setenv(k, v)
+			}
+			var expectedForwardedEnvVars []string
```
raj-prince: nit: will be more readable to keep the expected list of string also as a member in the test-case struct. Instead of creating on the fly. Also, we should try to minimize the logic in the test.

meet2mky: Thanks, Done!

PR #3891, opened 2025-10-14
refactor: readdirplus test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate readdirplus test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of readdirplus package so it can use config file.

[b/445951883](https://buganizer.corp.google.com/issues/445951883)

### Testing details
2. Unit tests - NA


tools/integration_tests/test_config.yaml
```diff
@@ -423,6 +423,27 @@ read_cache:
         run: TestCacheFileForIncludeRegexTest
         run_on_gke: true
 
+readdirplus:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "${BUCKET_NAME}"
+    log_file: # Not Required by readdirplus tests
+    run_on_gke: false
```
ashmeenkaur: this config has moved to configs section. I see we misconfigured it in some other packages as well. can you fix it? I also don't see a reason why these tests can't run in GKE environment.

tools/integration_tests/readdirplus/setup_test.go
```diff
@@ -108,42 +115,69 @@ func validateLogsForReaddirplus(t *testing.T, logFile string, dentryCacheEnabled
 	}
 }
 
-func mountGCSFuseAndSetupTestDir(flags []string, testDirName string) {
-	setup.MountGCSFuseWithGivenMountFunc(flags, mountFunc)
-	setup.SetMntDir(mountDir)
-	testDirPath = setup.SetupTestDirectory(testDirName)
+func mountGCSFuseAndSetupTestDir(flags []string, ctx context.Context, storageClient *storage.Client) {
+	setup.MountGCSFuseWithGivenMountWithConfigFunc(testEnv.cfg, flags, mountFunc)
+	setup.SetMntDir(testEnv.cfg.GCSFuseMountedDirectory)
```
ashmeenkaur: This should be GCSFuse or GKE mounted directory based on how tests are running. Can you check if this is working in both cases?

tools/integration_tests/readdirplus/readdirplus_without_dentry_cache_test.go
```diff
@@ -29,30 +32,39 @@ import (
...
-func (s *readdirplusWithoutDentryCacheTest) TestReaddirplusWithoutDentryCache() {
-	// Create directory structure
-	// testBucket/target_dir/                                                       -- Dir
-	// testBucket/target_dir/file		                                            -- File
-	// testBucket/target_dir/emptySubDirectory                                      -- Dir
-	// testBucket/target_dir/subDirectory                                           -- Dir
-	// testBucket/target_dir/subDirectory/file1                                     -- File
-	targetDir := path.Join(testDirPath, targetDirName)
+func (s *ReaddirplusWithoutDentryCacheTest) SetupSuite() {
+	// setupLogFileAndCacheDir(s.baseTestName)
```
ashmeenkaur:
```suggestion
```

same for other SetupSuite functions of this package.

tools/integration_tests/readdirplus/setup_test.go
PranjalC100: checked both are working!

tools/integration_tests/test_config.yaml
```diff
@@ -423,6 +423,28 @@ read_cache:
         run: TestCacheFileForIncludeRegexTest
         run_on_gke: true
 
+readdirplus:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "pranjal-bucket-1"
+    log_file: # added
```
```suggestion
    test_bucket: "${BUCKET_NAME}"
```

ashmeenkaur (commented):
LGTM! Please make sure that https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/run_tests_mounted_directory.sh#L662 commands still work. After testing, please add this to testing details in the PR description.

PranjalC100:
> LGTM! Please make sure that https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/run_tests_mounted_directory.sh#L662 commands still work. After testing, please add this to testing details in the PR description.

yes both are passing!

ashmeenkaur (approved):

PR #3894, opened 2025-10-15
feat(logging): Logging Improvement Part 3 - Refactor and simplify complex logger tests.

meet2mky:
### Description
Refactor logger tests to ensure they are readable and easy to develop.

b/451497800

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Not Required only unit tests are changed.

### DIFF
--- a/internal/logger/logger_test.go
+++ b/internal/logger/logger_test.go
@@ -76,14 +76,6 @@ func getTestLoggingFunctions() []func() {
 	}
 }
 
-func validateOutput(t *testing.T, expectedLogLineRegexes, actualLogLines []string) {
-	t.Helper()
-	require.Equal(t, len(expectedLogLineRegexes), len(actualLogLines))
-	for i := range actualLogLines {
-		assert.Regexp(t, expectedLogLineRegexes[i], actualLogLines[i])
-	}
-}
-
 // fetchAllLogLevelOutputsForSpecifiedSeverityLevel sets the log format and severity,
 // executes standard logging functions, and returns their output.
 func fetchAllLogLevelOutputsForSpecifiedSeverityLevel(t *testing.T, format string, level string) []string {
@@ -108,7 +100,10 @@ func fetchAllLogLevelOutputsForSpecifiedSeverityLevel(t *testing.T, format strin
 // validateLogOutputs compares the captured log line output with the expected log regex patterns.
 func validateLogOutputs(t *testing.T, expectedLogLineRegexes, actualLogLines []string) {
 	t.Helper()
-	validateOutput(t, expectedLogLineRegexes, actualLogLines)
+	require.Equal(t, len(expectedLogLineRegexes), len(actualLogLines))
+	for i := range actualLogLines {
+		assert.Regexp(t, expectedLogLineRegexes[i], actualLogLines[i])
+	}
 }
 
 // //////////////////////////////////////////////////////////////////////


internal/logger/logger_test.go
```diff
@@ -16,44 +16,37 @@ package logger
...
-	textErrorString   = "^time=\"[a-zA-Z0-9/:. ]{26}\" severity=ERROR message=\"TestLogs: www.errorExample.com\""
-
-	jsonTraceString   = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"TRACE\",\"message\":\"TestLogs: www.traceExample.com\"}"
-	jsonDebugString   = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"DEBUG\",\"message\":\"TestLogs: www.debugExample.com\"}"
-	jsonInfoString    = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"INFO\",\"message\":\"TestLogs: www.infoExample.com\"}"
-	jsonWarningString = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"WARNING\",\"message\":\"TestLogs: www.warningExample.com\"}"
-	jsonErrorString   = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"ERROR\",\"message\":\"TestLogs: www.errorExample.com\"}"
+	textLogPattern = `^time="[a-zA-Z0-9/:. ]{26}" severity=%s message="TestLogs: %s"\s*$`
+	jsonLogPattern = `^{\"timestamp\":{\"seconds\":\d{10},\"nanos\":\d{0,9}},\"severity\":\"%s\",\"message\":\"TestLogs: %s\"}\s*$`
```
gargnitingoogle: Do we need this and the other backslashes before the quotes in `jsonLogPattern` given that it's already defined by closing with backticks (`` ` ``) which escape any enclosed quotes in them. This will literally look for `\"` in the compared log message (i.e. including the backslashes). If yes, then this mismatches what we had on the LHS in the value of jsonDebugString . I mean LHS had `" \" "` which would match `"` in the log, while the RHS has `` `\"` `` which will match `\"` in the log.

internal/logger/logger_test.go
```diff
@@ -16,44 +16,37 @@ package logger
...
-	suite.Suite
-}
-
-func TestLoggerSuite(t *testing.T) {
-	suite.Run(t, new(LoggerTest))
-}
-
 // //////////////////////////////////////////////////////////////////////
-// Boilerplate
+// Helper
```
gargnitingoogle: nit:

```suggestion
// Helpers
```

internal/logger/logger_test.go
```diff
@@ -16,44 +16,37 @@ package logger
...
 // //////////////////////////////////////////////////////////////////////
 
+func expectedLogRegex(format, severity, message string) string {
+	switch format {
```
gargnitingoogle, line 41:
```suggestion
    t.Helper()
	switch format {
```

internal/logger/logger_test.go
```diff
@@ -101,17 +94,18 @@ func getTestLoggingFunctions() []func() {
 }
 
 func validateOutput(t *testing.T, expected []string, output []string) {
+	t.Helper()
 	for i := range output {
```
gargnitingoogle: I know that it's before this change, but what's guaranteeing that output and expected arrays have the same size ? Shouldn't we check that too to avoid panic'ing with array bounds check error without properly registering with t ?

internal/logger/logger_test.go
```diff
@@ -16,44 +16,37 @@ package logger
...
-	jsonErrorString   = "^{\"timestamp\":{\"seconds\":\\d{10},\"nanos\":\\d{0,9}},\"severity\":\"ERROR\",\"message\":\"TestLogs: www.errorExample.com\"}"
+	textLogPattern = `^time="[a-zA-Z0-9/:. ]{26}" severity=%s message="TestLogs: %s"\s*$`
+	jsonLogPattern = `^{\"timestamp\":{\"seconds\":\d{10},\"nanos\":\d{0,9}},\"severity\":\"%s\",\"message\":\"TestLogs: %s\"}\s*$`
 )
 
-type LoggerTest struct {
-	suite.Suite
-}
-
-func TestLoggerSuite(t *testing.T) {
```
gargnitingoogle, line 49: Removing the suite and this single top-level test function `TestLoggerSuite` will make it harder to filter out and run all the logging tests at once. Earlier we could have just passed `-test.run="TestLoggerSuite"` to the go test command to run all logging tests at once, but now we'll have do something like `-test.run='"^(TestTextFormatLogs_.*|TestJSONFormatLogs_.*...)$"` which is much harder to figure out and run on the command line if needed . Why are we making this structural change at all ?

internal/logger/logger_test.go
```diff
@@ -101,17 +94,18 @@ func getTestLoggingFunctions() []func() {
 }
 
 func validateOutput(t *testing.T, expected []string, output []string) {
```
gargnitingoogle:
```suggestion
func validateOutput(t *testing.T, expected, output []string) {
```

internal/logger/logger_test.go
```diff
@@ -101,17 +94,18 @@ func getTestLoggingFunctions() []func() {
...
+	t.Helper()
 	for i := range output {
 		if expected[i] == "" {
 			assert.Equal(t, expected[i], output[i])
```
gargnitingoogle: https://github.com/stretchr/testify/blob/feb1324bc3d000fed7b21dfe20bec72ecca27502/assert/assertions.go#L792

```suggestion
			assert.Empty(t, output[i])
```

gargnitingoogle (changes requested):
Review WIP.

internal/logger/logger_test.go
meet2mky: Suite here is redundant as we aren't using any suite feature like setup or teardown. It makes the tests more simple on removing it. Did this as part of suggestion on original PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3818#discussion_r2405403483

I agree that running tests from command line would be difficult with filters like this.

internal/logger/logger_test.go
meet2mky: Good catch. We don't need the `\` as it's inside `` ` `` ticks. Thank you!

gargnitingoogle (changes requested):
Review complete. One major comment. A few minor comments, and 1-2 readability comments.

internal/logger/logger_test.go
gargnitingoogle: > I agree that running tests from command line would be difficult with filters like this.

I often run things from command line as that's just faster than IDE. 

@kislaykishore can we revisit the decision to make this change ? It's making things worse IMO for no good reason.

kislaykishore: Independent tests are the default in Golang. It also reduces cognitive load.
While there is no easy way to run tests in a single file in Golang, you could either run all the tests in the package or name the tests such that they share a prefix.

internal/logger/logger_test.go
```diff
@@ -123,157 +117,201 @@ func validateLogOutputAtSpecifiedFormatAndSeverity(t *testing.T, format string,
...
 // //////////////////////////////////////////////////////////////////////
-func (t *LoggerTest) TestTextFormatLogs_LogLevelOFF() {
+func TestTextFormatLogs_LogLevelOFF(t *testing.T) {
 	var expected = []string{
 		"", "", "", "", "",
 	}
 
 	// Assert that nothing is logged when log level is OFF.
-	validateLogOutputAtSpecifiedFormatAndSeverity(t.T(), "json", cfg.OFF, expected)
+	validateLogOutputAtSpecifiedFormatAndSeverity(t, "json", cfg.OFF, expected)
```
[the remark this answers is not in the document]
kislaykishore: +1

internal/logger/logger_test.go
```diff
@@ -123,157 +117,201 @@ func validateLogOutputAtSpecifiedFormatAndSeverity(t *testing.T, format string,
...
+		"", // TRACE
+		"", // DEBUG
+		"", // INFO
+		"", // WARNING
+		"", // ERROR
 	}
 
 	// Assert that nothing is logged when log level is OFF.
-	validateLogOutputAtSpecifiedFormatAndSeverity(t.T(), "json", cfg.OFF, expected)
+	validateLogOutputAtSpecifiedFormatAndSeverity(t, "json", cfg.OFF, expected)
```
kislaykishore: These tests are not readable since the information about the Arrange, Act and Asserts blocks are all concealed in a function. I'd recommend we add them explicitly here even if that means duplicating the logic. [That's fine for tests](https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html).

internal/logger/logger_test.go
```diff
@@ -288,61 +326,65 @@ func (t *LoggerTest) TestInitLogFile() {
...
+			// Initialize logger factory for each subtest to ensure isolation.
+			logConfig := cfg.DefaultLoggingConfig()
+			defaultLoggerFactory = &loggerFactory{
+				file:      nil,
+				level:     string(logConfig.Severity), // setting log level to INFO by default
+				logRotate: logConfig.LogRotate,
+			}
+
+			SetLogFormat(tc.format)
+
```
kislaykishore: Remove this empty line

internal/logger/logger_test.go
meet2mky: Thinking more about this - 
1. Using defaults in golang simplifies test but running tests for a file from CMD increases cognitive load but i think usually we test a single method and test for that method are well prefixed with the method name (In this case where all tests of this package are within the same file so running package or file tests is same.)
2. Using suites gives the flexibility only to run all the suite tests at once. This may help if within a same package tests are required to organised within multiple suites or they have need of suite for common steup/teardown.

internal/logger/logger_test.go
meet2mky: Done!

internal/logger/logger_test.go
meet2mky: Done!

internal/logger/logger_test.go
meet2mky: Added a check to validate that size are same before proceeding to regex matches.

internal/logger/logger_test.go
meet2mky: There is no benefit of using suites in this package both the functionality wise or usage wise (I.e. Running tests using CMD). Removing this.

internal/logger/logger_test.go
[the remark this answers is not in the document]
meet2mky: Done!

internal/logger/logger_test.go
meet2mky: Removed one level of indirection and also split the Act and Assert part to be more intuitive and changes few variable names to show exact usage.

internal/logger/logger_test.go
meet2mky: Done!

internal/logger/logger_test.go
```diff
@@ -288,61 +326,65 @@ func (t *LoggerTest) TestInitLogFile() {
...
-	assert.Equal(t.T(), filePath, defaultLoggerFactory.file.Name())
-	assert.Nil(t.T(), defaultLoggerFactory.sysWriter)
-	assert.Equal(t.T(), format, defaultLoggerFactory.format)
-	assert.Equal(t.T(), cfg.DEBUG, defaultLoggerFactory.level)
-	assert.Equal(t.T(), fileSize, defaultLoggerFactory.logRotate.MaxFileSizeMb)
-	assert.Equal(t.T(), backupFileCount, defaultLoggerFactory.logRotate.BackupFileCount)
-	assert.True(t.T(), defaultLoggerFactory.logRotate.Compress)
+	t.Cleanup(func() {
+		defaultLoggerFactory.file.Close() // Close file handle to release resources.
+	})
```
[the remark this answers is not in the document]
meet2mky: Added Nil check in test expectation.

kislaykishore (approved):

PR #3896, opened 2025-10-15
refactor: only dir mounting [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes temporary solution/changes in config setup so that GKE can also run only dir mount tests using an environment variable.

https://buganizer.corp.google.com/issues/454584177

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/test_config.yaml
+++ b/tools/integration_tests/test_config.yaml
@@ -157,7 +157,6 @@ rename_dir_limit:
 local_file:
   - mounted_directory: "${MOUNTED_DIR}"
     test_bucket: "${BUCKET_NAME}"
-    only_dir: "${ONLY_DIR}"
     log_file: # Optional
     configs:
       - flags:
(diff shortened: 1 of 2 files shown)


tools/integration_tests/test_config.yaml
```diff
@@ -129,12 +128,13 @@ readonly:
           flat: true
           hns: true
           zonal: true
-        run_on_gke: true
 
 rename_dir_limit:
   - mounted_directory: "${MOUNTED_DIR}"
     test_bucket: "${BUCKET_NAME}"
+    only_dir: "${ONLY_DIR}"
```
ashmeenkaur, line 137: is this the only package running only dir tests?

ashmeenkaur (dismissed):
LGTM, could you please check if run_test_mounted_directory script and KOKORO tests are woking fine before merging?

tools/integration_tests/test_config.yaml
PranjalC100: rebased and added the same for read cache and local file configs

ashmeenkaur (dismissed):

ashmeenkaur (approved):

PR #3898, opened 2025-10-16
refactor: log rotation test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate log rotation test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of log rotation package so it can use config file.

[b/445952888](https://buganizer.corp.google.com/issues/445952888)

### Testing details
2. Unit tests - NA


tools/integration_tests/run_gke_tests.sh
PranjalC100, line 1: Still working on this script. 
It mounts the flags, chooses the tests it needs to run on according to the flags, unmounts it.

ashmeenkaur: can we add it in a separate PR

tools/integration_tests/test_config.yaml
```diff
@@ -519,3 +519,17 @@ interrupt:
           hns: true
           zonal: true
         run_on_gke: true
+
+log_rotation:
+  - mounted_directory: "${MOUNTED_DIR}"
+    test_bucket: "${BUCKET_NAME}"
+    log_file: /tmp/gcsfuse-tmp/LogRotationTest.log
```
ashmeenkaur: no longer required. In fact we decided to clean up this flag from the config file.

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -21,54 +21,73 @@ import (
...
+		log.Println("No configuration found for log rotation tests in config. Using flags instead.")
+		// Populate the config manually.
+		config.LogRotation = make([]test_suite.TestConfig, 1)
+		config.LogRotation[0].TestBucket = setup.TestBucket()
+		config.LogRotation[0].GKEMountedDirectory = setup.MountedDirectory()
+		config.LogRotation[0].LogFile = setup.LogFile()
+		config.LogRotation[0].Configs = make([]test_suite.ConfigItem, 1)
+		config.LogRotation[0].Configs[0].Flags = []string{
+			"--log-severity=TRACE --log-file=/gcsfuse-tmp/LogRotationTest.log --log-rotate-max-file-size-mb=2 --log-rotate-backup-file-count=2 --log-rotate-compress=false",
+			"--log-severity=TRACE --log-file=/gcsfuse-tmp/LogRotationTest.log --log-rotate-max-file-size-mb=2 --log-rotate-backup-file-count=2 --log-rotate-compress=true",
```
ashmeenkaur: Nit: can we use the exact same order as used in config file so it is easier to do string matching

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -77,42 +96,38 @@ func TestMain(m *testing.M) {
 		}
 	}()
 
-	setup.ExitWithFailureIfBothTestBucketAndMountedDirectoryFlagsAreNotSet()
-
-	// Run tests for mountedDirectory only if --mountedDirectory flag is set.
+	setupLogFileDir(testDirName)
```
ashmeenkaur: Nit: Not required

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -47,25 +47,28 @@ func runOperationsOnFileTillLogRotation(t *testing.T, wg *sync.WaitGroup, fileNa
 	}
 	// Setup file with 5 MiB content in test directory.
 	testDirPath := path.Join(setup.MntDir(), testDirName)
+	setupLogFileDir(testDirName)
```
ashmeenkaur: Nit: Not required

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -21,54 +21,73 @@ import (
...
-		"logging": map[string]any{
-			"severity":  "TRACE",
-			"file-path": logFilePath,
-			"log-rotate": map[string]any{
-				"max-file-size-mb":  maxFileSizeMB,
-				"backup-file-count": backupFileCount,
-				"compress":          compress,
-			},
-		},
+func setupLogFileDir(testName string) {
```
ashmeenkaur:
```suggestion
func setupLogFilePath(testName string) {
```

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -21,54 +21,73 @@ import (
...
+	var logFilePath string
+	logFilePath = path.Join(setup.TestDir(), GKETempDir, testName) + ".log"
+	if cfg.GKEMountedDirectory != "" { // GKE path
+		mountDir = cfg.GKEMountedDirectory
+		logFilePath = path.Join(GKETempDir, testName) + ".log"
+		if setup.ConfigFile() == "" {
+			// TODO: clean this up when GKE test migration completes.
+			logFilePath = "/gcsfuse-tmp/LogRotationTest.log"
```
ashmeenkaur:
```suggestion
			logFilePath = path.Join(GKETempDir, testName) + ".log"
```

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -21,54 +21,73 @@ import (
...
+		}
 	}
-	return yamlContent
+	cfg.LogFile = logFilePath
+	setup.SetLogFile(logFilePath)
```
ashmeenkaur: this is not required

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -47,25 +47,28 @@ func runOperationsOnFileTillLogRotation(t *testing.T, wg *sync.WaitGroup, fileNa
...
 	filePath := path.Join(testDirPath, fileName)
 	operations.CreateFileWithContent(filePath, filePerms, string(randomData), t)
-
+	currentLogFile := setup.LogFile()
```
ashmeenkaur: use cfg.LogFile in the tests

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -114,23 +116,32 @@ func TestLogRotation(t *testing.T) {
 	time.Sleep(1 * time.Second)
 
 	// Validate log files generated.
-	dirEntries := operations.ReadDirectory(logDirPath, t)
+	logFilesDirectory := path.Dir(setup.LogFile())
```
ashmeenkaur: use cfg.LogFile instead

tools/integration_tests/log_rotation/logrotate_logfile_test.go
```diff
@@ -114,23 +116,32 @@ func TestLogRotation(t *testing.T) {
...
+	dirEntries := operations.ReadDirectory(logFilesDirectory, t)
+
 	if len(dirEntries) != logFileCount {
 		t.Errorf("Expected log files in dirEntries folder: %d, got: %d",
 			logFileCount, len(dirEntries))
 	}
+
+	// Get the base name of the log file from the setup.
+	activeLogFileName := "LogRotationTest.log"
```
ashmeenkaur: can we use t.Name() here instead. Also change the config file flags to use the test name

tools/integration_tests/log_rotation/log_rotation_test.go
ashmeenkaur: Also is there a possibility of creating a common function in util package that can be reused in both these tests and read_cache tests

ashmeenkaur (commented):
The tests are backward incompatible as the script in run_test_mounted_directory.sh is failing.
https://github.com/GoogleCloudPlatform/gcsfuse/blob/6072d21aa32c950506e0522be980c38e8ee8f3fd/tools/integration_tests/run_tests_mounted_directory.sh#L333

Something like below conditional log file update needs to be added until GKE tests fully migrate:
https://github.com/GoogleCloudPlatform/gcsfuse/blob/6072d21aa32c950506e0522be980c38e8ee8f3fd/tools/integration_tests/read_cache/setup_test.go#L97

tools/integration_tests/run_gke_tests.sh
PranjalC100: sure, will remove it

tools/integration_tests/test_config.yaml
PranjalC100: okay will remove it, for now. Will clean this flag from the config files later in a separate PR. Wasn't using this log_file flag anywhere

tools/integration_tests/log_rotation/log_rotation_test.go
PranjalC100: removed!

tools/integration_tests/log_rotation/log_rotation_test.go
PranjalC100: Sure makes sense! Will update this file according to config

tools/integration_tests/log_rotation/log_rotation_test.go
PranjalC100: Yes, more generalized, updated!

tools/integration_tests/log_rotation/log_rotation_test.go
PranjalC100: Yes, since the function is not hardcoded, we could do that. Will raise this change in the next PR, along with the changes in the migrated packages

tools/integration_tests/log_rotation/log_rotation_test.go
PranjalC100: removed!

tools/integration_tests/log_rotation/logrotate_logfile_test.go
PranjalC100: removed!

tools/integration_tests/log_rotation/logrotate_logfile_test.go
PranjalC100: okay!

tools/integration_tests/log_rotation/log_rotation_test.go
```diff
@@ -77,42 +85,38 @@ func TestMain(m *testing.M) {
...
+	overrideFilePathsInFlagSet(cfg, setup.TestDir())
+
+	flags := setup.BuildFlagSets(*cfg, bucketType, "")
+	setupLogFilePath(testDirName)
+
+	successCode := static_mounting.RunTestsWithConfigFile(cfg, flags, m)
 
 	// Clean up test directory created.
-	setup.CleanupDirectoryOnGCS(ctx, storageClient, path.Join(setup.TestBucket(), testDirName))
+	setup.CleanupDirectoryOnGCS(ctx, storageClient, path.Join(setup.TestBucket(), "/gcsfuse-tmp/LogRotationTest"))
```
ashmeenkaur: This path won't be on the bucket right? We should clean up the testDirName on the bucket I believe. Can you check?

ashmeenkaur (approved):

PR #3902, opened 2025-10-17
test(storageutil): reduce flakiness caused by shorter timeout buffer in test

meet2mky:
### Description
Remove flakiness in test due to race between the stall duration and context timeout. 1ms is not enough and prone to race conditions where context deadline is not propagated quick enough. Running the original test with `--count=1000` fails almost always.
b/446816512

### Testing details
1. Manual - Done
2. Unit tests - Tested whole `retry_test.go` file many times with --count=10000 and it passed.
3. Integration tests - NA

### DIFF
--- a/output.txt
+++ b/output.txt
@@ -1,1 +1,0 @@
-ok  	github.com/googlecloudplatform/gcsfuse/v3/internal/storage/storageutil	104.179s


output.txt
```diff
@@ -0,0 +1 @@
+ok  	github.com/googlecloudplatform/gcsfuse/v3/internal/storage/storageutil	104.179s
```
Tulsishah: I think it is added my mistake.

meet2mky: Yeah, removed

internal/storage/storageutil/retry_test.go
```diff
@@ -248,24 +248,27 @@ func (t *ExecuteWithRetryTestSuite) TestExecuteWithRetry_RetryableThenNonRetryab
...
 		// Simulate a call that always takes longer than the per-attempt deadline.
 		select {
 		case <-time.After(stallDuration):
+			// This case should not be hit, as the context deadline
+			// is shorter by 10ms than stallDuration.
+			return "", errors.New("simulated apiCall finished before context timeout")
 		case <-ctx.Done():
 			return "", ctx.Err()
 		}
-		return "", ctx.Err()
```
Tulsishah, line 261: Isn't it require now?

meet2mky: Nope, it's not required as one of the select case will return.

PR #3905, opened 2025-10-17
refactor: dentry cache test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate dentry cache test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of dentry cache package so it can use config file.

[b/445953036](https://buganizer.corp.google.com/issues/445953036)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/dentry_cache/notifier_test.go
+++ b/tools/integration_tests/dentry_cache/notifier_test.go
@@ -39,7 +39,6 @@ type notifierTest struct {
 }
 
 func (s *notifierTest) SetupTest() {
-	//Truncate log file created.
 	testEnv.testDirPath = client.SetupTestDirectory(s.ctx, s.storageClient, testDirName)
 }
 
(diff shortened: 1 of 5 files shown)


tools/integration_tests/dentry_cache/notifier_test.go
```diff
@@ -15,49 +15,59 @@
...
-	flags []string
+	flags         []string
+	storageClient *storage.Client
+	ctx           context.Context
 	suite.Suite
 }
 
 func (s *notifierTest) SetupTest() {
-	mountGCSFuseAndSetupTestDir(s.flags, testDirName)
+	//Truncate log file created.
```
ashmeenkaur:
```suggestion
```

ashmeenkaur (dismissed):
One small nit, feel free to handle this in next PR.

ashmeenkaur (approved):

PR #3911, opened 2025-10-22
build(build-csi): Add .gcloudignore to reduce the amount of data transferred for CSI build

kislaykishore:
### Description
This reduces the amount of data from more than 69M to 4M.

### Testing details
1. Manual - yes
2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -1,6 +1,4 @@
 .gcloudignore
-# If you would like to upload your .git directory, .gitignore file or
-# files from your .gitignore file, remove the corresponding line below:
 .git
 .gitignore
 #!include:.gitignore


.gcloudignore
```diff
@@ -0,0 +1,51 @@
+# Binary names.
```
raj-prince: I see some common items in .gitignore and .gcloudignore. Have we tried re-using the list? - I see a way to do that here, under advanced topic section. (https://cloud.google.com/sdk/gcloud/reference/topic/gcloudignore).

Image

kislaykishore: Thanks! Fixed it.

PR #3912, opened 2025-10-22
feat(Unsupported Directory fix): Rename fix

Tulsishah:
### Description
Rename fix for unsupported directory 

[b/454485261](http://b/454485261)

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2570,8 +2570,13 @@ func (fs *fileSystem) renameNonHierarchicalDir(
 				return fmt.Errorf("delete file %q: %w", o.Name, err)
 			}
 		} else {
-			// For regular files, perform an in-place rename to the new directory.
-			if _, err = oldDir.RenameFile(ctx, o, newDir.Name().GcsObjectName()+nameDiff); err != nil {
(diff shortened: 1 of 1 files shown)


internal/fs/fs.go
```diff
@@ -2571,7 +2571,7 @@ func (fs *fileSystem) renameNonHierarchicalDir(
 			}
 		} else {
 			// For regular files, perform an in-place rename to the new directory.
-			if _, err = oldDir.RenameFile(ctx, o, path.Join(newDir.Name().GcsObjectName(), nameDiff)); err != nil {
+			if _, err = oldDir.RenameFile(ctx, o, newDir.Name().GcsObjectName()+nameDiff); err != nil {
```
vadlakondaswetha: will this work for regular files, does the nameDiff has a /? 
you have to add a comment to ensure that we dont change in future as part of adopting good code practices.

Tulsishah: We cannot create file names with a "/" in GCS. Any time a "/" is used, it is interpreted as creating a separate directory in GCS.

> you have to add a comment to ensure that we dont change in future as part of adopting good code practices.

Thanks. Added.

PR #3913, opened 2025-10-22
fix(b/454198297)): Upgrade Python along with gcloud installation

charith87:
Presubmit and Periodic failures due to incompatible python version with latest gcloud.
upgrading Python along with gcloud installation


kislaykishore:
I see that the integration test is failing. Could you check?

charith87:
> I see that the integration test is failing. Could you check?

Updated the files.

PR #3914, opened 2025-10-22
feat(logging): Logging Improvement Part 4 - Add MountInstanceID custom attribute to logger.

meet2mky:
### Description
Add custom attribute `fsName-MountInstanceID` to gcsfuse loggers to identify the logs from different mounts within same pod. This also helps in isolating logs from a single mount.

b/451499250

### Perf [Link](http://shortn/_ZWK7WZZ7t4)
image

### Testing details
Verified the logging in following scenarios

* Foreground mode with log file: ✅ 
* Background mode with log file: ✅ 
* Foreground mode without log file: ✅ 
* Background mode without log file: ✅  (Logs verified in syslog file)
* Verified on GKE mounting: ✅

1. Manual - Done
2. Unit tests - NA
3. Integration tests - Part of presubmit and perf test as well.

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -38,7 +38,7 @@ const (
 	ProgramName                  = "gcsfuse"
 	GCSFuseInBackgroundMode      = "GCSFUSE_IN_BACKGROUND_MODE"
 	GCSFuseMountInstanceIDEnvKey = "GCSFUSE_MOUNT_INSTANCE_ID"
-	GCSFuseMountInstanceIDKey    = "mount-id" // Combination of fsName and GCSFUSE_MOUNT_INSTANCE_ID
+	GCSFuseMountIDKey            = "mount-id" // Combination of fsName and GCSFUSE_MOUNT_INSTANCE_ID
 	textFormat                   = "text"
 	// Max possible length can be 32 as UUID has 32 characters excluding 4 hyphens.
 	mountInstanceIDLength = 8
@@ -226,7 +226,7 @@ func (f *loggerFactory) newLogger(level string) *slog.Logger {
 }
 
 func loggerAttr(fsName string) []slog.Attr {
-	return []slog.Attr{slog.String(GCSFuseMountInstanceIDKey, fmt.Sprintf("%s-%s", fsName, MountInstanceID()))}
+	return []slog.Attr{slog.String(GCSFuseMountIDKey, fmt.Sprintf("%s-%s", fsName, MountInstanceID()))}
 }
 
 // create a new logger with mountInstanceID set as custom attribute on logger.


internal/logger/logger.go
```diff
@@ -157,13 +158,10 @@ func MountInstanceID() string {
...
-func SetLogFormat(format string) {
-	if format == defaultLoggerFactory.format {
-		return
-	}
+// UpdateDefaultLogger updates the log format and creates a new logger with MountInstanceID set as custom attribute.
+func UpdateDefaultLogger(format, fsName string) {
 	defaultLoggerFactory.format = format
-	defaultLogger = defaultLoggerFactory.newLogger(defaultLoggerFactory.level)
+	defaultLogger = defaultLoggerFactory.newLoggerWithMountInstanceID(defaultLoggerFactory.level, fsName)
 }
```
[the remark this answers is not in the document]
meet2mky: Logger is re-initialised only once so it's not needed. Re-initialisation is needed to ensure the custom attribute is set. defaultLogger won't have the custom attribute even if it's of same format which is user provided.

internal/logger/logger.go
```diff
@@ -38,6 +38,7 @@ const (
 	ProgramName                  = "gcsfuse"
 	GCSFuseInBackgroundMode      = "GCSFUSE_IN_BACKGROUND_MODE"
 	GCSFuseMountInstanceIDEnvKey = "GCSFUSE_MOUNT_INSTANCE_ID"
+	GCSFuseMountInstanceIDKey    = "gcsfuse-mount-instance-id" // Combination of fsName and GCSFUSE_MOUNT_INSTANCE_ID
```
kislaykishore: Why not just gcsfuse-mount-id or mount-id? Smaller data will take less time to write as well, no?

internal/logger/logger.go
```diff
@@ -227,6 +225,19 @@ func (f *loggerFactory) newLogger(level string) *slog.Logger {
 	return logger
 }
 
+func loggerAttr(fsName string) []slog.Attr {
+	return []slog.Attr{slog.String(GCSFuseMountInstanceIDKey, fmt.Sprintf("%s-%s", fsName, MountInstanceID()))}
+}
+
+// create a new logger with mountInstanceID set as custom attribute on logger.
+func (f *loggerFactory) newLoggerWithMountInstanceID(level, fsName string) *slog.Logger {
```
kislaykishore, line 233: Should we rename `newLoggerWithMountInstanceID` to `newLogger` and delete the current `newLogger` method?

internal/logger/logger.go
meet2mky: `mount-id` seems short and concise enough. Changes done.

internal/logger/logger.go
meet2mky: `newLogger` is still used to initialize default Logger as soon as the logger package is initialise. See method at line 110 in this file. I don't think that defaultLogger initialize durring logger package initialization in `init` is used anywhere but we need careful understanding before removing so I am not considering that in scope of this change and can be taken up separately.

kislaykishore (approved):

PR #3915, opened 2025-10-22
refactor: benchmarking package migration [GKE-GCSFuse Test migration] 

PranjalC100:
### Description
This PR includes changes to migrate benchmarking package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of benchmarking package so it can use config file.

[b/445951822](https://buganizer.corp.google.com/issues/445951822)

### Testing details
2. Unit tests - NA


tools/integration_tests/test_config.yaml
```diff
@@ -420,6 +420,36 @@ read_cache:
         run: TestCacheFileForIncludeRegexTest
         run_on_gke: true
 
+benchmarking:
+  - test_bucket: "pranjal-bucket-1" # This would be populated by setup.TestBucket()
+    mounted_directory: "${MOUNTED_DIR}" # This would be populated by setup.MountedDirectory()
```
```suggestion
  - test_bucket: "${BUCKET_NAME}"
    mounted_directory: "${MOUNTED_DIR}"
```

tools/integration_tests/test_config.yaml
```diff
@@ -420,6 +420,36 @@ read_cache:
...
+          - "--stat-cache-ttl=0"
+          - "--enable-atomic-rename-object=true"
+        compatible:
+              flat: true
+              hns: true
+              zonal: true
+        run: "Benchmark_Rename"
+        run_on_gke: true
+      - flags:
+          - "--stat-cache-ttl=0"
```
ashmeenkaur, line 446: {"--client-protocol=grpc", "--stat-cache-ttl=0"} test missing

tools/integration_tests/test_config.yaml
```diff
@@ -420,6 +420,36 @@ read_cache:
...
+    configs:
+      - flags:
+          - "--stat-cache-ttl=0"
+        compatible:
+              flat: true
+              hns: true
+              zonal: true
+        run: "Benchmark_Stat"
+      - flags:
+          - "--stat-cache-ttl=0"
```
ashmeenkaur: these are the 2 flagsets
{"--stat-cache-ttl=0", "--enable-atomic-rename-object=true"}, 
{"--client-protocol=grpc", "--stat-cache-ttl=0", "--enable-atomic-rename-object=true"}

tools/integration_tests/test_config.yaml
```diff
@@ -420,6 +420,36 @@ read_cache:
         run: TestCacheFileForIncludeRegexTest
         run_on_gke: true
 
+benchmarking:
+  - test_bucket: "pranjal-bucket-1" # This would be populated by setup.TestBucket()
+    mounted_directory: "${MOUNTED_DIR}" # This would be populated by setup.MountedDirectory()
+    log_file: # Not required
+    configs:
+      - flags:
+          - "--stat-cache-ttl=0"
```
ashmeenkaur, line 429: {"--stat-cache-ttl=0"}
{"--client-protocol=grpc", "--stat-cache-ttl=0"}
are the 2 flagsets

tools/integration_tests/benchmarking/benchmark_delete_test.go
```diff
@@ -35,11 +35,12 @@ type benchmarkDeleteTest struct {
 }
 
 func (s *benchmarkDeleteTest) SetupB(b *testing.B) {
-	mountGCSFuseAndSetupTestDir(s.flags, testDirName)
+	mountGCSFuseAndSetupTestDir(s.flags, testEnv.ctx, testEnv.storageClient)
```
ashmeenkaur, line 38: can u check if this is happening at suite level? If not, can we do it at suite level for both gcsfuse and gke tests

PranjalC100: Since we only have 1 benchmark(test) in each file, so having a suite is the same as test level?

ashmeenkaur (approved):

PR #3921, opened 2025-10-23
refactor: read gcs algo test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate read gcs algo test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of read gcs algo cache package so it can use config file.

[b/454550408](https://buganizer.corp.google.com/issues/454550408)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/read_gcs_algo/read_gcs_algo_test.go
+++ b/tools/integration_tests/read_gcs_algo/read_gcs_algo_test.go
@@ -15,28 +15,67 @@
 package read_gcs_algo
 
 import (
+	"context"
+	"log"
 	"os"
 	"testing"
 
+	"cloud.google.com/go/storage"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/client"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting/static_mounting"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup"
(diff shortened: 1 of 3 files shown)


tools/integration_tests/read_gcs_algo/read_gcs_algo_test.go
```diff
@@ -15,28 +15,67 @@
...
+	// 1. Load and parse the common configuration.
+	cfg := test_suite.ReadConfigFile(setup.ConfigFile())
+	if len(cfg.ReadGCSAlgo) == 0 {
+		log.Println("No configuration found for list large dir tests in config. Using flags instead.")
+		// Populate the config manually.
+		cfg.ReadGCSAlgo = make([]test_suite.TestConfig, 1)
+		cfg.ReadGCSAlgo[0].TestBucket = setup.TestBucket()
+		cfg.ReadGCSAlgo[0].GKEMountedDirectory = setup.MountedDirectory()
+		cfg.ReadGCSAlgo[0].Configs = make([]test_suite.ConfigItem, 2)
+		// Do not enable fileCache as we want to test gcs read flow.
```
ashmeenkaur, line 50: can we move this comment to the config file instead for future understanding as this code will eventually be deleted once migration completes

ashmeenkaur (approved):

tools/integration_tests/read_gcs_algo/read_gcs_algo_test.go
PranjalC100: okay sure, will add it in the next PR!

PR #3923, opened 2025-10-23
fix(points-not-in-order): Add mount-id as a resource attribute for metrics and traces

kislaykishore:
### Description
This change adds the mount-id as a resource attribute to the OpenTelemetry metrics and trace exporters. This allows for the differentiation of metrics from multiple gcsfuse mounts leading to errors such as "Points not written in order". Additionally, the PID is removed from the metrics attributes since it was a no-op for some reason.

The mount-id is passed from the Mount function to the SetupOTelMetricExporters function, which then adds it as a service.instance.id attribute to the OpenTelemetry resource. The PID is removed from the WithFilteredResourceAttributes option in the setupCloudMonitoring function.

Repro:
```
for i in `seq 1 4`;do
rm -rf /tmp/mnt$i.log
mkdir -p ~/mnt$i
./gcsfuse --cloud-metrics-export-interval-secs=10 --log-file=/tmp/mnt$i.log <bkt> ~/mnt$i
sleep 2
done
```
Verified that service_instance_id gets populated for metrics: 
image

By reusing the getResource method, this change adds the mount-id as a resource attribute to the OpenTelemetry metrics exporter. This allows for the differentiation of traces from multiple gcsfuse mounts.

image

b/448007639, b/454621353,b/448006356

### Testing details
1. Manual - Tested

2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -414,7 +414,7 @@ func Mount(newConfig *cfg.Config, bucketName, mountPoint string) (err error) {
 			metricHandle = metrics.NewNoopMetrics()
 		}
 	}
-	shutdownTracingFn := monitor.SetupTracing(ctx, newConfig)
+	shutdownTracingFn := monitor.SetupTracing(ctx, newConfig, logger.MountInstanceID())
 	shutdownFn := common.JoinShutdownFunc(metricExporterShutdownFn, shutdownTracingFn)
 
 	// No-op if profiler is disabled.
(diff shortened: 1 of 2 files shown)


charith87 (approved):
Overall Approving. How do you validate if its working in GKE?

kislaykishore:
> Overall Approving. How do you validate if its working in GKE?

GKE has its own sets of identifiers like pod_uid, volume_name and bucket_name to ensure that the metrics timeseries for two mounts don't overlap.

PR #3925, opened 2025-10-24
feat(Unsupported Directory fix): Recursive delete implementation

Tulsishah:
### Description

When making an RmDir call, we begin by executing a Read operation to validate the presence of any unsupported directories. If any are present, we initiate a sequence of delete calls to remove all contained objects. Only once the directory is confirmed to be empty will the RmDir call proceed.

[b/454790614](https://b.corp.google.com/issues/454790614) 

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/internal/storage/storageutil/unsupported_object_util.go
+++ b/internal/storage/storageutil/unsupported_object_util.go
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	unsupportedObjectNameSubstrings = []string{"//", ".."}
+	unsupportedObjectNameSubstrings = []string{"//", "..", "/./"}
 	unsupportedObjectNamePrefixes   = []string{"/"}
 	unsupportedObjectNames          = []string{"", ".", ".."}
 )


internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,93 @@ func (d *dirInode) DeleteChildDir(
...
+			// Handle single file-like object deletion.
+			if err := d.deleteSingleObject(ctx, objectName); err != nil {
+				return fmt.Errorf("deleting unsupported object %q: %w", objectName, err)
+			}
+		}
+	}
+	return nil
+}
+
+// Helper to delete a single object, handling 'Not Found' errors gracefully.
```
vadlakondaswetha, line 1058: what is the need to handle NotFound errors gracefully? can you add a comment here.

internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,93 @@ func (d *dirInode) DeleteChildDir(
...
+func (d *dirInode) deleteSingleObject(ctx context.Context, objectName string) error {
+	if d.isBucketHierarchical() && strings.HasSuffix(objectName, "/") {
+		if err := d.bucket.DeleteFolder(ctx, objectName); err != nil {
```
vadlakondaswetha: what is the need to handle HNS here? only for completeness?

internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,93 @@ func (d *dirInode) DeleteChildDir(
...
+			if !errors.As(err, &notFoundErr) {
+				return err
+			}
+		}
+	} else {
+		if err := d.bucket.DeleteObject(ctx, &gcs.DeleteObjectRequest{Name: objectName}); err != nil {
+			var notFoundErr *gcs.NotFoundError
+			if !errors.As(err, &notFoundErr) {
+				return err
+			}
```
vadlakondaswetha: what about typecache - dont we need to delete type cache entries?

internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,93 @@ func (d *dirInode) DeleteChildDir(
...
+
+// Helper to delete a single object, handling 'Not Found' errors gracefully.
+func (d *dirInode) deleteSingleObject(ctx context.Context, objectName string) error {
+	if d.isBucketHierarchical() && strings.HasSuffix(objectName, "/") {
+		if err := d.bucket.DeleteFolder(ctx, objectName); err != nil {
+			var notFoundErr *gcs.NotFoundError
+			if !errors.As(err, &notFoundErr) {
+				return err
+			}
+		}
```
vadlakondaswetha, line 1070: you can return nil from if-block and avoid else. improves readability something like this
if d.isBucketHierarchial() {
 deleteFolder()
 return
}

deleteObject()

internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,93 @@ func (d *dirInode) DeleteChildDir(
...
+			if err := d.deleteSingleObject(ctx, objectName); err != nil {
+				return fmt.Errorf("deleting unsupported object %q: %w", objectName, err)
+			}
+		}
+	}
+	return nil
+}
+
+// Helper to delete a single object, handling 'Not Found' errors gracefully.
+func (d *dirInode) deleteSingleObject(ctx context.Context, objectName string) error {
```
vadlakondaswetha: nit: rename to deleteObjecT()

internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,93 @@ func (d *dirInode) DeleteChildDir(
...
+				// It's a file, delete it.
+				objName := obj.Name // Capture loop variable.
+				g.Go(func() error {
+					return d.deleteSingleObject(gCtx, objName)
+				})
+			}
+		}
+
+		for _, nestedPrefix := range objects.CollapsedRuns {
+			nestedPrefix := nestedPrefix // Capture loop variable.
```
vadlakondaswetha: why we need to copy ?

internal/fs/fs.go
```diff
@@ -2186,12 +2186,23 @@ func (fs *fileSystem) RmDir(
...
 		}
 
+		// If there are unsupported objects, delete them recursively.
+		if len(unsupportedDirs) > 0 {
+			err = childDir.DeleteObjects(ctx, unsupportedDirs)
+			if err != nil {
+				return fmt.Errorf("RmDir: failed to delete unsupported objects: %w", err)
+			}
+			// After deleting, we need to re-check for emptiness.
+			continue
```
vadlakondaswetha: is there any case where we end up in infinite loop - delete is returning NotFound error and REadDir is returning object?

I am wondering why do we need to check again?

internal/fs/inode/dir.go
Tulsishah: Removed. Thanks

internal/fs/inode/dir.go
Tulsishah: We have confirmed that unsupported directories correctly excluded from type cache insertion during the following operations:
1. [List](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/dir.go#L760)
2. Copy - Kernel can not read unsupported objects so it won't be able to send copy object call for them.
3. [Rename](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/dir.go#L1076)

This exclusion eliminates the need for any subsequent cleanup logic, meaning we no longer require a specific remove operation to delete these unsupported entries from the type cache.

internal/fs/fs.go
Tulsishah: > delete is returning NotFound error and REadDir is returning object

It should not happen ideally.

> I am wondering why do we need to check again?

I initially considered  to keep it like the original logic. Check if a directory contains no objects before proceeding with deletion. 
However, this approach risks creating an infinite loop if another process begins creating objects concurrently within that directory. To avoid this race condition and ensure the user receives an error promptly, I removed the `continue` statement.

internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,95 @@ func (d *dirInode) DeleteChildDir(
...
+		for _, obj := range objects.MinObjects {
+			// obj.Name is guaranteed to start with 'prefix'.
+			if !strings.HasSuffix(obj.Name, "/") {
+				// It's a file, delete it.
+				objName := obj.Name // Capture loop variable.
+				g.Go(func() error {
+					return d.deleteObject(gCtx, objName)
+				})
+			}
+		}
```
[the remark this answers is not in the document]
Tulsishah: This is not required because Collasperun already contains all the directory object prefixes. We then make a deletePrefixRecursively call for each directory prefix.

internal/fs/fs.go
vadlakondaswetha: another option would be check for emptyness after delete and throw an error if its not empty instead of doing it via a loop

internal/fs/inode/dir.go
```diff
@@ -1034,6 +1037,96 @@ func (d *dirInode) DeleteChildDir(
...
+			return fmt.Errorf("listing objects under prefix %q: %w", prefix, err)
+		}
+
+		// 1. Delete all file-like objects and recurse into subdirectories in parallel.
+		g, gCtx := errgroup.WithContext(ctx)
+		for _, obj := range objects.MinObjects {
+			// obj.Name is guaranteed to start with 'prefix'.
+			if !strings.HasSuffix(obj.Name, "/") {
+				// It's a file, delete it.
+				objName := obj.Name // Capture loop variable.
```
vadlakondaswetha, line 1103: remove copying the name into seperate variable here also.

vadlakondaswetha (approved):

internal/fs/fs.go
Tulsishah: That will increase one List Call which is class A operation. Let's think it separately. I'll raise separate task for this.

internal/fs/inode/dir.go
Tulsishah: Will do it in next PR to avoid running CI checks again.

PR #3927, opened 2025-10-24
refactor: interrupt test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate interrupt test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of interrupt package so it can use config file.

[b/454549489](https://buganizer.corp.google.com/issues/454549489)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/interrupt/interrupt_test.go
+++ b/tools/integration_tests/interrupt/interrupt_test.go
@@ -60,7 +60,6 @@ func TestMain(m *testing.M) {
 		}
 		cfg.Interrupt[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": true}
 		cfg.Interrupt[0].Configs[1].Flags = []string{
-			// TODO(b/417136852): Enable this test for Zonal Bucket also once read start working.
 			"--enable-streaming-writes=true",
 		}
 		cfg.Interrupt[0].Configs[1].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": false}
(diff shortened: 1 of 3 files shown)


tools/integration_tests/interrupt/interrupt_test.go
```diff
@@ -25,21 +25,50 @@ import (
...
+			"--implicit-dirs=true --enable-streaming-writes=false",
+			"--ignore-interrupts=true --enable-streaming-writes=false",
+			"--ignore-interrupts=false --enable-streaming-writes=false",
+		}
+		cfg.Interrupt[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": true}
+		cfg.Interrupt[0].Configs[1].Flags = []string{
+			// TODO(b/417136852): Enable this test for Zonal Bucket also once read start working.
+			"--enable-streaming-writes=true",
+		}
+		cfg.Interrupt[0].Configs[1].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": false}
```
ashmeenkaur, line 65: This is missing in the config file

PranjalC100: https://github.com/GoogleCloudPlatform/gcsfuse/pull/3927/files#diff-d60ee5bab6c5c65105f13462b7a991b31dfb804b810180bba4ddefd26b7c20d5R491-R496 can you check this out once?

ashmeenkaur: sorry my bad, missed it because of the ordering. LGTM!
Please move the TODO to config file instead.

ashmeenkaur (dismissed):

tools/integration_tests/interrupt/interrupt_test.go
PranjalC100: okay!

ashmeenkaur (approved):

PR #3929, opened 2025-10-24
feat(visualizer): adding read-manager wrapper to dump read-io visuals

raj-prince:
### Description
Implementing a wrapper for read-manager, which records read-io and dumps the visuals on Destroy().

b/454839875

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/read_manager/visual_read_manager.go
+++ b/internal/gcsx/read_manager/visual_read_manager.go
@@ -16,6 +16,7 @@ package read_manager
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"sync"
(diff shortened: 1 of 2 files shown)


internal/gcsx/read_manager/visual_read_manager.go
```diff
@@ -0,0 +1,147 @@
...
+			_, err = f.Write([]byte(output))
+			if err == nil {
+				return
+			}
+			logger.Warnf("Failed to write to output file: %v", err)
+		}
+	}
+
+	// Fallback to logging output
+	fmt.Println(output)
```
kislaykishore: Refactoring into another function would help reduce the code-complexity.

```suggestion
    func writeToFile(path, text string) error {
      if path == "" {
        return fmt.Errorf("empty path")
      }
      f, err := os.OpenFile(vrm.outputFilePath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0600)
      if err != nil {
            return fmt.Errorf("Failed to open output file: %w", err)
	  }
	  defer f.Close() 
	  if _, err = f.Write([]byte(output)); err != nil {
	    return fmt.Errorf("error while writing to file: %w", err)
	  }
    }
    if vrm.outputFilePath == "" {
      fmt.Println(output)
      return
    }
    if err := writeToFile(vrm.outputFilePath, output); err != nil {
      fmt.Println(output)
      logger.Warnf("Failed to write to output file: %v", err)
      return
    }
```

kislaykishore (approved):
LGTM - just one feedback.

internal/gcsx/read_manager/visual_read_manager.go
raj-prince: Thanks! Done.

internal/gcsx/read_manager/visual_read_manager.go
```diff
@@ -0,0 +1,161 @@
...
+// acceptRange records a read I/O range and merges it with existing ranges if possible.
+func (vrm *VisualReadManager) acceptRange(start, end uint64) {
+	if end <= start {
+		return // Invalid range, ignore
+	}
+
+	// Clamp end to object size.
+	if end > vrm.Object().Size {
+		end = vrm.Object().Size
+	}
```
kislaykishore, line 110: nit: modernize
```suggestion
		end = max(vrm.Object().Size, end)
```

raj-prince: I'll take this in the next PR.

PR #3931, opened 2025-10-24
feat(O_DIRECT): Introduce O-DIRECT flag while opening a file

alleaditya:
### Description
In order to enable O-DIRECT at GCSFuse level, add ability to enable O-DIRECT using mount options. This will be a hidden flag and default to `false`

[454888942](https://b.corp.google.com/issues/454888942)

### Testing details
1. Manual - NA
2. Unit tests - ✅
3. Integration tests - NA

NA. The default value of the flag is false

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2791,7 +2791,6 @@ func (fs *fileSystem) OpenFile(
 	op *fuseops.OpenFileOp) (err error) {
 	// Bypass the kernel's page cache for file reads and writes
 	if fs.newConfig.FileSystem.ODirect {
-		logger.Infof("O_DIRECT enabled for file: %d", op.Inode)
 		op.UseDirectIO = true
 	}
 


internal/fs/fs.go
```diff
@@ -2789,6 +2789,12 @@ func (fs *fileSystem) ReleaseDirHandle(
 func (fs *fileSystem) OpenFile(
 	ctx context.Context,
 	op *fuseops.OpenFileOp) (err error) {
+	// Bypass the kernel's page cache for file reads and writes
+	if fs.newConfig.FileSystem.ODirect {
+		logger.Infof("O_DIRECT enabled for file: %d", op.Inode)
```
raj-prince: We don't need this info log, o-direct flag will enable for all the openFile calls.

internal/fs/fs.go
```diff
@@ -2789,6 +2789,12 @@ func (fs *fileSystem) ReleaseDirHandle(
 func (fs *fileSystem) OpenFile(
 	ctx context.Context,
 	op *fuseops.OpenFileOp) (err error) {
+	// Bypass the kernel's page cache for file reads and writes
+	if fs.newConfig.FileSystem.ODirect {
```
raj-prince, line 2793: Can we also add a fs_test for o-direct - not exhaustive, two tests - one for o-direct enabled and other for o-direct disabled should be enough.

raj-prince (changes requested):
Couple of comments.

internal/fs/fs.go
alleaditya: Added it on recommendation from Gemini review.

internal/fs/fs.go
raj-prince: We can skip writing test at fs_level, but we need to add an e2e-test to test the o-direct behavior.

alleaditya (commented):
> LGTM.
> 
> Please manually test once, w/wo flag enabled once before merging. 

Verfied the results and found Page cache is enabled only when mounting without `--o-direct` flag.
https://screenshot.googleplex.com/4QeKZxAvMXXcbEb

PR #3940, opened 2025-10-27
fix: Propagate OTel tracing context in fs-ops

thrivikram-karur-g:
### Description
The OTel trace instrumentation setup in internal/fs/wrappers/tracing.go is not propagated when ignoreInterrupts is set default to true in internal/fs/fs.go file system interface methods. Because of this the trace context from WithTracing wrapper is lost and underlying libraries like go storage grpc instrumentation is not tracked under the same trace. This PR ensures the span context from parent is still passed down when tracing is enabled.

b/454805201

### Testing details
1. Manual - DONE

The following test plan was followed to test tracing is properly working as expected.
Enable experimental monitoring/tracing config with sampling ratio 1 to track all traces and export them to gcloud.
Using the mount instance id, filter the traces and check the trace for any of the file system methods like WriteFile or LookupInode. 
Compare these trace results with and without fix and it can be seen that the trace is as expected with the fix.

Screenshots below:

Trace with fix:

trace_with_fix

Trace without fix:

trace_without_fix

2. Unit tests - Added
Added unittests for all the fs ops method in fs/fs.go using fake bucker manager to check spans generated.

3. Integration tests - NA

NA

Perf results from the kokoro build run on the PR:

| Branch | File Size | Read BW | Write BW | RandRead BW | RandWrite BW |
| :---: | :---: | :---: | :---: | :---: | :---: |
| **Master** | 0.25MiB | 574.81MiB/s | 1.32MiB/s | 81.15MiB/s | 1.08MiB/s |
| **PR** | 0.25MiB | 583.88MiB/s | 1.21MiB/s | 77.8MiB/s | 1.2MiB/s |
| **Master** | 48.828MiB | 4725.19MiB/s | 79.43MiB/s | 1600.38MiB/s | 85.94MiB/s |
| **PR** | 48.828MiB | 4701.76MiB/s | 85.17MiB/s | 1630.13MiB/s | 85.4MiB/s |
| **Master** | 976.562MiB | 4724.65MiB/s | 37.67MiB/s | 864.03MiB/s | 40.05MiB/s |
| **PR** | 976.562MiB | 4732.25MiB/s | 38.19MiB/s | 977.97MiB/s | 39.7MiB/s |

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -573,6 +573,8 @@ type MetricsConfig struct {
 type MonitoringConfig struct {
 	ExperimentalTracingMode string `yaml:"experimental-tracing-mode"`
 
+	ExperimentalTracingProjectId string `yaml:"experimental-tracing-project-id"`
+
 	ExperimentalTracingSamplingRatio float64 `yaml:"experimental-tracing-sampling-ratio"`
 }
 
@@ -837,6 +839,12 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 	flagSet.StringP("experimental-tracing-mode", "", "", "Experimental: specify tracing mode")
 
 	if err := flagSet.MarkHidden("experimental-tracing-mode"); err != nil {
+		return err
+	}
+
+	flagSet.StringP("experimental-tracing-project-id", "", "", "Experimental: specify the GCP project-id to which traces will be exported. When unset, a project-id will be inferred as per the default credential detection process")
+
+	if err := flagSet.MarkHidden("experimental-tracing-project-id"); err != nil {
 		return err
 	}
 
@@ -1296,6 +1304,10 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 	}
 
 	if err := v.BindPFlag("monitoring.experimental-tracing-mode", flagSet.Lookup("experimental-tracing-mode")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("monitoring.experimental-tracing-project-id", flagSet.Lookup("experimental-tracing-project-id")); err != nil {
 		return err
 	}
 
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -865,6 +865,13 @@ params:
     default: ""
     hide-flag: true
 
+  - config-path: "monitoring.experimental-tracing-project-id"
+    flag-name: "experimental-tracing-project-id"
+    type: "string"
+    usage: "Experimental: specify the GCP project-id to which traces will be exported. When unset, a project-id will be inferred as per the default credential detection process"
+    default: ""
+    hide-flag: true
+
   - config-path: "monitoring.experimental-tracing-sampling-ratio"
     flag-name: "experimental-tracing-sampling-ratio"
     type: "float64"
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -92,16 +92,18 @@ func registerTerminatingSignalHandler(mountPoint string) {
 	}()
 }
 
-func getUserAgent(appName string, config string) string {
+func getUserAgent(appName, config, mountInstanceID string) string {
+	var userAgent string
 	gcsfuseMetadataImageType := os.Getenv("GCSFUSE_METADATA_IMAGE_TYPE")
 	if len(gcsfuseMetadataImageType) > 0 {
-		userAgent := fmt.Sprintf("gcsfuse/%s %s (GPN:gcsfuse-%s) (Cfg:%s)", common.GetVersion(), appName, gcsfuseMetadataImageType, config)
-		return strings.Join(strings.Fields(userAgent), " ")
+		userAgent = fmt.Sprintf("gcsfuse/%s %s (GPN:gcsfuse-%s) (Cfg:%s)", common.GetVersion(), appName, gcsfuseMetadataImageType, config)
+		userAgent = strings.Join(strings.Fields(userAgent), " ")
 	} else if len(appName) > 0 {
-		return fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-%s) (Cfg:%s)", common.GetVersion(), appName, config)
+		userAgent = fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-%s) (Cfg:%s)", common.GetVersion(), appName, config)
 	} else {
-		return fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse) (Cfg:%s)", common.GetVersion(), config)
+		userAgent = fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse) (Cfg:%s)", common.GetVersion(), config)
 	}
+	return fmt.Sprintf("%s (mount-id:%s)", userAgent, mountInstanceID)
 }
 
 func boolToBin(b bool) string {
@@ -172,7 +174,7 @@ func mountWithArgs(bucketName string, mountPoint string, newConfig *cfg.Config, 
 	// connection.
 	var storageHandle storage.StorageHandle
 	if bucketName != canned.FakeBucketName {
-		userAgent := getUserAgent(newConfig.AppName, getConfigForUserAgent(newConfig))
+		userAgent := getUserAgent(newConfig.AppName, getConfigForUserAgent(newConfig), logger.MountInstanceID(fsName(bucketName)))
 		logger.Info("Creating Storage handle...")
 		storageHandle, err = createStorageHandle(newConfig, userAgent, metricHandle)
 		if err != nil {
--- a/cmd/legacy_main_test.go
+++ b/cmd/legacy_main_test.go
@@ -70,27 +70,27 @@ func (t *MainTest) TestGetUserAgentWhenMetadataImageTypeEnvVarIsSet() {
 	t.T().Setenv("GCSFUSE_METADATA_IMAGE_TYPE", "DLVM")
 	mountConfig := &cfg.Config{}
 
-	userAgent := getUserAgent("AppName", getConfigForUserAgent(mountConfig))
+	userAgent := getUserAgent("AppName", getConfigForUserAgent(mountConfig), "testFS-123")
 
-	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s AppName (GPN:gcsfuse-DLVM) (Cfg:0:0:0:0:0:0)", common.GetVersion()))
+	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s AppName (GPN:gcsfuse-DLVM) (Cfg:0:0:0:0:0:0) (mount-id:testFS-123)", common.GetVersion()))
 	assert.Equal(t.T(), expectedUserAgent, userAgent)
 }
 
 func (t *MainTest) TestGetUserAgentWhenMetadataImageTypeEnvVarIsNotSet() {
 	mountConfig := &cfg.Config{}
 
-	userAgent := getUserAgent("AppName", getConfigForUserAgent(mountConfig))
+	userAgent := getUserAgent("AppName", getConfigForUserAgent(mountConfig), "testFS-123")
 
-	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0)", common.GetVersion()))
+	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0) (mount-id:testFS-123)", common.GetVersion()))
 	assert.Equal(t.T(), expectedUserAgent, userAgent)
 }
 
 func (t *MainTest) TestGetUserAgentConfigWithNoFileCache() {
 	mountConfig := &cfg.Config{}
 
-	userAgent := getUserAgent("AppName", getConfigForUserAgent(mountConfig))
+	userAgent := getUserAgent("AppName", getConfigForUserAgent(mountConfig), "testFS-123")
 
-	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0)", common.GetVersion()))
+	expectedUserAgent := strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0) (mount-id:testFS-123)", common.GetVersion()))
 	assert.Equal(t.T(), expectedUserAgent, userAgent)
 }
 
@@ -108,7 +108,7 @@ func (t *MainTest) TestGetUserAgentConfig() {
 					MaxSizeMb: 0,
 				},
 			},
-			expectedUserAgent: strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0)", common.GetVersion())),
+			expectedUserAgent: strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0) (mount-id:testFS-123)", common.GetVersion())),
 		},
 		{
 			name: "Config with file cache disabled where maxsize is set but cache dir is not set.",
@@ -117,7 +117,7 @@ func (t *MainTest) TestGetUserAgentConfig() {
 					MaxSizeMb: -1,
 				},
 			},
-			expectedUserAgent: strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0)", common.GetVersion())),
+			expectedUserAgent: strings.TrimSpace(fmt.Sprintf("gcsfuse/%s (GPN:gcsfuse-AppName) (Cfg:0:0:0:0:0:0) (mount-id:testFS-123)", common.GetVersion())),
 		},
 		{
 			name: "Config with file cache enabled but random read disabled.",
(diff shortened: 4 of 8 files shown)


internal/fs/fs.go
```diff
@@ -1681,7 +1682,7 @@ func (fs *fileSystem) LookUpInode(
 	if fs.newConfig.FileSystem.IgnoreInterrupts {
 		// When ignore interrupts config is set, we are creating a new context not
 		// cancellable by parent context.
-		ctx = context.Background()
+		ctx = monitor.GetTraceContext(ctx, fs.newConfig)
```
thrivikram-karur-g: from: kislay.
Pass old and new context and patch/augment trace/span context from the old context to the new when tracing is enabled to keep trace context method clean

thrivikram-karur-g: this is done now

internal/fs/fs.go
```diff
@@ -1681,7 +1682,8 @@ func (fs *fileSystem) LookUpInode(
 	if fs.newConfig.FileSystem.IgnoreInterrupts {
```
thrivikram-karur-g, line 1696: reduce this deduplication as well

internal/fs/fs.go
```diff
@@ -1674,15 +1675,24 @@ func (fs *fileSystem) StatFS(
...
+// getInterruptlessContext returns a new context that is not cancellable by the
+// parent context if the ignore-interrupts flag is set. Otherwise, it returns
+// the original context.
+func (fs *fileSystem) getInterruptlessContext(ctx context.Context) context.Context {
 	if fs.newConfig.FileSystem.IgnoreInterrupts {
 		// When ignore interrupts config is set, we are creating a new context not
 		// cancellable by parent context.
-		ctx = context.Background()
+		newCtx := context.Background()
+		return monitor.AugmentTraceContext(newCtx, ctx, fs.newConfig)
```
kislaykishore: In order to avoid pointer indirection for every fs-op, can we capture whether tracing is enabled in fs as a boolean data-member in fileSystem struct and use that.

internal/monitor/util.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"context"
+
+	newcfg "github.com/googlecloudplatform/gcsfuse/v3/cfg"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// When tracing is enabled ensure span & trace context from oldCtx is passed on to newCtx
+func AugmentTraceContext(newCtx context.Context, oldCtx context.Context, cfg *newcfg.Config) context.Context {
```
kislaykishore: Given that we are doing this only in fs, would it make sense to move this to fs itself? We can move it to a separate package if it sees more usage.

internal/monitor/util.go
```diff
@@ -0,0 +1,32 @@
...
+	if newcfg.IsTracingEnabled(cfg) {
+		span := trace.SpanFromContext(oldCtx)
+		newCtx = trace.ContextWithSpan(newCtx, span)
+	}
+
+	return newCtx
```
kislaykishore: The most common workflow would be where tracing is disabled. So, I'd recommend putting that first.
```suggestion
	if !newcfg.IsTracingEnabled(cfg) {
	  return newCtx
	}
	span := trace.SpanFromContext(oldCtx)
	return trace.ContextWithSpan(newCtx, span)
```

internal/monitor/util_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}{
+		{
+			name:                  "TracingEnabled_ShouldPropagateSpan",
+			tracingEnabled:        true,
+			expectSpanPropagation: true,
+		},
+		{
+			name:                  "TracingDisabled_ShouldNotPropagateSpan",
+			tracingEnabled:        false,
+			expectSpanPropagation: false,
```
kislaykishore: Instead of this test, I'd rather have tests that verify that traces are being correctly published. For instance, what if I remove the call to AugmentTraceContext in the `getInterruptlessContext` method. This test won't catch that.

Perhaps you could do something similar to what's being done in [metrics_test](https://github.com/GoogleCloudPlatform/gcsfuse/blob/1f0cd901d889603891ba2e96890b8f593d402242/internal/fs/metrics_test.go) for reference. 

* Implement a test-file-system
* Wrap with the [trace-wrapper](https://github.com/GoogleCloudPlatform/gcsfuse/blob/1f0cd901d889603891ba2e96890b8f593d402242/internal/fs/wrappers/tracing.go).
* Test with in-memory-exporter as is being done in [tracing_test](https://github.com/GoogleCloudPlatform/gcsfuse/blob/1f0cd901d889603891ba2e96890b8f593d402242/internal/fs/wrappers/tracing_test.go)
* Test for cases where ignore-interrupts is true and false.
* Don't write tests for `augmentTraceContext` which is an inner detail.

Let me know if it doesn't work.

internal/fs/fs.go
thrivikram-karur-g: Added isTracingEnabled as a boolean in file system struct.

internal/monitor/util.go
thrivikram-karur-g: yeah. moving for now, as stated will separate it out in future as per usage.

internal/monitor/util.go
thrivikram-karur-g: done

internal/monitor/util_test.go
thrivikram-karur-g: as discussed offline added the best extent of tests we could have for now, if fake bucket get's tracing support our test could actually update to cover the spans happening in child calls.

internal/fs/fs.go
```diff
@@ -1674,15 +1679,38 @@ func (fs *fileSystem) StatFS(
...
-	op *fuseops.LookUpInodeOp) (err error) {
+// When tracing is enabled ensure span & trace context from oldCtx is passed on to newCtx
+func augmentTraceContext(newCtx context.Context, oldCtx context.Context, isTracingEnabled bool) context.Context {
+	if !isTracingEnabled {
+		return newCtx
+	}
+
+	span := trace.SpanFromContext(oldCtx)
+	newCtx = trace.ContextWithSpan(newCtx, span)
+	logger.Info("Logging new context:", newCtx)
```
kislaykishore: Remove this log. If tracing is enabled, I think it'll take CPU cycles without offering any useful information.

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,116 @@
...
+	require.Len(t, ss, 1)
+	assert.Equal(t, "LookUpInode", ss[0].Name)
+	assert.Equal(t, trace.SpanKindServer, ss[0].SpanKind)
+}
+
+func TestTraceLookupInode(t *testing.T) {
+	ctx := context.Background()
+	bucket, server := createTestFileSystemWithTraces(ctx, t, true)
+
+	runTestWithLookUpNode(t, bucket, server)
```
kislaykishore: Please inline the test and don't hide it away in a separate method. You can use [table-driven](https://go.dev/wiki/TableDrivenTests) tests for code reuse. Apart from testing, tests are useful to understand the functionality. What tests are doing should also be easily apparent and the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), which is very useful for writing business logic should take a backseat for tests.

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	ctx = context.Background()
+	bucket, server = createTestFileSystemWithTraces(ctx, t, false)
+
+	runTestWithLookUpNode(t, bucket, server)
+}
```
kislaykishore, line 1343: Could you also add tests for other fs-ops as well?

internal/fs/fs.go
thrivikram-karur-g: apologies, my bad. I was testing something, will remove.

internal/fs/tracing_test.go
thrivikram-karur-g: sure, will do and update here.

internal/fs/tracing_test.go
thrivikram-karur-g: thanks for pointing, will make the code better

internal/fs/fs.go
thrivikram-karur-g: removed

internal/fs/tracing_test.go
thrivikram-karur-g: done

internal/fs/fs.go
```diff
@@ -1674,15 +1679,37 @@ func (fs *fileSystem) StatFS(
 	return
 }
 
-// LOCKS_EXCLUDED(fs.mu)
-func (fs *fileSystem) LookUpInode(
-	ctx context.Context,
-	op *fuseops.LookUpInodeOp) (err error) {
+// When tracing is enabled ensure span & trace context from oldCtx is passed on to newCtx
+func augmentTraceContext(newCtx context.Context, oldCtx context.Context, isTracingEnabled bool) context.Context {
```
kislaykishore: Rename it to `maybePropagateTraceContext`

internal/fs/fs.go
```diff
@@ -1674,15 +1679,37 @@ func (fs *fileSystem) StatFS(
...
+	if !isTracingEnabled {
+		return newCtx
+	}
+
+	span := trace.SpanFromContext(oldCtx)
+	newCtx = trace.ContextWithSpan(newCtx, span)
+
+	return newCtx
```
kislaykishore: You can merge these into a single statement.

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,118 @@
...
+	t.Helper()
+	ex := tracetest.NewInMemoryExporter()
+	otel.SetTracerProvider(sdktrace.NewTracerProvider(sdktrace.WithSyncer(ex)))
+	return ex
+}
+
+func TestTraceLookupInode(t *testing.T) {
+	ctx := context.Background()
+	var ignoreInterruptTestCases = []struct {
+		caseName string
```
kislaykishore: Just `name` is good enough.

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,118 @@
...
+		value    bool
```
kislaykishore: Rename to `ignoreInterrupts`

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,118 @@
...
+		{"disabled", false},
+	}
+	for _, tt := range ignoreInterruptTestCases {
+		t.Run(tt.caseName, func(t *testing.T) {
+			ex := newInMemoryExporter(t)
+			t.Cleanup(func() {
+				ex.Reset()
+			})
+			bucket, server := createTestFileSystemWithTraces(ctx, t, tt.value)
+
```
kislaykishore, line 900: Remove empty line

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,118 @@
...
+			ctx := context.Background()
+			fileName := "test.txt"
+			content := "test content"
+			createWithContents(ctx, t, bucket, fileName, content)
+
```
kislaykishore, line 929: Remove empty line

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,118 @@
...
+			lookupOp := &fuseops.LookUpInodeOp{
+				Parent: fuseops.RootInodeID,
+				Name:   fileName,
+			}
+
```
kislaykishore, line 1079: Remove empty line

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,118 @@
...
+			m := wrappers.WithTracing(server)
+			err := m.LookUpInode(context.Background(), lookupOp)
```
kislaykishore: Add a blank line before this.

internal/fs/fs.go
thrivikram-karur-g: done

internal/fs/fs.go
thrivikram-karur-g: done

internal/fs/tracing_test.go
thrivikram-karur-g: done

internal/fs/tracing_test.go
thrivikram-karur-g: done

internal/fs/tracing_test.go
thrivikram-karur-g: done

internal/fs/tracing_test.go
thrivikram-karur-g: done

internal/fs/tracing_test.go
thrivikram-karur-g: done

internal/fs/tracing_test.go
thrivikram-karur-g: done

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,115 @@
...
+func newInMemoryExporter(t *testing.T) *tracetest.InMemoryExporter {
+	t.Helper()
+	ex := tracetest.NewInMemoryExporter()
+	otel.SetTracerProvider(sdktrace.NewTracerProvider(sdktrace.WithSyncer(ex)))
+	return ex
+}
+
+func TestTraceLookupInode(t *testing.T) {
+	ctx := context.Background()
+	var ignoreInterruptTestCases = []struct {
```
kislaykishore: `testCases := [] struct {` is more succinct.

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,115 @@
...
+			createWithContents(ctx, t, bucket, fileName, content)
+			lookupOp := &fuseops.LookUpInodeOp{
+				Parent: fuseops.RootInodeID,
+				Name:   fileName,
+			}
+			m := wrappers.WithTracing(server)
+
+			err := m.LookUpInode(context.Background(), lookupOp)
+			require.NoError(t, err)
+			ss := ex.GetSpans()
```
kislaykishore, line 462: Add a blank line before this

internal/fs/tracing_test.go
thrivikram-karur-g: done but vscode go extension is automatically converting it into []struct { instead of [] struct { .
I think lint wise that maybe standard to not give space between [] and struct ?
For now, renamed though.

internal/fs/tracing_test.go
thrivikram-karur-g: done, and made few more readability changes and also added all fs-ops testcases

internal/fs/tracing_test.go
thrivikram-karur-g: added all fs-ops testcases

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,1406 @@
...
+				assert.Equal(t, spanName, ss[i].Name)
+				assert.Equal(t, trace.SpanKindServer, ss[i].SpanKind)
+			}
+		})
+	}
+}
+
+func TestTraceReadDir(t *testing.T) {
+	ctx := context.Background()
+	testCases := []struct {
```
thrivikram-karur-g, line 631: @kislaykishore , do you think we should pull this struct out and reuse named struct everywhere without this repetition or is it fine in testcases write up ?

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,1406 @@
...
+				Parent: fuseops.RootInodeID,
+				Name:   symlinkName,
+				Target: target,
+			}
+			err := m.CreateSymlink(ctx, createSymlinkOp)
+			require.NoError(t, err)
+			op := &fuseops.ReadSymlinkOp{
+				Inode: createSymlinkOp.Entry.Child,
+			}
+			err = m.ReadSymlink(ctx, op)
```
kislaykishore, line 1067: Add a newline before this to segregate the ACT and ARRANGE test sections. Please do the same for other tests.

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,1406 @@
...
+			ctx := context.Background()
+			fileName := "test.txt"
+			content := "test content"
+			createWithContents(ctx, t, bucket, fileName, content)
+			statFsOp := &fuseops.StatFSOp{}
+			err := m.StatFS(context.Background(), statFsOp)
+			require.NoError(t, err)
+
+			ss := ex.GetSpans()
+			require.Len(t, ss, len(tt.spans))
```
thrivikram-karur-g, line 148: @kislaykishore , this is also very repetitive, should we create a verifyTraces function or similar ?

internal/fs/tracing_test.go
kislaykishore: [Don't worry about DRY](https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html). It's okay to repeat in tests.

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,1406 @@
...
+	}
+	server, err := fs.NewFileSystem(ctx, serverCfg)
+	require.NoError(t, err, "NewFileSystem")
+	return bucket, server
+}
+
+func newInMemoryExporter(t *testing.T) *tracetest.InMemoryExporter {
+	t.Helper()
+	ex := tracetest.NewInMemoryExporter()
+	otel.SetTracerProvider(sdktrace.NewTracerProvider(sdktrace.WithSyncer(ex)))
```
kislaykishore, line 82: This will take care of cleaning up traces for all tests. With this, you can remove the cleanup statement in every test.
```suggestion
	ex := tracetest.NewInMemoryExporter()
	t.Cleanup(func() {
		ex.Reset()
	})
	otel.SetTracerProvider(sdktrace.NewTracerProvider(sdktrace.WithSyncer(ex)))
```

internal/fs/tracing_test.go
```diff
@@ -0,0 +1,1406 @@
...
+				ex.Reset()
+			})
+			bucket, server := createTestFileSystemWithTraces(ctx, t, tt.ignoreInterrupts)
+			m := wrappers.WithTracing(server)
+			ctx := context.Background()
+			fileName := "test.txt"
+			content := "test content"
+			createWithContents(ctx, t, bucket, fileName, content)
+			statFsOp := &fuseops.StatFSOp{}
+			err := m.StatFS(context.Background(), statFsOp)
```
kislaykishore, line 144: Add a space before this line. Same for other tests.

internal/fs/tracing_test.go
thrivikram-karur-g: cool. keeping it as it is then

internal/fs/tracing_test.go
thrivikram-karur-g: based on above discussion, keeping this as well without DRY

internal/fs/tracing_test.go
thrivikram-karur-g: updated to segregate AAA(arrange, act, assert) for all tests.

internal/fs/tracing_test.go
thrivikram-karur-g: updated to segregate AAA(arrange, act, assert) for all tests.

internal/fs/tracing_test.go
thrivikram-karur-g: makes sense, removed the test cleanup from individual tests and added it to newInMemoryExporter

PR #3945, opened 2025-10-28
ci(orbax): Fix orbax benchmark script

kislaykishore:
### Description
* Remove build as a dependency on build-csi since the code is built as part of build-csi anyway. It also helps get rid of the additional requirement of installing golang and goimports while building the CSI driver.
* Install dependencies like kubectl, make etc. in case they aren't installed.
* Add a flag to skip the CSI driver build.
* Add a call to `gcloud container cluster get-credentials` to fetch the credentials of the GKE cluster.
* Remove `shutil.rmtree` call in the `build_gcsfuse_image` function since the TemporaryDirectory will get automatically cleaned up on exit.
b/453871259

### Testing details
1. Manual - yes
2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/perfmetrics/scripts/continuous_test/gke/orbax_benchmark/run_benchmark.py
+++ b/perfmetrics/scripts/continuous_test/gke/orbax_benchmark/run_benchmark.py
@@ -439,7 +439,7 @@ async def main():
     await check_prerequisites()
 
     timestamp = datetime.now().strftime("%Y%m%d%H%M%S")
-    with tempfile.TemporaryDirectory(delete=False) as temp_dir:
+    with tempfile.TemporaryDirectory() as temp_dir:
         try:
             if args.skip_csi_driver_build:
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/continuous_test/gke/orbax_benchmark/run_benchmark.py
```diff
@@ -85,25 +84,75 @@ async def check_prerequisites():
...
+                    await run_command_async(["sudo", "apt", "install", "-y", "make"])
+                    await run_command_async(version_cmd)
+                except (FileNotFoundError, subprocess.CalledProcessError) as install_e:
+                    print(f"Error: Failed to install make: {install_e}", file=sys.stderr)
+                    sys.exit(1)
             if tool == "kubectl":
-                print("kubectl not found. Attempting to install via gcloud components...")
+                print("kubectl not found. Attempting to install...")
                 try:
-                    await run_command_async(["gcloud", "components", "install", "kubectl"])
```
Tulsishah, line 103: Why did it not working? Mostly people do this only

kislaykishore: Not sure - I don't think it worked for me ever.

PR #3954, opened 2025-10-30
test(list large dir): limit concurrency to prevent resource exhaustion error during test setup

anushka567:
### Description
As part of ZB testing, we added non-zero wait duration after file close/sync in our e2e tests setup, this led to increased test runtime especially when the setup involved creating multiple files synchronously such as list_large_dir, to avoid this, it was execessively parallelised so that the wait duration per file creation is parallelised . However, due to this, the test was   failing with resource exhausted error : 
```
runtime: program exceeds 10000-thread limit
fatal error: thread exhaustion
```

To avoid the issue above as well as to avoid completely synchronised wait times after file creation, the PR limits the max concurrency to 1024 so that the overall setup time equated to ~ N_FILES/MAX_CONCURRENCY secs.

b/455432790

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/util/operations/dir_operations.go
+++ b/tools/integration_tests/util/operations/dir_operations.go
@@ -100,6 +100,10 @@ func CreateDirectoryWithNFiles(numberOfFiles int, dirPath string, prefix string,
 		t.Fatalf("Error in creating directory %q: %v", dirPath, err)
 	}
 
+	// Limit the maximum number of I/O goroutines that can run simultaneously.
+	const maxConcurrency = 1024
+	sem := make(chan struct{}, maxConcurrency)
+
 	// 2. Setup a WaitGroup to manage concurrent Go routines
 	var wg sync.WaitGroup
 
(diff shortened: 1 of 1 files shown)


tools/integration_tests/util/operations/dir_operations.go
```diff
@@ -100,6 +100,10 @@ func CreateDirectoryWithNFiles(numberOfFiles int, dirPath string, prefix string,
 		t.Fatalf("Error in creating directory %q: %v", dirPath, err)
 	}
 
+	// Limit the maximum number of I/O goroutines that can run simultaneously.
+	const maxConcurrency = 1024
```
Tulsishah, line 104: Why we chose 1024?

anushka567: Safe number to not touch the upper limit even if multiple test packages are run in parallel  as well as not increase the test runtime by a lot (every thread is responsible for handling 12k/#threads number of files, so inversely related )

PR #3957, opened 2025-10-31
fix: Handle SIGTERM irrespective of unmount status

ameoghosh:
### Description
When we are returning from the go routine for signal handler, instead of return we should os.Exit with background thread with sleep 30sec alongside unmount. so that unconditionally the process terminates and does not get stuck in the main thread (What is the suspicion for delay reproduced during internal tests go/gcsfuse-sigterm-handling).

b/449182561

### Testing details
1. Manual - [Test doc](http://goto.google.com/gcsfuse-sigterm-handling)
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -84,7 +84,7 @@ func registerTerminatingSignalHandler(mountPoint string) {
 			//On signal receive wait in background and give 30 second for unmount to finish
 			//and then exit, so application is closed.
 			go func() {
-				logger.Warnf("Received %s, waiting for %d sec to let system gracefully unmount before killing the process", sigName, WaitTimeOnSignalReceive)
+				logger.Warnf("Received %s, waiting for %s to let system gracefully unmount before killing the process", sigName, WaitTimeOnSignalReceive)
 				time.Sleep(WaitTimeOnSignalReceive)
 				logger.Warnf("killing goroutines and exit")
 				//Forcefully exit to 0 so that caller get success on forcefull exit also.


cmd/legacy_main.go
```diff
@@ -75,8 +76,17 @@ func registerTerminatingSignalHandler(mountPoint string) {
...
-			logger.Infof("Received %s, attempting to unmount...", sigName)
 
+			//on signal receive wait in background and give 30 second for unmount to finish
+			// and then exit, so application is closed
+			go func() {
+				logger.Warnf("Received %s, waiting for 30s to kill goroutines", sigName)
+				time.Sleep(30 * time.Second)
+				logger.Warnf("killing goroutines and exit")
+				os.Exit(0)
+			}()
```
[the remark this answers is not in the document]
ameoghosh: 30 second is standard KubeAPI server retry interval for heart beat checks, so timeout should not be more than 30 second. will move it to constant.

cmd/legacy_main.go
```diff
@@ -75,8 +77,18 @@ func registerTerminatingSignalHandler(mountPoint string) {
...
-			logger.Infof("Received %s, attempting to unmount...", sigName)
 
+			//on signal receive wait in background and give 30 second for unmount to finish
+			// and then exit, so application is closed
+			go func() {
+				logger.Warnf("Received %s, waiting for 30s to kill goroutines", sigName)
+				time.Sleep(WaitTimeOnSignalReceive)
+				logger.Warnf("killing goroutines and exit")
+				//forcefully exit to 0 so that caller
+				os.Exit(0)
```
charith87, line 91: exit(1) would it work as well as suggested by gemini?

Also, is there anyway to verify the csi is also handling this gracefully.
can we add @saikat-royc or someone from GKE team as a reviewer as well, so its fine.

saikat-royc: CSI gke-gcsfuse-sidecar looks at cmd.ProcessState.Exited() only (not any specific exit code e.g ProcessStart.ExitCode()). So this change should be good from CSI side. We need to validate the behavior through an IO usecase with the CSI driver

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/0663993b7ee2ac972293a75bca0e147ef3887e4c/pkg/sidecar_mounter/sidecar_mounter.go#L144

cmd/legacy_main.go
```diff
@@ -53,6 +54,7 @@ const (
 	SuccessfulMountMessage         = "File system has been successfully mounted."
 	UnsuccessfulMountMessagePrefix = "Error while mounting gcsfuse"
 	DynamicMountFSName             = "gcsfuse"
+	WaitTimeOnSignalReceive        = 30 * time.Second
```
kislaykishore, line 60: Can we make it unexported (make first character lowercase)?

cmd/legacy_main.go
```diff
@@ -75,8 +77,18 @@ func registerTerminatingSignalHandler(mountPoint string) {
...
 			}
-			logger.Infof("Received %s, attempting to unmount...", sigName)
 
+			//on signal receive wait in background and give 30 second for unmount to finish
+			// and then exit, so application is closed
+			go func() {
+				logger.Warnf("Received %s, waiting for 30s to kill goroutines", sigName)
+				time.Sleep(WaitTimeOnSignalReceive)
+				logger.Warnf("killing goroutines and exit")
+				//forcefully exit to 0 so that caller
```
kislaykishore: The comment seems incomplete - "so that caller ..."?

cmd/legacy_main.go
```diff
@@ -75,8 +77,18 @@ func registerTerminatingSignalHandler(mountPoint string) {
 			case os.Interrupt:
 				sigName = "SIGINT"
 			}
-			logger.Infof("Received %s, attempting to unmount...", sigName)
 
+			//on signal receive wait in background and give 30 second for unmount to finish
+			// and then exit, so application is closed
+			go func() {
+				logger.Warnf("Received %s, waiting for 30s to kill goroutines", sigName)
```
kislaykishore: Should we get rid of this message. Say even in cases where the system unmounts gracefully, we'll see two log messages which could be confusing:
* Received SIGTERM, waiting for 30s to kill goroutines
* Received SIGTERM, attempting to unmount...

I'll recommend we get rid of the two messages and instead have a single log saying:
"Received SIGTERM, attempting to gracefully unmount"

And when the timeout expires:
"GCSFuse didn't terminate within 30s. Doing a force-kill"

cmd/legacy_main.go
ameoghosh: Updated

cmd/legacy_main.go
ameoghosh: Two logs at the kill is needed for debug when unmount is stuck and process forcefull exit. In case of unmount success no guarantee on main thread will exit and hence sub go routine log will be usefull.
So these are not just extra logs but help to improve debugablity.

cmd/legacy_main.go
ameoghosh: for long term solution this constant will be needed where gracefull exit from other loop and epoll kill need to honour same timeout.

cmd/legacy_main.go
kislaykishore: My concern is more about the first log message - "waiting for 30s" suggests that the process isn't doing anything. Can we rephrase it to "waiting for 30s to let system gracefully unmount before killing the process". Also, can we move it out of the goroutine, to ensure that the ordering of the two log messages is consistent?

ameoghosh: Logs can be rephrased.

But i am not clear on moving out is clear ? how can we certain that go routine spwan runs at what time ? they are  never guaranteed to schedule if system is busy and we will have wrong time in logs if we take it out. Not an ideal way of adding trace point.

cmd/legacy_main.go
kislaykishore: We can make it an exported-field then. Usually, we start with a principle of least visibility.

cmd/legacy_main.go
kislaykishore: Makes sense.

cmd/legacy_main.go
```diff
@@ -78,8 +80,18 @@ func registerTerminatingSignalHandler(mountPoint string) {
...
 				sigName = "SIGINT"
 			}
-			logger.Infof("Received %s, attempting to unmount...", sigName)
 
+			//on signal receive wait in background and give 30 second for unmount to finish
+			// and then exit, so application is closed
+			go func() {
+				logger.Warnf("Received %s, waiting for %d sec to let system gracefully unmount before killing the process", sigName, WaitTimeOnSignalReceive)
+				time.Sleep(WaitTimeOnSignalReceive)
+				logger.Warnf("killing goroutines and exit")
```
ashmeenkaur, line 89: NIT:
```suggestion
				logger.Warnf("Killing in progress operations and forcing exit.")
```

cmd/legacy_main.go
```diff
@@ -78,8 +80,18 @@ func registerTerminatingSignalHandler(mountPoint string) {
 			case os.Interrupt:
 				sigName = "SIGINT"
 			}
-			logger.Infof("Received %s, attempting to unmount...", sigName)
 
+			//on signal receive wait in background and give 30 second for unmount to finish
+			// and then exit, so application is closed
+			go func() {
+				logger.Warnf("Received %s, waiting for %d sec to let system gracefully unmount before killing the process", sigName, WaitTimeOnSignalReceive)
```
ashmeenkaur: I think WaitTimeOnSignalReceive will have the value in nano-second so we should remove the seconds multiplier from the constant.

cmd/legacy_main.go
```diff
@@ -78,8 +80,18 @@ func registerTerminatingSignalHandler(mountPoint string) {
 			case os.Interrupt:
 				sigName = "SIGINT"
 			}
-			logger.Infof("Received %s, attempting to unmount...", sigName)
 
+			//on signal receive wait in background and give 30 second for unmount to finish
+			// and then exit, so application is closed
```
ashmeenkaur: Nit: we usually start comments with capital letter and end them with a full stop 😅

ashmeenkaur (commented):
LGTM

cmd/legacy_main.go
ameoghosh: Thats not right, constant is declare with Seconds as WaitTimeOnSignalReceive        = 30 * time.Second

cmd/legacy_main.go
ameoghosh: Done :D

ashmeenkaur (approved):

cmd/legacy_main.go
kislaykishore: @ashmeenkaur is right - duration is basically an integer - https://go.dev/play/p/8IMJCH_8Wqa

ameoghosh: fix the representation, it has humanize internally - https://go.dev/play/p/eqxQ9pzWMtY

PR #3976, opened 2025-11-05
feat(Unsupported Directory fix): Handle scenarios for unsupported objects(e.g. "." , "..") 

Tulsishah:
### Description
- The objects "." and ".." on GCS can remain as objects rather than being treated as directories. This PR addresses these scenarios to prevent operation failures.
- https://screenshot.googleplex.com/9rHRnCW9zK9gR2L

[b457912799](https://b.corp.google.com/issues/457912799) 

### Testing details
1. Manual - Tested
2. Unit tests - Done

### DIFF
--- a/internal/storage/storageutil/unsupported_object_util.go
+++ b/internal/storage/storageutil/unsupported_object_util.go
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	unsupportedObjectNameSubstrings = []string{"//", "..", "/./"}
+	unsupportedObjectNameSubstrings = []string{"//", "/../", "/./"}
 	unsupportedObjectNamePrefixes   = []string{"/"}
 	unsupportedObjectNameSuffix     = []string{"/.", "/.."}
 	unsupportedObjectNames          = []string{"", ".", ".."}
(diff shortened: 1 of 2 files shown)


internal/fs/inode/dir.go
```diff
@@ -709,6 +709,15 @@ func (d *dirInode) readObjects(
 	}()
 
 	for _, o := range listing.MinObjects {
+		if storageutil.IsUnsupportedObjectName(o.Name) {
+			unsupportedDirs = append(unsupportedDirs, o.Name)
```
Tulsishah, line 716: For reviewer: I am storing unsupported objects as part of the unsupportedDirs array. I initially believed that only directories were possible, but objects like "." and ".." can also be created as files. Should I rename this array and all the variables to a more general term?

vadlakondaswetha: yes

vadlakondaswetha: We should rename the flag tool

internal/storage/storageutil/unsupported_object_util.go
```diff
@@ -22,6 +22,7 @@ import (
 var (
 	unsupportedObjectNameSubstrings = []string{"//", "..", "/./"}
```
vadlakondaswetha: what about object name /a/b/c..d/e -> is this supported or unsupported?

internal/fs/inode/dir.go
Tulsishah: Will raise separate PR to rename it at all places.

internal/storage/storageutil/unsupported_object_util.go
Tulsishah: It should be supported. Fixed and also added unit test.

Issue #3974, opened 2025-11-05
why ignore DeleteChildFile err


Peiqi0714:
<img width="980" height="860" alt="Image" src="https://github.com/user-attachments/assets/4b3f3deb-cede-47b0-9a7c-f1d2def0a3df" />
In func (fs *fileSystem) nonAtomicRename, the return err of DeleteChildFile is ignored, is it a bug?

anushka567:
Hi @Peiqi0714, thanks for raising this question! It's a great observation, and I appreciate you taking the time to carefully review the code.

The error from `DeleteChildFile` is captured in the err variable:
```
err = oldParent.DeleteChildFile(
	ctx,
	oldName,
	oldObject.Generation,
	&oldObject.MetaGeneration)

```
Next, we have another error check for `invalidateChildFileCacheIfExist`:
```
if err := fs.invalidateChildFileCacheIfExist(oldParent, oldObject.Name); err != nil {
	return fmt.Errorf("nonAtomicRename: while invalidating cache for delete file: %w", err)
}

```

This *if* statement uses a short variable declaration, so the `err` variable inside this *if* is a new variable that only exists within the scope of this *if* statement( shadowing ).

Finally, the original `err` variable from the `DeleteChildFile` call is checked and returned if it's not nil:
```
if err != nil {
	err = fmt.Errorf("DeleteChildFile: %w", err)
	return err
}

```

So, to summarize, the error from DeleteChildFile is indeed handled correctly.I hope this explanation helps clarify the code's logic. Please let me know if you have any more questions!

Peiqi0714:
@anushka567 Thanks for your answer, I get your point. But if DeleteChildFile returns err, it's not reasonable to call invalidateChildFileCacheIfExist. I think if DeleteChildFile returns err, we should return the function immediately, and no need to go to the next step.

Peiqi0714:
@anushka567 <img width="824" height="498" alt="Image" src="https://github.com/user-attachments/assets/0a1fd418-4b1b-4452-8d16-cda5d352c6ad" />  I think it's more reasonable and clearly after my modification. @anushka567

Peiqi0714:
<img width="901" height="900" alt="Image" src="https://github.com/user-attachments/assets/90b918e1-83bc-4864-8bab-49e5f8a08147" /> @anushka567 You need to see the code of `func (fs *fileSystem) Unlink`

anushka567:
[...]

PR #3977, opened 2025-11-05
fix(non atomic rename): handle different caches eviction appropriately based on error status of delete object operation

anushka567:
### Description
This PR is to handle the errors originating from deleting objects gracefully, by appropriately handling the cache eviction.

Delete succeeds :  Invalidate file cache and stat and type cache.

Delete returns precond error

- 	Precond error means metadata needs to be updated and the file cache cannot be used,
- 	Remove it from file cache
- 	Remove it from metadata (stat and type) cache

Delete returns any error which is not precondition error 

- 	In this case, can be that everything was valid but delete failed
- 	Metadata was valid, but delete failed -  so preserve the metadata cache
- 	Local file cache was valid but delete failed - so preserve the file cache

https://github.com/GoogleCloudPlatform/gcsfuse/issues/3974

### Testing details
1. Manual - Updated.
2. Unit tests - NA
3. Integration tests - Automated. Successful Run: [#14258](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/61612afb-087b-4be6-bc71-d7478a847bf1/summary)

### DIFF
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -402,11 +402,13 @@ func (b *fastStatBucket) DeleteObject(
 		b.addNegativeEntry(req.Name)
 		return
 	}
-	// If the delete failed due to a precondition error, invalidate the cache entry
-	// as the object's state is uncertain. For other errors, we don't touch the
-	// cache because the object likely still exists.
+	// If the delete failed due to a precondition error or not found error,
+	// invalidate the cache entry as the object's state is uncertain.
+	// For other errors, we don't touch the cache because the object likely
+	// still exists.
 	var preconditionErr *gcs.PreconditionError
-	if errors.As(err, &preconditionErr) {
+	var notFoundErr *gcs.NotFoundError
+	if errors.As(err, &preconditionErr) || errors.As(err, &notFoundErr) {
 		b.invalidate(req.Name)
 	}
 	return
--- a/internal/storage/caching/fast_stat_bucket_test.go
+++ b/internal/storage/caching/fast_stat_bucket_test.go
@@ -1009,7 +1009,7 @@ func (t *DeleteObjectTest) CallsWrapped() {
 	ExpectEq(name, wrappedReq.Name)
 }
 
-func (t *DeleteObjectTest) WrappedFails_NonPreconditionError() {
+func (t *DeleteObjectTest) WrappedFails_GenericError() {
 	const name = ""
 	var err error
 
@@ -1035,6 +1035,20 @@ func (t *DeleteObjectTest) WrappedReturnsPreconditionError() {
 	err := t.deleteObject(name)
 
 	ExpectThat(err, Error(HasSubstr("precondition failed")))
+}
+
+func (t *DeleteObjectTest) WrappedReturnsNotFoundError() {
+	const name = "taco"
+	// Erase
+	ExpectCall(t.cache, "Erase")(name)
+	// Wrapped
+	ExpectCall(t.wrapped, "DeleteObject")(Any(), Any()).
+		WillOnce(Return(&gcs.NotFoundError{Err: errors.New("object not found")}))
+
+	// Call.
+	err := t.deleteObject(name)
+
+	ExpectThat(err, Error(HasSubstr("object not found")))
 }
 
 func (t *DeleteObjectTest) WrappedSucceeds_AddsNegativeEntry() {


internal/fs/fs.go
```diff
@@ -2371,23 +2362,23 @@ func (fs *fileSystem) nonAtomicRename(
 	// Delete behind. Make sure to delete exactly the generation we cloned, in
 	// case the referent of the name has changed in the meantime.
 	oldParent.Lock()
+	defer oldParent.Unlock()
+
 	err = oldParent.DeleteChildFile(
```
ashmeenkaur: deleteChildFile can throw an error in case of stale object Generation or Metageneration (I think it will return ESTALE error in that case). 
IMO it is best to invalidate the entry in cache for those cases so follow up requests can succeed by not using the stale cached generation/metageneration. 

Can you also check if the above hypothesis is correct?

anushka567: I checked the above scenario.

Create a fresh object. Read the file via gcsfuse mount so file cache is populated.

Issue rename for the object. After the copying to destination succeeds, before the delete operation for source is issued, modify the object generation remotely.

The rename operation fails with I/O error. However reads against the "source" is served from the latest generation i.e. the updated data as the file cache is re populated with the latest data.

So i believe the file cache flow handles it implicitly, handling of file cache invalidation can be done after we have concluded that delete operation has succeeded properly.

anushka567: Checked the flows, whether or not the file cache is evicted, reads are always up to date because during the delete operation, the if it succeeds, we add entry to negative stat cache and if it fails, we invalidate metadata so further references made will result in look up calls which gets the latest data.

In any case, whether the file cache is invalidated or not , when a read call is made , the read manager stands invalidated as the minObj using which the read manager was created is staler than freshly updated inode's minobj. So in case the file cache was evicted or still present, the invalidation of the read manager triggers a fresh read manager creation which will be up-to-date in terms of generation/metageneration and the latest data will be served, so no references to stale generation.


That being said, I believe your point of keeping the MD cache and file cache in sync makes sense, which means invalidation regardless of the status of the delete operation. So, the change pending is to ensure the same happens in unlink() operation.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -397,10 +397,18 @@ func (b *fastStatBucket) DeleteObject(
...
+	// should not invalidate the stat cache, because the file might still exist.
+	if err != nil && !errors.As(err, &preconditionErr) {
+		return
+	}
+	// In case of successful delete or precondition error, we should invalidate the
+	// stat cache.
+	if err == nil {
 		b.addNegativeEntry(req.Name)
+	} else {
+		b.invalidate(req.Name)
```
abhishek10004, line 412: nit: can be cleaned slightly

if err == nil {
    addNegativeEntry...
} else if (precondition error) {
    invalidate...
}

return

internal/fs/inode/dir.go
```diff
@@ -977,7 +977,6 @@ func (d *dirInode) DeleteChildFile(
 	name string,
 	generation int64,
 	metaGeneration *int64) (err error) {
-	d.cache.Erase(name)
```
abhishek10004: Could you please add an UT for this as well?

anushka567: Updated

internal/fs/inode/dir.go
ashmeenkaur, line 1041: do you want to check precondition failure and delete the object conditionally from cache based on error?

internal/fs/fs.go
```diff
@@ -2374,21 +2374,25 @@ func (fs *fileSystem) nonAtomicRename(
...
 		&oldObject.MetaGeneration)
 
-	if err := fs.invalidateChildFileCacheIfExist(oldParent, oldObject.Name); err != nil {
-		return fmt.Errorf("nonAtomicRename: while invalidating cache for delete file: %w", err)
+	var preconditionErr *gcs.PreconditionError
+	// Do not invalidate the file cache in case of error while deleting file
+	// which is not precondition error.
+	if err != nil && !errors.As(err, &preconditionErr) {
+		return fmt.Errorf("DeleteChildFile: %w", err)
 	}
```
ashmeenkaur: precondition errors should also be returned from this function I believe based on the old flow but now we are returning nil in that case. 

Also can we simplify the code flow to explicitly remove file cache entry when error is nil or error is precondition failure to keep it similar to fast_stat_bucket.go?

Peiqi0714:
May I ask why this pr is still open? I think this pr is a fix to a potential bug.

internal/fs/inode/dir.go
anushka567: Since here we are concerned about type cache, It makes sense to evict if the "type" cached is no longer correct, an example of it would be when we have cached an object to be RegularFileType but then remotely it was updated to be ExplicitDirType. However, note that in this case, DeleteObject returns not found error.  I am planning to evict it based on this error , which when referenced to again, will cause a fresh stat to GCS and thus, lead to updated type .

Considering that type cache is soon to be deprecated, I'd suggest not blocking on this change .

Peiqi0714:
May i ask why type cache will be depracated
thanks！

发自我的iPhone


------------------ Original ------------------
From: anushka ***@***.***&gt;
Date: Wed,Dec 31,2025 0:44 AM
To: GoogleCloudPlatform/gcsfuse ***@***.***&gt;
Cc: Peiqi ***@***.***&gt;, Comment ***@***.***&gt;
Subject: Re: [GoogleCloudPlatform/gcsfuse] fix(non atomic rename): handledifferent caches eviction appropriately based on error status of delete objectoperation (PR #3977)





 
@anushka567 commented on this pull request.
 
 
In internal/fs/inode/dir.go:
 &gt;   
Since here we are concerned about type cache, It makes sense to evict if the "type" cached is no longer correct, an example of it would be when we have cached an object to be RegularFileType but then remotely it was updated to be ExplicitDirType. However, note that in this case, DeleteObject returns not found error.  I am planning to evict it based on this error , which when referenced to again, will cause a fresh stat to GCS and thus, lead to updated type .
 
Considering that type cache is soon to be deprecated, I'd suggest not blocking on this change .

anushka567:
Please note, Type cache deprecation is still in POC phase. The reason for the same is to maintain a single cache for metadata related operations in order to avoid scenarios of inconsistency between stat and type cache.

internal/fs/inode/dir_test.go
```diff
@@ -1547,6 +1547,48 @@ func (t *DirTest) CreateChildSymlink_TypeCaching() {
...
+	t.in.InsertFileIntoTypeCache(name)
+	AssertEq(metadata.RegularFileType, t.getTypeFromCache(name))
+
+	// Call the inode with a meta-generation that will cause a precondition error.
+	wrongMetaGeneration := o.MetaGeneration + 1
+	err = t.in.DeleteChildFile(t.ctx, name, o.Generation, &wrongMetaGeneration)
+
+	ExpectThat(err, Error(HasSubstr("DeleteObject: gcs.PreconditionError")))
+	AssertEq(metadata.RegularFileType, t.getTypeFromCache(name))
+}
```
abhishek10004, line 1590: Let's add a test when DeleteObject returns NotFound error as well

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -397,10 +397,17 @@ func (b *fastStatBucket) DeleteObject(
...
+	// In case of successful delete, add a negative entry to the cache.
+	if err == nil {
 		b.addNegativeEntry(req.Name)
+		return
+	}
+	// If the delete failed due to a precondition error, invalidate the cache entry
+	// as the object's state is uncertain. For other errors, we don't touch the
+	// cache because the object likely still exists.
+	var preconditionErr *gcs.PreconditionError
+	if errors.As(err, &preconditionErr) {
```
abhishek10004: What about not found error?

abhishek10004:
Looks good. Have left a few minor comments. Please check.

internal/fs/inode/dir_test.go
anushka567: Since we are using fake bucket here in the unit test, this situation cannot be simulated. 

One option would be to delete it twice so that the second delete would throw not found error but the first successful delete itself would evict it from type cache.

internal/storage/caching/fast_stat_bucket.go
anushka567: For not found error, it makes sense to add an entry in the negative stat cache instead of just invalidating. For now, I am going to keep the behavior same as previous implementation. There is some disparity in folder deletes and object deletes due to which it was kept as such, currently an [open problem](https://b.corp.google.com/issues/351945743#comment3), can be taken as a follow up as to when to add negative entry.

PR #3978, opened 2025-11-05
feat(Unsupported Directory fix):  E2e tests for unsupported Paths

Tulsishah:
### Description
1. Adding e2e tests for unsupported paths scenario for different operations like list, copy, rename and delete.
2. Setting includeFolderAsPrefix flag as while making a list call, the includeFolderAsPrefix parameter must be set to true. This is because the HNS/Rapid bucket only displays folders as part of the CollapsedRuns output when this flag is enabled.

[b/455706503](https://b.corp.google.com/issues/455706503)

### Testing details
1. Manual - NA
3. Unit tests - NA

### DIFF
--- a/tools/integration_tests/unsupported_path/unsupported_path_test.go
+++ b/tools/integration_tests/unsupported_path/unsupported_path_test.go
@@ -137,7 +137,7 @@ func (s *UnsupportedPathSuite) TestCopyDirWithUnsupportedPaths() {
 	// List the contents of the destination directory in the GCS bucket (to check actual objects created).
 	entries, err := client.ListDirectory(ctx, storageClient, setup.TestBucket(), destBucketPath)
 	require.NoError(s.T(), err, "Listing the destination directory in GCS should succeed.")
-	// Verify that only supported objects were copied (5 objects).
+	// Verify that only supported objects were copied (3 objects).
 	assert.Len(s.T(), entries, 3, "The number of copied objects should only match supported objects.")
 	s.Assert().ElementsMatch(expectedObjectNames, entries, "The copied object names must match the expected supported names.")
 }


tools/integration_tests/test_config.yaml
```diff
@@ -576,3 +576,24 @@ flag_optimizations:
...
+      flat: true
+      hns: true
+      zonal: false
+    run_on_gke: true
+  - flags:
+    - "--implicit-dirs --enable-unsupported-path-support=true --rename-dir-limit=200 --metadata-cache-negative-ttl-secs=0"
+    compatible:
+      flat: false
+      hns: false
+      zonal: true
```
ashmeenkaur, line 621:
```suggestion
  - flags:
    - "--implicit-dirs --client-protocol=grpc --enable-unsupported-path-support=true --rename-dir-limit=200 --metadata-cache-negative-ttl-secs=0"
    compatible:
      flat: true
      hns: true
      zonal: false
    run_on_gke: true
  - flags:
    - "--implicit-dirs --enable-unsupported-path-support=true --rename-dir-limit=200 --metadata-cache-negative-ttl-secs=0"
    compatible:
      flat: true
      hns: true
      zonal: true
```

on side note, what is the harm in testing both with zonal?

tools/integration_tests/unsupported_path/unsupported_path_test.go
```diff
@@ -0,0 +1,208 @@
...
+	require.NoError(s.T(), os.RemoveAll(path.Join(s.testDir, testDirName)))
+}
+
+func (s *UnsupportedPathSuite) TearDownSuite() {
+	require.NoError(s.T(), os.RemoveAll(s.testDir))
+}
+
+// createTestObjects populates the GCS bucket with objects having supported and unsupported names.
+func (s *UnsupportedPathSuite) createTestObjects() {
+	// Define objects with names that contain characters or sequences not supported by POSIX file systems.
```
ashmeenkaur, line 66: add t.Helper()

tools/integration_tests/test_config.yaml
Tulsishah: We can, but for zonal, we are already setting the gRPC client internally. Therefore, I felt there was no point in explicitly testing with an external gRPC client.

tools/integration_tests/unsupported_path/unsupported_path_test.go
```diff
@@ -0,0 +1,192 @@
...
+		// Moved unsupported objects
+		destBucketPath + "//",
+		destBucketPath + "/./",
+		destBucketPath + "/../",
+		destBucketPath + "/.",
+		destBucketPath + "/..",
+		// Moved supported objects
+		destBucketPath + "/supported_file.txt",
+		destBucketPath + "/supported_dir" + "/",
+	}
```
ashmeenkaur, line 172: shall we do an exact match with the unsupported objects declared above on line 68. This would also mean not using the '/' delimiter in the list call

tools/integration_tests/unsupported_path/unsupported_path_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	err := operations.RenameDir(sourceLocalPath, destLocalPath)
```
ashmeenkaur, line 174: What happens if we try to rename a directory containing nested unsupported directories?
for eg:
```
A/
  file1.txt 
  unsupportedDir1//  
        file2.txt
        unsupportedDir2//
               file3.txt
```    
Will it work for flat buckets and rename everything under A to B if tried via GCSFuse? Can we change the test objects to include nested unsupported objects?

Tulsishah: It will rename entire structure in flat bucket also. Also added structure in test.

tools/integration_tests/unsupported_path/unsupported_path_test.go
```diff
@@ -0,0 +1,201 @@
...
+		destBucketPath + "/unsupportedDir1/",
+	}
+
+	err := operations.CopyDir(sourceLocalPath, destLocalPath)
+
+	require.NoError(s.T(), err, "CopyDir operation should succeed.")
+	// List the contents of the destination directory in the GCS bucket (to check actual objects created).
+	entries, err := client.ListDirectory(ctx, storageClient, setup.TestBucket(), destBucketPath)
+	require.NoError(s.T(), err, "Listing the destination directory in GCS should succeed.")
+	// Verify that only supported objects were copied (5 objects).
```
ashmeenkaur:
```suggestion
	// Verify that only supported objects were copied (3 objects).
```

PR #3982, opened 2025-11-06
feat(Unsupported Directory fix):  Rename the flag

Tulsishah:
### Description
Rename the flag from unsupportedDir to unsupportedPath to use the more general term, as discussed in [this](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3976#discussion_r2493137196) thread.

[b/458222949](https://b.corp.google.com/issues/458222949)

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -802,7 +802,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.BoolP("enable-streaming-writes", "", true, "Enables streaming uploads during write file operation.")
 
-	flagSet.BoolP("enable-unsupported-path-support", "", false, "Enables support for unsupported files/directory fix implementation.")
+	flagSet.BoolP("enable-unsupported-path-support", "", false, "Enables support for file system paths with unsupported GCS names (e.g., names containing '//' or starting with /).  When set, GCSFuse will ignore these objects during listing and copying operations.  For rename and delete operations, the flag allows the action to proceed for all specified objects, including those with unsupported names.")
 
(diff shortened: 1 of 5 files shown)


cfg/config.go
```diff
@@ -802,9 +802,9 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.BoolP("enable-streaming-writes", "", true, "Enables streaming uploads during write file operation.")
 
-	flagSet.BoolP("enable-unsupported-dir-support", "", false, "Enables support for un-supported directory fix implementation.")
+	flagSet.BoolP("enable-unsupported-path-support", "", false, "Enables support for unsupported files/directory fix implementation.")
```
vadlakondaswetha: the description doesn't sound right

internal/fs/inode/dir.go
```diff
@@ -712,11 +712,11 @@ func (d *dirInode) readObjects(
 	}()
 
 	for _, o := range listing.MinObjects {
-		if storageutil.IsUnsupportedObjectName(o.Name) {
-			unsupportedDirs = append(unsupportedDirs, o.Name)
+		if storageutil.IsUnsupportedPathName(o.Name) {
```
vadlakondaswetha: call it isUnsupportedPath() or isUnsupportedObjectOrDirectory

cfg/config.go
Tulsishah: Updated.

vadlakondaswetha (approved):

PR #3983, opened 2025-11-06
feat(visualizer): adding functionality to merge based on threshold

raj-prince:
### Description
While analyzing vLLM read-pattern, we observed small IO sequential read with minor gaps. Adding a threshold config which will be used to merge two consecutive read ranges if gaps are less than the threshold. This will be helpful to know if read becomes sequential or not if we ignore some gaps b/w the reads.

GCSFuse default reader (or buffered reader) treats such kind of reads as sequential.

### Testing details
1. Manual - Yes.
2. Unit tests - yes.
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/read_manager/visual_read_manager.go
+++ b/internal/gcsx/read_manager/visual_read_manager.go
@@ -41,8 +41,11 @@ type VisualReadManager struct {
 	// Guards access to readIOs slice.
 	mu sync.Mutex
 
-	// Config for workload insight visualization.
+	// Configuration for workload insight visualization.
 	cfg cfg.WorkloadInsightConfig
+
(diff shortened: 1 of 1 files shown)


cfg/params.yaml
```diff
@@ -974,6 +974,17 @@ params:
     default: 1
     hide-flag: true
 
+  - config-path: "workload-insight.forward-merge-threshold-mb"
+    flag-name: "workload-insight-forward-merge-threshold-mb"
+    type: "int"
+    usage: >-
+      The threshold in MB for merging forward sequential reads for workload
```
kislaykishore, line 981:
```suggestion
      The threshold in MiB for merging forward sequential reads for workload
```

raj-prince: Will take in the next PR.

PR #3991, opened 2025-11-08
Implement sparse file partial download for random I/O optimization

xinyangge:
## Summary

This PR implements sparse file support for the GCSFuse file cache, enabling efficient random read operations by downloading only the requested byte ranges instead of entire files. This significantly improves performance for workloads with random access patterns on large files.

### Key Features

- **Sparse File Mode**: Files are cached as sparse files with on-demand chunk downloads (1MB chunks)
- **ByteRangeMap Tracking**: Efficient chunk-based tracking of downloaded byte ranges
- **Cache Hit Optimization**: Random reads served from cache in microseconds instead of milliseconds
- **LRU Eviction**: Proper size accounting ensures cache stays within configured limits

### Performance Impact

- Random reads to cached chunks: **~93µs** (vs ~37ms for GCS reads)
- Cache hit rate: Improves significantly for random I/O workloads
- Memory efficient: Only downloaded chunks count toward cache size limit

### Implementation Details

1. **ByteRangeMap** (`internal/cache/data/byte_range.go`): New data structure for tracking downloaded 1MB chunks in sparse files
2. **DownloadRange** (`internal/cache/file/downloader/job.go`): On-demand chunk download for random reads
3. **Cache Validation Fix**: Skip offset checks for sparse files since they track individual chunks, not contiguous ranges
4. **LRU Cache Fix**: Properly erase old entries before updating to ensure accurate size tracking

### Configuration

```yaml
file-cache:
  enable-sparse-file: true  # Enable sparse file mode (default: false)
```

### Bug Fixes Included

- Fixed a race condition when multiple threads are accessing the same file descriptor (CacheHandle in GCSFuse) concurrently (while it is being evicted at the same time).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/integration_tests/sparse_file/sparse_file_test.go
+++ b/tools/integration_tests/sparse_file/sparse_file_test.go
@@ -80,6 +80,8 @@ func (s *sparseFileTest) TearDownSuite() {
 }
 
 func TestSparseFile(t *testing.T) {
+	// TODO: Enable these tests using conditional t.SkipNow() if integrationTest flag is not set.
+	t.SkipNow()
 	suite.Run(t, new(sparseFileTest))
 }
 


ashmeenkaur (commented):
The diff on the config file is huge. Could you run "make" command on the PR and push the changes again? Please also sign the Google CLA - https://github.com/GoogleCloudPlatform/gcsfuse/pull/3991/checks?check_run_id=54890072877.

internal/cache/file/cache_handle.go
```diff
@@ -194,9 +195,52 @@ func (fch *CacheHandle) Read(ctx context.Context, bucket gcs.Bucket, object *gcs
 		requiredOffset = objSize
 	}
 
-	// If fileDownloadJob is not nil, it's better to get status of cache file
-	// from the job itself than to use file info cache.
-	if fch.fileDownloadJob != nil {
+	// Handle sparse file reads
+	if fileInfoData.SparseMode {
```
vadlakondaswetha, line 201: can we move all this logic into a seperate file called sparse_downloader_job.go similar to parallel_downloads_job.go, that way its become easier to maintain and understand.

internal/cache/file/cache_handle.go
```diff
@@ -194,9 +195,52 @@ func (fch *CacheHandle) Read(ctx context.Context, bucket gcs.Bucket, object *gcs
...
+		// For sparse files, check if the requested range is downloaded
+		if !fileInfoData.DownloadedRanges.ContainsRange(uint64(offset), uint64(requiredOffset)) {
+			// Calculate the chunk to download based on the configured chunk size
+			chunkSizeMb := fch.fileDownloadJob.SequentialReadSizeMb()
```
vadlakondaswetha: The default value of SequentialReadSizeMB is 200MB.
in the worst case we end up downloading 400MB of data when the offsets fall in the last 1MB of 200MB chunk.

Can we create a seperate flag to handle the chunk Size for sparse downloads? The goal is to provide best default values as much as possible without having the user to change the values.

internal/cache/file/downloader/job.go
```diff
@@ -403,6 +408,85 @@ func (job *Job) createCacheFile() (*os.File, error) {
 	return cacheFile, err
 }
 
+// DownloadRange downloads a specific byte range [start, end) from the GCS object
+// for sparse file support. It writes the data to the cache file at the appropriate
+// offset using O_DIRECT.
```
vadlakondaswetha: you are not using o_direct to write

internal/cache/file/downloader/job.go
```diff
@@ -403,6 +408,85 @@ func (job *Job) createCacheFile() (*os.File, error) {
...
+			ReadCompressed: job.object.HasContentEncodingGzip(),
+			ReadHandle:     nil,
+		})
+	if err != nil {
+		return fmt.Errorf("DownloadRange: error creating reader for range [%d, %d): %w", start, end, err)
+	}
+	defer newReader.Close()
+
+	// Open cache file for writing
+	cacheFile, err := os.OpenFile(job.fileSpec.Path, os.O_WRONLY, job.fileSpec.FilePerm)
```
vadlakondaswetha: if we are using the default value of sequential-read-size-mb flag which is 200MB, we are waiting for entire data to be written before serving to the application.
What we do incase of non-parallel downloads is notify the caller after every 8MB of data is written: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/cache/file/downloader/job.go#L307 We can either do the same thing here or reduce the chunk size.

internal/cache/data/byte_range.go
```diff
@@ -0,0 +1,180 @@
...
+	End   uint64 // exclusive
+}
+
+// ByteRangeMap tracks which chunk-aligned byte ranges have been downloaded in a sparse file.
+// It uses fixed-size chunks (1MB by default) for simplified tracking and assumes all downloads
+// are aligned to chunk boundaries.
+type ByteRangeMap struct {
+	mu        sync.RWMutex
+	chunkSize uint64
+	chunks    map[uint64]bool // chunk ID -> downloaded
```
vadlakondaswetha, line 35: why is the chunksize not same as the single chunk we are downloading from GCS. why is it 1MB?

internal/cache/file/cache_handle.go
xinyangge: Moved.

internal/cache/file/cache_handle.go
xinyangge: I actually think having too many knobs are confusing.  For example, we already have `file-cache-download-chunk-size-mb`, and adding a new knob specific to sparse file caching seems redundant.  My latest change reuses `file-cache-download-chunk-size-mb` for sparse file cache too.

internal/cache/file/downloader/job.go
xinyangge: I updated the comment to be consistent.

internal/cache/file/downloader/job.go
xinyangge: I prefer keeping it blocking given this feature is all about fetching one chunk at a time.

internal/cache/data/byte_range.go
xinyangge: Good catch.  I updated the implementation to make the chunk size consistent with `file-cache-download-chunk-size-mb`.

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+//
+// Returns:
+// - cacheHit: whether the range is present in the downloaded ranges
+// - err: any error that occurred while fetching the FileInfo
+func (job *Job) verifySparseRangeDownloaded(offset, requiredOffset int64) (cacheHit bool, err error) {
+	// Create file info key
+	fileInfoKey := data.FileInfoKey{
+		BucketName: job.bucket.Name(),
+		ObjectName: job.object.Name,
+	}
```
vadlakondaswetha: can we call getFileInfo() method here

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+	fileInfo := fileInfoVal.(data.FileInfo)
+	return fileInfo.SparseMode, nil
+}
+
+// DownloadRange downloads a specific byte range [start, end) from the GCS object
+// for sparse file support. It writes the data to the cache file at the appropriate
+// offset.
+//
+// Acquires and releases LOCK(job.mu)
+func (job *Job) DownloadRange(ctx context.Context, start, end uint64) error {
```
vadlakondaswetha: can we make this. method non-public

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+	cacheHit = fileInfo.DownloadedRanges != nil && fileInfo.DownloadedRanges.ContainsRange(uint64(offset), uint64(requiredOffset))
+
+	logger.Tracef("Sparse file cache hit check: offset=%d, requiredOffset=%d, DownloadedRanges=%v, cacheHit=%t",
+		offset, requiredOffset, fileInfo.DownloadedRanges != nil, cacheHit)
+
+	return cacheHit, nil
+}
+
+// IsSparseFile returns true if the object is configured for sparse file mode.
+func (job *Job) IsSparseFile(bucket gcs.Bucket, object *gcs.MinObject) (bool, error) {
```
vadlakondaswetha: who is using this method? i didnt find any references

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+// HandleSparseRead manages the download and validation of sparse file ranges.
+// It checks if the requested range is already downloaded, calculates chunk
+// boundaries, downloads missing chunks, and validates the cache hit status.
+//
+// Returns:
+// - cacheHit: true if the requested range is available in cache, false if fallback to GCS is needed
+// - error: non-nil if there was an error during the operation
+//
+// Note: The FileInfo in the cache is automatically updated by DownloadRange.
+func (job *Job) HandleSparseRead(ctx context.Context, offset, requiredOffset int64) (cacheHit bool, err error) {
```
vadlakondaswetha: in the context of method & file can we call offset and requiredOffset as startOffset, endOffest everywhere in this file

internal/cache/file/downloader/sparse_downloads_job_test.go
```diff
@@ -0,0 +1,304 @@
...
+			expectError:    false,
+		},
+		{
+			name:           "chunk end capped at object size",
+			offset:         90 * util.MiB,
+			requiredOffset: 95 * util.MiB,
+			chunkSizeMb:    20,
+			objectSize:     100 * util.MiB,
+			expectedStart:  80 * util.MiB,
+			expectedEnd:    100 * util.MiB, // capped at object size, not 120MB
```
vadlakondaswetha: with the current logic if the objectSize is 120MB, dont we end up downloading 2 chunks 80-100Mb and 100-120MB when we need only one chunk: 80-100MB.
Can we compute the end chunk as follows:
if requiredOffset%chunkSize > 0 
endChunk =( requiredOffset/chunkSize)+1

internal/cache/data/byte_range.go
```diff
@@ -0,0 +1,183 @@
...
+// ByteRangeMap tracks which chunk-aligned byte ranges have been downloaded in a sparse file.
+// The chunk size should match the actual download chunk size for efficient tracking.
+type ByteRangeMap struct {
+	mu        sync.RWMutex
+	chunkSize uint64
+	chunks    map[uint64]bool // chunk ID -> downloaded
+}
+
+// NewByteRangeMap creates a new empty ByteRangeMap with the specified chunk size.
+// The chunkSize should match the download chunk size (e.g., DownloadChunkSizeMb * 1MB).
```
vadlakondaswetha, line 39: Please remove all unused method from this file

internal/cache/file/downloader/sparse_downloads_job_test.go
```diff
@@ -0,0 +1,304 @@
...
+			AssertNe(nil, err, fmt.Sprintf("Test case %q: expected error but got none", tt.name))
+		} else {
+			AssertEq(nil, err, fmt.Sprintf("Test case %q: unexpected error: %v", tt.name, err))
+			AssertEq(tt.expectedStart, chunkStart, fmt.Sprintf("Test case %q: chunkStart mismatch", tt.name))
+			AssertEq(tt.expectedEnd, chunkEnd, fmt.Sprintf("Test case %q: chunkEnd mismatch", tt.name))
+		}
+	}
+}
+
+func (dt *sparseDownloaderTest) Test_DownloadRange() {
```
vadlakondaswetha, line 140: Please follow the following format for unit tests
Arrange
--new line
Act
--newline
Assert

no new lines inbetwen unless the logic is too complicated

tools/integration_tests/sparse_file/sparse_file_test.go
```diff
@@ -0,0 +1,243 @@
...
+func (s *sparseFileTest) TearDownSuite() {
+	setup.UnmountGCSFuseAndDeleteLogFile(setup.TestDir())
+}
+
+func TestSparseFile(t *testing.T) {
+	suite.Run(t, new(sparseFileTest))
+}
+
+// Test that random reads on a large file only download the necessary chunks
+func (s *sparseFileTest) TestRandomReadsDownloadOnlyRequiredChunks() {
```
vadlakondaswetha, line 89: From read functionality wise these are good testcases. But we dont know if only the required chunk is downloaded or full file or served directly from GCS. Good thing would be to test by reading the file from cached directory
Example: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/read_cache/helpers_test.go#L136

internal/cache/data/byte_range_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+	assert.Equal(t, uint64(2*MB), brm.TotalBytes())
+
+	brm.Clear()
+
+	assert.Equal(t, uint64(0), brm.TotalBytes())
+	assert.Empty(t, brm.Ranges())
+}
+
+func TestByteRangeMap_ConcurrentAccess(t *testing.T) {
```
raj-prince, line 252: nit: Add test timeout for easy debug in case of deadlock.

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return data.FileInfo{}, fmt.Errorf("error creating fileInfoKeyName: %w", err)
+	}
+
+	fileInfoVal := job.fileInfoCache.LookUpWithoutChangingOrder(fileInfoKeyName)
+	if fileInfoVal == nil {
+		return data.FileInfo{}, fmt.Errorf("file info not found in cache")
+	}
+
+	return fileInfoVal.(data.FileInfo), nil
```
raj-prince: Should we do type assertion and return error if not of correct type?

```
fileInfoData, ok := fileInfo.(data.FileInfo)
	if !ok {
		return nil, fmt.Errorf("getFileInfoData: wrong type")
	}
```

internal/cache/lru/lru.go
```diff
@@ -269,6 +269,27 @@ func (c *Cache) UpdateWithoutChangingOrder(
 	return nil
 }
 
+// UpdateSize updates the currentSize accounting when an entry's size has changed.
+// This is needed for entries whose size grows incrementally (e.g., sparse files).
+// Eviction is deferred until the next Insert() call.
+// The entry's order in the LRU is not changed.
+func (c *Cache) UpdateSize(key string, sizeDelta uint64) error {
```
raj-prince, line 276: Can we add unit test for this?

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+	if end > job.object.Size {
+		end = job.object.Size
+	}
+
+	// Check if this is a sparse file
+	fileInfoKey := data.FileInfoKey{
+		BucketName: job.bucket.Name(),
+		ObjectName: job.object.Name,
+	}
+	fileInfoKeyName, err := fileInfoKey.Key()
```
raj-prince: Can we move this initialization just before the use?

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+	// Update FileInfo with downloaded range
+	job.mu.Lock()
+	defer job.mu.Unlock()
+
+	// Re-fetch FileInfo in case it changed
+	fileInfoVal := job.fileInfoCache.LookUpWithoutChangingOrder(fileInfoKeyName)
+	if fileInfoVal == nil {
+		return fmt.Errorf("DownloadRange: file info not found in cache after download")
+	}
+	fileInfo := fileInfoVal.(data.FileInfo)
```
raj-prince: nit: plz ensure type safety check return error otherwise.

internal/cache/file/downloader/sparse_downloads_job.go
raj-prince: +1. Also, need to be clear about the requiredOffset should be the inclusive one, otherwise can trigger one extra chunk download.

internal/cache/file/downloader/sparse_downloads_job.go
```diff
@@ -0,0 +1,255 @@
...
+
+	// Get the configured chunk size from fileCacheConfig
+	chunkSize := uint64(job.fileCacheConfig.DownloadChunkSizeMb) * 1024 * 1024
+
+	// Align chunk start and end to chunk boundaries
+	// Start: round down to chunk boundary
+	chunkStart = (uint64(offset) / chunkSize) * chunkSize
+
+	// End: round up requiredOffset to chunk boundary to ensure full coverage
+	chunkEnd = ((uint64(requiredOffset) + chunkSize - 1) / chunkSize) * chunkSize
```
vadlakondaswetha: requiredOffset is the exclusive end to download from GCS. We need to do -1 to avoid downloadng extra chunk. We can pass it from the caller itself.

internal/cache/file/cache_handler.go
```diff
@@ -239,7 +253,7 @@ func (chr *CacheHandler) GetCacheHandle(object *gcs.MinObject, bucket gcs.Bucket
 	// If cacheForRangeRead is set to False, initialOffset is non-zero (i.e. random read)
 	// and entry for file doesn't already exist in fileInfoCache then no need to
 	// create file in cache.
-	if !cacheForRangeRead && initialOffset != 0 {
+	if !cacheForRangeRead && initialOffset != 0 && !chr.isSparse {
```
raj-prince, line 257: nit: can we update the comment with Sparse logic?

internal/cache/file/cache_handler.go
```diff
@@ -188,14 +192,24 @@ func (chr *CacheHandler) addFileInfoEntryAndCreateDownloadJob(object *gcs.MinObj
...
+		newFileInfo := data.FileInfo{
 			Key:              fileInfoKey,
 			ObjectGeneration: object.Generation,
 			Offset:           0,
 			FileSize:         object.Size,
+			SparseMode:       chr.isSparse,
+			DownloadedRanges: nil,
+		}
+		// For sparse files, set Offset to MaxUint64 as a sentinel to indicate
+		// sparse mode, so Offset < requiredOffset checks always fail
```
raj-prince, line 204: Just want to confirm, do we hit Offset < requiredOffset check for SparseMode is enabled. As flow is different in cacheHandle.Read() for Sparse mode.

internal/gcsx/file_cache_reader.go
```diff
@@ -50,8 +51,14 @@ type FileCacheReader struct {
 	// will be downloaded for random reads as well too.
 	cacheFileForRangeRead bool
 
+	// mu protects fileCacheHandle from concurrent access across multiple threads
```
raj-prince: Thanks for catching this!!

Taken care as part of this PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/4016, need to be rebased.

raj-prince (commented):
LGTM overall, couple of comments.

internal/cache/file/cache_handle.go
vadlakondaswetha: sounds good. We will discuss internally on how to reduce this flag value or introduce another flag without user intervention and update it in follow up PR.

internal/cache/file/cache_handler.go
xinyangge: It's possible.  In the sparse file mode, a random I/O can attempt to read a chunk that is beyond the latest position we have downloaded chunks for, and this should be supported.

internal/cache/file/downloader/sparse_downloads_job.go
xinyangge: I think the logic here correctly handles requiredOffset as an exclusive end.  Assume `requiredOffset` = `chunkSize` = 1MB, then `chunkEnd` will be 1MB.

internal/cache/file/downloader/sparse_downloads_job.go
xinyangge: No one.  I will delete this function.

internal/gcsx/file_cache_reader.go
xinyangge: Rebased.

internal/cache/file/cache_handler.go
xinyangge: updated.

internal/cache/file/downloader/sparse_downloads_job.go
xinyangge: done

internal/cache/file/downloader/sparse_downloads_job.go
xinyangge: done

internal/cache/lru/lru.go
xinyangge: I'll leave more tests to the GCS fuse team as discussed on slack.

internal/cache/data/byte_range.go
xinyangge: done

internal/cache/file/downloader/sparse_downloads_job.go
xinyangge: done

internal/cache/file/downloader/sparse_downloads_job.go
xinyangge: done

internal/cache/file/downloader/sparse_downloads_job.go
xinyangge: done

xinyangge:
> Please fix the lint issue and the test failure, otherwise we will not be able to merge the PR. test failure:
> 
> > [ RUN      ] sparseDownloaderTest.Test_DownloadRange
> > {"timestamp":{"seconds":1763449559,"nanos":241075907},"severity":"INFO","message":"GetStorageLayout <- (gcsfuse-default-bucket)"}
> > {"timestamp":{"seconds":1763449559,"nanos":241227171},"severity":"INFO","message":"GetStorageLayout -> (gcsfuse-default-bucket) 0 msec"}
> > sparse_downloads_job_test.go:196:
> > Expected: not(is nil)
> > Actual:   {{gcsfuse-default-bucket 0001-01-01 00:00:00 +0000 UTC test/sparse_download_range.txt} 1763449560468562 18446744073709551615 52428800 true 0xc02f0a39e0}, which cannot be compared to nil
> 
> [ FAILED ] sparseDownloaderTest.Test_DownloadRange (1.309347417s)

@vadlakondaswetha , fixed!

tools/integration_tests/sparse_file/sparse_file_test.go
ashmeenkaur, line 1: Could you please move these tests to read_cache test package which already exists. You can create a new test suite similar to https://github.com/GoogleCloudPlatform/gcsfuse/blob/ba1d8edf5492f0c72caffec54fd60053f46b9544/tools/integration_tests/read_cache/cache_file_for_include_regex_test.go
and add flags to run the tests with here https://github.com/GoogleCloudPlatform/gcsfuse/blob/ba1d8edf5492f0c72caffec54fd60053f46b9544/tools/integration_tests/read_cache/setup_test.go#L296.

xinyangge: > Could you please move these tests to read_cache test package which already exists. You can create a new test suite similar to https://github.com/GoogleCloudPlatform/gcsfuse/blob/ba1d8edf5492f0c72caffec54fd60053f46b9544/tools/integration_tests/read_cache/cache_file_for_include_regex_test.go and add flags to run the tests with here
> 
> https://github.com/GoogleCloudPlatform/gcsfuse/blob/ba1d8edf5492f0c72caffec54fd60053f46b9544/tools/integration_tests/read_cache/setup_test.go#L296
> 
> .

@ashmeenkaur As discussed in this morning's meeting, I'll leave testing-related changes as a follow up to the GCS fuse team.

tools/integration_tests/sparse_file/sparse_file_test.go
```diff
@@ -0,0 +1,243 @@
...
+func (s *sparseFileTest) TearDownTest() {
+	setup.SaveGCSFuseLogFileInCaseOfFailure(s.T())
+}
+
+func (s *sparseFileTest) TearDownSuite() {
+	setup.UnmountGCSFuseAndDeleteLogFile(setup.TestDir())
+}
+
+func TestSparseFile(t *testing.T) {
+	suite.Run(t, new(sparseFileTest))
```
ashmeenkaur:
```suggestion
    setup.ParseSetUpFlags()
	suite.Run(t, new(sparseFileTest))
```

ashmeenkaur, line 85:
```suggestion
	// TODO: Enable these tests using conditional t.SkipNow() if integrationTest flag is not set.
	t.SkipNow()
	suite.Run(t, new(sparseFileTest))
```

Making this change temporarily to unblock PR linux tests failure. I will raise another PR to fix it properly.

xinyangge:
> I am not able to merge this PR as it requires CLA from all authors and claude is one of the author. Made a clone of this PR and submitted: #4049
> 
> Based on our internal discussions and suggestions from our PM @trinadhkotturu we have renamed the flag to enable-block-cache and made it experimental. We will keep it experimental until we decide on the plan for block cache which works in all scenarios.
> 
> Let me know if you have any questions. otherwise we can close this PR now.

@vadlakondaswetha Thanks for all the prompt reviews!  It sounds good to me, and I'll close this PR.

PR #3993, opened 2025-11-08
refactor: Remove DataBuf from ReadResponse

vipnydav:
### Description
This PR removes the `DataBuf` field from the `ReadResponse` struct. This is because the `DataBuf` field is redundant, as the same slice is passed to `ReadAt` methods and written to, and the caller already has access to the modified slice. This change simplifies the code by removing the need for repeated assignments.

#### Original Rationale for Adding `DataBuf`
The initial design for integrating the multi-range downloader considered an optimization to achieve a "single data copy" and avoid extra data copies between the Go SDK, GCSFuse, and the FUSE kernel. The plan was to use a reusable buffer within GCSFuse, which would be passed to the Go SDK's multi-range downloader. The `DataBuf` field was part of this design to hold a reference to the data that would be returned to the FUSE layer.

#### Justification for Removal
The optimization for a single data copy using a reusable buffer is not being implemented in the current scope for the private preview. The current implementation passes a destination slice to the read methods, and the data from GCS is written directly into this slice.

In Go, slices are reference types that point to an underlying array. When a slice is passed to a function, any modifications to the underlying data are visible to the caller. Since our read methods write directly into the provided slice, the caller already has the data.

Given this, returning the data buffer in the `ReadResponse` is redundant. This PR removes the `DataBuf` field to simplify the `ReadResponse` struct and the associated code, making it more aligned with the current implementation.

b/458976462

### Testing details
1. Manual - Done
2. Unit tests - NA

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -266,7 +266,7 @@ func (p *BufferedReader) prepareQueueForOffset(offset int64) {
 //
 // LOCKS_EXCLUDED(p.mu)
 func (p *BufferedReader) ReadAt(ctx context.Context, inputBuf []byte, off int64) (gcsx.ReadResponse, error) {
-	resp := gcsx.ReadResponse{DataBuf: inputBuf}
+	resp := gcsx.ReadResponse{}
 	reqID := uuid.New()
 	start := time.Now()
 	initOff := off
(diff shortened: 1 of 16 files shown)


raj-prince (approved):
LGTM.

Could you please update the description with - what was the rationale behind adding the buffer as part of struct, but not required now?

vipnydav:
> LGTM.
> 
> Could you please update the description with rationale behind adding the buffer as part of struct, but not required now?

I've not added `ReadFileOp.Data` in this PR. Just removed `ReadFileOp.Dst` from the `ReadResponse` in this PR.

PR #3999, opened 2025-11-10
feat(bufferedread): Add RefCount and WasEvicted to PrefetchBlock

vipnydav:
### Description
This PR introduces reference count variable to `PrefetchBlock`. This addition is essential for managing the lifecycle of prefetch blocks, especially for zero copy in `BufferedReader`, to prevent race conditions where a block is evicted from `PrefetchBlockQueue` while still in use.

b/459451076

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/internal/block/prefetch_block.go
+++ b/internal/block/prefetch_block.go
@@ -66,11 +66,13 @@ type PrefetchBlock interface {
 	// - BlockStatusDownloadFailed: Download of this block has failed.
 	NotifyReady(val BlockStatus)
 
-	// IncrementRef increments the reference count of the block.
-	IncrementRef()
+	// IncRef increments the reference count of the block.
+	IncRef()
 
-	// DecrementRef decrements the reference count of the block.
-	DecrementRef() int32
+	// DecRef decrements the reference count of the block. It returns true if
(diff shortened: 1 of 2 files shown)


internal/block/prefetch_block.go
```diff
@@ -206,3 +192,15 @@ func (pmb *prefetchMemoryBlock) NotifyReady(val BlockStatus) {
 		panic("Expected to notify only once, but got multiple notifications.")
 	}
 }
+
+func (pmb *prefetchMemoryBlock) IncrementRef() {
+	pmb.refCount.Add(1)
+}
+
+func (pmb *prefetchMemoryBlock) DecrementRef() int32 {
```
raj-prince: Why are we returning refCount?

Instead you can return a bool, true if ref_count reaches 0, otherwise false.

Also, can we add panic with logs which ensure client is not calling DecRef more than IncRef?

raj-prince (changes requested):
Added a comment.

internal/block/prefetch_block.go
vipnydav: Makes sense.  I've updated the method with changes.

raj-prince (approved):

PR #4000, opened 2025-11-10
test migratiom concurrent operations read working

PranjalC100:
### Description
This PR includes changes to migrate concurrent test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of concurrent package so it can use config file.

[b/445952164](https://buganizer.corp.google.com/issues/445952164)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/concurrent_operations/concurrent_read_test.go
+++ b/tools/integration_tests/concurrent_operations/concurrent_read_test.go
@@ -16,15 +16,17 @@ package concurrent_operations
 
 import (
 	"bytes"
+	"context"
 	"fmt"
+	"log"
 	"math/rand"
 	"os"
 	"path"
 	"sync"
-	"syscall"
 	"testing"
 	"time"
 
+	"cloud.google.com/go/storage"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/client"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/operations"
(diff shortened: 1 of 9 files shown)


tools/integration_tests/concurrent_operations/concurrent_read_test.go
```diff
@@ -33,27 +35,51 @@ import (
...
+}
+
+func (s *concurrentReadTest) SetupSuite() {
+	setup.SetUpLogFilePath(s.baseTestName, GKETempDir, OldGKElogFilePath, testEnv.cfg)
+	mountGCSFuseAndSetupTestDir(s.flags, s.ctx, s.storageClient)
+}
+
+// createFileAndUploadToGCS is a helper function to create a local file of a given
+// size and upload it to GCS for subsequent validation.
+func (s *concurrentReadTest) createFileAndUploadToGCS(fileName string, fileSize int64) string {
```
PranjalC100, line 76: Calling ```operations.CreateFileOfSize```, creates a 500 MiB file inside the GCSfuse mount point. GCSfuse sees this new file and begins uploading it to your GCS bucket in the background. 

This upload may take time larger files. The problem I faced was that the test code didn't wait for this background upload to finish. It immediately proceeds to the next step, where goroutines try to read the file and validate its checksum against the object in GCS. 

Not sure if this approach is correct or not, we could also add a wait(5 second) call. Also could this be the reason for these errors? (https://b.corp.google.com/issues/447905694)

To avoid this, I am calling this function, ```client.UploadGcsObject```.

PR #4008, opened 2025-11-12
fix(Periodic tests on arm64 machines): Fix python upgrade

Tulsishah:
### Description
The environment variable CLOUDSDK_PYTHON and the required $PATH update were not being correctly carried over to sub-processes because they were only exported in the function's local scope. The function now permanently sets these variables in ~/.bashrc, guaranteeing their availability for all processes and shell sessions.

1. [b/459303275](https://b.corp.google.com/issues/459303275)
2. [b/459303385](https://b.corp.google.com/issues/459303385)

### Testing details
1. Manual - Done
4. Unit tests - NA
5. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/ml_tests/checkpoint/Jax/run_checkpoints.sh
+++ b/perfmetrics/scripts/ml_tests/checkpoint/Jax/run_checkpoints.sh
@@ -59,12 +59,6 @@ function mount_gcsfuse_and_run_test() {
   python3.11 ./perfmetrics/scripts/ml_tests/checkpoint/Jax/emulated_checkpoints.py --checkpoint_dir "${MOUNT_POINT}"
 }
 
-# Enable python virtual environment.
-# By default KOKORO VM installs python 3.8 which causes dependency issues.
-# Following commands are to explicitly set up python 3.11.
-sudo apt update
-sudo apt install -y software-properties-common
-sudo add-apt-repository -y ppa:deadsnakes/ppa
 # Install pip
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/ml_tests/checkpoint/Jax/run_checkpoints.sh
```diff
@@ -65,8 +65,6 @@ function mount_gcsfuse_and_run_test() {
 sudo apt update
```
ashmeenkaur: Are these 3 commands still required? Can you also update the comment above?

perfmetrics/scripts/install_latest_gcloud.sh
```diff
@@ -60,7 +60,10 @@ else
     # If this script is run in background or different shell then
     # export PATH needs to be called from the shell or use absolute gcloud path
     # or permanently add this path to path variable in bashrc.
-    export PATH="${INSTALL_DIR}/google-cloud-sdk/bin:$PATH"
+    echo 'export PATH="/usr/local/google-cloud-sdk/bin:$PATH"' >> "$HOME/.bashrc"
```
ashmeenkaur, line 63: Shall we rename the script to setup_environment.sh instead?

Tulsishah: I think the name is correct as it is upgrading gcloud and then setting env variable. We can discussed about it later as it is being used at many places I don't this we should rename it now.

ashmeenkaur (approved):

PR #4012, opened 2025-11-13
test: Fix periodic tests

Tulsishah:
### Description
- sourcing bashsrc was not working as it applicable only for parent process.  Explicitly making export statements to set particular version.

1. https://b.corp.google.com/issues/459303385
2. https://b.corp.google.com/issues/459303275

### Testing details
1. Manual - NA
3. Unit tests - NA
4. Integration tests - NA


meet2mky:
Let's merge this change temporary for fixing the jobs. I will improve it later to ensure these version and path issues are managed centrally from a single file which all scripts can use it.

charith87 (approved):
Relying on Mohit's approval and since its to fix tests.

PR #4015, opened 2025-11-13
feat(bufferedread): Implement zero-copy reads in buffered reader

vipnydav:
### Description
This PR enhances the performance of the `BufferedReader` by implementing a zero-copy read mechanism. This is achieved by modifying the `ReadAt` method to directly return slices of its internal buffers, avoiding an intermediate data copy and thereby improving read performance.

b/460428144

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -103,8 +103,9 @@ type BufferedReader struct {
 	// GUARDED by (mu)
 	blockPool *block.GenBlockPool[block.PrefetchBlock]
 
-	// A waitgroup to track outstanding references to data slices that have been returned
-	// directly from internal buffers to FUSE library.
+	// A WaitGroup to synchronize the destruction of the reader with any ongoing
+	// FUSE read callback goroutines. This ensures that all callbacks for
+	// in-flight data slices have completed before the reader is fully torn down.
 	inflightCallbackWg sync.WaitGroup
 }
 
@@ -553,10 +554,10 @@ func (p *BufferedReader) Destroy() {
 	p.blockPool = nil
 }
 
-// releaseOrMarkEvicted waits for any pending download on the given block to complete
-// and then handles its release. If the block has no outstanding references,
-// it is immediately returned to the block pool. If there are outstanding
-// references, the block is marked as evicted, and its final release
+// releaseOrMarkEvicted handles the release of a block that has been removed
+// from the prefetch queue. If the block has no outstanding references (i.e.,
+// it has not been returned to a FUSE read), it is immediately returned to the
+// block pool. Otherwise, the block is marked as evicted, and its final release
 // is deferred until the last reference's callback is executed.
 // LOCKS_REQUIRED(p.mu)
 func (p *BufferedReader) releaseOrMarkEvicted(entry *blockQueueEntry) {
--- a/internal/bufferedread/buffered_reader_test.go
+++ b/internal/bufferedread/buffered_reader_test.go
@@ -24,11 +24,11 @@ import (
 	"time"
 
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/block"
-	"github.com/googlecloudplatform/gcsfuse/v3/internal/cache/util"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/gcsx"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/fake"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/workerpool"
 	"github.com/googlecloudplatform/gcsfuse/v3/metrics"
 	"github.com/stretchr/testify/mock"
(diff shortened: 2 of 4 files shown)


internal/block/prefetch_block_test.go
```diff
@@ -126,54 +126,6 @@ func (testSuite *PrefetchMemoryBlockTest) TestPrefetchMemoryBlockReadAtEOF() {
 	assert.Equal(testSuite.T(), []byte("world"), readBuffer[0:n])
 }
 
-func (testSuite *PrefetchMemoryBlockTest) TestPrefetchMemoryBlockReadAtSliceSuccess() {
```
raj-prince, line 129: Why are we removing these tests?

vipnydav: There was some problem with the rebase. Added these tests again.

internal/bufferedread/buffered_reader.go
```diff
@@ -369,6 +383,12 @@ func (p *BufferedReader) ReadAt(ctx context.Context, inputBuf []byte, off int64)
 		}
 	}
 
+	resp.Data = data
+	resp.Callback = func() {
+		for _, cb := range callbacks {
+			cb()
```
raj-prince: Why extra function calls?

Instead:
1. Change the p.callback() to accept the list of block entry.
2. Instead of creating the slice of callbacks, create slice of entries.
3. Return a final callback which calls p.callbacks(entries).

internal/bufferedread/buffered_reader.go
```diff
@@ -230,19 +224,28 @@ func (p *BufferedReader) prepareQueueForOffset(offset int64) {
...
+// was returned directly from a block. It decrements the block's reference count
+// and releases it back to the pool if the count drops to zero and it was
+// previously marked for eviction.
+func (p *BufferedReader) callback(entry *blockQueueEntry) {
+	defer p.inflightCallbackWg.Done()
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if entry.block.DecRef() && entry.wasEvicted {
+		p.releaseBlock(entry)
```
raj-prince: Call here - p.blockPool.Release(entry.block)

And we can change the p.releaseBlock(entry) to something p.releaseOrMarkEvicted(entry), and we will call this over entry when we pop the entry from queue.

internal/bufferedread/buffered_reader.go
vipnydav: Done

internal/bufferedread/buffered_reader.go
```diff
@@ -109,6 +102,10 @@ type BufferedReader struct {
 	// of blocks that can be allocated across all BufferedReader instances.
 	// GUARDED by (mu)
 	blockPool *block.GenBlockPool[block.PrefetchBlock]
+
+	// A waitgroup to track outstanding references to data slices that have been returned
```
raj-prince: nit: Comment is not very clear. WaitGroup is used to synchronized go-routines, please write in terms of go-routine.

internal/bufferedread/buffered_reader.go
```diff
@@ -495,35 +516,60 @@ func (p *BufferedReader) scheduleBlockWithIndex(b block.PrefetchBlock, blockInde
...
-	err := p.blockPool.ClearFreeBlockChannel(true)
-	if err != nil {
+	if err := p.blockPool.ClearFreeBlockChannel(true); err != nil {
 		logger.Warnf("Destroy: clearing free block channel: %v", err)
 	}
 	p.blockPool = nil
 }
 
+// releaseOrMarkEvicted waits for any pending download on the given block to complete
+// and then handles its release. If the block has no outstanding references,
```
raj-prince: `waits for any pending download on the given block` - Is it correct?

internal/bufferedread/buffered_reader_test.go
```diff
@@ -1671,3 +1697,137 @@ func (t *BufferedReaderTest) TestReadAtConcurrentReads() {
...
+	initialFreeBlocks := reader.blockPool.TotalFreeBlocks()
+
+	reader.mu.Lock()
+	reader.releaseOrMarkEvicted(entry)
+	reader.mu.Unlock()
+
+	assert.True(t.T(), entry.wasEvicted, "Block should be marked as evicted")
+	assert.Equal(t.T(), initialFreeBlocks, reader.blockPool.TotalFreeBlocks(), "Block should not be released to pool yet")
+	reader.mu.Lock()
+	if entry.block.DecRef() {
```
raj-prince: Please re-write this test, not sure what do we want to test. Seems like, we have executed the requires instruction.

Please test the behavior of the method. If you want to test, the release, you can test via Read itself, that execution of callback releases the block to the pool.

internal/fs/handle/file_test.go
```diff
@@ -946,7 +946,13 @@ func (t *fileTest) Test_ReadWithReadManager_FullReadSuccessWithBufferedRead() {
 
 	assert.NoError(t.T(), err)
 	assert.Equal(t.T(), fileSize, resp.Size)
-	assert.Equal(t.T(), expectedData, buf)
+	// When buffered read is enabled, the data is returned in resp.Data and not
+	// directly in the buffer. We need to copy it to compare.
+	bytesCopied := 0
+	for _, dataSlice := range resp.Data {
```
raj-prince: Please create a test utility to convert from resp.Data -> string and use that everywhere. Keep the utility common place so that you can use in the Buffered-reader test too.

raj-prince (commented):
LGTM, few minor comments.

internal/bufferedread/buffered_reader.go
vipnydav: The comment was for a previous version of this method.
I updated the method but forgot to update the comment.

The comment has been updated now.

internal/bufferedread/buffered_reader.go
vipnydav: Updated the comment.

internal/bufferedread/buffered_reader_test.go
vipnydav: With the previous test I wanted to test that if block is being removed from prefetch queue but is still being used by jacobsa/fuse it doesn't get released to free blocks channel.

I've updated it to test via Read path.

raj-prince (approved):
LGTM.

Please don't merge if any release planned from master within a week.

PR #4016, opened 2025-11-13
fix: Adding synchronization logic in fileCache

abhishek10004:
### Description
Adding synchronization logic in FileCacheReader 

b/460267250

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - Yes

### DIFF
--- a/internal/gcsx/file_cache_reader.go
+++ b/internal/gcsx/file_cache_reader.go
@@ -38,6 +38,11 @@ const (
 	MiB    = 1 << 20
 )
 
+// FileCacheReader is a reader that attempts to satisfy read requests for a GCS
+// object from a local file cache. It is designed to be part of a layered
+// reading strategy, where it acts as the first-level cache.
+//
+// FileCacheReader supports parallel reads.
 type FileCacheReader struct {
 	Reader
 	object *gcs.MinObject
@@ -162,6 +167,8 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 	fc.mu.Unlock()
 
 	fc.mu.RLock()
+	// Because we're releasing write lock & then taking a read lock, we need to perform a nil check before accessing
+	// fileCacheHandle as some other thread could have set it to nil in between.
 	if fc.fileCacheHandle == nil {
 		fc.mu.RUnlock()
 		return 0, false, nil


internal/gcsx/file_cache_reader.go
```diff
@@ -150,8 +159,15 @@ func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte
 			}
 		}
 	}
+	fc.mu.Unlock()
 
+	fc.mu.RLock()
+	if fc.fileCacheHandle == nil {
```
raj-prince, line 172: Why is this check required?

internal/gcsx/file_cache_reader.go
raj-prince, line 46: Lets update the FileCacheReader documentation, that it supports concurrent reads.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -699,3 +700,79 @@ func (t *fileCacheReaderTest) Test_Destroy_NilCacheHandle() {
 
 	assert.Nil(nil, t.reader.fileCacheHandle)
 }
+
+func (t *fileCacheReaderTest) Test_Concurrent_ReadAt() {
```
raj-prince, line 704: Should we add a timeout logic for this test - high enough, to ensure failure for deadlock scenario.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -699,3 +700,79 @@ func (t *fileCacheReaderTest) Test_Destroy_NilCacheHandle() {
...
+	t.mockNewReaderWithHandleCallForTestBucket(t.object.Size, rd)
+	t.mockBucket.On("Name").Return("test-bucket")
+	t.mockBucket.On("BucketType").Return(t.bucketType)
+
+	var wg sync.WaitGroup
+	numGoroutines := 20
+	wg.Add(numGoroutines)
+
+	// Act: Concurrently try to read and destroy from a bunch of goroutines.
+	for i := 0; i < numGoroutines; i++ {
```
raj-prince: Can we convert into two for blocks, one for reader go-routine, other for destroyer routine? Just want to reduce nesting in the test.

internal/gcsx/random_reader.go
```diff
@@ -183,6 +183,9 @@ type randomReader struct {
 
 	// To synchronize reads served from range reader.
 	mu sync.Mutex
+
+	// To synchronize access to fileCacheHandle
+	fileCacheMu sync.RWMutex
```
raj-prince, line 188: Should we add test for this change too? If takes time, then we can take the random_reader changes separately.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -699,3 +700,79 @@ func (t *fileCacheReaderTest) Test_Destroy_NilCacheHandle() {
...
+
+func (t *fileCacheReaderTest) Test_Concurrent_ReadAt_And_Destroy() {
+	// Setup
+	t.object.Size = 100
+	testContent := testutil.GenerateRandomBytes(int(t.object.Size))
+	rd := &fake.FakeReader{ReadCloser: getReadCloser(testContent)}
+	t.mockNewReaderWithHandleCallForTestBucket(t.object.Size, rd)
+	t.mockBucket.On("Name").Return("test-bucket")
+	t.mockBucket.On("BucketType").Return(t.bucketType)
+
```
raj-prince: nit: 
```suggestion
```

internal/gcsx/file_cache_reader_test.go
```diff
@@ -699,3 +700,79 @@ func (t *fileCacheReaderTest) Test_Destroy_NilCacheHandle() {
...
+	// Setup
+	t.object.Size = 100
+	testContent := testutil.GenerateRandomBytes(int(t.object.Size))
+	rd := &fake.FakeReader{ReadCloser: getReadCloser(testContent)}
+	// Mock NewReaderWithReadHandle for the downloader job, which is triggered by GetCacheHandle.
+	// This should only be called once across all concurrent reads.
+	t.mockBucket.On("NewReaderWithReadHandle", mock.Anything, mock.Anything).Return(rd, nil).Once()
+	t.mockBucket.On("Name").Return("test-bucket")
+	t.mockBucket.On("BucketType").Return(t.bucketType)
+
```
raj-prince: nit: 
```suggestion
```

internal/gcsx/file_cache_reader_test.go
```diff
@@ -699,3 +700,79 @@ func (t *fileCacheReaderTest) Test_Destroy_NilCacheHandle() {
...
+			buf := make([]byte, t.object.Size)
+			readResponse, err := t.reader.ReadAt(t.ctx, buf, 0)
+
+			// Assert
+			assert.NoError(t.T(), err)
+			assert.Equal(t.T(), int(t.object.Size), readResponse.Size)
+			assert.Equal(t.T(), testContent, buf)
+		}()
+	}
+
```
raj-prince: nit: 
```suggestion
```

internal/gcsx/file_cache_reader.go
abhishek10004: Because we are releasing the write lock & taking a read lock, it's possible that some other thread can set it to nil in between. Hence, once we have taken a read lock, we have to do a nil check before accessing the variable

internal/gcsx/random_reader.go
abhishek10004: I did not add tests for this since this is not the default path right now. I will add it in a separate PR.

internal/gcsx/file_cache_reader.go
raj-prince: Aah, yes. Can we write a comment here?

internal/gcsx/file_cache_reader.go
abhishek10004: Yeah, will update it.

internal/gcsx/file_cache_reader_test.go
abhishek10004: Done

PR #4018, opened 2025-11-14
fix: fix managed folder tests in CD pipeline

Tulsishah:
### Description
Managed folder e2e tests are failing due to python dependency. Fixing it by upgrading it.

[b/460606295](https://b.corp.google.com/issues/460606295)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


ashmeenkaur (approved):
Approved but please check if we can use central script everywhere to upgrade python version

PR #4023, opened 2025-11-17
fix(dentry cache): mounted dir case correction for dentry cache package

PranjalC100:
### Description
**Issue:** GKE integration tests failed because the GCS FUSE mount directory was being erased.

**Cause:** The function responsible for setting up the test environment was unconditionally resetting the mount directory path. This action erased the correct path that had been configured for the GKE test, causing the test to fail as it could no longer find the mounted directory.

**Solution:** A check was added to only set the mount directory path if it wasn't already configured.

[b/464448423](https://buganizer.corp.google.com/issues/464448423)

### Testing details
2. Unit tests - 

### DIFF
--- a/tools/integration_tests/dentry_cache/setup_test.go
+++ b/tools/integration_tests/dentry_cache/setup_test.go
@@ -54,7 +54,7 @@ type env struct {
 
 func mountGCSFuseAndSetupTestDir(flags []string, ctx context.Context, storageClient *storage.Client) {
 	setup.MountGCSFuseWithGivenMountWithConfigFunc(testEnv.cfg, flags, mountFunc)
-	if setup.MountedDirectory() == "" {
+	if testEnv.cfg.GKEMountedDirectory == "" {
 		setup.SetMntDir(testEnv.cfg.GCSFuseMountedDirectory)
 	}
 	testEnv.testDirPath = client.SetupTestDirectory(ctx, storageClient, testDirName)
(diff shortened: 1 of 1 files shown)


tools/integration_tests/dentry_cache/setup_test.go
```diff
@@ -54,7 +54,9 @@ type env struct {
 
 func mountGCSFuseAndSetupTestDir(flags []string, ctx context.Context, storageClient *storage.Client) {
 	setup.MountGCSFuseWithGivenMountWithConfigFunc(testEnv.cfg, flags, mountFunc)
-	setup.SetMntDir(testEnv.cfg.GCSFuseMountedDirectory)
+	if setup.MountedDirectory() == "" {
```
ashmeenkaur: Were the tests failing without this change?

PranjalC100: yes

ashmeenkaur:
please update bug description

tools/integration_tests/dentry_cache/setup_test.go
ashmeenkaur: can we change this to
```suggestion
	if testEnv.cfg.GCSFuseMountedDirectory == "" {
```
?

PranjalC100:
Updated!

tools/integration_tests/dentry_cache/setup_test.go
PranjalC100: the tests pass with ```if testEnv.cfg.GKEMountedDirectory == ""```

ashmeenkaur (approved):

PR #4024, opened 2025-11-17
ci: create a periodic job cofig for daily runs on 8 core machine

anushka567:
### Description
As part of louhi pipeline stabilization, we have increased the machine size from 16 to 48. This has widened the existing infra gap between GCSFuse testing flow and GKE test grid, this can lead to surprises during release where test dependent on cores can suffer, To ensure such regressions are not introduced, we are introducing a new kokoro periodic job which will be run on 8 core machines.

b/459938372

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/e2e-tests-master-8-cores.cfg
+++ b/perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/e2e-tests-master-8-cores.cfg
@@ -1,0 +1,29 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


perfmetrics/scripts/continuous_test/gcp_ubuntu/e2e_tests/e2e-tests-master-8-cores.cfg
```diff
@@ -0,0 +1,29 @@
...
+  define_artifacts {
+    regex: "gcsfuse-failed-integration-test-logs-*"
+    strip_prefix: "github/gcsfuse/perfmetrics/scripts"
+    regex: "**/*sponge_log.*"
+    regex: "proxy*"
+  }
+}
+
+env_vars {
+  key: "PACKAGE_LEVEL_PARALLELISM"
```
Tulsishah, line 25: Currently you are not using this env variable also didn't mention in the cl/832218767

anushka567: the cl for only for resource pool creation. After this PR is merged , I have another final CL which will be responsible for putting the job in place.

PR #4025, opened 2025-11-17
fix(e2e-test): Reduce the read effort in the concurrent read-test

raj-prince:
### Description
Test_ConcurrentSegmentReadsSharedHandle fails on some of the ARM vm because of slowness. Slowness mainly happens due to lack of cpu resources on the system and different read-request size from the kernel on different system.

Two options to fix this:
1. Increase the timeout for the test from 5 min to even larger: rejected, as it will unnecessarily increase the test time.
2. Reduce the work done as part of this test - accepted, as it will test the concurrent read functionality without increasing the timeout.

b/447905694

### Testing details
1. Manual - Tested manually.
3. Unit tests - NA
4. Integration tests - NA


anushka567 (approved):

Tulsishah (approved):
Relying on @anushka567 's approval.

PR #4029, opened 2025-11-17
refactor: unfinalized object test package migration [GKE-GCSFuse Test migration]

PranjalC100:
### Description
This PR includes changes to migrate unfinalized obejct test package to use common config file. This common config will be used by both GCSFuse binary and GCSFuse csi driver tests.

#### Changes include:

Migration of unfinalized test package so it can use config file.

[b/461525255](https://buganizer.corp.google.com/issues/461525255)

### Testing details
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
+++ b/tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
@@ -47,7 +47,15 @@ func (t *unfinalizedObjectOperations) SetupTest() {
 	t.fileName = path.Base(t.T().Name()) + setup.GenerateRandomString(5)
 }
 
-func (t *unfinalizedObjectOperations) TeardownTest() {}
+func (s *unfinalizedObjectOperations) TearDownSuite() {
+	setup.UnmountGCSFuseWithConfig(testEnv.cfg)
+}
+
+func (s *unfinalizedObjectOperations) SetupSuite() {
+	setup.MountGCSFuseWithGivenMountWithConfigFunc(testEnv.cfg, s.flags, mountFunc)
+	setup.SetMntDir(testEnv.cfg.GCSFuseMountedDirectory)
+	testEnv.testDirPath = client.SetupTestDirectory(s.ctx, s.storageClient, testDirName)
+}
 
 ////////////////////////////////////////////////////////////////////////
 // Test scenarios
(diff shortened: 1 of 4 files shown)


tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
```diff
@@ -47,7 +47,17 @@ func (t *unfinalizedObjectOperations) SetupTest() {
...
-func (t *unfinalizedObjectOperations) TeardownTest() {}
+func (s *unfinalizedObjectOperations) TearDownSuite() {
+	setup.UnmountGCSFuseWithConfig(testEnv.cfg)
+}
+
+func (s *unfinalizedObjectOperations) SetupSuite() {
+	setup.MountGCSFuseWithGivenMountWithConfigFunc(testEnv.cfg, s.flags, mountFunc)
+	if setup.MountedDirectory() == "" {
+		setup.SetMntDir(testEnv.cfg.GCSFuseMountedDirectory)
+	}
```
ashmeenkaur:
```suggestion
	if testEnv.cfg.GCSFuseMountedDirectory == "" {
		setup.SetMntDir(testEnv.cfg.GCSFuseMountedDirectory)
	}
```

tools/integration_tests/unfinalized_object/unfinalized_read_test.go
```diff
@@ -48,7 +48,17 @@ func (t *unfinalizedObjectReads) SetupTest() {
...
-func (t *unfinalizedObjectReads) TeardownTest() {}
+func (s *unfinalizedObjectReads) TearDownSuite() {
+	setup.UnmountGCSFuseWithConfig(testEnv.cfg)
+}
+
+func (s *unfinalizedObjectReads) SetupSuite() {
+	setup.MountGCSFuseWithGivenMountWithConfigFunc(testEnv.cfg, s.flags, mountFunc)
+	if setup.MountedDirectory() == "" {
+		setup.SetMntDir(testEnv.cfg.GCSFuseMountedDirectory)
+	}
```
ashmeenkaur: here also same change

ashmeenkaur (commented):
LGTM otherwise

ashmeenkaur (approved):

PR #4030, opened 2025-11-18
ci: Automate Go code formatting in pull requests

vipnydav:
### Description

This PR introduces a new GitHub Actions workflow to automatically format Go files and update dependencies in pull requests.

Currently, PRs with improperly formatted Go files cause CI failures, requiring manual correction. This new workflow improves the developer experience by automatically applying `go generate`, `go mod tidy`, and `goimports` and committing the changes directly to the PR branch.

This automation eliminates manual formatting, reduces CI failures, and allows developers to focus on their code.

#### Security Enhancement:

This change uses a safe two-workflow approach.
An unprivileged workflow (`generate-and-format.yml`) runs on pull requests to generate a diff of formatting changes.
A separate, privileged workflow (`apply-auto-fixes.yml`) applies this diff, ensuring that untrusted code is never executed with write permissions or access to secrets.

b/461648699

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -161,7 +161,7 @@ func NewBufferedReader(opts *BufferedReaderOptions) (*BufferedReader, error) {
 // returns a gcsx.FallbackToAnotherReader error to signal that another reader
 // should be used. It takes handleID for logging purposes.
 // LOCKS_REQUIRED(p.mu)
-func (p *BufferedReader) handleRandomRead(offset int64,handleID int64) error {
+func (p *BufferedReader) handleRandomRead(offset int64, handleID int64) error {
 	// Exit early if we have already decided to fall back to another reader.
 	// This avoids re-evaluating the read pattern on every call when the random
 	// read threshold has been met.


.github/workflows/go-generate-and-format.yml
```diff
@@ -0,0 +1,57 @@
...
+         go-version: '1.24'
+
+     - name: Install goimports
+       run: go install golang.org/x/tools/cmd/goimports@latest
+
+     - name: Generate, Tidy, and Format
+       run: |
+         go generate ./...
+         go mod tidy
+         goimports -w .
```
kislaykishore: How about you install `make` and run `make fmt`

vipnydav: Using `make fmt` now. No need to explicitly install `make` since `ubuntu` image provided by github comes preinstalled with it.

.github/workflows/apply-auto-fixes.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Checkout PR branch
+        if: steps.check_patch.outputs.patch_exists == 'true'
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.workflow_run.head_repository.full_name }}
+          ref: ${{ github.event.workflow_run.head_branch }}
+          token: ${{ secrets.CODE_AUTO_FORMAT }}
+
+      - name: Apply patch
```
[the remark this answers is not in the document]
vipnydav: This is a false positive from the CodeQL scanner.

The alert is correctly identifying a pattern that is often dangerous: a privileged workflow checking out untrusted code from a pull request.

However, this workflow is secure because it never executes any code from the checked-out branch. Its only action is to run git apply on a .patch file. This patch file is a safe artifact generated by a separate, unprivileged workflow.

This two-workflow design intentionally separates the execution of untrusted code from privileged actions, which mitigates the vulnerability this alert is designed to prevent.

.github/workflows/generate-and-format.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          # Create a patch file if there are any changes
+          git diff --exit-code > auto_fixes.patch || true
+
+      - name: Upload patch artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: auto-fixes-patch
+          path: auto_fixes.patch
+          if-no-files-found: ignore
```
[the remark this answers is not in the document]
vipnydav: Added explicit permisions.

.github/workflows/generate-and-format.yml
```diff
@@ -0,0 +1,46 @@
+name: Generate and Format
+
+on:
+  pull_request:
+    paths:
+      - "**.go"
```
kislaykishore: You should also consider yaml files. For instance metrics.yaml change should trigger code-generation but I don't think it does.

vipnydav: Added All Files.

PR #4037, opened 2025-11-19
feat(bufferedread): Optimize prefetching for single-block reads

vipnydav:
### Description
**Problem:**

The `BufferedReader`'s prefetching logic is unnecessarily aggressive. It triggers an initial prefetch even for small or random reads that are fully contained within a single block. This leads to the allocation of unused blocks, causing block pool starvation and memory pressure in mixed-read workloads.

**Solution:**

This PR makes the initial prefetch conditional. A prefetch is now triggered only when a read request is guaranteed to cross a block boundary. This is achieved by calculating the remaining space in the current block and comparing it with the read size.

**Benefits:**

*   **Reduces Memory Pressure:** Prevents unnecessary block allocation for small reads which can be fulfilled from single block.
*   **Improves Block Availability:** Conserves blocks, making them more available for sequential reads.
*   **Enhances Throughput:** Reduces contention on the global block pool, improving performance in mixed-read workloads.

https://b.corp.google.com/issues/462137578

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -303,10 +303,11 @@ func (p *BufferedReader) ReadAt(ctx context.Context, inputBuf []byte, off int64)
 		p.prepareQueueForOffset(off)
 
 		if p.blockQueue.IsEmpty() {
-			// Prefetch only if the read spans across more than one block. This avoids aggressive
-			// prefetching for small reads contained within a single block.
+			// Prefetch if the read is sequential from the start, or if it spans multiple blocks.
+			// This ensures good performance for sequential reads while avoiding aggressive
+			// prefetching for small, random reads contained within a single block, thus saving blocks.
 			remainingReadSize := int64(len(inputBuf) - bytesRead)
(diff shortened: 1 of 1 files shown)


internal/bufferedread/buffered_reader.go
```diff
@@ -303,12 +303,13 @@ func (p *BufferedReader) ReadAt(ctx context.Context, inputBuf []byte, off int64)
 		p.prepareQueueForOffset(off)
 
 		if p.blockQueue.IsEmpty() {
-			if err = p.freshStart(off); err != nil {
+			shouldPrefetch := int64(len(inputBuf)-bytesRead) > p.config.PrefetchBlockSizeBytes-(off%p.config.PrefetchBlockSizeBytes)
```
raj-prince: Can we simplify this logic? Like, if request Range lies on the same block. lets request range is [rs, re], and bs is the block-size.

not prefetch if (rs / bs == re / bs) otherwise prefetch.

Also, please write a comment.

raj-prince (changes requested):
Overall LGTM, added a comment to simplify the prefetch logic.

raj-prince:
Discussed offline, this has another cons - when to trigger next prefetch if the a single block is downloaded as part of the first read.

PR #4043, opened 2025-11-20
feat(tailing reads): split open mode into access mode and file flags

anushka567:
### Description
This pull request refactors how file open modes are handled in gcsfuse to improve code clarity, maintainability, type safety, and to better align with POSIX standards. 

The main change is the introduction of a new OpenMode struct that separates the file access mode from other file flags, which is a cleaner and more extensible approach that better aligns with the POSIX model.

Also, the PR includes file handle creation during createfile call or openfile call while being aware whether O_DIRECT is enabled or not.

P.S: 
- ~The CI is currently failing due to a missing IsDirect() method in the jacobsa/fuse library. This is expected and will be resolved by an upcoming change in the jacobsa/fuse library. PR for that is in review.~

- AccessMode defaults to O_RDONLY . FileFlags is optional.  This should not interfere with any read/write flows as it is appropriately modified in cases where special handling is required.

MOTIVATION:
The current implementation used enum based approach to represent file open modes (mutually exclusive access modes and file flags which can mutually co-exist) , this is not ideal. Official documentation about the explicit [distinction](https://man7.org/linux/man-pages/man2/open.2.html).

[b/462588952](b/462588952)
[b/462281661](b/462281661)

### Testing details
1. Manual - Yes, checked manually if the correct open flags are being propagated.
2. Unit tests - Included.
3. Integration tests - Automated ~Will add once dependency is merged.~

### DIFF
--- a/internal/util/file_util.go
+++ b/internal/util/file_util.go
@@ -53,14 +53,6 @@ func (om OpenMode) FileFlags() int {
 	return om.fileFlags
 }
 
-func (om *OpenMode) SetAccessMode(accessMode int) {
-	om.accessMode = accessMode
-}
-
-func (om *OpenMode) SetFileFlags(fileFlags int) {
-	om.fileFlags = fileFlags
-}
-
 // IsAppend checks if the file was opened in append mode.
 // A file is considered in append mode (as suited for GCSFuse logic) if it is not
 // opened as read-only and the O_APPEND flag is set.


internal/fs/fs.go
```diff
@@ -2944,7 +2945,7 @@ func (fs *fileSystem) WriteFile(
 		gcsSynced, err = fh.Write(ctx, op.Data, op.Offset)
 	} else {
 		// Serve the request.
-		gcsSynced, err = in.Write(ctx, op.Data, op.Offset, util.Write)
+		gcsSynced, err = in.Write(ctx, op.Data, op.Offset, util.OpenMode{AccessMode: util.WriteOnly})
```
vadlakondaswetha: is there anythign wrong in passing the fh.openMode() here? Trying to avoid creating a new openMode when we already have one associated with fh

internal/util/file_util.go
```diff
@@ -14,36 +14,75 @@
...
-	case op.OpenFlags.IsAppend():
-		return Append
-	case op.OpenFlags.IsReadOnly():
-		return Read
-	default:
-		return Write
+// OpenMode represents the file open mode.
+type OpenMode struct {
+	// AccessMode defines the mutually exclusive access modes for opening a file.
+	AccessMode int
```
vadlakondaswetha: I am wondering if these should be public or should we expose them via methods. Because in some cases you are intializing by using OpenFlagAttributes and in other cases setting the variables directly. 

One option would be create another constructor which takes the required fields and initialize openMode struct and expose AccessMode and fileFlags via getter

does it makes sense to do it or will it become too complicated in the process of implementing o-direct chagnes

internal/fs/fs.go
anushka567: We can pass the openMode here, but then the enableRapidAppends flag will be no-op since the difference earlier was that while writing via file handle we attach the openmode as well before invoking inode write. 
Since, we have made the enableRapidAppends default on, we can remove this altogether to just 

```
gcsSynced, err = fh.Write(ctx, op.Data, op.Offset)
```


as fh.Write is defined as 
```
func (fh *FileHandle) Write(ctx context.Context, data []byte, offset int64) (bool, error) {
	return fh.inode.Write(ctx, data, offset, fh.openMode)
}

```

internal/util/file_util.go
anushka567: Agreed, makes sense to keep it private, Setters might be redundant as open mode is immutable but I had added for sense of completion.

internal/fs/fs.go
```diff
@@ -2003,7 +2003,7 @@ func (fs *fileSystem) createLocalFile(ctx context.Context, parentID fuseops.Inod
 	fs.mu.Unlock()
 	defer fs.mu.Lock()
 	fileInode.Lock()
-	err = fs.createBufferedWriteHandlerAndSyncOrTempWriter(ctx, fileInode, util.Write)
+	err = fs.createBufferedWriteHandlerAndSyncOrTempWriter(ctx, fileInode, openMode)
```
abhishek10004, line 2006: This seems to be a behavior change from the current code. Have we verified this?

anushka567: There is no functional change for `createLocalFile`. This function's purpose is to create a new, empty local file, for which the buffered write handler (BWH) is always triggered. The new `openMode` handling would only introduce a different behavior for an existing, unfinalized object opened in append mode, which is not a scenario for `createLocalFile`.

anushka567: Also tested via automated integration tests on ZB as well regional.

internal/util/file_util.go
```diff
@@ -14,36 +14,107 @@
...
-	switch {
-	case op.OpenFlags.IsAppend():
-		return Append
-	case op.OpenFlags.IsReadOnly():
-		return Read
-	default:
-		return Write
+// This abstraction allows a single function, `FileOpenMode`, to process flags
+// from different FUSE operations and also simplifies unit testing.
+type OpenFlagAttributes interface {
```
abhishek10004, line 76: Why can't we use OpenFlags directly here by importing fusekernel package?

anushka567: fusekernel is internal package to jacobsa/fuse library and cannot be imported directly.  [ref](https://go.dev/doc/go1.4#internalpackages)

internal/fs/fs.go
abhishek10004: I was wondering that since openMode can be different, would buffered write handler be even initialized or not but checked the code and for local files, we're initializing it irrespective of the open mode. So, there would be no change. Thanks for clarifying.

abhishek10004 (approved):
LGTM

internal/util/file_util.go
```diff
@@ -14,36 +14,107 @@
...
+
+func (om OpenMode) AccessMode() int {
+	return om.accessMode
+}
+
+func (om OpenMode) FileFlags() int {
+	return om.fileFlags
+}
+
+func (om *OpenMode) SetAccessMode(accessMode int) {
```
vadlakondaswetha: remove this, when you have a constructor and we dont expect these values to be changed after initialized, so setters are not required.

internal/fs/fs.go
vadlakondaswetha: I dont see openMode being used here: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/file.go#L597
Am i missing something?

internal/util/file_util.go
anushka567: Added it for completeness, but since open modes will be immutable, we can remove it, Updated.

internal/fs/fs.go
anushka567: We can remove openMode from the method definition here

- Originally, openMode was used to initialize a BWH object when inode.Write() was called.
- Later, the initialization of BWH was moved into a central place in fs.go
- Since BWH is guaranteed to be ready if applicable by the time this method runs, this method doesn't need openMode anymore.

PR #4045, opened 2025-11-20
feat(buffered reader default ON): Improve block implementation use slice for efficiency

meet2mky:
### Description
We are doing a redundant operation of clearing the block buffer upon re-use of buffer which is not required and causes regression on critical buffered read path as it takes around 1000 microseconds of extra work which is very high. This step actually is adding an overhead of around 1000 microseconds on critical buffered Read Path. On further investigating this step is redundant as this buffer slice is always protected by bound checks using `m.offset.end` and `m.offset.start` which are always checked before issuing any read of write on the slice so it's not required to reset each bits to zero.

In this PR we are refactoring the block implementation to use buffer slicing to ensure no overhead in the reuse of buffers and also removing the redundant variables to reduce memory in the blocs as the offset start is always 0 and offset end is always len(buffer). 

b/461362572

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - Part of pre-submits.

### DIFF
--- a/internal/block/prefetch_block.go
+++ b/internal/block/prefetch_block.go
@@ -182,7 +182,7 @@ func (pmb *prefetchMemoryBlock) SetAbsStartOff(startOff int64) error {
 
 	// If absStartOff is already set, then return an error.
 	if pmb.absStartOff >= 0 {
-		return fmt.Errorf("SetAbsStartOff: absStartOff is already set, re-setting is not allowed.")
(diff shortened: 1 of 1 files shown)


meet2mky:
A similar technique is used in `bytes.Buffer.Reset()`: https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/bytes/buffer.go;l=112

raj-prince (approved):

PR #4068, opened 2025-11-25
fix(buffered_read): make awaitReady concurrent safe

raj-prince:
### Description
- Concurrent calls to AwaitReady with one NotifyReady lead to a synchronisation issue. NotifyReady calls add a status to the notification channel that the AwaitReady go-routine listens to.
- After the notification, the channel was closed before the status was updated correctly. Closing the channel was leading to concurrent read to pv.Status and write to pv.Status.
- Fix: update the status first, then close the notification channel to concurrent reads. 

b/460615197

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


meet2mky:
Thanks @raj-prince for fixing this flakiness. Awesome work 💯

PR #4072, opened 2025-11-25
refactor(gcsreader): Use ReadTypeClassifier in GCSReader

vipnydav:
### Description
This PR refactors the logic for classifying read types (sequential vs. random) out of the `GCSReader` and into a new, dedicated `ReadTypeClassifier`. The `ReadTypeClassifier` is now created and managed by the `ReadManager`, and the read information is passed down to the `GCSReader` through the context.

This change centralizes the state and logic for read classification, making the code more modular and easier to maintain and shareable across all readers.

b/462587679

### Testing details
1. Manual - Done
2. Unit tests - Updated

### DIFF
--- a/internal/gcsx/client_readers/gcs_reader.go
+++ b/internal/gcsx/client_readers/gcs_reader.go
@@ -144,10 +144,9 @@ func (gr *GCSReader) read(ctx context.Context, readReq *gcsx.GCSReaderRequest) (
 		// acquire the lock. Hence, needs to be calculated again.
 		// Recalculating only for ZB and only when another read had been performed between now and
 		// the time when readerType was calculated for this request.
-		newReadInfo := readReq.ReadInfo
 		if gr.bucket.BucketType().Zonal && readReq.ExpectedOffset != gr.readTypeClassifier.NextExpectedOffset() {
-			newReadInfo = gr.readTypeClassifier.GetReadInfo(readReq.Offset, readReq.SeekRecorded)
-			reqReaderType = gr.readerType(newReadInfo.ReadType, gr.bucket.BucketType())
+			readReq.ReadInfo = gr.readTypeClassifier.GetReadInfo(readReq.Offset, readReq.SeekRecorded)
+			reqReaderType = gr.readerType(readReq.ReadType, gr.bucket.BucketType())
 		}
 		// If the readerType is range reader after re calculation, then use range reader.
 		// Otherwise fall back to MultiRange Downloder
@@ -156,7 +155,6 @@ func (gr *GCSReader) read(ctx context.Context, readReq *gcsx.GCSReaderRequest) (
 			// Calculate the end offset based on previous read requests.
 			// It will be used if a new range reader needs to be created.
 			readReq.EndOffset = gr.getEndOffset(readReq.Offset)
-			readReq.ReadType = newReadInfo.ReadType
 			readResp, err = gr.rangeReader.ReadAt(ctx, readReq)
 			return readResp.Size, err
 		}


internal/gcsx/read_manager/read_manager.go
```diff
@@ -61,6 +65,8 @@ func NewReadManager(object *gcs.MinObject, bucket gcs.Bucket, config *ReadManage
 	// Create a slice to hold all readers. The file cache reader will be added first if it exists.
 	var readers []gcsx.Reader
 
+	typeClassifier := gcsx.NewReadTypeClassifier(int64(config.SequentialReadSizeMB))
```
raj-prince: Please initialize the variable near the use. 

Also, if we can provide a better name - only type classifier doesn't make sense. Maybe readClassifier or classifier itself.

internal/gcsx/read_manager/read_manager.go
```diff
@@ -143,10 +150,14 @@ func (rr *ReadManager) ReadAt(ctx context.Context, p []byte, offset int64) (gcsx
 		return readResponse, nil
 	}
 
+	readInfo := rr.readTypeClassifier.GetReadInfo(offset, false)
```
raj-prince: You need to write a comment here - why is this required? Also, lets discuss about this in the standup.

internal/gcsx/read_manager/read_manager_test.go
```diff
@@ -331,12 +330,13 @@ func (t *readManagerTest) Test_ReadAt_R1FailsR2Succeeds() {
...
 	mockReader2 := new(gcsx.MockReader)
 	rm := &ReadManager{
-		object:  t.object,
-		readers: []gcsx.Reader{mockReader1, mockReader2},
+		object:             t.object,
+		readers:            []gcsx.Reader{mockReader1, mockReader2},
+		readTypeClassifier: gcsx.NewReadTypeClassifier(sequentialReadSizeInMb),
 	}
-	mockReader1.On("ReadAt", t.ctx, buf, offset).Return(gcsx.ReadResponse{}, gcsx.FallbackToAnotherReader).Once()
+	mockReader1.On("ReadAt", mock.Anything, buf, offset).Return(gcsx.ReadResponse{}, gcsx.FallbackToAnotherReader).Once()
```
raj-prince: Why did we need to change this?

internal/gcsx/reader.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs"
 )
 
+// readInfoKey is the unexported type for the read information in the context.
+type readInfoKey int
+
+// ReadInfoKey is the key for the read information in the context.
+const ReadInfoKey readInfoKey = iota
```
raj-prince: why iota?

internal/gcsx/client_readers/gcs_reader_test.go
```diff
@@ -98,15 +103,14 @@ func (t *gcsReaderTest) Test_NewGCSReader() {
...
-		Config:               nil,
+		MetricHandle:       metrics.NewNoopMetrics(),
+		MrdWrapper:         nil,
+		Config:             nil,
+		ReadTypeClassifier: gcsx.NewReadTypeClassifier(sequentialReadSizeInMb),
 	})
 
 	assert.Equal(t.T(), object, gcsReader.object)
 	assert.Equal(t.T(), t.mockBucket, gcsReader.bucket)
-	assert.Equal(t.T(), metrics.ReadTypeSequential, gcsReader.readType.Load())
```
raj-prince, line 112: We can use the classifier's IsSequential method.

internal/gcsx/read_manager/read_manager_test.go
vipnydav: Because the `ReadManager.ReadAt` method doesn't modify the context it receives. Instead, it creates a new child context using `context.WithValue` before passing it down to the readers.

Because the test doesn't have a reference to this new, internal context, it's impossible to use `t.ctx` in the mock expectation—they are different objects.

internal/gcsx/reader.go
vipnydav: I used iota to create a unique, collision-proof key for safely storing values in a context. Using iota is the standard Go idiom for assigning a constant value (0 in this case) to this key.

But since we would only be creating a single key. we can directly set it to 0. Should I update it or keep it same?

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -185,9 +144,10 @@ func (gr *GCSReader) read(ctx context.Context, readReq *gcsx.GCSReaderRequest) (
 		// acquire the lock. Hence, needs to be calculated again.
 		// Recalculating only for ZB and only when another read had been performed between now and
 		// the time when readerType was calculated for this request.
-		if gr.bucket.BucketType().Zonal && readInfo.expectedOffset != gr.expectedOffset.Load() {
-			readInfo = gr.getReadInfo(readReq.Offset, readInfo.seekRecorded)
-			reqReaderType = gr.readerType(readInfo.readType, gr.bucket.BucketType())
+		newReadInfo := readReq.ReadInfo
```
raj-prince: nit: please keep it readInfo only as it's not always a new readInfo.

raj-prince (approved):

PR #4077, opened 2025-11-26
feat(Unsupported dir path): Fix delete path for HNS bucket

Tulsishah:
### Description
- Ignoring delete object error, as in case of hns there is no way of knowing if underlying placeholder object exists or not in Hierarchical bucket. The DeleteFolder operation handles removing empty folders.

[b/463979343](https://b.corp.google.com/issues/463979343)

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -1070,7 +1070,14 @@ func (d *dirInode) DeleteObjects(ctx context.Context, objectNames []string) erro
 // between our List and Delete calls, we'd get a 'Not Found' error. By ignoring
 // it, we ensure the delete operation succeeds if the object is already gone.
 func (d *dirInode) deleteObject(ctx context.Context, objectName string) error {
+	// For HNS buckets, the directory entry might be backed by:
+	// 1. **Only a folder:** Deleting the 'object' will fail (not found), and the DeleteFolder call is needed.
+	// 2. **A 0-byte placeholder object + an empty folder:** We first attempt to delete the 0-byte object. The subsequent DeleteFolder handles the folder removal.
+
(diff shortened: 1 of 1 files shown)


internal/fs/inode/dir.go
```diff
@@ -1070,6 +1070,17 @@ func (d *dirInode) DeleteObjects(ctx context.Context, objectNames []string) erro
 // between our List and Delete calls, we'd get a 'Not Found' error. By ignoring
 // it, we ensure the delete operation succeeds if the object is already gone.
 func (d *dirInode) deleteObject(ctx context.Context, objectName string) error {
+	err := d.bucket.DeleteObject(ctx, &gcs.DeleteObjectRequest{Name: objectName})
+	if !d.isBucketHierarchical() && err != nil {
```
vadlakondaswetha, line 1081: why are we not ignoring notfound error for hns?

Tulsishah: For HNS buckets

1. Directory can remain only as folder(In such case it will return not found for objects) so we will avoid it and run DeleteFolder command.
2. Directory can remain as 0 byte object + empty folder --> In that case we first need to delete 0 byte object and then delete empty folder. 

The core logic of DeleteFolder is also same. [ref](https://github.com/GoogleCloudPlatform/gcsfuse/blob/7771fd15659fe1dba07fa06dd6d5279dddec139b/internal/fs/inode/dir.go#L1017)

vadlakondaswetha: can you add that as a comment in the code

Tulsishah: Done.

vadlakondaswetha (approved):

PR #4078, opened 2025-11-27
ci(louhi_release): build package from commit_hash with release_version intact

raj-prince:
### Description
- Improve the package build Docker file by adding support to build a package from a commit-hash with name as the provided release_version.
- Changing cd_script to pass commit_hash to the package builder
- Also, added a flag run_light_test to skip the run of all the e2e tests which can take more than 4 hrs. Running with run_light_test takes less than 1m (only e2e test part).

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/cd_scripts/package_gcsfuse.sh
+++ b/tools/cd_scripts/package_gcsfuse.sh
@@ -71,9 +71,9 @@ sudo apt-get install git -y
 sudo apt-get install qemu-user-static binfmt-support
 git clone https://github.com/GoogleCloudPlatform/gcsfuse.git
 cd gcsfuse/tools/package_gcsfuse_docker/
-git checkout "v$RELEASE_VERSION_TAG"
+git checkout "$COMMIT_HASH"
 echo "Building docker for ${architecture} ..."
-sudo docker buildx build --load . -t gcsfuse-release-${architecture}:"$RELEASE_VERSION_TAG" --build-arg GCSFUSE_VERSION="$RELEASE_VERSION" --build-arg ARCHITECTURE=${architecture} --build-arg BRANCH_NAME="v$RELEASE_VERSION_TAG" --build-arg COMMIT_HASH="$COMMIT_HASH" &> docker_${architecture}.log
+sudo docker buildx build --load . -t gcsfuse-release-${architecture}:"$RELEASE_VERSION_TAG" --build-arg GCSFUSE_VERSION="$RELEASE_VERSION" --build-arg ARCHITECTURE=${architecture} --build-arg BRANCH_NAME="$COMMIT_HASH" &> docker_${architecture}.log
 gsutil cp docker_${architecture}.log gs://"$UPLOAD_BUCKET"/v"$RELEASE_VERSION"/
 sudo docker run  -v $HOME/gcsfuse/release:/release gcsfuse-release-${architecture}:"$RELEASE_VERSION_TAG" cp -r /packages/. /release/v"$RELEASE_VERSION"
 
--- a/tools/package_gcsfuse_docker/Dockerfile
+++ b/tools/package_gcsfuse_docker/Dockerfile
@@ -16,8 +16,7 @@
 #
 # Build Arguments:
 #   GCSFUSE_VERSION  - Version string for the package (required, e.g., "0.39.2")
-#   COMMIT_HASH      - Specific git commit to build from (optional)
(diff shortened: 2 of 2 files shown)


tools/cd_scripts/e2e_test.sh
```diff
@@ -563,9 +571,16 @@ function run_e2e_tests_for_emulator_and_log() {
...
 
+
 # Declare an associative array to store the exit status of different test runs.
 declare -A exit_status
-if [[ "$RUN_READ_CACHE_TESTS_ONLY" == "true" ]]; then
+if [[ "$RUN_LIGHT_TEST" == "true" ]]; then
+    light_test_dir_non_parallel=("monitoring")
+    light_test_dir_parallel=("")
+
+    run_e2e_tests "flat"  light_test_dir_parallel light_test_dir_non_parallel false
```
ashmeenkaur, line 581: shall we run on all bucket types so we know that all buckets have been created properly?

tools/package_gcsfuse_docker/Dockerfile
```diff
@@ -40,10 +66,18 @@ ENV GO111MODULE=auto
 ARG GCSFUSE_VERSION
 ARG GCSFUSE_REPO="https://github.com/googlecloudplatform/gcsfuse/"
 ARG BRANCH_NAME="v${GCSFUSE_VERSION}"
```
ashmeenkaur, line 67: Would this Dockerfile have worked if we just passed BRANCH_NAME ?

I agree that COMMIT_HASH is a better name for readability but trying to reduce some toil for patch releases.

tools/cd_scripts/package_gcsfuse.sh
```diff
@@ -74,7 +73,7 @@ git clone https://github.com/GoogleCloudPlatform/gcsfuse.git
 cd gcsfuse/tools/package_gcsfuse_docker/
 git checkout "v$RELEASE_VERSION_TAG"
 echo "Building docker for ${architecture} ..."
-sudo docker buildx build --load . -t gcsfuse-release-${architecture}:"$RELEASE_VERSION_TAG" --build-arg GCSFUSE_VERSION="$RELEASE_VERSION" --build-arg ARCHITECTURE=${architecture} --build-arg BRANCH_NAME="v$RELEASE_VERSION_TAG" &> docker_${architecture}.log
+sudo docker buildx build --load . -t gcsfuse-release-${architecture}:"$RELEASE_VERSION_TAG" --build-arg GCSFUSE_VERSION="$RELEASE_VERSION" --build-arg ARCHITECTURE=${architecture} --build-arg BRANCH_NAME="v$RELEASE_VERSION_TAG" --build-arg COMMIT_HASH="$COMMIT_HASH" &> docker_${architecture}.log
```
ashmeenkaur: I am worried about patch releases. Since tag won't be created for patch release in the new release flow and GCSFuse repo (still at old versions) won't have this change, there is a chance of breaking the release. 

I think one thing that we can do for it is always pick up this script from master in the release flow and then checkout GCSFuse at the provided commit hash and pass RELEASE_VERSION argument to dockerfile always

tools/cd_scripts/package_gcsfuse.sh
```diff
@@ -74,7 +73,7 @@ git clone https://github.com/GoogleCloudPlatform/gcsfuse.git
 cd gcsfuse/tools/package_gcsfuse_docker/
 git checkout "v$RELEASE_VERSION_TAG"
```
ashmeenkaur: checkout commit hash instead

tools/cd_scripts/e2e_test.sh
raj-prince: We can iterate over this by adding new or changing the package. Currently, making this to simplify my testing.

tools/cd_scripts/package_gcsfuse.sh
raj-prince: Yeah. Sounds good. we can achieve using `startup-script-url` while creating gcp vm.

tools/package_gcsfuse_docker/Dockerfile
raj-prince: Yeah, makes sense. We can use the existing one only.

ashmeenkaur (approved):

PR #4085, opened 2025-11-28
chore(known-issues): Add metrics issues to known-issues.

kislaykishore:
### Description
Update known issues with metrics bugs.

N/A

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

No


docs/known-issues.md
```diff
@@ -15,4 +15,7 @@ This document lists known issues and bugs in GCSFuse, their impact, and the rele
 
 | Issue | Affected Versions | Fixed in | Reference |
 | :--- | :--- | :--- | :--- |
-| Input/output error when metrics are enabled. | Applications may receive input/output errors from GCSFuse mounts when metrics are enabled. | v2.12.0 | [#3870](https://github.com/GoogleCloudPlatform/gcsfuse/issues/3870) |
+| Input/output error when metrics are enabled. Applications may receive input/output errors from GCSFuse mounts when metrics are enabled. | v2.11.* | v2.12.0 | [#3870](https://github.com/GoogleCloudPlatform/gcsfuse/issues/3870) |
+| Incorrect gcs/reader_count and gcs/download_bytes_count metrics. gcs/reader_count and gcs/download_bytes_count are unreliable before v3.5. | v3.4 and older | v3.5 | [#3895](https://github.com/GoogleCloudPlatform/gcsfuse/pull/3895)
```
charith87: Can we add the earlier known version that way customers can know if they are impacted or not. ( e.g. v3.0.0 and above ). 

 v3.4 and older keeps it a little vague to know if they are getting affected or not.
Also, we could add "Metrics:" to reduce that its only limited to metrics and not other workflows.

PR #4086, opened 2025-11-28
feat(buffered reader default ON): Reuse read handle to optimize reader creation

meet2mky:
### Description
This pull request introduces a mechanism to reuse GCS read handles within the BufferedReader. By reusing the read handle, we can optimize the creation of subsequent GCS readers for the same GCS object, which is particularly beneficial for zonal buckets as it can bypass authorization and metadata checks.

### Changes

* A new ReadHandle struct is introduced to store the read handle and its expiration time (5 mins)
* BufferedReader now uses an atomic.Pointer to the ReadHandle struct for thread-safe access and updates
* The read handle is updated by the downloadTask after a successful read from a zonal bucket after which it will be refreshed if it's expired.
* Unit tests have been added to verify the behaviour of the read handle logic, including initial state, updates on expiration, and no updates when the handle is still valid.

#### Sequential Reads on RAPID
Master: **upto 4.2 GB/sec**
PR: **upto 4.5 GB/sec**

#### New Reader Creation

Master (AVG: 22ms):
image

PR (AVG: 15ms):
image

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - Part of pre-submit.

### DIFF
--- a/internal/bufferedread/buffered_reader.go
+++ b/internal/bufferedread/buffered_reader.go
@@ -630,7 +630,7 @@ func (br *BufferedReader) UpdateReadHandle(readHandle []byte) {
 	}
 
 	// Atomically store the new ReadHandle, overwriting the old one. We are intentionaly using
-	// Store here instead of CAS as we don't care if anyone other download task updated the
+	// Store here instead of CAS as we don't care if any other download task updated the
 	// atomic pointer from the time we detected the expiration as long as it's updated.
 	br.readHandle.Store(newReadHandle)
 }


internal/bufferedread/buffered_reader.go
```diff
@@ -50,9 +51,16 @@ type BufferedReadConfig struct {
 
 const (
 	defaultPrefetchMultiplier = 2
-	ReadOp                    = "readOp"
+	readOp                    = "readOp"
+	readHandleValidity        = 5 * time.Minute // source go/gcsfuse-zonal-buckets-read-design-doc
```
raj-prince, line 55: Could you please confirm, if this expiry time is client specific? Or server maintains a shared handle and refresh every 5 mins?

meet2mky: Expiry time is based on per buffered Reader instance. I couldn't find any resources apart from it's mention in the source go/gcsfuse-zonal-buckets-read-design-doc

meet2mky:
Closing as buffered reader is not going to be default ON for Zonal.

PR #4088, opened 2025-11-29
fix(metrics): Add missing GCS read metrics to sparse downloads

xinyangge:
## Summary

- Adds missing `metrics.CaptureGCSReadMetrics` call to `downloadSparseRange` in `sparse_downloads_job.go`
- This metric was present in `parallel_downloads_job.go` but missing from sparse downloads
- Uses `ReadTypeRandom` as the read type since sparse reads are on-demand/random reads

## Test plan

- [ ] Verify metrics are captured when sparse downloads occur
- [ ] Confirm no regression in existing metrics functionality

### DIFF
--- a/internal/fs/metrics_test.go
+++ b/internal/fs/metrics_test.go
@@ -367,7 +367,12 @@ func TestSparseReadFile_GCSReadMetrics(t *testing.T) {
 	bucket, server, mh, reader := createTestFileSystemWithMetrics(ctx, t, params)
 	server = wrappers.WithMonitoring(server, mh)
 	fileName := "sparse_test.txt"
-	content := "test content for sparse file download"
+	// Create a file larger than the chunk size (1MB) to test sparse behavior.
+	// With a 3MB file and 1MB chunks, reading from the middle should only
+	// download that chunk, not the entire file.
+	chunkSize := 1024 * 1024 // 1MB chunk size configured in test
+	fileSize := 3 * chunkSize
+	content := string(make([]byte, fileSize))
 	createWithContents(ctx, t, bucket, fileName, content)
 	lookupOp := &fuseops.LookUpInodeOp{
 		Parent: fuseops.RootInodeID,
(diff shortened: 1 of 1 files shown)


xinyangge:
cc @vadlakondaswetha @ashmeenkaur @raj-prince who reviewed the previous PR https://github.com/GoogleCloudPlatform/gcsfuse/pull/3991

internal/cache/file/downloader/sparse_downloads_job.go
kislaykishore, line 1: Can we add a metrics-test [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/metrics_test.go) to ensure that this code is covered?

xinyangge: Done.

internal/fs/metrics_test.go
```diff
@@ -357,6 +360,44 @@ func TestRandomReadFile_FileCacheMetrics(t *testing.T) {
...
+	}
+
+	// First read triggers sparse download from GCS
+	err = server.ReadFile(ctx, readOp)
+	require.NoError(t, err, "ReadFile")
+	waitForMetricsProcessing()
+
+	// Verify GCS read metrics for sparse download with Random read type
+	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/read_count", attribute.NewSet(attribute.String("read_type", string(metrics.ReadTypeRandomAttr))), int64(1))
+	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/download_bytes_count", attribute.NewSet(attribute.String("read_type", string(metrics.ReadTypeRandomAttr))), int64(len(content)))
```
kislaykishore: Could you also add a test where the sparse behavior is actually tested i.e. downloading less than the size of the file?

xinyangge: Done.

kislaykishore:
@raj-prince since you could you also review since you reviewed the sparse-download feature?

PR #4089, opened 2025-12-01
feat(dummy-io-bucket): Adding hidden flag to control dummy-io bucket

raj-prince:
### Description
1. Adding per-byte latency support for NewRangeReader.
2. Fixing perf issue happening because of zeroing the buffer (1.2 GB/s, after removing zeroing code started getting 7.8 GB/s, single threaded sequential read).

### Testing details
1. Manual - Yes.
4. Unit tests - NA
5. Integration tests - NA

### DIFF
--- a/internal/storage/dummy_io_bucket.go
+++ b/internal/storage/dummy_io_bucket.go
@@ -243,9 +243,9 @@ type dummyReader struct {
 // newDummyReader creates a new dummyReader with the specified total length.
 func newDummyReader(totalLen uint64, perMBLatency time.Duration) *dummyReader {
 	return &dummyReader{
-		totalLen:       totalLen,
-		bytesRead:      0,
-		readHandle:     []byte{}, // Always return a non-nil read handle
+		totalLen:   totalLen,
+		bytesRead:  0,
+		readHandle: []byte{}, // Always return a non-nil read handle
 		// TODO: fix precision issue in perByteLatency because of upper bound.
 		perByteLatency: time.Duration(perMBLatency.Nanoseconds()+MB-1) / MB,
 	}


internal/gcsx/bucket_manager.go
```diff
@@ -69,6 +70,11 @@ type BucketConfig struct {
 	TmpObjectPrefix          string
 	// Used in Zonal buckets to determine if objects should be finalized or not.
 	FinalizeFileForRapid bool
+
+	// Enable dummy I/O mode for testing purposes. In this mode all reads and writes are
```
abhishek10004: we're not doing anything for writes currently

internal/storage/dummy_io_bucket.go
```diff
@@ -226,21 +231,22 @@ func (d *dummyIOBucket) GCSName(object *gcs.MinObject) string {
...
-func newDummyReader(totalLen uint64) *dummyReader {
+func newDummyReader(totalLen uint64, perMBLatency time.Duration) *dummyReader {
 	return &dummyReader{
-		totalLen:   totalLen,
-		bytesRead:  0,
-		readHandle: []byte{}, // Always return a non-nil read handle
+		totalLen:       totalLen,
+		bytesRead:      0,
+		readHandle:     []byte{}, // Always return a non-nil read handle
+		perByteLatency: time.Duration(perMBLatency.Nanoseconds()+MB-1) / MB,
```
abhishek10004, line 250: Let's store perMBLatency here and do the calculation and rounding off in read call itself as this can introduce significant differences in expected throughput.

Eg: let's say per mb latency is 3.5ms. Here, we will store perByteLatency as 4ns/byte which will result in a big difference in final throughput.

raj-prince: Thanks! I have added a todo to fix this in the next PR. Requires a bit thinking to do it in right way.

internal/gcsx/bucket_manager.go
```diff
@@ -69,6 +70,11 @@ type BucketConfig struct {
 	TmpObjectPrefix          string
 	// Used in Zonal buckets to determine if objects should be finalized or not.
 	FinalizeFileForRapid bool
+
+	// Enable dummy I/O mode for testing purposes, simulated read without
```
abhishek10004, line 74: nit: simulates

abhishek10004 (approved):
LGTM

internal/gcsx/bucket_manager.go
raj-prince: will fix this in the next PR.

PR #4098, opened 2025-12-02
refactor(gcsreader): Optimize prefetching for regional buckets by recalculating read type

vipnydav:
### Description
This PR removes the zonal bucket check for recalculating the read type. The read type is now recalculated if the expected offset has changed for both zonal and regional buckets. This is important for regional buckets to adjust the prefetching window when the read pattern changes, and for zonal buckets, it allows switching to MRD for high-performance random reads

https://b.corp.google.com/issues/465375267

### Testing details
1. Manual - Done
2. Unit tests - Updated

### Performance Test: Sequential Read

The following table shows the results of a sequential read performance for a regional bucket in `us-west4` test on a 10GiB file, averaged over 3 runs. The test compares the `master` branch with the changes in this PR.

| Branch | Throughput (MiB/s) |
| :--- | :--- |
| `master` | 243.33 |
| `remove-zonal-check` | 304.67 |

**Conclusion:** No regression.


abhishek10004 (approved):
LGTM

Issue #4075, opened 2025-11-26
Add support for includeFoldersAsPrefixes=true when listing directories.


Kleptine:
**Describe the issue**
We have a bucket that contains ~1000 folders, each with ~10 files inside. Hierarchical namespaces is enabled.

Currently, listing this directory when mounted under gcsfuse takes ~3 seconds. Similarly, via a REST request, listing the directory takes 1.7s. 

However, with the new `includeFoldersAsPrefixes=true` argument included in the REST request, listing drops over 10x to 189ms! 

It would be great if gcsfuse would pass `includeFoldersAsPrefixes` when listing mounted directories. 

**System & Version (please complete the following information):**
 - OS: Any (tested on Debian/OSX)
 - Platform: GCE VM
 - Version: 3.4

Tulsishah:
We are already making list request with `includeFoldersAsPrefixes` as true for HNS buckets. [ref](https://github.com/GoogleCloudPlatform/gcsfuse/blob/cc3eca2c150316c4d43cd200bf72daf115e81dff/internal/fs/inode/dir.go#L685)

Kleptine:
Ah, thanks for letting me know. Yeah, I can see that gcsfuse reports a GCS request latency of ~75ms via cloud monitoring metrics, so it looks like that argument is being properly applied.

Unfortunately, the overall command `ls > /dev/null` under gcsfuse still takes 2 seconds. I suppose there is additional processing happening under the hood. Is that expected performance for this case?

I can't figure out why the REST API would return the results in 75ms, but gcsfuse and `ls` would take 2 seconds.

Tulsishah:
The observed latency for gcsfuse ls operations is  higher than a direct GCS API call because it includes the entire round trip: the application's request through the kernel, FUSE layer, and gcsfuse processing, in addition to the actual GCS API latency.

raj-prince:
@Kleptine Could you please run with GCSFuse logs enabled? You can do that using `--log-severity=trace` & `--log-file=logs.txt`.

This will help us in triaging if in-efficiencies are in GCSFuse or kernel.

Kleptine:
@raj-prince Sure, I've captured the log file. Is there a way I can send this to you in a secure way? I'm not fully confident in the data it contains to publish it to the web. I will also attempt to analyze it myself and see if I can discover anything.

Kleptine:
[...]

PR #4099, opened 2025-12-02
fix(HNS): improve folder listing performance by not setting includeTrailingDelimeter

raj-prince:
### Description

Passing `includeTrailingDelimiter=false` while listing the HNS bucket, as folder exists for every explicit directory and no explicit metadata is required. For a flat bucket, `includeTrailingDelimiter=true` is necessary to list an explicit directory.

Note: b/465295359 to confirm the correctness of the fix and see if we can fix server side itself.

We can see a huge perf boost ~20x, while performing listing of 1k folder.

**Before:**
```
Run 1: princer_google_com@tail-latency-test-vm:~$ time ls ~/gcs/a/b/ > /dev/null
real    0m0.648s
user    0m0.001s
sys     0m0.005s

Run 2: princer_google_com@tail-latency-test-vm:~$ time ls ~/gcs/a/b/ > /dev/null
real    0m0.638s
user    0m0.001s
sys     0m0.005s
```

**After:**
```
Run 1: princer_google_com@tail-latency-test-vm:~$ time ls ~/gcs/a/b/ > /dev/null
real    0m0.030s
user    0m0.001s
sys     0m0.004s

Run 2: princer_google_com@tail-latency-test-vm:~$ time ls ~/gcs/a/b/ > /dev/null
real    0m0.031s
user    0m0.003s
sys     0m0.003s
```

b/463907985
https://github.com/GoogleCloudPlatform/gcsfuse/issues/4075

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -685,10 +685,9 @@ func (d *dirInode) readObjects(
 	includeTrailingDelimeter := true // Important for flat bucket to list explicit directory.
 	if d.isBucketHierarchical() {
 		d.includeFoldersAsPrefixes = true
-		// Not needed for hierarchical bucket.
-		// And setting it to true causes performance issues - b/465295359
-		includeTrailingDelimeter = false // Not needed for hierarchical bucket.
-
+		// Listing explicit directories are not required for HNS bucket as already listed as
+		// folder resources and setting it to true causes perf issue (b/465295359)
+		includeTrailingDelimeter = false
 	}
 	// Ask the bucket to list some objects.
 	req := &gcs.ListObjectsRequest{


kislaykishore:
The "before" time in the PR description is better than the "after" time. Did we mistakenly swap them?

raj-prince:
> The "before" time in the PR description is better than the "after" time. Did we mistakenly swap them?

Yeah, corrected.

internal/fs/inode/dir.go
```diff
@@ -681,13 +681,18 @@ func (d *dirInode) ReadDescendants(ctx context.Context, limit int) (map[Name]*Co
...
 		d.includeFoldersAsPrefixes = true
+		// Listing explicit directories are not required for HNS bucket as already listed as
+		// folder resources and setting it to true causes perf issue (b/465295359)
+		includeTrailingDelimeter = false
 	}
 	// Ask the bucket to list some objects.
 	req := &gcs.ListObjectsRequest{
 		Delimiter:                "/",
-		IncludeTrailingDelimiter: true,
+		IncludeTrailingDelimiter: includeTrailingDelimeter,
```
Tulsishah, line 695: Can we make changes [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/dir.go#L1106) as well?

raj-prince: Not required. 

We are not setting `includeTrailingDelimiter=true`, so already being taken care of this optimization.

Issue #4066, opened 2025-11-24
Packaging issue: unexpected /DEBIAN folder


parametalol:
**Describe the issue**

The rpm package creates `/DEBIAN` folder on Fedora, with the `control` file in it.
The folder is not removed when the package is uninstalled.

**System & Version (please complete the following information):**

 - OS: Fedora 43
 - Platform: amd64 laptop
 - Version: 3.4.4

**Steps to reproduce the behavior with following information:**
1. Install the package
2. Observe the /DEBIAN folder

**Additional context**

NA

**SLO:**
We strive to respond to all bug reports within 24 business hours provided the information mentioned above is included.

abhishek10004:
Hi @parametalol, 

Thank you for reporting this issue. We have confirmed that this extra folder is being packaged in the RPM. We are currently implementing a change to fix this packaging issue. In the interim, you can safely delete the folder.

abhishek10004:
Closing the issue as I've merged the change in master.

PR #4100, opened 2025-12-02
fix(build): Excluding DEBIAN folder while building rpm package

abhishek10004:
### Description
Excluding DEBIAN folder while building rpm package to ensure it's not packaged in the RPM binary.

b/463363713
https://github.com/GoogleCloudPlatform/gcsfuse/issues/4066

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/package_gcsfuse_docker/Dockerfile
+++ b/tools/package_gcsfuse_docker/Dockerfile
@@ -109,7 +109,6 @@ WORKDIR ${GCSFUSE_PKG}
 # Build the package
 RUN dpkg-deb --build ${GCSFUSE_BIN}
 RUN mv ${GCSFUSE_BIN}.deb .
-RUN rm -rf ${GCSFUSE_BIN}/DEBIAN
 RUN fpm \
     -s dir \
     -t rpm \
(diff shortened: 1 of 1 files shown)


tools/package_gcsfuse_docker/Dockerfile
```diff
@@ -109,6 +109,7 @@ WORKDIR ${GCSFUSE_PKG}
 # Build the package
 RUN dpkg-deb --build ${GCSFUSE_BIN}
 RUN mv ${GCSFUSE_BIN}.deb .
+RUN rm -rf ${GCSFUSE_BIN}/DEBIAN
```
raj-prince: Don't we have something like `--exclude-folder` option in fpm? `rm -rf` seems scary a bit.

abhishek10004: Yes, there's an option for excluding folder in fpm but this is also safe as we're just removing the DEBIAN folder and that too after the debian package has been built.

raj-prince (approved):

PR #4118, opened 2025-12-05
test(file cache parallel reads): Enabling parallel reads from file cache

abhishek10004:
1. Enabled concurrent read tests for file cache as well
2. Updated the test name for one of the e2e tests

b/465027261

### Testing details
1. Manual - NA
3. Unit tests - NA
4. Integration tests - Done


tools/integration_tests/concurrent_operations/concurrent_read_test.go
```diff
@@ -269,9 +273,12 @@ func TestConcurrentRead(t *testing.T) {
 		return
 	}
 
+	cacheDirPath = fmt.Sprintf("/tmp/gcsfuse-file-cache-concurrent-read-%s", setup.GenerateRandomString(5))
```
raj-prince: Should we take the tmp directly programmatically as this can varied - /tmp vs /var/tmp or can vary based on platform too?

PR #4134, opened 2025-12-10
feat(rapid): tailing reads with o-direct

anushka567:
### Description
This PR contains the code changes to enable tailing reads for unfinalized objects when real-time appends is in progress. When the file handle is opened in o-direct mode and reads are attempted for an unfinalized object,if the read request is contained entirely within the cached object size , then the normal flow of reads is triggered, i.e. for a sequential read , range reader will be used to serve the request and for random request, then the MRD flow is invoked.

However if the request  is not entirely contained within the cached object size, then the read request goes via the MRD flow where all size related checks are ignored.

Other implementation justification is given [here](https://docs.google.com/document/d/1x3VqRh31TDraAkIr6q-HdIRwyAJumi5RjWDDg9xhsqk/edit?usp=sharing)

[b/463693893](b/463693893)

### Testing details
1. Manual - As specified in the linked document
4. Unit tests - Added

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2787,7 +2787,12 @@ func (fs *fileSystem) ReleaseDirHandle(
 func (fs *fileSystem) OpenFile(
 	ctx context.Context,
 	op *fuseops.OpenFileOp) (err error) {
-	// Bypass the kernel's page cache for file reads and writes
+	// For file handles opened in O_DIRECT mode, we must set UseDirectIO.
+	// This tells the kernel to bypass its page cache (for file reads and
+	// writes) and **size checks**, forwarding read requests directly to
+	// GCSFuse. This is critical for tailing reads on remotely updated files.
+	// This should be enabled for all file handles being opened, if direct
+	// IO is configured at file system level.
 	if fs.newConfig.FileSystem.ExperimentalODirect || op.OpenFlags.IsDirect() {
 		op.UseDirectIO = true
 	}
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -399,10 +399,18 @@ func (fh *FileHandle) OpenMode() util.OpenMode {
 // the known object size. This allows reading from an object that is being
 // concurrently written to.
 func (fh *FileHandle) shouldSkipSizeChecks(req *gcsx.ReadRequest) bool {
-	isUnfinalized := fh.readManager.Object().IsUnfinalized()
-	isDirectIO := fh.OpenMode().IsDirect()
-	isPositiveOffset := req.Offset >= 0
-	extendsBeyondSize := req.Offset+int64(len(req.Buffer)) > int64(fh.readManager.Object().Size)
+	if !fh.readManager.Object().IsUnfinalized() {
+		return false
+	}
+	if !fh.OpenMode().IsDirect() {
+		return false
+	}
+	if req.Offset < 0 {
+		return false
+	}
+	if req.Offset+int64(len(req.Buffer)) <= int64(fh.readManager.Object().Size) {
+		return false
+	}
 
-	return isUnfinalized && isDirectIO && isPositiveOffset && extendsBeyondSize
+	return true
 }
--- a/internal/fs/handle/file_test.go
+++ b/internal/fs/handle/file_test.go
@@ -1059,12 +1059,12 @@ func (t *fileTest) Test_ShouldSkipSizeChecks() {
 			expectedToSkip: true,
 		},
 		{
-			name:              "Nil read manager: should panic",
-			object:            unfinalizedObject,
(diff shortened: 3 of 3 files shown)


internal/fs/handle/file.go
```diff
@@ -389,3 +393,16 @@ func (fh *FileHandle) isValidReader() bool {
...
+// This is true for direct I/O reads on unfinalized objects that extend beyond
+// the known object size. This allows reading from an object that is being
+// concurrently written to.
+func (fh *FileHandle) shouldSkipSizeChecks(req *gcsx.ReadRequest) bool {
+	isUnfinalized := fh.readManager.Object().IsUnfinalized()
+	isDirectIO := fh.OpenMode().IsDirect()
+	isPositiveOffset := req.Offset >= 0
+	extendsBeyondSize := req.Offset+int64(len(req.Buffer)) > int64(fh.readManager.Object().Size)
+
+	return isUnfinalized && isDirectIO && isPositiveOffset && extendsBeyondSize
```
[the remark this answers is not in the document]
anushka567: Highly unlikely since the buffer is at max 1MB (limited by kernel), the overflow to happen requires the offset to be in order of petabyte.

internal/gcsx/client_readers/gcs_reader.go
```diff
@@ -85,7 +85,7 @@ func shouldRetryForShortRead(err error, bytesRead int, p []byte, offset int64, o
 		return false
 	}
 
-	if offset+int64(bytesRead) >= int64(objectSize) {
+	if offset+int64(bytesRead) >= int64(objectSize) && !skipSizeChecks {
 		return false
 	}
```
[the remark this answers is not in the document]
anushka567: Same as above.

internal/fs/fs.go
```diff
@@ -2788,7 +2788,7 @@ func (fs *fileSystem) OpenFile(
 	ctx context.Context,
 	op *fuseops.OpenFileOp) (err error) {
 	// Bypass the kernel's page cache for file reads and writes
-	if fs.newConfig.FileSystem.ODirect {
+	if fs.newConfig.FileSystem.ODirect || op.OpenFlags.IsDirect() {
```
vadlakondaswetha: can you add a comment why is this required.

internal/fs/handle/file.go
```diff
@@ -389,3 +393,16 @@ func (fh *FileHandle) isValidReader() bool {
 func (fh *FileHandle) OpenMode() util.OpenMode {
 	return fh.openMode
 }
+
+// shouldSkipSizeChecks determines if the read request should skip object size checks.
+// This is true for direct I/O reads on unfinalized objects that extend beyond
+// the known object size. This allows reading from an object that is being
+// concurrently written to.
+func (fh *FileHandle) shouldSkipSizeChecks(req *gcsx.ReadRequest) bool {
+	isUnfinalized := fh.readManager.Object().IsUnfinalized()
```
vadlakondaswetha: nit: generally we do short circuits i.e, as soon as we find a single condition is not true, we return from there. no point in evaluating all these conditions and adding a big if at the end. i.e,

if !fh.readManager.Object().IsUnfinalized() {
  return
}

if !fh.openMode().IsDirect() {
   return
}

internal/fs/handle/file_test.go
```diff
@@ -986,6 +986,113 @@ func (t *fileTest) Test_ReadWithReadManager_FullReadSuccessWithBufferedRead() {
...
+			offset:         100,
+			bufferSize:     10, // 100 + 10 > 100
+			expectedToSkip: true,
+		},
+		{
+			name:              "Nil read manager: should panic",
+			object:            unfinalizedObject,
+			openMode:          directIOReadMode,
+			offset:            100,
+			bufferSize:        10,
```
vadlakondaswetha: what about offset < size but offset+buff > size

internal/fs/handle/file_test.go
```diff
@@ -986,6 +986,113 @@ func (t *fileTest) Test_ReadWithReadManager_FullReadSuccessWithBufferedRead() {
...
+			useNilReadManager: true,
```
vadlakondaswetha: is this even a valid testcase? There should be no panics in the code ideally we should have handled the nil condition in the code or if we are sure it will never be nil dont add this testcase

internal/fs/handle/file_test.go
anushka567: Same behavior as this case, I have added an explicit test case for this.

internal/fs/handle/file_test.go
anushka567: Removing this testcase since this is redundant, In the code flow, we ensure that the read manager is non nil before this boolean method is invoked.

vadlakondaswetha (approved):

PR #4135, opened 2025-12-10
feat(buffered reader default ON): Improve copy to prefetch blocks from reader using ReadFrom interface implementation.

meet2mky:
### Description
Previously, when data was downloaded, it was copied from the network into an intermediate buffer of 32KB before being written to the final destination block. By implementing the io.ReaderFrom interface on the prefetchMemoryBlock, the data can be read directly from the source into the destination block's buffer, bypassing the intermediate buffer and the associated overhead.
The prefetchMemoryBlock now has a ReadFrom method. The io.Copy function in Go is optimized to detect if the destination implements the io.ReaderFrom interface. If it does, io.Copy will call the ReadFrom method directly, which is more efficient.

b/467631195

### Performance Improvement measurement using Dummy IO

image

### Note-

I have verified manually that both SRC doesn't implement io.WriterTo and DST doesn't implement io.ReaderFrom interface in dummy as well as rapid path.

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - Presubmit Tests


internal/block/prefetch_block.go
```diff
@@ -238,3 +238,28 @@ func (pmb *prefetchMemoryBlock) DecRef() bool {
...
+func (pmb *prefetchMemoryBlock) ReadFrom(r io.Reader) (n int64, err error) {
+	for {
+		// return if buffer is full.
+		if len(pmb.buffer) == cap(pmb.buffer) {
+			return n, nil
+		}
+
+		// Read into the remaining capacity of the buffer.
+		m, e := r.Read(pmb.buffer[len(pmb.buffer):cap(pmb.buffer)])
+		if m > 0 {
```
abhishek10004: nit: bytesRead instead of m

internal/block/prefetch_block.go
```diff
@@ -238,3 +238,28 @@ func (pmb *prefetchMemoryBlock) DecRef() bool {
...
+// to efficiently read from reader avoiding an intermediate buffer.
+func (pmb *prefetchMemoryBlock) ReadFrom(r io.Reader) (n int64, err error) {
+	for {
+		// return if buffer is full.
+		if len(pmb.buffer) == cap(pmb.buffer) {
+			return n, nil
+		}
+
+		// Read into the remaining capacity of the buffer.
+		m, e := r.Read(pmb.buffer[len(pmb.buffer):cap(pmb.buffer)])
```
abhishek10004: define m above & then use err here directly instead of e (minimizing extra allocation)

internal/block/prefetch_block.go
```diff
@@ -238,3 +238,28 @@ func (pmb *prefetchMemoryBlock) DecRef() bool {
...
+		if m > 0 {
+			pmb.buffer = pmb.buffer[:len(pmb.buffer)+m]
+			n += int64(m)
+		}
+		if e == io.EOF {
```
abhishek10004: just set err to nil here if err is io.EOF and then no need to return n & err explicitly

internal/block/prefetch_block_test.go
```diff
@@ -369,3 +370,95 @@ func (testSuite *PrefetchMemoryBlockTest) TestPrefetchMemoryBlockDecRefPanics()
...
+
+func (r *errorReader) Read(p []byte) (n int, err error) {
+	return 0, r.err
+}
+
+func (testSuite *PrefetchMemoryBlockTest) TestReadFrom_ReaderHasLessDataThanBufferCapacity() {
+	pmb, err := createPrefetchBlock(10)
+	require.NoError(testSuite.T(), err)
+
+	n, err := pmb.(*prefetchMemoryBlock).ReadFrom(bytes.NewReader([]byte("hello")))
```
abhishek10004: Can we create a string which stores the content and then use that in all the places in the same test (while creating reader as well while asserting length/content)?

abhishek10004: We should do this for all the tests

internal/block/prefetch_block.go
meet2mky: Done

internal/block/prefetch_block.go
meet2mky: Done

internal/block/prefetch_block.go
meet2mky: Done

internal/block/prefetch_block_test.go
meet2mky: Done.

internal/block/prefetch_block.go
```diff
@@ -238,3 +238,30 @@ func (pmb *prefetchMemoryBlock) DecRef() bool {
...
+		// Read into the remaining capacity of the buffer.
+		bytesRead, err = r.Read(pmb.buffer[len(pmb.buffer):cap(pmb.buffer)])
+		if bytesRead > 0 {
+			pmb.buffer = pmb.buffer[:len(pmb.buffer)+bytesRead]
+			n += int64(bytesRead)
+		}
+		if err == io.EOF {
+			// End of reader, we are done.
+			err = nil
+			return
```
raj-prince: nit: Lets avoid naked return for medium size function - https://go.dev/wiki/CodeReviewComments#naked-returns

be explicit while returning
```suggestion
			return n, nil
```

similarly for below return.

internal/bufferedread/download_task.go
```diff
@@ -98,9 +98,13 @@ func (p *downloadTask) Execute() {
 	}
 	defer newReader.Close()
 
-	n, err = io.CopyN(p.block, newReader, int64(end-start))
+	n, err = io.Copy(p.block, newReader)
```
raj-prince: Given CopyN also uses ReaderFrom if destination implements it, do we need any change here?

CopyN doc: https://pkg.go.dev/io#CopyN

raj-prince (commented):
LGTM overall, couple of comments.

internal/block/prefetch_block.go
meet2mky: Done

internal/bufferedread/download_task.go
meet2mky: io.CopyN is wrapper over Copy with Limit Reader overhead. Copy is better as it doesn't have overhead of creating Limit Reader. I am inclined to  keep these changes. Let me know if you think otherwise.

PR #4152, opened 2025-12-13
ci(e2e): Improve bucket deletion in e2e scripts.

meet2mky:
### Description
Improve bucket deletion in e2e scripts. Buckets would be eventually deleted after 30 days giving significant time to debug in case of failure test runs. This will also ensure buckets would always be deleted if some of runs fail to delete buckets due to errors etc.

b/468595921

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/improved_run_e2e_tests.sh
+++ b/tools/integration_tests/improved_run_e2e_tests.sh
@@ -333,22 +333,31 @@ create_bucket() {
 # Helper method to cleanup expired buckets.
 cleanup_expired_buckets() {
     log_info "Deleting buckets older than ${BUCKET_RETENTION_PERIOD_DAYS} days having prefix ${BUCKET_PREFIX}."
-    local -a bucket_uris
-    mapfile -t bucket_uris < <(gcloud storage buckets list --project=${PROJECT_ID} "gs://${BUCKET_PREFIX}*" \
+    local bucket_list_output
+    bucket_list_output=$(gcloud storage buckets list --project=${PROJECT_ID} "gs://${BUCKET_PREFIX}*" \
         --filter="creation_time < -P${BUCKET_RETENTION_PERIOD_DAYS}D" \
         --format="value(storage_url)")
-    if (( ${#bucket_uris[@]} == 0 )); then
+    if [[ $? -ne 0 ]]; then
+        log_error "Failed to list expired buckets. Cleanup of expired buckets is skipped."
+        return 1
+    fi
+
+    if [[ -z "$bucket_list_output" ]]; then
         log_info "No expired buckets found matching the criteria."
         return 0
     fi
+
+    local -a bucket_uris
+    mapfile -t bucket_uris <<< "$bucket_list_output"
+
     log_info "Attempting to delete ${#bucket_uris[@]} expired buckets."
     local bucket_deletion_logs
     bucket_deletion_logs=$(mktemp "/tmp/${TMP_PREFIX}_bucket_deletion_log.XXXXXX")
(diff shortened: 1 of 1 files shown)


tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -61,7 +61,8 @@ readonly INTEGRATION_TEST_PACKAGE_DIR="./tools/integration_tests"
 readonly INTEGRATION_TEST_PACKAGE_TIMEOUT_IN_MINS=90 
 readonly TMP_PREFIX="gcsfuse_e2e"
 readonly ZONAL_BUCKET_SUPPORTED_LOCATIONS=("us-central1" "us-west4")
-readonly DELETE_BUCKET_PARALLELISM=10 # Controls how many buckets are deleted in parallel.
+# e2e buckets created are retained for upto 30 days before deletion.
+readonly BUCKET_RETENTION_PERIOD_DAYS=30
```
Tulsishah: I feel 30 days is more, we should keep it less like 10.

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -325,23 +325,27 @@ create_bucket() {
...
+        --filter="creation_time < -P${BUCKET_RETENTION_PERIOD_DAYS}D" \
+        --format="value(storage_url)")
+    if (( ${#bucket_uris[@]} == 0 )); then
+        log_info "No expired buckets found matching the criteria."
+        return 0
+    fi
+    log_info "Attempting to delete ${#bucket_uris[@]} expired buckets."
+    local bucket_deletion_logs
+    bucket_deletion_logs=$(mktemp "/tmp/${TMP_PREFIX}_bucket_deletion_log.XXXXXX")
+    gcloud storage rm -r "${bucket_uris[@]}" > "$bucket_deletion_logs" 2>&1
```
Tulsishah: What if this command failed? won't it cause false failure in test status? Should we print log in case of failure?

tools/integration_tests/improved_run_e2e_tests.sh
meet2mky: Okay, Updated.

tools/integration_tests/improved_run_e2e_tests.sh
meet2mky: we don't use `set -e` option here so it won't fail the whole script. I have updated the logs to print based on failure or success.

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -325,23 +325,27 @@ create_bucket() {
...
+    log_info "Bucket cleanup complete."
 }
```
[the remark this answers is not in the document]
meet2mky: If it fails on listing it would eventually will be cleaned up in next runs.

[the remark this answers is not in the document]
Tulsishah: I am not sure if it is good practice to avoid failures. As in case of genuine error(e.g. gcloud version error) we will not get notified and it will constantly fail in the upcoming runs.

Tulsishah (approved):

tools/integration_tests/improved_run_e2e_tests.sh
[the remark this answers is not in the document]
meet2mky: Okay, Make sense. I think we can update the listing also.

Tulsishah (approved):

PR #4153, opened 2025-12-14
feat(mounting): Add warning log during mount based on mount time threshold

meet2mky:
### Description
Same as title.

b/457266735

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - Presubmit

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -59,7 +59,7 @@ const (
 	MountSlownessMessage           = "Mount slowness detected: mount time %v exceeded threshold %v"
 	DynamicMountFSName             = "gcsfuse"
 	WaitTimeOnSignalReceive        = 30 * time.Second
-	MountTimeThreshold             = 5 * time.Second
+	MountTimeThreshold             = 8 * time.Second
 )
 
 ////////////////////////////////////////////////////////////////////////


cmd/legacy_main.go
```diff
@@ -482,10 +484,15 @@ func Mount(mountInfo *mountInfo, bucketName, mountPoint string) (err error) {
...
+		startTime := time.Now()
 		err = daemonize.Run(path, args, env, os.Stdout, stderrFile)
 		if err != nil {
 			return fmt.Errorf("daemonize.Run: %w", err)
 		}
+		mountDuration := time.Since(startTime)
+		if mountDuration > MountTimeThreshold {
+			logger.Warnf(MountSlownessMessage, mountDuration, MountTimeThreshold)
+		}
 		logger.Infof(SuccessfulMountMessage)
```
abhishek10004, line 496: Should we add mount time in successful message as well?

abhishek10004 (approved):
LGTM

cmd/legacy_main.go
meet2mky: It's a good suggestion but in 95% of the cases it would become redundant don't you think because it's going to be always within the mount timings. Open for suggestions

abhishek10004: My intention was to get a sense of what is the normal mount time for cases where it mounts before 5s as well. But we can maybe use a metric for that. Let's discuss this offline. 
Let's go ahead with this change for now.

cmd/legacy_main.go
```diff
@@ -482,10 +484,15 @@ func Mount(mountInfo *mountInfo, bucketName, mountPoint string) (err error) {
 			}
 		}
 		// Run.
+		startTime := time.Now()
```
raj-prince, line 487: Shouldn't adding at one place (for foreground mode scenario) be sufficient?

meet2mky: I think it's fine to log at both places for consistency. It's going to be logged in rare scenarios so both place logging should be okay.

meet2mky: Discussed offline: There are no duplicate logs or double logs if it was interpreted in that regards. In case of background mode one instance prints it on the Stdout and other instance prints it to either syslog or log file.

cmd/legacy_main.go
```diff
@@ -56,8 +56,10 @@ import (
 const (
 	SuccessfulMountMessage         = "File system has been successfully mounted."
 	UnsuccessfulMountMessagePrefix = "Error while mounting gcsfuse"
+	MountSlownessMessage           = "Mount slowness detected: mount time %v exceeded threshold %v"
 	DynamicMountFSName             = "gcsfuse"
 	WaitTimeOnSignalReceive        = 30 * time.Second
+	MountTimeThreshold             = 5 * time.Second
```
raj-prince: Lets use 8 seconds, aline with our mount-timeout-test. Ref: https://github.com/GoogleCloudPlatform/gcsfuse/blob/a2554d6cbe8a2462e2f884c775b6f2fe5f604b64/tools/integration_tests/mount_timeout/mount_timeout_test.go#L68

meet2mky: Agree

meet2mky: Done!

raj-prince (approved):

PR #4154, opened 2025-12-15
test(metrics): add file cache metrics tests for disabled range read caching

alleaditya:
### Description
This change introduces new unit tests in [internal/fs/metrics_test.go](cci:7://file:///usr/local/google/home/alleaditya/workspace/gcsfuse/internal/fs/metrics_test.go:0:0-0:0) to verify file cache metrics behavior specifically when `enableFileCacheForRangeRead` is set to false. It ensures that read latencies and other relevant metrics are correctly recorded even when range read caching is disabled.

[463220507](https://b.corp.google.com/issues/463220507)

### Testing details
1. Manual - NA
2. Unit tests - Added comprehensive test cases in [internal/fs/metrics_test.go](cci:7://file:///usr/local/google/home/alleaditya/workspace/gcsfuse/internal/fs/metrics_test.go:0:0-0:0) covering sequential and random reads with range read caching disabled.
3. Integration tests - NA

No

### DIFF
--- a/internal/fs/metrics_test.go
+++ b/internal/fs/metrics_test.go
@@ -274,7 +274,7 @@ func TestSequentialReadFile_FileCacheMetrics(t *testing.T) {
 	)
 }
 
-func TestSequentialReadFile_FileCacheMetrics_DisabledRangeRead(t *testing.T) {
+func TestSequentialReadFile_FileCacheMetrics_DisabledFileCacheForRangeRead(t *testing.T) {
 	ctx := context.Background()
 	params := defaultServerConfigParams()
 	params.enableFileCache = true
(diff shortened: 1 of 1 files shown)


internal/fs/metrics_test.go
```diff
@@ -271,6 +274,80 @@ func TestSequentialReadFile_FileCacheMetrics(t *testing.T) {
 	)
 }
 
+func TestSequentialReadFile_FileCacheMetrics_DisabledRangeRead(t *testing.T) {
```
```suggestion
func TestSequentialReadFile_FileCacheMetrics_DisabledFileCacheForRangeRead(t *testing.T) {
```

kislaykishore: Could you run this test a 100 times using `go test -count=100 ...` to ensure that there is no race?

internal/fs/metrics_test.go
```diff
@@ -360,6 +436,67 @@ func TestRandomReadFile_FileCacheMetrics(t *testing.T) {
 	)
 }
 
+func TestRandomReadFile_FileCacheMetrics_DisabledRangeRead(t *testing.T) {
```
```suggestion
func TestRandomReadFile_FileCacheMetrics_DisabledFileCacheOnRangedRead(t *testing.T) {
```

internal/fs/metrics_test.go
alleaditya: I was able to run this in my console and did not see any issue.
https://paste.googleplex.com/6242185806741504

PR #4157, opened 2025-12-15
test(rapid): E2E tests for tailing reads with o-direct

anushka567:
### Description
This PR introduces end-to-end tests for the changes in PR #4134, which enables tailing reads on unfinalized objects in GCSFuse with zonal buckets.

[b/468140077](b/468140077)
### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - Added and automated.

### DIFF
--- a/tools/integration_tests/unfinalized_object/setup_test.go
+++ b/tools/integration_tests/unfinalized_object/setup_test.go
@@ -75,6 +75,11 @@ func TestMain(m *testing.M) {
 	testEnv.cfg = &cfg.UnfinalizedObject[0]
 	testEnv.bucketType = setup.TestEnvironment(testEnv.ctx, testEnv.cfg)
 
+	if !setup.IsZonalBucketRun() {
+		log.Printf("This test is only for Zonal buckets.")
+		os.Exit(0)
+	}
+
 	// 2. Create storage client before running tests.
 	var err error
 	testEnv.storageClient, err = client.CreateStorageClient(testEnv.ctx)
--- a/tools/integration_tests/unfinalized_object/unfinalized_read_test.go
+++ b/tools/integration_tests/unfinalized_object/unfinalized_read_test.go
@@ -98,7 +98,7 @@ func (t *unfinalizedObjectReads) setupAndAppend(filePath string, initialSize, ap
 	obj, err := t.storageClient.Bucket(setup.TestBucket()).Object(path.Join(testDirName, t.fileName)).Attrs(t.ctx)
 	require.NoError(t.T(), err)
 
-	writer, err := client.TakeoverWriter(t.ctx, t.storageClient, path.Join(testDirName, t.fileName), obj.Generation)
+	writer, err := client.AppendableWriter(t.ctx, t.storageClient, path.Join(testDirName, t.fileName), obj.Generation)
 	require.NoError(t.T(), err)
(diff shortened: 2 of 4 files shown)


internal/storage/debug_bucket.go
```diff
@@ -381,7 +381,6 @@ func (b *debugBucket) NewMultiRangeDownloader(
 	// Call through.
 	mrd, err = b.wrapped.NewMultiRangeDownloader(ctx, req)
 	if err != nil {
-		b.finishRequest(id, desc, start, &err)
```
anushka567, line 384: Removed since duplicate, there is already a defer statement above which logs the request.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -497,6 +497,14 @@ func AppendableWriter(ctx context.Context, client *storage.Client, object string
 	return wc, nil
 }
 
+func TakeoverWriter(ctx context.Context, client *storage.Client, object string, gen int64) (*storage.Writer, error) {
```
vadlakondaswetha: what is the difference between this and above method.
shouldn't this be appendable writer since we are appending to an existing object and above one is just creatWriter since we are creating a new object?

tools/integration_tests/unfinalized_object/unfinalized_read_test.go
```diff
@@ -83,6 +135,99 @@ func (t *unfinalizedObjectReads) TestUnfinalizedObjectsCanBeRead() {
 	assert.Equal(t.T(), writtenContent, string(readContent))
 }
 
+func (t *unfinalizedObjectReads) TestReadRemotelyModifiedUnfinalizedObject() {
+	if !setup.IsZonalBucketRun() {
```
vadlakondaswetha: why is this condition at test level? Since unfinalized is applicable only to zonal bucket shouldn't this be at setup_test.go file?

tools/integration_tests/util/client/storage_client.go
anushka567: As you mentioned, one is for creating new object and other is appending to exisitng generation. 

I have renamed the TakeoverWriter method to AppendableWriter and AppendableWriter to NewWriterWithPreconditionsSet(NewWriter already exists and this method is a wrapper over it)

tools/integration_tests/unfinalized_object/unfinalized_read_test.go
anushka567: Agreed, updated.

tools/integration_tests/unfinalized_object/setup_test.go
```diff
@@ -75,6 +75,11 @@ func TestMain(m *testing.M) {
 	testEnv.cfg = &cfg.UnfinalizedObject[0]
 	testEnv.bucketType = setup.TestEnvironment(testEnv.ctx, testEnv.cfg)
 
+	if !setup.IsZonalBucketRun() {
```
vadlakondaswetha, line 78: shouldn't this be after line 50?

anushka567: Due to changes related to config file based testing , setup.TestEnvironment (line 76) populates this field now. So, makes sense to keep it here.

PR #4162, opened 2025-12-15
feat(tailing reads): update existing inode instead of minting when only size changes

anushka567:
### Description
This PR improves support for `tail -f` and other streaming workloads on Zonal buckets.

Currently, GCSFuse creates a new inode whenever a file's size changes, which prevents applications from seeing new content as it's appended. This change modifies the behavior to update the size of the existing inode in-place if the object's generation has not changed, ensuring readers can see the most up-to-date file size in a timely manner.

[b/468968721](b/468968721)

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - Automated for any breaking change. E2E tests for scenario in a follow-up PR.

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -989,7 +989,7 @@ func (fs *fileSystem) lookUpOrCreateInodeIfNotStale(ic inode.Core) (in inode.Ino
 		// The existing inode has the same generation but a different size.
 		// Update the size and return the existing inode.
 		if cmp == 2 {
-			logger.Warnf("The size of object has changed remotely. Updating the existing inode to reflect the size change.\n")
+			logger.Warnf("The size of object has changed remotely at the same generation. Updating the existing inode to reflect the size change.\n")
 			existingInode.UpdateSize(oGen.Size)
 			in = existingInode
 			return


internal/fs/inode/inode.go
```diff
@@ -118,7 +121,7 @@ func (latest Generation) Compare(current Generation) int {
 	// generation or metageneration, the following case applies exclusively to
 	// zonal buckets.
 	if latest.Size > current.Size {
-		return 1
+		return 2
```
vadlakondaswetha, line 124: General interface of compare is if you zero the objects are same, >0 represents newer object is greater and <0 represents opposite. 
I would avoid keeping conditions on getting a specific integer value. Instead can we just compare the generation in compare method and do the size check directly in fs.go i.e,

if cmp == 0 {

  if latest.size > current.size {
     //updateSize
}

in = existingInode
return

}

anushka567: Since this was a custom struct (Generation) which we maintain and for ease of implementation, I thought it would be better to handle all conditions in the compare method itself with a disclaimer for use, so it is abstracted. but ideally it makes sense to have the return value -1,0, 1 

If we return 0 in compare method for our case (size changed, generation, metageneration remains same), [clobber check](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/inode/file.go#L250) will start returning false.  Semantically, does the scenario where only size has changed remotely mean clobbered or not?

anushka567: As discussed offline, we will continue to throw clobbered error in case of size change and to identify it uniquely, the compare method will return 2 in cases when the size has changed but the generation, metageneration still remains the same.

PR #4163, opened 2025-12-16
feat(fuse_settings): adding hidden config for max_background and congestion_threshold

raj-prince:
### Description
This paper offers a clear explanation about the `max_background` and `congestion_threshold`: <https://www.usenix.org/system/files/conference/fast17/fast17-vangoor.pdf>

Overall summary: 
- Async read-ahead is treated as an asynchronous request and is kept in the background queue.
- User file-system implementation reads from the pending queue (handling forgetInode and interrupts as priorities).
- max_background determines how many async read-requests are promoted to the pending queue.
- congestion_threshold (generally kept at 75% of max_background) is the limit for the number of async read-requests in the pending queue; once this threshold is hit, Fuse informs VFS that it is congested, and VFS then throttles the user application.

b/469190320

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -268,7 +268,6 @@ func setMaxReadAhead(mountPoint string, readAheadKb int) {
 		return
 	}
 	logger.Infof("Max read-ahead set to %d KB on mount-path %s successfully.", readAheadKb, mountPoint)
-	return
 }
 
 // setCongestionThreshold sets the fuse settings `congestion-threshold` by writing over
@@ -293,7 +292,6 @@ func setCongestionThreshold(mountPoint string, congestionThreshold int) {
 		return
 	}
 	logger.Infof("Congestion threshold set to %d on mount-path %s successfully.", congestionThreshold, mountPoint)
-	return
 }
 
 // setMaxBackground sets the fuse settings `max-background` by writing over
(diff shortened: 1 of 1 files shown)


cmd/legacy_main.go
```diff
@@ -268,6 +268,52 @@ func setMaxReadAhead(mountPoint string, readAheadKb int) error {
...
+
+	sysPathMxBg := filepath.Join("/sys/fs/fuse/connections", fmt.Sprintf("%d", minor), "max_background")
+	cmd := exec.Command("sudo", "-n", "tee", sysPathMxBg)
+	cmd.Stdin = strings.NewReader(fmt.Sprintf("%d\n", maxBackground))
+
+	var stderr strings.Builder
+	cmd.Stderr = &stderr
+	err = cmd.Run()
+	if err != nil {
+		err = fmt.Errorf("setting the max-background %d on mount-path %s: %v, stderr: %s", maxBackground, mountPoint, err, stderr.String())
```
vadlakondaswetha: why dont we just log it here. what is the need for constructing an error message, returning and calling doing the logging

raj-prince: Following the best practices of method and returned error - so that it can be easily tested.

But given no unit test and used only at one place so logging inside the method only - warning for error and info for success.

PR #4164, opened 2025-12-16
feat(simple_reader): bounded LRU cache for unused-connected MRD

raj-prince:
### Description
- Changes existing behavior of closing MRD as soon as it becomes unused (refcount == 0), now such MRD instances will be open forever (or closed by server).
- Added a hidden flag to control the size of LRU cache `unused-mrd-cache-size`. Currently disabled by default.
- Used existing LRU cache to store the unused-connected MRD, with key as MRDWrapper.
- Integrated LRU in MRDWrapper's IncrementRef & DecrementRef method. Ensured no-deadlock and race-condition.

b/470741582

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA


internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -105,37 +114,96 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
...
+	}
+	logger.Tracef("MRDWrapper (%s) added wrapper to cache", mrdWrapper.object.Name)
+	mrdWrapper.mu.Unlock()
+
+	// Evict outside all locks to avoid deadlock
+	for _, wrapper := range evictedValues {
+		mrdWrapper, ok := wrapper.(*MultiRangeDownloaderWrapper)
+		if !ok {
+			return fmt.Errorf("invalid value type during cache eviction")
+		}
```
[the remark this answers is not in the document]
raj-prince: Any error will be a fatal error, and should be handled well by the caller, hence returning from the first occurrence.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -90,6 +94,13 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) SetMinObject(minObj *gcs.MinObjec
 	return nil
 }
 
+// wrapperKey generates a unique key for the given MultiRangeDownloaderWrapper.
+// Uses the pointer address as the unique identifier, would be safe as long as
+// wrapper is uniquely associated with the lifecycle of the FileInode instance.
+func wrapperKey(wrapper *MultiRangeDownloaderWrapper) string {
+	return fmt.Sprintf("%p", wrapper)
```
vadlakondaswetha, line 101: Are we sure there is no overhead because of converting an instance to string type. why can't we use the inodeId?

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -105,37 +116,98 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
...
+	wasZero := mrdWrapper.refCount == 0
 	mrdWrapper.refCount++
 	if mrdWrapper.cancelCleanup != nil {
 		mrdWrapper.cancelCleanup()
 		mrdWrapper.cancelCleanup = nil
 	}
+
+	// If refCount was 0, remove from cache (file is being reopened)
+	if wasZero && mrdWrapper.mrdCache != nil {
+		deletedEntry := mrdWrapper.mrdCache.Erase(wrapperKey(mrdWrapper))
```
vadlakondaswetha, line 132: why do we need this extra condition ? why can't the cache handle the not exists scenario?

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -105,37 +116,98 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
...
-		mrdWrapper.Wrapped.Close()
-		mrdWrapper.Wrapped = nil
+	// Do nothing if MRD is in use.
+	if mrdWrapper.refCount > 0 {
+		mrdWrapper.mu.Unlock()
+		return
+	}
+
+	// No cache: close MRD immediately
+	if mrdWrapper.mrdCache == nil {
```
vadlakondaswetha: can we remove this check. We dont want to clean up the MRD when all fileHandles are closed. This is a long pending task . @abhishek10004 - FYI

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -105,37 +116,98 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
...
+		return
+	}
+	logger.Tracef("MRDWrapper (%s) added wrapper to cache", mrdWrapper.object.Name)
+
+	// Evict outside all locks to avoid deadlock.
+	mrdWrapper.mu.Unlock()
+	for _, wrapper := range evictedValues {
+		mrdWrapper, ok := wrapper.(*MultiRangeDownloaderWrapper)
+		if !ok {
+			return fmt.Errorf("invalid value type during cache eviction")
```
vadlakondaswetha: what do we want to do here? return error to the application or just proceed?

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -105,37 +116,98 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
...
+		return
+	}
+
+	// Check if wrapper was re-added to cache (refCount went 0→1→0 in between eviction and closure.)
+	// Lock order: wrapper.mu -> cache.mu (consistent with Increment/DecrementRefCount)
+	if mrdWrapper.mrdCache != nil && mrdWrapper.mrdCache.LookUpWithoutChangingOrder(wrapperKey(mrdWrapper)) != nil {
+		return
+	}
+
+	if mrdWrapper.closeLocked() {
```
vadlakondaswetha: Are you returning a bool just for logging purpose? you can just add a log in closeLocked to say that we are closing it.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -105,37 +116,98 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) GetRefCount() int {
...
+	evictedValues, err := mrdWrapper.mrdCache.Insert(wrapperKey(mrdWrapper), mrdWrapper)
+	if err != nil {
+		logger.Errorf("failed to insert wrapper (%s) into cache: %v", mrdWrapper.object.Name, err)
+		mrdWrapper.mu.Unlock()
+		return
+	}
+	logger.Tracef("MRDWrapper (%s) added wrapper to cache", mrdWrapper.object.Name)
+
+	// Evict outside all locks to avoid deadlock.
+	mrdWrapper.mu.Unlock()
```
vadlakondaswetha, line 174: why do we need a unlock here. I am trying to see if we can add a defer unlock at the top instead of adding unlock everywhere.

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -402,3 +403,289 @@ func (t *mrdWrapperTest) Test_EnsureMultiRangeDownloader_ForceRecreateMRD() {
...
+	t.mrdWrapper.Wrapped = fake.NewFakeMultiRangeDownloaderWithSleep(t.object, t.objectData, time.Microsecond)
+	t.mrdWrapper.refCount = 0
+}
+
+func (t *mrdWrapperCacheTest) Test_Cache_AddAndRemove() {
+	key := wrapperKey(t.mrdWrapper)
+
+	// Act: Open, close, and reopen file
+	t.mrdWrapper.IncrementRefCount()
+	err := t.mrdWrapper.DecrementRefCount()
```
vadlakondaswetha, line 437: can we check if its added to cache after this step.

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -402,3 +403,289 @@ func (t *mrdWrapperTest) Test_EnsureMultiRangeDownloader_ForceRecreateMRD() {
...
+		}()
+	}
+	wg.Wait()
+
+	// Act: Decrement refcount in parallel
+	for range maxRefCount {
+		wg.Add(1)
+		go func() {
+			err := t.mrdWrapper.DecrementRefCount()
+			assert.Nil(t.T(), err)
```
vadlakondaswetha, line 490: what happens if decrement gets called before increment. we will get an error. This test might become flaky

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -402,3 +403,289 @@ func (t *mrdWrapperTest) Test_EnsureMultiRangeDownloader_ForceRecreateMRD() {
...
+	// Assert: MRD reused and removed from cache on reopen
+	assert.NoError(t.T(), err)
+	assert.Equal(t.T(), 1, t.mrdWrapper.refCount)
+	assert.Nil(t.T(), t.cache.LookUpWithoutChangingOrder(key), "Wrapper should be removed from cache")
+	assert.NotNil(t.T(), t.mrdWrapper.Wrapped, "MRD should still exist (reused)")
+}
+
+// Override parent test - with cache enabled, MRD stays pooled instead of being closed
+func (t *mrdWrapperCacheTest) Test_IncrementRefCount_CancelCleanup() {
+	// Arrange
```
vadlakondaswetha, line 451: what is the difference betwen this and above test?

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -402,3 +403,289 @@ func (t *mrdWrapperTest) Test_EnsureMultiRangeDownloader_ForceRecreateMRD() {
...
+		)
+		assert.NoError(t.T(), err)
+		wrapper.Wrapped = fake.NewFakeMultiRangeDownloaderWithSleep(obj, t.objectData, time.Microsecond)
+		wrappers[i] = wrapper
+		wrappers[i].IncrementRefCount()
+		err = wrappers[i].DecrementRefCount()
+		assert.NoError(t.T(), err)
+	}
+
+	// Act: Reopen wrapper 0 and add 4th wrapper (triggers eviction)
```
vadlakondaswetha: looks like comment is wrong here.

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -402,3 +403,289 @@ func (t *mrdWrapperTest) Test_EnsureMultiRangeDownloader_ForceRecreateMRD() {
...
+				t.mrdWrapper.IncrementRefCount()
+				err := t.mrdWrapper.DecrementRefCount()
+				assert.NoError(t.T(), err)
+			}
+		}()
+	}
+	wg.Wait()
+
+	// Assert: Final state is refCount=0 (no deadlocks or panics)
+	assert.Equal(t.T(), 0, t.mrdWrapper.refCount, "RefCount should be 0 after all operations")
```
vadlakondaswetha, line 591: what are we testing wrt to cache in this test? can we check if the entry is present in cache?

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: - Conversion was required to use existing LRU implementation which uses key as `string`. I haven't benchmarked but there will be some overhead and tradeoff (mentioned in the design as well) not using the custom implementation. Also, cache only comes into picture when refcount changes 0->1 and 1->0 and not in the critical read path, so less severe.

image

- Don't have `inodeId` information in MRDWrapper, where MRD is created and destroyed. We need to pass explicitly and also need to convert that as string to use existing LRU implementation.

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: You mean the condition `deletedEntry != nil`? That is just for logging purposes, otherwise log will be for every wrapper when ref changes from 0 to 1.

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Sure, I'll change this behavior.

Just wanted to keep existing behavior if cache is disabled. After removal of this check, the behavior will be similar to infinite cache (no MRD closure) when cache is disabled.

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Ideally, this should not be happening. A panic makes more sense, but don't want to crash the system. Returning error to application makes more sense to me, as this represents a corrupt state (wrong type) of MRDWrapper.

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Given we can removed the immediate closure logic when cache is disabled, we can move the log inside closeLocked() only.

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Don't want to take lock while closing other MRDWrappers, as that might lead to deadlock in the worst case scenario.

internal/gcsx/multi_range_downloader_wrapper_test.go
raj-prince: Decrement will happen only after increment, we have a `wg.Wait` after the parallel IncrementRefCount invocation.

internal/gcsx/multi_range_downloader_wrapper_test.go
raj-prince: Yeah, almost same. Planning to remove the test from the next PR.

Added a todo to removed this test with removal of cancelClean and clock thing.

internal/gcsx/multi_range_downloader_wrapper_test.go
raj-prince: corrected.

internal/gcsx/multi_range_downloader_wrapper_test.go
raj-prince: Nothing, just stability and ensure no deadlock or race condition. Added the check.

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Discussed offline to go with the string conversion, later after the newMRDReader changed, iNodeID will be used as key.

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Simplified the wasZero check to refCount == 1

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Discussed offline - decided to not return and just log as error.

cfg/config.go
```diff
@@ -482,6 +482,8 @@ type FileSystemConfig struct {
 	TempDir ResolvedPath `yaml:"temp-dir"`
 
 	Uid int64 `yaml:"uid"`
+
+	UnusedMrdCacheSize int64 `yaml:"unused-mrd-cache-size"`
```
vadlakondaswetha: nit: rename it to inactiveMRDCacheSize

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -130,12 +152,52 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) DecrementRefCount() (err error) {
...
+	for _, wrapper := range evictedValues {
+		mrdWrapper, ok := wrapper.(*MultiRangeDownloaderWrapper)
+		if !ok {
+			logger.Errorf("invalid value type, expected MultiRangeDownloaderWrapper, got %T", wrapper)
+		}
+		mrdWrapper.CloseMRDForEviction()
```
vadlakondaswetha: shouldn't the close  be in else block?

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -130,12 +152,52 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) DecrementRefCount() (err error) {
...
+
+	// Cache with refCount 0: add the wrapper to cache and evict overflow wrappers.
+	evictedValues, err := mrdWrapper.mrdCache.Insert(wrapperKey(mrdWrapper), mrdWrapper)
+	if err != nil {
+		logger.Errorf("failed to insert wrapper (%s) into cache: %v", mrdWrapper.object.Name, err)
+		return
+	}
+	logger.Tracef("MRDWrapper (%s) added wrapper to cache", mrdWrapper.object.Name)
+
+	// Evict outside all locks to avoid deadlock.
```
vadlakondaswetha, line 173: should we check if any entries are evicted before doing unlock/lock?

internal/gcsx/multi_range_downloader_wrapper.go
raj-prince: Aah, sorry. Fixed it.

PR #4166, opened 2025-12-16
feat(tailing reads):  E2E tests for updating inode with latest size

anushka567:
### Description
This PR includes E2E tests for https://github.com/GoogleCloudPlatform/gcsfuse/pull/4162

The new tests verify two key scenarios:

- The inode ID is preserved when an object is appended to remotely (size increases, but generation does not change) .
- A new inode ID is created when an object is overwritten remotely (generation changes)

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - Updated and automated.

### DIFF
--- a/tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
+++ b/tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
@@ -60,6 +60,23 @@ func (s *unfinalizedObjectOperations) SetupSuite() {
 }
 
 ////////////////////////////////////////////////////////////////////////
+// Helper methods
+////////////////////////////////////////////////////////////////////////
+
+func (t *unfinalizedObjectOperations) setupUnfinalizedObjectAndGetInitialInode(initialContentSize int) (filePath string, initialInodeID uint64) {
+	filePath = path.Join(t.testDirPath, t.fileName)
+	initialContent := setup.GenerateRandomString(initialContentSize)
+
+	// 1. Create an unfinalized object.
+	_ = client.CreateUnfinalizedObject(t.ctx, t.T(), t.storageClient, path.Join(testDirName, t.fileName), initialContent)
+
+	// 2. Stat the file to get initial Inode ID.
+	initialStat := operations.StatFileOrFatal(filePath, t.T())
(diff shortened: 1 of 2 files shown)


tools/integration_tests/util/operations/file_operations.go
```diff
@@ -855,3 +855,14 @@ func CreateFileOnDiskAndCopyToMntDir(t *testing.T, filePathInLocalDisk string, f
 		t.Errorf("Error in copying file:%v", err)
 	}
 }
+
+// StatFileOrFatal stats the given file path and returns the syscall.Stat_t struct.
+// It fails the test if os.Stat or the type assertion fails.
+func StatFileOrFatal(filePath string, t *testing.T) *syscall.Stat_t {
```
anushka567, line 861: Needed new method because wanted to access the inode number returned with stat which is not available   directly in existing method return type.

tools/integration_tests/unfinalized_object/unfinalized_object_operations_test.go
```diff
@@ -139,6 +156,49 @@ func (t *unfinalizedObjectOperations) TestUnfinalizedObjectCanBeRenamedIfCreated
...
+	require.NoError(t.T(), err)
+	require.Equal(t.T(), obj.Generation, finalObject.Generation)
+
+	// Stat the file again.
+	// Since we are using StatCacheTTL=0, this should trigger LookupInode.
+	newStat := operations.StatFileOrFatal(filePath, t.T())
+
+	// Assert Inode ID is preserved and Size is updated.
+	assert.Equal(t.T(), initialInodeID, newStat.Ino, "Inode ID should be preserved")
+	assert.Equal(t.T(), int64(initialSize+appendSize), newStat.Size, "Size should be updated")
```
vadlakondaswetha, line 184: do we already have tests which validate if we are able to read the newer content?

anushka567: yes. Included in rapid_appends package.

anushka567: and overall tailing read support tests is in a follow-up PR #4206

PR #4173, opened 2025-12-17
perf: Refresh stat cache in case of cache hit

Tulsishah:
### Description
An issue has been identified in GCSFuse where under specific caching scenarios, the number of GCS calls is unnecessarily increased.

This occurs when the Type cache has an entry, correctly identifying an object as a file, but the corresponding Stat cache entry has expired. In this situation, an additional Stat call is made to GCS.

The root cause appears to be that the current lookup implementation only refreshes the Type cache and does not actively refresh the Stat cache, leading to this inconsistency

### Impact:

Running this FIO config - https://paste.googleplex.com/6387664402120704

### On v3.5.1

```
=======================================================
Operation            | Count    | Avg Latency (ms)
-------------------------------------------------------
StatObject           | 61       | 13.3066        // Double the stat calls(61 as compare to 31 after fix) as Stat cache data got expired.
GetFolder            | 32       | 24.1387        
ListObjects          | 1        | 35.5159        
ReadWithReadHandle   | 2869     | 46.7912        
=======================================================
```
### With the fix
```
=======================================================
Operation            | Count    | Avg Latency (ms)
-------------------------------------------------------
StatObject           | 31       | 15.1895        
GetFolder            | 31       | 28.3649        
ListObjects          | 1        | 40.8495        
ReadWithReadHandle   | 2770     | 48.6371        
=======================================================
```

The fix delivers a 0.4s improvement in latency. While throughput remains stable due to the read-heavy nature of the current workload, we anticipate more significant performance gains as we scale to larger workloads.

[b/469616037](https://b.corp.google.com/issues/469616037)

### Testing details
1. Manual - Tested
2. Unit tests - Updated
4. Perf tests - 89U6E8dZwAN9gZM

### DIFF
--- a/tools/integration_tests/rapid_appends/reads_after_appends_test.go
+++ b/tools/integration_tests/rapid_appends/reads_after_appends_test.go
@@ -137,10 +137,9 @@ func (t *DualMountReadsTestSuite) runAppendAndReadTest(verifyFunc readAndVerifyF
 			// Read only up to the cached file size (before append).
 			verifyFunc(t.T(), readPath, []byte(t.fileContent[:sizeBeforeAppend]))
 
-			// Wait for metadata cache to expire to fetch the latest size for the next read.
-			// Metadata update for appends in current iteration itself takes a minute, so the
-			// cached size will expire in ttl-60 secs from now, so wait accordingly.
-			time.Sleep(time.Duration(metadataCacheTTLSecs*time.Second - operations.WaitDurationAfterFlushZB))
+			// Wait for the metadata cache to expire to ensure the next read fetches the updated file size.
+			// Since the previous 'stat' call just refreshed the cache, we must wait for the duration of metadataCacheTTLSecs
+			time.Sleep(time.Duration(metadataCacheTTLSecs * time.Second))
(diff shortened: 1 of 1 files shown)


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -341,9 +341,11 @@ func (b *fastStatBucket) StatObject(
 			err = &gcs.NotFoundError{
 				Err: fmt.Errorf("negative cache entry for %v", req.Name),
 			}
+			b.addNegativeEntry(req.Name)
 
 			return
 		}
+		b.insertMinObject(entry)
```
vadlakondaswetha, line 348: Entry is the object fetched from cache only right? why are inserting into cache again?

vadlakondaswetha: Are you saying we are refreshing the entry everytime it is accessed. Thats not right approach for this cache since its a ttl based cache for the GCS data

Tulsishah:
After discussing with @vadlakondaswetha, we concluded that this is WAI on https://github.com/GoogleCloudPlatform/gcsfuse/pull/4173#discussion_r2629557566

PR #4174, opened 2025-12-18
feat(tracing): Add a global text map propagator for GCSFuse tracing

thrivikram-karur-g:
### Description
It is recommended to add a global TextMapPropagator for tracing to pass the trace context across service boundaries.

b/469952719

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

N/A

### DIFF
--- a/internal/monitor/traceexporter.go
+++ b/internal/monitor/traceexporter.go
@@ -23,9 +23,18 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+	"go.opentelemetry.io/otel/propagation"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
 )
+
+func initPropagators() {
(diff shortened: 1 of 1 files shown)


internal/monitor/traceexporter.go
```diff
@@ -36,6 +45,7 @@ func SetupTracing(ctx context.Context, c *cfg.Config, mountID string) common.Shu
 	}
 	if tp != nil {
 		otel.SetTracerProvider(tp)
+		initPropagators()
```
kislaykishore, line 48: Nit: the initPropagators method doesn't do a lot. Could be inlined.
```suggestion

// SetupTracing bootstraps the OpenTelemetry tracing pipeline.
func SetupTracing(ctx context.Context, c *cfg.Config, mountID string) common.ShutdownFn {
tp, shutdown, err := newTraceProvider(ctx, c, mountID)
@@ -36,6 +45,7 @@ func SetupTracing(ctx context.Context, c *cfg.Config, mountID string) common.Shu
}
if tp != nil {
otel.SetTracerProvider(tp)
otel.SetTextMapPropagator(
  propagation.NewCompositeTextMapPropagator(
		propagation.TraceContext{},
		propagation.Baggage{},
	)
)
```

PR #4177, opened 2025-12-18
test: Managed folder test fix in CD pipeline

Tulsishah:
### Description
Removing alpha from commands as now it is publicly available.

[b/469933367](https://b.corp.google.com/issues/469933367)

### Testing details
1. Manual - Tested on louhi
2. Unit tests - NA

### DIFF
--- a/tools/cd_scripts/e2e_test.sh
+++ b/tools/cd_scripts/e2e_test.sh
@@ -297,7 +297,6 @@ export PATH=${PATH}:/usr/local/go/bin
 git clone https://github.com/googlecloudplatform/gcsfuse |& tee -a ${LOG_FILE}
 cd gcsfuse
 bash ./perfmetrics/scripts/install_latest_gcloud.sh
-export PATH=/usr/local/google-cloud-sdk/bin:$PATH
 
 # Installation of crcmod is working through pip only on rhel and centos.
 # For debian and ubuntu, we are installing through sudo apt.


tools/cd_scripts/e2e_test.sh
```diff
@@ -297,6 +297,7 @@ export PATH=${PATH}:/usr/local/go/bin
 git clone https://github.com/googlecloudplatform/gcsfuse |& tee -a ${LOG_FILE}
 cd gcsfuse
 bash ./perfmetrics/scripts/install_latest_gcloud.sh
+export PATH=/usr/local/google-cloud-sdk/bin:$PATH
```
ashmeenkaur: Is line 185 not setting this already?

Tulsishah: In 299 line we are installing latest gcloud, that might require to export it again.

ashmeenkaur: Can you check once? I don't think it should be required because path variable won't change with installation.

Tulsishah: So @vipnydav ran the test without this and it is consistently failing. 

My opinion
While technically redundant in a standard shell, our script does two things that make the second export necessary:

1. When we run bash install_latest_gcloud.sh, it runs in a sub-shell. Any path changes made inside that script are lost the moment it finishes.
2. If that script replaces the google-cloud-sdk folder([which is happening](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/perfmetrics/scripts/install_latest_gcloud.sh)), the shell sometimes loses the hash/reference to the gcloud binary. Re-exporting the path "refreshes" the shell's lookup table.

Maybe above export(185 line) is unnecessary but I am not removing it incase it breaks in between things.

Tulsishah: As discussed offline, without this also it is working. @vipnydav not took removing alpha related changes, that might be cause of failure.

ashmeenkaur: I actually tested this flow, and it turns out we don't need to re-export the path after the installation. Since we are already using an absolute path for the initial SDK and the PATH variable itself persists across the entire bash -c block, the shell maintains the correct reference. I'd prefer to keep the script clean and avoid the redundant export since the current setup is stable.

If we do run into a lookup issue, the more standard fix would be to use hash -r, but for now, the script runs successfully.

ashmeenkaur (approved):

PR #4189, opened 2025-12-23
feat(simple_reader): bucket-type based optimizations for kernel-fuse-settings

raj-prince:
### Description
- Added hidden config `--enable-kernel-reader` to guard all the changes for simple_reader.
- Added bucket-type based config optimization support, keeping the precedence similar UserSet --> Profile --> (Machine Type / Bucket Type: assumes no common config with both machine and bucket type optimization) --> Default.
- Set `--enable-kernel-reader` true for zonal bucket.
- BucketType based ApplyOptimization is done once we know the bucket-type based on GetStorageLayout call. As getting BucketType before requires major changes without giving much benefit.

**Sample Generated logs for bucket-type optimization:**
```
{"timestamp":{"seconds":1767163639,"nanos":707946382},"severity":"INFO","message":"GCSFuse Config","mount-id":"princer-gcsfuse-test-zonal-us-west4a-0b81f5b8","Applied optimizations for bucket-type: ":"zonal","Full Config":{"file-system.enable-kernel-reader":{"final_value":true,"optimization_reason":"bucket-type \"zonal\""}}}
{"timestamp":{"seconds":1767163639,"nanos":708033431},"severity":"INFO","message":"Mounting file system \"princer-gcsfuse-test-zonal-us-west4a\"...","mount-id":"princer-gcsfuse-test-zonal-us-west4a-0b81f5b8"}
```

b/471209127

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


gargnitingoogle:
@raj-prince summarizing here what we discussed offline (and some I thought up later).
1. Add new/modified test cases in test_config.tpl for generating new test cases for bucket-type optimization
2. Add new unit tests for bucket-type optimization for function ApplyOptimizations in cfg/config_test.go, cfg/optimize_test.go
3. Add new e2e tests for bucket-type optimization in tools/integration_tests/flag_optimizations
4. No new handling is needed for isValueSet in root.go while passing it into ApplyOptimizations right now. It's working and tested through the existing unit tests.
5. Add new unit tests for bucket-type optimization flag in `TestParseParamsYAMLStr_*` tests in tools/config-gen/parser_test.go
6. Please document, ensure and test the precedence order of bucket-type-optimization wrt preexisting profile-optimization, and machine-type-optimization.

tools/config-gen/templates/config_test.tpl
```diff
@@ -32,6 +32,7 @@ func TestApplyOptimizations(t *testing.T) {
 			name            string
 			config          Config
 			isSet           *mockIsValueSet
+			input           OptimizationInput
```
gargnitingoogle: This should be an array, not a single fixed value. This way it can be tested as empty, 1 or even more than 1 if we want to test that.

```suggestion
			input           []OptimizationInput
```

cfg/params.yaml
```diff
@@ -345,6 +338,17 @@ params:
...
     hide-flag: true
 
+  - config-path: "file-system.enable-kernel-reader"
+    flag-name: "enable-kernel-reader"
+    type: "bool"
+    usage: "Enables kernel reader, disables prefetching gcsfuse side and relies on kernel read-ahead and page-cache."
+    default: false
+    hide-flag: true
+    optimizations:
+      bucket-based-optimization:
```
gargnitingoogle: This name is a bit misleading. It seems to indicate that it can vary arbitrarily for any bucket i.e it's a function of the bucket, but in reality it's a function of the bucket-type. It'll make more sense to rename it to `bucket-type-optimization`.


```suggestion
      bucket-type-optimization:
```

cfg/config_util.go
```diff
@@ -61,3 +61,16 @@ func IsMetricsEnabled(c *MetricsConfig) bool {
 func IsGKEEnvironment(mountPoint string) bool {
 	return strings.HasPrefix(mountPoint, "/dev/fd/")
 }
+
+// GetBucketType converts BucketType boolean flags to a BucketType enum.
+// The priority order is: Zonal > Hierarchical > Standard.
+// This is used to determine bucket-specific optimizations for kernel configs.
+func GetBucketType(hierarchical bool, zonal bool) BucketType {
```
gargnitingoogle:
```suggestion
func GetBucketType(hierarchical, zonal bool) BucketType {
```

cfg/types.go
```diff
@@ -114,3 +114,29 @@ func (p *ResolvedPath) UnmarshalText(text []byte) error {
...
+
+const (
+	// BucketTypeZonal represents a zonal bucket with single-zone storage.
+	BucketTypeZonal BucketType = "zonal"
+
+	// BucketTypeHierarchical represents a bucket with hierarchical namespace enabled.
+	BucketTypeHierarchical BucketType = "hierarchical"
+
+	// BucketTypeStandard represents a standard (regional or multi-regional) bucket.
+	BucketTypeStandard BucketType = "standard"
```
gargnitingoogle: Opposite of hierarchical and as we usually call this bucket-type

```suggestion
	BucketTypeFlat BucketType = "flat"
```

cfg/config_util_test.go
```diff
@@ -222,3 +222,47 @@ func TestIsGKEEnvironment(t *testing.T) {
...
+func TestGetBucketType(t *testing.T) {
+	tests := []struct {
+		name         string
+		hierarchical bool
+		zonal        bool
+		expected     BucketType
+	}{
+		{
+			name:         "Zonal bucket",
+			hierarchical: false,
```
gargnitingoogle: Actually this is never the case, GCS never returns zonal - true and hierarchical - false. So, if anything this should be an error case.

cfg/config_util_test.go
```diff
@@ -222,3 +222,47 @@ func TestIsGKEEnvironment(t *testing.T) {
...
+			zonal:        false,
+			expected:     BucketTypeStandard,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := GetBucketType(tt.hierarchical, tt.zonal)
+			if result != tt.expected {
+				t.Errorf("GetBucketType(%v, %v) = %s; want %s",
```
gargnitingoogle: Avoid typecasting type BucketType to string.

```suggestion
				t.Errorf("GetBucketType(%v, %v) = %v; want %v",
```

cmd/root.go
```diff
@@ -156,8 +159,9 @@ of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 				return fmt.Errorf("invalid config: %w", err)
 			}
 
-			isSet := &pflagAsIsValueSet{fs: cmd.PersistentFlags()}
-			optimizedFlags := mountInfo.config.ApplyOptimizations(isSet)
+			isUserSet := &pflagAsIsValueSet{fs: cmd.PersistentFlags()}
```
gargnitingoogle, line 162: I am a bit skeptical of whether isUserSet properly contains the values of the flags coming from config-file too. let's verify that once.

cfg/optimize.go
```diff
@@ -156,7 +158,7 @@ func convertToCamelCase(input string) string {
 }
 
 // setFlagValue uses reflection to set the value of a flag in ServerConfig.
-func setFlagValue(cfg *Config, flag string, override flagOverride, isSet isValueSet) error {
+func setFlagValue(cfg *Config, flag string, override flagOverride, isSet IsValueSet) error {
```
gargnitingoogle: I am not sure why this function is in optimize.go at all. This is used only in optimize_test.go now, so this function definition should also be moved to optimize_test.go itself. Please make that change.

cfg/optimize.go
```diff
@@ -229,9 +231,11 @@ func getOptimizedValue(
 	currentValue any,
 	profileName string,
 	machineType string,
+	input OptimizationInput,
```
gargnitingoogle: This argument should be optional to be in sync with the function ApplyOptimizations. I am not sure how it'll work otherwise. ALso, it's a new parameter, so it's safer design-wise to keep it as the last.




```suggestion
	input ...OptimizationInput,
```

cfg/optimize.go
```diff
@@ -229,9 +231,11 @@ func getOptimizedValue(
...
 	machineTypeToGroupMap map[string]string,
 ) OptimizationResult {
-	// Precedence: Profile -> Machine -> Default
+	// Precedence: Profile -> (Machine / Bucket) -> Default
```
gargnitingoogle: As discussed offline, let's not be ambiguous between machine-type and bucket-type .

```suggestion
	// Precedence: Profile > Machine-type > Bucket-Type > Default-Value
```

tools/config-gen/templates/config_test.tpl
raj-prince: Changed as pointer so now, given testing more than 1 is not a valid use-case.

cfg/optimize.go
raj-prince: Passed as pointer, and nil represents no input.

cfg/config_util_test.go
raj-prince: I feel, we shouldn't test GCS bucket creation logic, unnecessarily complicating the the code.

cfg/optimize.go
raj-prince: Seems unused function, I'll removed this in a separate PR. b/472177719

cmd/root.go
raj-prince: Tested manually, working for the current enable-kernel-reader flag.

cfg/config_util_test.go
gargnitingoogle: I agree, but let's not test a case which isn't a valid scenario, i.e. zonal but not hierarchical.

cmd/root.go
gargnitingoogle: Nice. thanks for confirming.

cfg/optimize.go
gargnitingoogle: thanks. Approved that PR.

tools/config-gen/parser.go
```diff
@@ -127,11 +127,11 @@ func validateParam(param Param) error {
 
 	// Validate bucket-based optimizations if present.
 	if param.Optimizations != nil {
-		for _, bbo := range param.Optimizations.BucketBasedOptimization {
-			validBucketTypes := []string{"zonal", "hierarchical", "standard"}
-			if !slices.Contains(validBucketTypes, bbo.BucketType) {
+		for _, bto := range param.Optimizations.BucketTypeOptimization {
+			validBucketTypes := []string{"zonal", "hierarchical", "flat"}
```
gargnitingoogle: This line should be outside the for-loop.

cfg/optimize.go
```diff
@@ -231,11 +231,12 @@ func getOptimizedValue(
 	currentValue any,
 	profileName string,
 	machineType string,
-	input OptimizationInput,
+	input *OptimizationInput,
 	machineTypeToGroupMap map[string]string,
 ) OptimizationResult {
-	// Precedence: Profile -> (Machine / Bucket) -> Default
-	// Assuming Machine and Bucket optimizations are applied on mutually exclusive flags.
+	// Precedence: Profile -> Machine -> Bucket -> Default
```
gargnitingoogle: Apologies, didn't fully think this through last time while commenting.

```go
// Precedence: Profile -> Machine-type -> Bucket-type -> Default
```

cfg/optimize.go
```diff
@@ -262,11 +263,11 @@ func getOptimizedValue(
 	}
 
 	// 3. Only if no profile is set, check for bucket-type optimization.
```
gargnitingoogle:
```go
// 3. If no profile is set, and no machine-type optimization applies, then check for bucket-type optimization.
```

cfg/optimize_test.go
gargnitingoogle, line 247: Add a similar positive test for bucket-type optimization here.

tools/config-gen/parser_test.go
```diff
@@ -203,7 +203,7 @@ params:
     default: "128"
     "usage": "Maximum read ahead in KB"
     optimizations:
-      bucket-based-optimization:
+      bucket-type-optimization:
         - bucket-type: zonal
```
gargnitingoogle, line 211: Could you add one or two optimizations with the other bucket-types e.g. "hierarchical", "flat" etc to exercise all the paths ?

tools/config-gen/parser_test.go
gargnitingoogle, line 393: Can you add a case here for non-supported bucket-types here to see that the parser fails?

tools/config-gen/templates/config_test.tpl
gargnitingoogle, line 129: We need to have a similar test to show that profile overrides bucket-type

gargnitingoogle: Similarly another test case to show that machine-type overrides bucket-type . While there may not be any case today where they collide, this is needed to make the generated code and test more robust in case this collision happens in future.

internal/fs/fs.go
```diff
@@ -244,13 +244,11 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		if serverCfg.IsUserSet != nil {
 			bucketType := syncerBucket.BucketType()
 			bucketTypeEnum := cfg.GetBucketType(bucketType.Hierarchical, bucketType.Zonal)
```
gargnitingoogle, line 262: In case of dynamic mount, bucketTypeEnum will come out to be the BucketTypeFlat which will call ApplyOptimizations with non-nil OptimizationInput which may not be the most efficient flow of ApplyOptimizations by unnecessarily checking bucket-type, and associated optimization. Should we just pass OptimizationInput as nil for it ? Another approach is to define a BucketTypeUnknown and use that for dynamic mounts.

raj-prince: Dynamic mount case will not hit here - You can check the parent if statement line: 233.

cfg/config_util_test.go
raj-prince: removed the test-case.

cfg/optimize_test.go
raj-prince: Isn't that duplicate of already covered in config_test.go?

gargnitingoogle: This is about adding a single static test as a back up to any bug in the generated tests. You can skip this or add it in a new PR.

internal/fs/fs.go
```diff
@@ -250,11 +250,14 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
...
-			if bucketTypeStr != "standard" {
-				logger.Infof("Applying bucket-type optimizations for %s bucket", bucketTypeStr)
-				_ = serverCfg.NewConfig.ApplyOptimizations(serverCfg.IsUserSet, bucketTypeStr)
-			}
+			bucketTypeEnum := cfg.GetBucketType(bucketType.Hierarchical, bucketType.Zonal)
+			logger.Tracef("Applying bucket-type optimizations for %s bucket", bucketTypeEnum)
+			optimizedFlags := serverCfg.NewConfig.ApplyOptimizations(serverCfg.IsUserSet, &cfg.OptimizationInput{
+				BucketType: bucketTypeEnum,
+			})
+			logger.Tracef("Bucket based optimized flags: %+v", optimizedFlags)
```
gargnitingoogle: Shouldn't this be an info log ? This is on par with the other mount flags logs, particularly this one: https://github.com/GoogleCloudPlatform/gcsfuse/blob/1266f3007d55e5393747e5b95b862a6867898e14/cmd/legacy_main.go#L468 . Ideally, this log should be in json element "GCSFuse config".

internal/fs/fs.go
```diff
@@ -250,11 +250,14 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
...
 		if serverCfg.IsUserSet != nil {
 			bucketType := syncerBucket.BucketType()
-			bucketTypeStr := cfg.GetBucketTypeString(bucketType.Hierarchical, bucketType.Zonal)
-			if bucketTypeStr != "standard" {
-				logger.Infof("Applying bucket-type optimizations for %s bucket", bucketTypeStr)
-				_ = serverCfg.NewConfig.ApplyOptimizations(serverCfg.IsUserSet, bucketTypeStr)
-			}
+			bucketTypeEnum := cfg.GetBucketType(bucketType.Hierarchical, bucketType.Zonal)
+			logger.Tracef("Applying bucket-type optimizations for %s bucket", bucketTypeEnum)
+			optimizedFlags := serverCfg.NewConfig.ApplyOptimizations(serverCfg.IsUserSet, &cfg.OptimizationInput{
```
gargnitingoogle, line 263: There is no call to Rationalize here (like we have at https://github.com/GoogleCloudPlatform/gcsfuse/blob/1266f3007d55e5393747e5b95b862a6867898e14/cmd/root.go#L165 after the ApplyOptimizations there). May be it's not needed now, but it might be needed later, we should capture this with a TODO, or a warning log.

internal/fs/fs.go
gargnitingoogle: Another important thing to remember here will be that any flag changes made here will not reflect in the "GCSFuse Config" "Full Config" logs in either gcsfuse logs or in gke logs, which is where we primarily look for in mount configuration. Should we insert these two keywords in this log to make it easier to find that bucket-type optimization had an impact on the final mount flags, conditional only of course i.e. `if len(optimizedFlags) > 0` etc (not making any changes to the previously existing logs in any way whatsoever)

internal/fs/fs.go
```diff
@@ -250,11 +250,14 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		// Apply bucket-type-specific optimizations now that we know the bucket type
 		if serverCfg.IsUserSet != nil {
 			bucketType := syncerBucket.BucketType()
-			bucketTypeStr := cfg.GetBucketTypeString(bucketType.Hierarchical, bucketType.Zonal)
-			if bucketTypeStr != "standard" {
-				logger.Infof("Applying bucket-type optimizations for %s bucket", bucketTypeStr)
-				_ = serverCfg.NewConfig.ApplyOptimizations(serverCfg.IsUserSet, bucketTypeStr)
-			}
+			bucketTypeEnum := cfg.GetBucketType(bucketType.Hierarchical, bucketType.Zonal)
+			logger.Tracef("Applying bucket-type optimizations for %s bucket", bucketTypeEnum)
```
gargnitingoogle: Can we add a comment here to explain what is going on here ? It's a bit complicated now with a 2nd call to ApplyOptimizations which updates the flags over and above the original set of optimizations done at https://github.com/GoogleCloudPlatform/gcsfuse/blob/1266f3007d55e5393747e5b95b862a6867898e14/cmd/root.go#L160 .

gargnitingoogle (changes requested):

internal/fs/fs.go
```diff
@@ -247,15 +249,27 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
...
 		if serverCfg.IsUserSet != nil {
 			bucketType := syncerBucket.BucketType()
 			bucketTypeEnum := cfg.GetBucketType(bucketType.Hierarchical, bucketType.Zonal)
-			logger.Tracef("Applying bucket-type optimizations for %s bucket", bucketTypeEnum)
 			optimizedFlags := serverCfg.NewConfig.ApplyOptimizations(serverCfg.IsUserSet, &cfg.OptimizationInput{
 				BucketType: bucketTypeEnum,
 			})
-			logger.Tracef("Bucket based optimized flags: %+v", optimizedFlags)
+			if len(optimizedFlags) > 0 {
+				logger.Info("GCSFuse Config", "Full Config", fmt.Sprintf("Optimizing flags for %s bucket: ", bucketTypeEnum), optimizedFlags)
```
gargnitingoogle: thanks @raj-prince. This looks good. Could you paste a sample generated log from this in the PR description ?

gargnitingoogle (approved):
LGTM

gargnitingoogle (changes requested):
The log in PR description `"!BADKEY"` indicates something wrong.

cfg/config_util.go
```diff
@@ -61,3 +61,16 @@ func IsMetricsEnabled(c *MetricsConfig) bool {
 func IsGKEEnvironment(mountPoint string) bool {
 	return strings.HasPrefix(mountPoint, "/dev/fd/")
 }
+
+// GetBucketType converts BucketType boolean flags to a BucketType enum.
+// The priority order is: Zonal > Hierarchical > Flat.
+// This is used to determine bucket-specific optimizations for kernel configs.
```
vadlakondaswetha, line 67: we always wanted to convert the buckeTType struct into enum like this. Can you add a todo and take it after this PR to convert bucketType struct in buckeT_handle.go to this enum

vadlakondaswetha (dismissed):

internal/fs/fs.go
```diff
@@ -241,6 +248,32 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
...
+		// IMPACT: Flags are used after this. Optimization/rationalization functions are called twice
+		// for non-dynamic mounts, but they are idempotent, so it's safe.
+		if serverCfg.IsUserSet != nil {
+			bucketType := syncerBucket.BucketType()
+			bucketTypeEnum := cfg.GetBucketType(bucketType.Hierarchical, bucketType.Zonal)
+			optimizedFlags := serverCfg.NewConfig.ApplyOptimizations(serverCfg.IsUserSet, &cfg.OptimizationInput{
+				BucketType: bucketTypeEnum,
+			})
+			if len(optimizedFlags) > 0 {
+				logger.Info("GCSFuse Config", "Full Config", fmt.Sprintf("Optimizing flags for %s bucket: ", bucketTypeEnum), optimizedFlags)
```
gargnitingoogle: Ideally, we should split this log into two separate logs, one for `Optimized Flags` and another for `Full Config` as https://github.com/GoogleCloudPlatform/gcsfuse/blob/6d80e720c4c347af1d968fe2a38998bd9179b2ae/cmd/legacy_main.go#L467. Here, both logs will be conditional based on `len(optimizedFlags) > 0`

**Note**: This is purely a debuggability and observability improvement, and is not a blocker for this PR. Please feel free to do this in next PR.

gargnitingoogle:
> The log in PR description `"!BADKEY"` indicates something wrong.

The fix according to gemini 3:

```go
logger.Info("GCSFuse Config", "Full Config", fmt.Sprintf("Optimizing flags for %s bucket: ", bucketTypeEnum), "OptimizedFlags", optimizedFlags)
```

It creates a log like this one: 

```sh
time="31/12/2025 06:33:47.011671" severity=INFO message="GCSFuse Config" mount-id=gargnitin-test-zb-usw4a-425ccaea "Full Config"="Optimizing flags for zonal bucket: " OptimizedFlags="map[file-system.enable-kernel-reader:{FinalValue:true OptimizationReason:bucket-type \"zonal\" Optimized:true} metadata-cache.stat-cache-max-size-mb:{FinalValue:-1 OptimizationReason:profile \"aiml-training\" Optimized:true} metadata-cache.ttl-secs:{FinalValue:-1 OptimizationReason:profile \"aiml-training\" Optimized:true}]"
```

internal/fs/fs.go
raj-prince: Will take as another PR.

gargnitingoogle: For future reference, this is how this log should be eventually:

```sh
{"timestamp":{"seconds":1767163017,"nanos":182709526},"severity":"INFO","message":"Applied bucket-type based optimizations for zonal bucket-type. OptimizedFlags="map[file-system.enable-kernel-reader:{FinalValue:true OptimizationReason:bucket-type \"zonal\" Optimized:true ","mount-id":"princer-gcsfuse-test-zonal-us-west4a-7cae37ee"}

time="31/12/2025 06:33:45.084241" severity=INFO message="GCSFuse Config" mount-id=gargnitin-test-zb-usw4a-425ccaea "Full Config"="&{AppName: CacheDir: CloudProfiler:{AllocatedHeap:true Cpu:true Enabled:false Goroutines:false Heap:true Label:gcsfuse-0.0.0 Mutex:false} Debug:{ExitOnInvariantViolation:false Fuse:false Gcs:false LogMutex:false} DisableAutoconfig:false DisableListAccessCheck:true DummyIo:{Enable:false PerMbLatency:0s ReaderLatency:0s} EnableAtomicRenameObject:true EnableGoogleLibAuth:true EnableHns:true EnableNewReader:true EnableUnsupportedPathSupport:true FileCache:{CacheFileForRangeRead:false DownloadChunkSizeMb:200 EnableCrc:false EnableODirect:false EnableParallelDownloads:false ExcludeRegex: ExperimentalEnableChunkCache:false ExperimentalParallelDownloadsDefaultOn:true IncludeRegex: MaxParallelDownloads:128 MaxSizeMb:-1 ParallelDownloadsPerFile:16 WriteBufferSize:4194304} FileSystem:{CongestionThreshold:0 DirMode:493 DisableParallelDirops:false EnableKernelReader:true ExperimentalEnableDentryCache:false ExperimentalEnableReaddirplus:false ExperimentalODirect:false FileMode:420 FuseOptions:[] Gid:-1 IgnoreInterrupts:true InactiveMrdCacheSize:0 KernelListCacheTtlSecs:0 MaxBackground:0 MaxReadAheadKb:0 PreconditionErrors:true RenameDirLimit:200000 TempDir: Uid:-1} Foreground:true GcsAuth:{AnonymousAccess:false KeyFile: ReuseTokenFromUrl:true TokenUrl:} GcsConnection:{BillingProject: ClientProtocol:http1 CustomEndpoint: EnableHttpDnsCache:true ExperimentalEnableJsonRead:false ExperimentalLocalSocketAddress: GrpcConnPoolSize:1 HttpClientTimeout:0s LimitBytesPerSec:-1 LimitOpsPerSec:-1 MaxConnsPerHost:0 MaxIdleConnsPerHost:100 SequentialReadSizeMb:200} GcsRetries:{ChunkTransferTimeoutSecs:10 MaxRetryAttempts:0 MaxRetrySleep:30s Multiplier:2 ReadStall:{Enable:true InitialReqTimeout:20s MaxReqTimeout:20m0s MinReqTimeout:1.5s ReqIncreaseRate:15 ReqTargetPercentile:0.99}} ImplicitDirs:true List:{EnableEmptyManagedFolders:false} Logging:{FilePath: Format:text LogRotate:{BackupFileCount:10 Compress:true MaxFileSizeMb:512} Severity:INFO} MachineType:a3-highgpu-8g MetadataCache:{DeprecatedStatCacheCapacity:20460 DeprecatedStatCacheTtl:1m0s DeprecatedTypeCacheTtl:1m0s EnableNonexistentTypeCache:false ExperimentalDirMetadataPrefetch:false ExperimentalMetadataPrefetchOnMount:disabled NegativeTtlSecs:0 StatCacheMaxSizeMb:17592186044415 TtlSecs:9223372036 TypeCacheMaxSizeMb:-1} Metrics:{BufferSize:256 CloudMetricsExportIntervalSecs:0 ExperimentalEnableGrpcMetrics:false PrometheusPort:0 StackdriverExportInterval:0s UseNewNames:false Workers:3} Monitoring:{ExperimentalTracingMode: ExperimentalTracingProjectId: ExperimentalTracingSamplingRatio:0} OnlyDir: Profile:aiml-training Read:{BlockSizeMb:16 EnableBufferedRead:false GlobalMaxBlocks:40 InactiveStreamTimeout:10s MaxBlocksPerHandle:20 MinBlocksPerHandle:4 RandomSeekThreshold:3 StartBlocksPerHandle:1} WorkloadInsight:{ForwardMergeThresholdMb:0 OutputFile: Visualize:false} Write:{BlockSizeMb:32 CreateEmptyFile:false EnableRapidAppends:true EnableStreamingWrites:true FinalizeFileForRapid:false GlobalMaxBlocks:1600 MaxBlocksPerFile:1}}"
```

Please ignore the fact that the 2nd log that I generated isn't json format. If it's complicated, it's fine, do it later.

PR #4190, opened 2025-12-23
feat(gRPC-metrics): enable gRPC metrics via Go-SDK inbuilt metrics-exporter in GCE

kislaykishore:
### Description
Currently, the grpc-metrics use the existing GCSFuse exporters when enable-grpc-metrics is true. However, in GCE environment, we'd like grpc-metrics to be exported via the Go SDK inbuilt exporter as opposed to using the GCSFuse exporter. This is required to ensure that these metrics get categorized correctly and remain free to the end-users.

b/471097206

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - NA

No

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -145,16 +145,15 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
 	clientOpts = append(clientOpts, option.WithGRPCConnectionPool(clientConfig.GrpcConnPoolSize))
 	clientOpts = append(clientOpts, option.WithUserAgent(clientConfig.UserAgent))
 
-	if clientConfig.EnableGrpcMetrics {
+	if clientConfig.EnableGrpcMetrics && clientConfig.IsGKE {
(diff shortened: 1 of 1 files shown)


cojenco (approved):

internal/storage/storage_handle.go
```diff
@@ -145,16 +145,15 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
 	clientOpts = append(clientOpts, option.WithGRPCConnectionPool(clientConfig.GrpcConnPoolSize))
 	clientOpts = append(clientOpts, option.WithUserAgent(clientConfig.UserAgent))
 
-	if clientConfig.EnableGrpcMetrics {
+	if clientConfig.EnableGrpcMetrics && clientConfig.IsGKE {
```
abhishek10004, line 148: For non-gke environment, when grpc metrics are enabled, does it default to a pre existing provider inside go sdk since we're not setting the meter provider anymore?

kislaykishore: yes.

PR #4191, opened 2025-12-23
fix(metrics-perm-denied-error): disable Cloud Monitoring exporter on permission-denied error

kislaykishore:
### Description
When the service account that runs GCSFuse with cloud-metrics-exporter enabled but lacks the monitoring/metricWriter permission, then currently, the exporter continuously attempts to export and fails causing an error log for every instance of failed export. This PR fixes it by disabling the exporter when a permission-denied error is detected.

b/471095417

### Testing details
1. Manual - yes
2. Unit tests - Added
3. Integration tests - NA

No

### DIFF
--- a/internal/monitor/otelexporters.go
+++ b/internal/monitor/otelexporters.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	cloudmetric "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric"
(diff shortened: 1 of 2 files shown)


internal/monitor/otelexporters.go
```diff
@@ -96,8 +99,45 @@ func setupCloudMonitoring(secs int64) []metric.Option {
...
+// necessary permissions.
+type permissionAwareExporter struct {
+	metric.Exporter
+	// disabled indicates whether the exporter has been permanently disabled.
+	disabled atomic.Bool
+}
+
+func (e *permissionAwareExporter) Export(ctx context.Context, rm *metricdata.ResourceMetrics) error {
+	// Check if disabled before attempting export to save resources and avoid noise.
+	if e.disabled.Load() {
```
vadlakondaswetha, line 122: do we really need a wrapper here or just incorporate this one condition into existing exporter itself? Because of all these wrappers we are seeing go routine stack size increasing which triggers stack.resize consuming cpu cycles

kislaykishore: There is no easy way to do disable export since the cloud-metrics library doesn't support it.
This is a separate code path that doesn't come in read/write flow - the metrics are updated asynchronously. So, this shouldn't lead to performance degradation.Plus, it's a single atomic boolean that we are adding. It shouldn't affect the stack size by much.

PR #4193, opened 2025-12-23
feat: Making MRD Reader compatible to integrate with REadmanager

vadlakondaswetha:
Renamed ReadAt() to Read method in GCSReader interface. This is to make mrd_reader.go compatible with both GCSReader and Reader interfaces.
Added ReadAt() method in mrd_reader.go. This helps in integrating mrdReader with read_manager.go directly.

### Testing details
1. Manual - done
2. Unit tests - done
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/reader.go
+++ b/internal/gcsx/reader.go
@@ -113,5 +113,5 @@ type GCSReader interface {
 	// ReadAt reads data into the `Buffer` field of the provided `GCSReaderRequest`,
 	// starting from the specified offset and ending at the specified end offset.
 	// It returns a `ReadResponse` indicating the number of bytes successfully read and any error encountered.
-	ReadAt(ctx context.Context, req *GCSReaderRequest) (ReadResponse, error)
+	Read(ctx context.Context, req *GCSReaderRequest) (ReadResponse, error)
 }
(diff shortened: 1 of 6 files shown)


internal/gcsx/reader.go
```diff
@@ -113,5 +113,5 @@ type GCSReader interface {
 	// ReadAt reads data into the `Buffer` field of the provided `GCSReaderRequest`,
```
abhishek10004, line 113: nit: Read ...

internal/gcsx/reader.go
```diff
@@ -113,5 +113,5 @@ type GCSReader interface {
...
 	// starting from the specified offset and ending at the specified end offset.
 	// It returns a `ReadResponse` indicating the number of bytes successfully read and any error encountered.
-	ReadAt(ctx context.Context, req *GCSReaderRequest) (ReadResponse, error)
+	Read(ctx context.Context, req *GCSReaderRequest) (ReadResponse, error)
```
abhishek10004, line 116: Should we use a different name here like ReadFromOffset or ReadAtOffset (or something else) since we'll be providing an explicit offset to read from? (io.Reader's Read is a stateful read where the offset to read from is being maintained & not passed explicitly)

PR #4196, opened 2025-12-23
fix(mounting): Add hidden flag to skip the mount access check using ListObject

meet2mky:
### Description
We are introducing a hidden flag `disable-list-access-check` which defaults to `True`. This flag skips the Initial ListObject call that GCSFuse makes during the mount operation to perform  mount access. Practically this is redundant because ListObject API only requires `storage.objects.list` permission to work. Before making this initial ListObject request we already make a GetStorageLayout API request which also requires exact same permission `storage.objects.list` to work. So this permission check again using ListObject is redundant here.

Note: This changes protected added behind the flag in case we need immediate reverts for any reason.

This pull request also improves the initial ListObjects request to `includeFoldersAsPrefixes` with `"/"` as delimiter. This ensures the ListObjects request return as soon as it finds an alive object or an empty folder.

Note: I Will be adding integration tests for all bucket types in a different PR. I have manually verified this using all types of buckets.

b/470840781

### Testing details
1. Manual - Tested with all types of bucket with a SA having storage.objects.list permission and all mount succeeds details added in bug.
2. Unit tests - Added
3. Integration tests - Automated presubmit tests [e2e zonal](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/6e66937f-b52d-491c-ad9d-f7278d93e274/log), [e2e regional](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/6e66937f-b52d-491c-ad9d-f7278d93e274/log).

None

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1521,18 +1521,18 @@ func TestArgsParsing_EnableAtomicRenameObjectFlag(t *testing.T) {
 
 func TestArgsParsing_DisableListAccessCheckFlag(t *testing.T) {
 	tests := []struct {
-		name                             string
-		args                             []string
+		name                           string
+		args                           []string
 		expectedDisableListAccessCheck bool
 	}{
 		{
-			name:                             "default",
-			args:                             []string{"gcsfuse", "abc", "pqr"},
+			name:                           "default",
+			args:                           []string{"gcsfuse", "abc", "pqr"},
 			expectedDisableListAccessCheck: true,
 		},
 		{
-			name:                             "normal",
-			args:                             []string{"gcsfuse", "--disable-list-access-check=false", "abc", "pqr"},
+			name:                           "normal",
+			args:                           []string{"gcsfuse", "--disable-list-access-check=false", "abc", "pqr"},
 			expectedDisableListAccessCheck: false,
 		},
 	}


internal/gcsx/bucket_manager.go
```diff
@@ -256,10 +259,11 @@ func (bm *bucketManager) SetUpBucket(
 	// Fetch bucket type from storage layout api and set bucket type.
 	b.BucketType()
 
-	// Check whether this bucket works, giving the user a warning early if there
-	// is some problem.
-	{
-		_, err = b.ListObjects(ctx, &gcs.ListObjectsRequest{MaxResults: 1})
+	// TODO(b/471129209): Cleanup this code after confirming the GetStorageLayout is sufficient for bucket access checks.
```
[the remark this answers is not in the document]
meet2mky: TODO bug is different for code cleanup.

cfg/params.yaml
```diff
@@ -159,6 +159,13 @@ params:
     default: false
     hide-flag: true
 
+  - config-path: "disable-list-access-check"
+    flag-name: "disable-list-access-check"
+    type: "bool"
+    usage: "Disables the list object based access check during mount operation"
```
raj-prince, line 165: nit:
```suggestion
    usage: "Disables the list object based access check during mount operation."
```

meet2mky: Thanks, Let me do it in the next PR where I will add integration tests.

PR #4200, opened 2025-12-24
feat(simple reader): Implementing MRD Pool

abhishek10004:
### Description
Implementing MRD Pool

b/470889482

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/mrd_pool.go
+++ b/internal/gcsx/mrd_pool.go
@@ -127,7 +127,8 @@ func (p *MRDPool) createRemainingMRDs(handle []byte) {
 // Please check returned MRD is non nil and valid (i.e. not in an error state) before using it.
 func (p *MRDPool) Next() *MRDEntry {
 	limit := p.currentSize.Load()
-	idx := p.current.Add(1) % limit
+	// Use post-increment style to get 0-based index for round-robin.
+	idx := (p.current.Add(1) - 1) % limit
 	return &p.entries[idx]
 }
 
--- a/internal/gcsx/mrd_pool_test.go
+++ b/internal/gcsx/mrd_pool_test.go
@@ -126,10 +126,10 @@ func (t *mrdPoolTest) TestNext() {
 	e3 := pool.Next()
 	e4 := pool.Next()
 
-	assert.Equal(t.T(), e1.mrd, fakeMRD1)
-	assert.Equal(t.T(), e2.mrd, fakeMRD2)
-	assert.Equal(t.T(), e3.mrd, fakeMRD3)
-	assert.Equal(t.T(), e4.mrd, fakeMRD1)
+	assert.Same(t.T(), e1.mrd, fakeMRD1)
+	assert.Same(t.T(), e2.mrd, fakeMRD2)
+	assert.Same(t.T(), e3.mrd, fakeMRD3)
+	assert.Same(t.T(), e4.mrd, fakeMRD1)
 }
 
 func (t *mrdPoolTest) TestDeterminePoolSize() {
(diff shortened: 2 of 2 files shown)


internal/gcsx/mrd_pool.go
```diff
@@ -0,0 +1,182 @@
...
+	cancelFunc context.CancelFunc
+	// creationWg is used to wait for the background creation of MRDs to finish.
+	creationWg sync.WaitGroup
+}
+
+// NewMRDPool initializes a new MRDPool.
+// It creates the first MRD synchronously to ensure immediate availability and starts a background goroutine to create the remaining MRDs.
+func NewMRDPool(config *MRDPoolConfig, handle []byte) (*MRDPool, error) {
+	// Limiting the pool size to 1 for files smaller than 500 MiB
+	if config.object.Size < 500*MiB {
```
raj-prince: Should we de-couple this logic in a separate method? Will be easier to test, if we need to any heuristic in near future.

abhishek10004: Yeah, makes sense. Will so some benchmarking to come up with a better heuristic as well.

internal/gcsx/mrd_pool.go
```diff
@@ -0,0 +1,198 @@
...
+// It uses a handle from an existing MRD or a fallback handle.
+func (p *MRDPool) RecreateMRD(entry *MRDEntry, fallbackHandle []byte) error {
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+
+	var handle []byte
+	if entry.mrd != nil {
+		handle = entry.mrd.GetHandle()
+	} else if fallbackHandle != nil {
+		handle = fallbackHandle
```
vadlakondaswetha, line 145: what is fallback handle, how will the caller pass it?

internal/gcsx/mrd_pool.go
```diff
@@ -0,0 +1,198 @@
...
+	} else {
+		for i := 0; i < int(p.currentSize.Load()); i++ {
+			if &p.entries[i] == entry {
```
vadlakondaswetha, line 148: in what scenarios entry.mrd is nil?

internal/gcsx/mrd_pool_test.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (t *mrdPoolTest) TestClose() {
+	t.poolConfig.PoolSize = 2
+	fakeMRD := fake.NewFakeMultiRangeDownloader(t.object, nil)
+	t.bucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(fakeMRD, nil).Times(2)
+
+	pool, err := NewMRDPool(t.poolConfig, nil)
+	assert.NoError(t.T(), err)
+	// pool.creationWg.Wait()
```
vadlakondaswetha: is this comment intended?

internal/gcsx/mrd_pool_test.go
```diff
@@ -0,0 +1,252 @@
...
+	assert.NotSame(t.T(), oldMRD, entry.mrd)
+
+	pool.Close()
+}
+
+func (t *mrdPoolTest) TestRecreateMRD_Error() {
+	t.poolConfig.PoolSize = 1
+	fakeMRD := fake.NewFakeMultiRangeDownloader(t.object, nil)
+	t.bucket.On("NewMultiRangeDownloader", mock.Anything, mock.Anything).Return(fakeMRD, nil).Once()
+
```
vadlakondaswetha: can you arrange all tests into 
arrange
-- new line
act 
-- new line
assert

internal/gcsx/mrd_pool_test.go
```diff
@@ -0,0 +1,252 @@
...
+	assert.NotSame(t.T(), e1, e2)
+	assert.NotSame(t.T(), e2, e3)
+	assert.NotSame(t.T(), e3, e1)
+	assert.Same(t.T(), e1, e4) // Wraps around
+
+	// The underlying mrd instances should also be distinct.
+	assert.NotNil(t.T(), e1.mrd)
+	assert.NotNil(t.T(), e2.mrd)
+	assert.NotNil(t.T(), e3.mrd)
+	assert.NotSame(t.T(), e1.mrd, e2.mrd)
```
vadlakondaswetha: why not assign fake instances to variables and compare with them

internal/gcsx/mrd_pool.go
abhishek10004: If the caller of RecreateMRD is caching an MRD handle, it can pass that handle as well while requesting the MRD to be recreated. That is being referred to as fallback handle here.

internal/gcsx/mrd_pool.go
abhishek10004: We create the first MRD in the pool synchronously and the rest are created asynchronously. So, if any of the MRD creation calls fails in this async path, then those entries will have nil MRD. This is why users of the MRDPool have to check MRD before use and call RecreateMRD in case it's nil or invalid.

internal/gcsx/mrd_pool.go
abhishek10004: Have decoupled the logic but will update the exact heuristic after we have the new MRD Reader path ready for easier benchmarking.

internal/gcsx/mrd_pool_test.go
abhishek10004: Not needed. Have removed it.

vadlakondaswetha (approved):

PR #4201, opened 2025-12-24
feat(tailing reads): returning latest size as part of getInodeAttributes call

anushka567:
### Description
Supporting tail -f operations without o-direct mode.

Tailing operations keeps a single long lived file handle, when the metadata cache expires , then kernel sends getInodeAttr() which in turn sends a lookup call to GCS to identify whether the object was clobbered or not. Now, with this result, we can access the latest size updates remotely and update the inode attributes so that further reads (tails) can be issued and served.

[b/470883351](b/470883351)

### Testing details
1. Manual - Yes, tested using custom scripts , tail -f operation is working with continuous remote appends 
2. Unit tests - Added.
3. Integration tests - Automated for backward compatibility, New test cases in follow-up PR.

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -495,6 +495,7 @@ func (f *FileInode) DeRegisterFileHandle(readOnly bool) {
 // from operating on stale object information.
 func (f *FileInode) UpdateSize(size uint64) {
 	f.src.Size = size
+	f.attrs.Size = size
 	f.updateMRDWrapper()
 }
 
@@ -577,7 +578,6 @@ func (f *FileInode) Attributes(
 			// the inode attributes to reflect latest size.
 			if o != nil {
 				f.UpdateSize(o.Size)
-				f.attrs.Size = f.src.Size
 				attrs = f.attrs
 				attrs.Nlink = 1
 				return


internal/fs/inode/file.go
```diff
@@ -565,12 +565,33 @@ func (f *FileInode) Attributes(
...
 			err = fmt.Errorf("clobbered: %w", err)
 			return
 		}
 		if clobbered {
+			// If clobbered check is true but the minObject returned is not nil, it means the clobber
+			// was due to update in object size remotely (appends case). In this scenario, we will update
+			// the inode attributes to reflect latest size.
+			if o != nil {
+				minObj := storageutil.ConvertObjToMinObject(o)
+				f.src = *minObj
```
vadlakondaswetha: what about the udpateSize method we added in this file. Are we not going to use it?

internal/fs/inode/file.go
```diff
@@ -565,12 +565,33 @@ func (f *FileInode) Attributes(
...
+				f.updateMRDWrapper()
+
+				f.attrs.Size = f.src.Size
+				f.attrs.Mtime = f.src.Updated
+
+				attrs.Size = f.src.Size
```
vadlakondaswetha: duplicate lines

internal/fs/inode/file.go
```diff
@@ -565,12 +565,33 @@ func (f *FileInode) Attributes(
...
+			// If clobbered check is true but the minObject returned is not nil, it means the clobber
+			// was due to update in object size remotely (appends case). In this scenario, we will update
+			// the inode attributes to reflect latest size.
+			if o != nil {
+				minObj := storageutil.ConvertObjToMinObject(o)
+				f.src = *minObj
+				f.updateMRDWrapper()
+
+				f.attrs.Size = f.src.Size
+				f.attrs.Mtime = f.src.Updated
```
vadlakondaswetha: updated field doesn't get updated for appends.

internal/fs/inode/file.go
```diff
@@ -565,12 +565,33 @@ func (f *FileInode) Attributes(
...
+
+				attrs.Size = f.src.Size
+				attrs.Mtime = f.src.Updated
+
+				attrs.Nlink = 1
```
vadlakondaswetha, line 582: what is the default value of Nlink, i see its not set in other positive flows

internal/fs/inode/file_mock_bucket_test.go
```diff
@@ -209,3 +240,89 @@ func (t *FileMockBucketTest) TestSync_GenerationMismatchClobber() {
...
+	// Mock StatObject to return the updated object.
+	statReq := &gcs.StatObjectRequest{
+		Name:                           fileName,
+		ForceFetchFromGcs:              false,
+		ReturnExtendedObjectAttributes: false,
+	}
+	t.bucket.On("StatObject", t.ctx, statReq).
+		Return(updatedMinObject, &gcs.ExtendedObjectAttributes{}, nil).Once()
+
+	attrs, err := f.Attributes(t.ctx, true)
```
vadlakondaswetha: can you verify the attributes before updating the minObject also.

abhishek10004 (dismissed):
LGTM

internal/fs/inode/file.go
anushka567: At line https://github.com/GoogleCloudPlatform/gcsfuse/blob/70a1af7a599cf5622bbbc8b1e3282f55391143c1/internal/fs/inode/file.go#L579


we set this to 1 when we conclude that file should not be considered clobbered. With size only change, we want to notify kernel to not mark this inode as stale. Nlink=0 triggers cleanup of resources by kernel.

internal/fs/inode/file.go
anushka567: I added this assuming Updated field would also change.  If that is not the case, then we can use updateSize method.

internal/fs/inode/file.go
anushka567: The first 2 lines update the inode's attributes which gcsfuse internally maintains and the second updates the return value as in the beginning of the method, we initialize the return value of attrs=f.attrs . 

so once, f.attrs is updated, we would want to reflect the same in the returned values. In hindsight, UpdateSize() method should also update internal state. Including the change here.

internal/fs/inode/file.go
```diff
@@ -565,12 +566,25 @@ func (f *FileInode) Attributes(
...
 			err = fmt.Errorf("clobbered: %w", err)
 			return
 		}
 		if clobbered {
+			// If clobbered check is true but the minObject returned is not nil, it means the clobber
+			// was due to update in object size remotely (appends case). In this scenario, we will update
+			// the inode attributes to reflect latest size.
+			if o != nil {
+				f.UpdateSize(o.Size)
+				attrs = f.attrs
```
anushka567, line 581: Update the attrs value which is returned from this method, to reflect the latest after size attrs update in internal state.

internal/fs/inode/file.go
anushka567: Using UpdateSize method since updated field will not be updated with appends.

internal/fs/inode/file.go
anushka567: UpdateSize() method will update both inode.source size attr and inode.attrs size attr.

internal/fs/inode/file_mock_bucket_test.go
anushka567: Updated unit test.

vadlakondaswetha (approved):

PR #4209, opened 2025-12-26
feat(simple_reader): remove timeout-clock logic from MRDWrapper

raj-prince:
### Description
- Removed clock and their usage from the MRDWrapper.
- Removed cancelCleanup function attributes and related logic from MRDWrapper.

b/471341773

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


abhishek10004 (approved):
LGTM

PR #4215, opened 2025-12-29
ci: group dependabot PRs together

ashmeenkaur:
### Description
Group dependabot PRs together so perf tests and e2e tests can be run together.

b/472231233

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,9 +11,8 @@ updates:
   - dependency-type: "direct"
   schedule:
     interval: weekly
-  open-pull-requests-limit: 10
   groups:
-    # Group all Go updates together
+    # Group all Go updates together in a single PR.
     go-dependencies:
       patterns:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -12,8 +12,17 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+  groups:
```
raj-prince, line 14: Optional: After grouping, do we need `open-pull-request-limit:10)? I think, default one (5, according to gemini) should be enough.

.github/dependabot.yml
```diff
@@ -12,8 +12,17 @@ updates:
...
+    # Group all Go updates together
```
raj-prince: nit: Group all Go updates together in a single PR?

PR #4216, opened 2025-12-29
feat(Type cache deprecation): Disable type cache checks behind the flag

Tulsishah:
### Description
Disable type cache checks behind the flag

[b/472219928](http://b/472219928)

### Testing details
1. Manual - NA
2. Unit tests - NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -292,14 +292,6 @@ func NewDirInode(
 		panic(fmt.Sprintf("Unexpected name: %s", name))
 	}
 
-	var cache metadata.TypeCache
-	if !isEnableTypeCacheDeprecation {
-		cache = metadata.NewTypeCache(typeCacheMaxSizeMB, typeCacheTTL)
-	} else {
-		// Disable Type cache.
-		cache = metadata.NewTypeCache(0, 0)
-	}
-
 	typed := &dirInode{
 		bucket:                          bucket,
 		mtimeClock:                      mtimeClock,
@@ -310,11 +302,15 @@ func NewDirInode(
 		enableNonexistentTypeCache:      enableNonexistentTypeCache,
 		name:                            name,
 		attrs:                           attrs,
-		cache:                           cache,
 		isHNSEnabled:                    isHNSEnabled,
 		isUnsupportedPathSupportEnabled: isUnsupportedPathSupportEnabled,
 		isEnableTypeCacheDeprecation:    isEnableTypeCacheDeprecation,
 		unlinked:                        false,
+	}
+	var cache metadata.TypeCache
+	if !isEnableTypeCacheDeprecation {
+		cache = metadata.NewTypeCache(typeCacheMaxSizeMB, typeCacheTTL)
+		typed.cache = cache
 	}
 
 	typed.lc.Init(id)
@@ -592,7 +588,6 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error) 
 		return
 	}
 	var cachedType metadata.Type
-	logger.Info("Deprecation: ", d.IsTypeCacheDeprecated())
 	if d.IsTypeCacheDeprecated() {
 		// TODO: Add deprecation logic.
 		cachedType = metadata.UnknownType
@@ -641,10 +636,12 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error) 
 		result = fileResult
 	}
 
-	if result != nil {
-		d.cache.Insert(d.cacheClock.Now(), name, result.Type())
-	} else if d.enableNonexistentTypeCache && cachedType == metadata.UnknownType {
-		d.cache.Insert(d.cacheClock.Now(), name, metadata.NonexistentType)
+	if !d.IsTypeCacheDeprecated() {
+		if result != nil {
+			d.cache.Insert(d.cacheClock.Now(), name, result.Type())
+		} else if d.enableNonexistentTypeCache && cachedType == metadata.UnknownType {
+			d.cache.Insert(d.cacheClock.Now(), name, metadata.NonexistentType)
+		}
 	}
 
 	return result, nil
@@ -729,12 +726,14 @@ func (d *dirInode) readObjects(
 	}
 
 	cores = make(map[Name]*Core)
-	defer func() {
-		now := d.cacheClock.Now()
-		for fullName, c := range cores {
-			d.cache.Insert(now, path.Base(fullName.LocalName()), c.Type())
-		}
-	}()
+	if !d.IsTypeCacheDeprecated() {
+		defer func() {
+			now := d.cacheClock.Now()
+			for fullName, c := range cores {
+				d.cache.Insert(now, path.Base(fullName.LocalName()), c.Type())
+			}
+		}()
+	}
 
 	for _, o := range listing.MinObjects {
 		if storageutil.IsUnsupportedPath(o.Name) {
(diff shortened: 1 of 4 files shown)


internal/fs/inode/dir.go
```diff
@@ -577,6 +580,10 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
 		dirResult, err = findExplicitFolder(ctx, d.Bucket(), NewDirName(d.Name(), name))
 		return
 	}
+	if d.isEnableTypeCacheDeprecation {
+		// TODO: Add deprecation logic.
+		return nil, nil
+	}
 
 	cachedType := d.cache.Get(d.cacheClock.Now(), name)
```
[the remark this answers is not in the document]
Tulsishah: Since flag is default false and hidden, we will not hit this condition.

internal/cache/metadata/type_cache_test.go
```diff
@@ -38,61 +39,77 @@ var (
...
 }
 
-func (t *TypeCacheTest) SetUp(ti *TestInfo) {
+func (t *TypeCacheTest) SetupTest() {
 	t.ttl = TTL
-	t.cache = createNewTypeCache(TypeCacheMaxSizeMB, t.ttl)
+	t.cache = createNewTypeCache(t.T(), TypeCacheMaxSizeMB, t.ttl)
+	if t.cache.IsDeprecated() {
+		t.T().Skip("Skipping TypeCacheTest because EnableTypeCacheDeprecation is true")
+	}
```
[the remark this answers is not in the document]
Tulsishah: Not required. With enabling flag default true, we will change this value to disable the test permanently.

internal/fs/inode/dir_test.go
```diff
@@ -255,7 +257,9 @@ func (t *DirTest) validateCore(cores map[Name]*Core, entryName string, isDir boo
 	AssertTrue(ok, "entry for "+entryName+" not found")
 	ExpectEq(expectedFullName, core.FullName.GcsObjectName())
 	ExpectEq(expectedType, core.Type())
-	ExpectEq(expectedType, t.getTypeFromCache(entryName))
+	if !t.tc.IsDeprecated() {
+		ExpectEq(expectedType, t.getTypeFromCache(entryName))
+	}
```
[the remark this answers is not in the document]
Tulsishah: Not required. With enabling flag default true, we will change this value to disable the test permanently.

internal/fs/inode/hns_dir_test.go
```diff
@@ -216,7 +218,9 @@ func (t *HNSDirTest) TestLookUpChildShouldCheckOnlyForExplicitHNSDirectory() {
 	assert.Nil(t.T(), err)
 	assert.Equal(t.T(), dirName, result.FullName.GcsObjectName())
 	assert.Equal(t.T(), dirName, result.Folder.Name)
-	assert.Equal(t.T(), metadata.ExplicitDirType, t.typeCache.Get(t.fixedTime.Now(), name))
+	if !t.typeCache.IsDeprecated() {
+		assert.Equal(t.T(), metadata.ExplicitDirType, t.typeCache.Get(t.fixedTime.Now(), name))
+	}
```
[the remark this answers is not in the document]
Tulsishah: Not required. With enabling flag default true, we will change this value to disable the test permanently.

internal/cache/metadata/type_cache.go
```diff
@@ -60,6 +60,8 @@ type TypeCache interface {
 	// If entry doesn't exist in the cache, then
 	// UnknownType is returned.
 	Get(now time.Time, name string) Type
+
+	IsDeprecated() bool
```
charith87: It might be cleaner to not call TypeCache interface calls itself (in the calling functions with the flag) rather than modifying internally TypeCache with the flag disablement changes unless you are using any of this implementation even in the new flow.,

Tulsishah: This is for backward compatibility. If any issues are found with the new implementation, we can disable it via a flag and revert to the current TypeCache + statCache implementation. In the future, once the new approach stabilizes, we can remove this component entirely.

Tulsishah: I am not using it in new flow but with new flow all the existing tests should work. That's why I am removing type cache checks for new implementation.

charith87: Agreeing with keeping backward compatibility with flag. But wouldnt the flag switch be better off done in dir.go or fs.go instead of inside type cache implmeentation. That way the whole typeCache structs and goroutines are not executed at all and you'll anyway need to know to do these eventually when you need to remove the TypeCache references in those files as well.

Tulsishah: > But wouldnt the flag switch be better off done in dir.go or fs.go instead of inside type cache implmeentation.

I thought about this, but it requires checking in a lot of places (like Insert and Erase function usage). If I implement it this way, it will just go through the type cache and check if it is deprecated or not.

Tulsishah: > That way the whole typeCache structs and goroutines are not executed at all and you'll anyway need to know to do these eventually when you need to remove the TypeCache references in those files as well.

For the main core logic I followed this only, like keeping flag in dirInode and call Type cache logic based on the flag value.

This deprecation function is specifically for test execution where we can not directly access flag from dirInode.

As you can see I only used this function in test assertions.

Tulsishah: As discussed offline removed checks from Type cache and moved it to dirInode.

internal/fs/inode/dir.go
```diff
@@ -281,12 +285,21 @@ func NewDirInode(
...
 	if !name.IsDir() {
 		panic(fmt.Sprintf("Unexpected name: %s", name))
 	}
 
+	var cache metadata.TypeCache
+	if !isEnableTypeCacheDeprecation {
+		cache = metadata.NewTypeCache(typeCacheMaxSizeMB, typeCacheTTL)
+	} else {
+		// Disable Type cache.
+		cache = metadata.NewTypeCache(0, 0)
```
Tulsishah: For reviewer: If we keep it nil then we have to make assertion at every function call, which increases unnecessary checks in code.

charith87: Why do we need an assertion for cache at every function call. If you are already checking for the flag isEnableTypeCacheDeprecation?

Tulsishah: If we have type cache set to nil, all the checks like [these](https://github.com/GoogleCloudPlatform/gcsfuse/blob/5466a3ca368a529b6afdc5ad4d60f1b86d5aedf8/internal/fs/inode/dir.go#L633) will fail as cache instance is nil and we are calling function through that instance.

Tulsishah: Initialized type cache with nil on deprecation flag enabled. This will make sure that we will not use Type cache at all with deprecation.

charith87: As discussed, eventually we'll need to change these when you remove the type cache references anyway when making this feature default ON. Doing it early will help in reducing the chances of regressions in the future where you might only need to remove the code blocks under the flag.

internal/fs/inode/dir.go
```diff
@@ -297,11 +302,16 @@ func NewDirInode(
...
 		isHNSEnabled:                    isHNSEnabled,
 		isUnsupportedPathSupportEnabled: isUnsupportedPathSupportEnabled,
+		isEnableTypeCacheDeprecation:    isEnableTypeCacheDeprecation,
 		unlinked:                        false,
 	}
+	var cache metadata.TypeCache
+	if !isEnableTypeCacheDeprecation {
+		cache = metadata.NewTypeCache(typeCacheMaxSizeMB, typeCacheTTL)
+		typed.cache = cache
+	}
```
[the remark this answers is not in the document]
Tulsishah: Discussed with reviewer, we won't to initialize it with nil to make sure that Type cache implementation is not being called at all.

PR #4225, opened 2025-12-31
fix(integration-tests): Support Non Debian/Ubuntu based Machines for executing improved_e2e integration tests

PranjalC100:
### Issue: Integration Scripts Incompatible with RHEL Environments

**1. Error in `improved_run_e2e_tests.sh`**
The script failed because it relies on `apt-get`, which is specific to Debian/Ubuntu systems.

**2. Error in `./perfmetrics/scripts/install_go.sh`**
The Go installer failed because it used `dpkg` to determine the system architecture, causing the installation to be skipped on non-Debian distributions.
### Solution
To support multiple Linux distributions (RHEL, Debian, etc.), the scripts were updated to be distro-agnostic:

* **Architecture Detection:** Switched to `system_arch=$(uname -m)` to detect the architecture without relying on package managers.
* **OS Detection:** Implemented a check using `/etc/os-release` to dynamically select the correct package manager (`apt-get`, `yum`, etc.):
    ```bash
    if [ -f /etc/os-release ]; then
        . /etc/os-release
        # Execute install commands based on $ID
    ```

Created a new centralized script perfmetrics/scripts/os_utils.sh to handle OS and architecture detection as well as package installation logic.

### Testing details
1. Manual - Ran the new Manual CD script in gcsfuse-tools to test on a rhel based machine. The run was successful, logs at http://shortn/_iakOVsppDv


tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -603,12 +603,62 @@ build_gcsfuse_once() {
...
+  
+  log_info "Detected OS ID: $os_id"
+
+  case "$os_id" in
+    ubuntu|debian)
+      sudo apt-get update
+      sudo apt-get install -y python3
+      # install python3-setuptools tools.
+      # Added curl, wget, tar for robust go installation
+      sudo apt-get install -y gcc python3-dev python3-setuptools curl wget tar
```
meet2mky: It's brittle if you satisfy the curl, wget tar requirement from here for robust go installation. It's better and isolated to do this inside the `install_go.sh` script. So whatever dependencies go_install.sh needs should be checked and installed in that script itself instead of managing from outside.

perfmetrics/scripts/install_go.sh
```diff
@@ -31,24 +31,57 @@ INSTALL_DIR="/usr/local" # Installation directory
...
+            ;;
+    esac
+
+    echo "Detected architecture: $system_arch mapped to Go arch: $architecture"
+
+    # Download with retries
+    local url="https://go.dev/dl/go${GO_VERSION}.linux-${architecture}.tar.gz"
+    echo "Downloading from: $url"
+    
+    # Check if wget exists, otherwise try curl
```
meet2mky: Check and install dependencies at the top of the script.

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -603,12 +603,62 @@ build_gcsfuse_once() {
...
+      # it requires to install crcmod.
+      sudo apt-get install -y python3-crcmod
+      ;;
+    rhel|centos|fedora|almalinux|rocky)
+      # RHEL/CentOS logic
+      sudo yum install -y python3
+      # Note: python3-devel is the name in RHEL-based systems, not python3-dev
+      # Added curl, wget, tar for robust go installation
+      sudo yum install -y gcc python3-devel python3-setuptools curl wget tar
+      # Try to install crcmod if available (e.g. via EPEL), otherwise continue.
```
meet2mky: If crcmod is a required dependency then install it else we can drop it?

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -603,12 +603,62 @@ build_gcsfuse_once() {
...
+      # Added curl, wget, tar for robust go installation
+      sudo pacman -S --noconfirm python gcc python-setuptools python-crcmod curl wget tar
+      ;;
+    *)
+      # Fallback or unknown
+      if command -v apt-get &> /dev/null; then
+         log_info "OS not explicitly matched, but apt-get found. Trying apt logic."
+         sudo apt-get update && sudo apt-get install -y python3 gcc python3-dev python3-setuptools python3-crcmod curl wget tar
+      elif command -v yum &> /dev/null; then
+         log_info "OS not explicitly matched, but yum found. Trying yum logic."
```
meet2mky: Just fail here with error with whatever OS type is found than trying which makes debugging easier to see why we didn't match OS Type.

tools/integration_tests/improved_run_e2e_tests.sh
PranjalC100: Sure, I added a new file `os_utils.sh` that handles this, and is called from within `install_go.sh`

tools/integration_tests/improved_run_e2e_tests.sh
PranjalC100: Upon checking, crcmod is not being used anywhere in our script and is only being installed. We only have it written in `/tools/cd_scripts/requirements.in` & `/tools/cd_scripts/requirements.txt`. But I dont think it is being used anymore. Removing it for now.

PranjalC100: Adding it back, for now. Will Add a TODO to remove it after we remove gsutil usage.

perfmetrics/scripts/os_utils.sh
meet2mky, line 1: add empty line at the end of file.

perfmetrics/scripts/install_go.sh
```diff
@@ -26,17 +26,48 @@ if [[ $# -ne 1 ]]; then
...
 INSTALL_DIR="/usr/local" # Installation directory
 
 # Function to download, extract, and install go
 install_go() {
-    local temp_dir architecture
+    local temp_dir architecture os_id system_arch
+    
+    if ! os_id=$(get_os_id); then
+      log_error "Failed to detect OS ID."
+      exit 1
```
meet2mky:
```suggestion
      return 1
```

Use return 1 inside everywhere in the install_go method.

perfmetrics/scripts/install_go.sh
```diff
@@ -48,7 +79,9 @@ echo "Installing Go version ${GO_VERSION} to ${INSTALL_DIR}"
 INSTALLATION_LOG=$(mktemp /tmp/go_install_log.XXXXXX)
 if ! install_go > "$INSTALLATION_LOG" 2>&1; then
     echo "Go version ${GO_VERSION} installation failed."
+    echo "--- Installation Log ---"
     cat "$INSTALLATION_LOG"
```
meet2mky, line 83: If you use exit in the install_go then this line 81 to 86 won't be reached because shell already exited.

perfmetrics/scripts/install_go.sh
PranjalC100: Okay, will make the changes.

meet2mky:
Looks good! have added e2e labels for running tests on kokoro.

meet2mky (approved):

Tulsishah (approved):
Relying on @meet2mky's approval.

PR #4227, opened 2025-12-31
refactor(getProjection): improving method readability and type safety

raj-prince:
### Description
- Improved type safety and readability of getProjection method

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -333,20 +333,17 @@ func (bh *bucketHandle) CopyObject(ctx context.Context, req *gcs.CopyObjectReque
 }
 
 func getProjectionValue(req gcs.Projection) storage.Projection {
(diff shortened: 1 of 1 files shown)


abhishek10004:
Looks good but I have one question. Why do we have gcs.Projection at all? Why don't we directly use storage.Projection in GCSFuse?

raj-prince:
> Looks good but I have one question. Why do we have gcs.Projection at all? Why don't we directly use storage.Projection in GCSFuse?

Initially GCSFuse was using jacobsa/gcloud library gcs.go was the interface, we haven't changed the interface as such, and keeping the gcs.go as a thin wrapper over storage.go. 
Yes, we can improve all these, but requires more change as fs implementation of GCSFuse uses gcs.go interface.

PR #4229, opened 2026-01-02
feat(simple_reader): rationzlize kernel-reader-flag based on file-cache

raj-prince:
### Description
Instead of adding file-cache checks at every place: (a) while setting fuse_system settings (b) while enabling async read, we should rationalize the newly added `enable-kernel-reader` flag. After that no need to add file-cache check anywhere.

**Thinking Ahead:** - Later if we want to set enable-kernel-reader as true with file-cache on regional bucket. We can modify the rationalize condition to when bucket-type is regional.  In that case, bucket-type information will be required in the rationalize which is true for other cases too.

b/472923739

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -152,6 +152,13 @@ func resolveLoggingConfig(config *Config) {
 	}
 }
 
+func disableKernelReaderFlagBasedOnFileCache(c *Config) {
+	if IsFileCacheEnabled(c) {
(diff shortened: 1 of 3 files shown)


raj-prince:
Closing as we don't want to disable the kernel-reader flag based on file-cache on, instead we should make file-cache as no-op.

PR #4234, opened 2026-01-02
feat(simple reader): Implementing read using new MRD Simple Reader

abhishek10004:
### Description
This pull request introduces a new 'MRD Simple Reader' for zonal buckets, enabled by a new enable-kernel-reader flag. This involves adding a new read path in the filesystem logic, and several new components like MrdInstance and MrdSimpleReader to manage multi-range downloads. The changes are extensive and include new logic for caching and lifecycle management of downloader instances, along with comprehensive tests.

b/472227021

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Will be part of a separate PR

### DIFF
--- a/tools/integration_tests/util/mounting/persistent_mounting/perisistent_mounting.go
+++ b/tools/integration_tests/util/mounting/persistent_mounting/perisistent_mounting.go
@@ -50,7 +50,7 @@ func mountGcsfuseWithPersistentMountingWithConfigFile(config *test_suite.TestCon
 		"-o",
 		"log_file=" + config.LogFile,
 		"-o",
-		"--enable-kernel-reader=false",
+		"enable-kernel-reader=false",
 	}
 
 	persistentMountingArgs := makePersistentMountingArgs(flags)


cmd/mount.go
```diff
@@ -182,6 +182,11 @@ func getFuseMountConfig(fsName string, newConfig *cfg.Config) *fuse.MountConfig
 		EnableReaddirplus: newConfig.FileSystem.ExperimentalEnableReaddirplus,
 	}
 
+	// Enable async reads if enable-kernel-reader flag is set to true.
```
raj-prince: nit: we can directly assign
```suggestion
	mountCfg.EnableAsyncReads = newConfig.FileSystem.EnableKernelReader
```

internal/fs/handle/file.go
```diff
@@ -95,6 +98,10 @@ func NewFileHandle(inode *inode.FileInode, fileCacheHandler *file.CacheHandler,
 		handleID:               handleID,
 	}
 
+	if c.FileSystem.EnableKernelReader && inode != nil && inode.Bucket().BucketType().Zonal {
```
raj-prince: Should we remove bucket-type check as handled as part of enable-kernel-reader flag - only enabled for zonal bucket right now?

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// EnsureMrdInstance ensures that the MRD pool is initialized. If the pool
+// already exists, this function is a no-op.
+func (mi *MrdInstance) EnsureMrdInstance() (err error) {
+	mi.poolMu.Lock()
+	defer mi.poolMu.Unlock()
+
+	// Return early if pool exists.
+	if mi.mrdPool != nil {
```
vadlakondaswetha, line 155: This check should also be before taking lock.

internal/gcsx/mrd_instance_test.go
```diff
@@ -0,0 +1,217 @@
...
+	t.mrdInstance.refCount = 1
+
+	t.mrdInstance.DecrementRefCount()
+
+	assert.Equal(t.T(), int64(0), t.mrdInstance.refCount)
+	key := strconv.FormatUint(uint64(t.inodeID), 10)
+	assert.NotNil(t.T(), t.cache.LookUpWithoutChangingOrder(key))
+}
+
+func (t *MrdInstanceTest) TestDecrementRefCount_Eviction() {
```
vadlakondaswetha, line 305: There are lot more tests which are added by Prince related to cache. what is the plan for that?

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+			// Safe to destroy. Hold refCountMu to prevent concurrent resurrection.
+			mrdInstance.Destroy()
+			mrdInstance.refCountMu.Unlock()
+		}
+	}
+}
+
+// Size returns the number of active MRDs.
+func (mi *MrdInstance) Size() uint64 {
```
vadlakondaswetha, line 293: who needs this information?

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,185 @@
...
+	defer msr.mu.Unlock()
+	if msr.mrdInstance != nil {
+		msr.mrdInstanceInUse.Store(false)
+		msr.mrdInstance.DecrementRefCount()
+		msr.mrdInstance = nil
+	}
+}
+
+// CheckInvariants is a no-op for this reader type. It's included to satisfy
+// the gcsx.Reader interface.
```
vadlakondaswetha: can you remove this code if not required

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,185 @@
...
+		err = msr.mrdInstance.EnsureMrdInstance()
+		if err != nil {
+			return nil, err
+		}
+		// After initialization, get the next available entry.
+		entry = msr.mrdInstance.GetMRDEntry()
+	} else {
+		// If an entry was retrieved, check if it's usable.
+		entry.mu.RLock()
+		needsRecreation := entry.mrd == nil || entry.mrd.Error() != nil
```
vadlakondaswetha: why do we need a lock to check these values?

internal/fs/handle/file.go
vadlakondaswetha: +1. we should remove it. Also inode!=nil is not required, inode can never be nil.

internal/fs/handle/file.go
```diff
@@ -246,6 +256,37 @@ func (fh *FileHandle) ReadWithReadManager(ctx context.Context, req *gcsx.ReadReq
...
 }
 
+// ReadWithMrdSimpleReader reads data at the given offset using the mrd simple reader.
+//
+// LOCKS_REQUIRED(fh.inode.mu)
+// UNLOCK_FUNCTION(fh.inode.mu)
+func (fh *FileHandle) ReadWithMrdSimpleReader(ctx context.Context, req *gcsx.ReadRequest) (gcsx.ReadResponse, error) {
+	// If content cache enabled, CacheEnsureContent forces the file handler to fall through to the inode
+	// and fh.inode.SourceGenerationIsAuthoritative() will return false.
+	if err := fh.inode.CacheEnsureContent(ctx); err != nil {
```
vadlakondaswetha: we are not supporting local filecache. We shouldn't add new code based on that.

internal/fs/handle/file.go
```diff
@@ -246,6 +256,37 @@ func (fh *FileHandle) ReadWithReadManager(ctx context.Context, req *gcsx.ReadReq
...
+	}
+
+	if !fh.inode.SourceGenerationIsAuthoritative() {
+		// Read from inode if source generation is not authoritative.
+		defer fh.inode.Unlock()
+		n, err := fh.inode.Read(ctx, req.Buffer, req.Offset)
+		return gcsx.ReadResponse{Size: n}, err
+	}
+
+	fh.lockHandleAndRelockInode(true)
```
vadlakondaswetha: why do we need to lock inode here. Also mrdSimpleReader is initialized in the constructor itself. why do we need a lock to check if its not nil.

internal/gcsx/mrd_instance.go
abhishek10004: This is needed for the lru cache.

internal/gcsx/mrd_instance.go
abhishek10004: No, it has to be inside an RLock at least since other functions write to it & hence there can be data race. I used a single RWLock otherwise, we'd have to have something like the following.

RLock
// check mrdpool & return early
RUnlock

Lock 
// check mrdpool & return early
// create mrd pool
Unlock

Have updated to this now.

internal/gcsx/mrd_simple_reader.go
abhishek10004: To prevent race condition & torn reads (since entry.mrd is an interface, so it is 2 words and we can read wrong data if done without locks or atomic operation)

internal/fs/handle/file.go
abhishek10004: Thanks. We do not need to lock inode here but we still need to do nil check for mrdSimpleReader inside fh's Rlock to ensure that there is no issue as we destroy the mrdSimpleReader in file handle's destroy function.

internal/fs/handle/file.go
abhishek10004: Done. Removed the code.

internal/gcsx/mrd_instance_test.go
abhishek10004: I can check them and add the required ones later in a different PR.

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+		return
+	}
+	return fmt.Errorf("MrdInstance::RecreateMRDEntry MRDPool is nil")
+}
+
+// RecreateMRD recreates the entire MRD pool. This is typically called by the
+// file inode when the backing GCS object's generation changes, invalidating
+// all existing downloader instances.
+func (mi *MrdInstance) RecreateMRD(object *gcs.MinObject) error {
```
raj-prince: I don't see passed object is being used  anywhere in this method.

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,180 @@
...
+		entry.mu.RUnlock()
+
+		if needsRecreation {
+			err = msr.mrdInstance.RecreateMRDEntry(entry)
+			if err != nil {
+				return nil, err
+			}
+			// After recreation, get the next available entry.
```
vadlakondaswetha: why do we need to do GetMRDEntry(), the entry gets updated as part of ReCreateMRDEntry right. cant we use that directly.

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+// instance is actively used (refCount > 0), it is removed from the inactive
+// MRD cache to prevent eviction.
+func (mi *MrdInstance) IncrementRefCount() {
+	mi.refCountMu.Lock()
+	defer mi.refCountMu.Unlock()
+	mi.refCount++
+
+	if mi.refCount == 1 && mi.mrdCache != nil {
+		// Remove from cache
+		deletedEntry := mi.mrdCache.Erase(strconv.FormatUint(uint64(mi.inodeId), 10))
```
raj-prince: nit: Can we create an inline function for key generation? A central place for key generation logic would be good and easy to change in future if required.

vadlakondaswetha (dismissed):

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+			logger.Tracef("MrdInstance::IncrementRefCount: MrdInstance (%s) erased from cache", mi.object.Name)
+		}
+	}
+}
+
+// DecrementRefCount decreases the reference count. When the count drops to zero, the
+// instance is considered inactive and is added to the LRU cache for potential
+// reuse. If the cache is full, this may trigger the eviction and closure of the
+// least recently used MRD instances.
+func (mi *MrdInstance) DecrementRefCount() {
```
raj-prince, line 252: Can we align with the implementation in MRDWrapper? I see, few things are missing.

1. lock simplification using defer.
2. Return simply if no evicted entry (minor optimization).

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+		} else {
+			// Check if the instance was resurrected.
+			mrdInstance.refCountMu.Lock()
+			if mrdInstance.refCount > 0 {
+				mrdInstance.refCountMu.Unlock()
+				continue
+			}
+			// Safe to destroy. Hold refCountMu to prevent concurrent resurrection.
+			mrdInstance.Destroy()
+			mrdInstance.refCountMu.Unlock()
```
raj-prince: Should we create a simple method to do this job, inside only we can check to not close if becomes active?

Doesn't look good, one MRD instance taking lock of other MRD instance.

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+	// inodeId is the ID of the file inode associated with this instance.
+	inodeId fuseops.InodeID
+	// object is the GCS object for which the downloaders are created.
+	object *gcs.MinObject
+	// bucket is the GCS bucket containing the object.
+	bucket gcs.Bucket
+	// refCount tracks the number of active users of this instance.
+	refCount int64
+	// refCountMu protects access to refCount.
+	refCountMu sync.Mutex
```
raj-prince, line 47: What are we achieving with separate lock for refCount? I see poolLock is being taken under refCount Lock, we need to make sure no deadlock. If we are not achieving much than keeping a single lock would be cleaner.

raj-prince (changes requested):
Reviewed the cache interaction and suggested few changes.

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+// RecreateMRDEntry recreates a specific, potentially failed, entry in the MRD pool.
+func (mi *MrdInstance) RecreateMRDEntry(entry *MRDEntry) (err error) {
+	mi.poolMu.RLock()
+	defer mi.poolMu.RUnlock()
+	if mi.mrdPool != nil {
+		err = mi.mrdPool.RecreateMRD(entry, nil)
+		if err != nil {
+			err = fmt.Errorf("MrdInstance::RecreateMRDEntry Error in recreating MRD: %w", err)
+		}
+		return
```
raj-prince:
```suggestion
		if err = mi.mrdPool.RecreateMRD(entry, nil); err != nil {
		     return fmt.Errorf("MrdInstance::RecreateMRDEntry Error in recreating MRD: %w", err)
		}
```

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+	mi.mrdPool, err = NewMRDPool(&MRDPoolConfig{PoolSize: int(mi.mrdConfig.PoolSize), object: mi.object, bucket: mi.bucket}, nil)
+	if err != nil {
+		err = fmt.Errorf("MrdInstance::EnsureMrdInstance Error in creating MRDPool: %w", err)
+	}
+	return
+}
+
+// RecreateMRDEntry recreates a specific, potentially failed, entry in the MRD pool.
+func (mi *MrdInstance) RecreateMRDEntry(entry *MRDEntry) (err error) {
+	mi.poolMu.RLock()
```
raj-prince: Why Rlock, we are kind of modifying the mrdPool?

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// Size returns the number of active MRDs.
+func (mi *MrdInstance) Size() uint64 {
+	mi.poolMu.RLock()
+	defer mi.poolMu.RUnlock()
+	if mi.mrdPool != nil {
+		return mi.mrdPool.Size()
+	}
+	return 1
```
raj-prince: Should we return 0, if pool is nil?

internal/gcsx/mrd_instance.go
abhishek10004: Thanks. It's not needed anymore since we will update the object details in MRDInstance directly when inode updates the minObject.

internal/gcsx/mrd_instance.go
abhishek10004: We need 2 locks because Size method requires us to take a lock and this gets called from the cache methods (insert, erase). If there was a single lock, then the logic with cache becomes complex. The 2 separate locks allows us to keep it clean.
Will add more UTs in the next PR for deadlocks.

internal/gcsx/mrd_instance.go
abhishek10004: Because this lock is only for preventing access to mrdPool variable. mrdPool internally has locks to prevent access to its internal state.

internal/gcsx/mrd_simple_reader.go
abhishek10004: Yeah, it's not needed. Have removed it.

internal/gcsx/mrd_instance.go
abhishek10004: Done. Please check.

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,234 @@
...
+	defer mi.poolMu.RUnlock()
+	if mi.mrdPool != nil {
+		return mi.mrdPool.Next()
+	}
+	return nil
+}
+
+// EnsureMrdInstance ensures that the MRD pool is initialized. If the pool
+// already exists, this function is a no-op.
+func (mi *MrdInstance) EnsureMrdInstance() (err error) {
```
raj-prince: nit: should we rename it to EnsureMrdPool?

abhishek10004: I get your point but users of MRDInstance should not worry about MRDPool. Let me think on this.

raj-prince: Ideally, it should be done automatically inside, MRDInstance during the Add call. Maybe something to improve in the next refactoring PR.

internal/fs/handle/file.go
```diff
@@ -95,6 +98,10 @@ func NewFileHandle(inode *inode.FileInode, fileCacheHandler *file.CacheHandler,
 		handleID:               handleID,
 	}
 
+	if c.FileSystem.EnableKernelReader {
```
raj-prince, line 101: Why are we creating the simple-reader during file-handle creation? why not on the first read? 

That way Increment/Decrement will happen during the simple_reader creation and destroy.

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,178 @@
...
+func NewMrdSimpleReader(mrdInstance *MrdInstance) *MrdSimpleReader {
+	return &MrdSimpleReader{
+		mrdInstance: mrdInstance,
+	}
+}
+
+// getValidEntry handles the logic of obtaining a usable entry from the MRDInstance,
+// including initialization and recreation of entries if they are in a bad state.
+func (msr *MrdSimpleReader) getValidEntry() (*MRDEntry, error) {
+	msr.mu.RLock()
```
raj-prince: Sorry, why this lock is required? As I see, we set mrdInstance to nil only in Destroy() method, which will not happen during active read.

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,178 @@
...
+		return ReadResponse{}, fmt.Errorf("MrdSimpleReader: failed to get a valid MRD entry")
+	}
+
+	// Prepare the buffer for the read operation.
+	buffer := bytes.NewBuffer(req.Buffer)
+	buffer.Reset()
+	done := make(chan readResult, 1)
+
+	// Local mutex for the callback closure to prevent race conditions on the 'done' channel.
+	var cbMu sync.Mutex
```
raj-prince: Sorry, didn't understand the usage of mutex? Do we need to close the channel, shouldn't it be garbage collected? 

If we explicitly wants to close then we can do once received inside the select statement?

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,178 @@
...
+		// After initialization, get the next available entry.
+		entry = msr.mrdInstance.GetMRDEntry()
+	} else {
+		// If an entry was retrieved, check if it's usable.
+		entry.mu.RLock()
+		needsRecreation := entry.mrd == nil || entry.mrd.Error() != nil
+		entry.mu.RUnlock()
+
+		if needsRecreation {
+			err = msr.mrdInstance.RecreateMRDEntry(entry)
```
raj-prince: As discussed offline, MRD re-creation logic should be abstracted out from the reader, either in MRDInstance of MRDPool.

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if msr.mrdInstanceInUse.CompareAndSwap(false, true) {
+		msr.mrdInstance.IncrementRefCount()
+	}
+
+	// Attempt to get an entry.
+	entry := msr.mrdInstance.GetMRDEntry()
+
+	// If no entry is available, the pool might not be initialized.
```
raj-prince: Reader shouldn't be worry about pool and all. MRD instance should have responsibility to return a valid one, if possible retry internally.

internal/fs/handle/file.go
abhishek10004: We're just creating the MRDReader during the file handle creation. The actual MRD creation happens during first read only. Ref count logic is already there using an atomic bool to ensure we increment ref count only once during first read.

internal/gcsx/mrd_simple_reader.go
abhishek10004: yes, will handle the improvement in next PR.

internal/gcsx/mrd_simple_reader.go
abhishek10004: yes, agreed. Reader is not aware of the pool at all. It tries to get an MRDEntry and in case it is nil, it means that MRDInstance is not initialized and it should EnsureMRDInstance. Will update the comment.

internal/fs/handle/file.go
abhishek10004: Also, we'll have to take a lock in ReadPath if we start creating the Reader in read path as it can be called in parallel.

internal/gcsx/mrd_simple_reader.go
abhishek10004: I was not sure how kernel would behave with asyncIO but it would even then send the release file handle call after all reads have completed. Hence, removed the lock.

internal/gcsx/mrd_simple_reader.go
abhishek10004: I added it to ensure that if read returns early, then we close the channel properly. 
But you're right. Since it's a buffered channel, the callback would not block and it will eventually be garbage collected. Removed it.

internal/gcsx/mrd_simple_reader.go
abhishek10004: Done in this PR itself.

internal/gcsx/mrd_instance.go
abhishek10004: Have removed this function as it's not needed after refactoring.

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,76 @@
...
+
+	n, err := msr.mrdInstance.Read(ctx, req.Buffer, req.Offset)
+	return ReadResponse{Size: n}, err
+}
+
+// Destroy cleans up the resources used by the reader, primarily by destroying
+// the associated MrdInstance. This should be called when the reader is no
+// longer needed.
+func (msr *MrdSimpleReader) Destroy() {
+	msr.mu.Lock()
```
raj-prince: Do we still need this lock?

internal/gcsx/mrd_simple_reader.go
```diff
@@ -0,0 +1,76 @@
...
+	"context"
+	"fmt"
+	"sync"
+	"sync/atomic"
+)
+
+// MrdSimpleReader is a reader that uses an MRD Instance to read data from a GCS object.
+// This reader is simpler than the GCSReader as it doesn't have complex logic
+// to switch between sequential and random read strategies.
+type MrdSimpleReader struct {
```
raj-prince, line 26: After the refactoring, I see MrdSimpleReader is not doing much work other than first read IncrementRef handling and few checks. Should we eliminate it all together, and move the functionality to handle/file.go?

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,300 @@
...
+		if res.err != nil && res.err != io.EOF {
+			return res.bytesRead, fmt.Errorf("Error in Add call: %w", res.err)
+		}
+		return res.bytesRead, res.err
+	}
+}
+
+// ensureMrdInstance ensures that the MRD pool is initialized. If the pool
+// already exists, this function is a no-op.
+func (mi *MrdInstance) ensureMrdInstance() (err error) {
```
raj-prince: nit: should we rename it to ensureMrdPool?

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// destroyEvictedCacheEntries is a helper function to destroy evicted MrdInstance objects.
+// It handles type assertion and ensures that only truly inactive instances are destroyed.
+// This function should not be called when refCountMu is held.
+func destroyEvictedCacheEntries(evictedValues []lru.ValueType) {
+	for _, instance := range evictedValues {
+		mrdInstance, ok := instance.(*MrdInstance)
+		if !ok {
+			logger.Errorf("destroyEvictedCacheEntries: Invalid value type, expected *MrdInstance, got %T", mrdInstance)
```
raj-prince:
```suggestion
			logger.Errorf("destroyEvictedCacheEntries: Invalid value type, expected *MrdInstance, got %T", instance)
```

internal/gcsx/mrd_instance.go
```diff
@@ -0,0 +1,300 @@
...
+	// object is the GCS object for which the downloaders are created.
+	object *gcs.MinObject
+	// bucket is the GCS bucket containing the object.
+	bucket gcs.Bucket
+	// refCount tracks the number of active users of this instance.
+	refCount int64
+	// refCountMu protects access to refCount.
+	refCountMu sync.Mutex
+	// poolMu protects access to mrdPool.
+	poolMu sync.RWMutex
```
raj-prince, line 49: I feel, we can get rid of this poolMu with atomic.Pointer[MRDPool] and we can use sync.Once for initialization.

In the RecreateMRD, will be a simple swap of atomic Pointer, WDUT?

internal/gcsx/mrd_simple_reader.go
abhishek10004: No, I think we should keep it and not pollute handle/file.go for a couple of reasons. 1. Readability & cleanliness of code 2. Additional logic, like retry for short reads etc. would go in here.

internal/gcsx/mrd_instance.go
abhishek10004: Discussed offline.
Would not be doing that since we close the oldPool in RecreateMRD as well and it will become hard to ensure no memory leak with atomic pointer.

raj-prince (approved):

PR #4237, opened 2026-01-05
feat(Type Cache Deprecation): Fast stat bucket cache implicit dir

Tulsishah:
### Description
In GCSFuse, an "implicit directory" is a folder that exists as a prefix to other objects but does not have a corresponding 0-byte object in GCS. Historically, these were tracked in a separate typeCache. This PR migrates that responsibility to the statCache by inferring directory existence from listing results.

[b/473410044](http://b/473410044)

### Testing details
1. Manual - NA
2. Unit tests - Added

### DIFF
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -132,7 +132,9 @@ func (b *fastStatBucket) insertListing(listing *gcs.Listing, dirName string) {
 			continue
 		}
 
-		// Ensure the prefix follows naming conventions (trailing slash).
+		// Ensure the prefix follows directory naming conventions (trailing slash).
+		// Although 'collapsedRuns' is expected to contain only directories, we perform
+		// this defensive check to prevent processing malformed prefixes.
 		if !strings.HasSuffix(p, "/") {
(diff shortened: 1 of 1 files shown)


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -91,6 +91,61 @@ func (b *fastStatBucket) insertMultiple(objs []*gcs.Object) {
...
+	// 3. Cache Sub-directories (Collapsed Runs)
+	// These represent folders discovered via prefixes in the ListObjects response.
+	for _, p := range listing.CollapsedRuns {
+		// Skip if this name was already cached as an explicit object.
+		if _, exists := minObjectNames[p]; exists {
+			continue
+		}
+
+		// Ensure the prefix follows naming conventions (trailing slash).
+		if !strings.HasSuffix(p, "/") {
```
vadlakondaswetha, line 138: ideally this shouldn't be the case right since all the objects comes under minObjects and only directories are present under collapsedRuns. if yes can you add that comment explicitly.

vadlakondaswetha (dismissed):

vadlakondaswetha (approved):

PR #4243, opened 2026-01-06
feat(Type Cache Deprecation): Add FetchOnlyFromCache flag and enable/disable caching with respect to flag value

Tulsishah:
### Description
- Request Struct Updates: The StatObjectRequest, ListObjectsRequest, and a newly introduced GetFolderRequest now include IsTypeCacheDeprecated and FetchOnlyFromCache flags.
- Cache Miss Handling: In the fastStatBucket implementation, if FetchOnlyFromCache is enabled and a cache miss occurs, the system returns a new CacheMissError instead of falling back to GCS.
- Interface Refactoring: The GetFolder method across various bucket implementations and interfaces (including Bucket, bucketHandle, and debugBucket) was refactored to accept a GetFolderRequest struct rather than a simple string for the folder name

[b/473671143](http://b/473671143)

### Testing details
1. Manual - NA
4. Unit tests - Added
6. Integration tests - Automated

### DIFF
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -418,7 +418,7 @@ func (b *fastStatBucket) StatObject(
 	}
 
 	// Cache Miss Handling
-	if req.IsTypeCacheDeprecated && req.FetchOnlyFromCache {
+	if req.FetchOnlyFromCache {
 		return nil, nil, &CacheMissError{
 			Err: fmt.Errorf("cache miss for %q", req.Name),
 		}
@@ -432,30 +432,6 @@ func (b *fastStatBucket) StatObject(
 func (b *fastStatBucket) ListObjects(
 	ctx context.Context,
 	req *gcs.ListObjectsRequest) (listing *gcs.Listing, err error) {
-	if req.IsTypeCacheDeprecated && req.FetchOnlyFromCache {
-		var hit bool
-		var entry any
-		hit, entry = b.lookUp(req.Prefix)
-
-		if hit {
-			if minObject, ok := entry.(*gcs.MinObject); ok {
-				// Negative entries result in NotFoundError.
-				if minObject == nil {
-					return nil, &gcs.NotFoundError{
-						Err: fmt.Errorf("negative cache entry for %q", req.Prefix),
-					}
-				}
-				if minObject.Generation == 0 { // Assumed to be a directory-like object from a collapsed run.
-					return &gcs.Listing{CollapsedRuns: []string{minObject.Name}}, nil
-				}
-				return &gcs.Listing{MinObjects: []*gcs.MinObject{minObject}}, nil
-			}
-		}
-		return nil, &CacheMissError{
-			Err: fmt.Errorf("cache miss for %q", req.Prefix),
-		}
-	}
-
 	// Fetch the listing.
 	listing, err = b.wrapped.ListObjects(ctx, req)
 	if err != nil {
@@ -580,7 +556,7 @@ func (b *fastStatBucket) GetFolder(ctx context.Context, req *gcs.GetFolderReques
 		return entry, nil
 	}
 
-	if req.IsTypeCacheDeprecated && req.FetchOnlyFromCache {
+	if req.FetchOnlyFromCache {
 		return nil, &CacheMissError{
 			Err: fmt.Errorf("cache miss for %q", req.Name),
 		}
--- a/internal/storage/gcs/request.go
+++ b/internal/storage/gcs/request.go
@@ -243,17 +243,12 @@ type StatObjectRequest struct {
 	// Controls whether StatObject response includes GCS ExtendedObjectAttributes.
 	ReturnExtendedObjectAttributes bool
 
-	// Flag to enable deprecation logic of Type cache.
-	IsTypeCacheDeprecated bool
-
 	// FetchOnlyFromCache determines if the request should be served exclusively from the stat cache.
 	//
 	// If true, the request performs a cache lookup. On a cache miss, it returns a CacheMissError
 	// and does not fall back to GCS.
 	//
 	// If false, the request falls back to GCS on a cache miss.
-	//
-	// Note: This flag is currently only respected when IsTypeCacheDeprecated is true.
 	FetchOnlyFromCache bool
 }
 
@@ -266,12 +261,7 @@ type GetFolderRequest struct {
 	// and does not fall back to GCS.
 	//
 	// If false, the request falls back to GCS on a cache miss.
-	//
-	// Note: This flag is currently only respected when IsTypeCacheDeprecated is true.
 	FetchOnlyFromCache bool
-
-	// Flag to enable deprecation logic of Type cache.
-	IsTypeCacheDeprecated bool
 }
 
 type Projection int64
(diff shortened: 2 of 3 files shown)


internal/storage/gcs/request.go
```diff
@@ -242,6 +242,20 @@ type StatObjectRequest struct {
 
 	// Controls whether StatObject response includes GCS ExtendedObjectAttributes.
 	ReturnExtendedObjectAttributes bool
+
+	// FetchFromCache is a boolean flag.
+	// If true, the request attempts to fetch the object metadata/listing only from the stat cache.
+	// If false (default), the request proceeds to the backend (GCS) if the cache miss occurs or if the cache is bypassed entirely.
+	FetchFromCache bool
```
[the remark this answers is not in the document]
Tulsishah: Thought to add in separate PR but anyways added in this only

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -30,6 +30,16 @@ import (
 	"github.com/jacobsa/timeutil"
 )
 
+// A *CacheMissError value is an error that indicates an object name or a
+// particular generation for that name were not found from cache.
+type CacheMissError struct {
+	Err error
+}
+
+func (nfe *CacheMissError) Error() string {
```
vadlakondaswetha: why nfe and not cme?

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -391,29 +401,62 @@ func (b *fastStatBucket) StatObject(
 		return
 	}
 
-	// Do we have an entry in the cache?
+	// Deprecated Skip: The deprecated logic skips cache if FetchFromCache is false.
+	if req.IsTypeCacheDeprecated && !req.FetchFromCache {
```
vadlakondaswetha: in the regular scenario we want to fetch from cache and then fallback to GCS. FetchFromCAche is to fetch only from cache.
Are you expecting callers to always set fetchFromCache=true, otherwise you are going to GCS always.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -391,29 +401,62 @@ func (b *fastStatBucket) StatObject(
...
 
+	// Standard fallback to GCS.
 	return b.StatObjectFromGcs(ctx, req)
 }
 
 // LOCKS_EXCLUDED(b.mu)
 func (b *fastStatBucket) ListObjects(
 	ctx context.Context,
 	req *gcs.ListObjectsRequest) (listing *gcs.Listing, err error) {
+	if req.IsTypeCacheDeprecated && req.FetchFromCache {
```
vadlakondaswetha: same for others too.

internal/storage/caching/fast_stat_bucket.go
vadlakondaswetha: can we also rename it to fetchonlyfromcache

Tulsishah: FetchFromCache determines if the request should be served exclusively from the stat cache.
If true, the request performs a cache lookup. On a cache miss, it returns a CacheMissError and does not fall back to GCS.

If false, the cache is bypassed entirely and the request is served directly from GCS.

This flag is currently only respected when IsTypeCacheDeprecated is true.

This flag works only with deprecated logic.

Reason:
- We already checked cache(with true) then with false value why should we check the cache again.
- Regular scenarios this flag won't make any sense. I kept old logic only.

Tulsishah: As discussed offline, this flow is being used in other places as well, with false value we will keep the old behaviour .

internal/storage/caching/fast_stat_bucket.go
Tulsishah: Discussed in above comment. Updated the flow.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -407,13 +417,45 @@ func (b *fastStatBucket) StatObject(
...
 }
 
 // LOCKS_EXCLUDED(b.mu)
 func (b *fastStatBucket) ListObjects(
 	ctx context.Context,
 	req *gcs.ListObjectsRequest) (listing *gcs.Listing, err error) {
+	if req.IsTypeCacheDeprecated && req.FetchOnlyFromCache {
+		var hit bool
+		var entry any
+		hit, entry = b.lookUp(req.Prefix)
```
vadlakondaswetha: the definition looks wrong here. Method name says listObjects and you are only fetching one object from cache.
Why can't we add a new method which says getImplicitDir() and in the bucket_handle it can throw unimplemented error.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -407,13 +417,45 @@ func (b *fastStatBucket) StatObject(
...
+		if hit {
+			if minObject, ok := entry.(*gcs.MinObject); ok {
+				// Negative entries result in NotFoundError.
+				if minObject == nil {
+					return nil, &gcs.NotFoundError{
+						Err: fmt.Errorf("negative cache entry for %q", req.Prefix),
+					}
+				}
+				if minObject.Generation == 0 { // Assumed to be a directory-like object from a collapsed run.
+					return &gcs.Listing{CollapsedRuns: []string{minObject.Name}}, nil
```
vadlakondaswetha: what is the difference between returning minObject vs collapsedRuns here>

internal/storage/gcs/request.go
```diff
@@ -242,6 +242,36 @@ type StatObjectRequest struct {
 
 	// Controls whether StatObject response includes GCS ExtendedObjectAttributes.
 	ReturnExtendedObjectAttributes bool
+
+	// Flag to enable deprecation logic of Type cache.
+	IsTypeCacheDeprecated bool
```
vadlakondaswetha: Thinking more on this. why do you need to check isTypeCacheDeprecated in the fast_stat_bucket. anyways fetchonlyFromCache is set by the caller only when IsTypeCacheDeprecated=true. no need to check that flag everywhere

internal/storage/caching/fast_stat_bucket.go
Tulsishah: So we are making ListObjects(dir/) call, dir/ can be explicit directory or implicit one. If it's generation is zero then we treat it as implicit directory and return CollapsedRuns otherwise we treat it as minObject.

internal/storage/caching/fast_stat_bucket.go
Tulsishah: With implicit dir flag set to true for directory(implicit or explicit) we always makes LIstObjects call for Lookup. So we can't create function like getImplicitDir() and call during lookup because it is being used by explicit dir also.

One idea in my mind is, as we only fetching from cache we can make stat call for only for cache lookup for directory  also.
So basically I am telling to move ListObject caching logic into StatObject and call statobject only for cache lookup for file and directory

We can discuss this further. I am removing this from this PR right now. Will raise separate PR.

internal/storage/gcs/request.go
Tulsishah: With old implementation it was required, removed it.

vadlakondaswetha (approved):

PR #4244, opened 2026-01-06
feat(tracing): Add read flow traces

thrivikram-karur-g:
### Description
Adding read flow traces for flows like file cache read, file cache write, file cache with parallel downloads.

Perf tests details we have run for the same:

| Branch | File Size | Read BW | Write BW | RandRead BW | RandWrite BW |
|---|---|---|---|---|---|
| Master | 0.25MiB | 623.15MiB/s | 1.16MiB/s | 85.46MiB/s | 1.3MiB/s |
| PR | 0.25MiB | 580.55MiB/s | 1.22MiB/s | 87.62MiB/s | 1.29MiB/s |
| Master | 48.828MiB | 4745.93MiB/s | 73.07MiB/s | 1547.57MiB/s | 79.59MiB/s |
| PR | 48.828MiB | 4711.62MiB/s | 76.42MiB/s | 1558.36MiB/s | 79.6MiB/s |
| Master | 976.562MiB | 4735.57MiB/s | 32.94MiB/s | 1008.71MiB/s | 38.77MiB/s |
| PR | 976.562MiB | 4724.48MiB/s | 37.88MiB/s | 1115.39MiB/s | 38.52MiB/s |

b/454481823
b/473678121

### Testing details
1. Manual - 
Tested that traces are working as expected and shows up with the complete details when exported to the google cloud trace explorer.

file_cache_trace

2. Unit tests - 
    Added basic unittests for otel_tracer and modified tests to check for no failures due to traceHandle where applicable.
3. Integration tests - NA

NA

### DIFF
--- a/tracing/otel_tracer_test.go
+++ b/tracing/otel_tracer_test.go
@@ -28,16 +28,6 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-func TestNewOTELTracer(t *testing.T) {
-	tracer := NewOTELTracer()
-
-	assert.NotNil(t, tracer)
-	tracerImpl, ok := tracer.(*otelTracer)
-	assert.True(t, ok)
-	assert.NotNil(t, tracerImpl.tracer)
-	assert.NotNil(t, tracerImpl.slicePool)
-}
-
 func TestOtelTracer_StartEndSpan(t *testing.T) {
 	recorder := tracetest.NewSpanRecorder()
 	provider := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(recorder))
@@ -103,33 +93,35 @@ func TestOtelTracer_SetCacheReadAttributes(t *testing.T) {
 	spanName := "test-cache-read-span"
 	bytesRead := 123
 
-	t.Run("cache_hit", func(t *testing.T) {
-		recorder.Reset()
+	testCases := []struct {
+		name     string
+		cacheHit bool
+	}{
+		{
+			name:     "cache_hit",
+			cacheHit: true,
+		},
+		{
+			name:     "cache_miss",
+			cacheHit: false,
+		},
+	}
 
-		_, span := tracer.StartSpan(context.Background(), spanName)
-		tracer.SetCacheReadAttributes(span, true, bytesRead)
-		tracer.EndSpan(span)
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			recorder.Reset()
 
-		spans := recorder.Ended()
-		assert.Len(t, spans, 1)
-		assert.Len(t, spans[0].Attributes(), 2)
-		assert.Contains(t, spans[0].Attributes(), attribute.Int(BYTES_READ, bytesRead))
-		assert.Contains(t, spans[0].Attributes(), attribute.Bool(IS_CACHE_HIT, true))
-	})
+			_, span := tracer.StartSpan(context.Background(), spanName)
+			tracer.SetCacheReadAttributes(span, tc.cacheHit, bytesRead)
+			tracer.EndSpan(span)
 
-	t.Run("cache_miss", func(t *testing.T) {
-		recorder.Reset()
-
-		_, span := tracer.StartSpan(context.Background(), spanName)
-		tracer.SetCacheReadAttributes(span, false, bytesRead)
-		tracer.EndSpan(span)
-
-		spans := recorder.Ended()
-		assert.Len(t, spans, 1)
-		assert.Len(t, spans[0].Attributes(), 2)
-		assert.Contains(t, spans[0].Attributes(), attribute.Int(BYTES_READ, bytesRead))
-		assert.Contains(t, spans[0].Attributes(), attribute.Bool(IS_CACHE_HIT, false))
-	})
+			spans := recorder.Ended()
+			assert.Len(t, spans, 1)
+			assert.Len(t, spans[0].Attributes(), 2)
+			assert.Contains(t, spans[0].Attributes(), attribute.Int(BYTES_READ, bytesRead))
+			assert.Contains(t, spans[0].Attributes(), attribute.Bool(IS_CACHE_HIT, tc.cacheHit))
+		})
+	}
 }
 
 func TestOtelTracer_PropagateTraceContext(t *testing.T) {


internal/cache/file/downloader/job.go
```diff
@@ -486,8 +493,15 @@ func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool
 	} else if job.status.Name == NotStarted {
 		// Start the async download
 		job.status.Name = Downloading
-		job.cancelCtx, job.cancelFunc = context.WithCancel(context.Background())
-		go job.downloadObjectAsync()
+		span := trace.SpanFromContext(ctx)
+		newCtx := context.Background()
+		newCtx = trace.ContextWithSpan(newCtx, span)
+		newCtx, downloadSpan := job.traceHandle.StartTraceLink(newCtx, tracing.FileCacheRead)
```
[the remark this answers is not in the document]
thrivikram-karur-g: thanks, fixing

internal/gcsx/read_manager/read_manager.go
```diff
@@ -162,7 +171,9 @@ func (rr *ReadManager) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (gcsx.
 
 	var err error
 	for _, r := range rr.readers {
+		ctx, span := rr.traceHandle.StartTrace(ctx, reflect.TypeOf(r).String())
```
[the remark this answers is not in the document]
thrivikram-karur-g: this is resolved.

internal/bufferedread/buffered_reader.go
```diff
@@ -136,7 +140,9 @@ func NewBufferedReader(opts *BufferedReaderOptions) (*BufferedReader, error) {
 	// the file, capped by the configured minimum.
 	blocksInFile := (int64(opts.Object.Size) + opts.Config.PrefetchBlockSizeBytes - 1) / opts.Config.PrefetchBlockSizeBytes
 	numBlocksToReserve := min(blocksInFile, opts.Config.MinBlocksPerHandle)
+	_, span := opts.TraceHandle.StartTrace(context.Background(), tracing.ReadPrefetchBlockPoolGen)
```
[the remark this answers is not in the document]
thrivikram-karur-g: This is expected whenever buffered read starts new

internal/bufferedread/buffered_reader.go
```diff
@@ -140,7 +140,9 @@ func NewBufferedReader(opts *BufferedReaderOptions) (*BufferedReader, error) {
 	// the file, capped by the configured minimum.
 	blocksInFile := (int64(opts.Object.Size) + opts.Config.PrefetchBlockSizeBytes - 1) / opts.Config.PrefetchBlockSizeBytes
 	numBlocksToReserve := min(blocksInFile, opts.Config.MinBlocksPerHandle)
+	_, span := opts.TraceHandle.StartSpan(context.Background(), tracing.ReadPrefetchBlockPoolGen)
```
kislaykishore: @raj-prince Is this call to context.Background() performance sensitive? If so, what could we do to verify that it doesn't impact the performance?

internal/bufferedread/buffered_reader.go
```diff
@@ -337,7 +343,9 @@ func (p *BufferedReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (gcs
 		entry := p.blockQueue.Peek()
 		blk := entry.block
 
-		status, waitErr := blk.AwaitReady(ctx)
+		fetchCtx, span := p.traceHandle.StartSpan(ctx, tracing.WaitForPrefetchBlock)
+		status, waitErr := blk.AwaitReady(fetchCtx)
```
kislaykishore: How about we add the spans in the AwaitReady function?

internal/bufferedread/buffered_reader.go
```diff
@@ -359,7 +367,9 @@ func (p *BufferedReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (gcs
 
 		relOff := readOffset - blk.AbsStartOff()
 		bytesToRead := len(req.Buffer) - bytesRead
+		_, span = p.traceHandle.StartSpan(ctx, tracing.ReadFromPrefetchBlock)
```
kislaykishore: Same - let's add the tracing in the ReadAtSlice method.

tracing/otel_tracer.go
```diff
@@ -30,6 +31,18 @@ func (o *otelTracer) StartSpan(ctx context.Context, traceName string) (context.C
 	return o.tracer.Start(ctx, traceName)
 }
 
+func (o *otelTracer) StartSpanLink(ctx context.Context, traceName string) (context.Context, trace.Span) {
+	span := trace.SpanFromContext(ctx)
+	traceOpts := make([]trace.SpanStartOption, 0, 1)
+	traceOpts = append(traceOpts, trace.WithLinks(trace.Link{
+		SpanContext: span.SpanContext(),
+		Attributes: []attribute.KeyValue{
```
kislaykishore: Can we store this slice as a var so that it doesn't need to be recreated on every call?

tracing/otel_tracer.go
```diff
@@ -30,6 +31,18 @@ func (o *otelTracer) StartSpan(ctx context.Context, traceName string) (context.C
...
+			attribute.Int64("gcp.cloud_trace.link_type", 1),
```
kislaykishore: Please add documentation to this attribute.

tracing/otel_tracer.go
```diff
@@ -43,6 +56,13 @@ func (o *otelTracer) RecordError(span trace.Span, err error) {
 	span.SetStatus(codes.Error, err.Error())
 }
 
+func (o *otelTracer) SetCacheReadAttributes(span trace.Span, isCacheHit bool, bytesRead int) {
+	span.SetAttributes(
```
kislaykishore: What's the performance characteristic of this method call? Could you add the details as a code-comment?

tracing/trace_handle.go
```diff
@@ -27,6 +27,8 @@ type TraceHandle interface {
 	// Start a span with a given name & context
 	StartSpan(ctx context.Context, traceName string) (context.Context, trace.Span)
 
+	StartSpanLink(ctx context.Context, traceName string) (context.Context, trace.Span)
```
kislaykishore: Please add documentation to an exported method.

tracing/otel_tracer.go
```diff
@@ -30,6 +31,18 @@ func (o *otelTracer) StartSpan(ctx context.Context, traceName string) (context.C
...
 
+func (o *otelTracer) StartSpanLink(ctx context.Context, traceName string) (context.Context, trace.Span) {
+	span := trace.SpanFromContext(ctx)
+	traceOpts := make([]trace.SpanStartOption, 0, 1)
+	traceOpts = append(traceOpts, trace.WithLinks(trace.Link{
+		SpanContext: span.SpanContext(),
+		Attributes: []attribute.KeyValue{
+			attribute.Int64("gcp.cloud_trace.link_type", 1),
+		},
+	}))
```
kislaykishore: This can be done in a single statement.
```suggestion
	traceOpts := []trace.SpanStartOption {
	                           trace.WithLinks(trace.Link{
			                           SpanContext: span.SpanContext(),
			                           Attributes: []attribute.KeyValue{
				                           attribute.Int64("gcp.cloud_trace.link_type", 1),
			                           },
			                         ),
		                         }
```

internal/gcsx/random_reader.go
```diff
@@ -359,8 +359,12 @@ func (rr *randomReader) ReadAt(
 	// then the file cache behavior is write-through i.e. data is first read from
 	// GCS, cached in file and then served from that file. But the cacheHit is
 	// false in that case.
+	ctx, span := rr.traceHandle.StartSpan(ctx, tracing.FileCacheRead)
+	defer rr.traceHandle.EndSpan(span)
 	n, cacheHit, err := rr.tryReadingFromFileCache(ctx, p, offset)
+	rr.traceHandle.SetCacheReadAttributes(span, cacheHit, n)
 	if err != nil {
+		rr.traceHandle.RecordError(span, err)
```
kislaykishore: We can move this logic inside of `tryReadingFromFileCache` method

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -230,7 +232,7 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader(forceR
 					handle = mrdWrapper.handle
 				}
 			}
-			mrd, err = mrdWrapper.bucket.NewMultiRangeDownloader(context.Background(), &gcs.MultiRangeDownloaderRequest{
+			mrd, err = mrdWrapper.bucket.NewMultiRangeDownloader(ctx, &gcs.MultiRangeDownloaderRequest{
 				Name:           mrdWrapper.object.Name,
 				Generation:     mrdWrapper.object.Generation,
 				ReadCompressed: mrdWrapper.object.HasContentEncodingGzip(),
```
kislaykishore, line 238: This is a logic change. @raj-prince could you review?

internal/gcsx/file_cache_reader.go
```diff
@@ -216,8 +220,12 @@ func (fc *FileCacheReader) ReadAt(ctx context.Context, req *ReadRequest) (ReadRe
 	// then the file cache behavior is write-through i.e. data is first read from
 	// GCS, cached in file and then served from that file. But the cacheHit is
 	// false in that case.
+	ctx, span := fc.traceHandle.StartSpan(ctx, tracing.FileCacheRead)
```
kislaykishore: Same comment as before: We can move this logic inside of tryReadingFromFileCache method

internal/gcsx/read_manager/read_manager.go
```diff
@@ -169,7 +174,9 @@ func (rr *ReadManager) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (gcsx.
 
 	var err error
 	for _, r := range rr.readers {
+		ctx, span := rr.traceHandle.StartSpan(ctx, r.ReaderName())
```
kislaykishore, line 196: Let's move this logic within ReadAt method?

internal/cache/file/downloader/job.go
```diff
@@ -492,8 +493,15 @@ func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool
 	} else if job.status.Name == NotStarted {
 		// Start the async download
 		job.status.Name = Downloading
-		job.cancelCtx, job.cancelFunc = context.WithCancel(context.Background())
-		go job.downloadObjectAsync()
+		span := trace.SpanFromContext(ctx)
+		newCtx := context.Background()
+		newCtx = trace.ContextWithSpan(newCtx, span)
```
kislaykishore:
```suggestion
		newCtx = trace.ContextWithSpan(context.Background(), span)
```

internal/cache/file/downloader/job.go
```diff
@@ -492,8 +493,15 @@ func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool
 	} else if job.status.Name == NotStarted {
 		// Start the async download
 		job.status.Name = Downloading
-		job.cancelCtx, job.cancelFunc = context.WithCancel(context.Background())
-		go job.downloadObjectAsync()
+		span := trace.SpanFromContext(ctx)
```
kislaykishore: Could you use PropagateTraceContext here?

tracing/span_attributes.go
```diff
@@ -0,0 +1,20 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tracing
+
+const (
+	IS_CACHE_HIT = "gcsfuse.is_cache_hit"
```
kislaykishore: Usually, the convention is to use "cache.hit" as the key and true/false as values.
```suggestion
	IS_CACHE_HIT = "cache.hit"
```

tracing/span_attributes.go
```diff
@@ -0,0 +1,20 @@
...
+	BYTES_READ   = "gcsfuse.bytes_read"
```
kislaykishore: Convention is to use *.size
```suggestion
	BYTES_READ   = "read.size"
```

tracing/span_names.go
```diff
@@ -21,35 +21,43 @@ package tracing
...
+	SyncFile                 = "SyncFile"
+	FlushFile                = "FlushFile"
+	ReleaseFileHandle        = "ReleaseFileHandle"
+	ReadSymlink              = "ReadSymlink"
+	RemoveXattr              = "RemoveXattr"
+	GetXattr                 = "GetXattr"
+	ListXattr                = "ListXattr"
+	SetXattr                 = "SetXattr"
+	Fallocate                = "Fallocate"
+	SyncFS                   = "SyncFS"
```
kislaykishore: Let's add an "fs." prefix to represent the fs operations.

tracing/span_names.go
```diff
@@ -21,35 +21,43 @@ package tracing
 // operations. Using these constants ensures consistency and better readability and a single source of truth where all the span names are listed.
 
 const (
```
kislaykishore, line 23: OTel suggests that the span names should be in lowercase. Please familiarize yourself with the [naming convention](https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations)

tracing/span_names.go
```diff
@@ -21,35 +21,43 @@ package tracing
...
-	FlushFile          = "FlushFile"
-	ReleaseFileHandle  = "ReleaseFileHandle"
-	ReadSymlink        = "ReadSymlink"
-	RemoveXattr        = "RemoveXattr"
-	GetXattr           = "GetXattr"
-	ListXattr          = "ListXattr"
-	SetXattr           = "SetXattr"
-	Fallocate          = "Fallocate"
-	SyncFS             = "SyncFS"
+	FileCacheRead            = "FileCacheRead"
```
kislaykishore: Let's add a docstring indicating the meaning of each span.

internal/bufferedread/download_task.go
```diff
@@ -58,6 +62,14 @@ func (p *downloadTask) Execute() {
 	stime := time.Now()
 	var err error
 	var n int64
+	var span trace.Span
+	p.ctx, span = p.traceHandle.StartSpan(p.ctx, tracing.DownloadPrefetchBlock)
+	logger.Info("Start span has been successfully triggerd")
```
kislaykishore:
```suggestion
```

internal/bufferedread/download_task.go
```diff
@@ -58,6 +62,14 @@ func (p *downloadTask) Execute() {
...
+	defer func() {
+		p.traceHandle.EndSpan(span)
+		logger.Info("End span has been successfully triggerd")
```
kislaykishore:
```suggestion
```

internal/bufferedread/download_task.go
```diff
@@ -58,6 +62,14 @@ func (p *downloadTask) Execute() {
 	stime := time.Now()
 	var err error
 	var n int64
+	var span trace.Span
```
kislaykishore: How do we record errors that happen in the download task?

internal/bufferedread/buffered_reader.go
```diff
@@ -512,20 +525,21 @@ func (p *BufferedReader) scheduleNextBlock(urgent bool) error {
...
 // LOCKS_REQUIRED(p.mu)
-func (p *BufferedReader) scheduleBlockWithIndex(b block.PrefetchBlock, blockIndex int64, urgent bool) error {
+func (p *BufferedReader) scheduleBlockWithIndex(ctx context.Context, b block.PrefetchBlock, blockIndex int64, urgent bool) error {
 	startOffset := blockIndex * p.config.PrefetchBlockSizeBytes
 	if err := b.SetAbsStartOff(startOffset); err != nil {
 		return fmt.Errorf("scheduleBlockWithIndex: setting start offset: %w", err)
 	}
 
-	ctx, cancel := context.WithCancel(p.ctx)
+	ctx, cancel := context.WithCancel(ctx)
```
kislaykishore: Seems to me that we are changing the context - instead of using the buffered-reader context, we are using the context of the Read request. Could you confirm my understanding? If so, could you provide the rationale?

internal/bufferedread/buffered_reader.go
rkachewar: Can we use somekind of decorator pattern instead of instrumenting every method with multiple lines?

kislaykishore: We used to use decorators earlier but very recently we got to know that increasing stack sizes were also becoming a problem. So, we are moving away from that pattern.

thrivikram-karur-g: I think this should not be a problem given that this is in NewBufferedReader which is only called in NewReadManager and not in every call to ReadFile from kernel, atleast that is my understanding, if you want, we can do the testing.B loop benchmark test if you think it is needed.

thrivikram-karur-g: resolving this comment

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: Sounds like a good idea from the perspective of code reuse but I am concerned adding the traceHandle as member of prefetchBlock struct seems a little concerning for me.

internal/cache/file/downloader/job.go
thrivikram-karur-g: yes, we can

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: If we do not use the context of the read request then the trace wouldn't be tied up properly all the way to the top until ReadFile fs-op.

internal/gcsx/read_manager/read_manager.go
thrivikram-karur-g: Might not be a right idea. I feel this is because you have to add this same code in each ReadAt method and defeats purpose of reuse and me adding just a span to indicate which reader is serving the readfile request from kernel

internal/cache/file/downloader/job.go
thrivikram-karur-g: we are replacing this with PropagateTraceContext anyways, so this will also be resolved.

tracing/span_names.go
thrivikram-karur-g: I used gemini to suggest me appropriate names based on the naming conventions listed in the otel site.

tracing/span_names.go
thrivikram-karur-g: sure

tracing/otel_tracer.go
thrivikram-karur-g: seems like a good idea, would reduce heap allocation and GC pressure

internal/bufferedread/download_task.go
thrivikram-karur-g: I have added the RecordError code for this, also moving these to the existing defer function where metric is already being recorded.

internal/bufferedread/download_task.go
thrivikram-karur-g: cleaned up

internal/bufferedread/download_task.go
thrivikram-karur-g: cleaned up

tracing/span_names.go
thrivikram-karur-g: added

tracing/span_names.go
thrivikram-karur-g: done

tracing/span_names.go
thrivikram-karur-g: done

tracing/span_attributes.go
thrivikram-karur-g: done

tracing/span_attributes.go
thrivikram-karur-g: done

internal/cache/file/downloader/job.go
thrivikram-karur-g: updated

internal/cache/file/downloader/job.go
thrivikram-karur-g: updated

internal/gcsx/file_cache_reader.go
thrivikram-karur-g: updated

internal/gcsx/multi_range_downloader_wrapper.go
thrivikram-karur-g: Once I have your go ahead, I can ask abhishek once for review.

internal/gcsx/random_reader.go
thrivikram-karur-g: updated

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: Also from gemini regarding context.Background() .

" " " 

Calling context.Background() has a negligible CPU cost because it does not "create" a new object every time you call it. Instead, it returns a pointer to a pre-allocated, global singleton.

Source code implementation:

var (
    background = new(emptyCtx)
    todo       = new(emptyCtx)
)

func Background() Context {
    return background
}

Don't worry about calling context.Background(). It's designed to be the "root" of your request and is optimized to be as lightweight as possible

" " "

tracing/otel_tracer.go
thrivikram-karur-g: updated

tracing/otel_tracer.go
thrivikram-karur-g: updated

tracing/trace_handle.go
thrivikram-karur-g: added

tracing/otel_tracer.go
thrivikram-karur-g: added, let me know if more needs to be added

tracing/otel_tracer.go
thrivikram-karur-g: added

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: This change requires changes all the way to NewGenBlockPool and several contructs concerned if this increases scope too much

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: Same comment as above:
This change requires changes all the way to NewGenBlockPool and several contructs concerned if this increases scope too much

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: moved these changes out of this PR

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: Moved these changes out of the PR

internal/bufferedread/buffered_reader.go
thrivikram-karur-g: moved these changes for not out of the PR

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	// ReadSymlink reads the target path stored within a symbolic link.
+	ReadSymlink = "fs.symlink.read"
+	// Rename changes the name or location of a file or directory.
+	Rename = "fs.rename"
+	// Unlink removes a name from the filesystem; if it was the last link, the file is deleted.
+	Unlink = "fs.unlink"
+	// GetXattr retrieves the value of an extended attribute.
+	GetXattr = "fs.xattr.get"
+	// SetXattr sets or updates an extended attribute.
+	SetXattr = "fs.SetXattr"
```
kislaykishore: Should it be in lowercase?
```suggestion
	SetXattr = "fs.set_xattr"
```

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	ReadDir = "fs.dir.read"
+	// ReadDirPlus reads directory entries along with their associated metadata/attributes.
+	ReadDirPlus = "fs.dir.read_plus"
+	// ReleaseDirHandle closes the handle for a directory and releases associated resources.
+	ReleaseDirHandle = "fs.dir.release_handle"
+
+	// --- File Lifecycle and I/O ---
+
+	// CreateFile creates and opens a new file within the filesystem.
+	CreateFile = "ffs.file.create"
```
kislaykishore: Typo:
```suggestion
	CreateFile = "fs.file.create"
```

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	ForgetInode = "fs.inode.forget"
+	// BatchForget allows the kernel to release multiple inode references in a single call.
+	BatchForget = "ffs.batch_forget"
+
+	// --- Directory Lifecycle ---
+
+	// MkDir creates a new directory entry.
+	MkDir = "fs.mkdir"
+	// RmDir removes an existing, empty directory.
+	RmDir = "fs.rmdir"
```
kislaykishore:
```suggestion
	RmDir = "fs.dir.rm"
```

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	SetInodeAttributes = "fs.inode.set_attributes"
+	// ForgetInode informs the kernel that it no longer needs to track a specific inode.
+	ForgetInode = "fs.inode.forget"
+	// BatchForget allows the kernel to release multiple inode references in a single call.
+	BatchForget = "ffs.batch_forget"
+
+	// --- Directory Lifecycle ---
+
+	// MkDir creates a new directory entry.
+	MkDir = "fs.mkdir"
```
kislaykishore:
```suggestion
	MkDir = "fs.dir.mk"
```

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	// LookUpInode resolves a specific filename within a directory to its unique inode.
+	LookUpInode = "fs.inode.lookup"
+	// GetInodeAttributes retrieves metadata such as size, mode, and timestamps for an inode.
+	GetInodeAttributes = "fs.inode.get_attributes"
+	// SetInodeAttributes updates metadata (e.g., permissions or ownership) for an inode.
+	SetInodeAttributes = "fs.inode.set_attributes"
+	// ForgetInode informs the kernel that it no longer needs to track a specific inode.
+	ForgetInode = "fs.inode.forget"
+	// BatchForget allows the kernel to release multiple inode references in a single call.
+	BatchForget = "ffs.batch_forget"
```
kislaykishore:
```suggestion
	BatchForget = "fs.batch_forget"
```

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	// FileCacheRead tracks read operations specifically from the local file cache.
+	FileCacheRead = "file.cache.read"
+	// FileCacheWrite tracks write or population operations into the local file cache.
+	FileCacheWrite = "file.cache.write"
+	// FileDownloadJob represents a background task responsible for fetching remote file data.
+	FileDownloadJob = "file.download.job"
+	// ReadPrefetchBlockPoolGen monitors the generation/lifecycle of the prefetch block pool.
+	ReadPrefetchBlockPoolGen = "read.prefetch.block_pool_gen"
+	// DownloadPrefetchBlock triggers a network request to pre-fill a specific data block.
+	DownloadPrefetchBlock = "download.prefetch.block"
```
kislaykishore:
```suggestion
	DownloadPrefetchBlock = "prefetch.block.download"
```

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	// --- Cache and Prefetching Operations ---
+
+	// FileCacheRead tracks read operations specifically from the local file cache.
+	FileCacheRead = "file.cache.read"
+	// FileCacheWrite tracks write or population operations into the local file cache.
+	FileCacheWrite = "file.cache.write"
+	// FileDownloadJob represents a background task responsible for fetching remote file data.
+	FileDownloadJob = "file.download.job"
+	// ReadPrefetchBlockPoolGen monitors the generation/lifecycle of the prefetch block pool.
+	ReadPrefetchBlockPoolGen = "read.prefetch.block_pool_gen"
```
kislaykishore: Why does this have "read" prefix while other attributes don't have that?

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
-	Fallocate          = "Fallocate"
-	SyncFS             = "SyncFS"
+	// --- Cache and Prefetching Operations ---
+
+	// FileCacheRead tracks read operations specifically from the local file cache.
+	FileCacheRead = "file.cache.read"
+	// FileCacheWrite tracks write or population operations into the local file cache.
+	FileCacheWrite = "file.cache.write"
+	// FileDownloadJob represents a background task responsible for fetching remote file data.
+	FileDownloadJob = "file.download.job"
```
kislaykishore: Should this be `file.cache.download`?

tracing/otel_tracer.go
```diff
@@ -43,6 +60,16 @@ func (o *otelTracer) RecordError(span trace.Span, err error) {
 	span.SetStatus(codes.Error, err.Error())
 }
 
+// This method is performant in the no-op implementation as the attribute.Bool and int methods perform heap allocation and they go as bad as 60-90 ns
+// It is not good that way so the whole attribute creation has been moved inside this method keeping no-op lightweight
+// Performance characteristics for this method along with StartSpan & EndSpan together for no-op
+func (o *otelTracer) SetCacheReadAttributes(span trace.Span, isCacheHit bool, bytesRead int) {
+	span.SetAttributes(
+		attribute.Bool(IS_CACHE_HIT, isCacheHit),
+		attribute.Int(BYTES_READ, bytesRead),
```
kislaykishore: Can you consider the following optimizations:
* IS_CACHE_HIT is gonna have only two values. Would it help performance if we reuse the values?
* We can reuse keys which will hopefully reduce some heap allocations.

```
var (
  bytesReadKey = attribute.Key(BYTES_READ)
  cacheHit = attribute.Bool(IS_CACHE_HIT, true)
  cacheMiss = attribute.Bool(IS_CACHE_HIT, false)
)
func (o *otelTracer) SetCacheReadAttributes(span trace.Span, isCacheHit bool, bytesRead int) {
  bytesReadAttr :=  bytesReadKey.Int(bytesRead)
  if isCacheHit {
    span.SetAttribute(cacheHit, bytesReadAttr)
  } else {
    span.SetAttribute(cacheMiss, bytesReadAttr)
  }
}

```


With these optimizations could you share the perf difference across the following dimensions:
* heap allocations
* Time taken / operation

tracing/otel_tracer.go
```diff
@@ -43,6 +60,16 @@ func (o *otelTracer) RecordError(span trace.Span, err error) {
 	span.SetStatus(codes.Error, err.Error())
 }
 
+// This method is performant in the no-op implementation as the attribute.Bool and int methods perform heap allocation and they go as bad as 60-90 ns
```
kislaykishore:
```suggestion
// attribute creation is a costly affair both from memory allocation and CPU time perspectives - <add numbers>. This method is specifically created so that the caller doesn't have to create the attributes themselves. Instead the implementation of the TraceHandle that's chosen decides whether to create the attributes. This allows skipping the attribute creation entirely in case of noop tracer which is selected when tracing is disabled.
```

tracing/otel_tracer.go
```diff
@@ -43,6 +60,16 @@ func (o *otelTracer) RecordError(span trace.Span, err error) {
...
+// It is not good that way so the whole attribute creation has been moved inside this method keeping no-op lightweight
```
kislaykishore: Please move this documentation to the trace_handle.go.

tracing/otel_tracer.go
```diff
@@ -18,10 +18,16 @@ import (
...
 
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 )
 
+// Identifies the span link as a parent relationship for proper trace visualization in GCP.
+var parentLinkAttributes = []attribute.KeyValue{
+	attribute.Int64("gcp.cloud_trace.link_type", 1),
```
kislaykishore: Where is `gcp.cloud_trace.link_type` documented? Please add that info as a comment.

internal/gcsx/read_manager/read_manager.go
kislaykishore: Sounds good!

tracing/span_attributes.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tracing
+
+const (
+	IS_CACHE_HIT = "cache.hit"
+	BYTES_READ   = "read.size"
```
kislaykishore: Add documentation.

tracing/span_names.go
thrivikram-karur-g: I am renaming it to match other attributes

tracing/span_names.go
thrivikram-karur-g: updated

tracing/span_names.go
thrivikram-karur-g: updated

tracing/span_names.go
thrivikram-karur-g: updated

tracing/span_names.go
thrivikram-karur-g: updated

tracing/span_names.go
thrivikram-karur-g: updated

tracing/span_names.go
thrivikram-karur-g: updated

tracing/span_names.go
thrivikram-karur-g: updated

tracing/span_names.go
thrivikram-karur-g: updated

tracing/otel_tracer.go
thrivikram-karur-g: updated

tracing/otel_tracer.go
thrivikram-karur-g: moved

tracing/otel_tracer.go
thrivikram-karur-g: removed usage of spanLink for now as per our discussion.

tracing/span_attributes.go
thrivikram-karur-g: updated

tracing/otel_tracer.go
thrivikram-karur-g: @kislaykishore 

I ran the below benchmarks and it is only slightly faster. Heap allocations is the same and I think the heap allocations happen because of the variadic operators of SetAttributes majorly, below is the benchmark code and it's results. if you still think it is worth it, we can make the change.

package tracing

import (
	"context"
	"testing"

	"go.opentelemetry.io/otel/attribute"
	"go.opentelemetry.io/otel/trace"
)

var OtelTraceHandle = NewOTELTracer()

func BenchmarkNoOpTracerSetCacheReadAttributes(b *testing.B) {
	ctx := context.Background()
	_, span := OtelTraceHandle.StartSpan(ctx, "TestSpanName")

	for b.Loop() {
		OtelTraceHandle.SetCacheReadAttributes(span, true, 100)
	}
	OtelTraceHandle.EndSpan(span)
}

var (
	bytesReadKey = attribute.Key(BYTES_READ)
	cacheHit     = attribute.Bool(IS_CACHE_HIT, true)
	cacheMiss    = attribute.Bool(IS_CACHE_HIT, false)
)

func SetCacheReadAttributes(span trace.Span, isCacheHit bool, bytesRead int) {
	span.SetAttributes(
		attribute.Bool(IS_CACHE_HIT, isCacheHit),
		attribute.Int(BYTES_READ, bytesRead),
	)
}

func SetCacheReadAttributesReuse(span trace.Span, isCacheHit bool, bytesRead int) {
	bytesReadAttr := bytesReadKey.Int(bytesRead)
	if isCacheHit {
		span.SetAttributes(cacheHit, bytesReadAttr)
	} else {
		span.SetAttributes(cacheMiss, bytesReadAttr)
	}
}

func BenchmarkNoOpTracerSetCacheReadAttributesAttributeReuse(b *testing.B) {
	ctx := context.Background()

	_, span := OtelTraceHandle.StartSpan(ctx, "TestSpanName")
	for b.Loop() {
		SetCacheReadAttributesReuse(span, true, 100)
	}
	OtelTraceHandle.EndSpan(span)
}

goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/tracing
cpu: Intel(R) Xeon(R) CPU @ 2.80GHz
BenchmarkNoOpTracerSetCacheReadAttributes-64                  	16890686	        70.22 ns/op	     128 B/op	       1 allocs/op
BenchmarkNoOpTracerSetCacheReadAttributesAttributeReuse-64    	18668442	        63.34 ns/op	     128 B/op	       1 allocs/op

internal/cache/file/downloader/job.go
```diff
@@ -498,8 +498,11 @@ func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool
 	} else if job.status.Name == NotStarted {
 		// Start the async download
 		job.status.Name = Downloading
-		job.cancelCtx, job.cancelFunc = context.WithCancel(context.Background())
+		// newCtx := job.traceHandle.PropagateTraceContext(context.Background(), ctx)
```
kislaykishore: Remove the commented out code.

internal/cache/file/downloader/job.go
```diff
@@ -498,8 +498,11 @@ func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool
...
+		newCtx, downloadSpan := job.traceHandle.StartSpan(ctx, tracing.FileDownloadJob)
+		job.cancelCtx, job.cancelFunc = context.WithCancel(newCtx)
 		go job.downloadObjectAsync()
+		job.traceHandle.EndSpan(downloadSpan)
```
kislaykishore: This doesn't seem right. The downloadSpan will alway finish within a few microseconds or even lower because all you are measuring here is the time taken to create a new context and span a goroutine (which is asynchronous). So it's not measuring anything useful.

tracing/otel_tracer.go
kislaykishore: Yup, let's use the more efficient one. Could you also consider the following optimization  that tries to keep a slice ready for consumption. 

```
var (
  bytesReadKey = attribute.Key(BYTES_READ)
  cacheHit = attribute.Bool(IS_CACHE_HIT, true)
  cacheMiss = attribute.Bool(IS_CACHE_HIT, false)
  ch chan []attribute.KeyValue
  flag atomic.Bool
)

func NewOTelTracer() {
  // Choose a large value (you can go higher too) for channel size so that the benchmark doesn't run out of slices ever.
  ch := make(chan []attribute.KeyValue, 10000)
  go func() {
    for {
      select {
        ch <- make([]attribute.KeyValue, 2):
        default:
           flag.CompareAndSwap(false, true)
           // Exit after signaling that it's now safe to run the benchmark.
           return
      }
    }
  }()
   go func() {
      for {
        ch <- make([]attribute.KeyValue, 2)
      }
    }()
}

func (o *otelTracer) SetCacheReadAttributes(span trace.Span, isCacheHit bool, bytesRead int) {
  for {
    if flg.Load() {
        // Good to trigger the benchmark loop.
	break
    }
  }
  attrSet := <-ch
  attrSet[0] :=  bytesReadKey.Int(bytesRead)
  if isCacheHit {
    attrSet[1] = cacheHit
  } else {
    attrSet[1] = cacheMiss
  }
  span.SetAttributes(attrSet...)
  
}

```

tracing/span_names.go
```diff
@@ -21,35 +21,99 @@ package tracing
...
+	// ReadSymlink reads the target path stored within a symbolic link.
+	ReadSymlink = "fs.symlink.read"
+	// Rename changes the name or location of a file or directory.
+	Rename = "fs.rename"
+	// Unlink removes a name from the filesystem; if it was the last link, the file is deleted.
+	Unlink = "fs.unlink"
+	// GetXattr retrieves the value of an extended attribute.
+	GetXattr = "fs.xattr.get"
+	// SetXattr sets or updates an extended attribute.
+	SetXattr = "fs.set_xattr"
```
kislaykishore: As per convention:
```suggestion
	SetXattr = "fs.xattr.get"
```

internal/cache/file/downloader/job.go
thrivikram-karur-g: yeah, agree, I checked how the traces look even without this span, I think it gives enough information for debugging and yes, this always takes few microseconds and doesn't seem to very useful. Removing it.

internal/cache/file/downloader/job.go
thrivikram-karur-g: updated

internal/cache/file/downloader/job.go
thrivikram-karur-g: removed

tracing/span_names.go
thrivikram-karur-g: updated

tracing/otel_tracer.go
thrivikram-karur-g: I have tried this but it is running into issues and benchmarking is stuck forever and never successfully ran.
Also assigning such a large channel and using the idea of buffered pooling is great but seems like would add much more changes to this PR. Would recommend this as a follow up instead.

thrivikram-karur-g: I was also actually able to run the benchmark tests as well.

Here is the comparision and it seems to be pretty efficient as no allocs are happening in the 

goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/tracing
cpu: Intel(R) Xeon(R) CPU @ 2.80GHz

BenchmarkNoOpTracerSetCacheReadAttributes-64                   	13449895	        90.86 ns/op	     128 B/op	       1 allocs/op
BenchmarkNoOpTracerSetCacheReadAttributesAttributeReuse-64     	13354520	        91.30 ns/op	     128 B/op	       1 allocs/op
BenchmarkNoOpTracerSetCacheReadAttributesBufferedPooling-64    	19592715	        60.66 ns/op	       0 B/op	       0 allocs/op

90 ns to 60 ns is really good but then originally the SetCacheReadAttributes itself is meant to be 60ns.
Let me reverify regarding this discrepancy.

thrivikram-karur-g: Here is the final code & result.

package tracing

import (
	"context"
	"testing"

	"go.opentelemetry.io/otel/attribute"
	"go.opentelemetry.io/otel/trace"
)

var OtelTraceHandle = NewOTELTracer()

func BenchmarkNoOpTracerSetCacheReadAttributes(b *testing.B) {
	ctx := context.Background()
	_, span := OtelTraceHandle.StartSpan(ctx, "TestSpanName")

	for b.Loop() {
		OtelTraceHandle.SetCacheReadAttributes(span, true, 100)
	}
	OtelTraceHandle.EndSpan(span)
}

func otelTracerAttributeChannelInitialisation() chan []attribute.KeyValue {
	// Choose a large value (you can go higher too) for channel size so that the benchmark doesn't run out of slices ever.
	ch := make(chan []attribute.KeyValue, 10000)
	for i := 0; i < cap(ch); i++ {
		ch <- make([]attribute.KeyValue, 2)
	}
	return ch
}

func SetCacheReadAttributes(span trace.Span, isCacheHit bool, bytesRead int) {
	span.SetAttributes(
		attribute.Bool(IS_CACHE_HIT, isCacheHit),
		attribute.Int(BYTES_READ, bytesRead),
	)
}

func SetCacheReadAttributesReuse(span trace.Span, isCacheHit bool, bytesRead int) {
	bytesReadAttr := bytesReadKey.Int(bytesRead)
	if isCacheHit {
		span.SetAttributes(cacheHit, bytesReadAttr)
	} else {
		span.SetAttributes(cacheMiss, bytesReadAttr)
	}
}

func SetCacheReadAttributesReuseBufferPooling(span trace.Span, isCacheHit bool, bytesRead int, ch chan []attribute.KeyValue) {
	attrSet := <-ch
	attrSet[0] = bytesReadKey.Int(bytesRead)
	if isCacheHit {
		attrSet[1] = cacheHit
	} else {
		attrSet[1] = cacheMiss
	}
	span.SetAttributes(attrSet...)
	ch <- attrSet
}

func BenchmarkNoOpTracerSetCacheReadAttributesAttributeReuse(b *testing.B) {
	ctx := context.Background()

	_, span := OtelTraceHandle.StartSpan(ctx, "TestSpanName")

	for b.Loop() {
		SetCacheReadAttributesReuse(span, true, 100)
	}
	OtelTraceHandle.EndSpan(span)
}

func BenchmarkNoOpTracerSetCacheReadAttributesBufferedPooling(b *testing.B) {
	ctx := context.Background()
	ch := otelTracerAttributeChannelInitialisation()

	_, span := OtelTraceHandle.StartSpan(ctx, "TestSpanName")

	for b.Loop() {
		SetCacheReadAttributesReuseBufferPooling(span, true, 100, ch)
	}
	OtelTraceHandle.EndSpan(span)
}

BenchmarkNoOpTracerSetCacheReadAttributes-64                   	13562260	        85.34 ns/op	     128 B/op	       1 allocs/op
BenchmarkNoOpTracerSetCacheReadAttributesAttributeReuse-64     	13642044	        86.85 ns/op	     128 B/op	       1 allocs/op
BenchmarkNoOpTracerSetCacheReadAttributesBufferedPooling-64    	19603456	        60.75 ns/op	       0 B/op	       0 allocs/op

thrivikram-karur-g: updated kislay

thrivikram-karur-g:
> I see this test failing: https://screenshot.googleplex.com/7L4dAiEjugBfQZF. Please check.

Yeah, I am debugging the same

thrivikram-karur-g:
> I still see tests failing - https://screenshot.googleplex.com/5PZZ2jQRu3DXRKg. BTW, would it be simpler to add a check [here](https://screenshot.googleplex.com/6VD4nRLbSsfBeFr) that
> 
> ```
> if config.TraceHandle == nil {
>   config.TraceHandle = tracing.NewNoopTracer()
> }
> ```

yeah, even I am thinking the same.

internal/cache/file/downloader/job.go
```diff
@@ -498,7 +498,7 @@ func (job *Job) Download(ctx context.Context, offset int64, waitForDownload bool
 	} else if job.status.Name == NotStarted {
 		// Start the async download
 		job.status.Name = Downloading
-		job.cancelCtx, job.cancelFunc = context.WithCancel(context.Background())
+		job.cancelCtx, job.cancelFunc = context.WithCancel(ctx)
```
kislaykishore: Not sure if it was discussed earlier. Could you explain why we are changing the context?

internal/gcsx/client_readers/multi_range_reader.go
```diff
@@ -38,12 +39,15 @@ type MultiRangeReader struct {
 	isMRDInUse atomic.Bool
 
 	metricHandle metrics.MetricHandle
+
+	traceHandle tracing.TraceHandle
 }
 
-func NewMultiRangeReader(object *gcs.MinObject, metricHandle metrics.MetricHandle, mrdWrapper *gcsx.MultiRangeDownloaderWrapper) *MultiRangeReader {
+func NewMultiRangeReader(object *gcs.MinObject, metricHandle metrics.MetricHandle, traceHandle tracing.TraceHandle, mrdWrapper *gcsx.MultiRangeDownloaderWrapper) *MultiRangeReader {
```
kislaykishore, line 46: As a defensive coding practice, I think it'd be safer to assign a Noop implementation of traceHandle if the value is nil - here and elsewhere. Otherwise, I'm afraid if we miss a case, it'd lead to a crash in production.

tracing/otel_tracer.go
```diff
@@ -26,6 +27,13 @@ type otelTracer struct {
 	tracer trace.Tracer
 }
 
+var (
+	bytesReadKey = attribute.Key(BYTES_READ)
+	cacheHit     = attribute.Bool(IS_CACHE_HIT, true)
+	cacheMiss    = attribute.Bool(IS_CACHE_HIT, false)
+	ch           chan []attribute.KeyValue
```
kislaykishore: This should be a data member in otelTracer struct. As explained later, I think instead of this, a pool would be a better implementation compared to channels.

tracing/otel_tracer.go
```diff
@@ -43,12 +51,29 @@ func (o *otelTracer) RecordError(span trace.Span, err error) {
...
 	span := trace.SpanFromContext(oldCtx)
 	return trace.ContextWithSpan(newCtx, span)
 }
 
 func NewOTELTracer() TraceHandle {
+	ch = make(chan []attribute.KeyValue, 5)
+	for i := 0; i < cap(ch); i++ {
+		ch <- make([]attribute.KeyValue, 2)
+	}
+
```
kislaykishore, line 80: It's a good idea to return the slice back to the channel. But, we are introducing a potential contention.
Say a 100 threads arrive concurrently, only 5 would be served at a time. The rest will be blocked waiting for the slices to become available again. I'd therefore advise for a more asynchronous solution. The following implementation prioritizes speed over memory allocations.
```suggestion

func (o *otelTracer) SetCacheReadAttributes(span trace.Span, isCacheHit bool, bytesRead int) {
	attrSet := o.slicePool.Get().([]attribute.KeyValue)
	defer func() {
	  o.slicePool.Put(attrSet)
	}
	attrSet[0] = bytesReadKey.Int(bytesRead)
	if isCacheHit {
		attrSet[1] = cacheHit
	} else {
		attrSet[1] = cacheMiss
	}
	span.SetAttributes(attrSet...)
}

func (o *otelTracer) PropagateTraceContext(newCtx context.Context, oldCtx context.Context) context.Context {
span := trace.SpanFromContext(oldCtx)
return trace.ContextWithSpan(newCtx, span)
}

func NewOTELTracer() TraceHandle {
    slicePool := sync.Pool {
      New: func() interface{} {
        return make([]attribute.KeyValue, 2)
      },
      }
      // construct otelTracer struct
      return &otelTracer {
        slicePool: slicePool,
        tracer: otel.Tracer(name),
      }
  }
```

internal/gcsx/client_readers/multi_range_reader.go
thrivikram-karur-g: yeah, actually after seeing so many linux tests failure where it was fixed only in test mocks, even I had the same concern afterwards. Let me actually add it everywhere where it is important. yes, same concern I had that one missed case is production issue.

thrivikram-karur-g: let me work on this and get back

tracing/otel_tracer.go
thrivikram-karur-g: Actually, I too wanted to bring this up as the benchmark loop on Parallel runs had very poor performance.
I will update code with your suggestions to see that it performs well with parallel execution.

internal/gcsx/file_cache_reader.go
```diff
@@ -101,6 +105,7 @@ func NewFileCacheReader(o *gcs.MinObject, bucket gcs.Bucket, fileCacheHandler *f
 // fileHandle to file in cache. So, we will get the correct data from fileHandle
 // because Linux does not delete a file until open fileHandle count for a file is zero.
 func (fc *FileCacheReader) tryReadingFromFileCache(ctx context.Context, p []byte, offset int64) (int, bool, error) {
+	ctx, span := fc.traceHandle.StartSpan(ctx, tracing.FileCacheRead)
 	if fc.fileCacheHandler == nil {
 		return 0, false, nil
 	}
```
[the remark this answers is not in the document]
thrivikram-karur-g: agreed, the startSpan should be after this when fileCacheHandler is even defined in the first place.

internal/cache/file/downloader/job.go
thrivikram-karur-g: this is so that the parent file cache reader which triggers there download jobs will have the visibility and be connected in one trace. if not the download jobs and parent are disconnected.
If you see, cancelCtx is further used down in the job.

https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/cache/file/downloader/job.go#L323

I too have once concern if this context.Background is explicitly done to ensure download is not cancelled if parent ends or is cancelled, that I am not sure but then I am not sure of an easy fix in the case.

kislaykishore: Wouldn't it work if we copy over the span context to the context returned by context.Background()?

thrivikram-karur-g: yes, we can do that. That is more better and should remove any concerns we have as well. let me change it.

tracing/otel_tracer.go
thrivikram-karur-g: thanks for this suggestion.

Here are the benchmarks with parallel runs as well.

goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/tracing
cpu: Intel(R) Xeon(R) CPU @ 2.80GHz

BenchmarkNoOpTracerSetCacheReadAttributes-64                           	23042205	        50.76 ns/op	      24 B/op	       1 allocs/op

BenchmarkNoOpTracerSetCacheReadAttributesParallel-64                   	45872264	        24.38 ns/op	      24 B/op	       1 allocs/op

BenchmarkNoOpTracerSetCacheReadAttributesAttributeReuse-64             	11538547	       103.5 ns/op	     128 B/op	       1 allocs/op

BenchmarkNoOpTracerSetCacheReadAttributesAttributeReuseParallel-64     	16647180	        67.36 ns/op	     128 B/op	       1 allocs/op

BenchmarkNoOpTracerSetCacheReadAttributesBufferedPooling-64            	21727167	        56.56 ns/op	       0 B/op	       0 allocs/op

BenchmarkNoOpTracerSetCacheReadAttributesBufferedPoolingParallel-64    	 1498582	      1006 ns/op	       0 B/op	       0 allocs/op

PASS
ok  	github.com/googlecloudplatform/gcsfuse/v3/tracing	9.443s

thrivikram-karur-g: this is with the sync Pool code incorporated.

tracing/otel_tracer.go
thrivikram-karur-g: Anyways, now moved only the syncPool to the struct

tracing/otel_tracer.go
```diff
@@ -43,12 +51,29 @@ func (o *otelTracer) RecordError(span trace.Span, err error) {
...
 func (o *otelTracer) PropagateTraceContext(newCtx context.Context, oldCtx context.Context) context.Context {
 	span := trace.SpanFromContext(oldCtx)
 	return trace.ContextWithSpan(newCtx, span)
 }
 
 func NewOTELTracer() TraceHandle {
+	ch = make(chan []attribute.KeyValue, 5)
+	for i := 0; i < cap(ch); i++ {
+		ch <- make([]attribute.KeyValue, 2)
+	}
```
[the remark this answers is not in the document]
thrivikram-karur-g: no longer applicable given we are using sync Pool now.

internal/gcsx/file_cache_reader.go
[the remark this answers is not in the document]
thrivikram-karur-g: updated

internal/cache/file/downloader/job.go
thrivikram-karur-g: Have updated

internal/gcsx/client_readers/multi_range_reader.go
thrivikram-karur-g: Have added nil checks in every reader constructor including even job constructor but feels pretty repetitive.

tracing/otel_tracer.go
thrivikram-karur-g: the allocations per op is actually there and hasn't reduced but it is way faster, not sure if this is more because of parallelism.

internal/gcsx/multi_range_downloader_wrapper.go
```diff
@@ -200,7 +201,8 @@ func (mrdWrapper *MultiRangeDownloaderWrapper) CloseMRDForEviction() {
 
 // Ensures that MultiRangeDownloader exists, creating it if it does not exist.
 // LOCK_REQUIRED(mrdWrapper.mu.RLock)
-func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader(forceRecreateMRD bool) (err error) {
+func (mrdWrapper *MultiRangeDownloaderWrapper) ensureMultiRangeDownloader(ctx context.Context, traceHandle tracing.TraceHandle, forceRecreateMRD bool) (err error) {
+	ctx = traceHandle.PropagateTraceContext(context.Background(), ctx)
```
thrivikram-karur-g: push it close to usage

thrivikram-karur-g: done

tracing/otel_tracer.go
thrivikram-karur-g: done

internal/gcsx/multi_range_downloader_wrapper.go
thrivikram-karur-g: as discussed, resolving this

tracing/otel_tracer_test.go
```diff
@@ -0,0 +1,150 @@
...
+		tracer.SetCacheReadAttributes(span, false, bytesRead)
+		tracer.EndSpan(span)
+
+		spans := recorder.Ended()
+		assert.Len(t, spans, 1)
+		assert.Len(t, spans[0].Attributes(), 2)
+		assert.Contains(t, spans[0].Attributes(), attribute.Int(BYTES_READ, bytesRead))
+		assert.Contains(t, spans[0].Attributes(), attribute.Bool(IS_CACHE_HIT, false))
+	})
+}
```
kislaykishore, line 125: You could use a table-driven test for this - https://go.dev/wiki/TableDrivenTests

tracing/otel_tracer_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func TestNewOTELTracer(t *testing.T) {
```
kislaykishore: Let's remove this test. These are internal details of the implementation and shouldn't be tested. We should only test the behavior i.e. whether traces are being logged correctly.

tracing/otel_tracer_test.go
thrivikram-karur-g: updated

tracing/otel_tracer_test.go
thrivikram-karur-g: removed

PR #4247, opened 2026-01-07
fix: network unreachable error

Tulsishah:
### Description
We encountered connect: network is unreachable errors during storage client calls within our end-to-end (e2e) tests. This client is initialized directly within the test suite to validate GCSFuse results against GCS results.

Created separate ShouldRetry function for test only so that core logic won't get change. 

[b/470721588](https://b.corp.google.com/issues/470721588)

### Testing details
1. Manual - Tested through louhi flow ([b/470721588#comment4](https://b.corp.google.com/issues/470721588#comment4))
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -57,12 +57,12 @@ func ShouldRetryForTest(err error) (b bool) {
 		return
 	}
 
-	if err != nil && strings.Contains(err.Error(), networkUnreachableError) {
+	// Convert err.Error() to lowercase to make the check case-insensitive
+	if err != nil && strings.Contains(strings.ToLower(err.Error()), networkUnreachableError) {
 		b = true
 		log.Printf("Retrying for 'network is unreachable' error: %v", err)
 		return
 	}
-
 	return
 }
 


internal/storage/storageutil/custom_retry.go
```diff
@@ -58,6 +61,13 @@ func ShouldRetry(err error) (b bool) {
 			return
 		}
 	}
+
+	if err != nil && strings.Contains(err.Error(), networkUnreachableError) {
```
raj-prince: Don't we get any error code? Message based retry doesn't look robust, ideally we should get a code from go-storage.

abhishek10004: +1

Tulsishah: No specific error code is being returned. Error is happening at network layer.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -35,17 +35,64 @@ import (
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
+const maxStorageClientRetryAttempts = 10
+const networkUnreachableError = "network is unreachable"
+
+func ShouldRetryForTest(err error) (b bool) {
+	b = storage.ShouldRetry(err)
+	if b {
```
abhishek10004, line 55: Instead of duplicating the code, let's call storageUtil.ShouldRetry and add our additional case after that here.

Tulsishah: Good Idea. Thanks. Done.

abhishek10004 (approved):
LGTM

PR #4251, opened 2026-01-07
refactor: propagate config to dir inode

ashmeenkaur:
### Description
Instead of passing individual config parameters, pass reference of config file to dir inode. 

b/473411326

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/googlecloudplatform/gcsfuse/v3/cfg"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/cache/metadata"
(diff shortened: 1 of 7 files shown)


internal/fs/inode/dir.go
```diff
@@ -276,16 +277,12 @@ func NewDirInode(
 	name Name,
 	attrs fuseops.InodeAttributes,
 	implicitDirs bool,
```
charith87, line 279: Does implicitDirs and enableNonexistentTypeCache and TypeCacheTTL also come in Config. Can they also be removed.

ashmeenkaur: These 3 can be removed from fs layer as well. I will do this in a separate PR.

charith87: Sg

charith87 (approved):

PR #4258, opened 2026-01-09
feat(simple reader): Adding metrics support in mrd_simple_reader

abhishek10004:
### Description
Collecting following metrics in mrd_simple_reader
1. Add call - request count & latency
2. read count & download_bytes_count

b/474070368

### Testing details
1. Manual - Done
3. Unit tests - NA
4. Integration tests - NA

### DIFF
--- a/internal/fs/metrics_test.go
+++ b/internal/fs/metrics_test.go
@@ -583,6 +583,7 @@ func TestReadFile_MrdSimpleReaderMetrics(t *testing.T) {
 	require.NoError(t, err, "ReadFile")
 	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/read_count", attribute.NewSet(attribute.String("read_type", string(metrics.ReadTypeParallelAttr))), int64(1))
 	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/download_bytes_count", attribute.NewSet(attribute.String("read_type", string(metrics.ReadTypeParallelAttr))), int64(len(content)))
+	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/request_count", attribute.NewSet(attribute.String("gcs_method", "MultiRangeDownloader::Add")), int64(1))
 	metrics.VerifyHistogramMetric(t, ctx, reader, "gcs/request_latencies", attribute.NewSet(attribute.String("gcs_method", "MultiRangeDownloader::Add")), uint64(1))
 }
 
--- a/internal/gcsx/mrd_instance.go
+++ b/internal/gcsx/mrd_instance.go
@@ -52,16 +52,13 @@ type MrdInstance struct {
 	poolMu sync.RWMutex
 	// mrdCache is a shared cache for inactive MrdInstance objects.
 	mrdCache *lru.Cache
-	// mrdConfig holds configuration for the MRD pool.
+	// holds config specified by the user using config-file flag and CLI flags.
 	config *cfg.Config
 }
 
 // NewMrdInstance creates a new MrdInstance for a given GCS object.
 // Passed config should not be nil. Code will panic otherwise.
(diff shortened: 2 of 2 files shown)


internal/gcsx/mrd_simple_reader.go
```diff
@@ -55,8 +59,9 @@ func (msr *MrdSimpleReader) ReadAt(ctx context.Context, req *ReadRequest) (ReadR
 		msr.mrdInstance.IncrementRefCount()
 	}
 
-	n, err := msr.mrdInstance.Read(ctx, req.Buffer, req.Offset)
-	return ReadResponse{Size: n}, err
+	bytesRead, err := msr.mrdInstance.Read(ctx, req.Buffer, req.Offset, msr.metrics)
+	metrics.CaptureGCSReadMetrics(msr.metrics, metrics.ReadTypeParallelAttr, int64(bytesRead))
```
vadlakondaswetha, line 93: dont we need a unit test which checks if this specific metric is getting captured?

abhishek10004: Added UT.

internal/gcsx/mrd_instance.go
```diff
@@ -50,17 +53,21 @@ type MrdInstance struct {
 	// mrdCache is a shared cache for inactive MrdInstance objects.
 	mrdCache *lru.Cache
 	// mrdConfig holds configuration for the MRD pool.
```
raj-prince: nit: mrdConfig -> config

internal/gcsx/mrd_instance.go
```diff
@@ -50,17 +53,21 @@ type MrdInstance struct {
...
-	mrdConfig cfg.MrdConfig
+	config *cfg.Config
 }
 
 // NewMrdInstance creates a new MrdInstance for a given GCS object.
-func NewMrdInstance(obj *gcs.MinObject, bucket gcs.Bucket, cache *lru.Cache, inodeId fuseops.InodeID, cfg cfg.MrdConfig) *MrdInstance {
+// Passed config should not be nil. Code will panic otherwise.
+func NewMrdInstance(obj *gcs.MinObject, bucket gcs.Bucket, cache *lru.Cache, inodeId fuseops.InodeID, config *cfg.Config) *MrdInstance {
+	if config == nil {
+		panic("NewMrdInstance: config cannot be nil")
```
raj-prince: nir: generally we avoid panic. can't we return error?

internal/fs/metrics_test.go
```diff
@@ -543,6 +549,43 @@ func TestSparseReadFile_GCSReadMetrics(t *testing.T) {
...
+	}
+
+	err = server.ReadFile(ctx, readOp)
+	require.NoError(t, err, "ReadFile")
+	waitForMetricsProcessing()
+
+	require.NoError(t, err, "ReadFile")
+	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/read_count", attribute.NewSet(attribute.String("read_type", string(metrics.ReadTypeParallelAttr))), int64(1))
+	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/download_bytes_count", attribute.NewSet(attribute.String("read_type", string(metrics.ReadTypeParallelAttr))), int64(len(content)))
+	metrics.VerifyHistogramMetric(t, ctx, reader, "gcs/request_latencies", attribute.NewSet(attribute.String("gcs_method", "MultiRangeDownloader::Add")), uint64(1))
```
raj-prince, line 587: Can we also assert for request_count for gcs_method "MultiRangeDownloader:Add"?

internal/gcsx/mrd_instance.go
abhishek10004: Config will never be nil as it will result in crashes early on. I'll add a check while creating filesystem for this. Have removed this check from here.

raj-prince (approved):

PR #4259, opened 2026-01-09
feat(simple reader): Updating MRDInstance when minObject is updated in inode

abhishek10004:
### Description
Adding changes to ensure minObject is updated in MrdInstance as well when it's updated in inode. Also ensuring that we decrement the refCount from simple reader only when it was incremented from that reader as well.

b/474070744

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/mrd_instance_test.go
+++ b/internal/gcsx/mrd_instance_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
@@ -672,7 +673,7 @@ func (t *MrdInstanceTest) TestGetMinObject() {
 
 func (t *MrdInstanceTest) TestClosePoolWithTimeout_LogWarningOnTimeout() {
 	// 1. Capture logs.
-	var buf bytes.Buffer
+	var buf logBuffer
 	logger.SetOutput(&buf)
 	defer logger.SetOutput(os.Stdout)
 	// 2. Create a pool that blocks on Close().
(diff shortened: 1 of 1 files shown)


internal/fs/inode/file.go
```diff
@@ -986,7 +986,7 @@ func (f *FileInode) updateInodeStateAfterSync(minObj *gcs.MinObject) {
 	if minObj != nil && !f.localFileCache {
 		f.src = *minObj
 		// Update MRDWrapper
-		f.updateMRDWrapper()
+		f.updateMRD()
```
vadlakondaswetha, line 992: after the inode gets updated with new generation we should invalidate the MRD instance and create it again

abhishek10004: Since we're leaving object unfinalized, then the generation won't change and we don't have to recreate the MRD. I can recreate it if FinalizeFileForRapid is set to true. What do you think?

abhishek10004: Handled generation change scenario in SetMinObject.

internal/gcsx/mrd_instance.go
```diff
@@ -57,15 +57,66 @@ type MrdInstance struct {
...
+	mi.mrdPool = newPool
+	mi.object = obj
+
+	if oldPool != nil {
+		go oldPool.Close()
+	}
+	return nil
+}
+
+// GetMinObject returns the gcs.MinObject stored in MrdInstance.
```
raj-prince: Is this only for testing? If yes, can we mention that in the comment.

internal/gcsx/mrd_instance.go
```diff
@@ -182,36 +233,29 @@ func (mi *MrdInstance) ensureMRDPool() (err error) {
...
 	mi.poolMu.Lock()
 	defer mi.poolMu.Unlock()
-	if mi.mrdPool != nil {
-		// Close the pool.
-		mi.mrdPool.Close()
-		mi.mrdPool = nil
+	pool := mi.mrdPool
+	mi.mrdPool = nil
+	if pool != nil {
+		go pool.Close()
```
raj-prince: Is this just for performance? Do we know, how much will be save? I am just worried about the go-routine resource leak, in case multiple such Close gets stuck.

abhishek10004: Yes, it's for performance. It will wait for the active requests on the MRD to complete and then destroy the MRD. It will do this serially for each MRD in the pool. 
Could you explain the resource-leak that you're thinking about in slightly more detail. Why do you think this can get stuck?

abhishek10004: Added observability.

raj-prince (approved):

raj-prince (approved):

PR #4260, opened 2026-01-09
feat(dir-metadata-prefetch):  implementation for async metadata prefetch

ashmeenkaur:
### Description
This PR implements the core logic for asynchronous directory metadata prefetching in GCSfuse. The feature is designed to reduce latency for sequential lookups by proactively fetching and caching metadata for sibling objects when a lookup for a specific child results in a cache miss.

**Key Changes**
1. Prefetch Logic in dirInode
   - Trigger: When LookUpChild encounters an UnknownType (cache miss), it now launches a background prefetch worker if the feature is enabled.
   - Background Worker: The runOnDemandPrefetch function performs asynchronous GCS list calls. It is designed to run without holding the inode lock, ensuring that other file system operations are not blocked while the cache is being populated
    - Concurrency Control: Introduced an atomic prefetchState (using prefetchReady and prefetchInProgress constants) to ensure that only one prefetch operation runs at a time per directory inode.
    - Large Directory Optimization: Added an isLargeDir flag. If a directory's size exceeds maxPrefetchCount, subsequent prefetches start from the last looked-up object's offset to prioritize caching its immediate lexicographical siblings.
    - Lifecycle Management: Prefetching is tied to a prefetchCtx which is cancelled when the inode is destroyed, preventing leaked background goroutines.

2. Storage Layer Enhancements
    - ListObjects API: Added a StartOffset field to ListObjectsRequest to support filtered GCS listing.
    - Fake Storage: Updated the fake storage bucket implementation to respect StartOffset, facilitating robust unit testing.
    - Internal Refactoring: Updated listObjectsAndBuildCores and readObjectsUnlocked to support configurable result limits (maxListCallResults) and offsets.

3. Configuration
    - The feature is gated behind an experimental configuration flag: `--experimental-dir-metadata-prefetch`

### Testing details
1. Manual - NA
2. Unit tests - Added + Updated

NA

### DIFF
--- a/internal/fs/inode/dir_prefetch_test.go
+++ b/internal/fs/inode/dir_prefetch_test.go
@@ -169,11 +169,6 @@ func (t *DirPrefetchTest) TestPrefetch_CancellationOnDestroy() {
 	_, _ = t.in.LookUpChild(t.ctx, "file0001")
 	t.in.mu.Unlock()
 
-	// Ensure it started.
-	assert.Eventually(t.T(), func() bool {
-		return t.in.prefetchState.Load() == prefetchInProgress
-	}, 1*time.Second, 5*time.Millisecond)
-
 	// Destroy the inode, which calls prefetchCancel().
 	err := t.in.Destroy()
 	require.NoError(t.T(), err)


internal/fs/inode/dir.go
```diff
@@ -481,6 +509,74 @@ func (d *dirInode) createNewObject(
...
+		if err != nil {
+			logger.Warnf("Prefetch failed for %s: %v", d.Name().GcsObjectName(), err)
+			return // Abort.
+		}
+
+		var keys []string
+		for _, a := range cores {
+			keys = append(keys, a.FullName.GcsObjectName())
+		}
+		slices.Sort(keys)
```
charith87: Do we need the sorting here. GCS response is anyway paginated with some sorting already.

internal/fs/inode/dir.go
```diff
@@ -481,6 +509,74 @@ func (d *dirInode) createNewObject(
...
+		logger.Infof("keys = %v", keys)
```
charith87: If this is temporary. Pls add a todo that this will be removed or trace.

internal/fs/inode/dir.go
```diff
@@ -42,6 +45,12 @@ import (
 // Defining a constant to set maxResults param.
 const MaxResultsForListObjectsCall = 5000
```
charith87, line 44: Add a Todo that there would be a hidden config for this.

charith87: I see there is a todo in the other part. Resolving this.

internal/fs/inode/dir.go
ashmeenkaur: Nope sorry, I added this for debugging. Removed now.

internal/fs/inode/dir.go
```diff
@@ -481,6 +508,67 @@ func (d *dirInode) createNewObject(
...
+		select {
+		case <-ctx.Done():
+			logger.Debugf("Metadata prefetch for directory inode %d aborted: context cancelled.", d.id)
+			return
+		default:
+		}
+
+		// Calculate how many results to ask for in this batch.
+		remaining := d.maxPrefetchCount - totalPrefetched
+		batchSize := int(math.Min(float64(remaining), float64(MaxResultsForListObjectsCall)))
```
charith87: *nit = Are these float64 type conversions needed since all are integers here.

ashmeenkaur: Thanks, figured out we could've used standard min instead!

charith87 (dismissed):
Dont want to block this PR. But i feel the vars and the business logic of the prefetch could all go into a struct. So the dir.go original functionality is isolated. Maybe check if at it in future PRs when you get a chance.

ashmeenkaur:
> Dont want to block this PR. But i feel the vars and the business logic of the prefetch could all go into a struct. So the dir.go original functionality is isolated. Maybe check if at it in future PRs when you get a chance.

Thanks, I kept the logic in `dir.go` to avoid duplicating the existing listing logic. However thinking a bit more about this, I agree that keeping it in a separate struct will be better. I'll address this in a future PR.

PR #4262, opened 2026-01-09
feat(simple reader): Destroy MRDinstance & remove it from cache when inode is destroyed

abhishek10004:
### Description
1. Destroy MRDinstance & remove it from cache when inode is destroyed.
2. Handled an edge case, when MRDInstance was re-added to the cache before eviction, in DecrementRefCount flow.
3. Refactored the code.
4. Added UTs

b/474070647

### Testing details
1. Manual - NA
5. Unit tests - Done
6. Integration tests - NA

### DIFF
--- a/internal/fs/inode/file_test.go
+++ b/internal/fs/inode/file_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/storageutil"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
+	"github.com/googlecloudplatform/gcsfuse/v3/metrics"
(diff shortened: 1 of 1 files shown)


internal/gcsx/mrd_instance_test.go
```diff
@@ -409,3 +381,163 @@ func (t *MrdInstanceTest) TestEnsureMRDPool_Failure() {
...
+	// Initialize pool.
+	err := mi.ensureMRDPool()
+	assert.NoError(t.T(), err)
+	mi.refCount = 1
+	// Capture logs to verify error message
+	var buf bytes.Buffer
+	logger.SetOutput(&buf)
+	defer logger.SetOutput(os.Stdout)
+
+	// This should fail to insert into cache (Size 2 > Cap 1) and trigger Destroy
```
raj-prince: Although, not a real life scenario.

But it will just close the pool not destroy the mrdInstance, right?

abhishek10004: yes, it will close the pool. Let me update the comment.

PR #4264, opened 2026-01-11
fix(cfg): respect config file values in optimizations

vipnydav:
### Description
Previously, GCSFuse relied on a custom `pflag` adapter and the `pflag.Changed()` method to determine if a configuration flag had been explicitly set by the user before applying automatic optimizations. This approach was flawed because it only detected flags provided via the command-line interface and failed to account for values defined in configuration files. Consequently, user-defined settings in a config file were being unintentionally ignored and overwritten by automatic optimization logic if the corresponding CLI flag was not also provided.

To resolve this issue, the optimization logic has been updated to leverage the existing `viper.IsSet()` functionality, which correctly identifies if a value was provided through either a CLI flag or the configuration file. Along with this change, internal flag references in the optimization logic and templates were transitioned to use full hierarchical configuration paths, such as `file-system.congestion-threshold`, to align with the structure used by the configuration manager.

https://b.corp.google.com/issues/474942076

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -45,7 +45,7 @@ type mountInfo struct {
 	// optimizedFlags contains the flags that were optimized
 	// based on either machine-type or profile.
 	optimizedFlags map[string]any
-	// userConfig is used to check if a flag was explicitly set by the user.
+	// viperConfig is used to check if a flag was explicitly set by the user.
 	// This is used to determine if optimization rules should be applied.
 	viperConfig *viper.Viper
 }
--- a/cfg/optimize_test.go
+++ b/cfg/optimize_test.go
@@ -1,3 +1,15 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+
 package cfg
 
 import (
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -179,7 +179,7 @@ func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle m
 ////////////////////////////////////////////////////////////////////////
 
 // Mount the file system according to arguments in the supplied context.
-func mountWithArgs(bucketName string, mountPoint string, newConfig *cfg.Config, metricHandle metrics.MetricHandle, traceHandle tracing.TraceHandle, userConfig *viper.Viper) (mfs *fuse.MountedFileSystem, err error) {
+func mountWithArgs(bucketName string, mountPoint string, newConfig *cfg.Config, metricHandle metrics.MetricHandle, traceHandle tracing.TraceHandle, viperConfig *viper.Viper) (mfs *fuse.MountedFileSystem, err error) {
 	// Enable invariant checking if requested.
 	if newConfig.Debug.ExitOnInvariantViolation {
 		locker.EnableInvariantsCheck()
@@ -215,7 +215,7 @@ func mountWithArgs(bucketName string, mountPoint string, newConfig *cfg.Config, 
 		storageHandle,
 		metricHandle,
 		traceHandle,
-		userConfig)
+		viperConfig)
 
 	if err != nil {
 		err = fmt.Errorf("mountWithStorageHandle: %w", err)
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -46,7 +46,7 @@ func mountWithStorageHandle(
 	storageHandle storage.StorageHandle,
 	metricHandle metrics.MetricHandle,
 	traceHandle tracing.TraceHandle,
-	userConfig *viper.Viper) (mfs *fuse.MountedFileSystem, err error) {
+	viperConfig *viper.Viper) (mfs *fuse.MountedFileSystem, err error) {
 
 	// Sanity check: make sure the temporary directory exists and is writable
 	// currently. This gives a better user experience than harder to debug EIO
@@ -128,7 +128,7 @@ be interacting with the file system.`)
 		SequentialReadSizeMb:       int32(newConfig.GcsConnection.SequentialReadSizeMb),
 		EnableNonexistentTypeCache: newConfig.MetadataCache.EnableNonexistentTypeCache,
 		NewConfig:                  newConfig,
-		UserConfig:                 userConfig,
+		ViperConfig:                viperConfig,
 		MetricHandle:               metricHandle,
 		TraceHandle:                traceHandle,
 	}
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -137,9 +137,9 @@ type ServerConfig struct {
 	// NewConfig has all the config specified by the user using config-file or CLI flags.
 	NewConfig *cfg.Config
 
-	// UserConfig tracks which flags were explicitly set by the user (vs defaults)
+	// ViperConfig tracks which flags were explicitly set by the user (vs defaults)
 	// This is used to determine if optimization rules should be applied.
-	UserConfig *viper.Viper
+	ViperConfig *viper.Viper
 
 	MetricHandle metrics.MetricHandle
 
@@ -260,16 +260,16 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		// bucketHandle to avoid duplicated network calls) would be needed to change this.
 		// IMPACT: Flags are used after this. Optimization/rationalization functions are called twice
 		// for non-dynamic mounts, but they are idempotent, so it's safe.
-		if serverCfg.UserConfig != nil {
+		if serverCfg.ViperConfig != nil {
 			bucketType := syncerBucket.BucketType()
 			bucketTypeEnum := cfg.GetBucketType(bucketType.Hierarchical, bucketType.Zonal)
-			optimizedFlags := serverCfg.NewConfig.ApplyOptimizations(serverCfg.UserConfig, &cfg.OptimizationInput{
+			optimizedFlags := serverCfg.NewConfig.ApplyOptimizations(serverCfg.ViperConfig, &cfg.OptimizationInput{
 				BucketType: bucketTypeEnum,
 			})
 			if len(optimizedFlags) > 0 {
 				logger.Info("GCSFuse Config", "Applied optimizations for bucket-type: ", bucketTypeEnum, "Full Config", optimizedFlags)
 				optimizedFlagNames := slices.Collect(maps.Keys(optimizedFlags))
-				if err := cfg.Rationalize(serverCfg.UserConfig, serverCfg.NewConfig, optimizedFlagNames); err != nil {
+				if err := cfg.Rationalize(serverCfg.ViperConfig, serverCfg.NewConfig, optimizedFlagNames); err != nil {
 					logger.Warnf("GCSFuse Config: error in rationalize after applying bucket-type optimizations: %v", err)
 				}
 			}


cfg/config_test.go
```diff
@@ -38,8 +38,8 @@ func TestApplyOptimizations(t *testing.T) {
 				config: Config{},
 				isSet: &mockIsValueSet{
 					setFlags: map[string]bool{
```
kislaykishore: Should this be renamed to setParams to suggest that this is not really flag name.

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+
+func TestOptimization_RespectsConfigFile(t *testing.T) {
+	testCases := []struct {
+		name          string
+		configContent string
+		args          []string
+		validate      func(*testing.T, *mountInfo)
+	}{
+		{
+			name: "machine_type_optimization_respects_config_file",
```
gargnitingoogle: Can you also add a test case for profile-based optimization having priority over machine-type-based optimization.

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+			},
+		},
+		{
+			name: "profile_optimization_respects_cli_flag",
+			args: []string{"--profile=" + cfg.ProfileAIMLTraining, "--implicit-dirs=false"},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.False(t, mi.config.ImplicitDirs, "Should respect CLI value false, not optimize to true")
+				assert.True(t, mi.isUserSet.IsSet("implicit-dirs"), "isUserSet should be true for implicit-dirs")
+			},
+		},
```
gargnitingoogle, line 2681: Can you add the test case `bucket_optimization_flag_is_detected_as_set` in this test as well ?

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
 		})
 	}
 }
+
+func TestOptimization_RespectsConfigFile(t *testing.T) {
```
gargnitingoogle:
```suggestion
func TestArgParsing_ConfigFileOverridesFlagOptimizations(t *testing.T) {
```

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+			err = cmd.Execute()
+
+			require.NoError(t, err)
+			require.NotNil(t, capturedMountInfo)
+			tc.validate(t, capturedMountInfo)
+		})
+	}
+}
+
+func TestOptimization_RespectsCliFlags(t *testing.T) {
```
gargnitingoogle:
```suggestion
TestArgParsing_CliFlagsOverridesFlagOptimizations(t *testing.T) {
```

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+			configContent: `implicit-dirs: false
+`,
+			args: []string{"--profile=" + cfg.ProfileAIMLTraining},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.False(t, mi.config.ImplicitDirs, "Should respect config file value false, not optimize to true")
+				assert.True(t, mi.isUserSet.IsSet("implicit-dirs"), "isUserSet should be true for implicit-dirs")
+			},
+		},
+		{
+			name: "bucket_optimization_flag_is_detected_as_set",
```
gargnitingoogle:
```suggestion
			name: "bucket_type_optimization_respects_config_file",
```

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+	}{
+		{
+			name: "machine_type_optimization_respects_config_file",
+			configContent: `write:
+  global-max-blocks: 123
+`,
+			args: []string{"--machine-type=a3-highgpu-8g"},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.Equal(t, int64(123), mi.config.Write.GlobalMaxBlocks, "Should respect config file value 123, not optimize to 1600")
+				assert.True(t, mi.isUserSet.IsSet("write.global-max-blocks"), "isUserSet should be true for write.global-max-blocks")
```
gargnitingoogle:
```suggestion
				assert.True(t, mi.isUserSet.IsSet("write.global-max-blocks"), "isUserSet should be true for write.global-max-blocks")
				assert.Contains(t, mi.optimizedFlags, "implicit-dirs", "Should not optimize implicit-dirs")
				assert.NotContains(t, mi.optimizedFlags, "write.global-max-blocks", "Should not optimize implicit-dirs")
```

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+		},
+		{
+			name: "bucket_optimization_flag_is_detected_as_set",
+			configContent: `file-system:
+  enable-kernel-reader: false
+`,
+			args: []string{},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.False(t, mi.config.FileSystem.EnableKernelReader)
+				assert.True(t, mi.isUserSet.IsSet("file-system.enable-kernel-reader"), "isUserSet should be true for file-system.enable-kernel-reader")
```
gargnitingoogle:
```suggestion
				assert.True(t, mi.isUserSet.IsSet("file-system.enable-kernel-reader"), "isUserSet should be true for file-system.enable-kernel-reader")
				assert.NotContains(t, mi.optimizedFlags, "implicit-dirs", "Should not optimize implicit-dirs")
```

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+		name     string
+		args     []string
+		validate func(*testing.T, *mountInfo)
+	}{
+		{
+			name: "machine_type_optimization_respects_cli_flag",
+			args: []string{"--machine-type=a3-highgpu-8g", "--write-global-max-blocks=123"},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.Equal(t, int64(123), mi.config.Write.GlobalMaxBlocks, "Should respect CLI value 123, not optimize to 1600")
+				assert.True(t, mi.isUserSet.IsSet("write.global-max-blocks"), "isUserSet should be true for write.global-max-blocks")
```
gargnitingoogle:
```suggestion
				assert.True(t, mi.isUserSet.IsSet("write.global-max-blocks"), "isUserSet should be true for write.global-max-blocks")
				assert.Contains(t, mi.optimizedFlags, "implicit-dirs", "Should optimize implicit-dirs")
				assert.NotContains(t, mi.optimizedFlags, "write.global-max-blocks", "Should not optimize write.global-max-blocks")
```

cmd/root_test.go
```diff
@@ -2546,3 +2546,111 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+			},
+		},
+		{
+			name: "profile_optimization_respects_cli_flag",
+			args: []string{"--profile=" + cfg.ProfileAIMLTraining, "--implicit-dirs=false"},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.False(t, mi.config.ImplicitDirs, "Should respect CLI value false, not optimize to true")
+				assert.True(t, mi.isUserSet.IsSet("implicit-dirs"), "isUserSet should be true for implicit-dirs")
```
gargnitingoogle:
```suggestion
				assert.True(t, mi.isUserSet.IsSet("implicit-dirs"), "isUserSet should be true for implicit-dirs")
				assert.NotContains(t, mi.optimizedFlags, "implicit-dirs", "Should not optimize implicit-dirs")
```

cmd/root_test.go
```diff
@@ -2631,6 +2644,18 @@ func TestOptimization_RespectsCliFlags(t *testing.T) {
 			validate: func(t *testing.T, mi *mountInfo) {
 				assert.False(t, mi.config.ImplicitDirs, "Should respect CLI value false, not optimize to true")
 				assert.True(t, mi.isUserSet.IsSet("implicit-dirs"), "isUserSet should be true for implicit-dirs")
+				assert.NotContains(t, mi.optimizedFlags, "implicit-dirs", "Should not optimize implicit-dirs")
+			},
+		},
+		{
+			name: "bucket_optimization_flag_is_detected_as_set",
```
gargnitingoogle:
```
         name: "bucket_type_optimization_respects_cli_flags",
```

cmd/root.go
```diff
@@ -52,36 +52,6 @@ type mountInfo struct {
 
 type mountFn func(mountInfo *mountInfo, bucketName, mountPoint string) error
 
-// pflagAsIsValueSet is an adapter that makes a pflag.FlagSet satisfy the
-// cfg.isValueSet interface, allowing us to check for user-set flags reliably.
-type pflagAsIsValueSet struct {
```
gargnitingoogle: As discussed offline, please remove the interface IsValueSet, it's not needed anymore I think.

cfg/optimize.go
```diff
@@ -104,12 +91,12 @@ func getMetadata(client *http.Client, endpoint string) ([]byte, error) {
 	return body, nil
 }
 
-// getMachineType fetches the machine type from the metadata server if not set in isSet, cfg.
-func getMachineType(isSet IsValueSet, cfg *Config) (string, error) {
+// getMachineType fetches the machine type from the metadata server if not set in userConfig, cfg.
+func getMachineType(userConfig *viper.Viper, cfg *Config) (string, error) {
```
gargnitingoogle: @vipnydav as discussed, passing Viper around makes it less intuitive and harder to understand what this parameter is for. I take back my last comment. Let's put back the IsValuSet interface and use that here and other places.

vipnydav: Done

cfg/config.go
```diff
@@ -229,23 +229,23 @@ var machineTypeToGroupMap = map[string]string{
 // ApplyOptimizations modifies the config in-place with optimized values.
 // input parameter is optional and provides runtime context for optimizations
 // such as bucket type. Pass nil if not available.
-func (c *Config) ApplyOptimizations(isSet IsValueSet, input *OptimizationInput) map[string]OptimizationResult {
```
gargnitingoogle, line 232: The change of the variable name from `isSet` to `userConfig` everywhere is both unnecessary and bad for readability, given that it's for type IsValueSet. Please revert this change as well.

cfg/optimize.go
```diff
@@ -49,17 +48,11 @@ type OptimizationResult struct {
 	Optimized bool `yaml:"-" json:"-"` // Field hidden from YAML and JSON to avoid it in logs.
 }
 
-// IsValueSet interface allows checking if a flag was explicitly set by the user.
-// This is used to determine whether to apply optimization rules or respect user choices.
+// IsValueSet provides access to user-provided configuration values,
```
gargnitingoogle: The new header makes it sounds like this is the same as the Config structure.

```suggestion
// IsValueSet interface allows checking if a flag was explicitly set by the user.
```

cfg/config.go
gargnitingoogle: In fact it's also inconsistent because now at some places, it's `userSetFlags` and others it is `userConfig`) .

gargnitingoogle:
@vipnydav please resolve conflicts in the PR.

cfg/rationalize.go
```diff
@@ -24,13 +24,6 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
 )
 
-// isSet interface is abstraction over the IsSet() method of viper, specially
-// added to keep rationalize method simple. IsSet will be used to resolve
-// conflicting deprecated flags and new configs.
-type isSet interface {
```
gargnitingoogle, line 30: I don't understand the need for removal of this structure. This change is causing a lot of unnecessary changes across a lot of files, especially so many tests. This interface is a subset of IsValueSet, so viper anyway can be passed as it is for it. At the most, you need to rename it from `isSet` to `IsSet` if needed.

vipnydav: I am currently reusing the `IsValueSet` interface defined in `optimize.go`, which is why I removed the local `isSet` interface in `rationalize.go` to avoid redundancy within the same package.

However, as discussed with @ashmeenkaur , while using interfaces is generally good practice, the goal here is to strictly use viper for configuration management. Keeping the `IsValueSet` interface (or a subset of it) leaves the door open for someone to pass in other implementations, like `pflag`, in the future. We want to avoid this to ensure we are always relying on viper's specific behavior for resolving configuration values.

If everyone is aligned on this, I can remove all the interfaces entirely and use `*viper.Viper` directly throughout the codebase. This would make the strict dependency explicit, even if it may reduce readability a bit.

CC : @kislaykishore

vipnydav: As discussed offline I have removed the `IsValueSet` interface.

cfg/config.go
vipnydav: Removed `IsValueSet` interface as discussed.

cfg/rationalize.go
gargnitingoogle: ok, discussed this offline with the others.

gargnitingoogle:
@vipnydav this PR looks obsolete now given the merged PR #4271 . If so, please close it.

vipnydav:
Closing this PR as this PR was too long to review. Hence created separate PRs for:
i) Using viper instead of pFlag : https://github.com/GoogleCloudPlatform/gcsfuse/pull/4271
ii) Removing IsSet interface : https://github.com/GoogleCloudPlatform/gcsfuse/pull/4291

PR #4265, opened 2026-01-12
feat(kernel-settings): Add a hidden flag to provide the kernel params file from CSI Driver.

meet2mky:
### Description
Adding a hidden flag to pass the kernel-params file from CSI Driver where GCSFuse would write the kernel parameters for CSI Driver in GKE environment.

b/475110502

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA

### DIFF
--- a/cfg/validate_test.go
+++ b/cfg/validate_test.go
@@ -322,27 +322,6 @@ func TestValidateConfigSuccessful(t *testing.T) {
 				},
 			},
 		},
-		{
-			name: "valid_kernel_params_file",
-			config: &Config{
-				Logging:   LoggingConfig{LogRotate: validLogRotateConfig()},
-				FileCache: validFileCacheConfig(t),
-				GcsConnection: GcsConnectionConfig{
-					SequentialReadSizeMb: 10,
-				},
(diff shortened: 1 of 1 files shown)


cmd/config_validation_test.go
```diff
@@ -1002,3 +1002,69 @@ func TestValidateConfigFile_MachineTypeConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateKernelParamsFileFlag(t *testing.T) {
```
raj-prince: I think, we should not add the test case if there is no validation logic. WDUT?

cfg/rationalize_test.go
```diff
@@ -753,3 +753,17 @@ func TestResolveLoggingConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestRationalize_KernelParamsFileUnchanged(t *testing.T) {
```
raj-prince: Here, also. We shouldn't increase the maintenance burden, if there is no rationalization code associated with the flag.

raj-prince (commented):
LGTM.

PR #4271, opened 2026-01-13
fix: Ensure optimizations honor user-defined values in configuration files

vipnydav:
### Description
Previously, GCSFuse relied on a custom `pflag` adapter and the `pflag.Changed()` method to determine if a configuration flag had been explicitly set by the user before applying automatic optimizations. This approach was flawed because it only detected flags provided via the command-line interface and failed to account for values defined in configuration files. Consequently, user-defined settings in a config file were being unintentionally ignored and overwritten by automatic optimization logic if the corresponding CLI flag was not also provided.

To resolve this issue, the optimization logic has been updated to leverage the existing `viper.IsSet()` functionality, which correctly identifies if a value was provided through either a CLI flag or the configuration file.

**Note** : Adding cleanup for `IsUserSet` Interface was making changes too big for review so would be making those changes in a separate PR.

**Note**: `BucketType` is not checked at `root` layer but at `fs` layer so it's tests can't be added in `root` layer.

https://b.corp.google.com/issues/474942076

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -2617,7 +2617,7 @@ func TestArgParsing_ConfigFileOverridesFlagOptimizations(t *testing.T) {
 			},
 		},
 		{
-			name: "machine_type_in_config_file_is_overridden_by_another_config_flag",
+			name: "machine_type_in_config_file_respects_other_config_file_configurations",
 			configContent: `
 machine-type: a3-highgpu-8g
 write:


cmd/root.go
```diff
@@ -159,9 +129,8 @@ of Cloud Storage FUSE, see https://cloud.google.com/storage/docs/gcs-fuse.`,
 				return fmt.Errorf("invalid config: %w", err)
 			}
 
-			isUserSet := &pflagAsIsValueSet{fs: cmd.PersistentFlags()}
-			mountInfo.isUserSet = isUserSet
-			optimizedFlags := mountInfo.config.ApplyOptimizations(isUserSet, nil)
+			mountInfo.isUserSet = v
```
ashmeenkaur: Nit: can we rename v to viperConfig for readability

cmd/root_test.go
```diff
@@ -2580,3 +2580,136 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+  global-max-blocks: 123
+`,
+			args: []string{"--machine-type=a3-highgpu-8g"},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.Equal(t, int64(123), mi.config.Write.GlobalMaxBlocks, "Should respect config file value 123, not optimize to 1600")
+				assert.True(t, mi.isUserSet.IsSet("write.global-max-blocks"), "isUserSet should be true for write.global-max-blocks")
+				assert.Contains(t, mi.optimizedFlags, "implicit-dirs", "Should optimize implicit-dirs")
+				if writeFlags, ok := mi.optimizedFlags["write"].(map[string]any); ok {
+					assert.NotContains(t, writeFlags, "global-max-blocks", "Should not optimize write.global-max-blocks")
+				}
```
ashmeenkaur: Nit:

To keep the tests simple and maintainable, could we change these assertions to only check the final config values and mi.isUserSet only?

Currently, the tests check mi.optimizedFlags, which is a hierarchical map currently being used for logging purpose only. IMO in future we can initialize the log file earlier to avoid storing these flags in mountInfo entirely.

```suggestion
				assert.Equal(t, int64(123), mi.config.Write.GlobalMaxBlocks) // config takes precendence
				assert.True(t, mi.isUserSet.IsSet("write.global-max-blocks"))
				assert.True(t, mi.config.ImplicitDirs) // default for high perf machine
				assert.False(t, mi.isUserSet.IsSet("implicit-dirs"))
```

cmd/root_test.go
```diff
@@ -2580,3 +2580,136 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+		configContent string
+		args          []string
+		validate      func(*testing.T, *mountInfo)
+	}{
+		{
+			name: "machine_type_optimization_respects_config_file",
+			configContent: `write:
+  global-max-blocks: 123
+`,
+			args: []string{"--machine-type=a3-highgpu-8g"},
```
ashmeenkaur, line 2599: Can we add tests where machine type/profile are set via config file if that is possible?

cmd/root_test.go
```diff
@@ -2580,3 +2580,136 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+					assert.NotContains(t, fsFlags, "enable-kernel-reader", "Should not optimize file-system.enable-kernel-reader")
+				}
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			configFile := createTempConfigFile(t, tc.configContent)
+			defer os.Remove(configFile)
+
```
ashmeenkaur: AAA format
Nit:
```suggestion
```

cmd/root_test.go
```diff
@@ -2580,3 +2583,139 @@ func TestArgsParsing_DummyIOConfigFile(t *testing.T) {
...
+  enable-kernel-reader: false
+`,
+			args: []string{},
+			validate: func(t *testing.T, mi *mountInfo) {
+				assert.False(t, mi.config.FileSystem.EnableKernelReader)
+				assert.True(t, mi.isUserSet.IsSet("file-system.enable-kernel-reader"), "isUserSet should be true for file-system.enable-kernel-reader")
+			},
+		},
+		{
+			name: "machine_type_in_config_file_is_overridden_by_another_config_flag",
```
ashmeenkaur:
```suggestion
			name: "machine_type_in_config_file_respects_other_config_file_configurations",
```

ashmeenkaur (approved):

PR #4278, opened 2026-01-16
fix: Add stall timeout for direct path connectivity

Tulsishah:
### Description
Addressed an issue where Direct Path connectivity calls would occasionally hang indefinitely, leading to a poor user experience. A 60-second context timeout has been implemented to cancel these stuck requests, allowing the mount operation to successfully fail over to the cloud path.

[b/475733879](https://buganizer.corp.google.com/issues/475733879)

### Testing details
1. Manual - tested with small ttl(60ns) and verified that it is giving context deadline exceeded error.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -52,8 +52,8 @@ const (
 	dynamicReadReqIncreaseRateEnv   = "DYNAMIC_READ_REQ_INCREASE_RATE"
 	dynamicReadReqInitialTimeoutEnv = "DYNAMIC_READ_REQ_INITIAL_TIMEOUT"
 
-	zonalLocationType             = "zone"
-stallTimeoutForDirectPath = 60 * time.Second
+	zonalLocationType         = "zone"
+	stallTimeoutForDirectPath = 60 * time.Second
 )
 
 type StorageHandle interface {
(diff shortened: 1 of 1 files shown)


mustvicky:
Can we repro the original situation as part of this fix to practically make sure that the timeout is taking effect? I am just concerned that the Go SDK might not have hooked the context properly

internal/storage/storage_handle.go
```diff
@@ -86,7 +87,11 @@ type gRPCDirectPathDetector struct {
 // from the environment where the client is running. A `nil` error represents Direct Connectivity was
 // detected.
 func (pd *gRPCDirectPathDetector) isDirectPathPossible(ctx context.Context, bucketName string) error {
-	return storage.CheckDirectConnectivitySupported(ctx, bucketName, pd.clientOptions...)
+	newCtx, cancel := context.WithTimeout(ctx, stallTimeoutForDirectPath)
```
raj-prince, line 90: LGTM from GCSFuse side.

I see, go-storage is not adhering the ctx while collecting metric inside the tool. Have we tested this, as I am not sure issue was because of grpc client creation or metrics collection?

We can create a bug for go-sdk to adhere the context - https://github.com/googleapis/google-cloud-go/blob/736b75806a04eee8f96f8e59665c2f7685740e12/storage/storage.go#L297

Tulsishah: I have already tested with small ttl like 60 nano secs and get context deadline exceeded error. So I assume they are handling it context deadline.

Tulsishah:
> Can we repro the original situation as part of this fix to practically make sure that the timeout is taking effect? I am just concerned that the Go SDK might not have hooked the context properly

I verified this by testing with a short TTL (60ns) and received the expected 'context deadline exceeded' error. The context is passed during [gRPC client creation](https://github.com/googleapis/google-cloud-go/blob/736b75806a04eee8f96f8e59665c2f7685740e12/storage/storage.go#L287) and is being handled correctly.

PR #4283, opened 2026-01-16
feat(simple_reader): changing the default value for max-background

raj-prince:
### Description
- Current default value of max-background was leading to high gcsfuse cpu usage and a way more than required to saturate the 200Gbps read throughput.
- Capping the default value based on the max supported nic value, kept the value slightly more than to saturate the network. With this we are kind of getting almost same performance with less gcsfuse cpu utilization.
- CSV file: https://github.com/GoogleCloudPlatform/gcsfuse-tools/blob/d_princer_micro_benchmarking/distributed-micro-benchmarking/final_reports/tune_kernel_settings/final_comparison.csv
-  Comparison Graph (the blue one is the proposed one):
image

b/475124487

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/config_util.go
+++ b/cfg/config_util.go
@@ -23,9 +23,9 @@ import (
 
 const (
 	// maxBackgroundLimit configures the upper limit for max-background kernel fuse
-	// setting. This is more than sufficient to saturated the 200 Gbps network bandwidth
+	// setting. This is more than sufficient to saturate the 200 Gbps network bandwidth
 	// on a single VM. Revise the numbers if you plan to support higher bandwidth VMs.
-	maxBackgroundLimit = 256
+	maxBackgroundLimit = 192
 )
 
 func DefaultMaxBackground() int {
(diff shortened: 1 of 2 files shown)


cfg/config_util.go
```diff
@@ -21,8 +21,15 @@ import (
...
+const (
+	// maxBackgroundLimit configures the upper limit for max-background kernel fuse
+	// setting. This is more than sufficient to saturate the 200 Gbps network bandwidth
+	// on a single VM. Revise the numbers if you plan to support higher bandwidth VMs.
+	maxBackgroundLimit = 192
+)
+
 func DefaultMaxBackground() int {
-	return max(12, 4*runtime.NumCPU())
+	return min(max(12, 2*runtime.NumCPU()), maxBackgroundLimit)
```
abhishek10004, line 32: Did we do any testing for limiting to 2*numCPUs for smaller machines?

raj-prince: Haven't performed any testing, just went ahead with reasonable allowed parallelism we used in the past, for example, we are using 2 * cpu for file-cache parallel download. If required, we can revisit on this.

PR #4290, opened 2026-01-19
feat(Type Cache Deprecation): LookUp Child changes

Tulsishah:
### Description
This PR updates the LookUpChild logic to support the deprecation of the typeCache by merging its functionality into a single, unified statCache. 

Key Updates:

1. Unified Cache Logic: Modifies LookUpChild to determine object types directly from metadata in the statCache
2. Feature Flag: Introduces the EnableTypeCacheDeprecation configuration flag to safely toggle this new behavior 
3. Cache-Only Phase: Implements a two-phase lookup that checks the local cache first before falling back to the GCS backend, optimizing performance.

[b/476918868](http://b/476918868)

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - Automated, also tested with type cache deprecation enabled

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -352,7 +352,7 @@ func (d *dirInode) lookUpChildDir(ctx context.Context, name string) (*Core, erro
 	}
 
 	if d.implicitDirs {
-		return findDirInode(ctx, d.Bucket(), childName, d.isEnableTypeCacheDeprecation)
+		return findDirInode(ctx, d.Bucket(), childName)
 	}
 	return findExplicitInode(ctx, d.Bucket(), childName, false)
 }
@@ -450,15 +450,14 @@ func findExplicitFolder(ctx context.Context, bucket *gcsx.SyncerBucket, name Nam
 
 // findDirInode finds the dir inode core where the directory is either explicit
 // or implicit. Returns nil if no such directory exists.
-func findDirInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name, isTypeCacheDeprecated bool) (*Core, error) {
+func findDirInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name) (*Core, error) {
 	if !name.IsDir() {
 		return nil, fmt.Errorf("%q is not directory", name)
 	}
 
 	req := &gcs.ListObjectsRequest{
-		Prefix:                name.GcsObjectName(),
-		MaxResults:            1,
-		IsTypeCacheDeprecated: isTypeCacheDeprecated,
+		Prefix:     name.GcsObjectName(),
+		MaxResults: 1,
 	}
 	listing, err := bucket.ListObjects(ctx, req)
 	if err != nil {
@@ -604,26 +603,31 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error) 
 	if d.IsTypeCacheDeprecated() && d.metadataCacheTtlSecs != 0 {
 		var cacheMissErr *caching.CacheMissError
 
-		// Try File
-		fileResult, err := findExplicitInode(ctx, d.Bucket(), NewFileName(d.Name(), name), true)
-		if err != nil && !errors.As(err, &cacheMissErr) {
-			return nil, err
-		}
-
-		// Try Directory
+		// 1. Try Directory FIRST (since it's the preferred return type)
 		var dirResult *Core
+		var err error
 		if d.Bucket().BucketType().Hierarchical {
 			dirResult, err = findExplicitFolder(ctx, d.Bucket(), NewDirName(d.Name(), name), true)
 		} else {
 			dirResult, err = findExplicitInode(ctx, d.Bucket(), NewDirName(d.Name(), name), true)
 		}
+
+		// If we found a directory, we're done. Return it now.
+		if dirResult != nil {
+			return dirResult, nil
+		}
+		// If we hit a real error (not a cache miss), exit early.
 		if err != nil && !errors.As(err, &cacheMissErr) {
 			return nil, err
 		}
 
-		if dirResult != nil {
-			return dirResult, nil
-		} else if fileResult != nil {
+		// 2. Try File ONLY if directory wasn't found
+		fileResult, err := findExplicitInode(ctx, d.Bucket(), NewFileName(d.Name(), name), true)
+		if err != nil && !errors.As(err, &cacheMissErr) {
+			return nil, err
+		}
+
+		if fileResult != nil {
 			return fileResult, nil
 		}
 	}
@@ -669,7 +673,7 @@ func (d *dirInode) fetchCoreEntity(ctx context.Context, name string, cachedType 
 		return
 	}
 	lookUpImplicitOrExplicitDir := func() (err error) {
-		dirResult, err = findDirInode(ctx, d.Bucket(), NewDirName(d.Name(), name), d.isEnableTypeCacheDeprecation)
+		dirResult, err = findDirInode(ctx, d.Bucket(), NewDirName(d.Name(), name))
 		return
 	}
 	lookUpHNSDir := func() (err error) {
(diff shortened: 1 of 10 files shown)


internal/fs/inode/dir.go
```diff
@@ -401,17 +406,24 @@ func findExplicitInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name
 		return nil, fmt.Errorf("StatObject: %w", err)
 	}
 
+	// This should be treated as an implicit directory, as implicit objects do not possess metadata.
+	// We are setting MinObject to nil because, for implicit directories, only the object name is saved.
+	if fetchOnlyFromCache && m.Generation == 0 {
+		m = nil
+	}
```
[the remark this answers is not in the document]
Tulsishah: > The findExplicitInode function returns nil, nil on gcs.NotFoundError. However, the LookUpChild method (lines 624-635) expects this function to return a caching.CacheMissError when fetchOnlyFromCache is true and an entry is not found in the cache.

Nofound error means cache hit but as negative cache so it does not expect `caching.CacheMissError` error. `caching.CacheMissError` means positive/negative both entries are not find in cache. So it is working as intended.

internal/fs/inode/dir.go
```diff
@@ -401,17 +406,24 @@ func findExplicitInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name
...
 }
 
-func findExplicitFolder(ctx context.Context, bucket *gcsx.SyncerBucket, name Name) (*Core, error) {
+func findExplicitFolder(ctx context.Context, bucket *gcsx.SyncerBucket, name Name, fetchOnlyFromCache bool) (*Core, error) {
 	// Call the bucket.
 	req := &gcs.GetFolderRequest{
-		Name: name.GcsObjectName(),
+		Name:               name.GcsObjectName(),
+		FetchOnlyFromCache: fetchOnlyFromCache,
 	}
```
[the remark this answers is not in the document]
Tulsishah: Nofound error means cache hit but as negative cache so it does not expect `caching.CacheMissError` error. `caching.CacheMissError` means positive/negative both entries are not find in cache. So it is working as intended.

internal/fs/inode/dir.go
```diff
@@ -336,19 +340,19 @@ func (d *dirInode) checkInvariants() {
...
-		return findExplicitFolder(ctx, d.Bucket(), childName)
+		return findExplicitFolder(ctx, d.Bucket(), childName, false)
 	}
 
 	if d.implicitDirs {
-		return findDirInode(ctx, d.Bucket(), childName)
+		return findDirInode(ctx, d.Bucket(), childName, d.isEnableTypeCacheDeprecation)
 	}
-	return findExplicitInode(ctx, d.Bucket(), childName)
+	return findExplicitInode(ctx, d.Bucket(), childName, false)
```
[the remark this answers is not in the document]
Tulsishah: It is expected as we are only changing LookUpChild behaviour with typecacheDeprecation flag on.  Other functions should work as currently it is working.

internal/fs/inode/dir.go
```diff
@@ -403,17 +408,24 @@ func findExplicitInode(ctx context.Context, bucket *gcsx.SyncerBucket, name Name
 		return nil, fmt.Errorf("StatObject: %w", err)
 	}
 
+	// This should be treated as an implicit directory, as implicit objects do not possess metadata.
+	// We are setting MinObject to nil because, for implicit directories, only the object name is saved.
+	if fetchOnlyFromCache && m.Generation == 0 {
+		m = nil
```
vadlakondaswetha, line 415: what is the reason for explicitly setting it to nil?

internal/fs/inode/dir.go
```diff
@@ -437,14 +449,15 @@ func findExplicitFolder(ctx context.Context, bucket *gcsx.SyncerBucket, name Nam
...
 	if !name.IsDir() {
 		return nil, fmt.Errorf("%q is not directory", name)
 	}
 
 	req := &gcs.ListObjectsRequest{
-		Prefix:     name.GcsObjectName(),
-		MaxResults: 1,
+		Prefix:                name.GcsObjectName(),
+		MaxResults:            1,
+		IsTypeCacheDeprecated: isTypeCacheDeprecated,
```
vadlakondaswetha: why is typeCacheDeprecated flag being passed as part of listObjects request>?

internal/fs/inode/dir.go
```diff
@@ -587,49 +600,54 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
-			MinObject: nil,
-		}
-	case metadata.ExplicitDirType:
-		if d.isBucketHierarchical() {
-			group.Go(lookUpHNSDir)
-		} else {
-			group.Go(lookUpExplicitDir)
+		// If metadata caching is enabled, attempt to retrieve the entry from the cache first.
+		// If a valid entry (file or directory) is found, return it immediately to avoid a GCS API call.
+		if d.metadataCacheTtlSecs != 0 {
```
vadlakondaswetha: why do we want to do this extra check? otherwise fast_stat_bucket doesnt get invoked?

internal/fs/inode/dir.go
```diff
@@ -587,49 +600,54 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
 	}
 	lookUpHNSDir := func() (err error) {
-		dirResult, err = findExplicitFolder(ctx, d.Bucket(), NewDirName(d.Name(), name))
+		dirResult, err = findExplicitFolder(ctx, d.Bucket(), NewDirName(d.Name(), name), false)
 		return
 	}
+
 	var cachedType metadata.Type
+	var cacheMissErr *caching.CacheMissError
 	if d.IsTypeCacheDeprecated() {
```
vadlakondaswetha: can we organize the code in a better way.

cachedType = metadata.UnknownType
if d.IsTypeCacheDeprecated && d.metadaCacheTtlSecs > 0 {
   // Get the entity
   // if it is not nil, return
}

if !d.IsTypeCacheDeprecated {
 cachedType = fetchCachedType()
}

getEntity(cachedType) -> all the existing logic of lookUpChild can be here without any isTypeCacheDeprecated flag.

internal/fs/inode/dir.go
```diff
@@ -681,10 +733,11 @@ func (d *dirInode) ReadDescendants(ctx context.Context, limit int) (map[Name]*Co
...
 		listing, err := d.bucket.ListObjects(ctx, &gcs.ListObjectsRequest{
-			Delimiter:         "", // recursively
-			Prefix:            d.Name().GcsObjectName(),
-			ContinuationToken: tok,
-			MaxResults:        limit + 1, // to exclude itself
+			Delimiter:             "", // recursively
+			Prefix:                d.Name().GcsObjectName(),
+			ContinuationToken:     tok,
+			MaxResults:            limit + 1, // to exclude itself
+			IsTypeCacheDeprecated: d.isEnableTypeCacheDeprecation,
```
vadlakondaswetha: same here. we should not pass flag values as part of request object for every call.

internal/fs/inode/dir.go
Tulsishah: We are instantiating inodes based on their specific Core Type. For the ImplicitDir Core Type, the MinObject field must be set to nil.
ref: https://github.com/GoogleCloudPlatform/gcsfuse/blob/44e7ca317e9dc39ffb1afd444ae3aca49e731e2a/internal/fs/inode/core.go#L54

internal/fs/inode/dir.go
Tulsishah: It is to enable new caching logic in fast stat bucket
ref:
https://github.com/GoogleCloudPlatform/gcsfuse/blob/44e7ca317e9dc39ffb1afd444ae3aca49e731e2a/internal/storage/caching/fast_stat_bucket.go#L439

internal/fs/inode/dir.go
Tulsishah: If we do this then with list call caching will happen correctly with type cache deprecation otherwise implicit directories will not get cached correctly.

internal/fs/inode/dir.go
Tulsishah: If we don't put think check then,

with metadata cache ttl 0 also we will make sequential call to GCS. 

Also updated in the comment.

internal/fs/inode/dir.go
Tulsishah: Also added comment.

internal/fs/inode/dir.go
Tulsishah: Thought to have less diff as we are anyway going to deprecate the old logic. But NVM I have organize the code.

internal/fs/inode/dir.go
vadlakondaswetha: This is a mount level config and should be initialized as part of fast_stat_bucket instance instead of passing it as part of ListObjectrEquest proto.

internal/fs/inode/dir.go
```diff
@@ -583,40 +597,93 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
+		} else {
+			dirResult, err = findExplicitInode(ctx, d.Bucket(), NewDirName(d.Name(), name), true)
+		}
+		if err != nil && !errors.As(err, &cacheMissErr) {
+			return nil, err
+		}
+
+		if dirResult != nil {
+			return dirResult, nil
+		} else if fileResult != nil {
```
vadlakondaswetha: if dirREsult is given preference why not make dir call first, if result is not nil just return. Why are we making 2 calls always?

PR #4291, opened 2026-01-19
refactor: remove redundant IsValueSet interface

vipnydav:
### Description

This PR removes the custom `IsValueSet` interface and its various adapters (such as `pflagAsIsValueSet`), replacing them with direct usage of `*viper.Viper` across the configuration and optimization logic.

**Background**

In the previous PR #4271, we implemented a fix to ensure that optimizations honor user-defined values in configuration files. That implementation still had the `IsValueSet` interface to abstract the check for whether a configuration key was explicitly set by the user as removing it were making the changes bigger to review.

Now, we want to make sure that in future the developers only use `*viper.Viper` and don't end up using any other data structure which satisfies `IsSet` interface as it may lead to same behavior where CLI flags or Config flags are overriden. Hence we directly pass `*viper.Viper` simplifying the code and removing the need for boilerplate adapter structures.

https://b.corp.google.com/issues/476900087

### Testing details
1. Manual - Done
2. Unit tests - Updated

### DIFF
--- a/cfg/config_test.go
+++ b/cfg/config_test.go
@@ -80,6 +80,7 @@ func TestApplyOptimizations(t *testing.T) {
 				for key, val := range tc.userSetFlags {
 					v.Set(key, val)
 				}
+
 				optimizedFlags := c.ApplyOptimizations(v, tc.input)
 
 				if tc.expectOptimized {
@@ -148,6 +149,7 @@ func TestApplyOptimizations(t *testing.T) {
 				for key, val := range tc.userSetFlags {
 					v.Set(key, val)
 				}
+
 				optimizedFlags := c.ApplyOptimizations(v, tc.input)
 
 				if tc.expectOptimized {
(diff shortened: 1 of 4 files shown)


cfg/config_test.go
```diff
@@ -28,19 +29,17 @@ func TestApplyOptimizations(t *testing.T) {
...
 				name:   "user_set",
 				config: Config{},
-				isSet: &mockIsValueSet{
-					setFlags: map[string]bool{
-						"file-system.congestion-threshold": true,
-						"machine-type":                     true,
-					},
+				userSetFlags: map[string]any{
+					"file-system.congestion-threshold": true,
+					"machine-type":                     true,
```
ashmeenkaur: Nit: flags are not boolean by definition

cfg/optimize_test.go
```diff
@@ -139,34 +103,26 @@ func TestGetMachineType_FlagIsSet(t *testing.T) {
...
-				setFlags:    map[string]bool{"machine-type": true},
-				stringFlags: map[string]string{"machine-type": "cli-machine-type"},
-			},
-			config: &Config{
-				MachineType: "config-file-machine-type",
+			userSetFlags: map[string]string{
+				"machine-type": "cli-machine-type",
 			},
 			expectedMachineType: "cli-machine-type",
 		},
```
ashmeenkaur: the 2 tests are same

PR #4300, opened 2026-01-20
fix(test): prevent panic in read cache validation when chunks are empty

vipnydav:
### Description
The validation logic in `tools/integration_tests/read_cache/helpers_test.go` was attempting to access the last element of the `logEntry.Chunks` slice using `logEntry.Chunks[len(logEntry.Chunks)-1]` without first verifying if the slice contained any elements. This caused a panic when the slice was empty.

This change adds a guard clause to return early if no chunks are found in the logs, preventing the index out-of-bounds error and allowing the test to fail gracefully or proceed correctly.

b/477184829

### Testing details
1. Manual - Done
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/read_cache/helpers_test.go
+++ b/tools/integration_tests/read_cache/helpers_test.go
@@ -92,6 +92,10 @@ func validate(expected *Expected, logEntry *read_logs.StructuredReadLogEntry, is
 		t.Errorf("chunks read don't match! Expected: %d, Got from logs: %d",
 			chunkCount, len(logEntry.Chunks))
 	}
+	// If no chunks are found in logs, we can't validate further.
+	if len(logEntry.Chunks) == 0 {
+		return
+	}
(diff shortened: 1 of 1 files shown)


tools/integration_tests/read_cache/helpers_test.go
```diff
@@ -92,6 +92,10 @@ func validate(expected *Expected, logEntry *read_logs.StructuredReadLogEntry, is
 		t.Errorf("chunks read don't match! Expected: %d, Got from logs: %d",
 			chunkCount, len(logEntry.Chunks))
 	}
+	// If no chunks are found in logs, we can't validate further.
```
abhishek10004, line 95: Is bailing out early without logging an error the correct decision here? It seems like we expected the logs for file cache here and validation should fail if that's not the case.

vipnydav: We are already logging it in Line#91. And if `len(logEntry.Chunks) !=0` we can perform other checks as well. That's why I didn't return when we were getting unexpected number of chunks.

PR #4303, opened 2026-01-20
test(tests): Enable kernel reader for e2e tests and disable where needed

vipnydav:
### Description
This PR enables the kernel reader by default for integration tests. Previously, the kernel reader was disabled by default for all the integration tests. This change removes that default and instead explicitly disables the kernel reader only in specific test suites that require it to be OFF.  The kernel reader is now explicitly disabled only in specific suites where it must be off to verify other reader paths, such as `buffered_read` and `read_cache`. Additionally, several packages have been updated to test both enabled and disabled states to maintain coverage of the previous flow for zonal buckets.

b/474484680

### Testing details
1. Manual - Done
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/monitoring/setup_test.go
+++ b/tools/integration_tests/monitoring/setup_test.go
@@ -177,24 +177,24 @@ func TestMain(m *testing.M) {
 		testEnv.cfg.GKEMountedDirectory = setup.MountedDirectory()
 
 		testEnv.cfg.Configs = make([]test_suite.ConfigItem, 6)
-		testEnv.cfg.Configs[0].Flags = []string{"--prometheus-port=9190 --cache-dir=/gcsfuse-tmp/PromOTELSuite --log-file=/gcsfuse-tmp/monitoring.log"}
+		testEnv.cfg.Configs[0].Flags = []string{"--prometheus-port=9190 --cache-dir=/gcsfuse-tmp/PromOTELSuite --log-file=/gcsfuse-tmp/monitoring.log --enable-kernel-reader=false"}
 		testEnv.cfg.Configs[0].Compatible = map[string]bool{"flat": true, "hns": false, "zonal": false}
 		testEnv.cfg.Configs[0].Run = "TestPromOTELSuite"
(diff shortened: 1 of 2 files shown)


tools/integration_tests/rapid_appends/setup_test.go
abhishek10004, line 69: Let's run these 2 without kernel reader as well

tools/integration_tests/rapid_appends/setup_test.go
```diff
@@ -81,8 +81,12 @@ func TestMain(m *testing.M) {
 		cfg.RapidAppends[0].Configs[2].Flags = []string{
 			"--metadata-cache-ttl-secs=0",  // NoCache
 			"--metadata-cache-ttl-secs=70", // MetadataCache
-			"--file-cache-max-size-mb=-1 --cache-dir=/gcsfuse-tmp/cache --metadata-cache-ttl-secs=0",  // FileCache
-			"--metadata-cache-ttl-secs=70 --file-cache-max-size-mb=-1 --cache-dir=/gcsfuse-tmp/cache", // MetadataAndFileCache
+			"--file-cache-max-size-mb=-1 --cache-dir=/gcsfuse-tmp/cache --metadata-cache-ttl-secs=0",                               // FileCache
```
abhishek10004, line 84: Instead of manually duplicating the flags and appending an extra flag, would it make sense to have a helper function do that?

tools/integration_tests/rapid_appends/setup_test.go
vipnydav: There tests validate using storage client and don't read via gcsfuse mount so it doesn't make sense to run these without kernel reader as well.

tools/integration_tests/rapid_appends/setup_test.go
vipnydav: We are keeping these configurations for backward compatibility in the short term, but they will be removed in the next few releases. As we shift toward using `test_config.yaml` as our main configuration file, the extra effort required to refactor these older files doesn't provide much benefit.

PR #4304, opened 2026-01-21
docs(Unsupported Path Support): Update documentation

Tulsishah:
### Description
Update documentation for semantic and troubleshooting doc.

[b/477461017](http://b/477461017)

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -110,7 +110,7 @@ Fuse package is not installed. It may throw this error. Run the following comman
 ### Encountered unsupported prefixes during listing, or ls: reading directory \<mountpath>/\<path>:  Input/output error
 Unable to list unsupported objects in a mounted bucket, i.e. objects which have names with `//` in them or have names starting with `/` e.g. `gs://<bucket>//A` or `gs://<bucket>/A//B` etc. Such objects can be listed by the following command: `gcloud storage ls gs://<bucket>/**//**`.
 
-The `Input/output error` has been supressed from v3.6.0 onwards. In older vetsions, this error can be mitigated in one of the following two ways.
+The `Input/output error` has been supressed from v3.6.0 onwards. In older versions, this error can be mitigated in one of the following two ways.
 
 * Move/Rename such objects e.g. for objects of names like `A//B`, use `gcloud storage mv gs://<bucket>/A//* gs://<bucket>/A/`), Or
 * Delete such objects e.g. for objects of names like `A//B`, use `gcloud storage rm gs://<bucket>/A//**`.


docs/semantics.md
```diff
@@ -519,15 +519,9 @@ Traditional file systems do not allow multiple directory entries with the same n
...
 
-### Unsupported object names
-
-- Objects in GCS with `double slashes '//'` as a name or prefix are not supported in GCSfuse. Accessing a directory with such named files will cause an 'input/output error', as the Linux filesystem does not support files or directories named with a '/'. The most common example of this is an object called, for example 'A//C.txt' where 'A' indicates a directory and 'C.txt' indicates a file, and is missing directory 'B/' between 'A/' and 'C.txt'.
-
-
-- Objects in GCS with suffix `/\n` like, `gs://gcs-bkt/a/\n`:
-Mounting bucket with such objects leads to crash `sync: unlock of unlocked mutex` or `Panic: Inode 'a/' cannot have child file`.
-`\n` in GCSFuse is used to resolve the name conflicts, in case there is a file and directory exists with the same name. Ref: [name-conflict](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#name-conflicts) section.
+### Unsupported Path names
```
gargnitingoogle, line 522: let's keep it **object** names only.

```suggestion
### Unsupported Object names
```

docs/semantics.md
```diff
@@ -519,15 +519,9 @@ Traditional file systems do not allow multiple directory entries with the same n
...
 
+- GCSFuse (from v3.6.0 onwards) gracefully handles objects with names containing `//`, `/./` or `/../`, ensuring the file operations will not crash or throw errors. While listing shows only the supported objects, directory-level deletions and renames apply to all contained objects, whether supported or unsupported.
```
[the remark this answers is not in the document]
Tulsishah: > . Since docs/troubleshooting.md still documents this as an existing issue, it should probably be retained here as a known limitation under the 'Unsupported Path names' section.

Discussed with @gargnitingoogle that we always keep latest data in semantics doc.

docs/semantics.md
```diff
@@ -519,15 +519,9 @@ Traditional file systems do not allow multiple directory entries with the same n
...
-
-
-- Objects in GCS with suffix `/\n` like, `gs://gcs-bkt/a/\n`:
-Mounting bucket with such objects leads to crash `sync: unlock of unlocked mutex` or `Panic: Inode 'a/' cannot have child file`.
-`\n` in GCSFuse is used to resolve the name conflicts, in case there is a file and directory exists with the same name. Ref: [name-conflict](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#name-conflicts) section.
+### Unsupported Path names
 
+- GCSFuse (from v3.6.0 onwards) gracefully handles objects with names containing unsupported path segments (such as `//`, `/./`, `/../`), names starting with `/`, or names ending with `/.` or `/..`. This ensures that file operations will not crash or throw errors. While listing shows only the supported objects, directory-level deletions and renames apply to all contained objects, whether supported or unsupported.
```
gargnitingoogle:
```suggestion
- GCSFuse (from v3.6.0 onwards) gracefully handles objects with names containing unsupported path segments (such as `//`, `/./`, `/../`). This ensures that file operations will not crash or throw errors. While listing shows only the supported objects, directory-level deletions and renames apply to all contained objects, whether supported or unsupported.
```

docs/semantics.md
```diff
@@ -519,15 +519,9 @@ Traditional file systems do not allow multiple directory entries with the same n
 
 Instead, when a conflicting pair of foo and ```foo/``` objects both exist, it appears in the Cloud Storage FUSE file system as if there is a directory named foo and a file or symlink named ```foo\n``` (i.e. foo followed by U+000A, line feed). This is what will appear when the parent's directory entries are read, and Cloud Storage FUSE will respond to requests to look up the inode named ```foo\n``` by returning the file inode. ```\n``` in particular is chosen because it is not legal in Cloud Storage object names, and therefore is not ambiguous.
 
-### Unsupported object names
-
-- Objects in GCS with `double slashes '//'` as a name or prefix are not supported in GCSfuse. Accessing a directory with such named files will cause an 'input/output error', as the Linux filesystem does not support files or directories named with a '/'. The most common example of this is an object called, for example 'A//C.txt' where 'A' indicates a directory and 'C.txt' indicates a file, and is missing directory 'B/' between 'A/' and 'C.txt'.
-
-
-- Objects in GCS with suffix `/\n` like, `gs://gcs-bkt/a/\n`:
```
gargnitingoogle, line 527: Please confirm that `\n` isn't supported in gcs object names now before removing it. If it's still supported in GCS, then leave it as it is.

docs/semantics.md
gargnitingoogle: As discussed, let's either add the full set, which is too long, or only the most common use case.

docs/semantics.md
Tulsishah: It can be folders also in HNS, that's why I put generalize name "Path".

docs/semantics.md
Tulsishah: GCS doesn't support creating these kind of objects.
https://docs.cloud.google.com/storage/docs/objects#naming

docs/semantics.md
gargnitingoogle: > It can be folders also in HNS, that's why I put generalize name "Path".

Sure, but the term `object` applies to all kinds of gcs objects including files and folders.

Tulsishah: Still as we have mentioned unsupported-path-support in entire code base I would like to stick with same terminology.

docs/semantics.md
```diff
@@ -519,15 +519,9 @@ Traditional file systems do not allow multiple directory entries with the same n
...
-Mounting bucket with such objects leads to crash `sync: unlock of unlocked mutex` or `Panic: Inode 'a/' cannot have child file`.
-`\n` in GCSFuse is used to resolve the name conflicts, in case there is a file and directory exists with the same name. Ref: [name-conflict](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#name-conflicts) section.
+### Unsupported Path names
 
+- GCSFuse (from v3.6.0 onwards) gracefully handles objects with names containing unsupported path segments (such as `//`, `/./`, `/../`). This ensures that file operations will not crash or throw errors. While listing shows only the supported objects, directory-level deletions and renames apply to all contained objects, whether supported or unsupported.
```
charith87: It might be good to mention its a limitation of Linux Filesystem bit which was mentioned before here.

docs/troubleshooting.md
```diff
@@ -110,7 +110,7 @@ Fuse package is not installed. It may throw this error. Run the following comman
 ### Encountered unsupported prefixes during listing, or ls: reading directory \<mountpath>/\<path>:  Input/output error
 Unable to list unsupported objects in a mounted bucket, i.e. objects which have names with `//` in them or have names starting with `/` e.g. `gs://<bucket>//A` or `gs://<bucket>/A//B` etc. Such objects can be listed by the following command: `gcloud storage ls gs://<bucket>/**//**`.
 
-This error can be mitigated in one of the following two ways.
+The `Input/output error` has been supressed from v3.6.0 onwards. In older vetsions, this error can be mitigated in one of the following two ways.
```
charith87: Typo is there - vetsions -> versions

charith87 (approved):

docs/semantics.md
```diff
@@ -519,15 +519,12 @@ Traditional file systems do not allow multiple directory entries with the same n
...
-
-
-- Objects in GCS with suffix `/\n` like, `gs://gcs-bkt/a/\n`:
-Mounting bucket with such objects leads to crash `sync: unlock of unlocked mutex` or `Panic: Inode 'a/' cannot have child file`.
-`\n` in GCSFuse is used to resolve the name conflicts, in case there is a file and directory exists with the same name. Ref: [name-conflict](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#name-conflicts) section.
+- Due to limitations in the Linux filesystem, path segments such as `//`, `/./`, or `/../` are not supported locally, even though they are valid object names in GCS. From v3.6.0 onwards, GCSFuse handles these objects gracefully:
 
+   1. Listing: To prevent system errors or crashes, these unsupported objects are hidden from file listings.
+   2. Rename/Delete: Directory-level operations (like deletions and renames) still apply to all contained objects, ensuring that unsupported objects are not accidentally left behind.
```
gargnitingoogle, line 527: Redundant information

```suggestion
   2. Rename/Delete: Directory-level operations still apply to all contained objects, ensuring that unsupported objects are not accidentally left behind.
```

docs/troubleshooting.md
```diff
@@ -110,7 +110,7 @@ Fuse package is not installed. It may throw this error. Run the following comman
 ### Encountered unsupported prefixes during listing, or ls: reading directory \<mountpath>/\<path>:  Input/output error
 Unable to list unsupported objects in a mounted bucket, i.e. objects which have names with `//` in them or have names starting with `/` e.g. `gs://<bucket>//A` or `gs://<bucket>/A//B` etc. Such objects can be listed by the following command: `gcloud storage ls gs://<bucket>/**//**`.
 
-This error can be mitigated in one of the following two ways.
+The `Input/output error` has been supressed from v3.6.0 onwards. In older versions, this error can be mitigated in one of the following two ways.
```
gargnitingoogle, line 113: Typo

```suggestion
The `Input/output error` has been suppressed from v3.6.0 onwards. In older versions, this error can be mitigated in one of the following two ways.
```

PR #4316, opened 2026-01-26
feat(kernel-settings): Write kernel config only for zonal buckets in GKE env when kernel reader is enabled

meet2mky:
### Description
Same as title. This change is as per revised plan of using continuous enforcement of kernel params.

b/478858189

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -277,17 +277,14 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		} else {
 			logger.Warnf("Cannot apply bucket-type optimizations as IsUserSet is nil")
 		}
-		// Write post mount kernel settings for Zonal Buckets in GKE environments for non dynamic mounts before user space mounting in GCSFuse.
-		// Mounting in GKE is already done at this point but writing kernel settings early ensures the asynchronous
-		// application of these settings happens as early as possible in GKE.
-		if serverCfg.NewConfig.FileSystem.KernelParamsFile != "" && bucketType.Zonal {
+		// Write post mount kernel settings for Zonal Buckets when kernel reader is enabled in GKE environments for
+		// non dynamic mounts before user space mounting in GCSFuse. Mounting in GKE is already done at this point but
+		// writing kernel settings early ensures the asynchronous application of these settings happens as early as possible in GKE.
+		if serverCfg.NewConfig.FileSystem.KernelParamsFile != "" && bucketType.Zonal && serverCfg.NewConfig.FileSystem.EnableKernelReader {
 			kernelParams := kernelparams.NewKernelParamsManager()
 			kernelParams.SetReadAheadKb(int(serverCfg.NewConfig.FileSystem.MaxReadAheadKb))
-			// Set max-background and congestion window when async read is enabled via kernel reader.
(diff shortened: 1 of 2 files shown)


internal/fs/fs.go
```diff
@@ -277,10 +277,10 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		} else {
 			logger.Warnf("Cannot apply bucket-type optimizations as IsUserSet is nil")
 		}
-		// Write post mount kernel settings in GKE environments for non dynamic mounts before user space mounting in GCSFuse.
+		// Write post mount kernel settings for Zonal Buckets in GKE environments for non dynamic mounts before user space mounting in GCSFuse.
 		// Mounting in GKE is already done at this point but writing kernel settings early ensures the asynchronous
 		// application of these settings happens as early as possible in GKE.
-		if serverCfg.NewConfig.FileSystem.KernelParamsFile != "" {
+		if serverCfg.NewConfig.FileSystem.KernelParamsFile != "" && bucketType.Zonal {
```
raj-prince: Instead of this, I think we should guard all the kernel setting change in GKE/Non-GKE under kernel-reader flag. 

This will keep the existing kernel-read=false behavior intact.

meet2mky: Done, PTAL!

internal/fs/fs.go
```diff
@@ -277,17 +277,14 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
...
 			logger.Warnf("Cannot apply bucket-type optimizations as IsUserSet is nil")
 		}
-		// Write post mount kernel settings in GKE environments for non dynamic mounts before user space mounting in GCSFuse.
-		// Mounting in GKE is already done at this point but writing kernel settings early ensures the asynchronous
-		// application of these settings happens as early as possible in GKE.
-		if serverCfg.NewConfig.FileSystem.KernelParamsFile != "" {
+		// Write post mount kernel settings for Zonal Buckets when kernel reader is enabled in GKE environments for
+		// non dynamic mounts before user space mounting in GCSFuse. Mounting in GKE is already done at this point but
+		// writing kernel settings early ensures the asynchronous application of these settings happens as early as possible in GKE.
+		if serverCfg.NewConfig.FileSystem.KernelParamsFile != "" && bucketType.Zonal && serverCfg.NewConfig.FileSystem.EnableKernelReader {
```
raj-prince, line 283: nit: should we remove the bucketType check from here also?

meet2mky: I think, we can! Let me change it as well.

meet2mky: No, I think we shouldn't because we don't want to write kernel config file to GKE for regional buckets in any scenario as it could write the unintended configuration if somehow customer set `--max-read-ahead-kb` hidden flag which would override the CSI `read_ahead_kb`. I think we should keep this check here.

raj-prince: Okay, sounds good. Although it's rare if any customer will enable kernel-reader and kernel-settings both on regional bucket.

raj-prince (approved):

internal/fs/fs.go
meet2mky: Discussed offline, current changes are sufficient.

PR #4329, opened 2026-01-28
feat: optimize sparse downloads with in-flight tracking and accurate accounting

vipnydav:
### Description
This PR optimizes the sparse file cache implementation to reduce latency and improve cache integrity. Key improvements include:

- **Deduplicated Downloads**: Implemented an `inflightChunks` map to synchronize concurrent requests for the same missing blocks, preventing redundant GCS egress and local disk I/O.
- **Targeted Segment Fetching**: Refactored the download logic to identify and fetch only missing segments within a requested range, rather than re-downloading entire contiguous spans.
- **Concurrent Block Fetching**: Optimized multi-block requests (e.g., reads spanning multiple chunks) by creating separate readers for each missing segment and downloading them in parallel using goroutines to minimize latency.
- **Accurate Usage Accounting**: Updated **ByteRangeMap** to be file-size aware, fixing an issue where cache usage was over-counted for files not aligned to chunk boundaries.

https://b.corp.google.com/issues/479985831

### Testing details
1. Manual - Done
2. Unit tests - Added


internal/cache/data/byte_range_map.go
```diff
@@ -52,8 +49,21 @@ func (brm *ByteRangeMap) chunkID(offset uint64) uint64 {
 	return offset / brm.chunkSize
 }
 
+// chunkSizeOf returns the size of a specific chunk, handling the last chunk which might be smaller.
+func (brm *ByteRangeMap) chunkSizeOf(chunkID uint64) uint64 {
```
vadlakondaswetha, line 53: call it chunksize

vadlakondaswetha, line 53: please add unit test for this method

internal/cache/file/downloader/sparse_downloads_job_test.go
```diff
@@ -121,22 +116,85 @@ func (dt *sparseDownloaderTest) Test_calculateSparseChunkBoundaries() {
...
+	// Should have 1 wait channel
+	AssertEq(1, len(waitChans))
+	AssertEq(inflightCh, waitChans[0])
+	// Verify inflightChunks map was updated
+	dt.job.mu.Lock()
+	_, ok0 := dt.job.inflightChunks[0]
+	_, ok1 := dt.job.inflightChunks[1]
+	_, ok2 := dt.job.inflightChunks[2]
+	dt.job.mu.Unlock()
+	AssertTrue(ok0, "Chunk 0 should be marked inflight")
```
vadlakondaswetha, line 193: i guess the assert message here is wrong ? there is already one inflightChunk, so ok0 should remain inflight, ok1, ok2 should be marked inflight.

vipnydav: ok0 maps to chunk0 but In Line#160 we marked chunk1 to be inflight so it would remain inflight. And after than in Line#177 we mark chunk0 and chunk2 inflight. So the assert message is correct.

internal/cache/data/byte_range_map.go
vipnydav: Added

vipnydav: We can't call it `chunkSize` since `ByteRangeMap` already has a field named `chunkSize` and we get this error:

```
field and method with the same name chunkSize
```

vipnydav:
CC: @xinyangge as they were the original contributor of the sparse downloads feature and may have valuable insights on these changes.

vadlakondaswetha (approved):

PR #4335, opened 2026-02-02
fix(exclude regex tests): Correction for only dir tests flags for exclude regex test

PranjalC100:
### Description
This PR fixes a configuration gap in the integration test `TestCacheFileForExcludeRegexTest`.

Currently, when running tests with `OnlyDirMounted` (mounting a specific directory rather than the root bucket), the `--enable-kernel-reader=false` flag is not passed to GCSFuse. This flag is required to ensure consistent behavior with other read cache tests and to prevent test flakiness related to kernel readahead interacting with the test's cache validation logic.

### Changes
- Updated `tools/integration_tests/read_cache/cache_file_for_exclude_regex_test.go` to include `--enable-kernel-reader=false` in the flag set for directory mounts.

https://buganizer.corp.google.com/issues/480942316

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - NA


vipnydav (approved):
LGTM. Did you test it out manually that before this change we were not getting `--enable-kernel-reader=false` in the flags for `only-dir` mounts and after this change we started getting the flag and the test passes?

If so, update the Manual Testing to `Done`.

PranjalC100:
> LGTM. Did you test it out manually that before this change we were not getting `--enable-kernel-reader=false` in the flags for `only-dir` mounts and after this change we started getting the flag and the test passes?
> 
> If so, update the Manual Testing to `Done`.

Yes I've verified the same!

ashmeenkaur (approved):

PR #4359, opened 2026-02-09
docs(buffered-reader): update semantics for phase 2 improvements

vipnydav:
### Description
This PR updates the `docs/semantics.md` file to reflect the improvements introduced in Phase 2 of the Buffered Reader feature.

b/479099905

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/docs/semantics.md
+++ b/docs/semantics.md
@@ -19,7 +19,7 @@ The feature is **disabled by default** and can be enabled using:
 **Note:** Buffered reads are designed to operate exclusively when the file cache is disabled. If both features are enabled, the file cache takes precedence and buffered reads will be ignored.
 
 **Best Use Cases:**
-- Applications using fewer threads (typically less than 8) reading large files (> 100MB) with sequential access patterns.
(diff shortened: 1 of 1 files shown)


docs/semantics.md
```diff
@@ -19,10 +19,10 @@ The feature is **disabled by default** and can be enabled using:
 **Note:** Buffered reads are designed to operate exclusively when the file cache is disabled. If both features are enabled, the file cache takes precedence and buffered reads will be ignored.
 
 **Best Use Cases:**
-- Single-threaded applications reading large files (> 100MB) sequentially.
+- Applications using fewer threads (typically less than 8) reading large files (> 100MB) with sequential access patterns.
```
raj-prince: Thinking again, I feel, we should remove the thread constraint and use memory constraint.

We can write something like following:
Applications reading large file (> 100MB) with sequential access patterns, Cx needs to pass enough buffer memory <via global-max-blocks> to allow for higher parallelism (default buffer will allow up to 4 parallelism).

raj-prince (approved):

PR #4365, opened 2026-02-09
feat(shared-chunk-cache-gc): adding support of running cron based lru for shared-cache

raj-prince:
### Description
- Complete shared-cache support: https://github.com/GoogleCloudPlatform/gcsfuse/pull/4346
- Provides command utility to run cron based lru cache over the shared cache directory

b/484232306

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/gcsfuse-scc-gc/README.md
+++ b/tools/gcsfuse-scc-gc/README.md
@@ -18,8 +18,19 @@ go build .
 **Options:**
 - `-cache-dir` (required): Path to cache directory
 - `-target-size-mb`: Target size in MB (default: 10240)
+- `-concurrency`: Maximum concurrent file operations (default: 16)
 - `-dry-run`: Show what would be deleted
 - `-debug`: Enable debug logging
+
+## File System Requirements
+
+This garbage collector tool requires a file system that supports:
+
+1. **Atomic rename**: The tool uses atomic rename operations to safely expire cache files (renaming `.bin` to `.bak`) without disrupting concurrent reads from multiple GCSFuse instances.
+
+2. **Access time (atime) tracking**: The LRU eviction algorithm depends on file access times to determine which chunks are least recently used. The file system should track atime, either through:
+
+**Note:** NFS and most POSIX-compliant file systems meet these requirements.
 
 ## Scheduling
 
@@ -82,14 +93,7 @@ concurrent read and eviction of chunked file.
 ```
 Output:
 ```
-time=2026-02-10T10:38:30.989Z level=INFO msg="Starting LRU cache eviction" cache_dir=/mnt/nfs-cache target_size_mb=1024
-time=2026-02-10T10:38:31.244Z level=DEBUG msg="Manifest created" files=100 total_size_mb=800 scan_duration=126.255393ms
-time=2026-02-10T10:38:31.244Z level=INFO msg="Cache below target, nothing to do" cache_size_mb=800 target_size_mb=1024
+time=2026-02-12T19:06:36.237Z level=INFO msg="Starting LRU cache eviction" cache_dir=/mnt/nfs-cache target_size_mb=10240
+time=2026-02-12T19:06:36.640Z level=DEBUG msg="Manifest created" files=100 total_size_mb=800 scan_duration=120.731382ms
+time=2026-02-12T19:06:36.640Z level=INFO msg="Cache below target, nothing to do" cache_size_mb=800 target_size_mb=10240
 ```
-
-## Notes
-
-- **Multi-host safe**: Can run concurrently on multiple hosts
-- **Error tolerant**: Handles ENOENT/ENOTEMPTY gracefully
-- **Parallel operations**: 10 concurrent file operations
-- Works with SHA256-based cache structure: `cache-dir/XX/YY/hash/offset_offset.bin`
--- a/tools/gcsfuse-scc-gc/main.go
+++ b/tools/gcsfuse-scc-gc/main.go
@@ -15,13 +15,16 @@
 /**
 * How does gcsfuse-scc-gc (GCSFuse Shared Chunk Cache Garbage Collector) work?
 *
-* 1. Cleans up `.bak` files expired during the previous run.
-* 2. Scans cache directory for `.bin` files with atime and size
-* 3. If total size < target, then exit without expiration or eviction.
-* 4. Sorts by atime and selects oldest files to expire.
-* 5. Renames selected files to `.bak` (kept until next run for ongoing reads).
-* 6. Removes old `.tmp` files (older than 1 hour)
-* 7. Cleans up empty directories.
+* 1. Performs a single walk to scan the cache directory, collecting:
+*    - `.bin` files (cache chunks) with atime/mtime and size
+*    - `.bak` files (previously expired files from last run)
+*    - `.tmp` files (incomplete downloads)
+*    - Directories (for cleanup)
+* 2. Cleans up `.bak` files expired during the previous run.
+* 3. Sorts `.bin` files by atime and selects oldest files to expire, only if cache size exceeds target.
+* 4. Renames selected files to `.bak` (kept until next run for ongoing reads).
+* 5. Removes old `.tmp` files (older than 1 hour).
+* 6. Cleans up empty directories (deepest first).
  */
 
 package main
@@ -32,7 +35,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
-	"sort"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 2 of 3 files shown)


tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,373 @@
...
+			return nil
+		}
+
+		atime := time.Unix(stat.Atim.Sec, stat.Atim.Nsec)
+		mtime := time.Unix(stat.Mtim.Sec, stat.Mtim.Nsec)
+
+		// Use mtime as fallback if atime looks stale (not updated due to noatime/relatime)
+		// If atime equals mtime, it likely means atime isn't being updated
+		effectiveTime := atime
+		if atime.Equal(mtime) || atime.Before(mtime) {
```
vadlakondaswetha: why not just take the mtime always

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,373 @@
...
+		return []FileInfo{}
+	}
+
+	bytesToExpire := manifest.TotalSize - targetSize
+
+	// Sort by atime (oldest first)
+	sorted := make([]FileInfo, len(manifest.Files))
+	copy(sorted, manifest.Files)
+
+	sort.Slice(sorted, func(i, j int) bool {
```
vadlakondaswetha: why not sort directly the manifest.files?

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,373 @@
...
+// expireFiles renames files to .bak in parallel
+func expireFiles(files []FileInfo) {
+	var wg sync.WaitGroup
+	semaphore := make(chan struct{}, 10) // Limit concurrency
+
+	for _, f := range files {
+		wg.Add(1)
+		go func(file FileInfo) {
+			defer wg.Done()
+			semaphore <- struct{}{}
```
vadlakondaswetha: why not use a semaphore.Acquire method here?

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,385 @@
...
+	if *cacheDir == "" {
+		slog.Error("cache-dir is required")
+		os.Exit(1)
+	}
+
+	slog.Info("Starting LRU cache eviction", "cache_dir", *cacheDir, "target_size_mb", *targetSize)
+
+	// Step 1: Clean up any previous .bak expired as part of previous run.
+	if !*dryRun {
+		removeAllBakFiles(*cacheDir)
```
geertj: I recommend we drive removing the .bak files from the manifest. The path walk in the cache directory can be extremely expensive (hours for a very large cache). We therefore only want to walk the directory once per invocation of the GC. I think that the code as-is does 4 full walks of the cache dir?

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,385 @@
...
+			if err := os.Rename(file.Path, bakPath); err != nil && !os.IsNotExist(err) {
+				slog.Warn("Failed to rename file", "path", file.Path, "error", err)
+			}
+		}(f)
+	}
+
+	wg.Wait()
+}
+
+// removeOldTmpFiles removes .tmp files older than 1 hour
```
geertj: Mention why the 1 hour. It is a safety margin so that in-progress serving request for those files can complete. When we renamed the .bin file to .bak, no new requests will be started, and because rename preserves the file handle, existing request can finish.

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+	}
+
+	return selected
+}
+
+// expireFiles renames files to .bak in parallel
+func expireFiles(files []FileInfo) {
+	var wg sync.WaitGroup
+	semaphore := make(chan struct{}, 10) // Limit concurrency
```
geertj: Can we make the concurrency a flag? A hard coded default is tricky as it is very dependent on the NFS server, and whether you run the expiration on a separate server or not. The maximum default concurrency per NFS mount is 64. Therefore I think the default value of the flag can be a bit higher, maybe 16 or 20.

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,385 @@
...
+			}
+		}(f)
+	}
+
+	wg.Wait()
+}
+
+// removeOldTmpFiles removes .tmp files older than 1 hour
+func removeOldTmpFiles(cacheDir string) {
+	cutoff := time.Now().Add(-1 * time.Hour)
```
geertj: Drive this from the manifest as well. I would include any file in it, and also capture the atime and the mtime in the manifest. When removing tmp files I think you want to test "mtime".

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,385 @@
...
+
+	bytesToExpire := manifest.TotalSize - targetSize
+
+	// Sort by atime (oldest first)
+	sorted := make([]FileInfo, len(manifest.Files))
+	copy(sorted, manifest.Files)
+
+	sort.Slice(sorted, func(i, j int) bool {
+		return sorted[i].Atime.Before(sorted[j].Atime)
+	})
```
geertj: Nit: use slices.Sort() to sort in-line?

tools/gcsfuse-scc-gc/README.md
```diff
@@ -0,0 +1,95 @@
...
+```
+time=2026-02-10T10:38:30.989Z level=INFO msg="Starting LRU cache eviction" cache_dir=/mnt/nfs-cache target_size_mb=1024
+time=2026-02-10T10:38:31.244Z level=DEBUG msg="Manifest created" files=100 total_size_mb=800 scan_duration=126.255393ms
+time=2026-02-10T10:38:31.244Z level=INFO msg="Cache below target, nothing to do" cache_size_mb=800 target_size_mb=1024
+```
+
+## Notes
+
+- **Multi-host safe**: Can run concurrently on multiple hosts
+- **Error tolerant**: Handles ENOENT/ENOTEMPTY gracefully
```
geertj: Do you mean that this allows you to use multiple concurrent instances of the GC? If so you can say that we're safe against that, but that it's not recommended.

tools/gcsfuse-scc-gc/README.md
```diff
@@ -0,0 +1,95 @@
+# GCSFuse Shared Chunk Cache (SCC) Garbage Collector
```
geertj, line 1: Can we add a section here on file system requirements? I think there are just two:
- atomic rename
- atime tracking
We should also add some more general documentation on this feature but we can do that later when if/when we will officially support it.

tools/gcsfuse-scc-gc/main.go
raj-prince: Aah, blunder.  Yeah, we can get all the required fileInfo in one scan. Thanks!

tools/gcsfuse-scc-gc/main.go
raj-prince: For system not using relatime, atime is more accurate.

For system using relatime, atime may be more recent. So, used recent of aTime and mTime.

tools/gcsfuse-scc-gc/README.md
raj-prince: Removing these notes, doesn't add much value. If required, will add later.

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,410 @@
...
+	start := time.Now()
+	manifest := &Manifest{
+		Files:    make([]FileInfo, 0),
+		BakFiles: make([]FileInfo, 0),
+		TmpFiles: make([]FileInfo, 0),
+		Dirs:     make([]string, 0),
+	}
+
+	// Look for gcsfuse-file-cache subdirectory
+	actualCacheDir := filepath.Join(cacheDir, "gcsfuse-file-cache")
```
geertj, line 175: I've not checked this, but can you make sure this directory is different from what filecache would use? If a user would switch between filecache and scc on the same host, then we don't want that to be in different directories.

tools/gcsfuse-scc-gc/main.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	bytesToExpire := manifest.TotalSize - targetSize
+
+	// Sort by atime (oldest first)
+	slices.SortFunc(manifest.Files, func(a, b FileInfo) int {
+		recentA := a.Atime
+		if recentA.Before(a.Mtime) {
+			recentA = a.Mtime
+		}
```
geertj, line 331: I /think/ that if mtime is updated, atime is always updated. But let's not depend on that and keep this as-is.

tools/gcsfuse-scc-gc/main.go
raj-prince: Yeah, make sense. Will make change for both core implementation and gc-change together in the separate PR.

PR #4370, opened 2026-02-10
feat(dir-metadata-prefetch): Add context-based cancellation to metadata prefetcher

ashmeenkaur:
### Description
This PR implements the foundation for handling race conditions between the asynchronous metadata prefetcher and foreground write operations. 

It introduces a context-based cancellation mechanism where:
1.  **`DirInode`** now maintains a lifecycle `Context` that is cancelled on `Destroy`.
2.  **`MetadataPrefetcher`** respects this context and allows cancelling specific prefetch runs via `CancelCurrDirPrefetcher`.
3.  **`FastStatBucket`** (stat cache) now accepts a `Context` in insert methods and performs a critical section check (`ctx.Err()`) inside the fast stat lock to prevent updating the cache with stale data if the operation was cancelled.

**Note:** The integration of `CancelCurrDirPrefetcher` and `CancelSubdirectoryPrefetches` into `DirInode` write operations (like `Unlink`, `Rename`) will be added in a follow-up PR.

Fixes b/483297261

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

NA

### DIFF
--- a/internal/fs/inode/base_dir.go
+++ b/internal/fs/inode/base_dir.go
@@ -300,13 +300,13 @@ func (d *baseDirInode) IsTypeCacheDeprecated() bool {
 	return d.isEnableTypeCacheDeprecation
 }
 
-// CancelCurrDirPrefetcher stops only the *current* prefetch run for this dir.
 func (d *baseDirInode) CancelCurrDirPrefetcher() {}
 
-// CancelSubdirectoryPrefetches permanently stops the context for this inode AND all its descendants.
-// This is used for Directory Rename/Delete operations. This context is not refreshed
-// with the assumption that Rename/Delete directory operations will lead to new inodes with fresh contexts.
 func (d *baseDirInode) CancelSubdirectoryPrefetches() {}
 
-// Context provides the lifecycle context of the inode.
-func (d *baseDirInode) Context() context.Context { return nil }
+func (d *baseDirInode) Context() context.Context {
+	// TODO: Consider implementing Context() if it simplifies the code in the future.
+	// Currently, baseDirInode is the root for dynamic mounts where listing (and thus prefetching)
+	// is not allowed, so a nil context suffices.
+	return nil
+}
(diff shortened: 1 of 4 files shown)


internal/fs/inode/dir_prefetcher.go
```diff
@@ -78,31 +83,42 @@ func NewMetadataPrefetcher(
...
 // It uses an atomic state to prevent concurrent execution.
+// This function is already protected by directory mutex so no new mutex is required here for the setup part.
 func (p *MetadataPrefetcher) Run(fullObjectName string) {
 	// Do not trigger prefetching if:
 	// 1. metadata prefetch config is disabled.
 	// 2. metadata cache ttl is 0 (disabled).
 	// 3. stat cache size is 0.
-	if !p.enabled || p.metadataCacheTTL == 0 || p.statCacheSize == 0 {
+	// 4. The inode context is already cancelled (dir inode is dead/renamed).
+	if !p.enabled || p.metadataCacheTTL == 0 || p.statCacheSize == 0 || p.inodeCtx.Err() != nil {
```
charith87: should p.InodeCtx != nil be also here.

internal/fs/inode/base_dir.go
```diff
@@ -299,3 +299,14 @@ func (d *baseDirInode) Unlink() {
...
+// CancelCurrDirPrefetcher stops only the *current* prefetch run for this dir.
+func (d *baseDirInode) CancelCurrDirPrefetcher() {}
+
+// CancelSubdirectoryPrefetches permanently stops the context for this inode AND all its descendants.
+// This is used for Directory Rename/Delete operations. This context is not refreshed
+// with the assumption that Rename/Delete directory operations will lead to new inodes with fresh contexts.
+func (d *baseDirInode) CancelSubdirectoryPrefetches() {}
+
+// Context provides the lifecycle context of the inode.
+func (d *baseDirInode) Context() context.Context { return nil }
```
charith87: Would the nil be replaced in a followup

ashmeenkaur: Base dir inode is the root inode for dynamic mounting. We don't allow listing on the root inode in case of dynamic mounts, so we won't do prefetching here. I think we can handle without a context here, but in a follow-up PR, if I see that adding a context makes the code simpler, I will add it. Added a TODO comment.

PR #4375, opened 2026-02-13
test: Randomize test directory names in negative stat cache tests

vadlakondaswetha:
Use random directory names for negative stat cache tests. This prevents collisions 

- when running tests in parallel or concurrently on the same bucket.
- When the cleanup doesn't happen properly

---
*PR created automatically by Jules for task [10178962003327661698](https://jules.google.com/task/10178962003327661698) started by @vadlakondaswetha*

### DIFF
--- a/tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
+++ b/tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
@@ -28,13 +28,14 @@ import (
 )
 
 type disabledNegativeStatCacheTest struct {
-	flags []string
+	flags   []string
+	testDir string
 	suite.Suite
 }
 
 func (s *disabledNegativeStatCacheTest) SetupTest() {
-	testDir := testDirName + setup.GenerateRandomString(5)
-	mountGCSFuseAndSetupTestDir(s.flags, testDir)
+	s.testDir = testDirName + setup.GenerateRandomString(5)
+	mountGCSFuseAndSetupTestDir(s.flags, s.testDir)
 }
 
 func (s *disabledNegativeStatCacheTest) TearDownTest() {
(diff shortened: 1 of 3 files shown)


tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
```diff
@@ -28,16 +28,19 @@ import (
...
 
 func (s *disabledNegativeStatCacheTest) SetupTest() {
-	mountGCSFuseAndSetupTestDir(s.flags, testDirName)
+	s.testDir = testDirName + setup.GenerateRandomString(5)
+	mountGCSFuseAndSetupTestDir(s.flags, s.testDir)
 }
 
 func (s *disabledNegativeStatCacheTest) TearDownTest() {
 	setup.UnmountGCSFuse(testEnv.rootDir)
+	setup.CleanupDirectoryOnGCS(testEnv.ctx, testEnv.storageClient, path.Join(setup.TestBucket(), s.testDir))
```
Tulsishah: Will it work for onlyDirMounting or Dynamic mounting case? Instead of this can we use GCSFuse cleanup function like clean up before unmount? That will take care of onlyDirMounting or Dynamic mounting cases.

tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
```diff
@@ -28,16 +28,19 @@ import (
...
 type disabledNegativeStatCacheTest struct {
-	flags []string
+	flags   []string
+	testDir string
 	suite.Suite
 }
 
 func (s *disabledNegativeStatCacheTest) SetupTest() {
-	mountGCSFuseAndSetupTestDir(s.flags, testDirName)
+	s.testDir = testDirName + setup.GenerateRandomString(5)
```
charith87: Should we also add the TestName also in this. example negativeStatCache/DisabledNegativeStatCacheTest/<Random> That way the tests will use a heirarchy for the objects.

tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
charith87, line 54: Should this also be modified to new path?

vadlakondaswetha: testEnv.testDirPath is updated to new path. so its not required here again

tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
vadlakondaswetha: As long as the directory for each test is different, it shouldn;t matter.

tools/integration_tests/negative_stat_cache/disabled_negative_stat_cache_test.go
vadlakondaswetha: done.

Tulsishah (dismissed):

PR #4378, opened 2026-02-13
fix(shared-chunk-cache): creating different sub-directory for shared-chunk-cache than traditional file-cache

raj-prince:
### Description
- Keeping the shared-chunk-cache in the same directory lead to some toil for the customer who would want to switch b/w the shared-chunk-cache and file-cache on the same VM. They need to pass different cache-directory to do so. Hence, keeping separate directory for shared-chunk-cache and traditional file-cache.
- Also, adding some error/retry related test in shared_chunk_cache_reader_test.go.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -318,7 +318,7 @@ func createFileCacheHandler(serverCfg *ServerConfig) (fileCacheHandler *file.Cac
 	}
 
 	// Regular gcsfuse file-cache with memory based LRU cache.
-	fileCacheHandler, err = createInMemoryFileCacheHandler(baseCacheDir, filePerm, dirPerm, serverCfg)
+	fileCacheHandler, err = createSingleMountFileCacheHandler(baseCacheDir, filePerm, dirPerm, serverCfg)
 	return fileCacheHandler, nil, err
 }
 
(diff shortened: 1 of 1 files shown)


internal/fs/fs.go
```diff
@@ -307,36 +307,54 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
...
+	if err != nil {
+		return nil, fmt.Errorf("createSharedChunkCacheManager: while creating shared chunk cache manager: %w", err)
+	}
+
+	logger.Infof("File Cache: Shared chunk cache created successfully at %s", cacheDir)
+	return sharedCacheManager, nil
+}
+
+// createInMemoryFileCacheHandler creates a file cache handler with in-memory LRU eviction for single-instance usage.
+func createInMemoryFileCacheHandler(baseCacheDir string, filePerm, dirPerm os.FileMode, serverCfg *ServerConfig) (*file.CacheHandler, error) {
```
geertj: I think "InMemoryFileCacheHandler" is a bit of a misleading name. When I read this for the first time, I thought this was going to set up a filecache on tmpfs.

Maybe "LocalFileCacheHandler" or "SingleNodeFileCacheHandler"?

PR #4383, opened 2026-02-17
fix(dir-metadata-prefetch): metadata prefetch race condition with active writer tracking

ashmeenkaur:
## Problem
Metadata prefetching is an asynchronous process. A race condition exists where:

A prefetch is triggered (e.g., by a Lookup call).
A write operation (e.g., Delete) occurs, updating the cache with fresh "Unknown" status for the deleted file.
The slow prefetch finishes and writes "RegularFile" (stale data) back into the cache, overwriting the fresh update.
## Solution
The fix introduces Active Writer Tracking and Proactive Prefetch Cancellation:

1. Active Writer Tracking: Added an atomic counter activeWriters to dirInode to track ongoing write operations.
2. Writer-Aware Prefetch: The MetadataPrefetcher now accepts a shouldRun callback that prevents a new prefetch from starting if there are active writers in the directory.
3. Explicit Cancellation: Critical write operations (including CloneToChildFile, DeleteChildFile, DeleteChildDir, RenameFile, and RenameFolder) now increment the writer count and explicitly cancel any in-flight prefetchers for that directory.
4. FS-Level Integration: Writer tracking and prefetch cancellation are also integrated into flushFile and syncFile in fs.go to protect the cache during file synchronization.
5. Recursive Cancellation: For directory-level changes (like renaming or deleting a folder), the cancellation is propagated recursively to all child directory prefetchers.
## Bug ID
b/483297572
## Key Changes
- internal/fs/inode/dir.go: Implemented IncrementActiveWriters and DecrementActiveWriters logic.
- internal/fs/inode/dir_prefetcher.go: Updated the prefetcher to check the shouldRun condition before execution.
- internal/fs/fs.go: Added findParentDirInode helper to manage writer tracking during file flushes.
## Testing Performed
- TestPrefetch_SkipIfActiveWriters: Verifies that prefetching is skipped when a write operation is in progress.
- TestPrefetch_RaceWithDeleteChildFile: Simulates a slow GCS list call returning stale data and verifies that a concurrent DeleteChildFile (which updates the cache) is not overwritten by the stale prefetch result.
- TestRecursiveCancellation: Ensures that cancelling a parent directory's prefetcher correctly propagates through the hierarchy.

### DIFF
--- a/internal/fs/inode/dir_prefetcher.go
+++ b/internal/fs/inode/dir_prefetcher.go
@@ -94,14 +94,14 @@ func (p *MetadataPrefetcher) Run(fullObjectName string) {
 		return
 	}
 
-	//6. Do not trigger prefetch if the last prefetch result is still within the TTL.
+	// Do not trigger prefetch if the last prefetch result is still within the TTL.
 	lastPrefetchTime := p.lastPrefetchTime.Load()
 	now := p.cacheClock.Now()
 	if lastPrefetchTime != nil && now.Sub(*lastPrefetchTime) < p.metadataCacheTTL {
 		return
 	}
 
-	// 6. Ensure only one prefetch runs at a time for this directory.
+	// Ensure only one prefetch runs at a time for this directory.
 	if !p.state.CompareAndSwap(prefetchReady, prefetchInProgress) {
 		return
 	}
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -666,6 +666,7 @@ type fileSystem struct {
 // Helpers
 ////////////////////////////////////////////////////////////////////////
 
+// LOCKS_REQUIRED(fs.mu)
 func (fs *fileSystem) findParentDirInode(childName inode.Name) inode.DirInode {
 	parentName, err := childName.ParentName()
 	if err != nil {
@@ -1364,6 +1365,17 @@ func (fs *fileSystem) flushFile(
 		return nil
 	}
 
+	fs.mu.Lock()
+	parInode := fs.findParentDirInode(f.Name())
+	fs.mu.Unlock()
+	if parInode != nil {
+		// Increment active writers so no new prefetch gets triggered until write operation completes.
+		parInode.IncrementActiveWriters()
+		defer parInode.DecrementActiveWriters()
+		// Cancel current directory's prefetch so stale data is not updated in cache.
+		parInode.CancelCurrDirPrefetcher()
+	}
+
 	// Flush the inode.
 	err := f.Flush(ctx)
 	if err != nil {
@@ -1392,6 +1404,17 @@ func (fs *fileSystem) syncFile(
 	// when file to be synced has been unlinked from the same mount.
 	if f.IsUnlinked() {
 		return nil
+	}
+
+	fs.mu.Lock()
+	parInode := fs.findParentDirInode(f.Name())
+	fs.mu.Unlock()
+	if parInode != nil {
+		// Increment active writers so no new prefetch gets triggered until write operation completes.
+		parInode.IncrementActiveWriters()
+		defer parInode.DecrementActiveWriters()
+		// Cancel current directory's prefetch so stale data is not updated in cache.
+		parInode.CancelCurrDirPrefetcher()
 	}
 
 	// Sync the inode.


internal/fs/inode/dir.go
```diff
@@ -1082,6 +1108,12 @@ func (d *dirInode) EraseFromTypeCache(name string) {
 
 // LOCKS_REQUIRED(d)
 func (d *dirInode) CloneToChildFile(ctx context.Context, name string, src *gcs.MinObject) (*Core, error) {
```
vadlakondaswetha, line 1117: can we rename to an existing file - i meant if the destination exists can we rename source to the destination?

internal/fs/inode/dir.go
```diff
@@ -1139,6 +1173,7 @@ func (d *dirInode) CreateChildSymlink(ctx context.Context, name string, target s
 
 // LOCKS_REQUIRED(d)
 func (d *dirInode) CreateChildDir(ctx context.Context, name string) (*Core, error) {
+	// No need to cancel prefetch here as creation of new file can not lead to stale data in metadata cache.
```
vadlakondaswetha: type - creation of new directory not file

internal/fs/inode/dir.go
```diff
@@ -1424,6 +1483,17 @@ func (d *dirInode) RenameFile(ctx context.Context, fileToRename *gcs.MinObject,
 }
 
 func (d *dirInode) RenameFolder(ctx context.Context, folderName string, destinationFolderName string, folderInode DirInode) (*gcs.Folder, error) {
```
vadlakondaswetha: rename the folderInode to dstFolderInode so its clear what we are referring to source or destination.

internal/fs/inode/dir_prefetcher.go
```diff
@@ -90,11 +96,12 @@ func (p *MetadataPrefetcher) Run(fullObjectName string) {
 	// 2. metadata cache ttl is 0 (disabled).
 	// 3. stat cache size is 0.
 	// 4. The inode context is nil or already cancelled (dir inode is dead/renamed).
-	if !p.enabled || p.metadataCacheTTL == 0 || p.statCacheSize == 0 || p.inodeCtx == nil || p.inodeCtx.Err() != nil {
+	// 5. If there are active writers in the directory, do not trigger prefetch. This is determined by shouldRun function.
+	if !p.enabled || p.metadataCacheTTL == 0 || p.statCacheSize == 0 || p.inodeCtx == nil || p.inodeCtx.Err() != nil || !p.shouldRun() {
```
vadlakondaswetha: I haven;t checked the previous PRs but when prefetcher is not enabled - i am guessing its a mount level config why is the prefetcher getting initialized?

vadlakondaswetha: if any write is in progress, we are not triggering the prefetch. Are we not triggering it after write is done?

internal/fs/inode/dir.go
ashmeenkaur: Yes it is allowed, logs for ref:

```
{"timestamp":{"seconds":1772033688,"nanos":176077143},"message":"fuse_debug: Op 0x00000012        connection.go:468] <- LookUpInode (parent 1, name \"b.txt\", PID 3340852)"}
{"timestamp":{"seconds":1772033688,"nanos":176243521},"message":"fuse_debug: Op 0x00000012        connection.go:567] -> LookUpInode (inode 2)"}
{"timestamp":{"seconds":1772033688,"nanos":176338929},"message":"fuse_debug: Op 0x00000014        connection.go:468] <- LookUpInode (parent 1, name \"a.txt\", PID 3340852)","mount-id":"193ef04c"}
{"timestamp":{"seconds":1772033688,"nanos":176477234},"message":"fuse_debug: Op 0x00000014        connection.go:567] -> LookUpInode (inode 3)","mount-id":"193ef04c"}
{"timestamp":{"seconds":1772033688,"nanos":177116527},"message":"fuse_debug: Op 0x00000016        connection.go:468] <- Rename (PID 3340852, old_parent 1, old_name \"b.txt\", new_parent 1, new_name \"a.txt\")","mount-id":"193ef04c"}
{"timestamp":{"seconds":1772033688,"nanos":177338236},"message":"gcs: Req              0x4: <- CopyObject(\"ashmeen-test/b.txt\", \"ashmeen-test/a.txt\")","mount-id":"193ef04c"}
{"timestamp":{"seconds":1772033688,"nanos":237079557},"message":"gcs: Req              0x4: -> CopyObject(\"ashmeen-test/b.txt\", \"ashmeen-test/a.txt\") (59.745018ms): OK","mount-id":"193ef04c"}
{"timestamp":{"seconds":1772033688,"nanos":237140013},"message":"gcs: Req              0x5: <- DeleteObject(\"ashmeen-test/b.txt\")","mount-id":"193ef04c"}
{"timestamp":{"seconds":1772033688,"nanos":273515206},"message":"gcs: Req              0x5: -> DeleteObject(\"ashmeen-test/b.txt\") (36.358336ms): OK","mount-id":"193ef04c"}
{"timestamp":{"seconds":1772033688,"nanos":273630670},"message":"fuse_debug: Op 0x00000016        connection.go:567] -> Rename ()","mount-id":"193ef04c"}
```

internal/fs/inode/dir_prefetcher.go
ashmeenkaur: The prefetch will be triggered on the next lookup, provided there are no active write operations in progress. We've designed it this way (not maintaining any queue) to avoid unnecessary prefetching in cases where the workload might have moved away from that specific directory.

ashmeenkaur: Thanks, created a separate PR to fix this - https://github.com/GoogleCloudPlatform/gcsfuse/pull/4419

PR #4401, opened 2026-02-18
feat(gcsx): treat first read at non-zero offset as random

vipnydav:
### Description
This PR improves the GCSFuse read prefetch algorithm by treating the first request at a non-zero offset as a random read by default. Currently, GCSFuse assumes sequential intent even for the first request, which often leads to unnecessary `200 MiB` prefetches for random access workloads.

As part of this change, we have removed the reliance on `minSeeksForRandomRead` (or equivalent seek-based thresholds) for initial classification. Previously, the algorithm required multiple seeks to transition to a random read type. Now, the first request itself is classified as random if it occurs at a non-zero offset, which triggers the random-friendly heuristic immediately.

This change is supported by data analysis of Bigstore logs, which shows that less than `7%` of read requests starting at non-zero offsets are actually sequential. By treating these as random from the start, we mitigate several critical issues:

- **Memory Ballooning**: Reduces server-side memory swell caused by large readers for small data portions.
- **Tail Latencies**: Decreases "connection reset by peer" errors and retries by ensuring readers are more likely to be fully drained and TCP connections to be reused.

This logic aligns with Linux kernel behavior, which uses a growing window approach for non-zero offsets while maintaining immediate readahead for offset zero.

For more details, please refer to [go/improve-gcsfuse-read-prefetch-algorithm](http://goto.google.com/improve-gcsfuse-read-prefetch-algorithm)

b/484168094

### Testing details
1. Manual - Done
2. Unit tests - Updated

### DIFF
--- a/internal/gcsx/client_readers/gcs_reader_test.go
+++ b/internal/gcsx/client_readers/gcs_reader_test.go
@@ -436,6 +436,13 @@ func (t *gcsReaderTest) Test_ReadAt_WithAndWithoutReadConfig() {
 
 // This test validates the bug fix where seeks are not updated correctly in case of zonal bucket random reads (b/410904634).
 func (t *gcsReaderTest) Test_ReadAt_ValidateZonalRandomReads() {
+	// Re-initialize GCSReader with initialOffset 13 MiB to force Random read type.
+	t.gcsReader = NewGCSReader(t.object, t.mockBucket, &GCSReaderConfig{
+		MetricHandle:       metrics.NewNoopMetrics(),
+		MrdWrapper:         nil,
+		Config:             nil,
+		ReadTypeClassifier: gcsx.NewReadTypeClassifier(int64(sequentialReadSizeInMb), 13*MiB),
+	})
 	t.gcsReader.rangeReader.reader = nil
 	t.gcsReader.mrr.isMRDInUse.Store(false)
 	t.object.Size = 20 * MiB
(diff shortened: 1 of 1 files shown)


internal/gcsx/read_type_classifier.go
```diff
@@ -25,6 +25,12 @@ const (
 	MB = 1 << 20
 )
 
+const (
```
raj-prince: What about storing the initial offset itself? We can initialize with -1 and change it only once

Or check if we can pass the initial value while constructing ReadClassifier.

internal/gcsx/read_type_classifier.go
```diff
@@ -116,22 +126,37 @@ func (rtc *ReadTypeClassifier) isSeekNeeded(offset int64) bool {
...
 	if !seekRecorded && rtc.isSeekNeeded(offset) {
 		numSeeks = rtc.seeks.Add(1)
 		seekRecorded = true
 	}
 
-	if numSeeks >= minSeeksForRandom {
-		readType = metrics.ReadTypeRandom
+	// Identify if the first read is at a non-zero offset.
+	state := rtc.firstReadState.Load()
+	if state == firstReadStateUnset {
```
raj-prince: Multiple concurrent go routine can pass this check in the start. Can we simplify?

internal/gcsx/read_type_classifier.go
```diff
@@ -154,17 +179,27 @@ func (rtc *ReadTypeClassifier) GetReadInfo(offset int64, seekRecorded bool) Read
...
 // handle that separately.
 func (rtc *ReadTypeClassifier) ComputeSeqPrefetchWindowAndAdjustType() int64 {
-	if seeks := rtc.seeks.Load(); seeks >= minSeeksForRandom {
+	currentReadType := rtc.readType.Load()
+	seeks := rtc.seeks.Load()
+
+	// Evaluate for Random read type if:
+	// 1. We have recorded seeks (seeks > 0).
+	// 2. OR we are already in Random mode because the first read was non-zero.
+	if seeks > 0 || rtc.firstReadState.Load() == firstReadStateNonZero {
```
raj-prince: So, averageBytes will be zero for the first non-zero offset read, and eventually this function will return 1MB.

Can we handle explicitly for seek = 0 scenario, as that will be more readable, given average calculation doesn't make sense with seek = 0?

internal/gcsx/read_type_classifier.go
vipnydav: Agreed, that would make it more readable. Updated it.

internal/gcsx/read_type_classifier.go
vipnydav: I used average calculation because, it already handled seek=0, but agreed that explictly handling would make it more readable. Updated it.

internal/gcsx/read_type_classifier.go
vipnydav: While multiple goroutines can pass this initial check when the value is still `-1`, only one will succeed in the subsequent `CompareAndSwap` operation. We have kept this check as a guard because `CompareAndSwap` is a relatively expensive operation compared to an atomic load. This pattern ensures that for all subsequent `ReadAt` calls (once the initial offset is initialized), we avoid the overhead of the `CompareAndSwap` logic entirely, optimizing the performance of the hot path. WDYT?

internal/gcsx/read_type_classifier.go
```diff
@@ -57,17 +57,22 @@ type ReadTypeClassifier struct {
 
 	// sequentialReadSizeMb is the configured sequential read size in MB.
 	sequentialReadSizeMb int64
+
+	// initialOffset tracks the offset of the first read.
```
raj-prince:
```suggestion
	// initialOffset stores the first read offset, helps in determining the initial prefetch size.
```

internal/gcsx/read_type_classifier.go
```diff
@@ -57,17 +57,22 @@ type ReadTypeClassifier struct {
...
+	initialOffset atomic.Int64
```
raj-prince: why atomic?

internal/gcsx/read_type_classifier.go
```diff
@@ -116,22 +121,25 @@ func (rtc *ReadTypeClassifier) isSeekNeeded(offset int64) bool {
...
 		numSeeks = rtc.seeks.Add(1)
 		seekRecorded = true
 	}
 
-	if numSeeks >= minSeeksForRandom {
-		readType = metrics.ReadTypeRandom
-	}
-	averageReadBytes := avgReadBytes(rtc.totalReadBytes.Load(), numSeeks)
+	// Identify if the first read is at a non-zero offset.
+	isFirstReadNonZero := rtc.initialOffset.Load() > 0
```
raj-prince: this boolean is not required, once we change the initialOffset to normal int.

internal/gcsx/read_type_classifier.go
vipnydav: We can remove atomic and make it int since it is initialized during creation of readmaanger and is thread safe. Updated it.

internal/gcsx/read_type_classifier.go
vipnydav: Removed it.

internal/gcsx/read_type_classifier.go
vipnydav: Moved the logic for `firstReadOffset` while creating `ReadTypeClassifier`.

raj-prince (dismissed):

PR #4406, opened 2026-02-18
feat(wire-log): plumb through wire log

geertj:
### Description

This adds a new flag, `-wire-log=<path>`, that enables the wire log that was recently added to jacobsa/fuse. When enabled, this will serialize every FUSE operation as a JSON object and append it to the specified file.

The wire log is useful when analyzing workload patterns.

### Testing details

The wirelog feature has unit tests in jacobsa/fuse. This merely plumbs through the flag. I have run this manually adding `-wire-log wirelog.json` and verified that the wire log is correctly generated.

None

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -621,6 +621,8 @@ type LoggingConfig struct {
 	LogRotate LogRotateLoggingConfig `yaml:"log-rotate"`
 
 	Severity LogSeverity `yaml:"severity"`
+
+	WireLog ResolvedPath `yaml:"wire-log"`
 }
 
 type MetadataCacheConfig struct {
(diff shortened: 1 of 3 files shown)


cmd/mount.go
```diff
@@ -189,6 +189,15 @@ func getFuseMountConfig(fsName string, newConfig *cfg.Config) *fuse.MountConfig
 		EnableAsyncReads: newConfig.FileSystem.EnableKernelReader,
 	}
 
+	if newConfig.Logging.WireLog != "" {
+		wireLog, err := os.Create(string(newConfig.Logging.WireLog))
```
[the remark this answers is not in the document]
geertj: The original code was intentional. The entires are appending during a run. When a new gcsfuse instance starts up, the wire log is cleared. The wire log's purpose is to serve as a debugging aid for a specific gcsfuse application interaction.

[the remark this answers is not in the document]
geertj: Furthermore, the life cycle of the wire log is tied to the life cycle of the mount. The file will be closed by when the gcsfuse process exits.

cfg/params.yaml
```diff
@@ -813,6 +813,13 @@ params:
     usage: "Specifies the logging severity expressed as one of [trace, debug, info, warning, error, off]"
     default: "info"
 
+  - config-path: "logging.wire-log"
```
raj-prince, line 842: Do we want to keep this flag public? Not sure, if people may confused with `logging.file-path`.

If we want to keep it private, then we need to add `hide-flag=true`.

abhishek10004: IMO, this should be a hidden flag only.

geertj: Updated

geertj:
@abhishek10004 I made the flag hidden, please take a look

cfg/config.go
```diff
@@ -1912,6 +1916,10 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	if err := v.BindPFlag("logging.wire-log", flagSet.Lookup("wire-log")); err != nil {
+		return err
```
abhishek10004, line 1924: Could you please run "make build" once? There should be code for flagSet.MarkHidden as well for this flag.

geertj: Ah, did not realize this was needed. Updated.

PR #4408, opened 2026-02-23
fix(Type cache Deprecation): Implicit Dir memory Consumption

Tulsishah:
### Description
This PR addresses a significant memory regression introduced during the Type Cache Deprecation project.

**Problem:** When the typeCache was deprecated, implicit directory information is moved into the unified statCache. However, each implicit directory entry was being stored as a full gcs.MinObject struct, which consumes approximately 1454 bytes per entry even if most fields are empty. In flat buckets with many objects (e.g., 319,000), this caused the cache to balloon to over 936.5 MiB, compared to the ~168 bytes per entry in the legacy typeCache.

**Solution:** This PR optimizes the internal entry struct in the statCache to handle implicit directories without allocating a full MinObject:

1. **Specialized Entry Type:** Introduced an implicitDir boolean flag to mark entries as implicit directories.
2. **Lazy Construction:** Implicit entries now store only the name. The MinObject is constructed on the fly during a LookUp if needed, reducing the memory footprint from ~1.4 KB to ~168-200 bytes per entry.
3. **Conflict Resolution:** Updated InsertImplicitDir to ensure it does not overwrite explicit metadata (files or explicit directories) which contains actual GCS generation data.
4. **Constant Adjustment:** Updated `AverageSizeOfPositiveStatCacheEntry` and the default `stat-cache-max-size-mb` (from 33 to 34 MiB) to account for the base struct size increase due to the new field and memory alignment.

#### Before Fix

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-23ec0f63-7fff-26a5-0628-c0941a6cd828"><div dir="ltr" style="margin-left:0pt;" align="left">
Cache Metric Per Entry | Before Deprecation | After Deprecation
-- | -- | --
Type Cache(files) | 168 bytes | 0 bytes
Type Cache (Implicit Dir) | 168 bytes | 0 bytes
Stat Cache (Files) | 1484 bytes | 1484 bytes
Stat Cache (Implicit Dir) | 0 bytes | 1454 bytes
Total diff in memory usage | ~ 1118 bytes increased after deprecation

</div></b>

#### After Fix

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-a71aa91a-7fff-2d7a-93ec-871d2d6660f1"><div dir="ltr" style="margin-left:0pt;" align="left">
Cache Metric Per Entry | Before Deprecation | After Deprecation
-- | -- | --
Type Cache(files) | 168 bytes | 0 bytes
Type Cache (Implicit Dir) | 168 bytes | 0 bytes
Stat Cache (Files) | 1484 bytes | 1484 bytes
Stat Cache (Implicit Dir) | 0 bytes | 198 bytes
Total diff in memory usage | ~ 138 bytes decreased after deprecation

</div></b>

[b/486527347](https://b.corp.google.com/issues/486527347)

### Testing details
1. Manual
7xDgJMfgwPpKu3p

3. Unit tests - Added

### DIFF
--- a/internal/cache/metadata/stat_cache.go
+++ b/internal/cache/metadata/stat_cache.go
@@ -206,11 +206,20 @@ func (sc *statCacheBucketView) InsertImplicitDir(objectName string, expiration t
 	// Is there already a better entry?
 	if existing := sc.sharedCache.LookUp(name); existing != nil {
 		e := existing.(entry)
-		// The ListObject function returns explicit directory objects in both the MinObject and CollapseRun lists.
-		// Since these objects are already captured in the MinObject collection, we no longer need to overwrite or re-add them during the CollapseRun iteration.
-		// If existing entry is a positive entry (m != nil), we prefer it over implicit directory
-		// because implicit directory is inferred and has Generation 0.
-		// Even if existing is old generation, it's explicit.
+		// The ListObjects response handles directories in two ways:
+		// 1. 'MinObject' returns explicit directory objects containing full metadata.
+		// 2. 'CollapseRun' generates placeholders for these same directories; if no
+		//    explicit object exists, it treats them as "implicit" (inferred).
+		//
+		// We attempt to create implicit directories for all entries in 'CollapseRun'.
+		// However, since 'ListObject' returns explicit directories in the 'MinObject'
+		// list as well, this could result in redundant implicit entries for
+		// every explicit directory already processed.
+		//
+		// To prevent this, we check if an entry with the same name already exists
+		// with non-nil metadata. If metadata is present, we skip the implicit
+		// creation to avoid overwriting a real, explicit object with an inferred
+		// placeholder (which would lack metadata and have 'Generation 0').
 		if e.m != nil {
 			return
 		}


internal/cache/metadata/stat_cache.go
```diff
@@ -195,6 +199,32 @@ func (sc *statCacheBucketView) Insert(m *gcs.MinObject, expiration time.Time) {
...
+	// Insert an entry.
+	e := entry{
+		implicitDir: true,
+		expiration:  expiration,
+		key:         name,
+	}
+
+	if _, err := sc.sharedCache.Insert(name, e); err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
Tulsishah: Let's not do it in this PR as other functions also follows same pattern.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -122,37 +122,24 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
...
 	// 2. Cache Explicit Objects
 	for _, o := range listing.MinObjects {
 		b.cache.Insert(o, expiration)
-		minObjectNames[o.Name] = struct{}{}
 	}
 
 	// 3. Cache Sub-directories (Collapsed Runs)
 	// These represent folders discovered via prefixes in the ListObjects response.
 	for _, p := range listing.CollapsedRuns {
-		// Skip if this name was already cached as an explicit object.
```
Tulsishah, line 151: For reviewer: We don't require this logic anymore as `InsertImplicitDir` handles this.

cfg/config.go
```diff
@@ -1320,13 +1320,13 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.IntP("stat-cache-capacity", "", 20460, "How many entries can the stat-cache hold (impacts memory consumption). This flag has been deprecated (starting v2.0) and in favor of stat-cache-max-size-mb. For now, the value of stat-cache-capacity will be translated to the next higher corresponding value of stat-cache-max-size-mb (assuming stat-cache entry-size ~= 1688 bytes, including 1448 for positive entry and 240 for corresponding negative entry), if stat-cache-max-size-mb is not set.\"")
```
vadlakondaswetha, line 1323: do we need to update the cloud docs for this change?

internal/cache/metadata/stat_cache.go
```diff
@@ -195,6 +199,32 @@ func (sc *statCacheBucketView) Insert(m *gcs.MinObject, expiration time.Time) {
 	}
 }
 
+func (sc *statCacheBucketView) InsertImplicitDir(objectName string, expiration time.Time) {
+	name := sc.key(objectName)
+
+	// Is there already a better entry?
+	if existing := sc.sharedCache.LookUp(name); existing != nil {
+		e := existing.(entry)
+		// If existing entry is a positive entry (m != nil), we prefer it over implicit directory
```
vadlakondaswetha: in what flow does this happen?

internal/cache/metadata/stat_cache.go
```diff
@@ -195,6 +199,32 @@ func (sc *statCacheBucketView) Insert(m *gcs.MinObject, expiration time.Time) {
...
+
+	// Insert an entry.
+	e := entry{
+		implicitDir: true,
+		expiration:  expiration,
+		key:         name,
+	}
+
+	if _, err := sc.sharedCache.Insert(name, e); err != nil {
+		panic(err)
```
vadlakondaswetha: why we want to panic if we can't insert into cache. just log and proceed?

Tulsishah: Thought it  but as [rest of the methods](https://github.com/GoogleCloudPlatform/gcsfuse/blob/0dbc7e3e4f20ac9e82f250225aa02c2bc7c35986/internal/cache/metadata/stat_cache.go#L194)are following same pattern, I didn't change it. May be we can fix them together in later PR.

internal/cache/metadata/stat_cache.go
Tulsishah: The ListObject function returns explicit directory objects in both the MinObject and CollapseRun lists. Since these objects are already captured in the MinObject collection, we no longer need to overwrite or re-add them during the CollapseRun iteration.

Previously, we handled this by [maintaining a separate list](https://github.com/GoogleCloudPlatform/gcsfuse/blob/0dbc7e3e4f20ac9e82f250225aa02c2bc7c35986/internal/storage/caching/fast_stat_bucket.go#L152) and checking for existing entries before insertion. However, that approach increased the memory footprint. This new implementation avoids that overhead.

cfg/config.go
Tulsishah: Yeah will raise separate bug ticket after this got merged.

internal/cache/metadata/stat_cache.go
vadlakondaswetha: can you please add this detail in the comment for future reference.

internal/cache/metadata/stat_cache.go
vadlakondaswetha: Atleast for the new methods, lets not panic

PR #4420, opened 2026-02-28
feat: Setting directConnectivityEnforced option for grpc

vadlakondaswetha:
Setting directConnectivityEnforced option while creating grpcClient. This header is used by server to decide whether to reject non direct path requests. Right now it will not reject but only tracks the non directpath calls.

### Testing details
1. Manual - done
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -54,8 +54,6 @@ const (
 
 	zonalLocationType         = "zone"
 	stallTimeoutForDirectPath = 60 * time.Second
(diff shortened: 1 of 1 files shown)


abhishek10004 (approved):
Approving with the assumption that server will wait for fallback to be implemented on GCSFuse side before rejecting requests.

vadlakondaswetha:
I will also do the fallback before the release. Hence merging this PR.

PR #4421, opened 2026-02-28
feat: added directpath strategy param

vadlakondaswetha:
Add a config for the user to configure what to do when directpath is not available for grpc

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cmd/config_validation_test.go
+++ b/cmd/config_validation_test.go
@@ -546,7 +546,7 @@ func TestValidateConfigFile_GCSConnectionConfigSuccessful(t *testing.T) {
 					ClientProtocol:             "http2",
 					CustomEndpoint:             "www.abc.com",
 					ExperimentalEnableJsonRead: true,
-					GrpcPathStrategy:           "direct-path-with-fallback",
+					GrpcPathStrategy:           "direct-path-only",
 					GrpcConnPoolSize:           200,
 					HttpClientTimeout:          400 * time.Second,
 					LimitBytesPerSec:           20,
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -809,14 +809,14 @@ func TestArgsParsing_GCSConnectionFlags(t *testing.T) {
 	}{
 		{
 			name: "Test gcs connection flags.",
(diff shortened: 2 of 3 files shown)


cmd/config_validation_test.go
```diff
@@ -545,6 +546,7 @@ func TestValidateConfigFile_GCSConnectionConfigSuccessful(t *testing.T) {
 					ClientProtocol:             "http2",
 					CustomEndpoint:             "www.abc.com",
 					ExperimentalEnableJsonRead: true,
+					GrpcPathStrategy:           "direct-path-with-fallback",
```
abhishek10004: Should we add a config to test manually setting this parameter (validate both options) in the config file?

cmd/root_test.go
```diff
@@ -816,6 +816,7 @@ func TestArgsParsing_GCSConnectionFlags(t *testing.T) {
 					ClientProtocol:             "http2",
 					CustomEndpoint:             "www.abc.com",
 					ExperimentalEnableJsonRead: true,
+					GrpcPathStrategy:           "direct-path-with-fallback",
```
abhishek10004: Similar to above, we should validate by setting the option manually (to both valid values)?

cfg/params.yaml
```diff
@@ -623,6 +623,15 @@ params:
     deprecated: true
     deprecation-warning: "Experimental flag: can be removed in a minor release."
 
+  - config-path: "gcs-connection.grpc-path-strategy"
+    flag-name: "grpc-path-strategy"
+    type: "directPathStrategy"
+    usage: >-
+      Strategy for DirectPath connectivity when client-protocol=grpc.
```
abhishek10004, line 630: What would the behavior for this parameter be with rapid buckets?

vadlakondaswetha: This is applicable only for non-rapid buckets for now i.e, when user explicitly sets the client-protocol=grpc and not rapid buckets.

cmd/config_validation_test.go
vadlakondaswetha: added

cmd/root_test.go
vadlakondaswetha: added

cfg/params.yaml
abhishek10004: Thanks. Should we mention in the description that this is for non-rapid only?

abhishek10004 (approved):
LGTM.

PR #4423, opened 2026-03-02
feat(consistent symlink): Flag for enabling new symbolic link behavior

anushka567:
### Description
This PR introduces the `enable-standard-symlinks `hidden flag to align GCSFuse's symbolic link representation with the GCS standard used by other GCS clients. When enabled, new symlinks will be  created as objects containing the target path as content and the goog-reserved-file-is-symlink: true metadata key

Marking it as default false until all changes are checked in and productionized.

[b/488905594](b/488905594)

### Testing details
1. Manual - Yes.
2. Unit tests - As part of PR.
3. Integration tests - Not needed.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1585,7 +1585,7 @@ func TestArgsParsing_EnableStandardSymlinks(t *testing.T) {
 		},
 		{
 			name:                           "normal",
-			args:                           []string{"gcsfuse", "--enable-standard-symlinks=true", "abc", "pqr"},
+			args:                           []string{"gcsfuse", "--experimental-enable-standard-symlinks=true", "abc", "pqr"},
 			expectedEnableStandardSymlinks: true,
 		},
 	}


cfg/config.go
```diff
@@ -1549,6 +1557,10 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	if err := v.BindPFlag("enable-standard-symlinks", flagSet.Lookup("enable-standard-symlinks")); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
anushka567: Auto generated code. Strict lexicographical order is not maintained in this code section.

cfg/config.go
```diff
@@ -443,6 +443,8 @@ type Config struct {
 
 	EnableNewReader bool `yaml:"enable-new-reader"`
 
+	EnableStandardSymlinks bool `yaml:"enable-standard-symlinks"`
```
vadlakondaswetha: we should name the flag as experimental when the feature is under development.

anushka567: Done.

ashmeenkaur:
relying on @vadlakondaswetha's review

PR #4424, opened 2026-03-02
feat(consistent symlink):  changes for CreateSymlink flow for updated representation

anushka567:
### Description
This PR implements the creation phase of the updated GCS symlink representation in GCSFuse to align with the standard used by other GCS clients like Storage Transfer Service (STS) . When enabled via a flag, new symlinks will be created as GCS objects where the target path is stored in the object content and marked with the standard metadata key `goog-reserved-file-is-symlink: true `.

Context: [go/sts-consistent-symlink-handling-in-gcsfuse](http://goto.google.com/sts-consistent-symlink-handling-in-gcsfuse)

[b/488952518](b/488952518)

### Testing details
1. Manual - Yes.
2. Unit tests - Included.

No.

### DIFF
--- a/internal/fs/inode/dir_test.go
+++ b/internal/fs/inode/dir_test.go
@@ -1568,6 +1568,64 @@ func (t *DirTest) TestCreateChildSymlink_TypeCaching() {
 	assert.Equal(t.T(), dirObjName, result.MinObject.Name)
 }
 
+func (t *DirTest) TestCreateChildSymlink_StandardSymlinkEnabled() {
+	// Re-create inode with standard symlinks enabled.
+	t.in.Unlock()
+	config := &cfg.Config{
+		List:                               cfg.ListConfig{EnableEmptyManagedFolders: true},
+		MetadataCache:                      cfg.MetadataCacheConfig{TypeCacheMaxSizeMb: 4},
+		EnableHns:                          false,
+		EnableUnsupportedPathSupport:       true,
+		EnableTypeCacheDeprecation:         isTypeCacheDeprecationEnabled,
+		ExperimentalEnableStandardSymlinks: true,
+	}
+	parInodeCtx := context.Background()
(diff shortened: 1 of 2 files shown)


internal/fs/inode/dir_test.go
```diff
@@ -1568,6 +1568,64 @@ func (t *DirTest) TestCreateChildSymlink_TypeCaching() {
...
+	// Check metadata for standard symlink
+	assert.Equal(t.T(), "true", result.MinObject.Metadata[StandardSymlinkMetadataKey])
+	// Check that the old metadata key is NOT present
+	_, ok := result.MinObject.Metadata[SymlinkMetadataKey]
+	assert.False(t.T(), ok)
+	// Check content for standard symlink (should be target)
+	content, err := storageutil.ReadObject(t.ctx, t.bucket, objName)
+	require.NoError(t.T(), err)
+	assert.Equal(t.T(), target, string(content))
+}
```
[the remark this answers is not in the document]
anushka567: Only validating creation behavior and that object created with symlink target as content. Actual operation behavior is validated (manually) and will be a part of the overall integration tests which is in a draft follow-up PR.

internal/fs/inode/symlink.go
```diff
@@ -27,6 +27,7 @@ import (
 // treated as a symlink. For use in testing only; other users should detect
 // this with IsSymlink.
 const SymlinkMetadataKey = "gcsfuse_symlink_target"
+const StandardSymlinkMetadataKey = "goog-reserved-file-is-symlink"
```
vadlakondaswetha, line 30: i hope you will add a comment here that symlinkMetadataKey is deprecated and isSymlink method will be updated in the follow up PRs

anushka567: Yes, it is the next PR in fact.

PR #4426, opened 2026-03-02
fix(Type Cache Deprecation): Implicit directories visible without flag being passed

Tulsishah:
### Description
This PR fixes a bug where implicit directories were incorrectly being cached and displayed even when the implicit-dirs flag was disabled. This issue surfaced following the implementation of the Type Cache Deprecation project, which merged typeCache functionality into the statCache.

[b/488971967](https://buganizer.corp.google.com/issues/488971967)

### Testing details
1. Manual - NA
2. Unit tests - Added

### DIFF
--- a/tools/integration_tests/implicit_dir/list_test.go
+++ b/tools/integration_tests/implicit_dir/list_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup/implicit_and_explicit_dir_setup"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestListImplicitObjectsFromBucket(t *testing.T) {
@@ -141,5 +142,28 @@ func TestListImplicitObjectsFromBucket(t *testing.T) {
 	if err != nil {
 		t.Errorf("error walking the path : %v\n", err)
 		return
+	}
+}
+
+func TestStatImplicitDirAfterList(t *testing.T) {
+	testDirPath := setup.SetupTestDirectory(DirForImplicitDirTests)
+	// TODO: Remove the condition and keep the storage-client flow for non-ZB too.
+	if setup.IsZonalBucketRun() {
+		implicit_and_explicit_dir_setup.CreateImplicitDirectoryStructureUsingStorageClient(testEnv.ctx, t, testEnv.storageClient, DirForImplicitDirTests)
+	} else {
+		implicit_and_explicit_dir_setup.CreateImplicitDirectoryStructure(DirForImplicitDirTests)
+	}
+
+	// List the directory
+	_, err := os.ReadDir(testDirPath)
+	if err != nil {
+		t.Fatalf("ReadDir failed: %v", err)
+	}
+
+	// Stat the implicit directory
+	implicitDirPath := path.Join(testDirPath, implicit_and_explicit_dir_setup.ImplicitDirectory)
+	f, err := os.Stat(implicitDirPath)
+	if assert.NoError(t, err) {
+		assert.True(t, f.IsDir())
 	}
 }


cmd/mount.go
```diff
@@ -108,6 +108,7 @@ be interacting with the file system.`)
 		DisableListAccessCheck:             newConfig.DisableListAccessCheck,
 		DummyIOCfg:                         newConfig.DummyIo,
 		IsTypeCacheDeprecated:              newConfig.EnableTypeCacheDeprecation,
+		ImplicitDir:                        newConfig.ImplicitDirs,
```
vadlakondaswetha, line 111: what is this config for? how is it different from enableTypeCacheDeprecation?

tools/integration_tests/explicit_dir/list_test.go
```diff
@@ -105,3 +107,28 @@ func TestListOnlyExplicitObjectsFromBucket(t *testing.T) {
 		return
 	}
 }
+
+func TestStatImplicitDirAfterList(t *testing.T) {
+	testDirPath := setup.SetupTestDirectory(DirForExplicitDirTests)
+	// TODO: Remove the condition and keep the storage-client flow for non-ZB too.
+	if setup.IsZonalBucketRun() {
```
vadlakondaswetha, line 114: Zonal buckets are hns. how are you creating implicit dirs in zb?

tools/integration_tests/implicit_dir/list_test.go
```diff
@@ -143,3 +144,26 @@ func TestListImplicitObjectsFromBucket(t *testing.T) {
 		return
 	}
 }
+
+func TestStatImplicitDirAfterList(t *testing.T) {
+	testDirPath := setup.SetupTestDirectory(DirForImplicitDirTests)
+	// TODO: Remove the condition and keep the storage-client flow for non-ZB too.
+	if setup.IsZonalBucketRun() {
+		implicit_and_explicit_dir_setup.CreateImplicitDirectoryStructureUsingStorageClient(testEnv.ctx, t, testEnv.storageClient, DirForImplicitDirTests)
```
vadlakondaswetha, line 152: same here

cmd/mount.go
Tulsishah: enableTypeCacheDeprecation flag is hidden flag to manage type cache related changes, which we will deprecate in future. ImplicitDir is for logic to cache implicit-dirs only if flag is passed otherwise ignore.

tools/integration_tests/explicit_dir/list_test.go
Tulsishah: This logic was originally [implemented](https://github.com/GoogleCloudPlatform/gcsfuse/blob/e842b6c70481049c3774a1fb1e29efdfccb2d09d/tools/integration_tests/explicit_dir/list_test.go#L44) by @gargnitingoogle. These common tests run across all three bucket types. For Zonal buckets, we provide [additional arguments](https://github.com/GoogleCloudPlatform/gcsfuse/blob/553cec17c7203665ec067c567ce05b3965353744/tools/integration_tests/util/client/storage_client.go#L243) during client creation, which likely necessitates this specific setup.

The intent is to ensure that HNS and Zonal buckets run these tests regardless of the implicit-dir flag. Because their directory structures are functionally identical in this context, we treat HNS and Zonal buckets as having 'implicit directory' behavior by default.

tools/integration_tests/implicit_dir/list_test.go
Tulsishah: This logic was originally [implemented](https://github.com/GoogleCloudPlatform/gcsfuse/blob/e842b6c70481049c3774a1fb1e29efdfccb2d09d/tools/integration_tests/explicit_dir/list_test.go#L44) by @gargnitingoogle. These common tests run across all three bucket types. For Zonal buckets, we provide [additional arguments](https://github.com/GoogleCloudPlatform/gcsfuse/blob/553cec17c7203665ec067c567ce05b3965353744/tools/integration_tests/util/client/storage_client.go#L243) during client creation, which likely necessitates this specific setup.

The intent is to ensure that HNS and Zonal buckets run these tests regardless of the implicit-dir flag. Because their directory structures are functionally identical in this context, we treat HNS and Zonal buckets as having 'implicit directory' behavior by default.

PR #4428, opened 2026-03-03
feat(consistent symlink): modify the IsSymlink method to identify both legacy as well as standard representation

anushka567:
### Description
This PR updates the IsSymlink method in GCSFuse to recognize both the legacy GCSFuse-specific symlink representation and the standard GCS symlink representation used by clients like Storage Transfer Service (STS) 

Context: [go/sts-consistent-symlink-handling-in-gcsfuse](http://goto.google.com/sts-consistent-symlink-handling-in-gcsfuse)

[b/489240501](b/489240501)

### Testing details
1. Manual - Yes
2. Unit tests - Added.
3. Integration tests - NA

No

### DIFF
--- a/internal/fs/inode/symlink.go
+++ b/internal/fs/inode/symlink.go
@@ -26,8 +26,8 @@ import (
 // When this custom metadata key is present in an object record, it is to be
 // treated as a symlink. For use in testing only; other users should detect
 // this with IsSymlink.
-// Note: SymlinkMetadataKey is deprecated and retained solely for backward
-// compatibility.
+// Note: SymlinkMetadataKey is deprecated in favor of StandardSymlinkMetadataKey
+// and retained solely for backward compatibility.
 const SymlinkMetadataKey = "gcsfuse_symlink_target"
 const StandardSymlinkMetadataKey = "goog-reserved-file-is-symlink"
 


internal/fs/inode/symlink.go
```diff
@@ -26,6 +26,8 @@ import (
 // When this custom metadata key is present in an object record, it is to be
 // treated as a symlink. For use in testing only; other users should detect
 // this with IsSymlink.
+// Note: SymlinkMetadataKey is deprecated and retained solely for backward
```
vadlakondaswetha: SymlinkMetadataKey is deprecated in favor of StandardSymlinkMetadataKey

internal/fs/inode/symlink.go
```diff
@@ -34,9 +36,15 @@ func IsSymlink(m *gcs.MinObject) bool {
...
-
-	_, ok := m.Metadata[SymlinkMetadataKey]
-	return ok
+	// 1. Check legacy/custom key presence
+	if _, ok := m.Metadata[SymlinkMetadataKey]; ok {
+		return true
+	}
+	// 2. Check standard reserved key value
+	if val, ok := m.Metadata[StandardSymlinkMetadataKey]; ok {
+		return val == "true"
```
vadlakondaswetha, line 45: does STS also uses small case? what happens when we use capital case?

anushka567: STS uses small case, that is what is declared explicitly in the official documentation ([ref](https://docs.google.com/document/d/1dWjeR9uRDYF7vfM5b54TYnvO3rNdAsrGBUOT-aOuOLg/edit?resourcekey=0-ic5syINaeA_5qvKZDQ1Y5A&tab=t.0#bookmark=id.fpqqoitw7wsj)). 

If the user however tries to explicitly pass capital/mixed, it will not be treated as a symlink by GCSFuse. This case is also not feasible since STS and other clients always ensure that creation of symlink will result in `goog-reserved-file-is-symlink: true` and even GCSFuse will follow the same.

anushka567: Validated that same behavior is observed with other clients.

gcloud with preserve symlinks set for GCS object with `goog-reserved-file-is-symlink: TRUE`, does not consider it as a symlink.

PR #4431, opened 2026-03-03
feat(consistent symlink): update symlink inode struct to include metadata for direct references

anushka567:
### Description
This PR updates the SymlinkInode struct to store the complete metadata of the backing GCS object and ensures that the Source() method returns the full metadata instead of a manually constructed map

Context: 
[go/sts-consistent-symlink-handling-in-gcsfuse](http://goto.google.com/sts-consistent-symlink-handling-in-gcsfuse)

[b/489239761](b/489239761)

### Testing details
1. Manual - Yes.
2. Unit tests - Added.
3. Integration tests - NA

No

### DIFF
--- a/internal/fs/inode/symlink_test.go
+++ b/internal/fs/inode/symlink_test.go
@@ -167,7 +167,7 @@ func (t *SymlinkTest) TestSource() {
 		MetaGeneration: 2,
 		Size:           100,
 		Metadata:       map[string]string{inode.StandardSymlinkMetadataKey: "true"},
-		Updated:        time.Now(),
+		Updated:        time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC),
 	}
 	attrs := fuseops.InodeAttributes{}
 	name := inode.NewFileName(inode.NewRootName("some-bucket"), m.Name)


internal/fs/inode/symlink.go
```diff
@@ -99,7 +100,8 @@ func NewSymlinkInode(
 			Ctime: m.Updated,
 			Mtime: m.Updated,
 		},
-		target: m.Metadata[SymlinkMetadataKey],
+		target:   m.Metadata[SymlinkMetadataKey],
+		metadata: m.Metadata,
```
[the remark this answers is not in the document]
anushka567: This change is in a follow-up PR ([draft changes](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4435)) .

internal/fs/inode/symlink.go
```diff
@@ -99,7 +100,8 @@ func NewSymlinkInode(
 			Ctime: m.Updated,
 			Mtime: m.Updated,
 		},
-		target: m.Metadata[SymlinkMetadataKey],
+		target:   m.Metadata[SymlinkMetadataKey],
```
vadlakondaswetha, line 103: I am assuming target field will be removed in the follow up prs

anushka567: Yes, that is correct.

PR #4439, opened 2026-03-03
feat: Added a stat call to check if dp is working and fallback when it doesn't work

vadlakondaswetha:
After creating the grpcClient we will make a dummy stat call.  When the call fails with any error other than notFound path, that means server is rejecting the cloud path rejects. In that case, we will fallback to http if the user chooses that option.

### Testing details
1. Manual - Done
2. Unit tests - Not possible to write as the client creation is coupled with the stat call. FakeServer doesn't support grpc today. Will add integration tests to verify
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/concurrent_operations/concurrent_listing_test.go
+++ b/tools/integration_tests/concurrent_operations/concurrent_listing_test.go
@@ -542,15 +542,15 @@ func (s *concurrentListingTest) Test_ListWithMoveFile() {
 		}
 	}()
 
-	// Create file
-	err := os.WriteFile(path.Join(testDirPath, "move_file.txt"), []byte("Hello, world!"), setup.FilePermission_0600)
-	require.NoError(s.T(), err)
-
 	// Goroutine 2: Move file
 	go func() {
 		defer wg.Done()
 		for range iterationsForHeavyOperations { // Adjust iteration count if needed
 			fileName := setup.GenerateRandomString(5) + "move_file.txt"
+			// Create file
+			err := os.WriteFile(path.Join(testDirPath, fileName), []byte("Hello, world!"), setup.FilePermission_0600)
+			require.NoError(s.T(), err)
+
 			// Move File in the target directory
 			err = operations.Move(path.Join(testDirPath, fileName), path.Join(targetDir, fileName))
 			require.NoError(s.T(), err)
@@ -599,15 +599,16 @@ func (s *concurrentListingTest) Test_ListWithMoveDir() {
 			require.NoError(s.T(), f.Close())
 		}
 	}()
-	// Create Dir
-	err := os.Mkdir(path.Join(testDirPath, "move_dir"), setup.DirPermission_0755)
-	require.NoError(s.T(), err)
 
 	// Goroutine 2: Move Dir
 	go func() {
 		defer wg.Done()
 		for range iterationsForHeavyOperations { // Adjust iteration count if needed
 			dirName := setup.GenerateRandomString(5) + "move_dir"
+			// Create Dir
+			err := os.Mkdir(path.Join(testDirPath, dirName), setup.DirPermission_0755)
+			require.NoError(s.T(), err)
+
 			// Move Dir in the target dir
 			err = operations.Move(path.Join(testDirPath, dirName), path.Join(targetDir, dirName))
 			require.NoError(s.T(), err)


internal/storage/storage_handle.go
```diff
@@ -403,6 +446,32 @@ func (sh *storageClient) getClient(ctx context.Context, isbucketZonal bool) (*st
 	return nil, fmt.Errorf("invalid client-protocol requested: %s", sh.clientConfig.ClientProtocol)
 }
 
+func (sh *storageClient) createNonBidiGRPCClientWithHttpFallback(ctx context.Context, bucketName string) (*storage.Client, error) {
```
abhishek10004, line 466: Why are we calling this createNonBidi* and not createGRPCClientWithHttpFallback? We can take the enableBidi config as a param which can be false for now but can be changed to true in future?

internal/storage/storage_handle.go
```diff
@@ -195,9 +186,28 @@ func setRetryConfig(ctx context.Context, sc *storage.Client, clientConfig *stora
 	}
 }
 
-// Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
-func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig, enableBidiConfig bool) (sc *storage.Client, err error) {
+// applyDPDetectionRetryConfig applies a lenient retry configuration for DirectPath detection phase.
+// This config is designed to fail fast during the initial connection check.
+func applyDPDetectionRetryConfig(ctx context.Context, sc *storage.Client, clientConfig *storageutil.StorageClientConfig) {
```
abhishek10004: since we already have setRetryConfig, should we also call this setDPDetectionRetryConfig?

internal/storage/storage_handle.go
```diff
@@ -76,22 +83,6 @@ type storageClient struct {
...
-
-// isDirectPathPossible checks if gRPC direct connectivity is available for a specific bucket
-// from the environment where the client is running. A `nil` error represents Direct Connectivity was
-// detected.
-func (pd *gRPCDirectPathDetector) isDirectPathPossible(ctx context.Context, bucketName string) error {
-	newCtx, cancel := context.WithTimeout(ctx, stallTimeoutForDirectPath)
-	defer cancel()
-
-	// The storage library will see the timeout in 'newCtx' and abort the request if it takes too long.
-	return storage.CheckDirectConnectivitySupported(newCtx, bucketName, pd.clientOptions...)
```
abhishek10004, line 94: Was it not possible to use this for non zonal grpc as well?

internal/storage/storage_handle.go
```diff
@@ -436,22 +505,11 @@ func (sh *storageClient) BucketHandle(ctx context.Context, bucketName string, bi
 		return nil, fmt.Errorf("storageLayout call failed: %s", err)
 	}
 
-	client, err = sh.getClient(ctx, bucketType.Zonal)
+	client, err = sh.getClient(ctx, bucketType.Zonal, bucketName)
 	if err != nil {
 		return nil, err
 	}
 
-	if bucketType.Zonal || sh.clientConfig.ClientProtocol == cfg.GRPC {
```
abhishek10004, line 444: Do we not want to have any direct path check for zonal at all? While this is not very useful now, it was helpful during the initial phase.

internal/storage/storage_handle.go
vadlakondaswetha: Agreed. For now i am just restricting it to the current scenario - Zonal buckets uses bidi without fallback and non ZB uses non bidi with fallback. We can make it generic when need arises. Its a trivial change.

internal/storage/storage_handle.go
vadlakondaswetha: Not sure. It was removed in the Prince PR, so removed it. I will revert it for now and will discuss it when he is back.

internal/storage/storage_handle.go
vadlakondaswetha: same as other comment. REverting it for now.

internal/storage/storage_handle_test.go
```diff
@@ -339,21 +340,21 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleWithInvalidClientProtoco
 	assert.Contains(testSuite.T(), err.Error(), "invalid client-protocol requested: test-protocol")
 }
 
-func (testSuite *StorageHandleTest) TestNewStorageHandleDirectPathDetector() {
```
abhishek10004, line 342: Since we're not removing the directPathDetector code for now, we should keep the original tests as well.

internal/storage/storage_handle.go
abhishek10004, line 532: Do we want this check for non zonal gRPC protocol as well with the fallback logic in place? (we can have http client due to fallback and then doing this check won't make sense and can create confusion)

internal/storage/storage_handle.go
```diff
@@ -403,6 +462,32 @@ func (sh *storageClient) getClient(ctx context.Context, isbucketZonal bool) (*st
...
+
+	// We will reach here when we failed to create a grpc client with direct path.
+	// Decide whether to create a http client based on grpPathStrategy param.
+	if sh.clientConfig.GrpcPathStrategy == cfg.DirectPathOnly {
+		return nil, err
+	}
+
+	// When grpcPathStrategy=DirectPathWithFallback, create a http client.
+	if sh.httpClient == nil {
+		sh.httpClient, err = createHTTPClientHandle(ctx, &sh.clientConfig)
```
abhishek10004, line 488: Should we add an info log for fallback?

internal/storage/storage_handle.go
vadlakondaswetha: lets not touch this until we decide on what to do with direct path detector - should be removed completely or not.

internal/storage/storage_handle_test.go
```diff
@@ -376,24 +377,16 @@ func (testSuite *StorageHandleTest) TestNewStorageHandleDirectPathDetector() {
 	}
 }
 
-func (testSuite *StorageHandleTest) TestCreateGRPCClientHandle() {
-	sc := storageutil.GetDefaultStorageClientConfig(keyFile)
```
abhishek10004, line 380: Why are we removing these 2 tests?

vadlakondaswetha: They are failing as we also stat call after grpc client creation. Since grpc client is created as part of the method, we can't mock it and fakeserver doesn't support grpc.

PR #4453, opened 2026-03-09
test(fs): add unit tests for gRPC metrics

alleaditya:
### Description

This PR adds comprehensive unit-level test coverage for gRPC metrics in GCSFuse, fulfilling the need to formally cross-verify gRPC trace emissions against the respective active filesystem and GCS operations. 

In this addition, we:
* Emitted explicit metrics assertions ensuring traces execute across active states:
  * Regular Object Access (LookUpInode / ReadFile).
  * File Cache Miss scenarios where a GCS reader is queried directly.
  * File Write paths (`CreateFile` & `SyncFile`).
  * Pick failure simulations (`StatObject` failures).

[493422480](https://b.corp.google.com/issues/493422480)

### Testing details
1. **Manual** - NA
2. **Unit tests** - 
   - `go test ./internal/fs -v -run TestGrpcMetrics` validates assertions specifically returning `PASS` across configurations.
3. **Integration tests** - `make test` runs smoothly with no regression impacts. Existing tests in gcs_metrics_test.go and metrics_test.go execute safely without pollution.

None.

### DIFF
--- a/internal/fs/grpc_metrics_test.go
+++ b/internal/fs/grpc_metrics_test.go
@@ -340,9 +340,6 @@ func createTestFileSystemWithGrpcMetrics(ctx context.Context, t *testing.T, para
 
 	if params.enableFileCache || params.enableSparseFileCache {
 		cacheDir := t.TempDir()
-		t.Cleanup(func() {
-			os.RemoveAll(cacheDir)
-		})
 		serverCfg.NewConfig.CacheDir = cfg.ResolvedPath(cacheDir)
 		serverCfg.NewConfig.FileCache = cfg.FileCacheConfig{
 			MaxSizeMb:                              100,


internal/fs/grpc_metrics_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+func (m *mockGrpcBucket) emitGrpcMetric(ctx context.Context, method string, isFailed bool) {
+	attr := otelmetric.WithAttributes(attribute.String("grpc_method", method))
+	m.started.Add(ctx, 1, attr)
+	m.callDur.Record(ctx, 1, attr)
+	m.defaultPicks.Add(ctx, 1, attr)
+	if isFailed {
+		m.failedPicks.Add(ctx, 1, attr)
+	}
+}
```
kislaykishore: Let's not mock the metrics generation because that's the functionality we want to test.

alleaditya: Removed this part

internal/fs/grpc_metrics_test.go
```diff
@@ -0,0 +1,594 @@
...
+					for _, kv := range subsetAttrs {
+						val, ok := dp.Attributes.Value(kv.Key)
+						if !ok || val.Emit() != kv.Value.Emit() {
+							matches = false
+							break
+						}
+					}
+					if matches {
+						foundDataPoint = true
+						assert.GreaterOrEqual(t, dp.Value, expectedMinLimit, "metric value too low for metric: %s", metricName)
```
kislaykishore: Why can't we do an exact match?

internal/fs/grpc_metrics_test.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+	// SyncFile might fail if BidiWriteObject is unimplemented, but we check metrics.
+	_ = server.SyncFile(ctx, syncOp)
+
+	waitForMetricsProcessing()
+	time.Sleep(501 * time.Millisecond)
+
+	// Verify BidiWriteObject metric.
+	metrics.VerifyCounterMetric(t, ctx, reader, "grpc.client.attempt.started",
+		attribute.NewSet(attribute.String("grpc.method", "google.storage.v2.Storage/BidiWriteObject")),
```
kislaykishore, line 489: Could you check whether BidiWriteObject is actually used in GCSFuse?

internal/fs/grpc_metrics_test.go
```diff
@@ -0,0 +1,494 @@
...
+		1, metrics.AtLeast(), metrics.Subset())
+	metrics.VerifyHistogramMetric(t, ctx, reader, "grpc.client.call.duration",
```
kislaykishore, line 491: Overall, I feel that the tests are much more complicated than what they should be. Let's not block the gRPC metrics enable-by-default PR due to this. But let's discuss how to simplify the tests.

kislaykishore: On second thought, while it's complicated, I guess it's a bullet that we have to bite.

kislaykishore (approved):

internal/fs/grpc_metrics_test.go
kislaykishore: Discussed offline - the set of labels can change say in case of retries. So, we only test the subset of labels.

internal/fs/grpc_metrics_test.go
alleaditya: It indeed is. This was the fix for previous [pipeline failures](https://screenshot.googleplex.com/9DmssUSt7ZRoaCr) after enabling default gRPC metrics.

The change ensures a concrete MeterProvider is always passed to the storage client when gRPC metrics are enabled, or explicitly disabled if no provider is found. This prevents the "input/output error" caused by the storage SDK's default handling of bidirectional streams.

alleaditya: Additionally I added a new comprehensive test case to storage_handle_test.go verifying the correct gRPC metrics configuration across all scenarios (GKE vs non-GKE, provider found vs not found).

alleaditya: I was able to test this and replicate it when testing append on zonal buckets and found the following metrics from the exporter. (test script - [5519880049786880](https://paste.googleplex.com/5519880049786880))

```
grpc_client_call_duration_count{grpc_method="google.storage.v2.Storage/BidiWriteObject",grpc_status="OK",grpc_target="google-c2p:///storage.googleapis.com"} 2
```

Context:

The GCS gRPC API uses BidiWriteObject (Bidirectional Write) for advanced write operations. In GCSFuse, this is primarily triggered in two scenarios:

* [Verified] **Zonal Buckets**: GCSFuse sets the Append flag on the storage writer for zonal buckets (
bucket_handle.go:L200
). The Go SDK uses BidiWriteObject to handle these appends.
* [Not verified] **Modern gRPC Writes**: Newer versions of the Go Storage SDK may prefer BidiWriteObject over the standard WriteObject for certain flushing or synchronization behaviors when using the gRPC protocol.

PR #4468, opened 2026-03-11
feat(metrics): enable gRPC metrics by default in configuration

alleaditya:
### Description
Enable gRPC metrics by default. The changes include test fixes that were affected by this change.

[442903788](https://b.corp.google.com/issues/442903788)

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - NA

No


cfg/config.go
```diff
@@ -966,7 +966,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.BoolP("experimental-enable-grpc-metrics", "", false, "Enables support for gRPC metrics")
+	flagSet.BoolP("experimental-enable-grpc-metrics", "", true, "Enables support for gRPC metrics")
```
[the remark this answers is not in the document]
alleaditya: Continuing to keep the flag in experimental phase.

kislaykishore (dismissed):
Let's submit this once the presubmits succeed.

kislaykishore:
One of the non-ZB tests apparently failed. Could you check?

internal/storage/storage_handle.go
```diff
@@ -157,15 +157,17 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
 	clientOpts = append(clientOpts, option.WithGRPCConnectionPool(clientConfig.GrpcConnPoolSize))
 	clientOpts = append(clientOpts, option.WithUserAgent(clientConfig.UserAgent))
 
-	if clientConfig.EnableGrpcMetrics && clientConfig.IsGKE {
+	if clientConfig.EnableGrpcMetrics {
```
kislaykishore: Is this intended?

kislaykishore (changes requested):
Seems to me that this PR has been mistakenly corrupted. Please check.

internal/storage/storage_handle.go
alleaditya: It appears that the previous zonal bucket write test [pipeline failures](https://screenshot.googleplex.com/9DmssUSt7ZRoaCr) were resulting from `internal codes.Internal error during the close operation of a bidirectional gRPC stream` reported as input/output error (EIO).

#### Existing configuration:
On non-GKE platforms (like a standard Compute Engine instance or local machine), even if gRPC metrics were enabled, no MeterProvider was explicitly passed.

#### Further investigation (from Gemini)
In such scenario, the Go storage SDK falls back to its default internal metrics instrumentation. In the SDK version currently used (v1.61.2), "there is a known issue (or at least an incompatibility with gcsfuse's lifecycle)" where this default instrumentation causes an internal codes.Internal error during the close operation of a bidirectional gRPC stream (like BidiWriteObject, which is used for writes to zonal buckets). This error propagates back to FUSE as an input/output error (EIO).

The solution was to pass the metric provider, which seems correct either way. A failure occurred when the metric provider wasn't passed, but with default gRPC metrics enabled, passing the MetricProvider is valid.

Alternatively I am highlighting this to go sdk. Meanwhile the changes seem intended.

alleaditya: We are not seeing the `input / output` errors now even with the changes reverted.

internal/storage/storage_handle.go
```diff
@@ -164,6 +164,8 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
 		if sdkmp, ok := mp.(*sdkmetric.MeterProvider); ok {
 			// pass in if sdkmp is of type *sdkmetric.MeterProvider (not a No-op)
 			clientOpts = append(clientOpts, experimental.WithMeterProvider(sdkmp))
+		} else {
+			clientOpts = append(clientOpts, storage.WithDisabledClientMetrics())
```
kislaykishore: Why is this needed?

alleaditya: Definitely an unwanted intruder change. Removed it.

kislaykishore (approved):

PR #4473, opened 2026-03-12
feat(chunkRetryDeadline): Propagate hidden chunkRetryDeadline flag and use new default for chunkRetryDeadline

meet2mky:
### Description
The PR aims to provide better control over how GCSFuse handles retries for chunk uploads, particularly during write stalls. It increases the default chunkRetryDeadline from 32 seconds to 120 seconds, allowing more time for successful retries of temporary failures (like 503 errors or potential stalls due to lock contention during high scale checkpoint workloads)
Changes:
* It propagates the hidden chunk-retry-deadline-secs flag to ensure the configurable deadline is respected throughout the object writing process.
* The default value for the chunk retry deadline is moved to 120 seconds to improve robustness.
* The PR adds documentation notes for gRPC object writers clarifying that chunkRetryDeadline and chunkTransferTimeout do not apply there.
* New integration tests have been added to verify the behavior of the chunk-retry-deadline-secs flag during write stalls, along with a helper function to parse the flag.
* Several constants related to chunk retries and timeouts in inode tests were unexported to improve the codebase's internal structure.

b/491635218

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - Part of pre-submits

None

### DIFF
--- a/internal/fs/inode/symlink_test.go
+++ b/internal/fs/inode/symlink_test.go
@@ -47,7 +47,10 @@ var _ TearDownInterface = &CoreTest{}
 func init() { RegisterTestSuite(&SymlinkTest{}) }
 
 func (t *SymlinkTest) SetUp(ti *TestInfo) {
-	bucket := gcsx.NewSyncerBucket( /*appendThreshold=*/ 1 /*chunkRetryDeadlineSecs=*/, 120 /*chunkTransferTimeoutSecs=*/, 10,
+	bucket := gcsx.NewSyncerBucket(
+		/*appendThreshold=*/ 1,
+		/*chunkRetryDeadlineSecs=*/ 120,
+		/*chunkTransferTimeoutSecs=*/ 10,
 		".gcsfuse_tmp/", fake.NewFakeBucket(timeutil.RealClock(), "some-bucket", gcs.BucketType{}))
 	t.bucket = &bucket
 }


internal/fs/handle/file_test.go
```diff
@@ -697,7 +697,7 @@ func (t *fileTest) Test_ReadWithMrdKernelReader_Success() {
 
 func (t *fileTest) Test_ReadWithMrdKernelReader_NotAuthoritative() {
 	// 1. Setup
-	zonalBucket := gcsx.NewSyncerBucket(1, 10, ".gcsfuse_tmp/", fake.NewFakeBucket(&t.clock, "zonal_bucket", gcs.BucketType{Zonal: true}))
+	zonalBucket := gcsx.NewSyncerBucket(1 /* appendThreshold */, 120 /* chunkRetryDeadlineSecs */, 10 /* chunkTransferTimeoutSecs */, ".gcsfuse_tmp/", fake.NewFakeBucket(&t.clock, "zonal_bucket", gcs.BucketType{Zonal: true}))
```
vadlakondaswetha: the comment here looks odd. please see what is the recommended way for golang

meet2mky: golang doesn't have any rules regarding these comments but other languages do have and I have changed it similar to C++ convention as per go/c-readability-advice#comment-argument-literals-with-their-parameter-names and it improves readability for these tests. Alternate option is to define constants for all which seems to be more work and not immediately clear what value the test is operating on. Similar advice is present here go/bugpattern/ParameterName#the-problem

internal/fs/inode/symlink_test.go
```diff
@@ -47,12 +47,8 @@ var _ TearDownInterface = &CoreTest{}
 func init() { RegisterTestSuite(&SymlinkTest{}) }
 
 func (t *SymlinkTest) SetUp(ti *TestInfo) {
-	bucket := gcsx.NewSyncerBucket(
-		1,
-		10, // ChunkTransferTimeoutSecs
-		".gcsfuse_tmp/",
-		fake.NewFakeBucket(timeutil.RealClock(), "some-bucket", gcs.BucketType{}),
-	)
+	bucket := gcsx.NewSyncerBucket( /*appendThreshold=*/ 1 /*chunkRetryDeadlineSecs=*/, 120 /*chunkTransferTimeoutSecs=*/, 10,
```
raj-prince: Nit: Could you please make sure, [comment][value],

internal/fs/gcs_metrics_test.go
```diff
@@ -57,7 +57,7 @@ func (bm *fakeBucketManagerWithMetrics) SetUpBucket(
 
 	// Wrap bucket with monitor.NewMonitoringBucket to enable GCS metrics.
 	sb = gcsx.NewSyncerBucket(
-		0, 10, ".gcsfuse_tmp/",
+		/*appendThreshold=*/ 0 /*chunkRetryDeadlineSecs=*/, 120 /*chunkTransferTimeoutSecs=*/, 10, ".gcsfuse_tmp/",
```
raj-prince: nit: here also

internal/fs/inode/symlink_test.go
meet2mky: Thanks for catching this. Done.

internal/fs/gcs_metrics_test.go
meet2mky: Thanks for catching this. Done.

PR #4480, opened 2026-03-16
test: add proper cleanups to fix flaky file cache tests

vipnydav:
### Description
This PR introduces robust resource cleanup logic across the file cache and storage unit test suites to resolve long-standing flakiness.

Following recent upgrades to the Google Cloud Go SDK and the fake-gcs-server, several tests began failing due to inter-test interference. The primary issues addressed are:

- **Leaked Background Goroutines**: Background download jobs were not being cancelled, allowing them to persist and compete for resources with subsequent tests.
- **Stale Mock Expectations**: Mock storage clients were retaining "Expected Calls" from previous tests, leading to immediate failures in new test environments.
- **Unclosed Server Handles**: Failure to shut down the fake GCS server between tests led to leaked storage handles and inconsistent state.

b/493137434

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/bucket_handle_test.go
+++ b/internal/storage/bucket_handle_test.go
@@ -64,8 +64,11 @@ func minObjectsToMinObjectNames(minObjects []*gcs.MinObject) (objectNames []stri
 
 func createBucketHandle(testSuite *BucketHandleTest, resp *controlpb.StorageLayout) {
 	var err error
-	testSuite.mockClient.ExpectedCalls = nil
-	testSuite.mockClient.Calls = nil
+
+	testSuite.mockClient = new(MockStorageControlClient)
+	testSuite.fakeStorage = NewFakeStorageWithMockClient(testSuite.mockClient, cfg.HTTP2)
+	testSuite.storageHandle = testSuite.fakeStorage.CreateStorageHandle()
+
 	testSuite.mockClient.On("GetStorageLayout", mock.Anything, mock.Anything, mock.Anything).
 		Return(resp, nil)
 	testSuite.bucketHandle, err = testSuite.storageHandle.BucketHandle(context.Background(), TestBucketName, "", false)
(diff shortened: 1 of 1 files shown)


internal/storage/bucket_handle_test.go
```diff
@@ -64,6 +64,8 @@ func minObjectsToMinObjectNames(minObjects []*gcs.MinObject) (objectNames []stri
 
 func createBucketHandle(testSuite *BucketHandleTest, resp *controlpb.StorageLayout) {
 	var err error
+	testSuite.mockClient.ExpectedCalls = nil
```
raj-prince: why are you coupling with createBucketHandle? - shouldn't we reset it as part of TearDownTest or SetupTest separately?

internal/storage/bucket_handle_test.go
```diff
@@ -1473,6 +1457,8 @@ func (testSuite *BucketHandleTest) TestBucketHandleWithError() {
 	var x *controlpb.StorageLayout
 	var err error
 	// Test when the client returns an error.
+	testSuite.mockClient.ExpectedCalls = nil
```
raj-prince: Seems, you are resetting again after the common createBucketHandle call as part of SetupTest method. What's the issue in calling createBucketHandle separately in each test and reset (setting nil) as part of common Setup or TearDown?

internal/storage/bucket_handle_test.go
vipnydav: I've removed the mock resets (ExpectedCalls = nil) entirely. Instead of trying to wipe old state, we now cleanly instantiate a fresh mock and fake storage from scratch to avoid any coupling.

internal/storage/bucket_handle_test.go
vipnydav: Agreed, Earlier I was trying to follow DRY approach but this is a much cleaner approach. I have updated the code to follow exactly this pattern, SetupTest strictly initializes fields to nil, TearDownTest handles cleanup, and createBucketHandle is now called explicitly at the start of each individual test.

PR #4492, opened 2026-03-18
fix: correct folder memory size calculation

kislaykishore:
### Description

In `stat_cache.go`, the memory footprint of `gcs.Folder` is now correctly computed by introducing a new `NestedSizeOfGcsFolder` function, tested in `sizeof_test.go`. Previously, `util.UnsafeSizeOf` was used on a pointer, resulting in only 8 bytes being counted instead of the full struct and its dynamically allocated string fields.

b/493761002

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

No

### DIFF
--- a/internal/cache/metadata/stat_cache.go
+++ b/internal/cache/metadata/stat_cache.go
@@ -119,17 +119,21 @@ type entry struct {
 	implicitDir bool
 }
 
-// Size returns the memory-size (resident set size) of the receiver entry.
-// The size calculated by the unsafe.Sizeof calls, and
-// NestedSizeOfGcsMinObject etc. does not account for
-// hidden members in data structures like maps, slices, linked-lists etc.
-// To account for those, we are adding a fixed constant of 515 bytes (deduced from
-// benchmark runs) to heap-size per positive stat-cache entry
-// to calculate a size closer to the actual memory utilization.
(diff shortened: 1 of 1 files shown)


internal/cache/metadata/stat_cache.go
```diff
@@ -126,23 +126,23 @@ type entry struct {
...
 	if e.f != nil {
-		size += uint64(util.UnsafeSizeOf(&e.f))
+		size += uint64(util.NestedSizeOfGcsFolder(e.f))
 	}
 
 	// Convert heap-size to RSS (resident set size).
 	size = uint64(math.Ceil(util.HeapSizeToRssConversionFactor * float64(size)))
 
-	return
+	return size
```
vadlakondaswetha, line 149: implicitDir in entry is a new boolean added as part of type cache deprecation. can you add that also to the size calculation
https://github.com/GoogleCloudPlatform/gcsfuse/blob/f156c3ee2765fec4fcc534aaab60b25b43d2bee4/internal/cache/metadata/stat_cache.go#L119

kislaykishore: I think, the `util.UnsafeSizeOf(&e)` should take care of that. Lemme know if that's not the case.

PR #4500, opened 2026-03-19
feat(force-dp-check): adding test to validate the force-dp using proxy server

raj-prince:
### Description
1. Update emulator.sh to start gRPC emulator server.
2. Implemented gRPC proxy server to accept gRPC traffic from GCSFuse and pass it on emulator service (although right now not needed but useful in future for gRPC related emulator based test). Added header validation check in proxy server.
3. Added test to confirm gRPC traffic contains the required header - (a) force_direct_connectivity & (b) direct_connectivity_diagnostic

b/493748028

### Testing details
1. Manual - Yes.
4. Unit tests - NA
5. Integration tests - Yes (tested manually).

### DIFF
--- a/tools/proxy_server/main.go
+++ b/tools/proxy_server/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -29,8 +30,6 @@ import (
 	"strings"
 	"syscall"
 	"time"
-
-	"google.golang.org/grpc"
 )
 
 const PortAndProxyProcessIdInfoLogFormat = "Listening Proxy Server On Port [%s] with Process ID [%d]"
@@ -243,19 +242,13 @@ func main() {
 	gOpManager = NewOperationManager(*gConfig)
 
 	// Determine proxy type from config (default to http if not specified)
-	proxyType := strings.ToLower(gConfig.ProxyType)
-	if proxyType == "" {
-		proxyType = "http"
-	}
-
-	if proxyType == "grpc" {
+	switch strings.ToLower(gConfig.ProxyType) {
+	case "grpc":
 		log.Println("Starting gRPC proxy server...")
-		gs := NewGRPCProxyServer()
-		gs.Start()
-	} else {
+		NewGRPCProxyServer().Start()
+	default:
 		log.Println("Starting HTTP proxy server...")
-		ps := NewProxyServer()
-		ps.Start()
+		NewProxyServer().Start()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tools/proxy_server/grpc_proxy.go
```diff
@@ -0,0 +1,220 @@
...
+				}
+			}
+		}
+
+		// Return first error encountered, or nil if both succeeded
+		if err1 != nil {
+			return err1
+		}
+		return err2
+	}
```
meet2mky, line 208: Is it required to wait for the cases to complete? Can't we return as soon as any of them fail?

tools/proxy_server/main.go
```diff
@@ -238,6 +242,59 @@ func main() {
...
+
+	if proxyType == "grpc" {
+		log.Println("Starting gRPC proxy server...")
+		gs := NewGRPCProxyServer()
+		gs.Start()
+	} else {
+		log.Println("Starting HTTP proxy server...")
+		ps := NewProxyServer()
+		ps.Start()
+	}
```
meet2mky, line 252: Nit: Simplify
```suggestion
// Start gRPC proxy server if explicitly specified else HTTP.
switch strings.ToLower(gConfig.ProxyType) {
case "grpc":
	log.Println("Starting gRPC proxy server...")
	NewGRPCProxyServer().Start()
default:
	log.Println("Starting HTTP proxy server...")
	NewProxyServer().Start()
}
```

tools/proxy_server/main.go
```diff
@@ -238,6 +242,59 @@ func main() {
...
+		log.Fatalf("Error on listening: %v", err)
+	}
+	gPort = strconv.Itoa(listener.Addr().(*net.TCPAddr).Port)
+	log.Printf(PortAndProxyProcessIdInfoLogFormat, gPort, os.Getpid())
+
+	// Start gRPC server in goroutine
+	go func() {
+		if err := startGRPCProxy(listener, gConfig.TargetHost, gConfig.HeaderValidation); err != nil && err != grpc.ErrServerStopped {
+			log.Fatalf("gRPC server error: %v", err)
+		}
```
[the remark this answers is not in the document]
meet2mky: please check if this needs handling.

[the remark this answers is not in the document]
raj-prince: Yup, requires.

tools/proxy_server/grpc_proxy.go
raj-prince: Yeah, we should wait for both the error to keep the state clean. Returning early may create a go-routine leak (as other routine will run and eventually write to a channel, but nobody would be reading). We need to handle that leak separately and given no major gain with early return, ROI is less.

meet2mky (approved):
LGTM

PR #4503, opened 2026-03-19
feat(shared-chunk-cache): adding basic e2e test for shared-chunk-cache

raj-prince:
### Description
Adding cache-hit/miss test for shared-chunk-cache using two mounts scenario. To keep the test-execution robust we used modification time to test the cache-hit/cache-miss logic, not via log-parsing.

b/495298967

### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/shared_chunk_cache/shared_chunk_cache_test.go
+++ b/tools/integration_tests/shared_chunk_cache/shared_chunk_cache_test.go
@@ -66,7 +66,7 @@ func (t *BaseSuite) SetupTest() {
 		t.sharedCacheDir = path.Join(GKETempDir, "shared-cache", "gcsfuse-shared-chunk-cache")
 		// Clean up cache directory before each test to ensure clean state
 		operations.RemoveDir(t.sharedCacheDir)
-		
+
 		// GKE Mode: Already mounted
 		t.primaryMount.mntDir = testEnv.cfg.GKEMountedDirectory
 		t.primaryMount.testDirPath = path.Join(t.primaryMount.mntDir, testDirName)


tools/integration_tests/shared_chunk_cache/setup_test.go
```diff
@@ -0,0 +1,373 @@
...
+	// Verify cache files were NOT modified (proving they weren't re-downloaded)
+	newCacheFileTimes := t.getCacheFileModTimes()
+	for filePath, oldInfo := range cacheFileTimes {
+		newInfo, exists := newCacheFileTimes[filePath]
+		require.True(t.T(), exists, "Cache file should still exist: %s", filePath)
+		require.Equal(t.T(), oldInfo.ModTime(), newInfo.ModTime(),
+			"Cache file should not be modified (proving cache hit): %s", filePath)
+	}
+
+	speedup := float64(primaryCacheMissTime) / float64(cacheHitTime)
```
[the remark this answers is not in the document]
raj-prince: cache hit will never be zero.

tools/integration_tests/shared_chunk_cache/setup_test.go
```diff
@@ -0,0 +1,114 @@
...
+	setup.OverrideFilePathsInFlagSet(testEnv.cfg, setup.TestDir())
+
+	// For dual mount, we create another directory.
+	secondaryDir, err := os.MkdirTemp(setup.TestDir(), "gcsfuse-secondary-mount")
+	if err != nil {
+		log.Fatalf("Failed to create secondary mount directory: %v", err)
+	}
+	testEnv.cfg.GCSFuseMountedDirectorySecondary = secondaryDir
+
+	log.Println("Running static mounting tests for shared chunk cache...")
```
meet2mky: Nit: Let's remove this log message from here and add it the setupTest where mounting with static is actually happening?

tools/integration_tests/shared_chunk_cache/shared_chunk_cache_test.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		t.T().Logf("Error walking cache directory: %v", err)
+	}
+	return count
+}
+
+func (t *BaseSuite) getCacheFileModTimes() map[string]os.FileInfo {
```
meet2mky, line 174: Is it the best way to get file cache hit verification? I have seen in read_cache that a lot of verifications are done via trace logs and verified via presence of cacheHit=true and false?

tools/integration_tests/shared_chunk_cache/shared_chunk_cache_test.go
```diff
@@ -0,0 +1,372 @@
...
+		t.primaryMount.setupTestDir(testEnv.cfg.GCSFuseMountedDirectory, testEnv.cfg.LogFile)
+		t.mountGcsfuse(t.primaryMount, "primary", t.primaryFlags)
+
+		if len(t.secondaryFlags) > 0 {
+			secondaryLog := path.Join(path.Dir(testEnv.cfg.LogFile), "gcsfuse_secondary.log")
+			t.secondaryMount.setupTestDir(testEnv.cfg.GCSFuseMountedDirectorySecondary, secondaryLog)
+			t.mountGcsfuse(t.secondaryMount, "secondary", t.secondaryFlags)
+		}
+	}
+}
```
meet2mky, line 94: Nit: can we combine both if else together if it's possible for simplicity.

tools/integration_tests/shared_chunk_cache/shared_chunk_cache_test.go
```diff
@@ -0,0 +1,372 @@
...
+		readOffset   = 10 * util.MiB // Start at 10MB (in the 2nd chunk)
+	)
+
+	// Arrange: Set up test file and verify initial cache state
+	t.createTestFile(testFileName, fileSize)
+	initialCacheCount := t.getCachedChunkCount()
+	require.Equal(t.T(), 0, initialCacheCount, "Cache should be empty initially")
+
+	// Act: Read 2MB from the 2nd chunk (triggers download and caching of entire 10MB chunk)
+	primaryFilePath := path.Join(t.primaryMount.testDirPath, testFileName)
```
meet2mky, line 215: Is this operation fully synchronous. I mean that it is guaranteed to have downloaded and cached 10MB before returning the response so next read is always hit. If it's not then in slow network or high load condition it would make the tests flaky.

meet2mky (commented):
Some comments, Does these tests need to be in a separate package and not read_cache. What's the reason for it?

tools/integration_tests/shared_chunk_cache/shared_chunk_cache_test.go
raj-prince: Yes, given log based validation complicates the test with regex-based parsing and validation, and here we need to parse and validate log files from two mounts.

This seems the best approach to validate the MVP.

raj-prince:
> Some comments, Does these tests need to be in a separate package and not read_cache. What's the reason for it?

Read_cache test are single mount test, while this is multi-mount test, hence created a separate package for this.

tools/integration_tests/shared_chunk_cache/shared_chunk_cache_test.go
raj-prince: Yes, it's read-through-cache and guarantees the download will happen first then only returns the response.

meet2mky:
Looks good.

Let's add the information of this package in improved_e2e_test script and in test_config.yaml for GKE in separate PR for continuous testing.

PR #4514, opened 2026-03-23
feat: Round-up file-sizes by block-size in file-cache (opt-out by experimental-file-cache-disable-size-calculation-fix)

gargnitingoogle:
### Description
- This fixes the accounting of the sizes of the cached files in the cache-dir, by rounding them up to the block-size of the cache-dir. This is the first phase of the fix needed to contain the actual disk-utilization of `cache-dir` at the specified limit of `file-cache-max-size-mb`.
- Adds new experimental hidden flag new hidden flag
`experimental-file-cache-disable-size-calculation-fix` to disable the fix, if needed as a fallback for unexpected behavior change.

**Note**: This fix makes no differences to the cases where `file-cache-max-size-mb` is specified as `-1`, which is the default value for that flag. So, this fix doesn't change anything unless `file-cache-max-size-mb` is explicitly set to a value other than `-1`.

**Note**: This fix doesn't completely ensure that the actual disk-size of `cache-dir` doesn't go beyond `file-cache-max-size-mb` as it doesn't address the space taken up by directories in `cache-dir`. It also doesn't clear up empty directories in `cache-dir`. For that, more changes are planned in further phases.

**Note**: This fix doesn't apply when the cache is sparse-enabled (i.e. `file-cache-experimental-enable-chunk-cache=true`), as the file-size rounding-up is not needed for them, so this fix is disabled for that case.

[b/496408968](http://b/496408968)

### Testing details
1. Manual - Tested locally with `file-cache-max-size-mb` set to `1` and `number of directories ~= number of files` in the read workload. With that set up, I let a multi-thread read application run for about 2 minutes. Saw the effect of the change.
   - Without the fix
      - the size of cache-dir (output of `du -sh <cache-dir>`) grew up to 4.5 MB
      - Number of files in cache-dir grew to about 554 files.
   - With the fix
      - the size of cache-dir grew to and stayed at about 2.5 MB (because of 1.5 MB taken by directories)
      - the number of files in the cache-dir grew to and stayed at 256 (all small files in the test set-up), which makes sense because 4k * 256 = 1 MB.
2. Unit tests - Added/updated unit tests for the changes.
   - e2e tests for non-zb: 
     - [run1](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/ef5cf880-5046-4bef-9908-3a749f5f628d/log) - passed
     - [run2](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/3ba39858-24cd-42ed-bf48-7daa720fe7cc/summary) - passed
     - [run3](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/181a05da-c4d4-4007-8529-f2d0b72077de)
     - [run4 - with the block-size stored in FileInfo and DownloadJob](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/f968fdd6-3296-490a-b1e4-de2c9a4823d6)
     - [run5](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/3fd61520-0fe4-4bda-9c2b-27a397a26cf8/summary)
     - [run6](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/71806983-3234-46fa-8135-d8685fac4845)
     - [run7 - latest changes with addressing Kislay's comments](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/b82a2873-e374-437f-9290-7e5a4f70a6b9)
     - [run8 with latest comments addressed](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/3a5f4dde-f5ee-4b24-84d6-c4ae704efe3c/log)
     - [run9 - latest code](http://fusion2/ci/kokoro/prod%3Agcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/d6b277dc-27fc-4513-b50f-bed032af6b89)
   - e2e test with zb:
     - [run1](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/6f08eafa-adb4-4bff-a663-c54adab6a891/summary)
   - perf run: 
      - [run with the block-size stored in FileInfo and DownloadJob](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/1064d047-e766-48a8-8d73-fbdbb1ca456b/log)
      ```md
      +--------+------------+--------------+------------+--------------+--------------+
      +--------+------------+--------------+------------+--------------+--------------+
      | Master |  0.25MiB   | 558.53MiB/s  |  1.3MiB/s  |  80.15MiB/s  |  1.11MiB/s   |
      |   PR   |  0.25MiB   | 576.11MiB/s  | 1.05MiB/s  |  82.28MiB/s  |  1.27MiB/s   |
      | Master | 48.828MiB  | 4410.43MiB/s | 78.73MiB/s | 1546.12MiB/s |  79.43MiB/s  |
      |   PR   | 48.828MiB  | 4450.18MiB/s | 77.0MiB/s  | 1567.94MiB/s |  78.95MiB/s  |
      | Master | 976.562MiB | 4432.69MiB/s | 35.19MiB/s | 765.57MiB/s  |  38.76MiB/s  |
      |   PR   | 976.562MiB | 4429.37MiB/s | 35.21MiB/s | 1089.37MiB/s |  39.1MiB/s   |
      +--------+------------+--------------+------------+--------------+--------------+
      ```
   - Scale tests
     - Checkpointing benchmark
       - With fix: Run in progress. Will add data when available.

The fix can potentially change the user experience indirectly, as the `cache-dir` disk-utilization will now be closer to `file-cache-max-size-mb`, and it can reduce how many files are actually cached, if there are many small files.

### DIFF
--- a/internal/util/diskutil/disk_util.go
+++ b/internal/util/diskutil/disk_util.go
@@ -50,6 +50,7 @@ func GetVolumeBlockSize(path string) uint64 {
 		return defaultVolumeBlockSize
 	}
 	// Prefer Frsize (fragment size) over Bsize for actual disk allocation if available.
+	// Refer https://github.com/cockroachdb/pebble/pull/1072.
 	blockSize := uint64(stat.Bsize)
 	if stat.Frsize > 0 {
 		blockSize = uint64(stat.Frsize)


internal/cache/lru/lru.go
```diff
@@ -155,7 +169,7 @@ func (c *Cache) Insert(
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	valueSize := value.Size()
+	valueSize := c.sizeCalcFunc(value)
```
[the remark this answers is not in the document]
gargnitingoogle: done

cfg/config.go
```diff
@@ -489,6 +489,8 @@ type FileCacheConfig struct {
 
 	ExcludeRegex string `yaml:"exclude-regex"`
 
+	ExperimentalDisableSizeCalculationFix bool `yaml:"experimental-disable-size-calculation-fix"`
```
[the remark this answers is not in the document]
gargnitingoogle: yes, it was followed.

internal/util/disk_util_test.go
```diff
@@ -0,0 +1,90 @@
...
+	blockSize, err := GetVolumeBlockSize(tempDir)
+	if err != nil {
+		t.Fatalf("GetVolumeBlockSize failed: %v", err)
+	}
+
+	// Assuming a standard block size like 4096 for most modern filesystems
+	// This might fail on specific edge-case filesystems, but should work for common CI environments
+	if blockSize <= 0 {
+		t.Errorf("Expected positive block size, got: %d", blockSize)
+	}
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/cache/file/cache_handler.go
```diff
@@ -62,15 +63,36 @@ type CacheHandler struct {
...
 	var compiledIncludeRegex *regexp.Regexp
 
 	compiledExcludeRegex = compileRegex(excludeRegex)
 	compiledIncludeRegex = compileRegex(includeRegex)
 
+	if sizeCalcFix {
+		volumeBlockSize, err := baseutil.GetVolumeBlockSize(cacheDir)
+		if err != nil {
+			logger.Warnf("Failed to get volume block size for cacheDir %q: %v. Using default 4096.", cacheDir, err)
+			volumeBlockSize = 4096
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/cache/lru/lru.go
```diff
@@ -297,3 +311,10 @@ func (c *Cache) EraseEntriesWithGivenPrefix(prefix string) {
 		}
 	}
 }
+
+// SetSizeCalcFunc overrides the default size calculation logic for the LRU entries.
+func (c *Cache) SetSizeCalcFunc(sizeCalcFunc func(ValueType) uint64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.sizeCalcFunc = sizeCalcFunc
+}
```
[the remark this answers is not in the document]
gargnitingoogle: done

internal/util/diskutil/disk_util.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2026 Google LLC
```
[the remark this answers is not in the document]
gargnitingoogle: bad comment!

internal/util/diskutil/disk_util_test.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2026 Google LLC
```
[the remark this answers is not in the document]
gargnitingoogle: bad comment!

internal/util/disk_util.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"syscall"
+)
+
+// GetSpeculativeFileSizeOnDisk calculates the theoretical disk space a file will
+// consume given its actual content size and the filesystem's block size. It rounds
+// up the content size to the nearest block boundary to simulate block allocation.
+func GetSpeculativeFileSizeOnDisk(fileContentSize uint64, volumeBlockSize uint64) uint64 {
```
gargnitingoogle:
```suggestion
func GetSpeculativeFileSizeOnDisk(fileContentSize, volumeBlockSize uint64) uint64 {
```

internal/cache/lru/lru_test.go
```diff
@@ -362,3 +376,90 @@ func (t *CacheTest) TestRaceCondition() {
...
+	t.insertAndAssert("item3", testData{Value: 3, DataSize: 2}, []int64{1}, nil)
+
+	// item1 got evicted because insertion of item3 required space.
+	// This eviction would not have happened if SizeCalcFunc was not applied.
+	ExpectEq(nil, t.cache.LookUp("item1"))
+	ExpectEq(2, t.cache.LookUp("item2").(testData).Value)
+	ExpectEq(3, t.cache.LookUp("item3").(testData).Value)
+}
+
+func (t *CacheTest) TestCustomSizeCalcFunc_PanicOnNonEmptyCache() {
```
gargnitingoogle:
```suggestion
func (t *CacheTest) TestSetSizeCalcFunc_PanicOnNonEmptyCache() {
```

internal/util/disk_util_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+func TestGetSpeculativeFileSizeOnDisk(t *testing.T) {
+	tests := []struct {
+		name            string
+		fileContentSize uint64
+		volumeBlockSize uint64
+		expectedSize    uint64
+	}{
+		{
+			name:            "Zero Block Size",
```
gargnitingoogle:
```suggestion
			name:            "Zero_Block_Size",
```

cmd/root_test.go
```diff
@@ -570,8 +570,8 @@ func TestArgsParsing_FileCacheFlags(t *testing.T) {
 		expectedConfig *cfg.Config
 	}{
 		{
-			name: "Test file cache flags.",
-			args: []string{"gcsfuse", "--file-cache-cache-file-for-range-read", "--file-cache-download-chunk-size-mb=20", "--file-cache-enable-crc", "--cache-dir=/some/valid/dir", "--file-cache-exclude-regex=.*", "--file-cache-include-regex=.*", "--file-cache-enable-parallel-downloads", "--file-cache-max-parallel-downloads=40", "--file-cache-max-size-mb=100", "--file-cache-parallel-downloads-per-file=2", "--file-cache-enable-o-direct=false", "abc", "pqr"},
+			name: "Test_file_cache_flags",
```
kislaykishore: Same for other subtests.
```suggestion
			name: "file_cache_flags",
```

internal/cache/lru/lru.go
```diff
@@ -46,9 +46,12 @@ type Cache struct {
 	// Mutable state
 	/////////////////////////
 
-	// Sum of entry.Value.Size() of all the entries in the cache.
+	// Sum of the effective sizes of all the entries in the cache.
 	currentSize uint64
 
+	// Function to calculate the actual contribution of the given entry towards currentSize.
+	sizeCalcFunc func(ValueType) uint64
```
kislaykishore: Consider keeping a static size in the entry itself as a field as opposed to computing it dynamically. I understand it'll increase the size of the cache by 8 bytes per entry but I feel it's not a bad thing to do from a readability and performance point of view.

internal/cache/file/cache_handler.go
```diff
@@ -64,13 +69,32 @@ type CacheHandler struct {
...
+		if isSparse {
+			logger.Info("file-cache disk-utilization fix is not supported with sparse-mode and is disabled.")
+		} else {
+			volumeBlockSize, err := baseutil.GetVolumeBlockSize(cacheDir)
+			if err != nil {
+				logger.Warnf("Failed to get volume block size for cacheDir %q: %v. Using default %d.", cacheDir, err, defaultCacheDirVolumeBlockSize)
+				volumeBlockSize = defaultCacheDirVolumeBlockSize
+			}
+
+			fileInfoCache.SetSizeCalcFunc(func(v lru.ValueType) uint64 {
```
kislaykishore: Instead of doing this, consider modifying the Size() function in the FileInfo object so that it takes the block size into account.

internal/util/disk_util.go
```diff
@@ -0,0 +1,43 @@
...
+// using statfs sys-call. The block-size can be 0 or 2^n. The most common value is 4096.
+func GetVolumeBlockSize(path string) (uint64, error) {
+	var stat syscall.Statfs_t
+	if err := syscall.Statfs(path, &stat); err != nil {
+		return 0, fmt.Errorf("failed to get stats for path %q: %w", path, err)
+	}
+	// Bsize is int64 on some platforms (like Linux/amd64) and uint32 on others (like Linux/arm).
+	// Casting to uint64 ensures consistency.
+	return uint64(stat.Bsize), nil
+}
```
kislaykishore: This function looks cacheable since file-cache path is not going to change throughout the lifetime of the process. Could we optimize this?

internal/cache/file/cache_handler.go
```diff
@@ -64,13 +69,32 @@ type CacheHandler struct {
...
+				if _, ok := v.(data.FileInfo); ok {
+					return baseutil.GetSpeculativeFileSizeOnDisk(v.Size(), volumeBlockSize)
```
kislaykishore: Trying to keep the interface simple - how about we just take size and directory as inputs and let the GetSpeculativeFileSizeOnDisk take care of computing the volumeBlockSize math?

cmd/root_test.go
gargnitingoogle: will change

internal/cache/file/cache_handler.go
gargnitingoogle: This will require GetVolumeBlockSize() (indirectly statfs call) to be made for every single cached file separately, which will kill the whole performance.

internal/cache/file/cache_handler.go
gargnitingoogle: This will either require the FileInfo's Size() function's interface to be changed to also take a block-size, or for FileInfo to store a block-size. The latter will just lead to unnecessary memory increase by adding a 8-byte block-size integer in every FileInfo variable. The former doesn't add memory, but changes a function signature, which is also used in file-cache downloader, which is an unnecessary change going beyond the component which actually needs the fix.

internal/util/disk_util.go
gargnitingoogle: No, this is already as optimal as I think as it can be. This is called from NewCacheHandler, which is called exactly once per mount.

cmd/root_test.go
gargnitingoogle: done

internal/cache/file/cache_handler.go
gargnitingoogle: Discussed offline that this is not a good idea for the solution without losing performance.

internal/util/disk_util.go
gargnitingoogle: Discussed offline that this is done only once per mount. So, this is not a problem.

internal/cache/file/cache_handler.go
gargnitingoogle: Had a discussion with @kislaykishore offline, who suggested wrapping the size in a new struct e.g. a FileCacheEntry which also stores the size, rather than saving a function to calculate the size for each file-cache-entry. This is simply about design nicety and readability.

internal/fs/fs.go
```diff
@@ -367,6 +367,17 @@ func createSingleMountFileCacheHandler(baseCacheDir string, filePerm, dirPerm os
 	}
 
 	fileInfoCache := lru.NewCache(sizeInBytes)
+
+	var volumeBlockSize uint64 = 1
+	if !serverCfg.NewConfig.FileCache.ExperimentalDisableSizeCalculationFix && !serverCfg.NewConfig.FileCache.EnableExperimentalSharedChunkCache {
```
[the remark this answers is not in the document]
gargnitingoogle: good catch! fixed.

internal/cache/file/cache_handler.go
gargnitingoogle: I explored 2 different approaches for doing this cleanly, which involved (a) adding a volumeBlockSize in each FileInfo struct, and (b) create a new struct CacheEntry in internal/cache/file as a wrapper over FileInfo. Overall, the changes for both would be in nearly all the same files, but (b) would result in slightly higher number of changes overall. So, I've stuck with (a), and have pushed the changes for (a).

internal/fs/fs.go
```diff
@@ -367,6 +367,17 @@ func createSingleMountFileCacheHandler(baseCacheDir string, filePerm, dirPerm os
...
+
+	var cacheDirVolumeBlockSize uint64 = 1
+	if !serverCfg.NewConfig.FileCache.ExperimentalDisableSizeCalculationFix && !serverCfg.NewConfig.FileCache.ExperimentalEnableChunkCache {
+		var err error
+		cacheDirVolumeBlockSize, err = util.GetVolumeBlockSize(cacheDir)
+		if err != nil {
+			logger.Warnf("Failed to get volume block size for cacheDir %q: %v. Using default 4096.", cacheDir, err)
+			cacheDirVolumeBlockSize = 4096
+		}
+	}
```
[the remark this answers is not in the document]

internal/fs/fs.go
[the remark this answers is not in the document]

internal/util/disk_util.go
```diff
@@ -0,0 +1,43 @@
...
+	numBlocks := (fileContentSize + volumeBlockSize - 1) / volumeBlockSize
+	return numBlocks * volumeBlockSize
+}
+
+// GetVolumeBlockSize retrieves the block size of the file system containing the given path
+// using statfs sys-call. The block-size can be 0 or 2^n. The most common value is 4096.
+func GetVolumeBlockSize(path string) (uint64, error) {
+	var stat syscall.Statfs_t
+	if err := syscall.Statfs(path, &stat); err != nil {
+		return 0, fmt.Errorf("failed to get stats for path %q: %w", path, err)
```
kislaykishore: Let's return 4KiB here in case of errors. Also, please log the error so that it's clear that we are falling back. This means that this method definition would change to not return error.

internal/fs/fs.go
```diff
@@ -367,6 +367,21 @@ func createSingleMountFileCacheHandler(baseCacheDir string, filePerm, dirPerm os
...
+			logger.Info("file-cache disk-utilization fix is not supported with sparse-mode and is disabled.")
+		} else {
+			var err error
+			cacheDirVolumeBlockSize, err = util.GetVolumeBlockSize(cacheDir)
+			if err != nil {
+				logger.Warnf("Failed to get volume block size for cacheDir %q: %v. Using default %d.", cacheDir, err, file.DefaultCacheDirVolumeBlockSize)
+				cacheDirVolumeBlockSize = file.DefaultCacheDirVolumeBlockSize
+			}
+		}
+	}
```
kislaykishore: Extract this code into a separate method.

internal/fs/fs.go
```diff
@@ -367,6 +367,21 @@ func createSingleMountFileCacheHandler(baseCacheDir string, filePerm, dirPerm os
...
+	if !serverCfg.NewConfig.FileCache.ExperimentalDisableSizeCalculationFix {
+		if serverCfg.NewConfig.FileCache.ExperimentalEnableChunkCache {
+			logger.Info("file-cache disk-utilization fix is not supported with sparse-mode and is disabled.")
+		} else {
+			var err error
+			cacheDirVolumeBlockSize, err = util.GetVolumeBlockSize(cacheDir)
+			if err != nil {
+				logger.Warnf("Failed to get volume block size for cacheDir %q: %v. Using default %d.", cacheDir, err, file.DefaultCacheDirVolumeBlockSize)
+				cacheDirVolumeBlockSize = file.DefaultCacheDirVolumeBlockSize
+			}
```
kislaykishore: Assuming that you make the change to ensure that `GetVolumeBlockSize` doesn't return an error, this block and the `DefaultCacheDirVolumeBlockSize` constant won't be needed.

internal/util/diskutil/disk_util.go
kislaykishore, line 1: Let's move this to internal/diskutil/diskutil.go so that its purpose is clear.

internal/gcsx/file_cache_reader_test.go
```diff
@@ -93,8 +93,13 @@ func (t *fileCacheReaderTest) SetupTest() {
 	t.cacheDir = path.Join(os.Getenv("HOME"), "test_cache_dir")
 	lruCache := lru.NewCache(cacheMaxSize)
 	fileCacheConfig := &cfg.FileCacheConfig{EnableCrc: false}
-	t.jobManager = downloader.NewJobManager(lruCache, util.DefaultFilePerm, util.DefaultDirPerm, t.cacheDir, sequentialReadSizeInMb, fileCacheConfig, metrics.NewNoopMetrics(), tracing.NewNoopTracer())
-	t.cacheHandler = file.NewCacheHandler(lruCache, t.jobManager, t.cacheDir, util.DefaultFilePerm, util.DefaultDirPerm, "", "", false)
+	cacheDirVolumeBlockSize, err := testutil.GetVolumeBlockSize(t.cacheDir)
+	if err != nil {
+		cacheDirVolumeBlockSize = 4096
```
kislaykishore: It's inconsistent - in some case we use DefaultCacheDirVolumeBlockSize.

internal/util/disk_util.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	// Bsize is int64 on some platforms (like Linux/amd64) and uint32 on others (like Linux/arm).
```
kislaykishore: Could you link the relevant documentation to support this?

internal/cache/data/file_info.go
```diff
@@ -50,16 +52,31 @@ type FileInfo struct {
...
 		return fi.DownloadedChunks.TotalBytes()
 	}
 	return fi.FileSize
 }
 
+// Size returns the speculative physical size on disk, rounded up to the volume block size.
+// If CacheDirVolumeBlockSize is 0, this returns 0 (no disk consumption).
+// If CacheDirVolumeBlockSize is 1, it returns the exact logical ContentSize.
+// This satisfies the LRU ValueType interface for eviction accounting.
+func (fi FileInfo) Size() uint64 {
```
kislaykishore, line 75: I'd recommend using pointer-receivers to avoid copies. Feel free to benchmark it to see if it helps.
```suggestion
func (fi *FileInfo) Size() uint64 {
```

internal/cache/file/cache_handler.go
```diff
@@ -193,12 +201,13 @@ func (chr *CacheHandler) addFileInfoEntryAndCreateDownloadJob(object *gcs.MinObj
 
 	if addEntryToCache {
 		newFileInfo := data.FileInfo{
```
kislaykishore: Could we create a `NewFileInfo` method to ensure that we don't mistakenly forget setting the CacheDirVolumeBlockSize field anywhere?

internal/cache/file/cache_handler_test.go
```diff
@@ -1122,3 +1130,60 @@ func Test_Destroy(t *testing.T) {
...
+	// The handler should configure the block size (default 4096 in testing if statfs isn't mocked properly,
+	// or the actual tmpdir block size, which is usually 4096).
+	require.GreaterOrEqual(t, handler.volumeBlockSize, uint64(512))
+	// Verify that inserting a 1-byte file via the handler's calculated block size would overflow a 100-byte cache.
+	fi := data.FileInfo{
+		Key: data.FileInfoKey{
+			ObjectName: "test.txt",
+		},
+		FileSize:                1,
+		CacheDirVolumeBlockSize: handler.volumeBlockSize,
```
kislaykishore:
```suggestion
		CacheDirVolumeBlockSize: cacheDirVolumeBlockSize,
```

internal/util/disk_util.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"syscall"
+)
+
+// GetSpeculativeFileSizeOnDisk calculates the theoretical disk space a file will
+// consume given its actual content size and the filesystem's block size. It rounds
+// up the content size to the nearest block boundary to simulate block allocation.
+func GetSpeculativeFileSizeOnDisk(fileContentSize, volumeBlockSize uint64) uint64 {
+	if volumeBlockSize == 0 {
+		return 0
```
kislaykishore: Should we return `fileContentSize` if `volumeBlockSize` is 0? That will help maintain parity in case we have forgotten to correctly set the `volumeBlockSize` value and ended up passing 0 breaking the file-cache.

internal/cache/file/cache_handler_test.go
```diff
@@ -1122,3 +1130,60 @@ func Test_Destroy(t *testing.T) {
...
+	require.NotNil(t, handler)
+	// The handler should configure the block size to 1.
+	require.Equal(t, handler.volumeBlockSize, uint64(1))
+	// Verify that inserting a 1-byte file via the handler's block size of would work.
+	fi := data.FileInfo{
+		Key: data.FileInfoKey{
+			ObjectName: "test.txt",
+		},
+		FileSize:                5,
+		CacheDirVolumeBlockSize: handler.volumeBlockSize,
```
kislaykishore: Same as above - let's declare a variable for block-size and reuse it in both places. That'll make the test more readable.

internal/gcsx/file_cache_reader_test.go
gargnitingoogle: this one looks like a miss. This if block will be removed if we remove the error return-value from GetVolumeBlockSize.

internal/util/disk_util.go
gargnitingoogle: I couldn't find a definiteve document, only conversations about it. Removing this information.

internal/cache/data/file_info.go
gargnitingoogle: I followed the pattern from the existing Size() function which used non-pointer-receiver.
While pointer does better in performance by avoiding copy-by-value, it enables the possibility of unintentionally modifying the member variables of fi in it.
Anyway, for now, I've switched this and original Size method (now called ContentSize) to pointer-receivers.

gargnitingoogle: No, wait, I can't change the signature of Size() in FileInfo from non-pointer receiver to pointer-receiver otherwise it becomes incompatible with the ValueType interface at https://github.com/GoogleCloudPlatform/gcsfuse/blob/ce54e44c89204c86c0d3bc0f9de617f3fa602a46/internal/cache/lru/lru.go#L69 , and to make it work, I'll have to change a lot of code in lru.go. So, I don't think it's a good change to make, also given that FileInfo isn't so big a structure that we should worry about the cost of its copy.

kislaykishore: I guess all you would need to do is pass &fileInfo to the cache instead of fileInfo. But let me know if that understanding is flawed.

gargnitingoogle: Yes, it can be handled by passing the pointer to lru from the file-cache handler, and receiving the same pointer back from file-cache lru when we call `Get()`.

According to gemini, the change is about 52 lines across 10 files (including the file cache handler, downloader and different unit test files). But more than that, this is an unsafe change, because then `lru.Get(key)` will return the file-cache-entry as a pointer, whose fields are now mutable. This can easily make the lru editable through backdoor, potentially jeopardizing the encapsulation and lock/unlock mechanism.

I suggest that we keep such a change out of scope of the current PR (which is pretty safe from functionality perspective). If you want to further explore down this path (pointer-receiver), let's take this up separately.

kislaykishore: Sure. No problem

internal/cache/file/cache_handler_test.go
gargnitingoogle: done

internal/util/disk_util.go
gargnitingoogle: yes, done

internal/cache/file/cache_handler_test.go
gargnitingoogle: done

internal/cache/file/cache_handler.go
gargnitingoogle: done

internal/fs/fs.go
gargnitingoogle: yes, removed

internal/fs/fs.go
gargnitingoogle: done. added unit tests too for it.

internal/util/disk_util.go
gargnitingoogle: done

internal/util/diskutil/disk_util.go
gargnitingoogle: done

internal/cache/data/file_info_test.go
```diff
@@ -67,12 +67,14 @@ func TestKeyMethodWithEmptyObjectName(t *testing.T) {
 	assert.Equal(t, "", key)
 }
 
+func TestContentSizeMethod(t *testing.T) {
+	fi := NewFileInfo(getTestFileInfoKey(), TestGeneration, TestDataFileSize, 0, false, nil, 4096)
+
+	assert.Equal(t, TestDataFileSize, fi.ContentSize())
```
kislaykishore: It's difficult to go all the way up to check what's the value of `TestDataFileSize`
```suggestion
    fileContentSize := 23
	fi := NewFileInfo(getTestFileInfoKey(), TestGeneration, fileContentSize, 0, false, nil, 4096)

	assert.Equal(t, fileContentSize, fi.ContentSize())
```

internal/cache/data/file_info_test.go
```diff
@@ -67,12 +67,14 @@ func TestKeyMethodWithEmptyObjectName(t *testing.T) {
...
 func TestSizeMethod(t *testing.T) {
-	fi := FileInfo{
-		Key:              getTestFileInfoKey(),
-		ObjectGeneration: TestGeneration,
-		FileSize:         TestDataFileSize,
-	}
+	fi := NewFileInfo(getTestFileInfoKey(), TestGeneration, TestDataFileSize, 0, false, nil, 4096)
 
-	assert.Equal(t, TestDataFileSize, fi.Size())
+	assert.Equal(t, uint64(4096), fi.Size())
```
kislaykishore: It's difficult to go all the way up to check what's the value of `TestDataFileSize`
```suggestion
    fileContentSize := 23
	fi := NewFileInfo(getTestFileInfoKey(), TestGeneration, fileContentSize, 0, false, nil, 4096)

	assert.Equal(t, uint64(4096), fi.Size())
```

internal/util/diskutil/disk_util.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"syscall"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
+)
+
+// DefaultCacheDirVolumeBlockSize is the block-size used for cache-dir in case
+// statfs call fails for it.
+const DefaultVolumeBlockSize = 4096
```
kislaykishore: Do we need it to be exported?
```suggestion
const defaultVolumeBlockSize = 4096
```

internal/fs/fs_internal_test.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func TestFsInternalTestSuite(t *testing.T) {
+	suite.Run(t, new(FsInternalTestSuite))
+}
+
+func (suite *FsInternalTestSuite) SetupTest() {
+	var err error
+	suite.cacheDir, err = os.MkdirTemp("", "fs_internal_test")
+	assert.NoError(suite.T(), err)
```
kislaykishore:
```suggestion
	require.NoError(suite.T(), err)
```

internal/fs/fs_internal_test.go
```diff
@@ -0,0 +1,93 @@
...
+			FileCache: cfg.FileCacheConfig{
+				ExperimentalDisableSizeCalculationFix: false,
+				ExperimentalEnableChunkCache:          true,
+			},
+		},
+	}
+
+	// Sparse mode overrides and explicitly disables block size up-rounding
+	blockSize := cacheDirVolumeBlockSize(serverCfg, suite.cacheDir)
+	assert.Equal(suite.T(), uint64(1), blockSize)
```
kislaykishore: Add a newline before this to have separate Arrange, Act and Assert segments.

internal/fs/fs_internal_test.go
```diff
@@ -0,0 +1,93 @@
...
+	serverCfg := &ServerConfig{
+		NewConfig: &cfg.Config{
+			FileCache: cfg.FileCacheConfig{
+				ExperimentalDisableSizeCalculationFix: false,
+				ExperimentalEnableChunkCache:          false,
+			},
+		},
+	}
+
+	actualBlockSize := diskutil.GetVolumeBlockSize(suite.cacheDir)
```
kislaykishore: Move this to assert block.

internal/fs/fs_internal_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	blockSize := cacheDirVolumeBlockSize(serverCfg, suite.cacheDir)
```
kislaykishore: Add a blank line after this to segregate Act and Assert segments.

internal/fs/fs_internal_test.go
```diff
@@ -0,0 +1,93 @@
...
+		NewConfig: &cfg.Config{
+			FileCache: cfg.FileCacheConfig{
+				ExperimentalDisableSizeCalculationFix: true,
+				ExperimentalEnableChunkCache:          false,
+			},
+		},
+	}
+
+	// Because the size calculation fix is completely disabled, the block size returned should be 1
+	blockSize := cacheDirVolumeBlockSize(serverCfg, suite.cacheDir)
```
kislaykishore: Add a new line after this.

internal/fs/fs.go
```diff
@@ -345,6 +347,18 @@ func createSharedChunkCacheManager(baseCacheDir string, filePerm, dirPerm os.Fil
...
+	var cacheDirVolumeBlockSize uint64 = 1
+	if !serverCfg.NewConfig.FileCache.ExperimentalDisableSizeCalculationFix {
+		if serverCfg.NewConfig.FileCache.ExperimentalEnableChunkCache {
+			logger.Info("file-cache disk-utilization fix is not supported with sparse-mode, so is disabled.")
+		} else {
+			cacheDirVolumeBlockSize = diskutil.GetVolumeBlockSize(cacheDir)
+		}
+	}
+	return cacheDirVolumeBlockSize
+}
```
kislaykishore: Simplify - reduce nesting - https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html
```suggestion
	if serverCfg.NewConfig.FileCache.ExperimentalDisableSizeCalculationFix {
	  return 1
	}
	if serverCfg.NewConfig.FileCache.ExperimentalEnableChunkCache {
	  logger.Info("file-cache disk-utilization fix is not supported with sparse-mode, so is disabled.")
	  return 1
	}
	return diskutil.GetVolumeBlockSize(cacheDir)
}
```

internal/cache/data/file_info_test.go
gargnitingoogle: done

internal/util/diskutil/disk_util.go
gargnitingoogle: we could actually use it in test code rather than hardcoding 4096 in them, but it's ok to not do that now. So, make it internal.

internal/fs/fs_internal_test.go
gargnitingoogle: done

internal/fs/fs_internal_test.go
gargnitingoogle: done

internal/fs/fs_internal_test.go
gargnitingoogle: done

internal/fs/fs.go

internal/fs/fs_internal_test.go
gargnitingoogle: done

internal/fs/fs_internal_test.go
gargnitingoogle: done

internal/cache/data/file_info_test.go
gargnitingoogle: done

internal/fs/fs_internal_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/util/diskutil"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/cfg"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+type FsInternalTestSuite struct {
+	suite.Suite
```
kislaykishore: Let's avoid using suits if we can. In this case, could we just use os.TempDir() instead of creating a new directory for each test?
Also, consider making these tests table-driven.

internal/util/diskutil/disk_util.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// DefaultCacheDirVolumeBlockSize is the block-size used for cache-dir in case
+// statfs call fails for it.
+const defaultVolumeBlockSize = 4096
+
+// GetSpeculativeFileSizeOnDisk calculates the theoretical disk space a file will
+// consume given its actual content size and the filesystem's block size. It rounds
+// up the content size to the nearest block boundary to simulate block allocation.
+func GetSpeculativeFileSizeOnDisk(fileContentSize, volumeBlockSize uint64) uint64 {
```
kislaykishore, line 36: Should we cap the upper limit of block size as well? What if the statfs call returns and obscenely high value and ends up breaking file-cache? Note that statfs is not standardized in POSIX.
I would recommend we cap the block size to a high enough number such as 1024KiB.

internal/util/diskutil/disk_util.go
kislaykishore: You should ideally refrain from using such constants in tests. Instead use literals in tests

internal/fs/fs_internal_test.go
gargnitingoogle: done

internal/util/diskutil/disk_util.go
gargnitingoogle: Done. Added special handling for Bsize = 0 and Bsize > 1 MiB, and also according unit test changes.

internal/util/diskutil/disk_util.go
```diff
@@ -0,0 +1,67 @@
...
+		blockSize = uint64(stat.Frsize)
+	}
+	// Sanity check: If the value is 0 or suspiciously large, fallback to default or max
+	if blockSize == 0 {
+		logger.Errorf("statfs for %q returned Bsize = 0, so defaulting to %d", path, defaultVolumeBlockSize)
+		return defaultVolumeBlockSize
+	}
+	if blockSize > maxVolumeBlockSize {
+		logger.Errorf("statfs for %q returned Bsize (%d), which is too high, so truncating it to %d", path, blockSize, maxVolumeBlockSize)
+		return maxVolumeBlockSize
```
kislaykishore: If it's too high, use the default and not 1024Ki.

gargnitingoogle: done

internal/util/diskutil/disk_util.go
```diff
@@ -0,0 +1,67 @@
...
+
+// GetVolumeBlockSize retrieves the block size of the file system containing the given path
+// using statfs sys-call. The block-size can be 0 or 2^n. The most common value is 4096.
+func GetVolumeBlockSize(path string) uint64 {
+	var stat syscall.Statfs_t
+	if err := syscall.Statfs(path, &stat); err != nil {
+		logger.Errorf("statsfs failed for %q: %v. Defaulting to block-size %d for this directory.", path, err, defaultVolumeBlockSize)
+		return defaultVolumeBlockSize
+	}
+	// Prefer Frsize (fragment size) over Bsize for actual disk allocation if available.
```
kislaykishore, line 52: Please add documentation explaining why we prefer this.

internal/util/diskutil/disk_util_test.go
```diff
@@ -0,0 +1,113 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			blockSize := diskutil.GetVolumeBlockSize(tc.cachedir)
+
+			if tc.expectedVolumeBlockSize == 0 {
+				assert.True(t, blockSize > 0 && ((blockSize&(blockSize-1)) == 0), "Block-size of a directory should be a power of 2. Got: %d", blockSize)
+			} else {
+				assert.Equal(t, tc.expectedVolumeBlockSize, blockSize)
+			}
+		})
+	}
```
kislaykishore: Let's split this into two tests. "expected volume block size == 0" is kind of a flag here. Instead we are just verifying that it's a power of two.

gargnitingoogle: done

internal/util/diskutil/disk_util.go
gargnitingoogle: I have only found this in conversations and references over internet, e.g. https://github.com/cockroachdb/cockroach/issues/61182 but not on any linux documentations or journals. So, I have not been able to quote the source on it, but this reasoning seems consistent from. Gemini suggests the same too.

gargnitingoogle: Added reference to https://github.com/cockroachdb/pebble/pull/1072 as discussed offline.

PR #4515, opened 2026-03-23
fix: Fix data-race in lru-cache and 200x improvement in EraseEntriesWithGivenPrefix performance for deleting 10K entries out of 20K

kislaykishore:
### Description
- Fix a data race bug (fatal error: concurrent map iteration and map write) in lru where `c.index` was previously iterated without holding a lock while other operations might insert concurrently. While it doesn't cause issues in the applications since the operations to this module are already serialized, it's unintuitive and a code-smell.
- Refactored `Erase` and `EraseEntriesWithGivenPrefix` to use a common internal lock-free `eraseInternal` utility function.
- Ensured that `EraseEntriesWithGivenPrefix` will acquire the write lock once for the entire batch instead of taking it for every entry.
- Add benchmark tests for lru-cache erasure for a given prefix that show 18x performance improvement for erasing entries in bulk.

### Perf
**Before fix**  
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/cache/lru
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkEraseEntriesWithGivenPrefix-16                     1628            747389 ns/op
BenchmarkEraseEntriesWithGivenPrefix_Concurrent-16      fatal error: concurrent map iteration and map write
```

**After fix**  
```
[----------] Finished with tests from CacheTest
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/cache/lru
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkEraseEntriesWithGivenPrefix-16                    30418             39771 ns/op
BenchmarkEraseEntriesWithGivenPrefix_Concurrent-16        242079              4817 ns/op
```

b/495217870

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - NA

No

### DIFF
--- a/internal/cache/lru/lru_benchmark_test.go
+++ b/internal/cache/lru/lru_benchmark_test.go
@@ -101,7 +101,7 @@ func BenchmarkInsert1Million(b *testing.B) {
 
 		// Insert 1 million entries
 		for j := range numEntries {
-			key = fmt.Sprintf("prefix/key-%d", j)
+			key := fmt.Sprintf("prefix/key-%d", j)
 			_, _ = cache.Insert(key, data)
 		}
 


internal/cache/lru/lru.go
```diff
@@ -291,9 +308,16 @@ func (c *Cache) UpdateSize(key string, sizeDelta uint64) error {
...
 	for key := range c.index {
 		if strings.HasPrefix(key, prefix) {
-			c.Erase(key)
+			keysToDelete = append(keysToDelete, key)
 		}
 	}
+	c.mu.RUnlock()
+
+	if len(keysToDelete) > 0 {
+		c.eraseKeys(keysToDelete)
```
vadlakondaswetha, line 321: This was intentional at that time to avoid blocking other operations if there are too many objects to delete

kislaykishore: Are you suggesting we should continue to take locks for every delete operation? Also, is there any backing data for this assumption?

internal/cache/lru/lru.go
```diff
@@ -291,9 +308,16 @@ func (c *Cache) UpdateSize(key string, sizeDelta uint64) error {
 }
 
 func (c *Cache) EraseEntriesWithGivenPrefix(prefix string) {
+	c.mu.RLock()
```
vadlakondaswetha, line 311: All other methods take c.mu.Lock, so effectively we are blocking all other operations. Can we make read operations to use Rlock.

internal/cache/lru/lru.go
vadlakondaswetha: I am checking the code - I have 2 things.
1. It is called from fast_stat_bucket where a lock is taken. That lock is exclusive for all operations to stat_cache. how did we get this crash issue.
2. This method is called in the rename flow, clearing the cache of all entries in the folder.  A single directory can have any number of entries, now blocking all other operations till the data is deleted will slow down.

kislaykishore: 1. Understood. Then it'll probably not cause a crash. I'll double check. I was looking at this module as a unit. Doesn't it look like a code-smell?
2. Deleting a million entries is expected to take a fraction of a second. Should we worry about it?

vadlakondaswetha: 1. yes, its a code-smell we should fix it
2. If we have the numbers that it will take <1sec for million entires, we should be good.

kislaykishore: Added tests to verify the same.
``` 
 % go test -v -bench=BenchmarkEraseEntriesWithGivenPrefix_1M -run=^$ -benchmem ./internal/cache/lru/...
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/cache/lru
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkEraseEntriesWithGivenPrefix_1Million
BenchmarkEraseEntriesWithGivenPrefix_1Million-16    	       3	 393560310 ns/op	44714304 B/op	      33 allocs/op
PASS
ok  	github.com/googlecloudplatform/gcsfuse/v3/internal/cache/lru	8.525s

```

internal/cache/lru/lru.go
kislaykishore: I think the read-only operations already are taking Rlock. Let me know if that's not the case.

PR #4519, opened 2026-03-23
refactor(improved cd pipeline script): modify improved cd script

PranjalC100:
### Description
Refactors improved_e2e_test.sh into a modular wrapper script that uses getopt for robust argument parsing and adds support for multiple Linux distributions (Debian, Ubuntu, RHEL, and CentOS). It dynamically determines the system architecture and OS to install the appropriate gcsfuse package before delegating test execution to improved_run_e2e_tests.sh.

https://b.corp.google.com/issues/496078241

### Testing details
1. Through Louhi

No


tools/cd_scripts/improved_e2e_test.sh
```diff
@@ -13,351 +13,160 @@
...
+# A long option name followed by a colon indicates it requires an argument.
+LONG=local-run,zonal,release-version:,output-dir:,help
+
+# Parse the options using getopt
+# --options "" specifies that there are no short options.
+PARSED=$(getopt --options "" --longoptions "$LONG" --name "$0" -- "$@")
+if [[ $? -ne 0 ]]; then
+    # getopt will have already printed an error message
+    usage 1
+fi
```
meet2mky, line 56:
```suggestion
# Parse the options using getopt
# --options "" specifies that there are no short options.
if ! PARSED=$(getopt --options "" --longoptions "$LONG" --name "$0" -- "$@"); then
    usage 1
fi
```

meet2mky: We are using set -e option so if getopt fails due to unknown arguments it's not going to print usage. So let's put it in the `if` block

tools/cd_scripts/improved_e2e_test.sh
```diff
@@ -13,351 +13,160 @@
...
+            RELEASE_VERSION="$2"
+            shift 2
+            # Regex breakdown:
+            # ^      : Start of string
+            # [0-9]+ : One or more digits
+            #  \.     : A literal dot
+            # $      : End of string
+            RE="^[0-9]+\.[0-9]+\.[0-9]+$"
+            if [[ ! $RELEASE_VERSION =~ $RE ]]; then
+                log_error "--release-version is incorrectly formatted."
```
meet2mky: Nit: we can add the value also for better logs.
```suggestion
                log_error "--release-version value '$RELEASE_VERSION' is incorrectly formatted."
```

tools/cd_scripts/improved_e2e_test.sh
```diff
@@ -13,351 +13,160 @@
...
-# Build GCSFuse from source if it's a local run or requires a custom build
-if [ "${LOCAL_RUN}" = "true" ] || [[ "${BUCKET_NAME_TO_USE}" != "gcsfuse-release-packages" ]]; then
-    echo "Building GCSFuse from source (Version: ${VERSION})..."
-    go run tools/build_gcsfuse/main.go . . "${VERSION}"
-    sudo cp bin/* /usr/bin/
-    sudo cp sbin/* /usr/sbin/
 fi
 
+# Set parallelism to 4
+ARGS+=( "--package-level-parallelism=4")
```
meet2mky:
```suggestion
# Set parallelism to 4 as it is optimal for all of the release VM(s).
ARGS+=("--package-level-parallelism=4")
```

meet2mky (dismissed):
LGTM, minor comments.

tools/cd_scripts/improved_e2e_test.sh
PranjalC100: Ohhh, understood!!

meet2mky (approved):
Looks good. Thanks!

PR #4520, opened 2026-03-24
feat(tracing): Add write flow traces

thrivikram-karur-g:
### Description
Adding traces in write flows to help quicker debugging.
Traces added are as below.

| Trace Name | Identifier | Description |
| :--- | :--- | :--- |
| **Write File Staged** | `write.staged` | Traces a write operation using the legacy staged writes. |
| **Write File Streaming** | `write.streaming` | Traces a write operation using the buffered writes handler. |
| **Sync File Staged** | `write.staged.sync` | Traces the synchronization of the staged temp file to GCS. |
| **Sync File Streaming** | `write.streaming.sync` | Traces the synchronization of buffered writes to GCS. |
| **Streaming Upload Block** | `write.streaming.upload.block` | Traces the upload of a single streaming block to GCS. |
| **Streaming Upload Finalize** | `write.streaming.upload.finalize` | Traces the finalization of the streaming upload. |
| **Streaming Upload Flush** | `write.streaming.upload.flush` | Traces the flushing of pending streaming writes. |
| **Streaming Uploader** | `write.streaming.uploader` | Traces the go routine uploading blocks to GCS when received |

Perf results run on the PR:

Screenshot 2026-04-10 at 6 45 39 PM

There is no degradation on the performance of writes seen.

Attaching golang benchmark test results for the traceHandle interface. Added a method Trace to the handle for better code reuse.

goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/tracing
cpu: Intel(R) Xeon(R) CPU @ 2.80GHz
BenchmarkTrace/BenchmarkStartSpan_Otel-64         	 6412791	       188.3 ns/op	     240 B/op	       3 allocs/op
BenchmarkTrace/BenchmarkStartServerSpan_Otel-64   	 5128497	       234.1 ns/op	     272 B/op	       5 allocs/op
BenchmarkTrace/BenchmarkRecordError_Otel-64       	265906203	         4.510 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithErrorNoBytes_Otel-64         	 4450281	       269.1 ns/op	     305 B/op	       4 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithErrorWithBytes_Otel-64       	 4457209	       267.2 ns/op	     305 B/op	       4 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorNoBytes_Otel-64      	 4598284	       260.4 ns/op	     305 B/op	       4 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorWithBytes_Otel-64    	 4535978	       262.5 ns/op	     305 B/op	       4 allocs/op
BenchmarkTrace/BenchmarkSetCacheReadAttributes_Otel-64              	56318166	        21.30 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkSetUploadAttributes_Otel-64                 	40176628	        29.71 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkPropagateTraceContext_Otel-64               	25519111	        45.39 ns/op	      48 B/op	       1 allocs/op
BenchmarkTrace/BenchmarkStartSpan_Noop-64                           	504248462	         2.380 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkStartServerSpan_Noop-64                     	448551762	         2.675 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkRecordError_Noop-64                         	644497599	         1.861 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithErrorNoBytes_Noop-64         	447284846	         2.682 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithErrorWithBytes_Noop-64       	504384649	         2.386 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorNoBytes_Noop-64      	446925662	         2.684 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorWithBytes_Noop-64    	504111261	         2.379 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkSetCacheReadAttributes_Noop-64              	745226696	         1.572 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkSetUploadAttributes_Noop-64                 	799512030	         1.501 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkPropagateTraceContext_Noop-64               	534620713	         2.233 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/googlecloudplatform/gcsfuse/v3/tracing	23.932s

We check no-op is very performant to ensure tracing when disabled has minimal if not no impact to the performance.

b/454482833

### Testing details
1. Manual 

Triggered a single file upload with sequential writes for a new file and observed the following trace(Note: GCS writer doesn't take in context, so we only see one single clear trace). Along with the object name and the upload size details.

Screenshot 2026-04-10 at 5 08 20 PM

Triggered a single file upload with staged writes, below shows WriteFile making a read call to download the complete object from GCS 

Screenshot 2026-04-09 at 4 17 43 PM

Flush fs op shows the file content being uploaded to GCS from the tempFile.

Screenshot 2026-04-10 at 5 02 23 PM

2. Unit tests - NA
3. Integration tests - NA

N/A

### DIFF
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -105,7 +105,9 @@ func (uh *UploadHandler) Upload(block block.Block) error {
 	}
 	// Start the uploader goroutine but only once.
 	uh.startUploader.Do(func() {
+		_, span := uh.traceHandle.StartSpan(context.Background(), tracing.UploadAsync)
 		go uh.uploader()
+		uh.traceHandle.EndSpan(span)
 	})
 	uh.uploadCh <- block
 	return nil
@@ -157,7 +159,9 @@ func (uh *UploadHandler) uploader() {
 // If there is an error during upload, it returns after storing the error in uploadError.
 func (uh *UploadHandler) uploadBlock(b block.Block) {
 	_, span := uh.traceHandle.StartSpan(context.Background(), tracing.StreamingUploadBlock)
+	var written int64
 	defer func() {
+		uh.traceHandle.SetUploadAttributes(span, written, uh.objectName)
 		uh.traceHandle.EndSpan(span)
 	}()
 	if b == nil {
@@ -177,7 +181,8 @@ func (uh *UploadHandler) uploadBlock(b block.Block) {
 		return
 	}
 
-	_, err := io.Copy(uh.writer, b)
+	var err error
+	written, err = io.Copy(uh.writer, b)
 	if errors.Is(err, context.Canceled) {
 		// Context canceled error indicates that the file was deleted from the
 		// same mount. In this case, we suppress the error to match local
@@ -198,6 +203,8 @@ func (uh *UploadHandler) Finalize() (obj *gcs.MinObject, err error) {
 	defer func() {
 		if err != nil {
 			uh.traceHandle.RecordError(span, err)
+		} else if obj != nil {
+			uh.traceHandle.SetUploadAttributes(span, int64(obj.Size), uh.objectName)
 		}
 		uh.traceHandle.EndSpan(span)
 	}()
@@ -237,6 +244,8 @@ func (uh *UploadHandler) FlushPendingWrites() (o *gcs.MinObject, err error) {
 	defer func() {
 		if err != nil {
 			uh.traceHandle.RecordError(span, err)
+		} else if o != nil {
+			uh.traceHandle.SetUploadAttributes(span, int64(o.Size), uh.objectName)
 		}
 		uh.traceHandle.EndSpan(span)
 	}()
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -139,7 +139,8 @@ func (uh *UploadHandler) UploadError() (err error) {
 
 // uploader is the single-threaded goroutine that uploads blocks.
 func (uh *UploadHandler) uploader(ctx context.Context) {
-	_, span := uh.traceHandle.StartSpan(context.Background(), tracing.StreamingUploader)
+	_, _, finishSpan := uh.traceHandle.Trace(context.Background(), tracing.StreamingUploader, nil)
+	defer finishSpan()
 	for currBlock := range uh.uploadCh {
 		uh.uploadBlock(ctx, currBlock)
 
@@ -148,7 +149,6 @@ func (uh *UploadHandler) uploader(ctx context.Context) {
 		uh.blockPool.Release(currBlock)
 		uh.wg.Done()
 	}
-	uh.traceHandle.EndSpan(span)
 }
 
 // uploadBlock uploads the block content to GCS writer.
@@ -157,15 +157,12 @@ func (uh *UploadHandler) uploader(ctx context.Context) {
 // If there is already an error in uploadError, it returns without doing anything.
 // If there is an error during upload, it returns after storing the error in uploadError.
 func (uh *UploadHandler) uploadBlock(ctx context.Context, b block.Block) {
-	_, span := uh.traceHandle.StartSpan(ctx, tracing.StreamingUploadBlock)
 	var written int64
 	var err error
+	_, span, finishSpan := uh.traceHandle.Trace(ctx, tracing.StreamingUploadBlock, &err)
 	defer func() {
 		uh.traceHandle.SetUploadAttributes(span, written, uh.objectName)
-		if err != nil {
-			uh.traceHandle.RecordError(span, err)
-		}
-		uh.traceHandle.EndSpan(span)
+		finishSpan()
 	}()
 	if b == nil {
 		logger.Warnf("uploadBlock: received nil block for object %s", uh.objectName)
@@ -201,14 +198,12 @@ func (uh *UploadHandler) uploadBlock(ctx context.Context, b block.Block) {
 // Finalize finalizes the upload.
 func (uh *UploadHandler) Finalize(ctx context.Context) (obj *gcs.MinObject, err error) {
 	ctx = uh.traceHandle.PropagateTraceContext(context.Background(), ctx)
-	_, span := uh.traceHandle.StartSpan(ctx, tracing.StreamingUploadFinalize)
+	_, span, finishSpan := uh.traceHandle.Trace(ctx, tracing.StreamingUploadFinalize, &err)
 	defer func() {
-		if err != nil {
-			uh.traceHandle.RecordError(span, err)
-		} else if obj != nil {
+		if err == nil && obj != nil {
 			uh.traceHandle.SetUploadAttributes(span, int64(obj.Size), uh.objectName)
 		}
-		uh.traceHandle.EndSpan(span)
+		finishSpan()
 	}()
 	uh.wg.Wait()
 	close(uh.uploadCh)
@@ -241,14 +236,12 @@ func (uh *UploadHandler) ensureWriter(ctx context.Context) error {
 
 // FlushPendingWrites uploads any data in the write buffer.
 func (uh *UploadHandler) FlushPendingWrites(ctx context.Context) (o *gcs.MinObject, err error) {
-	_, span := uh.traceHandle.StartSpan(ctx, tracing.StreamingUploadFlush)
+	_, span, finishSpan := uh.traceHandle.Trace(ctx, tracing.StreamingUploadFlush, &err)
 	defer func() {
-		if err != nil {
-			uh.traceHandle.RecordError(span, err)
-		} else if o != nil {
+		if err == nil && o != nil {
 			uh.traceHandle.SetUploadAttributes(span, int64(o.Size), uh.objectName)
 		}
-		uh.traceHandle.EndSpan(span)
+		finishSpan()
 	}()
 	uh.wg.Wait()
 
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -210,14 +210,14 @@ func (uh *UploadHandler) Finalize(ctx context.Context) (obj *gcs.MinObject, err 
 	}
 
 	obj, err = uh.bucket.FinalizeUpload(ctx, uh.writer)
-	if obj != nil {
-		bytes = int64(obj.Size)
-	}
 	if err != nil {
 		// FinalizeUpload already returns GCSerror so no need to convert again.
 		uh.uploadError.Store(&err)
 		logger.Errorf("FinalizeUpload failed for object %s: %v", uh.objectName, err)
 		return nil, err
+	}
+	if obj != nil {
+		bytes = int64(obj.Size)
 	}
 	return obj, nil
 }
@@ -246,14 +246,14 @@ func (uh *UploadHandler) FlushPendingWrites(ctx context.Context) (o *gcs.MinObje
 	}
 
 	o, err = uh.bucket.FlushPendingWrites(ctx, uh.writer)
-	if o != nil {
-		bytes = int64(o.Size)
-	}
 	if err != nil {
 		// FlushUpload already returns GCS error so no need to convert again.
 		uh.uploadError.Store(&err)
 		logger.Errorf("FlushUpload failed for object %s: %v", uh.objectName, err)
 		return nil, err
+	}
+	if o != nil {
+		bytes = int64(o.Size)
 	}
 	return o, nil
 }
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -106,7 +106,7 @@ func (uh *UploadHandler) Upload(ctx context.Context, block block.Block) error {
 	// Start the uploader goroutine but only once.
 	uh.startUploader.Do(func() {
 		_, span := uh.traceHandle.StartSpan(context.Background(), tracing.UploadAsync)
-		go uh.uploader()
+		go uh.uploader(ctx)
 		uh.traceHandle.EndSpan(span)
 	})
 	uh.uploadCh <- block
@@ -140,9 +140,9 @@ func (uh *UploadHandler) UploadError() (err error) {
 }
 
 // uploader is the single-threaded goroutine that uploads blocks.
-func (uh *UploadHandler) uploader() {
+func (uh *UploadHandler) uploader(ctx context.Context) {
 	for currBlock := range uh.uploadCh {
-		uh.uploadBlock(currBlock)
+		uh.uploadBlock(ctx, currBlock)
 
 		// Put back the uploaded block to the pool for re-use,
 		// irrespective of whether the upload was successful or not.
@@ -156,8 +156,8 @@ func (uh *UploadHandler) uploader() {
 // If the block is nil, it logs a warning and returns.
 // If there is already an error in uploadError, it returns without doing anything.
 // If there is an error during upload, it returns after storing the error in uploadError.
-func (uh *UploadHandler) uploadBlock(b block.Block) {
-	_, span := uh.traceHandle.StartSpan(context.Background(), tracing.StreamingUploadBlock)
+func (uh *UploadHandler) uploadBlock(ctx context.Context, b block.Block) {
+	_, span := uh.traceHandle.StartSpan(ctx, tracing.StreamingUploadBlock)
 	var written int64
 	defer func() {
 		uh.traceHandle.SetUploadAttributes(span, written, uh.objectName)
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -96,10 +96,10 @@ func newUploadHandler(req *CreateUploadHandlerRequest) *UploadHandler {
 }
 
 // Upload adds a block to the upload queue.
-func (uh *UploadHandler) Upload(block block.Block) error {
+func (uh *UploadHandler) Upload(ctx context.Context, block block.Block) error {
 	uh.wg.Add(1)
 
-	err := uh.ensureWriter()
+	err := uh.ensureWriter(ctx)
 	if err != nil {
 		return fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
@@ -114,12 +114,11 @@ func (uh *UploadHandler) Upload(block block.Block) error {
 }
 
 // createObjectWriter creates a GCS object writer.
-func (uh *UploadHandler) createObjectWriter() (err error) {
+func (uh *UploadHandler) createObjectWriter(ctx context.Context) (err error) {
 	req := gcs.NewCreateObjectRequest(uh.obj, uh.objectName, nil, uh.chunkRetryDeadline, uh.chunkTransferTimeout)
 	// We need a new context here, since the first writeFile() call will be complete
 	// (and context will be cancelled) by the time complete upload is done.
-	var ctx context.Context
-	ctx, uh.cancelFunc = context.WithCancel(context.Background())
+	ctx, uh.cancelFunc = context.WithCancel(uh.traceHandle.PropagateTraceContext(context.Background(), ctx))
 	if uh.bucket.BucketType().Zonal && (uh.obj != nil && uh.obj.Finalized.IsZero()) {
 		chunkWriterReq := gcs.CreateObjectChunkWriterRequest{
 			CreateObjectRequest: *req,
@@ -197,8 +196,8 @@ func (uh *UploadHandler) uploadBlock(b block.Block) {
 }
 
 // Finalize finalizes the upload.
-func (uh *UploadHandler) Finalize() (obj *gcs.MinObject, err error) {
-	ctx := context.Background()
+func (uh *UploadHandler) Finalize(ctx context.Context) (obj *gcs.MinObject, err error) {
+	ctx = uh.traceHandle.PropagateTraceContext(context.Background(), ctx)
 	_, span := uh.traceHandle.StartSpan(ctx, tracing.StreamingUploadFinalize)
 	defer func() {
 		if err != nil {
@@ -213,7 +212,7 @@ func (uh *UploadHandler) Finalize() (obj *gcs.MinObject, err error) {
 
 	// Writer may not have been created for empty file creation flow or for very
 	// small writes of size less than 1 block.
-	err = uh.ensureWriter()
+	err = uh.ensureWriter(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
@@ -228,9 +227,9 @@ func (uh *UploadHandler) Finalize() (obj *gcs.MinObject, err error) {
 	return obj, nil
 }
 
-func (uh *UploadHandler) ensureWriter() error {
+func (uh *UploadHandler) ensureWriter(ctx context.Context) error {
 	if uh.writer == nil {
-		if err := uh.createObjectWriter(); err != nil {
+		if err := uh.createObjectWriter(ctx); err != nil {
 			return fmt.Errorf("createObjectWriter failed for object %s: %w", uh.objectName, err)
 		}
 	}
@@ -238,8 +237,8 @@ func (uh *UploadHandler) ensureWriter() error {
 }
 
 // FlushPendingWrites uploads any data in the write buffer.
-func (uh *UploadHandler) FlushPendingWrites() (o *gcs.MinObject, err error) {
-	ctx := context.Background()
+func (uh *UploadHandler) FlushPendingWrites(ctx context.Context) (o *gcs.MinObject, err error) {
+	ctx = uh.traceHandle.PropagateTraceContext(context.Background(), ctx)
 	_, span := uh.traceHandle.StartSpan(ctx, tracing.StreamingUploadFlush)
 	defer func() {
 		if err != nil {
@@ -253,7 +252,7 @@ func (uh *UploadHandler) FlushPendingWrites() (o *gcs.MinObject, err error) {
 
 	// Writer may not have been created for empty file creation flow or for very
 	// small writes of size less than 1 block.
-	err = uh.ensureWriter()
+	err = uh.ensureWriter(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -156,8 +156,7 @@ func (uh *UploadHandler) uploader() {
 // If there is already an error in uploadError, it returns without doing anything.
 // If there is an error during upload, it returns after storing the error in uploadError.
 func (uh *UploadHandler) uploadBlock(b block.Block) {
-	ctx := context.Background()
-	ctx, span := uh.traceHandle.StartSpan(ctx, tracing.StreamingUploadBlock)
+	_, span := uh.traceHandle.StartSpan(context.Background(), tracing.StreamingUploadBlock)
 	defer func() {
 		uh.traceHandle.EndSpan(span)
 	}()
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -139,7 +139,7 @@ func (uh *UploadHandler) UploadError() (err error) {
 
 // uploader is the single-threaded goroutine that uploads blocks.
 func (uh *UploadHandler) uploader(ctx context.Context) {
-	_, _, finishSpan := uh.traceHandle.Trace(context.Background(), tracing.StreamingUploader, nil)
+	_, finishSpan := uh.traceHandle.TraceUpload(context.Background(), tracing.StreamingUploader, "", nil, nil)
 	defer finishSpan()
 	for currBlock := range uh.uploadCh {
 		uh.uploadBlock(ctx, currBlock)
@@ -159,11 +159,9 @@ func (uh *UploadHandler) uploader(ctx context.Context) {
 func (uh *UploadHandler) uploadBlock(ctx context.Context, b block.Block) {
 	var written int64
 	var err error
-	_, span, finishSpan := uh.traceHandle.Trace(ctx, tracing.StreamingUploadBlock, &err)
-	defer func() {
-		uh.traceHandle.SetUploadAttributes(span, written, uh.objectName)
-		finishSpan()
-	}()
+	_, finishSpan := uh.traceHandle.TraceUpload(ctx, tracing.StreamingUploadBlock, uh.objectName, &written, &err)
+	defer finishSpan()
+
 	if b == nil {
 		logger.Warnf("uploadBlock: received nil block for object %s", uh.objectName)
 		return
@@ -198,13 +196,9 @@ func (uh *UploadHandler) uploadBlock(ctx context.Context, b block.Block) {
 // Finalize finalizes the upload.
 func (uh *UploadHandler) Finalize(ctx context.Context) (obj *gcs.MinObject, err error) {
 	ctx = uh.traceHandle.PropagateTraceContext(context.Background(), ctx)
-	_, span, finishSpan := uh.traceHandle.Trace(ctx, tracing.StreamingUploadFinalize, &err)
-	defer func() {
-		if err == nil && obj != nil {
-			uh.traceHandle.SetUploadAttributes(span, int64(obj.Size), uh.objectName)
-		}
-		finishSpan()
-	}()
+	bytes := int64(obj.Size)
+	_, finishSpan := uh.traceHandle.TraceUpload(ctx, tracing.StreamingUploadFinalize, uh.objectName, &bytes, &err)
+	defer finishSpan()
 	uh.wg.Wait()
 	close(uh.uploadCh)
 
@@ -236,13 +230,9 @@ func (uh *UploadHandler) ensureWriter(ctx context.Context) error {
 
 // FlushPendingWrites uploads any data in the write buffer.
 func (uh *UploadHandler) FlushPendingWrites(ctx context.Context) (o *gcs.MinObject, err error) {
-	_, span, finishSpan := uh.traceHandle.Trace(ctx, tracing.StreamingUploadFlush, &err)
-	defer func() {
-		if err == nil && o != nil {
-			uh.traceHandle.SetUploadAttributes(span, int64(o.Size), uh.objectName)
-		}
-		finishSpan()
-	}()
+	bytes := int64(o.Size)
+	_, finishSpan := uh.traceHandle.TraceUpload(ctx, tracing.StreamingUploadFlush, uh.objectName, &bytes, &err)
+	defer finishSpan()
 	uh.wg.Wait()
 
 	// Writer may not have been created for empty file creation flow or for very
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -89,9 +89,6 @@ func newUploadHandler(req *CreateUploadHandlerRequest) *UploadHandler {
 		chunkTransferTimeout: req.ChunkTransferTimeoutSecs,
 		traceHandle:          req.TraceHandle,
 	}
-	if uh.traceHandle == nil {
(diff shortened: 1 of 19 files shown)


internal/bufferedwrites/upload_handler.go
```diff
@@ -80,6 +84,10 @@ func newUploadHandler(req *CreateUploadHandlerRequest) *UploadHandler {
 		obj:                  req.Object,
 		blockSize:            req.BlockSize,
 		chunkTransferTimeout: req.ChunkTransferTimeoutSecs,
+		traceHandle:          req.TraceHandle,
+	}
+	if uh.traceHandle != nil {
+		uh.traceHandle = tracing.NewNoopTracer()
 	}
```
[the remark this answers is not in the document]
thrivikram-karur-g: thanks, must be a typo

thrivikram-karur-g:
working on checking how context needs to passed down from parent which makes the call

internal/bufferedwrites/upload_handler.go
```diff
@@ -83,33 +87,38 @@ func newUploadHandler(req *CreateUploadHandlerRequest) *UploadHandler {
...
-	err := uh.ensureWriter()
+	err := uh.ensureWriter(ctx)
 	if err != nil {
 		return fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
 	// Start the uploader goroutine but only once.
 	uh.startUploader.Do(func() {
+		_, span := uh.traceHandle.StartSpan(context.Background(), tracing.UploadAsync)
 		go uh.uploader()
+		uh.traceHandle.EndSpan(span)
```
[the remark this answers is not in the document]
thrivikram-karur-g: @kislaykishore , I feel this is not a very useful span and I am removing this.

internal/fs/inode/file.go
```diff
@@ -922,7 +943,14 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
...
+	_, span := f.traceHandle.StartSpan(ctx, tracing.SyncFileStaged)
+	defer func() {
+		if err != nil {
+			f.traceHandle.RecordError(span, err)
+		}
+		f.traceHandle.EndSpan(span)
+	}()
 	var latestGcsObj *gcs.Object
 	if !f.local {
 		var err error
```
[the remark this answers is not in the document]
thrivikram-karur-g: resolved

internal/bufferedwrites/upload_handler.go
```diff
@@ -148,6 +157,12 @@ func (uh *UploadHandler) uploader() {
 // If there is already an error in uploadError, it returns without doing anything.
 // If there is an error during upload, it returns after storing the error in uploadError.
 func (uh *UploadHandler) uploadBlock(b block.Block) {
+	_, span := uh.traceHandle.StartSpan(context.Background(), tracing.StreamingUploadBlock)
```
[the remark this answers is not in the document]
thrivikram-karur-g: resolved

internal/fs/inode/file.go
```diff
@@ -675,6 +682,13 @@ func (f *FileInode) Write(
...
+	_, span := f.traceHandle.StartSpan(ctx, tracing.WriteFileStaged)
+	defer func() {
+		if err != nil {
+			f.traceHandle.RecordError(span, err)
+		}
+		f.traceHandle.EndSpan(span)
+	}()
 	// Make sure f.content != nil.
 	err = f.ensureContent(ctx)
 	if err != nil {
```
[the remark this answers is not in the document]
thrivikram-karur-g: resolved

internal/fs/inode/file.go
```diff
@@ -693,7 +707,12 @@ func (f *FileInode) writeUsingTempFile(ctx context.Context, data []byte, offset
 //
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) writeUsingBufferedWrites(ctx context.Context, data []byte, offset int64) (bool, error) {
-	err := f.bwh.Write(data, offset)
+	_, span := f.traceHandle.StartSpan(ctx, tracing.WriteFileStreaming)
+	f.traceHandle.SetUploadAttributes(span, int64(len(data)), f.src.Name)
+	defer func() {
+		f.traceHandle.EndSpan(span)
+	}()
+	err := f.bwh.Write(ctx, data, offset)
```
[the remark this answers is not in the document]
thrivikram-karur-g: resolved

internal/fs/inode/file.go
```diff
@@ -922,7 +943,14 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
...
 // LOCKS_REQUIRED(f.mu)
-func (f *FileInode) syncUsingContent(ctx context.Context) error {
+func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
+	_, span := f.traceHandle.StartSpan(ctx, tracing.SyncFileStaged)
+	defer func() {
+		if err != nil {
+			f.traceHandle.RecordError(span, err)
+		}
+		f.traceHandle.EndSpan(span)
+	}()
```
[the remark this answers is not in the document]
thrivikram-karur-g: resolved

internal/bufferedwrites/upload_handler.go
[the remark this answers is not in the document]
thrivikram-karur-g: keeping this for now, after you take a look, we can decide if it needs to be kept or removed.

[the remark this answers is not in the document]
kislaykishore: What purpose is this span serving?

internal/bufferedwrites/upload_handler.go
```diff
@@ -131,9 +140,9 @@ func (uh *UploadHandler) UploadError() (err error) {
 }
 
 // uploader is the single-threaded goroutine that uploads blocks.
-func (uh *UploadHandler) uploader() {
+func (uh *UploadHandler) uploader(ctx context.Context) {
```
kislaykishore, line 138: Should we have a span tracking this function?

internal/bufferedwrites/upload_handler.go
```diff
@@ -165,7 +180,8 @@ func (uh *UploadHandler) uploadBlock(b block.Block) {
 		return
 	}
 
-	_, err := io.Copy(uh.writer, b)
+	var err error
+	written, err = io.Copy(uh.writer, b)
```
kislaykishore, line 179: Should we record the error as well in the span?

internal/bufferedwrites/upload_handler.go
```diff
@@ -180,18 +196,28 @@ func (uh *UploadHandler) uploadBlock(b block.Block) {
 }
 
 // Finalize finalizes the upload.
-func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
+func (uh *UploadHandler) Finalize(ctx context.Context) (obj *gcs.MinObject, err error) {
+	ctx = uh.traceHandle.PropagateTraceContext(context.Background(), ctx)
```
kislaykishore, line 195: Could you explain with a comment as to why we are creating a new context for the span.

internal/bufferedwrites/upload_handler.go
```diff
@@ -201,27 +227,37 @@ func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
...
 			return fmt.Errorf("createObjectWriter failed for object %s: %w", uh.objectName, err)
 		}
 	}
 	return nil
 }
 
 // FlushPendingWrites uploads any data in the write buffer.
-func (uh *UploadHandler) FlushPendingWrites() (*gcs.MinObject, error) {
+func (uh *UploadHandler) FlushPendingWrites(ctx context.Context) (o *gcs.MinObject, err error) {
+	ctx = uh.traceHandle.PropagateTraceContext(context.Background(), ctx)
```
kislaykishore: Same here - please add a comment explaining why a new context needs to be created.

internal/fs/inode/file.go
```diff
@@ -705,13 +725,15 @@ func (f *FileInode) writeUsingBufferedWrites(ctx context.Context, data []byte, o
...
-		err = f.flushUsingBufferedWriteHandler()
+		err = f.flushUsingBufferedWriteHandler(ctx)
 		if err != nil {
+			f.traceHandle.RecordError(span, err)
 			return false, fmt.Errorf("could not finalize what has been written so far: %w", err)
 		}
 		return true, f.writeUsingTempFile(ctx, data, offset)
 	}
 	if err != nil {
+		f.traceHandle.RecordError(span, err)
```
kislaykishore: Instead of recording errors at multiple places, I think you can use defer to look at the error value and record it in the span.

internal/fs/inode/file.go
```diff
@@ -922,10 +944,22 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
 // object, syncs the content and updates the inode state.
 //
 // LOCKS_REQUIRED(f.mu)
-func (f *FileInode) syncUsingContent(ctx context.Context) error {
+func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
+	ctx, span := f.traceHandle.StartSpan(ctx, tracing.SyncFileStaged)
+	st, err := f.content.Stat()
```
kislaykishore: Why do we need to make this Stat() call?

internal/fs/tracing_test.go
```diff
@@ -899,10 +899,14 @@ func (s *TracingTestSuite) TestTraceWriteFile() {
...
-				assert.Equal(t, spanName, ss[i].Name)
-				assert.Equal(t, trace.SpanKindServer, ss[i].SpanKind)
+			require.GreaterOrEqual(t, len(ss), len(tt.spans))
+			spanNamesInSS := make(map[string]trace.SpanKind)
+			for _, s := range ss {
+				spanNamesInSS[s.Name] = s.SpanKind
+			}
+			for _, spanName := range tt.spans {
+				assert.Contains(t, spanNamesInSS, spanName, "span %s not found", spanName)
+				assert.Equal(t, spanNamesInSS[spanName], trace.SpanKindServer)
```
kislaykishore: Expected value comes first.
```suggestion
				assert.Equal(t, trace.SpanKindServer, spanNamesInSS[spanName])
```

tracing/span_attributes.go
```diff
@@ -15,6 +15,8 @@
 package tracing
 
 const (
-	IS_CACHE_HIT = "cache.hit" // Indicates if the response was served from cache or not.
-	BYTES_READ   = "read.size" // Indicates the number of bytes read from the given reader
+	IS_CACHE_HIT   = "cache.hit"          // Indicates if the response was served from cache or not.
+	BYTES_READ     = "read.size"          // Indicates the number of bytes read from the given reader
+	BYTES_UPLOADED = "upload.size"        // Indicates the number of bytes uploaded
+	OBJECT_NAME    = "upload.object_name" // Indicates the object name uploaded
```
kislaykishore: Why not "write"?
```suggestion
	BYTES_UPLOADED = "write.chunk.size"        // Indicates the number of bytes uploaded
	OBJECT_NAME    = "write.object_name" // Indicates the object name uploaded
```

tracing/span_names.go
```diff
@@ -114,4 +114,23 @@ const (
...
+	// SyncFileStreaming traces the synchronization of buffered writes to GCS.
+	SyncFileStreaming = "sync.streaming"
+	// StreamingUploadBlock traces the upload of a single streaming block to GCS.
+	StreamingUploadBlock = "streaming.upload.block"
+	// StreamingUploadFinalize traces the finalization of the streaming upload.
+	StreamingUploadFinalize = "streaming.upload.finalize"
+	// StreamingUploadFlush traces the flushing of pending streaming writes.
+	StreamingUploadFlush = "streaming.upload.flush"
+	// Tracks the go routine and async upload of the write blocks received
+	UploadAsync = "upload.async"
```
kislaykishore: Let's review if this is needed.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: I think this would be a better choice than the span wrap around go routine trigger, will add here and remove there

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: thinking aloud and doing some git blame, I think in this case the pre-existing context.Background does not seem to be a intended for ensuring no cancellation propagation, actually removing this and passing context as it is.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: Thinking aloud, I do not think a new context is needed here either as well.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: yes, sounds like a good thing to add. have added it.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: yes, updated defer code to record error as well.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: as commented, not relevant and removed this

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: updated

internal/fs/inode/file.go
thrivikram-karur-g: updated

internal/fs/inode/file.go
thrivikram-karur-g: this is not a real stat call to GCS, but rather gets the Size details of the gcs.tempFile struct locally created. This is to log the bytesUploaded attribute.

internal/fs/tracing_test.go
thrivikram-karur-g: updated 👍

tracing/span_attributes.go
thrivikram-karur-g: most terminology was upload but makes sense from tracing perpective. updated

tracing/span_names.go
thrivikram-karur-g: I do not think it is needed, especially now that we are tracing the uploader function which is actually what is triggered in a go routine.

thrivikram-karur-g: removed it

tracing/span_names.go
```diff
@@ -114,4 +114,23 @@ const (
...
+	// SyncFileStaged traces the synchronization of the staged temp file to GCS.
+	SyncFileStaged = "sync.staged"
+	// SyncFileStreaming traces the synchronization of buffered writes to GCS.
+	SyncFileStreaming = "sync.streaming"
+	// StreamingUploadBlock traces the upload of a single streaming block to GCS.
+	StreamingUploadBlock = "streaming.upload.block"
+	// StreamingUploadFinalize traces the finalization of the streaming upload.
+	StreamingUploadFinalize = "streaming.upload.finalize"
+	// StreamingUploadFlush traces the flushing of pending streaming writes.
+	StreamingUploadFlush = "streaming.upload.flush"
```
kislaykishore: These appear inconsistent - should we have "write.streaming" as the prefix?

internal/fs/inode/file.go
```diff
@@ -922,10 +945,22 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
...
```
kislaykishore: While the cost of doing this stat call would likely be small, I'll recommend we don't make this additional call to get size unless we have confirmed the performance characteristics even if it means that we don't have the size.
Alternatively, we can make this call depending on whether tracing is enabled.

thrivikram-karur-g: I can add the tracing check for now.

tracing/span_names.go
thrivikram-karur-g: yeah, let us make it write.streaming & write.staged prefix based everywhere.

internal/fs/inode/file.go
thrivikram-karur-g: I have seen that tracing check also has it's own complications, we can temporarily put it as -1 indicating that it means that full file upload has been triggered.

tracing/span_names.go
thrivikram-karur-g: updated

internal/fs/inode/file.go
thrivikram-karur-g: updated

internal/fs/inode/file.go
thrivikram-karur-g: as discussed, wrapped this call with tracingEnabled to check to ensure no impact to the write flows when tracing is disabled.

internal/fs/inode/file.go
```diff
@@ -918,14 +941,40 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
...
 
+// get the temp content size when tracing is enabled to set the BYTES_UPLOADED attribute
+func (f *FileInode) getTempContentSizeForSpan() int64 {
+	if !cfg.IsTracingEnabled(f.config) {
+		return -1
+	}
+
+	st, err := f.content.Stat()
+	if err != nil {
+		logger.Errorf("stat: %v", err)
```
kislaykishore: Add more details to this log message so that it's clear where the error is.

internal/fs/inode/file.go
```diff
@@ -918,14 +941,40 @@ func (f *FileInode) Sync(ctx context.Context) (gcsSynced bool, err error) {
...
+
 // syncUsingContent syncs the inode content to GCS. It fetches the latest GCS
 // object, syncs the content and updates the inode state.
 //
 // LOCKS_REQUIRED(f.mu)
-func (f *FileInode) syncUsingContent(ctx context.Context) error {
+func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
+	ctx, span := f.traceHandle.StartSpan(ctx, tracing.SyncFileStaged)
+	if err != nil {
+		return fmt.Errorf("stat: %w", err)
```
kislaykishore: Please add more context regarding the error.

kislaykishore (commented):
Please ensure that the PR description is consistent with the code.

internal/fs/inode/file.go
thrivikram-karur-g: updated

internal/fs/inode/file.go
thrivikram-karur-g: updated

kislaykishore: I don't see it being updated.

thrivikram-karur-g:
> Please ensure that the PR description is consistent with the code.

Updated description with trace output screenshots as well & how they appear in trace explorer

internal/fs/inode/file.go
thrivikram-karur-g: this error no longer makes sense, this is removed. you can take a look above in the same file, you will see a more descriptive error.

kislaykishore (dismissed):
LGTM. Please get an approval from Swetha before submitting.

internal/fs/inode/file.go
```diff
@@ -721,17 +744,17 @@ func (f *FileInode) writeUsingBufferedWrites(ctx context.Context, data []byte, o
...
 	if errors.As(err, &preconditionErr) {
 		return &gcsfuse_errors.FileClobberedError{
-			Err:        fmt.Errorf("f.bwh.Flush(): %w", err),
+			Err:        fmt.Errorf("f.bwh.Flush(ctx): %w", err),
 			ObjectName: f.src.Name,
 		}
 	}
 	if err != nil {
-		return fmt.Errorf("f.bwh.Flush(): %w", err)
+		return fmt.Errorf("f.bwh.Flush(ctx): %w", err)
```
vadlakondaswetha: can. you share a sample error log. I want to understand why do we need to add ctx as part of this message

internal/bufferedwrites/upload_handler.go
```diff
@@ -131,23 +138,35 @@ func (uh *UploadHandler) UploadError() (err error) {
...
+	uh.traceHandle.EndSpan(span)
 }
 
 // uploadBlock uploads the block content to GCS writer.
 // It is called by the uploader goroutine.
 // If the block is nil, it logs a warning and returns.
 // If there is already an error in uploadError, it returns without doing anything.
 // If there is an error during upload, it returns after storing the error in uploadError.
-func (uh *UploadHandler) uploadBlock(b block.Block) {
+func (uh *UploadHandler) uploadBlock(ctx context.Context, b block.Block) {
```
vadlakondaswetha, line 156: every method is polluted with the same set of code. I prefer doing this in a generic fashion to avoid complicating the code. Gemini suggests this. See if we can make it more generic.

In tracing/trace_handle.go:
type TraceHandle interface {
    // ... existing methods ...
    
    // Trace starts a span and returns a finisher function.
    // Use it like: ctx, span, done := th.Trace(ctx, name, &err); defer done()
    Trace(ctx context.Context, name string, err *error) (context.Context, trace.Span, func())
}

In tracing/otel_tracer.go:
func (o *otelTracer) Trace(ctx context.Context, name string, err *error) (context.Context, trace.Span, func()) {
    ctx, span := o.StartSpan(ctx, name)
    return ctx, span, func() {
        if err != nil && *err != nil {
            o.RecordError(span, *err)
        }
        o.EndSpan(span)
    }
}

Eg:
func (f *FileInode) writeUsingTempFile(ctx context.Context, data []byte, offset int64) (err error) {
    ctx, span, done := f.traceHandle.Trace(ctx, tracing.WriteFileStaged, &err)
    defer done()
    
    // Set specific attributes after starting
    f.traceHandle.SetUploadAttributes(span, int64(len(data)), f.src.Name)
    
    // ... logic ...
}

vadlakondaswetha (commented):
can you also upload a large file like 100MB and share the traces.

internal/fs/inode/file.go
thrivikram-karur-g: actually it does not print the context, it is just modification of the function signature in the log given now it has context as a parameter. anyways, we can keep the log same as well without modification as it is not very useful.

thrivikram-karur-g: removed and kept it same.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: Refactored it as recommended by gemini, also added golang benchmark tests to verify that the new method to traceHandle interface is very performant.

thrivikram-karur-g:
> can you also upload a large file like 100MB and share the traces.

The trace screenshots in the PR are for files of size 1G, the fio command run was.

`fio --name=append_test --ioengine=libaio --rw=write --bs=1M --size=1G --numjobs=1 --direct=1 --filename=test_file_12.dat`

if you are asking for the link, I am sharing it offline.

internal/bufferedwrites/upload_handler.go
```diff
@@ -201,27 +225,34 @@ func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
...
 	}
 	return nil
 }
 
 // FlushPendingWrites uploads any data in the write buffer.
-func (uh *UploadHandler) FlushPendingWrites() (*gcs.MinObject, error) {
+func (uh *UploadHandler) FlushPendingWrites(ctx context.Context) (o *gcs.MinObject, err error) {
+	_, span, finishSpan := uh.traceHandle.Trace(ctx, tracing.StreamingUploadFlush, &err)
+	defer func() {
+		if err == nil && o != nil {
```
vadlakondaswetha: This will simplify the uploading attributes also:

func TraceUpload(th TraceHandle, ctx context.Context, name string, objName string, bytes *int64, err *error) (context.Context, func()) {
    ctx, span := th.StartSpan(ctx, name)
    return ctx, func() {
        if bytes != nil {
            th.SetUploadAttributes(span, *bytes, objName)
        }
        if err != nil && *err != nil {
            th.RecordError(span, *err)
        }
        th.EndSpan(span)
    }
}

func (uh *UploadHandler) uploadBlock(ctx context.Context, b block.Block) {
    var written int64
    var err error
    ctx, done := tracing.TraceUpload(uh.traceHandle, ctx, tracing.StreamingUploadBlock, uh.objectName, &written, &err)
    defer done()

    // ... logic ...
    written, err = io.Copy(uh.writer, b)
    // ...
}


is this feasible to do ?

thrivikram-karur-g: yes, should be feasible. I can update.

tracing/benchmark_test.go
```diff
@@ -68,6 +68,26 @@ func BenchmarkTrace(b *testing.B) {
...
+				done()
+			}
+		})
+
+		b.Run(fmt.Sprintf("BenchmarkTraceWithoutError_%s", prefix), func(b *testing.B) {
+			ctx := context.Background()
+			err := fmt.Errorf("TestError")
+
+			for b.Loop() {
+				_, _, done := th.Trace(ctx, "TestSpanName", &err)
```
thrivikram-karur-g: fix this.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: I have run golang b.Loop benchmark tests and this is highly inefficient.
We have other way though to add it as an interface method which is feasible

thrivikram-karur-g: These are the benchmark results below.

" " "
BenchmarkTrace/BenchmarkTraceUploadWithErrorNoBytes_Noop-64         	19654833	        58.96 ns/op	      80 B/op	       1 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithErrorWithBytes_Noop-64       	20406697	        58.71 ns/op	      80 B/op	       1 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorNoBytes_Noop-64      	21114004	        58.01 ns/op	      80 B/op	       1 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorWithBytes_Noop-64    	20737923	        57.47 ns/op	      80 B/op	       1 allocs/op
" " "

Test run.

		b.Run(fmt.Sprintf("BenchmarkTraceUploadWithErrorWithBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			err := fmt.Errorf("TestError")
			bytes := int64(33554432)

			for b.Loop() {
				_, finishSpan := TraceUpload(th, ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, &err)
				finishSpan()
			}
		})

thrivikram-karur-g: benchmark tests run below.

`

        b.Run(fmt.Sprintf("TestBenchmarkTraceUploadWithErrorNoBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			err := fmt.Errorf("TestError")
			bytes := int64(0)

			for b.Loop() {
				_, finishSpan := TraceUpload(th, ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, &err)
				finishSpan()
			}
		})
		b.Run(fmt.Sprintf("TestBenchmarkTraceUploadWithErrorWithBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			err := fmt.Errorf("TestError")
			bytes := int64(33554432)

			for b.Loop() {
				_, finishSpan := TraceUpload(th, ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, &err)
				finishSpan()
			}
		})

		b.Run(fmt.Sprintf("TestBenchmarkTraceUploadWithoutErrorNoBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			bytes := int64(0)

			for b.Loop() {
				_, finishSpan := TraceUpload(th, ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, nil)
				finishSpan()
			}
		})

		b.Run(fmt.Sprintf("TestBenchmarkTraceUploadWithoutErrorWithBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			bytes := int64(33554432)

			for b.Loop() {
				_, finishSpan := TraceUpload(th, ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, nil)
				finishSpan()
			}
		})

		b.Run(fmt.Sprintf("BenchmarkTraceUploadWithErrorNoBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			err := fmt.Errorf("TestError")
			bytes := int64(0)

			for b.Loop() {
				_, finishSpan := th.TraceUpload(ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, &err)
				finishSpan()
			}
		})

		b.Run(fmt.Sprintf("BenchmarkTraceUploadWithErrorWithBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			err := fmt.Errorf("TestError")
			bytes := int64(33554432)

			for b.Loop() {
				_, finishSpan := th.TraceUpload(ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, &err)
				finishSpan()
			}
		})

		b.Run(fmt.Sprintf("BenchmarkTraceUploadWithoutErrorNoBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			bytes := int64(0)

			for b.Loop() {
				_, finishSpan := th.TraceUpload(ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, nil)
				finishSpan()
			}
		})

		b.Run(fmt.Sprintf("BenchmarkTraceUploadWithoutErrorWithBytes_%s", prefix), func(b *testing.B) {
			ctx := context.Background()
			bytes := int64(33554432)

			for b.Loop() {
				_, finishSpan := th.TraceUpload(ctx, "TestSpanName", "A/B/C/test_file.text", &bytes, nil)
				finishSpan()
			}
		})`
		
Results:
BenchmarkTrace/TestBenchmarkTraceUploadWithErrorNoBytes_Noop-64         	20345972	        56.17 ns/op	      80 B/op	       1 allocs/op
BenchmarkTrace/TestBenchmarkTraceUploadWithErrorWithBytes_Noop-64       	21374916	        57.69 ns/op	      80 B/op	       1 allocs/op
BenchmarkTrace/TestBenchmarkTraceUploadWithoutErrorNoBytes_Noop-64      	21641968	        55.15 ns/op	      80 B/op	       1 allocs/op
BenchmarkTrace/TestBenchmarkTraceUploadWithoutErrorWithBytes_Noop-64    	21605773	        56.69 ns/op	      80 B/op	       1 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithErrorNoBytes_Noop-64             	447856023	         2.679 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithErrorWithBytes_Noop-64           	448953206	         2.672 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorNoBytes_Noop-64          	447635182	         2.677 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrace/BenchmarkTraceUploadWithoutErrorWithBytes_Noop-64        	448691851	         2.675 ns/op	       0 B/op	       0 allocs/op

Having this method on the traceHandle is necessary for efficiency.

tracing/benchmark_test.go
thrivikram-karur-g: fixed

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: have updated but with this method on traceHandle

kislaykishore: Makes sense - it's inefficient because it causes heap-escape.

internal/bufferedwrites/upload_handler.go
```diff
@@ -180,18 +194,25 @@ func (uh *UploadHandler) uploadBlock(b block.Block) {
...
 	// small writes of size less than 1 block.
-	err := uh.ensureWriter()
+	err = uh.ensureWriter(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
 
-	obj, err := uh.bucket.FinalizeUpload(context.Background(), uh.writer)
+	obj, err = uh.bucket.FinalizeUpload(ctx, uh.writer)
+	if obj != nil {
```
vadlakondaswetha: this check should be after err != nil check

internal/bufferedwrites/upload_handler.go
```diff
@@ -201,27 +222,33 @@ func (uh *UploadHandler) Finalize() (*gcs.MinObject, error) {
...
 	// small writes of size less than 1 block.
-	err := uh.ensureWriter()
+	err = uh.ensureWriter(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
 
-	o, err := uh.bucket.FlushPendingWrites(context.Background(), uh.writer)
+	o, err = uh.bucket.FlushPendingWrites(ctx, uh.writer)
+	if o != nil {
```
vadlakondaswetha: same here also

internal/bufferedwrites/upload_handler.go
```diff
@@ -83,33 +87,36 @@ func newUploadHandler(req *CreateUploadHandlerRequest) *UploadHandler {
 		blockSize:            req.BlockSize,
 		chunkRetryDeadline:   req.ChunkRetryDeadlineSecs,
 		chunkTransferTimeout: req.ChunkTransferTimeoutSecs,
+		traceHandle:          req.TraceHandle,
+	}
+	if uh.traceHandle == nil {
```
vadlakondaswetha: why should the upload handler be worried about tracehandle being nil or not. Wherever the intialization is being done, there only we should set it to NewNoOpTracer if tracer is not enabled.

internal/bufferedwrites/upload_handler.go
```diff
@@ -131,9 +138,11 @@ func (uh *UploadHandler) UploadError() (err error) {
 }
 
 // uploader is the single-threaded goroutine that uploads blocks.
-func (uh *UploadHandler) uploader() {
+func (uh *UploadHandler) uploader(ctx context.Context) {
+	_, finishSpan := uh.traceHandle.TraceUpload(context.Background(), tracing.StreamingUploader, "", nil, nil)
```
vadlakondaswetha, line 139: Please do the similar optimizatons for read flow as well in a seperate PR

vadlakondaswetha (commented):
minor comments. otherwise LGTM

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: sure, will pick it up in a separate PR. 👍

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: we wanted to be extra careful that any wrong initializations of incomplete initializations of trace handler should not break the code. It should never happen but wanted to ensure there is no way for it to happen with this in place.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: makes sense. updating.

internal/bufferedwrites/upload_handler.go
thrivikram-karur-g: makes sense, updating

vadlakondaswetha (approved):

internal/bufferedwrites/upload_handler.go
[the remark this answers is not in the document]
rkachewar: Did we not test it before raising PR or have a unit test?

PR #4526, opened 2026-03-25
test(e2e): test packages to correctly parse and utilize the common configuration file

PranjalC100:
### Description

This PR updates several end-to-end integration test packages to correctly parse and utilize the common configuration file (`test_config.yaml`). It also includes a bug fix for object listing verification in the `managed_folders` package.

**Key changes include:**

*   **Go Test Package Migrations:**
    *   **Cloud Profiler:** Updated `TestMain` to dynamically replace labels and service names in the configuration using the test environment's values.
    *   **Managed Folders:** Updated flag replacement logic to correctly identify and replace the `--key-file` flag.
    *   **Rapid Appends:** Added logic to handle comma-separated flags within the configuration by converting them to space-separated fields.
    *   **Requester Pays:** Updated the replacement logic for billing projects and key files to match the new flag format.
*   **Bug Fix in `managed_folders`:**
    *   Fixed an issue in `test_helper.go` where the code incorrectly attempted to access `objs[3]` when only one object was expected.
    *   Added missing `return` statements after error logging to prevent nil pointer dereferences.
*   **Configuration Update:**
    *   Updated `test_config.yaml` to include the `${PROFILE_SERVICE_NAME}` placeholder for the cloud profiler configuration.

[b/497647007](https://buganizer.corp.google.com/issues/497647007)

### Testing details

1.  **Integration tests:** Confirmed that migrated packages (`cloud_profiler`, `managed_folders`, `rapid_appends`, and `requester_pays_bucket`) run successfully with the configuration file approach.

### Any backward incompatible change?

None.

### DIFF
--- a/tools/integration_tests/improved_run_e2e_tests.sh
+++ b/tools/integration_tests/improved_run_e2e_tests.sh
@@ -729,15 +729,16 @@ test_package() {
 
   # Build go package test command.
   local go_test_cmd_parts=()
-  if ${SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE}; then
-    go_test_cmd_parts+=("-short")
-  fi
+
   if ${USE_COMMON_CONFIG}; then
     go_test_cmd_parts+=("BUCKET_NAME=${bucket_name}")
   fi
   go_test_cmd_parts+=("GODEBUG=asyncpreemptoff=1" "go" "test" "-v" "-timeout=${INTEGRATION_TEST_PACKAGE_TIMEOUT_IN_MINS}m" "${INTEGRATION_TEST_PACKAGE_DIR}/${package_name}")
   if [[ "$package_name" == "benchmarking" ]]; then
     go_test_cmd_parts+=("-bench=." "-benchtime=100x")
+  fi
+  if ${SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE}; then
+    go_test_cmd_parts+=("-short")
   fi
   # Test Binary flags after this.
   go_test_cmd_parts+=("-args" "--integrationTest")
--- a/tools/integration_tests/improved_run_e2e_tests.sh
+++ b/tools/integration_tests/improved_run_e2e_tests.sh
@@ -727,17 +727,9 @@ test_package() {
   local bucket_type="$3"
   local attempt_number="$4"
 
-  local is_migrated=true
-  for unmigrated in "${UNMIGRATED_PACKAGES[@]}"; do
-    if [[ "$package_name" == "$unmigrated" ]]; then
-      is_migrated=false
-      break
-    fi
-  done
-
   # Build go package test command.
   local go_test_cmd_parts=()
-  if ${CONFIG_FILE_RUN} && ${is_migrated}; then
+  if ${CONFIG_FILE_RUN}; then
     go_test_cmd_parts+=("BUCKET_NAME=${bucket_name}")
   fi
   go_test_cmd_parts+=("GODEBUG=asyncpreemptoff=1" "go" "test" "-v" "-timeout=${INTEGRATION_TEST_PACKAGE_TIMEOUT_IN_MINS}m" "${INTEGRATION_TEST_PACKAGE_DIR}/${package_name}")
@@ -747,7 +739,7 @@ test_package() {
   # Test Binary flags after this.
   go_test_cmd_parts+=("-args" "--integrationTest")
 
-  if ${CONFIG_FILE_RUN} && ${is_migrated}; then
+  if ${CONFIG_FILE_RUN}; then
     local config_file_path
     config_file_path=$(realpath "${INTEGRATION_TEST_PACKAGE_DIR}/test_config.yaml")
     go_test_cmd_parts+=("--config-file=${config_file_path}")
@@ -778,7 +770,7 @@ test_package() {
   test_package_log_file=$(create_file_helper "running_package_logs/${bucket_type}/${package_name}_attempt_${attempt_number}.txt")
   # Run the package test command and capture log output with runtime stats.
   log_info "Started running test package [$package_name] for bucket type [$bucket_type] with bucket name [$bucket_name] (Attempt: $attempt_number)"
-  log_info "Test is run using test_config.yaml:$is_migrated"
+  log_info "Test is run using test_config.yaml:${CONFIG_FILE_RUN}"
   log_info "Test Command: $go_test_cmd"
 
   if ! eval "$go_test_cmd" > "$test_package_log_file" 2>&1; then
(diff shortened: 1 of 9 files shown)


tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -350,6 +356,18 @@ TEST_PACKAGES_FOR_ZB=("${TEST_PACKAGES_COMMON[@]}" "rapid_appends" "unfinalized_
...
 
+# Packages that have not yet been migrated to the new config file method.
+readonly UNMIGRATED_PACKAGES=(
+  "buffered_read"
+  "negative_stat_cache"
+  "readonly_creds"
+  "mount_timeout"
+  "managed_folders"
+  "concurrent_operations"
+  "interrupt"
```
PranjalC100:
```
managed_folders, concurrent_operations need to be migrated soon, as they are run in GKE as well.
```

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -23,6 +23,7 @@ usage() {
   echo "    --test-installed-package                     Test installed gcsfuse package. (Default: false)"
   echo "    --install-package-from-path   <path>         Google Cloud Storage bucket path for GCSFuse package for testing (e.g. gs://<bucket-name>/my-gcsfuse-package.rpm)"
   echo "                                                 This option is mutually exclusive with --test-installed-package. (Default: "")"
+  echo "    --config-file-run                            Run migrated test packages using the config file method. (Default: false)"
```
meet2mky: can we rename this to `--use-common-config` and the description to 
`"Run test packages with the common config being used for GKE/GCSFuse test packages."`

meet2mky: make changes to variables as well accordingly.

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -651,15 +657,25 @@ test_package() {
   local bucket_type="$3"
 
   # Build go package test command.
-  local go_test_cmd_parts=("GODEBUG=asyncpreemptoff=1" "go" "test" "-v" "-timeout=${INTEGRATION_TEST_PACKAGE_TIMEOUT_IN_MINS}m" "${INTEGRATION_TEST_PACKAGE_DIR}/${package_name}")
-  if ${SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE}; then
-    go_test_cmd_parts+=("-short")
+  local go_test_cmd_parts=()
+  if ${CONFIG_FILE_RUN}; then
+    go_test_cmd_parts+=("BUCKET_NAME=${bucket_name}")
```
meet2mky: This part looks weird. You are populating on the environment variable in config file using this way whereas other variables like CLOUD_PROFILER, etc and MOUNTED_DIR are not populated like this. Is there any uniform way to do this?

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -653,16 +673,34 @@ test_package() {
...
+
   # Build go package test command.
-  local go_test_cmd_parts=("GODEBUG=asyncpreemptoff=1" "go" "test" "-v" "-timeout=${INTEGRATION_TEST_PACKAGE_TIMEOUT_IN_MINS}m" "${INTEGRATION_TEST_PACKAGE_DIR}/${package_name}")
-  if ${SKIP_NON_ESSENTIAL_TESTS_ON_PACKAGE}; then
-    go_test_cmd_parts+=("-short")
+  local go_test_cmd_parts=()
+  if ${CONFIG_FILE_RUN} && ${is_migrated}; then
+    go_test_cmd_parts+=("BUCKET_NAME=${bucket_name}")
   fi
+  go_test_cmd_parts+=("GODEBUG=asyncpreemptoff=1" "go" "test" "-v" "-timeout=${INTEGRATION_TEST_PACKAGE_TIMEOUT_IN_MINS}m" "${INTEGRATION_TEST_PACKAGE_DIR}/${package_name}")
```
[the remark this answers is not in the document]
meet2mky: +1

tools/integration_tests/improved_run_e2e_tests.sh
```diff
@@ -683,6 +699,8 @@ test_package() {
   test_package_log_file=$(create_file_helper "running_package_logs/${bucket_type}/${package_name}.txt")
   # Run the package test command and capture log output with runtime stats.
   log_info "Started running test package [$package_name] for bucket type [$bucket_type] with bucket name [$bucket_name]"
+  log_info "Test is run using test_config.yaml:${CONFIG_FILE_RUN}"
```
meet2mky: Let's remove these as it would clutter the output more with same repeated infomration. Instead of this add a single Log when you parse the config file flag and print once that this e2e test run is using common config.

tools/integration_tests/improved_run_e2e_tests.sh
PranjalC100: Here is a breakdown of the variables in the `test_config.yaml` file and why only `BUCKET_NAME` needs to be updated:

*   **`MOUNTED_DIR`**, **`MOUNTED_DIR_SECONDARY`**, and **`ONLY_DIR`**: These are exclusively used by GKE.
*   **`PROFILE_LABEL`**: This is used in the `cloud_profiler` package. Since we use the default `testVersionName` and `testServiceName` for GCE environments, this variable is primarily intended for GKE. We don't need to set it for our test runs.
*   **`BILLING_PROJECT`**: This is used in the `requester_pays_bucket` package. We already hardcode the billing project when running in non-GKE environments, making this variable primarily for GKE as well.
*   **`KEY_FILE`**: We are already replacing this flag in the `managed_folders` package (which doesn't run in GKE), and it is also conditionally set within the code for the `requester_pays_bucket` package.
*   **`BUCKET_NAME`**: This is the only variable required for our setup.

Given this context, replacing only the `BUCKET_NAME` should be the correct approach.

tools/integration_tests/improved_run_e2e_tests.sh
PranjalC100: SGTM

tools/integration_tests/improved_run_e2e_tests.sh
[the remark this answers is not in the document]
PranjalC100: Added.

tools/integration_tests/managed_folders/test_helper.go
```diff
@@ -188,33 +188,36 @@ func listNonEmptyManagedFolders(t *testing.T) {
 			// numberOfObjects - 1
 			if len(objs) != 1 {
 				t.Errorf("Incorrect number of objects in the directory %s expected %d: got %d: ", dir.Name(), 1, len(objs))
+				return nil
 			}
 			// testBucket/NonEmptyManagedFoldersTest/managedFolder1/testFile  -- FileInNonEmptyManagedFoldersTest
 			if objs[0].Name() != FileInNonEmptyManagedFoldersTest || objs[0].IsDir() {
-				t.Errorf("Listed incorrect object expected %s: got %s: ", FileInNonEmptyManagedFoldersTest, objs[3].Name())
+				t.Errorf("Listed incorrect object expected %s: got %s: ", FileInNonEmptyManagedFoldersTest, objs[0].Name())
```
PranjalC100: Fixed a bug where it tried to print objs[3].Name() when it expected only 1 object

tools/integration_tests/improved_run_e2e_tests.sh
PranjalC100: Outdated comment btw. This file changes will be taken in a different PR.

meet2mky (approved):
Fix PR description based on current code changes and let's merge once the code changes in GKE are merged.

PR #4529, opened 2026-03-25
chore(emulator): Improve emulator script use pre-built binary and other things for stability.

meet2mky:
### Description

This PR refactors the emulator test script to improve the stability and reliability of integration testing

* Readiness Checks: Replaced fixed sleep intervals with a wait_for_emulator function that polls the storage emulator until it is ready to accept requests.
* Log Streaming: Captured emulator Docker logs into a local file and displayed them during the cleanup phase to simplify debugging of test failures.
* Sequential Execution: Configured tests to run sequentially (-p 1) to prevent high CPU usage from impacting stability.
* Increased Gunicorn Timeout: Set the Gunicorn timeout within the emulator container to 10 minutes to prevent worker processes from terminating under high load.
* Improved Argument Handling: Added support for passing a prebuilt gcsfuse directory as a positional argument
Logging Helpers: Introduced log_info and log_error functions for clearer and more consistent output formatting.

b/496187949

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - Part of presubmit run.

### DIFF
--- a/tools/integration_tests/emulator_tests/emulator_tests.sh
+++ b/tools/integration_tests/emulator_tests/emulator_tests.sh
@@ -13,20 +13,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Fail on any error
-set -eo pipefail
-
-# Display commands being run
+# Exit on error, treat unset variables as errors, and propagate pipeline errors.
+set -euo pipefail
 set -x
+
+# Script uses positional arguments:
+#   $1: testInstalledPackage (default: false)
+#   $2: gcsfuse prebuilt directory (default: "")
+
+# Logging Helpers
+log_info() {
+  echo "[INFO] $(date +"%Y-%m-%d %H:%M:%S"): $1"
+}
+
+log_error() {
+  echo "[ERROR] $(date +"%Y-%m-%d %H:%M:%S"): $1"
+}
+
+
+
+TEST_INSTALLED_PACKAGE=${1:-false}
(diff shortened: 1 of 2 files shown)


tools/integration_tests/emulator_tests/emulator_tests.sh
```diff
@@ -92,45 +117,79 @@ sudo docker pull $DOCKER_IMAGE
...
+    fi
+    sleep 1
+    count=$((count+1))
+  done
+  log_error "Emulator failed to become ready after $timeout seconds."
+  return 1
+}
+
+# Run the emulator container in the background and stream its logs to a file.
+# GUNICORN_CMD_ARGS="--timeout 600" is used to increase the Gunicorn timeout from the default 30 seconds
```
raj-prince, line 142: Do we know, what is this timeout for?

raj-prince (approved):
Looks good. just added a question.

tools/integration_tests/emulator_tests/emulator_tests.sh
meet2mky: This timeout is for `GUNICORN` worker process that is spawned when emulator starts. This process has default timeout of 30s which is too less and causes the worker to die and restart and which creates fresh state of GCS database with it. If during the emulator tests any time this worker dies then emulator tests will fail certainly. With this timeout emulator tests are more resilient and we removed the building of GCSFuse process for each package to make it even less cpu intensive.

PR #4547, opened 2026-03-30
perf:⚡ Bolt: Optimize integer to string conversion

kislaykishore:
💡 **What:** Replaced `fmt.Sprintf("%d", ...)` with `strconv.FormatInt(...)` and `strconv.Itoa(...)`.
🎯 **Why:** The `fmt.Sprintf` function uses reflection and parses the format string at runtime, making it significantly slower and more memory-intensive than direct type conversion using the `strconv` package. These integer-to-string conversions happen frequently enough in kernel param processing and file info cache key generation to warrant optimization.
📊 **Impact:** Reduces CPU time and garbage collection pressure by avoiding unnecessary allocations and reflection overhead.
🔬 **Measurement:** Can be verified via Go benchmarks, though the impact is localized to these specific hot paths. Tests for both `data` and `kernelparams` pass successfully.

---
*PR created automatically by Jules for task [11733946182043549447](https://jules.google.com/task/11733946182043549447) started by @kislaykishore*

### DIFF
--- a/internal/cache/data/file_info_benchmark_test.go
+++ b/internal/cache/data/file_info_benchmark_test.go
@@ -15,19 +15,9 @@
 package data
 
 import (
-	"fmt"
 	"testing"
 	"time"
 )
-
-func BenchmarkGetFileInfoKeyName_Original(b *testing.B) {
-	bucketCreationTime := time.Unix(TestTimeInEpoch, 0)
-	b.ResetTimer()
-	for b.Loop() {
-		unixTimeString := fmt.Sprintf("%d", bucketCreationTime.Unix())
-		_ = TestBucketName + unixTimeString + TestObjectName
-	}
-}
 
 func BenchmarkGetFileInfoKeyName_Optimized(b *testing.B) {
 	bucketCreationTime := time.Unix(TestTimeInEpoch, 0)


internal/cache/data/file_info.go
```diff
@@ -41,7 +41,7 @@ func GetFileInfoKeyName(objectName string, bucketCreationTime time.Time, bucketN
 	if bucketName == "" || objectName == "" {
 		return "", errors.New(InvalidKeyAttributes)
 	}
-	unixTimeString := fmt.Sprintf("%d", bucketCreationTime.Unix())
+	unixTimeString := strconv.FormatInt(bucketCreationTime.Unix(), 10)
 	return bucketName + unixTimeString + objectName, nil
```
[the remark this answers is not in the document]
kislaykishore: @Jules please check this comment.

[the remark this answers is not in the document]
kislaykishore: @jules While replacing fmt.Sprintf is a good performance improvement, we can optimize this further by avoiding the string concatenations on the next line, which cause additional memory allocations. Since this function is in a potentially hot path for cache key generation, building the string with a pre-allocated byte slice would be more efficient.
```
size := len(bucketName) + len(objectName) + 20
keyBytes := make([]byte, 0, size)
keyBytes = append(keyBytes, bucketName...)
keyBytes = strconv.AppendInt(keyBytes, bucketCreationTime.Unix(), 10)
keyBytes = append(keyBytes, objectName...)
return string(keyBytes), nil
```

internal/cache/data/file_info.go
```diff
@@ -41,8 +41,12 @@ func GetFileInfoKeyName(objectName string, bucketCreationTime time.Time, bucketN
 	if bucketName == "" || objectName == "" {
 		return "", errors.New(InvalidKeyAttributes)
 	}
-	unixTimeString := fmt.Sprintf("%d", bucketCreationTime.Unix())
-	return bucketName + unixTimeString + objectName, nil
+	size := len(bucketName) + len(objectName) + 20
```
kislaykishore, line 44: @jules could you add benchmarks that demonstrate the impact of this change?

internal/cache/data/file_info_benchmark_test.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"testing"
+	"time"
+)
+
+func BenchmarkGetFileInfoKeyName_Original(b *testing.B) {
+	bucketCreationTime := time.Unix(TestTimeInEpoch, 0)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
kislaykishore: @jules could you modernize
```suggestion
	for b.Loop() {
```

internal/cache/data/file_info_benchmark_test.go
```diff
@@ -0,0 +1,38 @@
...
+		unixTimeString := fmt.Sprintf("%d", bucketCreationTime.Unix())
+		_ = TestBucketName + unixTimeString + TestObjectName
+	}
+}
+
+func BenchmarkGetFileInfoKeyName_Optimized(b *testing.B) {
+	bucketCreationTime := time.Unix(TestTimeInEpoch, 0)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
kislaykishore: @jules could you Modernize - for
```suggestion
	for b.Loop() {
```

internal/cache/data/file_info_benchmark_test.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func BenchmarkGetFileInfoKeyName_Original(b *testing.B) {
+	bucketCreationTime := time.Unix(TestTimeInEpoch, 0)
+	b.ResetTimer()
+	for b.Loop() {
+		unixTimeString := fmt.Sprintf("%d", bucketCreationTime.Unix())
+		_ = TestBucketName + unixTimeString + TestObjectName
+	}
+}
```
kislaykishore: @jules remove this benchmark.

PR #4548, opened 2026-03-30
refactor(kokoro perf tests): increase timeout to 9hrs

PranjalC100:
### Description
Because of a recent addition of fio configs in our Kokoro perf tests, the perf test was timing out.

`ERROR: Aborting VM command due to timeout of 21600 seconds`
 
I am setting this timeout as 9 hours in this PR(current value is set as 6hrs). I still need to add Zonal tests to the perf tests

b/487460676

### Testing - via Kokoro


PranjalC100:
We will parallelize the perf tests instead.

PR #4558, opened 2026-03-31
fix(read flow): when metadata is cached due to list object api call

ashmeenkaur:
### Description
This PR fixes a bug where reads on non-rapid (legacy) buckets would fail with an input/output error if the object's metadata was previously cached via a List operation. The fix involves enabling size-checks for legacy buckets in the read flow.

### Background
The JSON List Object API for non-rapid buckets does not return the finalized object attribute (b/497852269) when using selective attributes. When metadata is cached via a list API call, objects are incorrectly identified as "unfinalized".

This causes the read flow to set req.SkipSizeChecks = true. Because GCSFuse attempts to use a Multi-Range Reader (MRR) when size checks are skipped, it encounters an Invalid MultiRangeDownloaderWrapper error, leading to a read failure.

Ref logs:
```
{"timestamp":{"seconds":1774889016,"nanos":859250248},"severity":"TRACE","message":"fuse_debug: Op 0x000000de        connection.go:474] <- OpenFile (inode 6, PID 2390362)","mount-id":"-0ec19849"}
{"timestamp":{"seconds":1774889016,"nanos":859341089},"severity":"TRACE","message":"fuse_debug: Op 0x000000de        connection.go:573] -> OpenFile (handle 17)","mount-id":"-0ec19849"}
{"timestamp":{"seconds":1774889016,"nanos":860076436},"severity":"TRACE","message":"fuse_debug: Op 0x000000e0        connection.go:474] <- ReadFile (inode 6, PID 2390362, handle 17, offset 0, 1048576 bytes)","mount-id":"-0ec19849"}
{"timestamp":{"seconds":1774889016,"nanos":860225889},"severity":"INFO","message":"!fh.readManager.Object().IsUnfinalized() = false","mount-id":"-0ec19849"}
{"timestamp":{"seconds":1774889016,"nanos":860319871},"severity":"INFO","message":"Reader type = RangeReaderType and req.SkipSizeChecks = true","mount-id":"-0ec19849"}
{"timestamp":{"seconds":1774889016,"nanos":860350168},"severity":"ERROR","message":"ReadFile: input/output error, fh.readManager.ReadAt: MultiRangeReader.readFromMultiRangeReader: Invalid MultiRangeDownloaderWrapper","mount-id":"-0ec19849"}
{"timestamp":{"seconds":1774889016,"nanos":860369490},"severity":"ERROR","message":"fuse: Op 0x000000e0 *fuseops.ReadFileOp] -> Error: \"input/output error\"","mount-id":"-0ec19849"}
```

Changes
- Modified shouldSkipSizeChecks to explicitly return false if the bucket is not a rapid (zonal) bucket.
- Added a new integration test TestReadFileWorksAfterListDir to verify that reads succeed after a directory listing.
- Updated unit tests in internal/fs/handle/file_test.go to cover rapid vs. non-rapid bucket scenarios.

[b/497852269](http://b/497852269)

### Testing details
1. Manual - manually verfied
2. Unit tests - added
3. Integration tests - added

NA

### DIFF
--- a/tools/integration_tests/operations/list_dir_test.go
+++ b/tools/integration_tests/operations/list_dir_test.go
@@ -194,9 +194,8 @@ func TestReadFileWorksAfterListDir(t *testing.T) {
 	_ = operations.ReadDirectory(testDir, t)
 	fh, err := os.OpenFile(path.Join(testDir, obj1), syscall.O_DIRECT, operations.FilePermission_0600)
 	require.NoError(t, err)
-	chunk := make([]byte, util.MiB)
-	n, err := fh.Read(chunk)
+	content, err := operations.ReadFileSequentially(fh, util.MiB)
 
 	require.NoError(t, err)
-	require.EqualValues(t, objSize, n)
+	require.EqualValues(t, objSize, len(content))
 }


anushka567:
LGTM on the code change

PR #4568, opened 2026-04-02
perf: ⚡ Bolt: [performance improvement] Optimize BytesToHigherMiBs using integer arithmetic

kislaykishore:
💡 **What**: Replaced the `math.Ceil(float64(bytes) / float64(bytesInOneMiB))` calculation in `util.BytesToHigherMiBs` with equivalent integer addition and bitwise shifting `(bytes + bytesInOneMiB - 1) >> 20`. I've also added a corresponding benchmark test (`util_benchmark_test.go`) utilizing the modern Go `b.Loop()` syntax.

🎯 **Why**: The original implementation forced slow `uint64` to `float64` conversions, a floating-point division, a function call to `math.Ceil`, and finally a float to integer conversion. Since `bytesInOneMiB` is a power of 2, this is computationally wasteful.

📊 **Impact**: Micro-benchmark indicates execution time drops from ~0.77ns/op to ~0.39ns/op. This reduces CPU overhead on a hot path where capacity constraints and limits are frequently calculated.

🔬 **Measurement**: Run `go test -bench=BenchmarkBytesToHigherMiBs ./internal/util/` to verify the execution time.

---
*PR created automatically by Jules for task [7645361226353554684](https://jules.google.com/task/7645361226353554684) started by @kislaykishore*

### DIFF
--- a/internal/util/util_benchmark_test.go
+++ b/internal/util/util_benchmark_test.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


internal/util/util_benchmark_test.go
```diff
@@ -0,0 +1,12 @@
+package util
```
kislaykishore, line 15: @jules add copyright message

kislaykishore: Copyright year is incorrect. It should be 2026.

PR #4570, opened 2026-04-02
fix: 🕵️‍♂️Sherlock: Fix response body resource leak token-source endpoint

kislaykishore:
🐛 Bug: What was going wrong (the root cause)
`proxyTokenSource.Token` in `internal/auth/token_source.go` fetches tokens via `http.Get`. It never closed the resulting response body, which leaks underlying TCP connections and file descriptors whenever the function is called. This leaks resources.

💡 Fix: How it was resolved
Added `defer resp.Body.Close()` immediately after confirming the `err` from `http.Get` is `nil`, ensuring the body gets closed on successful token fetch.

b/499872519

---
*PR created automatically by Jules for task [15327027354465123042](https://jules.google.com/task/15327027354465123042) started by @kislaykishore*

### DIFF
--- a/internal/auth/token_source.go
+++ b/internal/auth/token_source.go
@@ -74,13 +74,7 @@ func (ts proxyTokenSource) Token() (token *oauth2.Token, err error) {
 		return nil, err
 	}
 	defer func() {
-		closeErr := resp.Body.Close()
-		if closeErr == nil {
-			return
-		}
-		if err == nil {
-			err = fmt.Errorf("proxyTokenSource cannot close body: %w", closeErr)
-		}
+		_ = resp.Body.Close()
 	}()
 
 	body, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))


internal/auth/token_source.go
```diff
@@ -73,6 +73,9 @@ func (ts proxyTokenSource) Token() (token *oauth2.Token, err error) {
 		err = fmt.Errorf("proxyTokenSource cannot fetch token: %w", err)
 		return nil, err
 	}
+	// Sherlock: Add missing resp.Body.Close() to prevent file descriptor/TCP connection leak
+	// when proxyTokenSource.Token() is called repeatedly.
+	defer resp.Body.Close()
```
kislaykishore: @jules Error return value of `resp.Body.Close` is not checked (errcheck). Please remember to check for lint failures.

internal/auth/token_source_test.go
```diff
@@ -70,6 +71,60 @@ func TestProxyTokenSource_TokenFetch_ServerError(t *testing.T) {
 	assert.Nil(t, token)
 }
 
+func TestProxyTokenSource_TokenFetch_ClosesBody(t *testing.T) {
```
kislaykishore: @jules I think this test is unnecessary and too intrusive. Please get rid of this.

internal/auth/token_source.go
```diff
@@ -73,6 +73,9 @@ func (ts proxyTokenSource) Token() (token *oauth2.Token, err error) {
 		err = fmt.Errorf("proxyTokenSource cannot fetch token: %w", err)
 		return nil, err
 	}
+	// Sherlock: Add missing resp.Body.Close() to prevent file descriptor/TCP connection leak
+	// when proxyTokenSource.Token() is called repeatedly.
```
kislaykishore: @jules Please get rid of this comment.

patch.diff
kislaykishore, line 1: @jules Please remove this file.

internal/auth/token_source_test.go
kislaykishore: @jules Please delete the test.

internal/auth/token_source.go
kislaykishore: @jules delete this comment

internal/auth/token_source.go
```diff
@@ -73,6 +73,12 @@ func (ts proxyTokenSource) Token() (token *oauth2.Token, err error) {
 		err = fmt.Errorf("proxyTokenSource cannot fetch token: %w", err)
 		return nil, err
 	}
+	defer func() {
+		closeErr := resp.Body.Close()
+		if err == nil && closeErr != nil {
```
kislaykishore: @jules Simplify by using early returns.

internal/auth/token_source.go
meet2mky, line 101: Is it possible that now we can send a perfectly valid token as well as an error reported from the closing the response body. I am just thinking out loud here if it can cause any regression when we are not able to close response body correctly. Although I don't think it should happen in normal circumstances.

kislaykishore: Sounds good. I'll make the change to duck this error.

PR #4578, opened 2026-04-03
test(integration): add script to run migrated tests on mounted directory

PranjalC100:
### Description
This PR introduces a new bash script (`tools/integration_tests/run_migrated_tests_mounted_directory.sh`) to automate the execution of integration tests against a mounted GCS directory. 

Key features of the script:
- Parses `test_config.yaml` using `yq` to dynamically extract test configurations, mounted directories, and FUSE flags.
- Automatically builds the `gcsfuse` binary if it is not found before starting the tests.
- Safely handles dynamically creating buckets (supporting `flat`, `hns`, and `zonal` types, mounting the bucket, running the go test command with the appropriate arguments, and unmounting/deleting the temporary buckets via an exit trap.
- Supports running tests for a specific package (`--package`) or all packages (`--all`).
- Supports concurrent dual-mounting for testing configurations that require secondary flags.
- Dynamically skips tests if `run_on_gke` is set to false in the test configuration.
- Automatically enforces trace logging (`--log-severity=trace`) and handles --only-dir logic for specific tests.
- Generates and prints a rich execution summary table at the end of the test suite run.

[bug](https://buganizer.corp.google.com/issues/499225878)

### Testing details
1. Manual - Verified the script properly parses the YAML config, mounts the directory, runs the tests, and cleans up the mount afterward.
2. Unit tests - NA
3. Integration tests - NA

No.

### DIFF
--- a/tools/integration_tests/run_migrated_tests_mounted_directory.sh
+++ b/tools/integration_tests/run_migrated_tests_mounted_directory.sh
@@ -1,0 +1,447 @@
+#!/bin/bash
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


PranjalC100:
Consider adding an option to handle different bucket types. We could approach this in two ways:

* Accept different flag sets depending on the specific bucket type.
* Accept a single bucket type flag, and validate it against the compatible section (e.g., flat, hns, zonal) in the test config file before executing the tests."

PranjalC100:
Some tests require a log_file and trace enabled logs. Will have to set this up in this script as well. For example: buffered_read package

PR #4592, opened 2026-04-06
feat(observability): Use --token-url for metrics and trace exporter if no default cr…

iamsumit:
### Description

Metrics client doesn't use `--token-url` in case of default credentials are not present. This PR changes that.

https://github.com/GoogleCloudPlatform/gcsfuse/issues/4586

### Testing details
1. Manual - Everything works as expected and I could see the metrics being exported properly.
2. Unit tests - NA
3. Integration tests - NA


chaitanyamaili:
@thrivikram-karur-g can we get for review/input for this pr?

thrivikram-karur-g:
> @thrivikram-karur-g can we get for review/input for this pr?

Hello @chaitanyamaili , thanks for this contribution but as communicated we are discussing this internally and treat this as a feature request. We will be owning it if we accept the request.

PR #4611, opened 2026-04-10
fix(requester pays): Fix requester pays for gRPC and also setup integration tests with gRPC with explicit SA.

meet2mky:
### Description
This PR fixes a bug in gRPC protocol requester pays feature. Where the billing project is not populated correctly causing errors of the following types.

```
InvalidArgument desc = Bucket is a requester pays bucket but no user project provided.\nerror details: name = BadRequest field =  desc = Bucket is a requester pays bucket but no user project provided."}
```
### Testing
Created separate SA with the required permission to test the requester pays feature correctly. Previously ran tests used overly Privileged account which had `resourcemanager.projects.createBillingAssignment` permission on the project to which the bucket in test belonged. This causes the the requester pays tests to pass even if the billing project is not provided thus masking the failures.

The SA and relevant permissions have been setup accordingly for the `gcs-fuse-test` and `gcs-fuse-test-ml` prjoects.

Tests are passing on GKE as well: https://b.corp.google.com/issues/500550846#comment19

b/500550846

### Testing details
1. Manual - Yes
2. Unit tests - NA
3. Integration tests - Yes

NONE

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -31,6 +31,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/storageutil"
 	"google.golang.org/api/iterator"
+	"google.golang.org/grpc/metadata"
 )
 
 const FullFolderPathHNS = "projects/_/buckets/%s/folders/%s"
@@ -43,6 +44,7 @@ type bucketHandle struct {
 	bucketType           *gcs.BucketType
 	controlClient        StorageControlClient
 	finalizeFileForRapid bool
+	billingProject       string
 }
 
 func (bh *bucketHandle) Name() string {
@@ -375,6 +377,10 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 	if err != nil {
 		err = fmt.Errorf("error while setting attribute selection for List Object query :%w", err)
 		return
+	}
+
+	if bh.billingProject != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx, "x-goog-user-project", bh.billingProject)
 	}
 
 	itr := bh.bucket.Objects(ctx, query) // Returning iterator to the list of objects.
--- a/tools/integration_tests/requester_pays_bucket/setup_test.go
+++ b/tools/integration_tests/requester_pays_bucket/setup_test.go
@@ -95,12 +95,27 @@ func TestMain(m *testing.M) {
 		cfg.RequesterPaysBucket[0].Configs[0].Compatible = map[string]bool{"flat": true, "hns": true, "zonal": false}
 	}
 
-	// Replace ${BILLING_PROJECT} placeholder in flags with the default billing project.
-	// when not running in GKE environment.
+	testEnv.ctx = context.Background()
+
+	// When not running in GKE environment.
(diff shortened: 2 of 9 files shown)


internal/storage/bucket_handle.go
```diff
@@ -377,6 +379,10 @@ func (bh *bucketHandle) ListObjects(ctx context.Context, req *gcs.ListObjectsReq
 		return
 	}
 
+	if bh.billingProject != "" {
```
raj-prince, line 382: Why are we only setting in ListObjects? Ideally, it should be done in the Go SDK during dial call, irrespective of operations.

tools/integration_tests/requester_pays_bucket/setup_test.go
```diff
@@ -35,6 +36,20 @@ import (
...
+	// (e.g., "gcs-fuse-test" or "gcs-fuse-test-ml").
+	// The test expects a JSON key for this SA to be stored in Secret Manager
+	// within the same project, under the secret name specified by requesterPaysCredsSecretName.
+	// Note: The user or service account running the test package, as well as the
+	// requester-pays service account (requester-pays-tester), must be granted the
+	// Storage Admin and Service Usage Consumer roles on the project.
+	// For example, adding the Service Usage Consumer role to
+	// requester-pays-tester@gcs-fuse-test.iam.gserviceaccount.com in the gcs-fuse-test project.
+	requesterPaysServiceAccountName = "requester-pays-tester"
+	requesterPaysCredsSecretName    = "requester-pays-tester"
```
raj-prince, line 51: nit: "requester-pays-tester" --> "requester-pays-tester-secret"

tools/integration_tests/util/creds_tests/creds.go
```diff
@@ -92,7 +95,52 @@ func CreateCredentials(ctx context.Context) (serviceAccount, localKeyFilePath st
...
 	}
-	operations.CloseFile(file)
+	if err := file.Close(); err != nil {
+		log.Printf("Failed to close credentials file: %v", err)
+	}
+
+	return
+}
+
+func CreateCredentialsForSA(ctx context.Context, serviceAccountName, saCredentialsSecretName string) (serviceAccountEmail, localKeyFilePath string) {
```
raj-prince, line 67: Can we re-use the code in CreateCredentials & CreateCredentialsForSA?

raj-prince (commented):
Looks good overall.

Added couple of minor comments.

internal/storage/bucket_handle.go
meet2mky: Yes, Ideally, It should be in the Go-SDK. I will file the bug in their component to investigate and make a permanent fix for this. We will ask them to ensure they add this header only if it's not already added by GCSFuse to prevent any duplication of header. Once the Go-SDK fix is done we can remove the changes safely from our codebase.

tools/integration_tests/requester_pays_bucket/setup_test.go
meet2mky: I have already created the secret with the above name. So changing it would require us to create secret with different name. I think, this is fine to have SA name and secret name for the SA credentials to be same. We can change it later if this presents any problem which I think it shouldn't

tools/integration_tests/util/creds_tests/creds.go
meet2mky: Ack, Checking if this can be improved.

meet2mky: Done!

raj-prince (approved):

PR #4614, opened 2026-04-13
feat(hns): remove empty-dir check before HNS folder rename

PranjalC100:
### Description
This PR optimizes the `renameHierarchicalDir` operation for Hierarchical Namespace (HNS) buckets by removing the preliminary `checkDirNotEmpty` validation step. 

Previously, GCSFuse performed a recursive listing of the destination directory via `ReadDescendants` (translating to a `ListObjects` call with `MaxResults=2`) to verify it was empty before attempting a rename. 

By eliminating this redundant client-side check, we shift the validation burden to the native GCS backend. The logic now attempts to delete the destination directory directly; if the directory is not empty, GCS natively rejects the deletion with a `PreconditionError,` which we catch and translate to a POSIX `ENOTEMPTY` error. This saves a Class A List API operation, reducing latency and operational costs for customers, while maintaining strict POSIX compliance.

https://b.corp.google.com/issues/495920968

### Testing details
1. Manual - Ran custom Python `os.rename` tests against mounted HNS buckets with TRACE logs enabled. Verified that renaming to a non-empty directory natively fails and returns `ENOTEMPTY` without triggering any preliminary `ListObjects` calls. 
2. Unit tests - NA
3. Integration tests - Ran existing HNS rename integration tests to ensure expected POSIX behavior is maintained.

No backward incompatible changes. The error propagation to the OS (e.g., `ENOTEMPTY`) remains functionally identical.

image

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2700,6 +2700,7 @@ func (fs *fileSystem) renameHierarchicalDir(ctx context.Context, oldParent inode
 			if errors.As(deleteErr, &precondErr) {
 				return fuse.ENOTEMPTY
 			}
+			return fmt.Errorf("DeleteChildDir: %w", deleteErr)
 		}
 	}
 
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2656,18 +2656,6 @@ func (fs *fileSystem) ensureNoLocalFilesInDirectory(dir inode.BucketOwnedDirInod
 	return nil
 }
 
-func (fs *fileSystem) checkDirNotEmpty(dir inode.BucketOwnedDirInode, name string) error {
-	unexpected, err := dir.ReadDescendants(context.Background(), 1)
-	if err != nil {
-		return fmt.Errorf("read descendants of the new directory %q: %w", name, err)
-	}
-
-	if len(unexpected) > 0 {
-		return fuse.ENOTEMPTY
-	}
-	return nil
-}
-
 // Rename an old folder to a new folder in a hierarchical bucket. If the new folder already
 // exists and is non-empty, return ENOTEMPTY. If old folder have open files then return
 // ENOTSUP.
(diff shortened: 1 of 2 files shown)


internal/fs/fs.go
```diff
@@ -2700,17 +2688,19 @@ func (fs *fileSystem) renameHierarchicalDir(ctx context.Context, oldParent inode
...
 		newParent.Lock()
-		_ = newParent.DeleteChildDir(ctx, newName, false, newDirInode)
+		deleteErr := newParent.DeleteChildDir(ctx, newName, false, newDirInode)
 		newParent.Unlock()
+		if deleteErr != nil {
+			var precondErr *gcs.PreconditionError
+			if errors.As(deleteErr, &precondErr) {
+				return fuse.ENOTEMPTY
+			}
+		}
```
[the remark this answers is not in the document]
PranjalC100: Added a `return fmt.Errorf("DeleteChildDir: %w", deleteErr)` to catch all possible errors (similar to the other usage of DeleteChildDir in fs.go)

internal/fs/fs.go
```diff
@@ -2700,17 +2688,20 @@ func (fs *fileSystem) renameHierarchicalDir(ctx context.Context, oldParent inode
...
-
-		// This refers to an empty destination directory.
-		// The RenameFolder API does not allow renaming to an existing empty directory.
-		// To make this work, we delete the empty directory first from gcsfuse and then perform rename.
+		// The RenameFolder API does not allow renaming to an empty existing directory.
+		// To make this work, we attempt to delete the destination directory first.
+		// If it is non-empty, this deletion will fail with a PreconditionError,
+		// in which case we immediately return ENOTEMPTY.
 		newParent.Lock()
-		_ = newParent.DeleteChildDir(ctx, newName, false, newDirInode)
```
vadlakondaswetha, line 2713: do we have exisitng integration tests which are covering this scenario?

PranjalC100: Added the following test case in rename_dir_test package. 

TestRenameDirectoryWithExistingNonEmptyDestDirectory - this test case was missing. Added it following the AAA.

tools/integration_tests/rename_dir_limit/rename_dir_test.go
```diff
@@ -197,3 +197,24 @@ func TestRenameDirectoryWithExistingEmptyDestDirectory(t *testing.T) {
...
+	newDirPath := path.Join(testDir, "destNonEmptyDirectory")
+	operations.CreateDirectoryWithNFiles(1, newDirPath, PrefixTempFile, t)
+
+	// Go's Rename function does not support renaming a directory into an existing directory.
+	// To achieve this, we call a Python rename function as a workaround.
+	cmd := exec.Command("python3", "-c", fmt.Sprintf("import os; os.rename('%s', '%s')", oldDirPath, newDirPath))
+	output, err := cmd.CombinedOutput()
+
+	assert.Error(t, err)
+	assert.Contains(t, string(output), "Directory not empty")
```
vadlakondaswetha: is it possible to validate the error code? ENOTEMPTY

PranjalC100: Yes, doing that right now. Catching the error code and validating if it's equal to  `int(syscall.ENOTEMPTY)`

vadlakondaswetha (approved):

PR #4620, opened 2026-04-15
test(logs): enforce log-file strictness and centralize log filename parsing

PranjalC100:
## Description
This PR centralizes and strengthens log file handling within the integration test suite. It ensures that log files are consistently parsed with the correct `.log` suffix and that failures to open log files result in immediate test failure rather than silent errors during mounted directory tests.
##
### 1. The Root Cause
In `inactive_stream_timeout`, test validations check for the **absence** of activity logs via `doesNotHaveInactiveReaderClosedLogLineInLogFile`. Previously, if the target file handle was unavailable, the system printed an error but completely skipped blocking assertions because it evaluated `err == nil` as false:

**Original Unsafe Code (before fixes):**
```go
// Flawed logic inside Setup
_, err := hasInactiveReaderClosedLogLineInLogFile(t, objectName, logFile, startTime, endTime)
if err != nil { // This printed the error but allowed the test to continue
    fmt.Println("INACTIVE STREAM CHECK ERROR:", err)
}
if err == nil {
    // If the file was missing, err != nil, so this assertion was skipped!
    t.Fatalf("Unexpected 'Inactive Reader Closed' log message found...")
}
```
### The Output
The test package prints success `(PASS)` without actually verifying from the logs:

```
INACTIVE STREAM CHECK ERROR: failed to open log file: open /gcsfuse-tmp/TestTimeoutEnabledSuite.log: no such file or directory
--- PASS: TestTimeoutEnabledSuite (12.05s)
```

## Key Changes:

* **Strict Log File Access:** Replaced soft error returns with require.NoError in the inactive_stream_timeout test setup to ensure that a missing or inaccessible log file correctly fails the test.
* **Centralized Suffix Enforcement:** Modified ParseLogFileFromFlags to automatically append a .log extension to filenames extracted from flags if it is missing.
* **Improved Log Path Discovery:** Updated SetUpLogFilePath to attempt to infer log filenames from the first configuration block if they are not explicitly provided in the top-level flags.

## Link to the issue in case of a bug fix.
[link](https://b.corp.google.com/issues/502912150)

## Testing details
* Manual - Ran the mounted directory tests for inactive_stream_timeout.
* Unit tests - NA
* Integration tests - Validated that log paths are correctly constructed for both mounted directory and local environments and that tests fail appropriately when logs are missing.

## Any backward incompatible change? If so, please explain.
No.

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -851,15 +851,9 @@ func ParseLogFileFromFlags(flags []string) string {
 		parts := strings.Fields(flagStr)
 		for _, part := range parts {
 			if strings.HasPrefix(part, "--log-file=") {
-				logPath := strings.TrimPrefix(part, "--log-file=")
-				if logPath == "" {
-					continue
+				if logPath := strings.TrimPrefix(part, "--log-file="); logPath != "" {
+					return path.Base(logPath)
 				}
-				fileName := path.Base(logPath)
-				if !strings.HasSuffix(fileName, ".log") {
-					fileName += ".log"
-				}
-				return fileName
 			}
 		}
 	}
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -847,11 +847,19 @@ func OverrideFilePathsInFlagSet(t *test_suite.TestConfig, GCSFuseTempDirPath str
 
 func ParseLogFileFromFlags(flags []string) string {
 	for _, flagStr := range flags {
+		flagStr = strings.ReplaceAll(flagStr, ",", " ")
 		parts := strings.Fields(flagStr)
 		for _, part := range parts {
 			if strings.HasPrefix(part, "--log-file=") {
-				// Get just the filename from the path
-				return path.Base(strings.TrimPrefix(part, "--log-file="))
+				logPath := strings.TrimPrefix(part, "--log-file=")
+				if logPath == "" {
+					continue
+				}
+				fileName := path.Base(logPath)
+				if !strings.HasSuffix(fileName, ".log") {
+					fileName += ".log"
+				}
+				return fileName
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


tools/integration_tests/util/setup/setup.go
```diff
@@ -861,14 +869,25 @@ func ParseLogFileFromFlags(flags []string) string {
...
-		logFilePath = path.Join(GKETempDir, parsedLogFileName) + ".log"
+
+	// Infer log filename directly from the parsed config block.
+	if parsedLogFileName == "" && cfg != nil && len(cfg.Configs) > 0 {
+		parsedLogFileName = ParseLogFileFromFlags(cfg.Configs[0].Flags)
+	}
+
+	// Default logFile name.
+	if parsedLogFileName == "" {
+		parsedLogFileName = "gcsfuse.log"
```
PranjalC100: I'm setting the default name to `gcsfuse-tmp.log` to maintain backward compatibility and prevent unexpected breakages across existing environments.

tools/integration_tests/util/setup/setup.go
```diff
@@ -861,14 +869,25 @@ func ParseLogFileFromFlags(flags []string) string {
...
+
+	// Infer log filename directly from the parsed config block.
+	if parsedLogFileName == "" && cfg != nil && len(cfg.Configs) > 0 {
+		parsedLogFileName = ParseLogFileFromFlags(cfg.Configs[0].Flags)
+	}
+
+	// Default logFile name.
+	if parsedLogFileName == "" {
+		parsedLogFileName = "gcsfuse-tmp.log"
+	}
```
meet2mky, line 874: In which cases we would fallback to default logfile name?

tools/integration_tests/util/setup/setup.go
```diff
@@ -847,11 +847,19 @@ func OverrideFilePathsInFlagSet(t *test_suite.TestConfig, GCSFuseTempDirPath str
...
-				return path.Base(strings.TrimPrefix(part, "--log-file="))
+				logPath := strings.TrimPrefix(part, "--log-file=")
+				if logPath == "" {
+					continue
+				}
+				fileName := path.Base(logPath)
+				if !strings.HasSuffix(fileName, ".log") {
+					fileName += ".log"
+				}
+				return fileName
```
meet2mky: Why can't we just enforce the file name to either end in .log file in the config file and in the flags in our test setup. Is there any need to add .log suffix if everything is managed properly in config file and direct flags.

meet2mky (commented):
few comments.

tools/integration_tests/util/setup/setup.go
PranjalC100: No, there's no need to add the suffix .log then. Should I verify and correct the logfile names then? 
This might act as a fallback incase someone adds new tests, and doesn't add `.log` in the filename.

meet2mky: Yes, it's better to just use the log file whatever is given in the --log-file arg. It doesn't necessarily needs to have .log extension.

tools/integration_tests/util/setup/setup.go
PranjalC100: This keeps the default logfile name whenever parsedLogFilename is empty (which happens when we don't explicitly set the --log-file flag, however we still use the default logfile for all packages). 

Previously, we were implicitly defaulting to `gcsfuse-tmp.log` by just adding a `.log` suffix to the directory path(`gcsfuse-tmp`). I've updated the code to make this default fallback more obvious.

PranjalC100: Gemini had suggested me this. Line 876

meet2mky (approved):
LGTM

PR #4628, opened 2026-04-16
feat: making direct-path verification non-fatal until dummy-stat calls becomes reliable

raj-prince:
### Description
- Making direct-path verification (during mount) non-fatal to stop mount failure because of unreliable dummy-stat calls. We may need to increase overall timeout with stall support to make the dummy-stat call reliable.
- Removing the ctx-timeout while client creation, given it's necessary for grpc traffic and we don't want to fail because of delay in client creation.

b/503303141

### Testing details
1. Manual - Yes, verified manually. (a) grpc with error in dummy-stat (b) rapid with error in dummy-stat (c) grpc with success in dummy-stat (d) rapid with success in dummy-stat
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -57,7 +57,7 @@ const (
 
 	// DirectPath detection parameters - used for fast-fail detection during client creation
 	directPathDetectionMaxAttempts      = 5
-	directPathDetectionTimeout          = 5 * time.Minute
+	directPathDetectionTimeout          = 10 * time.Second
 	directPathDetectionMaxBackoff       = 5 * time.Second
 	directPathDetectionMaxRetryDuration = 1 * time.Minute
 )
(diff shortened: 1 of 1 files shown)


internal/storage/storage_handle.go
```diff
@@ -454,7 +462,7 @@ func (sh *storageClient) createNonBidiGRPCClientWithHttpFallback(ctx context.Con
 	}
 
 	var err error
-	sh.grpcClient, err = createGRPCClientHandle(ctx, &sh.clientConfig, false, bucketName, billingProject)
+	sh.grpcClient, err = createGRPCClientHandle(ctx, &sh.clientConfig, false, bucketName, billingProject, false)
```
abhishek10004: IIRC, we decided to skip this for regional as well for now.

internal/storage/storage_handle.go
```diff
@@ -217,21 +218,16 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
-		setRetryConfig(ctx, sc, clientConfig)
 	}
+	setRetryConfig(ctx, sc, clientConfig)
 
-	err = verifyDirectPathConnectivity(ctx, clientConfig, bucketName, sc, billingProject)
-	unSetDirectPathEnvVariable()
-	if err != nil {
-		return nil, err
+	// TODO(b/503624405): Make the direct-path verification fatal after making the dummy-stat reliable.
+	if verifyErr := verifyDirectPathConnectivity(ctx, clientConfig, bucketName, sc, billingProject); verifyErr != nil {
```
abhishek10004, line 229: IIRC, this context would not have any timeout. Can this cause any issues like permanent stall?

raj-prince: This may cause, but we don't have other option. We need to wait for the client to be ready.

abhishek10004: Discussed with Prince. verifyDirectPathConnectivity uses a different context with timeout for stat call. Hence, resolving this.

internal/storage/storage_handle.go
```diff
@@ -217,21 +218,16 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
+		logger.Warnf("DirectPath verification failed, continuing without DirectPath guarantee: %v", verifyErr)
+		// Apply production retry config since verification skipped it on failure.
```
abhishek10004: This can be put in defer in verifyDirectPathConnectivity since we're manually restoring this at the end (line 260). So, we can remove that and put it under defer so that it gets applied in all cases.

abhishek10004 (dismissed):
LGTM. Left a minor comment.

abhishek10004 (approved):
LGTM

PR #4630, opened 2026-04-17
feat(metrics): Adding streaming write fallback count metric

thrivikram-karur-g:
### Description
Adding streaming write fallback metrics that tracks the reason for the fallback and also which file mode & flag it fell back to staging writes. Useful for understanding why streaming writes fallback hasn't been picked .

b/499889281

### Testing details
1. Manual 

Ran GCSFuse with streaming writes enabled.

Executed the command:  

Existing File fallback test: 
echo "test_data" > existing_file_test.txt on a new file.  Grepped the prometheus endpoint for the metrics using below:   curl -s http://localhost:8080/metrics | grep "streaming_write_fallback" 
No output for first time given that the file is new and streaming writes are allowed, got output like below:  fs_streaming_write_fallback_count{open_mode="write_only",write_fallback_reason="existing_file"} 1  Executed the command multiple times in loops and given each time streaming writes cannot be used for the same reason of existing file, saw the below.  fs_streaming_write_fallback_count{open_mode="write_only",write_fallback_reason="existing_file"} 2  fs_streaming_write_fallback_count{open_mode="write_only",write_fallback_reason="existing_file"} 7  fs_streaming_write_fallback_count{open_mode="write_only",write_fallback_reason="existing_file"} 14   echo "test_data" > not_enabled_test.txt  got an output like below:  fs_streaming_write_fallback_count{open_mode="write_only",write_fallback_reason="not_enabled"} 2  Ran the following to test the concurrent limit breached.  for i in {1..10}; do    dd if=/dev/zero of=limit_breached_test_$i.txt bs=1M count=10 &  done; wait  Got the output for the metric fallback with appropriate reason. 
fs_streaming_write_fallback_count{open_mode="write_only",write_fallback_reason="concurrent_limit_breached"} 7
 

similarly tested for not enabled and out_of_order_writes

2. Unit tests - added
3. Integration tests - NA

N/A


internal/fs/inode/file.go
```diff
@@ -1098,6 +1104,31 @@ func (f *FileInode) Truncate(
...
+		}
+	case util.WriteOnly:
+		if openMode.IsAppend() {
+			metricOpenMode = metrics.OpenModeWriteOnlyAppendAttr
+		} else {
+			metricOpenMode = metrics.OpenModeWriteOnlyAttr
+		}
+	}
+	if metricOpenMode != "" {
+		f.metricHandle.FsStreamingWriteFallbackCount(1, metricOpenMode, reason)
```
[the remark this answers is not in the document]
thrivikram-karur-g: There is no mention of histogram, cummulative distribution is a commonly used term here, so we are good.

internal/fs/inode/file.go
```diff
@@ -1058,6 +1063,7 @@ func (f *FileInode) truncateUsingBufferedWriteHandler(ctx context.Context, size
 	// If truncate size is less than the total file size resulting in OutOfOrder write, finalize and fall back to temp file.
 	if errors.Is(err, bufferedwrites.ErrOutOfOrderWrite) {
 		logger.Infof("Out of order write detected. File %s will now use legacy staged writes. "+StreamingWritesSemantics, f.name.String())
+		f.recordFallback(util.NewOpenMode(util.WriteOnly, 0), metrics.WriteFallbackReasonOutOfOrderAttr)
```
thrivikram-karur-g: It is hard to understand which file mode contributes to this. Either add a fallback like this or some unknown or not applicable.

internal/fs/metrics_test.go
```diff
@@ -137,7 +145,7 @@ func createWithContents(ctx context.Context, t *testing.T, bucket gcs.Bucket, na
 }
 
 func waitForMetricsProcessing() {
-	time.Sleep(5 * time.Millisecond)
+	time.Sleep(50 * time.Millisecond)
```
thrivikram-karur-g: Timeout increase not required

internal/fs/inode/file.go
```diff
@@ -1098,6 +1104,31 @@ func (f *FileInode) Truncate(
 	return false, f.truncateUsingTempFile(ctx, size)
 }
 
+// recordFallback maps util.OpenMode to metrics.OpenMode and records the fallback metric.
+func (f *FileInode) recordFallback(openMode util.OpenMode, reason metrics.WriteFallbackReason) {
+	if f.metricHandle == nil {
+		return
+	}
```
kislaykishore: Let's not add it here. This should be a safe change because we know that fs.metricHandle is never nil.
```suggestion
```

internal/fs/inode/file.go
```diff
@@ -1098,6 +1104,31 @@ func (f *FileInode) Truncate(
...
+		} else {
+			metricOpenMode = metrics.OpenModeReadWriteAttr
+		}
+	case util.WriteOnly:
+		if openMode.IsAppend() {
+			metricOpenMode = metrics.OpenModeWriteOnlyAppendAttr
+		} else {
+			metricOpenMode = metrics.OpenModeWriteOnlyAttr
+		}
+	}
```
kislaykishore: Create a function for this resolution. That'll help reduce the cyclomatic complexity of this change.

internal/fs/inode/file.go
```diff
@@ -1098,6 +1104,31 @@ func (f *FileInode) Truncate(
...
+	if metricOpenMode != "" {
```
kislaykishore: The metric must count every fallback instance. If metricOpenMode is not empty it means that the fallback happened but we don't have the open-mode. That should not obstruct us from logging the metric. We can create an "unknown" category for such cases.

internal/fs/metrics_test.go
```diff
@@ -1382,3 +1391,150 @@ func TestSetInodeAttributes_Metrics(t *testing.T) {
...
+	lookUpOp := &fuseops.LookUpInodeOp{
+		Parent: fuseops.RootInodeID,
+		Name:   fileName,
+	}
+	err := server.LookUpInode(ctx, lookUpOp)
+	require.NoError(t, err)
+	openOp := &fuseops.OpenFileOp{
+		Inode: lookUpOp.Entry.Child,
+	}
+	reflect.ValueOf(openOp).Elem().FieldByName("OpenFlags").Set(reflect.ValueOf(os.O_RDWR).Convert(reflect.ValueOf(openOp).Elem().FieldByName("OpenFlags").Type()))
```
kislaykishore: Can't really figure out what's happening here.

kislaykishore: The code looks quite complex. Would this work?
```suggestion
	openOp := &fuseops.OpenFileOp{
		Inode: lookUpOp.Entry.Child,
		OpenFlags: syscall. O_RDWR,
	}
```

internal/fs/metrics_test.go
```diff
@@ -1382,3 +1391,150 @@ func TestSetInodeAttributes_Metrics(t *testing.T) {
...
+		Inode:  lookUpOp.Entry.Child,
+		Handle: openOp.Handle,
+		Offset: 0,
+		Data:   []byte("test"),
+	}
+
+	err = server.WriteFile(ctx, op)
+	waitForMetricsProcessing()
+
+	assert.NoError(t, err)
```
```suggestion
	require.NoError(t, err)
```

kislaykishore: Move this up; right after server.WriteFile call.

internal/fs/metrics_test.go
```diff
@@ -1382,3 +1391,150 @@ func TestSetInodeAttributes_Metrics(t *testing.T) {
...
+		Inode:  lookUpOp.Entry.Child,
+		Handle: openOp.Handle,
+		Offset: 10,
+		Data:   []byte("data"),
+	}
+
+	err = server.WriteFile(ctx, op2)
+	waitForMetricsProcessing()
+
+	assert.NoError(t, err)
```
```suggestion
	require.NoError(t, err)
```

kislaykishore: Move it right after the WriteFile call.

internal/fs/metrics_test.go
```diff
@@ -1382,3 +1391,150 @@ func TestSetInodeAttributes_Metrics(t *testing.T) {
...
+	lookupOp2 := &fuseops.LookUpInodeOp{
+		Parent: fuseops.RootInodeID,
+		Name:   fileName2,
+	}
+	err = server.LookUpInode(ctx, lookupOp2)
+	require.NoError(t, err)
+	openOp2 := &fuseops.OpenFileOp{
+		Inode: lookupOp2.Entry.Child,
+	}
+	reflect.ValueOf(openOp2).Elem().FieldByName("OpenFlags").Set(reflect.ValueOf(os.O_RDWR).Convert(reflect.ValueOf(openOp2).Elem().FieldByName("OpenFlags").Type()))
```
kislaykishore: Same comment as above. Could we simplify this?

internal/fs/metrics_test.go
```diff
@@ -1382,3 +1391,150 @@ func TestSetInodeAttributes_Metrics(t *testing.T) {
...
+		Inode:  lookupOp2.Entry.Child,
+		Handle: openOp2.Handle,
+		Offset: 0,
+		Data:   []byte("data"),
+	}
+
+	err = server.WriteFile(ctx, op2)
+	waitForMetricsProcessing()
+
+	assert.NoError(t, err)
```
kislaykishore: Move it right after the WriteFile call
```suggestion
	require.NoError(t, err)
```

internal/fs/metrics_test.go
thrivikram-karur-g: yes, feasible with syscall attributes

internal/fs/metrics_test.go
thrivikram-karur-g: syscall attributes works.

internal/fs/inode/file.go
thrivikram-karur-g: done

internal/fs/inode/file.go
thrivikram-karur-g: done, infact moved it to metrics module

internal/fs/inode/file.go
thrivikram-karur-g: added other category support, updated metrics-gen code as well.

internal/fs/metrics_test.go
thrivikram-karur-g: was trying add the type of OpenFlags accordingly as per jacobsa fuse op expectation, hence all the reflect and type casting, seems like syscall works, so updated now.

internal/fs/metrics_test.go
thrivikram-karur-g: updated

thrivikram-karur-g: updated

internal/fs/metrics_test.go
thrivikram-karur-g: updated

thrivikram-karur-g: updated

internal/fs/metrics_test.go
thrivikram-karur-g: updated

internal/fs/metrics_test.go
thrivikram-karur-g: updated

metrics/metrics.yaml
kislaykishore:
```suggestion
    - "concurrency_limit_breached"
```

metrics/helper.go
```diff
@@ -14,9 +14,37 @@
...
+		if isAppend {
+			return OpenModeWriteOnlyAppendAttr
+		}
+		return OpenModeWriteOnlyAttr
+	default:
+		return OpenModeOtherAttr
+	}
+}
+
+func RecordStreamingWriteFallbackMetric(mh MetricHandle, openMode util.OpenMode, reason WriteFallbackReason) {
```
kislaykishore: It's an exported method, so it should have a test case. Either add a test for this method. Alternatively and preferably, move it to inode/ package and make it an internal (unexported method) method.

thrivikram-karur-g: I would prefer a metric based method to be with metric module for reuse somewhere else, I have added a test instead for the helper function.

metrics/metrics.yaml
thrivikram-karur-g: updated, this updated lots of format lines in the metrics auto-gen files.

internal/fs/metrics_test.go
kislaykishore, line 1: The presubmit failed. Please check.

internal/fs/inode/file.go
```diff
@@ -714,6 +718,7 @@ func (f *FileInode) writeUsingBufferedWrites(ctx context.Context, data []byte, o
 	// Fall back to temp file for Out-Of-Order Writes.
 	if errors.Is(err, bufferedwrites.ErrOutOfOrderWrite) {
 		logger.Infof("Out of order write detected. File %s will now use legacy staged writes. "+StreamingWritesSemantics, f.name.String())
+		metrics.RecordStreamingWriteFallbackMetric(f.metricHandle, openMode, metrics.WriteFallbackReasonOutOfOrderAttr)
```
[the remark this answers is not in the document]
thrivikram-karur-g: fine, fair enough I guess, do not see major difference but sounds more accurate record, so will incorporate.

internal/fs/inode/file.go
```diff
@@ -1058,6 +1063,7 @@ func (f *FileInode) truncateUsingBufferedWriteHandler(ctx context.Context, size
 	// If truncate size is less than the total file size resulting in OutOfOrder write, finalize and fall back to temp file.
 	if errors.Is(err, bufferedwrites.ErrOutOfOrderWrite) {
 		logger.Infof("Out of order write detected. File %s will now use legacy staged writes. "+StreamingWritesSemantics, f.name.String())
+		metrics.RecordStreamingWriteFallbackMetric(f.metricHandle, util.NewOpenMode(util.WriteOnly, 0), metrics.WriteFallbackReasonOutOfOrderAttr)
```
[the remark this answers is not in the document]
thrivikram-karur-g: same reply as above, makes sense, will incorporate

internal/fs/inode/file.go
```diff
@@ -1181,9 +1187,11 @@ func (f *FileInode) InitBufferedWriteHandlerIfEligible(ctx context.Context, open
 				"limit (set by --write-global-max-blocks) has been reached. To allow more concurrent files "+
 				"to use streaming writes, consider increasing this limit if sufficient memory is available. "+
 				"For more details on memory usage, see: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#writes", f.name.String())
+			metrics.RecordStreamingWriteFallbackMetric(f.metricHandle, openMode, metrics.WriteFallbackReasonConcurrencyLimitBreachedAttr)
 			return false, nil
 		}
 		if err != nil {
+			metrics.RecordStreamingWriteFallbackMetric(f.metricHandle, openMode, metrics.WriteFallbackReasonOtherAttr)
```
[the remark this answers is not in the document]
thrivikram-karur-g: no, this is accurate enough as buffered write handler creation failed and couldn't be initialized or used. I think it is good to go ahead with this.

internal/fs/metrics_test.go
thrivikram-karur-g: I think it just did not run, I am checking...

metrics/helper.go
```diff
@@ -14,9 +14,37 @@
...
+	metricOpenMode := getMetricOpenMode(openMode)
+	mh.FsStreamingWriteFallbackCount(1, metricOpenMode, reason)
+}
```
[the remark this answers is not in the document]
thrivikram-karur-g: updated the check

internal/fs/metrics_test.go
thrivikram-karur-g: It is passing now.

internal/fs/inode/file.go
```diff
@@ -1181,9 +1215,11 @@ func (f *FileInode) InitBufferedWriteHandlerIfEligible(ctx context.Context, open
 				"limit (set by --write-global-max-blocks) has been reached. To allow more concurrent files "+
 				"to use streaming writes, consider increasing this limit if sufficient memory is available. "+
 				"For more details on memory usage, see: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#writes", f.name.String())
+			recordStreamingWriteFallbackMetric(f.metricHandle, openMode, metrics.WriteFallbackReasonConcurrencyLimitBreachedAttr)
```
kislaykishore, line 1219: Ideally, we should return different errors depending on the reason for the fallback. That way, the caller can simply look at the error and record the fallback reason there. However, doing that at this point could be quite complicated. So, just wanted to make a note and keep this concern as non-blocking for this PR.

kislaykishore (dismissed):

internal/fs/inode/file.go
thrivikram-karur-g: makes sense, I was checking quickly where is the error caught and all and yes seems like a complicated change at this point. Closing this discussion.

internal/fs/inode/file.go
```diff
@@ -659,6 +664,33 @@ func (f *FileInode) Read(
...
+			return metrics.OpenModeReadWriteAppendAttr
+		}
+		return metrics.OpenModeReadWriteAttr
+	case util.WriteOnly:
+		if isAppend {
+			return metrics.OpenModeWriteOnlyAppendAttr
+		}
+		return metrics.OpenModeWriteOnlyAttr
+	default:
+		return metrics.OpenModeOtherAttr
```
vadlakondaswetha, line 683: when does this happen? I see openMode.AccessMode() is returning one of these 3
ReadWrite, readonly, writeonly

internal/fs/metrics_test.go
```diff
@@ -1503,3 +1511,147 @@ func TestReadFile_ReadBlockSizesMetric(t *testing.T) {
 		})
 	}
 }
+
+func TestStreamingWrite_Fallback_ExistingFile(t *testing.T) {
```
vadlakondaswetha: nit: streamingwrites

internal/fs/metrics_test.go
```diff
@@ -1503,3 +1511,147 @@ func TestReadFile_ReadBlockSizesMetric(t *testing.T) {
...
+	attrs := attribute.NewSet(attribute.String("open_mode", "write_only"), attribute.String("write_fallback_reason", "out_of_order"))
+	metrics.VerifyCounterMetric(t, ctx, reader, "fs/streaming_write_fallback_count", attrs, 1)
+}
+
+func TestStreamingWrite_Fallback_ConcurrencyLimitBreached(t *testing.T) {
+	ctx := context.Background()
+	params := defaultServerConfigParams()
+	params.enableStreamingWrites = true
+	params.writeGlobalMaxBlocks = 1
+	bucket, server, mh, reader := createTestFileSystemWithMetrics(ctx, t, params, false)
```
vadlakondaswetha, line 1608: please arrange all tests in AAA format

thrivikram-karur-g: It is already in AAA format 😅  , I think because of lengthy ARRANGE & ACT code it might be  confusing to reach through, let me see if adding comments separating sections can improve readability.

internal/fs/inode/file.go
thrivikram-karur-g: Though today there is only ReadWrite, ReadOnly & WriteOnly, purely from code perspective it returns an int value i.e accessMode . So default and fallback code for other have been added.

```
func (om OpenMode) AccessMode() int {
	return om.accessMode
}

type OpenMode struct {
	// accessMode defines the mutually exclusive access modes for opening a file.
	accessMode int

	// fileFlags defines flags that modify the open/read/write behavior.
	// These can be combined using bitwise OR.
	fileFlags int
}

func NewOpenMode(accessMode int, fileFlags int) OpenMode {
	return OpenMode{
		accessMode: accessMode,
		fileFlags:  fileFlags,
	}
}

```

internal/fs/metrics_test.go
thrivikram-karur-g: updated to make it more clear

internal/fs/metrics_test.go
thrivikram-karur-g: okay 👍

internal/fs/inode/file.go
vadlakondaswetha: Should we add a warning or something? Can we dynamically add values to the metrics. I am worried that we see the mode as other in metrics and dont understand what does it mean

internal/fs/metrics_test.go
```diff
@@ -1503,3 +1511,159 @@ func TestReadFile_ReadBlockSizesMetric(t *testing.T) {
...
+	op2 := &fuseops.WriteFileOp{
+		Inode:  lookUpOp.Entry.Child,
+		Handle: openOp.Handle,
+		Offset: 10,
+		Data:   []byte("data"),
+	}
+
+	// Act
+	err = server.WriteFile(ctx, op2)
+
```
vadlakondaswetha: comments should be sufficient. We dont need new lines within the AAA

internal/fs/inode/file.go
thrivikram-karur-g: we can't dynamically add values to the metric labels(if you mean code generated as this could theoretically mean infinite cardinality) but if you meant adding it in the future to the list, then it is possible

thrivikram-karur-g: here values refer to the metric labels values

vadlakondaswetha: can you add a warning in that case. Ideally it shouldn't happen since accessmode is initialized using this method which only return one of the 2 values, so adding a warning should be fine.

thrivikram-karur-g: yes, I do not see any concern in adding a warning.

internal/fs/metrics_test.go
thrivikram-karur-g: removed lines

internal/fs/inode/file.go
thrivikram-karur-g: added warning

thrivikram-karur-g:
debugging the test failure

internal/fs/inode/file.go
vadlakondaswetha: getMetricOpenMode: unexpected accessMode

vadlakondaswetha (approved):

internal/fs/inode/file.go
kislaykishore: A couple of questions:
1. Could this cause log spam if in future someone forgets to add a new access mode here in the future?
2. Is this warning log actionable. If it's just about adding a new case in the switch-case, could we just look at the metric-label: "others" and figure it out.
3. Note that this warning will be generated regardless of whether the metrics are enabled. So, there is no way to turn it off in production.

Because of these concerns, I would rather not have this warning.

PR #4633, opened 2026-04-20
feat: Add Read Path Support for Pirlo Buckets

vipnydav:
### Description
This change implements the read path support for Pirlo buckets by categorizing them as "rapid" buckets, consistent with the treatment of zonal buckets. This classification enables high-performance read features and specialized handling required for these bucket types.

b/503161174

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -386,6 +386,8 @@ params:
       bucket-type-optimization:
         - bucket-type: "zonal"
           value: "DefaultCongestionThreshold()"
+        - bucket-type: "pirlo"
+          value: "DefaultCongestionThreshold()"
 
   - config-path: "file-system.dir-mode"
     flag-name: "dir-mode"
@@ -409,6 +411,8 @@ params:
     optimizations:
       bucket-type-optimization:
         - bucket-type: "zonal"
+          value: true
+        - bucket-type: "pirlo"
           value: true
 
   - config-path: "file-system.experimental-enable-dentry-cache"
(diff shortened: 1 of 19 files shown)


internal/gcsx/client_readers/gcs_reader_test.go
```diff
@@ -476,39 +476,51 @@ func (t *gcsReaderTest) Test_ReadAt_ValidateZonalRandomReads() {
...
+			bucketType: gcs.BucketType{Zonal: true},
+		},
+		{
+			name:       "PirloBucket",
+			bucketType: gcs.BucketType{Pirlo: true},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func() {
+			t.SetupTest()
```
abhishek10004, line 495: Is this needed?

internal/fs/handle/file_test.go
```diff
@@ -1213,7 +1222,7 @@ func (t *fileTest) Test_ShouldSkipSizeChecks() {
 			openMode:       readOnlyMode,
 			offset:         50,
 			bufferSize:     60,
-			isRapidBucket:  true,
+			bucketType:     gcs.BucketType{Zonal: true},
```
abhishek10004: There are a lot of zonal=true scenarios. See if there's a clean way we can test them without too much duplication for pirlo as well.

abhishek10004 (commented):
Overall looks good. Left a few minor comments.

internal/gcsx/client_readers/gcs_reader_test.go
vipnydav: Yes, because `SetupTest` and `TearDownTest` are only called once for this test and not for each sub test.

cfg/params.yaml
abhishek10004, line 1: These changes will enable kernel reader on pirlo buckets as well by default. Please measure the perf with regional with these changes once as well to compare the performance with standard read path.

abhishek10004 (approved):

PR #4637, opened 2026-04-21
fix: prevent symlink attack in log file creation

vadlakondaswetha:
This PR fixes a reported symlink attack vulnerability in the `cmd/` package.

Specifically:
- In `cmd/mount.go`, the `--wire-log` file is now opened using `os.OpenFile` with the `syscall.O_NOFOLLOW` flag instead of `os.Create`.
- In `cmd/legacy_main.go`, the daemon `.stderr` log file is opened with the `syscall.O_NOFOLLOW` flag.

This prevents an unprivileged user from triggering arbitrary file overwrites on the host when gcsfuse runs as root, addressing the reported Container-to-Host escape vector.

---
*PR created automatically by Jules for task [451784451141055946](https://jules.google.com/task/451784451141055946) started by @vadlakondaswetha*

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -51,7 +51,6 @@ import (
 	"github.com/kardianos/osext"
 	"github.com/spf13/viper"
 	"golang.org/x/net/context"
-	"syscall"
 )
 
 const (
@@ -442,7 +441,7 @@ func Mount(mountInfo *mountInfo, bucketName, mountPoint string) (err error) {
 		var stderrFile *os.File
 		if newConfig.Logging.FilePath != "" {
 			stderrFileName := string(newConfig.Logging.FilePath) + ".stderr"
-			if stderrFile, err = os.OpenFile(stderrFileName, os.O_WRONLY|os.O_CREATE|os.O_APPEND|syscall.O_NOFOLLOW, 0644); err != nil {
+			if stderrFile, err = os.OpenFile(stderrFileName, os.O_WRONLY|os.O_CREATE|os.O_APPEND|unix.O_NOFOLLOW, 0644); err != nil {
 				return err
 			}
 		}
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -51,6 +51,7 @@ import (
 	"github.com/kardianos/osext"
 	"github.com/spf13/viper"
 	"golang.org/x/net/context"
+	"syscall"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


cmd/legacy_main.go
```diff
@@ -441,7 +441,7 @@ func Mount(mountInfo *mountInfo, bucketName, mountPoint string) (err error) {
 		var stderrFile *os.File
 		if newConfig.Logging.FilePath != "" {
 			stderrFileName := string(newConfig.Logging.FilePath) + ".stderr"
-			if stderrFile, err = os.OpenFile(stderrFileName, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644); err != nil {
+			if stderrFile, err = os.OpenFile(stderrFileName, os.O_WRONLY|os.O_CREATE|os.O_APPEND|unix.O_NOFOLLOW, 0644); err != nil {
 				return err
```
meet2mky, line 445: With this change, GCSFuse will not be able use any log file path which is symbolic link as it would fail. Do you think it's good idea to return a better error message here something like if it failed because of file being symlink we can report that symlink based log files are not supported? Similarly in other places.

meet2mky (commented):
Looks good. Just one comment about observability.

cmd/legacy_main.go
vadlakondaswetha: added unit tests.
The error returned by the openFile call itself has symlink related error information. so no need of customer error message.

meet2mky (dismissed):
Looks good.

cmd/mount_test.go
```diff
@@ -148,3 +150,29 @@ func TestGetFuseMountConfig_EnableReaddirplus(t *testing.T) {
...
+
+func TestGetFuseMountConfig_WireLogNoFollowSymlink(t *testing.T) {
+	tempDir := t.TempDir()
+	logFile := tempDir + "/wire.log"
+	symlinkPath := tempDir + "/symlink.log"
+
+	// Create a dummy log file
+	f, err := os.Create(logFile)
+	assert.NoError(t, err)
+	assert.NoError(t, f.Close())
```
kislaykishore:
```suggestion
	require.NoError(t, err)
	require.NoError(t, f.Close())
```

cmd/mount_test.go
```diff
@@ -148,3 +150,29 @@ func TestGetFuseMountConfig_EnableReaddirplus(t *testing.T) {
...
+
+	// Create a symlink pointing to the log file
+	err = os.Symlink(logFile, symlinkPath)
+	assert.NoError(t, err)
+
+	newConfig := &cfg.Config{
```
kislaykishore, line 166: Let's remove blank lines to keep the "arrange" block together.

internal/logger/logger_test.go
```diff
@@ -566,3 +568,26 @@ func TestGetLogFHandler(t *testing.T) {
 		assert.Regexp(t, expectedTraceRegex, logs[1])
 	})
 }
+
+func TestInitLogFile_NoFollowSymlink(t *testing.T) {
+	tempDir := t.TempDir()
+	logFile := tempDir + "/main.log"
+	symlinkPath := tempDir + "/symlink.log"
+
```
kislaykishore: Same comments as before - Let's keep the arrange block together.

internal/logger/logger_test.go
```diff
@@ -566,3 +568,26 @@ func TestGetLogFHandler(t *testing.T) {
...
+	// Create a dummy log file
+	f, err := os.Create(logFile)
+	assert.NoError(t, err)
+	assert.NoError(t, f.Close())
```
kislaykishore:
```suggestion
	require.NoError(t, err)
	require.NoError(t, f.Close())
```

abhishek10004 (dismissed):
LGTM. Please fix the minor comments left by Kislay.

kislaykishore (approved):

PR #4639, opened 2026-04-21
feat(metrics): Streaming write fallback metric-gen code

thrivikram-karur-g:
### Description
Adding metric for tracking streaming write fallback with reasons metric-gen code.

b/504892393

### Testing details
1. Manual - NA (given it is auto generated code, existing tests passing and new tests added getting passed is good enough)
2. Unit tests - otel_metrics_test.go updated.
3. Integration tests - NA

N/A


thrivikram-karur-g:
gemini review wouldn't be relevant but still adding it as it is auto-gen code.

metrics/metrics.yaml
```diff
@@ -140,6 +140,26 @@
     attribute-type: string
     values: *fs_ops_list
 
+- metric-name: "fs/streaming_write_fallback_count"
+  description: "The cumulative distribution of the number of streaming write fallbacks with reason attached"
```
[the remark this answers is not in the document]
thrivikram-karur-g: okay, making the change though this term is used for other such metrics.

metrics/otel_metrics_test.go
```diff
@@ -4570,6 +4570,237 @@ func TestFsOpsLatency(t *testing.T) {
...
+		},
+		{
+			name: "open_mode_write_only_append_write_fallback_reason_out_of_order",
+			f: func(m *otelMetrics) {
+				m.FsStreamingWriteFallbackCount(5, "write_only_append", "out_of_order")
+			},
+			expected: map[attribute.Set]int64{
+				attribute.NewSet(attribute.String("open_mode", "write_only_append"), attribute.String("write_fallback_reason", "out_of_order")): 5,
+			},
+		}, {
```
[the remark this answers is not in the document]
thrivikram-karur-g: I do not see any redudancy here and updating test case template is out of scope for this PR.

metrics/metrics.yaml
```diff
@@ -140,6 +140,26 @@
...
+    - "write_only_append"
+    - "read_write"
+    - "read_write_append"
+  - attribute-name: write_fallback_reason
+    attribute-type: string
+    values:
+    - "out_of_order"
+    - "existing_file"
+    - "concurrent_limit_breached"
+    - "not_enabled"
```
kislaykishore: I'd recommend we get rid of the `not_enabled`. It's a configuration and we should not have a metric for it. It doesn't justify the name: "fallback"

thrivikram-karur-g: cc: @vadlakondaswetha

thrivikram-karur-g: Confirmed with swetha that as well and agreed that it is better to remove. Removing this.

kislaykishore (dismissed):

kislaykishore (approved):

PR #4640, opened 2026-04-22
feat(metrics): Read block size metric gen code and tpl changes

thrivikram-karur-g:
### Description
Adding the metric gen code for read/block_sizes metric.
This is a histogram metric and the details for the metric are available in metrics.yaml .

b/504892196

### Testing details
1. Manual - NA
2. Unit tests - Auto generated unittests have been added and need to pass.
3. Integration tests - NA

N/A

### DIFF
--- a/tools/metrics-gen/main.go
+++ b/tools/metrics-gen/main.go
@@ -70,20 +70,18 @@ type TemplateData struct {
 
 // Helper functions for the template.
 var funcMap = template.FuncMap{
-	"toPascal":         toPascal,
-	"toCamel":          toCamel,
-	"getVarName":       getVarName,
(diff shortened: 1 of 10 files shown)


tools/metrics-gen/main.go
```diff
@@ -159,6 +160,10 @@ func getUnitMethod(unit string) string {
 	}
 }
 
+func isTimeHistogram(m Metric) bool {
+	return m.Type == "int_histogram" && (m.Unit == "us" || m.Unit == "ms" || m.Unit == "s")
```
kislaykishore: Let's change the type to `time_histogram`

kislaykishore (approved):

PR #4641, opened 2026-04-22
Fix lock leak in renameHierarchicalDir when destination directory is not empty

vadlakondaswetha:
Fixes a resource leak and possible deadlock in `renameHierarchicalDir`. When renaming an object to a new directory that is not empty, `checkDirNotEmpty` returns an error, but the `newDirInode` lock wasn't added to `pendingInodes` and so it was never released. By moving the append to `pendingInodes` up, right after `err == nil`, we ensure it always gets unlocked and the lookup count is decremented.

---
*PR created automatically by Jules for task [4274856973442780388](https://jules.google.com/task/4274856973442780388) started by @vadlakondaswetha*

### DIFF
--- a/tools/integration_tests/operations/rename_dir_test.go
+++ b/tools/integration_tests/operations/rename_dir_test.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/tools/integration_tests/operations/rename_dir_test.go
+++ b/tools/integration_tests/operations/rename_dir_test.go
@@ -1,0 +1,68 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


tools/integration_tests/operations/rename_dir_test.go
```diff
@@ -0,0 +1,68 @@
...
+	operations.CreateFileWithContent(destFilePath, setup.FilePermission_0600, Content, t)
+
+	// Attempt to rename the source directory to the destination directory.
+	err = os.Rename(srcDirPath, destDirPath)
+
+	// Assert that an error occurred (because destination is not empty).
+	assert.Error(t, err)
+	assert.True(t, strings.Contains(err.Error(), "file exists") || strings.Contains(err.Error(), "directory not empty") || strings.Contains(err.Error(), "not empty"), "Error message should mention file exists or not empty, but got: %v", err)
+
+	// If the issue was not fixed, this process could hang on further operations due to leaked locks.
```
meet2mky: > 	// If the issue was not fixed, this process could hang on further operations due to leaked locks.

Nit: This comment wouldn't make sense unless read in the context of this PR. We can remove it and also arrange test in AAA format.

meet2mky (dismissed):
Looks good only a minor nit.

meet2mky (approved):

abhishek10004 (approved):

PR #4642, opened 2026-04-22
feat(native retries): Add hidden enable-native-retries flag for GKE native mount retries

meet2mky:
### Description

This PR introduces a new hidden boolean flag, `enable-mount-retries` (default: `false`), to support native GCSFuse mount retries under GKE environments, as described in the GCSFuse X GKE Native Retries design (go/gcsfuse-gke-native-retries).
### Key Changes
1. **Configuration Schema (`cfg/params.yaml`)**:
   - Added `gcs-retries.enable-mount-retries` as a hidden boolean flag under `gcs-retries` section.
   - Defaulted to `false`.
   - Positioned alphabetically to keep parameters sorted in ascending order.
2. **Test Suite Coverage (`cmd/config_validation_test.go` & `cmd/root_test.go`)**:
   - Updated default GCSFuse configuration validation expectations to include `EnableMountRetries: false`.
   - Added a new unit test `"Test with enable-mount-retries explicitly true"` to verify CLI flag parsing works as expected when overridden.

b/504517897

### Testing details
1. Manual - NA
2. Unit tests - Added
3. Integration tests - Part of Presubmit

NONE

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -636,6 +636,8 @@ type GcsRetriesConfig struct {
 
 	ChunkTransferTimeoutSecs int64 `yaml:"chunk-transfer-timeout-secs"`
 
+	EnableMountRetries bool `yaml:"enable-mount-retries"`
+
 	ExperimentalNonrapidFolderApiStallRetry bool `yaml:"experimental-nonrapid-folder-api-stall-retry"`
 
 	MaxRetryAttempts int64 `yaml:"max-retry-attempts"`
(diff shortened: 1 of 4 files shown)


cfg/config.go
```diff
@@ -636,6 +636,8 @@ type GcsRetriesConfig struct {
 
 	ChunkTransferTimeoutSecs int64 `yaml:"chunk-transfer-timeout-secs"`
 
+	EnableNativeRetries bool `yaml:"enable-native-retries"`
```
vadlakondaswetha: the project is called native retries from the GKE perspective that GCSFuse ishandling the retries. how will you explain this flag to other customers> Should we call it enable mount retries or something else?

meet2mky: It's not meant for other customers apart from GKE and it's hidden flag but if it make sense then let's add the mount in the flag and make it `--enable-mount-native-retries`?
What do you think?

meet2mky: Or maybe `--enable-mount-retries` and drop the native as well?

meet2mky: Thanks, renaming it to --enable-mount-retries as per offline discussion.

vadlakondaswetha (approved):

PR #4651, opened 2026-04-27
fix(tests): dynamically resolve log file path based on active test suite

PranjalC100:
### Description
This PR addresses a critical issue in the integration test framework where log file paths were incorrectly resolved for packages containing multiple test suites in GKE mounted directory test cases.

**Problem:**
The `SetUpLogFilePath` function was previously hardcoded to always infer the log file name from the first configuration block (`cfg.Configs[0]`) whenever a `--log-file` flag was not explicitly provided. 

In GKE environments, we deploy test pods with specific mount flags where the log file is passed via the `--log-file` flag in the `test_config.yaml`. While the hardcoded logic worked for simple packages, it caused failures in multi-suite packages like `buffered_read`, `inactive_stream_timeout`, and `readdirplus`. For example, when running `TestTimeoutDisabledSuite` in isolation, the framework would incorrectly look for `TestTimeoutEnabledSuite.log`. 

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/test/e2e/testsuites/gcsfuse_integration.go#L327

**Local Testing Observation:**
This bug was initially difficult to reproduce locally because the `/gcsfuse-tmp` directory was not being cleared between test suite runs. As a result, the framework was accessing stale log files from previous suites, allowing tests to pass incorrectly despite pointing to the wrong file path. 

Once the testing process was updated to clear the `/gcsfuse-tmp` directory after every test suite run within a package, the issue became visible as the framework could no longer fallback to these stale files. This mirrors the behavior in clean GKE environments where the expected file is never created, leading to immediate "file not found" errors.

**Solution:**
The resolution logic has been refactored to dynamically identify the currently running test suite using Go's standard `flag.Lookup("test.run")`. The framework now matches this active test name against the `Run` identifier in the configuration blocks to retrieve the correct log file path. A fallback to the first configuration block is maintained for cases where a match is not found.

### Linked Issues
- Fixes [b/506212867](http://b/506212867)
- Fixes [b/505825331](http://b/505825331)
- Fixes [b/505831192](http://b/505831192)

## Testing Details
1. **Manual:** Verified that log files are correctly created and named during local runs of individual suites after clearing the `/gcsfuse-tmp` directory.
2. **GKE Environment:** Verified running tests in the GKE environment on this commit, confirming that log paths are now correctly resolved dynamically.
3. **Issue Reproduction:** Successfully reproduced the main issue on the `3.9.0_release` branch by clearing the temporary directories, validating that the previous hardcoded logic was causing "file not found" errors (e.g., `open /gcsfuse-tmp.log: no such file or directory`).
4. **Integration tests:** Kokoro presubmits.

### Any backward incompatible change?
None. This change only improves the accuracy of log file path resolution within the integration test framework.

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -895,13 +895,23 @@ func ParseLogFileFromFlags(flags []string) string {
 	return ""
 }
 
-func SetUpLogFilePath(flags []string, GKETempDir string, OldGKElogFilePath string, cfg *test_suite.TestConfig) {
+func SetUpLogFilePath(testName string, flags []string, GKETempDir string, OldGKElogFilePath string, cfg *test_suite.TestConfig) {
 	var logFilePath string
 	parsedLogFileName := ParseLogFileFromFlags(flags)
 
 	// Infer log filename directly from the parsed config block.
-	if parsedLogFileName == "" && cfg != nil && len(cfg.Configs) > 0 {
-		parsedLogFileName = ParseLogFileFromFlags(cfg.Configs[0].Flags)
+	if parsedLogFileName == "" && cfg != nil {
+		currentTest := strings.Trim(testName, "^$")
+		currentTest = strings.Split(currentTest, "/")[0]
+
+		if currentTest != "" {
+			for _, c := range cfg.Configs {
+				if c.Run == currentTest {
+					parsedLogFileName = ParseLogFileFromFlags(c.Flags)
+					break
+				}
+			}
+		}
 	}
 
 	// Default logFile name.
--- a/tools/integration_tests/buffered_read/fallback_test.go
+++ b/tools/integration_tests/buffered_read/fallback_test.go
@@ -41,7 +41,7 @@ type fallbackSuiteBase struct {
 }
 
 func (s *fallbackSuiteBase) SetupSuite() {
-	setup.SetUpLogFilePath(s.flags, GKETempDir, "", testEnv.cfg)
+	setup.SetUpLogFilePath(s.T().Name(), s.flags, GKETempDir, "", testEnv.cfg)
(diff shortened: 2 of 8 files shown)


meet2mky:
Hi @PranjalC100, I see multiple time back & forth happening on the migration. Could you run the all these tests reported in bugs on this GCSFuse branch and verify if they are passing also attach the output here in description.

meet2mky (commented):
Please resolve this once you have completed the testing. Thanks!

PranjalC100:
Hey @meet2mky, I've tested this commit in a GKE environment following the steps provided in the bugs logged. The tests for `inactive_stream_timeout`, `readdirplus` & `buffered_read`  have passed. I've added the detailed logs to this gPaste: [link](https://paste.googleplex.com/6080124731654144). Thanks!

meet2mky (commented):
Kokoro is failing. PTAL

tools/integration_tests/util/setup/setup.go
```diff
@@ -895,13 +895,23 @@ func ParseLogFileFromFlags(flags []string) string {
 	return ""
 }
 
-func SetUpLogFilePath(flags []string, GKETempDir string, OldGKElogFilePath string, cfg *test_suite.TestConfig) {
+func SetUpLogFilePath(testName string, flags []string, GKETempDir string, OldGKElogFilePath string, cfg *test_suite.TestConfig) {
```
meet2mky, line 898: Let's refactor this method in more simpler terms as to what logic is executing in case of GKE vs non-GKE and the fallback behaviour in both scenarios. Not blocking on this.

meet2mky (approved):

PranjalC100:
TODO: https://buganizer.corp.google.com/issues/507323874

PR #4656, opened 2026-04-28
fix(bufferedwrite): correct out-of-order write detection for stale truncated offsets

vipnydav:
### Description
This PR fixes a bug in the streaming writes path where an out of order write could incorrectly bypass the `ErrOutOfOrderWrite` safeguard. 

Previously, if an application truncated a file upwards, wrote past that truncation point, and then performed a backward seek exactly to the `truncatedSize`, the out-of-order check would evaluate to false. This bypassed the safe fallback to legacy staged writes and caused the new data to be incorrectly appended to the end of the file.

The fix ensures that `truncatedSize` is only treated as a valid offset if the file's `totalSize` has not yet surpassed it. Once `wh.totalSize >= wh.truncatedSize`, any subsequent write at the `truncatedSize` offset is correctly identified as a backward seek, safely returning `ErrOutOfOrderWrite` and triggering the local staged writes fallback.

b/498246786

### Testing details
1. Manual - Done
2. Unit tests - Added

No

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler.go
+++ b/internal/bufferedwrites/buffered_write_handler.go
@@ -141,7 +141,7 @@ func (wh *bufferedWriteHandlerImpl) Write(ctx context.Context, data []byte, offs
 	if err != nil {
 		return
 	}
-	if offset != wh.totalSize && offset != wh.truncatedSize {
+	if offset != wh.totalSize && (offset != wh.truncatedSize || wh.totalSize >= wh.truncatedSize) {
 		logger.Errorf("BufferedWriteHandler.OutOfOrderError for object: %s, expectedOffset: %d, actualOffset: %d",
 			wh.uploadHandler.objectName, wh.totalSize, offset)
 		return ErrOutOfOrderWrite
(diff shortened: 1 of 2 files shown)


internal/bufferedwrites/buffered_write_handler.go
```diff
@@ -141,7 +141,7 @@ func (wh *bufferedWriteHandlerImpl) Write(ctx context.Context, data []byte, offs
 	if err != nil {
 		return
 	}
-	if offset != wh.totalSize && offset != wh.truncatedSize {
+	if offset != wh.totalSize && (offset != wh.truncatedSize || wh.totalSize >= wh.truncatedSize) {
```
vadlakondaswetha, line 148: can you add a comment here once we write post the truncated size, any writes starting from the truncated offset are considered out of order and give an example.

vadlakondaswetha (dismissed):

internal/bufferedwrites/buffered_write_handler.go
vipnydav: Done

vadlakondaswetha (approved):

PR #4660, opened 2026-04-29
test(managed_folders): replace static sleep with dynamic IAM polling

PranjalC100:
### Description
Replaces the hardcoded 60-second sleep with a **dynamic polling mechanism** `(waitForIAMPropagation)` for **IAM policy propagation** in the `managed folders` integration tests. 
This allows tests to proceed as soon as the expected IAM role propagates rather than waiting the full 60 seconds every time, improving overall test execution speed.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - These changes directly update and optimize the managed folders integration tests `admin_permissions_test.go` and `view_permissions_test.go`

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -679,7 +679,6 @@ func getMetricOpenMode(openMode util.OpenMode) metrics.OpenMode {
 		}
 		return metrics.OpenModeWriteOnlyAttr
 	default:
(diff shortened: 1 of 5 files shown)


PranjalC100:
The used API is served from the Control Plane and might lead to false positives, as the tests might fail on the assumption that the permission has been propagated correctly in Data Plane. Closing this refactoring for now.

PR #4664, opened 2026-04-30
fix: validate object size after upload to detect incomplete writes

vipnydav:
### Description
This change introduces a validation check immediately following file upload operations to ensure data integrity.

It has been observed that the backend can occasionally return a successful "OK" status for a write stream closure, even if the data persistence failed. In these cases, the resulting object is recorded with a size of `0` despite the client having transmitted non-empty data.

To prevent these failures from being silently ignored, GCSFuse now explicitly validates the object size returned by the server against the expected size during Flush and Sync operations.

b/508075834

### Testing details
1. Manual - Done
2. Unit tests - Added

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -976,6 +976,13 @@ func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
 	if err != nil {
 		return fmt.Errorf("SyncObject: %w", err)
 	}
+
+	if newObj != nil && f.content != nil {
+		if st, statErr := f.content.Stat(); statErr == nil && newObj.Size != uint64(st.Size) {
+			return fmt.Errorf("could not upload entire data, expected size %d, Got %d", st.Size, newObj.Size)
+		}
+	}
+
 	minObj := storageutil.ConvertObjToMinObject(newObj)
 	// If we wrote out a new object, we need to update our state.
 	f.updateInodeStateAfterFlush(minObj)
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -978,8 +978,13 @@ func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
 	}
 
 	if newObj != nil && f.content != nil {
-		if st, statErr := f.content.Stat(); statErr == nil && newObj.Size != uint64(st.Size) {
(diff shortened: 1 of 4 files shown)


anushka567:
@vipnydav Can you share a reproducer/references if any for the following
`It has been observed that the backend can occasionally return a successful "OK" status for a write stream closure, even if the data persistence failed. `

vipnydav:
@anushka567 
> @vipnydav Can you share a reproducer/references if any for the following `It has been observed that the backend can occasionally return a successful "OK" status for a write stream closure, even if the data persistence failed. `

While the behavior is not currently reproducible with a consistent set of steps, it is the issue being referenced in [b/505585459](https://b.corp.google.com/issues/505585459). That bug tracks an instance where an application encountered a JSONDecodeError because a metadata file was read as empty or stale immediately after it was successfully written. This PR implements the suggested improvement to explicitly validate the object size returned by the server after the upload call which would help us debug these issues faster.

internal/fs/inode/file.go
```diff
@@ -976,6 +976,18 @@ func (f *FileInode) syncUsingContent(ctx context.Context) (err error) {
 	if err != nil {
 		return fmt.Errorf("SyncObject: %w", err)
 	}
+
+	if newObj != nil && f.content != nil {
+		st, statErr := f.content.Stat()
```
vadlakondaswetha, line 981: how long is the f.content persisted after sync?

vadlakondaswetha (approved):

internal/fs/inode/file.go
vipnydav: After the `sync` operation completes, the `updateInodeStateAfterSync` method is called. Inside this method, the code checks if the sync was successful `(minObj != nil)` and whether the local file cache is disabled `(!f.localFileCache)`.

If both of these conditions are met, it immediately calls `f.content.Destroy()` and sets it to `nil`. This means that unless the local file caching feature is enabled (which in my understanding is always disabled), `f.content` is destroyed and its local disk space is freed up in [Line 993](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4664/changes#diff-ef57b9a0715508504d96904e54ecc0d11df9624040c29a850dce217d9eda9497R993).

PR #4674, opened 2026-05-07
fix: Making dp check more reliable

vadlakondaswetha:
### Description
Make the stat call during dp check more reliable by using executeWithRetry method to handle stall and failure scenarios.

### Testing details
1. Manual - Done
2. Unit tests - Added as applicable
3. Integration tests - Already present

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -237,7 +237,7 @@ func verifyDirectPathConnectivity(ctx context.Context, clientConfig *storageutil
 		RetryMultiplier:  1.5,
 		MaxRetryAttempts: directPathDetectionMaxAttempts,
 	}
-	retryConfig := storageutil.NewRetryConfig(dpClientConfig, directPathDetectionTimeout, directPathDetectionMaxRetryDuration, 1*time.Second)
+	retryConfig := storageutil.NewRetryConfig(dpClientConfig, directPathDetectionTimeout, directPathDetectionMaxRetryDuration, storageutil.DefaultInitialBackoff)
 
 	apiCall := func(attemptCtx context.Context) (*storage.ObjectAttrs, error) {
 		return bucketHandle.Object(testObject).Attrs(attemptCtx)
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -234,7 +234,7 @@ func verifyDirectPathConnectivity(ctx context.Context, clientConfig *storageutil
 
 	dpClientConfig := &storageutil.StorageClientConfig{
 		MaxRetrySleep:    directPathDetectionMaxBackoff,
-		RetryMultiplier:  1.5,
+		RetryMultiplier:  clientConfig.RetryMultiplier,
 		MaxRetryAttempts: directPathDetectionMaxAttempts,
 	}
 	retryConfig := storageutil.NewRetryConfig(dpClientConfig, directPathDetectionTimeout, directPathDetectionMaxRetryDuration, storageutil.DefaultInitialBackoff)
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -202,6 +202,12 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		return nil, fmt.Errorf("NewGRPCClient: %w", err)
 	}
 
+	// Set the production level retry config.
+	defer func() {
+		logger.Infof("Applying production retry config after DirectPath verification.")
+		setRetryConfig(ctx, sc, clientConfig)
+	}()
+
 	// Direct-path verification is fatal for regional. Todo(b/503624405): Make it fatal for all after making the dummy-stat reliable.
 	if verifyErr := verifyDirectPathConnectivity(ctx, clientConfig, bucketName, sc, billingProject); verifyErr != nil {
 		logger.Warnf("DirectPath verification failed with error: %v", verifyErr)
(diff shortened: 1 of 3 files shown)


internal/storage/storage_handle.go
```diff
@@ -239,26 +218,36 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
 		bucketHandle = bucketHandle.UserProject(billingProject)
 	}
-	_, statErr := bucketHandle.Object(testObject).Attrs(verifyCtx)
+
+	dpClientConfig := &storageutil.StorageClientConfig{
+		MaxRetrySleep:    directPathDetectionMaxBackoff,
+		RetryMultiplier:  1.5,
+		MaxRetryAttempts: directPathDetectionMaxAttempts,
+	}
+	retryConfig := storageutil.NewRetryConfig(dpClientConfig, directPathDetectionTimeout, directPathDetectionMaxRetryDuration, 1*time.Second)
```
meet2mky: Nit: Use existing default (1s) initial backoff constant.
```suggestion
	retryConfig := storageutil.NewRetryConfig(dpClientConfig, directPathDetectionTimeout, directPathDetectionMaxRetryDuration, storageutil.DefaultInitialBackoff)
```

internal/storage/storage_handle.go
```diff
@@ -239,26 +218,36 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
...
 	var testObject = "gcsfuse-dp-object"
 	bucketHandle := sc.Bucket(bucketName)
 	if billingProject != "" {
 		bucketHandle = bucketHandle.UserProject(billingProject)
 	}
-	_, statErr := bucketHandle.Object(testObject).Attrs(verifyCtx)
+
+	dpClientConfig := &storageutil.StorageClientConfig{
+		MaxRetrySleep:    directPathDetectionMaxBackoff,
+		RetryMultiplier:  1.5,
```
meet2mky: What's the rationale behind using RetryMultiplier to 1.5 instead of 2.0 default on other places. I don't think it's going to make much difference. 

| Attempt | Multiplier = 1.5 Sleep Range | Multiplier = 2.0 Sleep Range |
| :---: | :--- | :--- |
| **1** | `[1 ns, 1.00s]` | `[1 ns, 1.00s]` |
| **2** | `[1 ns, 1.50s]` | `[1 ns, 2.00s]` |
| **3** | `[1 ns, 2.25s]` | `[1 ns, 4.00s]` |
| **4** | `[1 ns, 3.375s]` | `[1 ns, 5.00s]` *(Capped)* |
| **5** | `[1 ns, 5.00s]` *(Capped)* | `[1 ns, 5.00s]` *(Capped)* |
| **6** | `[1 ns, 5.00s]` *(Capped)* | `[1 ns, 5.00s]` *(Capped)* |
| **7** | `[1 ns, 5.00s]` *(Capped)* | `[1 ns, 5.00s]` *(Capped)* |

So before each attempt the the sleep duration would be anything random value between these range so we can just go ahead with the default 2.0.

Note- Our code is incorrect for max backoff and we have PR open: #4672

meet2mky (dismissed):
Looks good to me just few comments.

meet2mky:
Do you want to build sidecar image from your branch and verify the code change in inference at 512 nodes and see how stall and retries are working together? @vadlakondaswetha

internal/storage/storage_handle.go
vadlakondaswetha: Prince defined these values as he wanted an aggressive retry. Changing it to 2.

vadlakondaswetha:
> Do you want to build sidecar image from your branch and verify the code change in inference at 512 nodes and see how stall and retries are working together? @vadlakondaswetha

I will raise a ticket to vendor after this PR is merged.

meet2mky (dismissed):
LGTM

internal/storage/storage_handle.go
```diff
@@ -221,7 +201,6 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 	if sc, err = storage.NewGRPCClient(ctx, clientOpts...); err != nil {
 		return nil, fmt.Errorf("NewGRPCClient: %w", err)
 	}
-	setRetryConfig(ctx, sc, clientConfig)
```
abhishek10004, line 224: IMO the production retry config should be set in createGRPCClientHandle only. It is not verifyDirectPathConnectivity's responsibility. Earlier it made sense as verifyDirectPathConnectivity was just restoring the original config (i.e. no side effects) but now it is setting it.

IMO, we should move the defer block in verifyDirectPathConnectivity to here (with an updated msg.

abhishek10004 (dismissed):
LGTM

internal/storage/storage_handle.go
```diff
@@ -183,22 +183,23 @@ func setRetryConfig(ctx context.Context, sc *storage.Client, clientConfig *stora
...
 
 // Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
-func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig, isbucketRapid bool, enableBidiConfig bool, bucketName string, billingProject string) (sc *storage.Client, err error) {
+func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig, isbucketRapid bool, enableBidiConfig bool, bucketName string, billingProject string) (*storage.Client, error) {
 	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
 		return nil, fmt.Errorf("error setting direct path env var: %w", err)
 	}
 	defer unSetDirectPathEnvVariable()
 
 	var clientOpts []option.ClientOption
```
abhishek10004, line 192: nit: should not be needed anymore

abhishek10004 (approved):

internal/storage/storage_handle.go
vadlakondaswetha: done.

PR #4675, opened 2026-05-07
feat(kernel range reader): Adding support for fuse-max-pages-limit param

abhishek10004:
### Description
Adding support for fuse-max-pages-limit param. While this will limit the max number of pages that fuse will process in a single request, the machine level value will only be updated if that is lower than the currently set value.

b/510726102

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


cfg/config.go
```diff
@@ -1143,14 +1118,20 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
...
+	flagSet.BoolP("finalize-file-for-rapid", "", false, "Finalizes the files on close for Rapid storage. Appends will be slower on finalized files.")
 
-	if err := flagSet.MarkHidden("finalize-file-on-close"); err != nil {
+	if err := flagSet.MarkHidden("finalize-file-for-rapid"); err != nil {
 		return err
 	}
 
 	flagSet.BoolP("foreground", "", false, "Stay in the foreground after mounting.")
 
+	flagSet.IntP("fuse-max-pages-limit", "", DefaultFuseMaxPagesLimit(), "Sets the limit for the maximum number of pages that fuse can process in a single request. This will update the node level value (on supported kernels) if the new value is greater than the current node level value.")
```
vadlakondaswetha: is node a common terminology outside GKE. should we rephrase it differently - i.e,This gets applied at machine level and not specific to a single mount. The value will be updated only if the provided value is greater than the current value.

meet2mky: Nit: Can we make it more concise here?
```suggestion
	flagSet.IntP("fuse-max-pages-limit", "", DefaultFuseMaxPagesLimit(), "Sets the limit for the maximum number of pages that fuse can process in a single request. On supported kernels, This limit would be increased upto the fuse-max-pages-limit if the existing limit is lower.")
```

internal/kernelparams/kernelparams.go
```diff
@@ -191,9 +210,27 @@ func (m *KernelParamsManager) addParam(name ParamName, value string) {
 	})
 }
 
-// SetMaxPagesLimit adds the max_pages_limit parameter to the config.
+// SetMaxPagesLimit adds the max_pages_limit parameter to the config only if the
+// requested limit is greater than the current host max_pages_limit value.
 func (m *KernelParamsManager) SetMaxPagesLimit(limit int) {
```
meet2mky: I think this function should just do m.addParam and the business logic to know if it's supported or not could be handled by the caller method as it would separate the utility from the functional logic.

meet2mky (commented):
Added few comments, PTAL

cfg/config.go
abhishek10004: Rephrased it. It's slightly verbose but more clear now.

internal/kernelparams/kernelparams.go
abhishek10004: Done. Added a helper for checking if we should update the limit or not.

internal/kernelparams/kernelparams.go
```diff
@@ -31,6 +31,25 @@ import (
...
+	}
+	return val, nil
+}
+
+// SupportsFuseMaxPagesLimit checks if the host kernel supports configuring
+// the FUSE max_pages_limit attribute by reading the procfs entry.
+func SupportsFuseMaxPagesLimit() bool {
+	_, err := readMaxPagesLimitFunc()
+	return err == nil
+}
```
meet2mky: Nit: This is not required as everytime we would need the current max pages value and error together to determine if it's supported and what should be next value based on current value.
```suggestion
```

meet2mky (dismissed):
LGTM

internal/kernelparams/kernelparams.go
abhishek10004: Yeah, we'd need this in the next change while determining whether to enable the kernel reader flag. I'll move it to that PR. Thanks.

meet2mky (approved):

vadlakondaswetha (approved):
LGTM based on Mohit's review

PR #4676, opened 2026-05-09
feat(kernel range reader): Adding implementation for kernel range reader

abhishek10004:
### Description
Adding implementation for kernel range reader

b/513189024

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -281,7 +281,7 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		} else {
 			logger.Warnf("Cannot apply bucket-type optimizations as ViperConfig is nil")
 		}
-		// Write post mount kernel settings for Zonal Buckets when kernel reader is enabled in GKE environments for
+		// Write post mount kernel settings when kernel reader is enabled in GKE environments for
 		// non dynamic mounts before user space mounting in GCSFuse. Mounting in GKE is already done at this point but
 		// writing kernel settings early ensures the asynchronous application of these settings happens as early as possible in GKE.
 		if serverCfg.NewConfig.FileSystem.KernelParamsFile != "" && bucketType.Zonal && serverCfg.NewConfig.FileSystem.EnableKernelReader {
@@ -3125,7 +3125,7 @@ func (fs *fileSystem) ReadFile(
 			Buffer: op.Dst,
 			Offset: op.Offset,
 		}
-		resp, err = fh.ReadWithMrdKernelReader(ctx, req)
+		resp, err = fh.ReadWithKernelReader(ctx, req)
 		op.BytesRead = resp.Size
 		op.Data = resp.Data
 		op.Callback = resp.Callback
--- a/internal/fs/handle/file.go
+++ b/internal/fs/handle/file.go
@@ -57,9 +57,9 @@ type FileHandle struct {
 	// GUARDED_BY(mu)
 	readManager gcsx.ReadManager
 
-	// MrdKernelReader is a reader that uses an MRD instance to read data from a GCS
-	// object. This reader is kernel-optimized & reads whatever is requested as is.
-	mrdKernelReader *gcsx.MrdKernelReader
+	// kernelReader is a reader that uses a kernel-optimized approach to read data
+	// from a GCS object based on the bucket type (MRD for Zonal, Range for Regional).
+	kernelReader gcsx.Reader
 	// fileCacheHandler is used to get file cache handle and read happens using that.
 	// This will be nil if the file cache is disabled.
 	// Exactly one of these should be set:
@@ -118,7 +118,7 @@ func NewFileHandle(
 	}
 
 	if c.FileSystem.EnableKernelReader {
-		fh.mrdKernelReader = gcsx.NewMrdKernelReader(inode.GetMRDInstance(), metricHandle)
+		fh.kernelReader = gcsx.NewKernelReader(inode.Bucket(), inode.Source(), inode.GetMRDInstance(), metricHandle)
 	}
 
 	fh.inode.RegisterFileHandle(fh.openMode.AccessMode() == util.ReadOnly)
@@ -143,9 +143,9 @@ func (fh *FileHandle) Destroy() {
 	if fh.readManager != nil {
 		fh.readManager.Destroy()
 	}
-	if fh.mrdKernelReader != nil {
-		fh.mrdKernelReader.Destroy()
-		fh.mrdKernelReader = nil
+	if fh.kernelReader != nil {
+		fh.kernelReader.Destroy()
+		fh.kernelReader = nil
 	}
 }
 
@@ -279,11 +279,11 @@ func (fh *FileHandle) ReadWithReadManager(ctx context.Context, req *gcsx.ReadReq
 	return readResponse, nil
 }
 
-// ReadWithMrdKernelReader reads data at the given offset using the mrd kernel reader.
+// ReadWithKernelReader reads data at the given offset using the kernel reader.
 //
 // LOCKS_REQUIRED(fh.inode.mu)
 // UNLOCK_FUNCTION(fh.inode.mu)
-func (fh *FileHandle) ReadWithMrdKernelReader(ctx context.Context, req *gcsx.ReadRequest) (gcsx.ReadResponse, error) {
+func (fh *FileHandle) ReadWithKernelReader(ctx context.Context, req *gcsx.ReadRequest) (gcsx.ReadResponse, error) {
 	if !fh.inode.SourceGenerationIsAuthoritative() {
 		// Read from inode if source generation is not authoritative.
 		defer fh.inode.Unlock()
@@ -295,11 +295,11 @@ func (fh *FileHandle) ReadWithMrdKernelReader(ctx context.Context, req *gcsx.Rea
 	fh.mu.RLock()
 	defer fh.mu.RUnlock()
 
-	if fh.mrdKernelReader == nil {
-		return gcsx.ReadResponse{}, errors.New("mrdKernelReader is not initialized")
+	if fh.kernelReader == nil {
+		return gcsx.ReadResponse{}, errors.New("kernelReader is not initialized")
 	}
 
-	return fh.mrdKernelReader.ReadAt(ctx, req)
+	return fh.kernelReader.ReadAt(ctx, req)
 }
 
 // Equivalent to locking fh.Inode() and calling fh.Inode().Read, but may be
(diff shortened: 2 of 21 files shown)


internal/fs/handle/file.go
```diff
@@ -118,7 +119,7 @@ func NewFileHandle(
 	}
 
 	if c.FileSystem.EnableKernelReader {
-		fh.mrdKernelReader = gcsx.NewMrdKernelReader(inode.GetMRDInstance(), metricHandle)
+		fh.kernelReader = gcsx.NewKernelReader(inode.Bucket(), inode.Source(), inode.GetMRDInstance(), metricHandle)
```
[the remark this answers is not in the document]
abhishek10004: This could happen when a local file was synced to GCS and then read again. Have fixed this with the new set of changes.

internal/fs/handle/file.go
```diff
@@ -57,9 +58,9 @@ type FileHandle struct {
 	// GUARDED_BY(mu)
 	readManager gcsx.ReadManager
 
-	// MrdKernelReader is a reader that uses an MRD instance to read data from a GCS
-	// object. This reader is kernel-optimized & reads whatever is requested as is.
-	mrdKernelReader *gcsx.MrdKernelReader
+	// kernelReader is a reader that uses a kernel-optimized approach to read data
+	// from a GCS object based on the bucket type (MRD for Zonal, Range for Regional).
```
vadlakondaswetha: Range -> RangeReader or kernelRangeReader?

internal/fs/handle/file_test.go
```diff
@@ -643,7 +643,7 @@ func (t *fileTest) Test_Read_ReaderInvalidatedByGenerationChange() {
 	assert.Equal(t.T(), content2, output)
 }
 
-func (t *fileTest) Test_ReadWithMrdKernelReader_Success() {
```
vadlakondaswetha, line 646: All these methods are using MRDReader for reading data. 
Either make these methods run for both readers or include MRD in the test method name.
Test_ReadWithMRDKernelReader_Success or 
Test_ReadWithKernelReader_MRDReader

Also please add tests covering kernel rangereader also.

internal/gcsx/kernel_range_reader.go
```diff
@@ -0,0 +1,105 @@
...
+func NewKernelRangeReader(bucket gcs.Bucket, instance *KernelRangeReaderInstance, metricHandle metrics.MetricHandle) *KernelRangeReader {
+	return &KernelRangeReader{
+		bucket:   bucket,
+		instance: instance,
+		metrics:  metricHandle,
+	}
+}
+
+// CheckInvariants performs internal consistency checks on the reader state.
+func (rkr *KernelRangeReader) CheckInvariants() {
```
vadlakondaswetha: nit: rkr -> krr?

internal/gcsx/kernel_readers/kernel_range_reader_instance.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"sync"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs"
+)
+
+// KernelRangeReaderInstance manages a reference to the latest gcs.MinObject
+// for a standard/regional bucket, providing a thread-safe way to update
+// it when the object is synced or mutated on GCS.
+type KernelRangeReaderInstance struct {
```
vadlakondaswetha, line 26: why do we need seperate instance. why not put this minObject inside kernel_range_reader.go itself

internal/gcsx/kernel_readers/kernel_range_reader.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2026 Google LLC
```
vadlakondaswetha, line 1: can we put all these kernel readers and factory files under a seperate folder
gcsx/kernel_readers/

internal/gcsx/kernel_range_reader.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+// ReadAt reads data from the object by creating a new range reader.
+func (rkr *KernelRangeReader) ReadAt(ctx context.Context, req *ReadRequest) (ReadResponse, error) {
+	var resp ReadResponse
+
+	obj := rkr.instance.GetMinObject()
+	if obj == nil {
+		return resp, io.EOF
```
vadlakondaswetha: when object is nil, why are we returning EOF?

internal/gcsx/kernel_range_reader.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return resp, fmt.Errorf("failed to create range reader: %w", err)
+	}
+	defer func() { _ = reader.Close() }()
+
+	n, err := io.ReadFull(reader, req.Buffer[:endOffset-req.Offset])
+	resp.Size = n
+
+	if rkr.metrics != nil {
+		metrics.CaptureGCSReadMetrics(rkr.metrics, metrics.ReadTypeParallelAttr, int64(n))
```
vadlakondaswetha: what is the difference between these 2 metrics?
why do we want to tag the read as ReadTypePArallelAttr if we know thats the only flow thats going to get triggered?

internal/gcsx/mrd_kernel_reader.go
```diff
@@ -66,6 +66,15 @@ func isShortRead(bytesRead int, bufferSize int, err error) bool {
 	return false
 }
 
+// ReaderName returns the name of the reader.
+func (mkr *MrdKernelReader) ReaderName() string {
```
vadlakondaswetha: can we keep the naming consistent - kernelMRDReader

internal/fs/inode/file.go
```diff
@@ -190,6 +191,10 @@ func NewFileInode(
 		}
 	}
 
+	if !f.bucket.BucketType().Zonal {
```
vipnydav: Can we put this statement in `else` block? 

`f.bucket.BucketType().IsRapid()` contains `Zonal` and `Pirlo` buckets and I believe Kernel Range Reader would be used for other buckets.

internal/fs/fs.go
```diff
@@ -281,7 +281,7 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		} else {
 			logger.Warnf("Cannot apply bucket-type optimizations as ViperConfig is nil")
 		}
-		// Write post mount kernel settings for Zonal Buckets when kernel reader is enabled in GKE environments for
+		// Write post mount kernel settings when kernel reader is enabled in GKE environments for
 		// non dynamic mounts before user space mounting in GCSFuse. Mounting in GKE is already done at this point but
```
vipnydav, line 285: Would it still be done only for non dynamic mounts?

internal/gcsx/kernel_reader_factory.go
```diff
@@ -0,0 +1,37 @@
...
+// throughput via connection pooling and multi-range downloads.
+// For Standard (Regional) buckets, it uses the Range-based kernel reader which
+// leverages the kernel read-ahead and creates specific range requests.
+func NewKernelReader(
+	bucket gcs.Bucket,
+	kernelRangeReaderInstance *KernelRangeReaderInstance,
+	mrdInstance *MrdInstance,
+	metricsHandle metrics.MetricHandle,
+) Reader {
+	if bucket.BucketType().Zonal {
```
vipnydav: We should use `IsRapid()` method here and at other places as well because `NewMrdKernelReader` will be used for both Zonal and Pirlo Buckets. Update the comment accordingly.

internal/gcsx/kernel_range_reader.go
abhishek10004: Thanks for catching. Fixed it.

internal/gcsx/kernel_readers/kernel_range_reader_instance.go
abhishek10004: Because inode does not have access to the kernel range reader instance to update the min object. There were 2 ways to handle it, I can either keep the kernel reader at the inode level itself or follow the current design pattern (that we have for MRD). I chose the 2nd for better readability.

internal/gcsx/kernel_range_reader.go
abhishek10004: Both of these are being captured in the current read flow. Hence, I'm capturing these here as well (both would be same in this case).

For range reader, DownloadBytes gets incremented by the length of the reader when the reader is created (irrespective of what is read from it), whereas read bytes track the exact bytes read from the read.

vadlakondaswetha: as both the metrics are same for kernel reader, can we deprecate one

abhishek10004: Agreed but we do not want inconsistencies in metrics in different read paths. Checked with Kislay as well and he also wants to keep them consistent in both read paths.

internal/gcsx/kernel_readers/kernel_range_reader.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+// ReadAt reads data from the object by creating a new range reader.
+func (krr *KernelRangeReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (gcsx.ReadResponse, error) {
+	var resp gcsx.ReadResponse
+
+	obj := krr.instance.GetMinObject()
+	if obj == nil {
+		err := fmt.Errorf("KernelRangeReader::ReadAt: Nil MinObject")
```
vadlakondaswetha: Wont this error get logged by default before returning to application. do we need to log it explicitly?
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/wrappers/error_mapping.go#L123

internal/gcsx/kernel_readers/kernel_range_reader.go
```diff
@@ -0,0 +1,109 @@
...
+			Range: &gcs.ByteRange{
+				Start: uint64(req.Offset),
+				Limit: uint64(endOffset),
+			},
+			ReadCompressed: obj.HasContentEncodingGzip(),
+		})
+	if err != nil {
+		return resp, fmt.Errorf("failed to create range reader: %w", err)
+	}
+	defer func() { _ = reader.Close() }()
```
vadlakondaswetha: should we atleast log a warning that something failed while closing the reader?

vadlakondaswetha (dismissed):

internal/gcsx/kernel_readers/kernel_range_reader.go
abhishek10004: Yes, it will get logged. Removed the log from here.

vadlakondaswetha (approved):

PR #4678, opened 2026-05-11
feat(caching): add negative entry caching to stat cache

alleaditya:
### Description

This change implements negative entry caching (non-existent path caching) to optimize workloads that aggressively poll missing files (e.g., JupyterLab).

- **New Flag**: `--enable-nonexistent-entry-caching` (boolean, default `false`).
- **Proactive Listing Cache**: Empty directory listings (`ListObjects` returning 0 results) are proactively cached as negative directory entries to fully protect against implicit directory network probes.
- **VFS Routing**: `LookUpChild` tracks definitive negative cache hits and short-circuits immediately in memory, avoiding network fallback.

#### Benchmark Results
A custom benchmarking script was executed against a locally mounted test bucket to measure performance over a **sustained 60-second polling window** per scenario (aggregating over 1.3 million total VFS operations). The bucket was configured with a `5s` negative cache TTL (`--metadata-cache-negative-ttl-secs=5`).

##### 1. VFS Throughput & Latency Distributions
Memory interception nearly **doubles total application throughput** while slashing tail latencies by over 40%:

| Metric | Negative Caching `DISABLED` | Negative Caching `ENABLED` | Improvement |
| :--- | :--- | :--- | :--- |
| **Total Ops Completed** | `458,234 lookups` | `877,848 lookups` | **91.5% more throughput** |
| **Average Latency** | `0.1303 ms` (130 µs) | `0.0679 ms` (68 µs) | **48% faster** |
| **Median (P50)** | `0.1171 ms` (117 µs) | `0.0602 ms` (60 µs) | **49% faster** |
| **P99 Tail Latency** | `0.2939 ms` (294 µs) | `0.1729 ms` (173 µs) | **41% faster tail** |

##### 2. TTL Cache Expiration Mechanics (Trace Logs)
Over a 60-second window with a 5-second TTL, exactly **12 cache expirations** are mathematically expected.

- **Disabled**: Triggers **82,268 backend network calls** continuously over the 60-second window.
- **Enabled**: Triggers exactly **24 backend network calls** for the entire minute. 
  - **Proof**: Each of the 12 TTL expirations triggers exactly 2 backend calls (re-verifying both `missing_file` and `missing_file/` to refresh the cache). `12 expirations × 2 calls = 24 calls`. Every other request is intercepted entirely in user-space memory.

https://b.corp.google.com/issues/511786738

### Testing details

1. Manual - Verified sustained 60-second polling behavior against local mounts, validating exact TTL expiration counts and throughput improvements. ([Test script](https://paste.googleplex.com/5738355712196608))
2. Unit tests - Added `WrappedSaysNotFound_NegativeCachingDisabled`, `CacheHit_Negative_Disabled_FetchOnly`, and `EmptyListing_NegativeCaching` to verify strict bypass when disabled and proper tombstone storage when enabled.
3. Integration tests - Ran existing E2E caching suites successfully.
    * Perf test and HNS integration test: [Sponge Link](https://fusion2.corp.google.com/ci/kokoro/prod:gcsfuse%2Fgcp_ubuntu%2Fpresubmits%2Fperf_tests%2Fpresubmit/activity/3888bf26-e4d9-42ea-a11e-c5d617fe4333/summary)

| Master |  0.25MiB   | 526.26MiB/s  |  1.3MiB/s  |  72.87MiB/s  |  1.24MiB/s   |
|   PR   |  0.25MiB   | 529.28MiB/s  | 1.34MiB/s  |  75.82MiB/s  |  1.26MiB/s   |
| Master | 48.828MiB  | 4162.55MiB/s | 77.69MiB/s | 1456.75MiB/s |  79.01MiB/s  |
|   PR   | 48.828MiB  | 4214.7MiB/s  | 75.63MiB/s | 1402.22MiB/s |  76.78MiB/s  |
| Master | 976.562MiB | 4176.32MiB/s | 35.85MiB/s | 804.39MiB/s  |  38.21MiB/s  |
|   PR   | 976.562MiB | 4052.01MiB/s | 34.0MiB/s  |  971.4MiB/s  |  37.4MiB/s   |

No.

---

TAG=agy
CONV=3da12d68-a026-456e-9f11-16369d509c21

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -682,6 +682,8 @@ type MetadataCacheConfig struct {
 
 	EnableMetadataPrefetch bool `yaml:"enable-metadata-prefetch"`
 
+	EnableNonexistentEntryCaching bool `yaml:"enable-nonexistent-entry-caching"`
+
 	EnableNonexistentTypeCache bool `yaml:"enable-nonexistent-type-cache"`
 
 	ExperimentalMetadataPrefetchOnMount string `yaml:"experimental-metadata-prefetch-on-mount"`
@@ -992,6 +994,8 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 	if err := flagSet.MarkHidden("enable-mount-retries"); err != nil {
 		return err
 	}
+
+	flagSet.BoolP("enable-nonexistent-entry-caching", "", false, "Cache paths confirmed to not exist (404s) in the unified stat cache.")
 
 	flagSet.BoolP("enable-new-reader", "", true, "Enables support for new reader implementation.")
 
@@ -1601,6 +1605,10 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 	}
 
 	if err := v.BindPFlag("gcs-retries.enable-mount-retries", flagSet.Lookup("enable-mount-retries")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("metadata-cache.enable-nonexistent-entry-caching", flagSet.Lookup("enable-nonexistent-entry-caching")); err != nil {
 		return err
 	}
 
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -1002,7 +1002,6 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 	if err := flagSet.MarkHidden("enable-new-reader"); err != nil {
 		return err
 	}
-
 	flagSet.BoolP("enable-nonexistent-type-cache", "", false, "Once set, if an inode is not found in GCS, a type cache entry with type NonexistentType will be created. This also means new file/dir created might not be seen. For example, if this flag is set, and metadata-cache-ttl-secs is set, then if we create the same file/node in the meantime using the same mount, since we are not refreshing the cache, it will still return nil. This flag has been deprecated in favour of a single unified flag metadata-cache-negative-ttl-secs.")
 
 	flagSet.BoolP("enable-rapid-appends", "", true, "Enables support for appends to unfinalized object using streaming writes")
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -995,13 +995,14 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.BoolP("enable-nonexistent-entry-caching", "", false, "Cache paths confirmed to not exist (404s) in the unified stat cache.")
-
 	flagSet.BoolP("enable-new-reader", "", true, "Enables support for new reader implementation.")
 
 	if err := flagSet.MarkHidden("enable-new-reader"); err != nil {
 		return err
 	}
+
+	flagSet.BoolP("enable-nonexistent-entry-caching", "", false, "Cache paths confirmed to not exist (404s) in the unified stat cache.")
+
 	flagSet.BoolP("enable-nonexistent-type-cache", "", false, "Once set, if an inode is not found in GCS, a type cache entry with type NonexistentType will be created. This also means new file/dir created might not be seen. For example, if this flag is set, and metadata-cache-ttl-secs is set, then if we create the same file/node in the meantime using the same mount, since we are not refreshing the cache, it will still return nil. This flag has been deprecated in favour of a single unified flag metadata-cache-negative-ttl-secs.")
 
 	flagSet.BoolP("enable-rapid-appends", "", true, "Enables support for appends to unfinalized object using streaming writes")
@@ -1607,11 +1608,11 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	if err := v.BindPFlag("metadata-cache.enable-nonexistent-entry-caching", flagSet.Lookup("enable-nonexistent-entry-caching")); err != nil {
+	if err := v.BindPFlag("enable-new-reader", flagSet.Lookup("enable-new-reader")); err != nil {
 		return err
 	}
 
-	if err := v.BindPFlag("enable-new-reader", flagSet.Lookup("enable-new-reader")); err != nil {
+	if err := v.BindPFlag("metadata-cache.enable-nonexistent-entry-caching", flagSet.Lookup("enable-nonexistent-entry-caching")); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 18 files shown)


metrics/noop_metrics.go
```diff
@@ -20,43 +20,41 @@ import (
...
 
-func (*noopMetrics) GcsRequestLatencies(ctx context.Context, latency time.Duration, gcsMethod GcsMethod) {
-}
+	func (*noopMetrics) GcsRetryCount(inc int64, retryErrorCategory RetryErrorCategory){}
 
-func (*noopMetrics) GcsRetryCount(inc int64, retryErrorCategory RetryErrorCategory) {}
+	func (*noopMetrics) TestUpdownCounter(inc int64){}
 
-func (*noopMetrics) TestUpdownCounter(inc int64) {}
+	func (*noopMetrics) TestUpdownCounterWithAttrs(inc int64, requestType RequestType){}
```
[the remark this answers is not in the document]
alleaditya: Reverted the formatting changes.

metrics/metric_handle.go
```diff
@@ -22,177 +22,181 @@ import (
...
 )
-
 // RetryErrorCategory is a custom type for the retry_error_category attribute.
 type RetryErrorCategory string
-
 const (
-	RetryErrorCategoryOTHERERRORSAttr        RetryErrorCategory = "OTHER_ERRORS"
+	RetryErrorCategoryOTHERERRORSAttr RetryErrorCategory = "OTHER_ERRORS"
 	RetryErrorCategorySTALLEDREADREQUESTAttr RetryErrorCategory = "STALLED_READ_REQUEST"
 )
```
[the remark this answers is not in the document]
alleaditya: Haven't modified this file and is not present in the change list either.

cfg/config.go
```diff
@@ -999,6 +1001,8 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	flagSet.BoolP("enable-nonexistent-entry-caching", "", false, "Cache paths confirmed to not exist (404s) in the unified stat cache.")
```
vadlakondaswetha, line 1004: we already have a negative-stat-cache-ttl. why not put this feature also under that flag.
can we do a bit of thinking here for pros and cons.

Now we have one flag for ttl and another for enabling/disabing the feature

cfg/config.go
```diff
@@ -682,6 +682,8 @@ type MetadataCacheConfig struct {
 
 	EnableMetadataPrefetch bool `yaml:"enable-metadata-prefetch"`
 
+	EnableNonexistentEntryCaching bool `yaml:"enable-nonexistent-entry-caching"`
```
vadlakondaswetha, line 685: nit: what does entry-caching mean?

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -253,6 +267,11 @@ func (b *fastStatBucket) addNegativeEntryForFolder(name string) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
+	if !b.enableNonexistentEntryCaching {
```
vadlakondaswetha, line 270: why did you put this check for folders?

internal/fs/inode/dir.go
```diff
@@ -696,19 +697,27 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
+			dirCacheMiss = true
+		} else if err != nil {
 			return nil, err
 		}
 
 		// 2. Try File ONLY if directory wasn't found
+		var fileCacheMiss bool
 		fileResult, err := findExplicitInode(ctx, d.Bucket(), NewFileName(d.Name(), name), true)
-		if err != nil && !errors.As(err, &cacheMissErr) {
+		if fileResult != nil {
```
vadlakondaswetha, line 709: i didnt follow. why these changes are required

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -124,6 +128,11 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 		return
 	}
 
+	if len(listing.MinObjects) == 0 && len(listing.CollapsedRuns) == 0 && b.enableNonexistentEntryCaching && strings.HasSuffix(dirName, "/") {
```
vadlakondaswetha, line 131: can you add a comment what is this check doing. when does minObjects and collapsed Runs are empty.

alleaditya:
Closing this PR in favor of two smaller, independent PRs:
1. Emulator test fixes: #4728
2. Negative stat cache implementation: #4729

cfg/config.go
alleaditya: It was implying "creating an entry".

cfg/config.go
alleaditya: Should moving to use "negative-ttl-secs" alone be a good measure without needing to maintain two separate flags? I am in favor of the decision.

I chose entry-caching to align with the terminology of 'entries' in the stat cache (which stores directory entries).

internal/storage/caching/fast_stat_bucket.go
alleaditya: The fastStatBucket handles caching for both files and folders. addNegativeEntryForFolder is called when a folder lookup fails. If `enable-nonexistent-entry-caching` is set to false, we must ensure we do not cache negative entries for folders either.
This check ensures that when the feature is disabled, we do not add a negative entry for the folder, and instead we explicitly call `b.cache.Erase(name)` to invalidate any potentially stale positive entry from the cache.

internal/fs/inode/dir.go
alleaditya: * Currently, `LookUpChild` checks the cache for a directory first. If it hits a negative cache entry (returning NotFoundError), it immediately returns ENOENT to the kernel without checking if a file with the same name exists. This is incorrect because a file foo can exist even if the directory foo/ does not.
* These changes ensure we check the cache for both the `directory (foo/)` and the `file (foo)`. We only short-circuit and return ENOENT (definitive negative hit) if both lookups hit the cache and both confirmed non-existence (i.e., neither was a cache miss). If either check results in a cache miss, we must fall back to GCS (fetchCoreEntity) to ensure correctness.

internal/storage/caching/fast_stat_bucket.go
alleaditya: * "I will add a comment. To clarify: MinObjects and CollapsedRuns are empty when GCS returns an empty listing for a directory prefix. This happens when the directory is implicit and all its children have been deleted (meaning it no longer exists), or when the directory never existed. In these cases, if enable-nonexistent-entry-caching is true, we cache the directory path (with a trailing slash) as a negative entry to prevent future redundant listing requests.
* Explicit empty directories contain a placeholder object (e.g., dir/) which is returned in MinObjects, so they are not cached as negative entries. I will add this explanation as a comment in the next commit."

PR #4689, opened 2026-05-13
feat(pirlo): support rapid writes state and write behavior changes for Pirlo buckets

vipnydav:
### Description
This PR updates GCSFuse to support specialized write behaviors for Pirlo buckets, specifically introducing a state-based approach for rapid writes. It transitions the bucket type classification from a boolean to an enum state and adjusts synchronization, upload, and read logic accordingly.

### **Key Changes**

*   **Pirlo State Enum**: Updated the `Pirlo` field in `BucketType` from a boolean to a `PirloState` enum, which can now represent `PirloStateRapidEnabled` or `PirloStateRapidDisabled`.
*   **Synchronized Flushing**: Modified the buffered write handler to flush pending writes during a sync operation specifically for Pirlo buckets that have rapid writes enabled.
*   **Appendable Object Writer**: Configured the storage handle to use appendable semantics (`wc.Append = true`) for Pirlo buckets only when they are in the rapid writes enabled state.
*   **Compose API Restriction**: Disabled the use of the `Compose` API for Pirlo buckets when rapid writes are enabled, as this API is not supported for those buckets; instead, the system relies on the appendable API.
*   **Read Logic & Size Checks**: Updated the file handle logic to ensure size checks are performed (not skipped) when rapid writes as disabled for pirlo bucket.
*   **Upload Handling**: Adjusted the upload handler to prefer creating an appendable object writer for Pirlo buckets that support rapid writes.

### Testing details
1. Manual - Done
2. Unit tests - Updated

### DIFF
--- a/internal/bufferedwrites/buffered_write_handler_test.go
+++ b/internal/bufferedwrites/buffered_write_handler_test.go
@@ -305,6 +305,15 @@ func (testSuite *BufferedWriteTest) TestFlush_SizeMismatch_ReturnsError() {
 			bucketType: gcs.BucketType{Zonal: true},
 			obj:        &gcs.Object{Name: "testObject", Size: 0},
 		},
+		{
+			name:       "pirlo_new_file_rapid_writes",
+			bucketType: gcs.BucketType{Pirlo: gcs.PirloStateRapidWritesEnabled},
+		},
+		{
+			name:       "pirlo_append_rapid_writes",
+			bucketType: gcs.BucketType{Pirlo: gcs.PirloStateRapidWritesEnabled},
+			obj:        &gcs.Object{Name: "testObject", Size: 0},
+		},
 	}
 	for _, tc := range testCases {
 		testSuite.Run(tc.name, func() {
@@ -312,7 +321,7 @@ func (testSuite *BufferedWriteTest) TestFlush_SizeMismatch_ReturnsError() {
 			mockBucket.On("BucketType").Return(tc.bucketType)
 			writer := &storagemock.Writer{}
 			writer.On("Write", mock.Anything).Return(2, nil)
-			if tc.bucketType.Zonal && tc.obj != nil {
+			if (tc.bucketType.Zonal || tc.bucketType.Pirlo == gcs.PirloStateRapidWritesEnabled) && tc.obj != nil {
 				mockBucket.On("CreateAppendableObjectWriter", mock.Anything, mock.Anything).Return(writer, nil)
 			} else {
 				mockBucket.On("CreateObjectChunkWriter", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(writer, nil)
@@ -358,6 +367,15 @@ func (testSuite *BufferedWriteTest) TestSync_SizeMismatch_ReturnsError() {
 			bucketType: gcs.BucketType{Zonal: true},
 			obj:        &gcs.Object{Name: "testObject", Size: 0},
 		},
+		{
+			name:       "pirlo_new_file_rapid_writes",
+			bucketType: gcs.BucketType{Pirlo: gcs.PirloStateRapidWritesEnabled},
+		},
+		{
+			name:       "pirlo_append_rapid_writes",
+			bucketType: gcs.BucketType{Pirlo: gcs.PirloStateRapidWritesEnabled},
+			obj:        &gcs.Object{Name: "testObject", Size: 0},
+		},
 	}
 	for _, tc := range testCases {
 		testSuite.Run(tc.name, func() {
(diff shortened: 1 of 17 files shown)


internal/storage/gcs/bucket.go
```diff
@@ -21,15 +21,29 @@ import (
 	"golang.org/x/net/context"
 )
 
+type PirloState int
+
+const (
```
anushka567, line 26: Can you please add a short comment regarding the three states mentioned here?

internal/storage/bucket_handle_test.go
```diff
@@ -594,6 +594,56 @@ func (testSuite *BucketHandleTest) TestBucketHandle_CreateObjectChunkWriter() {
...
+			bucketType:          gcs.BucketType{Pirlo: gcs.PirloStateRapidWritesEnabled},
+			finalizeFileOnClose: true,
+			expectedAppend:      true,
+		},
+		{
+			name:                "PirloBucket_RapidDisabled",
+			bucketType:          gcs.BucketType{Pirlo: gcs.PirloStateRapidWritesDisabled},
+			finalizeFileOnClose: false,
+			expectedAppend:      false,
+		},
```
anushka567, line 627: Do we need any test corresponding to the following state : `PirloStateNone` , if it is already covered under standard bucket , then maybe you could highlight that.

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -103,6 +103,26 @@ func (t *UploadHandlerTest) TestCreateObjectWriter_CreateAppendableObjectWriterC
...
+func (t *UploadHandlerTest) TestCreateObjectWriter_CreateObjectChunkWriterCalledForPirloRapidDisabled() {
+	t.createUploadHandlerWithObjectOfGivenSize(objectSize, time.Time{})
+	t.mockBucket.On("BucketType").Return(gcs.BucketType{Pirlo: gcs.PirloStateRapidWritesDisabled})
+	t.mockBucket.On("CreateObjectChunkWriter", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(&storagemock.Writer{}, nil)
+
+	_ = t.uh.createObjectWriter(context.Background())
+
+	t.mockBucket.AssertCalled(t.T(), "CreateObjectChunkWriter", mock.Anything, mock.Anything)
+}
+
```
anushka567, line 145: Nit: Can you please group these tests into a table-driven test using subtests in order to reduce code duplication.

internal/bufferedwrites/buffered_write_handler_test.go
```diff
@@ -484,7 +512,7 @@ func (testSuite *BufferedWriteTest) TestSyncPartialBlockTableDriven() {
 			assert.Equal(testSuite.T(), 0, bwhImpl.uploadHandler.blockPool.TotalFreeBlocks())
 			// Read the object from back door.
 			content, err := storageutil.ReadObject(context.Background(), bwhImpl.uploadHandler.bucket, bwhImpl.uploadHandler.objectName)
-			if tc.bucketType.Zonal {
+			if tc.bucketType.Zonal || tc.bucketType.Pirlo == gcs.PirloStateRapidWritesEnabled {
```
anushka567: I see this condition repeated several times in this test file. Can we create a utility method for the same?

internal/gcsx/syncer.go
```diff
@@ -59,10 +59,10 @@ func NewSyncer(
 		bucket: bucket,
 	}
 
-	// Zonal buckets do not currently support Compose, so we always write objects
-	// in their entirety.
+	// Zonal buckets and Pirlo buckets with rapid writes enabled do not currently
```
anushka567: Nit: `Zonal buckets and Pirlo buckets with rapid writes enabled` ->  Rapid buckets, since this is also how we have collectively referenced such buckets at internal/bufferedwrites/buffered_write_handler.go#219

internal/storage/bucket_handle_test.go
vipnydav: Since `PirloStateNone` is the first element in the iota enumeration, it acts as the default zero value for the `PirloState` type. Therefore, when the test initializes `gcs.BucketType{}`, the Pirlo field implicitly defaults to `gcs.PirloStateNone`.

internal/bufferedwrites/upload_handler_test.go
vipnydav: Utilized `BucketType`'s existing `RapidWritesEnabled()` method .

anushka567 (approved):
LGTM.

vadlakondaswetha (commented):
LGTM on the code changes.
Relying on Anushka's review for unit tests

vadlakondaswetha (approved):
LGTM on the code changes.
Relying on Anushka's review for unit tests

PR #4690, opened 2026-05-14
chore(cd scripts): Make release package bucket configurable option in CD scripts

meet2mky:
### Description
Make release package bucket configurable option in cd scripts.

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - NA


abhishek10004:
Have we made changes at all the places where this script is called to pass the release bucket as a parameter now?

meet2mky:
> Have we made changes at all the places where this script is called to pass the release bucket as a parameter now?

Will do once this change is done. It's called only from the louhi flow 1.

abhishek10004 (approved):

PR #4724, opened 2026-05-23
test: add config dump integration tests

### LINKED ISSUE #2134: Add integration tests to detect that the configuration is dumped in the stdout/logfile.
Background:
GCSFuse dumps the configuration that it was started with in the log file/standard output depending upon whether it was started in background/foreground mode. We'd like to add integration tests to verify that the logging does happen.

Motivation:
Configuration logging helps determine what was the configuration of GCSFuse when it was run. It comes in handy while debugging issues.

Acceptance Criteria:
Add an integration test that should test whether the configuration is being dumped into the log file. We don't want to compare exactly whether all the flags are being written to since that'll lead to it becoming a change-detector test; we just want to know whether the dumping is happening.

Haihan-Jiang:
Fixes #2134

## What changed
- Added mounting integration coverage for config dumps in foreground mode without a log file, where logs go to stdout.
- Added mounting integration coverage for config dumps in background mode with `--log-file`, where logs go to the configured log file.
- Assert only stable config-dump markers instead of comparing the full flag/config output.

## Why
Configuration dumps are useful debugging evidence, and this verifies that dumping keeps happening without turning the test into a full config snapshot.

## Validation
- `gofmt -w tools/integration_tests/mounting/gcsfuse_test.go`
- `git diff --check`
- `go test ./tools/integration_tests/mounting`

### DIFF
--- a/tools/integration_tests/mounting/gcsfuse_test.go
+++ b/tools/integration_tests/mounting/gcsfuse_test.go
@@ -129,6 +129,19 @@ func waitForLogFileToContainConfigDump(logFile string) {
 	assertContainsConfigDump(output)
 }
 
+func readStdoutUntilMounted(stdout io.Reader) []byte {
+	var output []byte
+	buf := make([]byte, 4096)
+
+	for !bytes.Contains(output, []byte("successfully mounted")) {
+		n, err := stdout.Read(buf)
+		output = append(output, buf[:n]...)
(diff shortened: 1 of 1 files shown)


Haihan-Jiang:
This is ready for review from my side. The visible checks are green; could someone take a look when convenient?

PR #4729, opened 2026-05-25
feat(caching): implement negative stat cache to optimize polling of missing files

alleaditya:
### Description

This change implements negative entry caching (non-existent path caching) to optimize workloads that aggressively poll missing files (e.g., JupyterLab).

• **Activation**: Controlled by `metadata-cache: negative-ttl-secs` config parameter (or `--metadata-cache-negative-ttl-secs` flag). It is enabled by default with a 5-second TTL. Setting it to `0` disables the feature.
• **Proactive Listing Cache**: Empty directory listings (`ListObjects` returning 0 results) are proactively cached as negative directory entries to fully protect against implicit directory network probes.
• **VFS Routing**: `LookUpChild` tracks definitive negative cache hits and short-circuits immediately in memory, avoiding network fallback.

#### Benchmark Results

A custom sustained benchmarking script (`bench.sh`) was executed against a locally mounted test bucket to measure performance over sustained 60-second polling windows.

The metrics demonstrate that negative caching successfully eliminates redundant network traffic across all lookup scenarios, including those requiring implicit directory checking.

##### 1. Implicit Dirs vs Explicit Dirs (Throughput & Network Calls)

In `--implicit-dirs` scenarios, FUSE previously required *both* a `StatObject` and a fallback `ListObjects` call per lookup. With this change, both are aggressively short-circuited:

| Scenario (60s Polling) | Negative Caching TTL=0s | Negative Caching TTL=5s | Result |
| :--- | :--- | :--- | :--- |
| **Explicit Dirs** | Ops: 771 <br> `StatObject` calls: 1,544 <br> `ListObjects` calls: 1 | Ops: 845,720 <br> `StatObject` calls: 24 <br> `ListObjects` calls: 1 | **98.4% network drop.** <br> Throughput: +109,591% (>1,000x) |
| **Implicit Dirs** | Ops: 735 <br> `StatObject` calls: 736 <br> `ListObjects` calls: 737 | Ops: 874,744 <br> `StatObject` calls: 12 <br> `ListObjects` calls: 15 | **98.2% network drop.** <br> Throughput: +118,912% (>1,000x) |

*(Note: Network calls exactly match mathematically expected TTL expiration rates - approximately 1 network call per 5 seconds).*

https://b.corp.google.com/issues/511786738

### Testing details
1. **Manual** - Verified sustained polling behavior against local mounts with and without `--implicit-dirs`, validating exact TTL expiration counts and throughput improvements.
2. **Unit tests** - Added `TestLookUpChild_NegativeCacheHit` in `dir_test.go` and `TestGCSMetrics_RequestCount_NegativeCachingShortCircuit` in `gcs_metrics_test.go`. Verified negative caching upon empty listings in `ListObjectsTest_InsertListing.EmptyListing`.
3. **Integration tests** - Ran existing E2E caching suites successfully 
   * [Perf Sponge](http://sponge2.corp.google.com/a28aa343-03d4-4079-a753-a95bc52f8a11)
   * [Integration tests](http://sponge2.corp.google.com/05760153-9061-4dc8-8275-b5af34976c87)
   * [ZB Integration tests](http://sponge2/2024ab3c-75e9-4fb6-bdae-0d9382a2d060)

No.


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -124,6 +124,19 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
...
+	// has any children (because they were deleted) or a non-existent directory.
+	// In either case, if negative caching is enabled (negativeCacheTTL > 0), we cache
+	// this directory name with a trailing slash as a negative entry to short-circuit
+	// subsequent lookups.
+	// Note: Explicit empty directories have a placeholder object (e.g., "dir/"),
+	// which will be returned in MinObjects, so they will not trigger this negative caching.
+	if len(listing.MinObjects) == 0 && len(listing.CollapsedRuns) == 0 && b.negativeCacheTTL > 0 && strings.HasSuffix(dirName, "/") {
+		b.cache.AddNegativeEntry(dirName, b.clock.Now().Add(b.negativeCacheTTL))
+		return
+	}
```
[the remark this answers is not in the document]
alleaditya: Thanks for the catch. I agree that speculative negative caching in 'insertListing' is unsafe due to pagination (empty pages/last page overwriting positive entries). I have removed the negative caching from 'insertListing' in the latest commit (063c48a9ac). Definitive negative caching is still handled correctly in the 'StatObject' and 'GetFolder' lookup paths.

docs/semantics.md
```diff
@@ -209,7 +209,10 @@ The behavior of stat cache is controlled by the following flags/config parameter
    If this config variable is missing, then the value of ```--stat-cache-ttl``` is used.
    * ```--stat-cache-ttl``` commandline flag, which can be set to a value like ```10s``` or ```1.5h```. The default is one minute. This has been deprecated (starting v2.0) and is currently only available for backward compatibility. If ```metadata-cache: ttl-secs``` is set, ```--stat-cache-ttl``` is ignored.
    
-   Positive and negative stat results will be cached for the specified amount of time.
+   Positive stat results (existing files) will be cached for this specified amount of time.
+
+3. **Negative Stat-cache (Non-existent Entry Caching)**: Controls caching of non-existent paths (`ErrNotExist` / 404 results) to optimize workloads that aggressively poll missing files (e.g., JupyterLab).
+   * `metadata-cache: negative-ttl-secs` in the config-file (integer). Sets the TTL in seconds for negative entries. Default is 5 seconds. Use -1 for infinite TTL. Setting this to 0 disables negative caching.
```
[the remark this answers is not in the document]
alleaditya: Good point. Added a warning block in docs/semantics.md detailing the risks of using infinite TTL (-1) for negative caching when files are modified by external clients. Fixed in commit 3bf09de03 (and preserved in 063c48a9ac).

internal/fs/inode/dir.go
```diff
@@ -686,6 +686,7 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
 		// 1. Try Directory FIRST (since it's the preferred return type)
 		var dirResult *Core
 		var err error
+		var dirCacheMiss bool
```
raj-prince: Why can't we add negative caching logic in the fast-stat-bucket when the list response is empty, instead of changing here?

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -244,6 +248,11 @@ func (b *fastStatBucket) addNegativeEntry(name string) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
+	if b.negativeCacheTTL <= 0 {
+		b.cache.Erase(name)
```
raj-prince: Erasing inside addNegativeEntry method doesn't make much sense. Also, we don't need to erase, instead we will not add. 

We can have a check on the top (outside lock to return) or even better if calling method would not call this method if TTL is negative.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -271,6 +285,9 @@ func (b *fastStatBucket) lookUp(name string) (hit bool, m *gcs.MinObject) {
 	defer b.mu.Unlock()
 
 	hit, m = b.cache.LookUp(name, b.clock.Now())
+	if hit && m == nil && b.negativeCacheTTL <= 0 {
```
raj-prince: Do we need this? I feel, b.cache takes care of expiry.

internal/storage/caching/fast_stat_bucket_test.go
```diff
@@ -645,6 +645,15 @@ func (t *StatObjectTest) CacheHit_Negative() {
 	ExpectThat(err, HasSameTypeAs(&gcs.NotFoundError{}))
 }
 
+func (t *StatObjectTest) CacheHit_Negative_Disabled_FetchOnly() {
```
raj-prince: Lets use AAA while writing unit test and separate the section by newline.

raj-prince (changes requested):

internal/storage/caching/fast_stat_bucket_test.go
```diff
@@ -1016,6 +1036,11 @@ func (t *ListObjectsTest_InsertListing) callAndVerify(ctx context.Context, isHNS
 		ExpectCall(t.cache, "InsertImplicitDir")(dir, Any())
 	}
 
+	dirHasContents := len(listing.MinObjects) > 0 || len(listing.CollapsedRuns) > 0
+	if !dirHasContents && prefix != "" && ctx.Err() == nil && !isHNS {
+		ExpectCall(t.cache, "AddNegativeEntry")(prefix, Any())
+	}
```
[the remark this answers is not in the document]
alleaditya: Outdated

internal/storage/caching/fast_stat_bucket.go
alleaditya: This is deleting the positive cache. 

Example:
1. List call caches a positive entry.
2. Delete the object
3. Now look up the same object

#3 to avoid positive cache returning values, we are clearing cache for the positive entry.

internal/fs/inode/dir.go
alleaditya: Resovled

internal/storage/caching/fast_stat_bucket.go
raj-prince: You don't need to do explicitly, as part of 2nd step cache will be invalidated.

internal/fs/gcs_metrics_test.go
```diff
@@ -463,3 +468,82 @@ func TestGCSMetrics_RetryCount(t *testing.T) {
...
+		Parent: fuseops.RootInodeID,
+		Name:   "missing_file",
+	}
+
+	// 1. First Lookup (Cache Miss) -> Emits backend probes through monitoring layer
+	_ = server.LookUpInode(ctx, lookupOp)
+	waitForMetricsProcessing()
+
+	// Probing a missing file checks dir then file -> 2 actual backend network calls.
+	metrics.VerifyCounterMetric(t, ctx, reader, "gcs/request_count",
```
raj-prince, line 538: I am not sure, if we are testing the right flow. Ideally, I would expect one ListObject when implicitDir = true.

internal/fs/gcs_metrics_test.go
```diff
@@ -463,3 +468,82 @@ func TestGCSMetrics_RetryCount(t *testing.T) {
...
+}
+
+func (bm *fakeBucketManagerForShortCircuit) SetUpBucket(ctx context.Context, name string, _ bool, _ metrics.MetricHandle) (gcsx.SyncerBucket, error) {
+	return gcsx.NewSyncerBucket(0, 120, 10, ".gcsfuse_tmp/", gcsx.NewContentTypeBucket(bm.bucket)), nil
+}
+func (bm *fakeBucketManagerForShortCircuit) ShutDown() {}
+
+// TestGCSMetrics_RequestCount_NegativeCachingShortCircuit validates that when negative entry caching is enabled,
+// repeated lookups for non-existent files short-circuit in memory and do not emit redundant backend GCS requests.
+func TestGCSMetrics_RequestCount_NegativeCachingShortCircuit(t *testing.T) {
```
raj-prince, line 484: nit: Can we use newline to separate the AAA instead using newlines at multiple places?

raj-prince (commented):
Overall looks good. Added couple of minor comments.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -135,6 +135,9 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 	if b.implicitDir && dirHasContents && !isDirInListing {
 		b.cache.InsertImplicitDir(dirName, expiration)
 	}
+	if b.implicitDir && !dirHasContents && dirName != "" && b.negativeCacheTTL > 0 {
+		b.cache.AddNegativeEntry(dirName, b.clock.Now().Add(b.negativeCacheTTL))
+	}
```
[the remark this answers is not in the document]
alleaditya: We can rely on the calling method to check for the feature being enabled. Adding the check in here will make it difficult for custom addition of negative entry.

internal/fs/gcs_metrics_test.go
alleaditya: We have now changes this to look for list objects and settings push for `implicitDir = true`

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -546,7 +585,12 @@ func (b *fastStatBucket) StatObjectFromGcs(ctx context.Context,
 	if err != nil {
 		// Special case: NotFoundError -> negative entry.
 		if _, ok := err.(*gcs.NotFoundError); ok {
-			b.addNegativeEntry(req.Name)
+			// Do not add negative cache for directories if implicitDir is enabled,
+			// because they might be implicit directories (which are verified via ListObjects).
+			// We only negatively cache directories after ListObjects confirms they have no children.
+			if !b.implicitDir || !strings.HasSuffix(req.Name, "/") {
```
raj-prince: This will skip adding negative entry for explictDir scenario, could you please check once?

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -146,6 +148,13 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 		return
 	}
 
+	if isDirPath && !dirHasContents && dirName != "" && b.negativeCacheTTL > 0 {
+		hit, m := b.cache.LookUp(dirName, b.clock.Now())
+		if !hit || m == nil {
+			b.cache.AddNegativeEntry(dirName, b.clock.Now().Add(b.negativeCacheTTL))
```
raj-prince: Any reason for doing LookUp before inserting negative entry?
(1) We would be doing lookUp in the ListObject method, right?
(2) Ideally, we should update the expiry timestamp to - latest-fetch-time + ttl, for that we need to update the cache every time we get the latest from the server.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -465,6 +491,19 @@ func (b *fastStatBucket) ListObjects(
 	} else {
 		// note anything we found.
 		b.insertMultipleMinObjects(ctx, listing.MinObjects)
+
+		// Negatively cache empty directories to short-circuit future ListObjects checks.
```
raj-prince: This seems duplicate. This is mostly required `--enable-nonexistent-type-cache` flag is false by default.

Maybe we can skip this change, as we have a way to cache negative implict-dir type for the previous gcsfuse version.
CC: @vadlakondaswetha 

If required, then please create a helper method to re-use the code for negative-implicit-dir caching.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -449,6 +458,23 @@ func (b *fastStatBucket) StatObject(
...
+	// is used to verify the existence of an implicit directory. If this prefix
+	// has a negative cache entry (meaning it doesn't exist and has no descendants),
+	// we can safely short-circuit the network call and return an empty listing.
+	if b.implicitDir && req.Prefix != "" && strings.HasSuffix(req.Prefix, "/") {
+		hit, m := b.lookUp(req.Prefix)
+		if hit && m == nil {
+			if !b.BucketType().Hierarchical {
+				return &gcs.Listing{}, nil
+			}
+			folderHit, f := b.lookUpFolder(req.Prefix)
```
raj-prince: This won't be required mostly. As for HNS bucket we wouldn't be calling Listobject but StatFolder. So, should be handled automatically.

raj-prince (changes requested):

internal/storage/caching/fast_stat_bucket.go
alleaditya: For #2,  isn't the code saying the same with `b.clock.Now().Add(b.negativeCacheTTL)`?

raj-prince: Nope, we won't be updating for the hit scenario. Ideally, we should update for the hit scenario too.

alleaditya: Removing this logic causes managed folder tests to fail due to race conditions. Without the lock, concurrent network requests can lead to stale negative entries overwriting valid positive ones when the bucket state changes in-flight. This check is necessary to maintain cache consistency.

1. **Slow Network Calls:** The `ListObjects` network request is made to GCS without holding the cache lock (``b.mu``). Because network calls are relatively slow, the state of the GCS bucket can change while the request is in flight.
2. **Concurrent Modifications:** Imagine Goroutine A initiates `ListObjects("dir/")` and gets an empty response. While Goroutine A is waiting for that response, Goroutine B successfully creates a file inside that directory (e.g., ``dir/file.txt``), and then populates the cache with a positive entry for `dir/` (as an implicit directory).
3. **Preventing Overwrites:** If Goroutine A's `insertListing` unconditionally called `AddNegativeEntry("dir/")` after receiving its stale, empty listing response, it would overwrite Goroutine B's newer, valid positive entry. The cache would incorrectly report that `dir/` does not exist.

internal/storage/caching/fast_stat_bucket.go
alleaditya: Removed this check

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -449,6 +458,23 @@ func (b *fastStatBucket) StatObject(
...
+			if folderHit && f == nil {
+				return &gcs.Listing{}, nil
+			}
+		}
+	}
```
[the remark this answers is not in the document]
alleaditya: No more relevant

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -465,6 +491,19 @@ func (b *fastStatBucket) ListObjects(
...
+		// of the TypeCache deprecation flag.
+		isDirPath := strings.HasSuffix(req.Prefix, "/")
+		dirHasContents := len(listing.MinObjects) > 0 || len(listing.CollapsedRuns) > 0
+
+		if isDirPath && !dirHasContents && req.Prefix != "" && b.negativeCacheTTL > 0 {
+			hit, m := b.cache.LookUp(req.Prefix, b.clock.Now())
+			if !hit || m == nil {
+				b.cache.AddNegativeEntry(req.Prefix, b.clock.Now().Add(b.negativeCacheTTL))
+			}
+		}
```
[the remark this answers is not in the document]
alleaditya: Not relevant and tests break

internal/fs/inode/dir.go
```diff
@@ -683,31 +683,37 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
 		}
 
 		// If we found a directory, we're done. Return it now.
 		if dirResult != nil {
 			return dirResult, nil
 		}
 		// If we hit a real error (not a cache miss), exit early.
-		if err != nil && !errors.As(err, &cacheMissErr) {
-			return nil, err
+		if dirErr != nil && !errors.As(dirErr, &cacheMissErr) {
```
raj-prince, line 694: Lets do the error handling before we return the dirResult.

internal/fs/inode/dir.go
```diff
@@ -683,31 +683,37 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
+		fileResult, fileErr := findExplicitInode(ctx, d.Bucket(), NewFileName(d.Name(), name), true)
+		if fileErr != nil && !errors.As(fileErr, &cacheMissErr) {
+			return nil, fileErr
 		}
 
 		if fileResult != nil {
 			return fileResult, nil
 		}
+
+		// 3. Both lookups resulted in cache hits (no errors) with no results found,
```
raj-prince:
```suggestion
		// 3. Both lookups resulted in cache hits (no cacheMiss error) with no results found,
```

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -132,7 +136,9 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 	// implicit directory.
 	dirHasContents := len(listing.MinObjects) > 0 || len(listing.CollapsedRuns) > 0
 	isDirInListing := len(listing.MinObjects) > 0 && listing.MinObjects[0].Name == dirName
-	if b.implicitDir && dirHasContents && !isDirInListing {
+	isDirPath := strings.HasSuffix(dirName, "/")
```
raj-prince: Do we need this check, given ListObject is called only for directory only?

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -146,6 +152,12 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 		return
 	}
 
+	isNegativeCacheEnabled := b.negativeCacheTTL > 0 && !b.enableEmptyManagedFolders
```
raj-prince, line 155: Can we add a comment with reasoning to not enable negativeCaching when enableEmptyManagedFolder is true?

internal/storage/caching/fast_stat_bucket_test.go
```diff
@@ -1008,13 +1010,18 @@ func (t *ListObjectsTest_InsertListing) callAndVerify(ctx context.Context, isHNS
 		WillOnce(Return(gcs.BucketType{Hierarchical: isHNS}))
 	ExpectCall(t.wrapped, "ListObjects")(Any(), Any()).
 		WillOnce(Return(listing, nil))
-	// Register expectations.
+	if prefix != "" {
+		ExpectCall(t.cache, "LookUp")(prefix, Any()).WillRepeatedly(Return(false, nil))
```
raj-prince: do we need this?

tools/integration_tests/dentry_cache/stat_test.go
```diff
@@ -45,6 +45,9 @@ func (s *statWithDentryCacheEnabledTest) SetupTest() {
 
 func (s *statWithDentryCacheEnabledTest) TearDownTest() {
 	setup.SaveGCSFuseLogFileInCaseOfFailure(s.T())
+	// Wait for the metadata cache TTL to expire so that negatively cached entries
+	// don't pollute subsequent tests.
+	time.Sleep(2500 * time.Millisecond)
```
raj-prince: Why do we need this change? Also, kind of brittle if someone change the value of metadata-cache-ttl.

raj-prince (commented):
Changes looks good, few minor comments.

internal/storage/caching/fast_stat_bucket.go
alleaditya: If that is the assumption, should we not have an early return at the start of the function?

The reason of adding this into variable is that it was *already* being used in for inserting cache using `InsertImplicitDir`

This line simply extracts out for re-use?

internal/fs/inode/dir.go
alleaditya: Thought I would handle this in a separate bug. However changed if after the comments.

internal/storage/caching/fast_stat_bucket_test.go
alleaditya: Removed. Doesn't seem necessary change

tools/integration_tests/dentry_cache/stat_test.go
alleaditya: Prince, removing this is causing denrty flat bucket [test failures](http://sponge2.corp.google.com/33d8299f-b8fe-4fc0-ad1d-f0951fe19f5c)
> "Running tests with flags: [--implicit-dirs --experimental-enable-dentry-cache --metadata-cache-ttl-secs=2 --client-protocol=grpc --metadata-cache-negative-ttl-secs=2]"

**Why `time.Sleep(2100 * time.Millisecond)` is required in `TestStatWithDentryCacheEnabled`:**

1. **Stale Cache Prevention:** The test suite configures a 2-second stat cache TTL (`--metadata-cache-ttl-secs=2`). Because the test updates the file out-of-band directly on GCS (`client.WriteToObject`), GCSFuse is unaware of the modification. Sleeping `>2s` ensures the cached metadata expires so the subsequent `stat` call fetches the updated object size from GCS rather than returning a stale cache hit.
2. **Preventing Cascading Rate Limits & Cache Pollution:** Removing the sleep causes the test to fail immediately and disrupt timing intervals between test iterations (`#01`). This leads to GCS backend throttling (`HTTP 429 Too Many Requests`) during directory setup and causes negative dentry cache lookups (`ENOENT`) from preceding cleanup phases to bleed into subsequent test runs.

internal/storage/caching/fast_stat_bucket.go
alleaditya: added comments

tools/integration_tests/dentry_cache/stat_test.go
raj-prince: Why we were not seeing the failure before? Is it because of negative implicit-directory caching? And if the test is not implemented with negative implicit-directory caching then we can explicitly disable it.

internal/storage/caching/fast_stat_bucket.go
raj-prince: We can add an early return, but I don't see it was being used already -https://github.com/GoogleCloudPlatform/gcsfuse/blob/9252f247496887783c1aced3c50672672e919141/internal/storage/caching/fast_stat_bucket.go#L117

alleaditya: Removed it as we understand this was actually not present earlier and the call is being made only for directory.

raj-prince (approved):

tools/integration_tests/dentry_cache/stat_test.go
alleaditya: Resolved this offline. Determined it is indeed needed due to negative caching happening in the prior test of dentry cache.

internal/storage/caching/fast_stat_bucket.go
alleaditya: Removed

internal/storage/caching/fast_stat_bucket.go
alleaditya: Removed the check and rephrase the check

internal/storage/caching/fast_stat_bucket.go
alleaditya: We explicitly bypassed this for managed folders now determining the extra lookup wasn't necessary.

Issue #4696, opened 2026-05-18
Feature Request: Warn and block writes when a directory exceeds a configurable entry limit


chaitanyapantheor:
## Summary

When a GCS bucket directory contains a very large number of files, Cloud Storage
FUSE currently has no mechanism to warn operators or degrade gracefully. In
production we observe:

- **~50,000 files in a single directory** → significant read/list performance
  degradation
- **100,000+ files in a single directory** → very poor performance that can
  lead to filesystem state corruption, causing extended downtime and potential
  file loss

There is also no way to prevent new files from being written into an already
over-populated directory, so the problem compounds silently until it causes an
outage.

## Current Behavior

`gcsfuse` performs a full `ListObjects` (paginated) on every `readdir()` call
(when `kernel-list-cache-ttl-secs=0`) or on first access within a TTL window.
With 100k objects under one prefix this means:

- 10+ sequential paginated GCS API calls per listing
- Unbounded memory growth in the metadata/stat cache
- No warning in logs that a directory is approaching a problematic size
- No way to prevent further writes into an already over-populated directory
- No documented hard limit — operators discover the problem only after
  degradation or corruption occurs

### 1. Write guard — block new file creation above a threshold

When a directory's entry count (known from the stat/list cache, or from a
lightweight `ListObjects` prefix count) meets or exceeds a configurable limit,
reject `O_CREAT` operations on that directory with `ENOSPC` and emit a
structured `WARNING` log.

```yaml
file-system:
  max-dir-entries: 50000  # block new creates and warn when parent dir hits this
```

Returning `ENOSPC` gives applications and scripts a clear, standard error
signal (`No space left on device`) rather than silently allowing the directory
to grow into the corruption zone.

### 2. Warning threshold log (pre-limit signal)

Emit a structured `WARNING` log when a `ListObjects` response for a single
directory prefix exceeds a lower configurable threshold, giving operators
time to act before the hard limit is reached.

```yaml
file-system:
  large-dir-warning-threshold: 10000  # log WARNING when readdir exceeds this
```

### 3. Streaming / incremental readdir

Instead of buffering the entire `ListObjects` result before returning to the
kernel, stream entries back incrementally. This caps peak memory usage and
allows the kernel to begin iterating entries before the full GCS list
completes — directly addressing latency and OOM risk at extreme scale.

### 4. Documentation of limits

Publish a supported maximum directory size (entries per prefix) in the
official documentation with explicit guidance on mitigation strategies
[...]

PR #4736, opened 2026-05-27
feat: add experimental large directory warning threshold

Haihan-Jiang:
## What

Adds a hidden experimental `--experimental-large-dir-warning-threshold` flag / config field under `file-system`. When the threshold is greater than zero, directory reads log a warning after the final directory entry list reaches or exceeds the threshold. A value of `0` keeps the behavior disabled.

This covers the warning signal requested in #4696 without changing write-path behavior or claiming to solve the full large-directory guard/streaming work.

## Validation

- `go test ./cfg`
- `docker run --rm -v "$PWD":/work -v gcsfuse-go-mod-cache:/go/pkg/mod -v gcsfuse-go-build-cache:/root/.cache/go-build -w /work golang:1.26.3 go test ./cfg ./internal/fs/inode ./internal/fs/handle`
- `git diff --check`

Updates #4696

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -605,6 +605,8 @@ type FileSystemConfig struct {
 
 	ExperimentalEnableReaddirplus bool `yaml:"experimental-enable-readdirplus"`
 
+	ExperimentalLargeDirWarningThreshold int64 `yaml:"experimental-large-dir-warning-threshold"`
+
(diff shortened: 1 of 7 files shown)


Haihan-Jiang:
This is ready for review from my side. The title check has a later successful run after the initial failure, and the other visible checks are green. Could someone take a look when convenient?

PR #4746, opened 2026-06-01
fix(retry): add Unauthenticated to retryable gRPC status codes to handle transient token expiration in storage control client

meet2mky:
### Description
This PR adds support for retrying gRPC Unauthenticated errors in the Storage Control Client to handle transient credential/token expirations. It also aligns the documentation and code comments across different retry mechanism.

b/517636259

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - Part of pre-submit

NONE

### DIFF
--- a/internal/storage/control_client_wrapper_test.go
+++ b/internal/storage/control_client_wrapper_test.go
@@ -733,6 +733,29 @@ func (testSuite *ControlClientGaxRetryWrapperTest) TestStorageControlClientGaxRe
 	require.Len(testSuite.T(), gaxOpts, 2)
 }
 
+func (testSuite *ControlClientGaxRetryWrapperTest) TestStorageControlClientGaxRetryOptions_UnauthenticatedIsRetryable() {
+	// Arrange
+	clientConfig := storageutil.GetDefaultStorageClientConfig(keyFile)
+	gaxOpts := storageControlClientGaxRetryOptions(&clientConfig)
+	var settings gax.CallSettings
+	for _, opt := range gaxOpts {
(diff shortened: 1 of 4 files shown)


internal/storage/control_client_wrapper_test.go
```diff
@@ -733,6 +733,29 @@ func (testSuite *ControlClientGaxRetryWrapperTest) TestStorageControlClientGaxRe
 	require.Len(testSuite.T(), gaxOpts, 2)
 }
 
+func (testSuite *ControlClientGaxRetryWrapperTest) TestStorageControlClientGaxRetryOptions_UnauthenticatedIsRetryable() {
```
raj-prince, line 736: nit: Test is doing more than the function name, either only test the Unauthenticated code in the test or create a table driven test for all or few error codes and with the retriable behavior.

raj-prince (dismissed):

internal/storage/control_client_wrapper_test.go
```diff
@@ -733,6 +733,29 @@ func (testSuite *ControlClientGaxRetryWrapperTest) TestStorageControlClientGaxRe
...
+	// Arrange
```
kislaykishore: nit: these labels are not required.

internal/storage/control_client_wrapper_test.go
kislaykishore: +1

meet2mky: Done

kislaykishore (approved):

PR #4747, opened 2026-06-02
perf: set EnableAsyncDIO for kernel reader

geertj:
Set EnableAsyncDIO for the kernel reader.

This enables the same benefits that we get from EnableAsyncReads but for O_DIRECT reads and writes.

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -186,8 +186,10 @@ func getFuseMountConfig(fsName string, newConfig *cfg.Config) *fuse.MountConfig 
 		// Enables ReadDirPlus, allowing the kernel to retrieve directory entries and their
 		// attributes in a single operation.
 		EnableReaddirplus: newConfig.FileSystem.ExperimentalEnableReaddirplus,
-		// Enable async reads if enable-kernel-reader flag is set to true.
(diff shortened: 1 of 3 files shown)


cmd/mount.go
```diff
@@ -185,8 +186,10 @@ func getFuseMountConfig(fsName string, newConfig *cfg.Config) *fuse.MountConfig
 		// Enables ReadDirPlus, allowing the kernel to retrieve directory entries and their
 		// attributes in a single operation.
 		EnableReaddirplus: newConfig.FileSystem.ExperimentalEnableReaddirplus,
-		// Enable async reads if enable-kernel-reader flag is set to true.
+		// Enable async reads if the kernel reader is enabled
 		EnableAsyncReads: newConfig.FileSystem.EnableKernelReader,
+		// Enable async direct IO if the kernel reader is enabled
+		EnableAsyncDIO: newConfig.FileSystem.EnableKernelReader,
```
abhishek10004, line 192: There's one issue I see with this during writes. Because this will cause concurrent requests to come during writes as well, it might force the write path to staging writes (download to disk, then edit and then upload) instead of streaming writes (in case of out of order chunks).

PR #4748, opened 2026-06-03
fix(syncObject): Fix Unfinalized Zonal Object Sync Failures (Staged Writes & Truncate to Stale Size)

meet2mky:
### Problem & Root Cause
When executing streaming writes (and fallback legacy staged writes) on **Zonal/Rapid buckets**, files exist on GCS in an unfinalized state. This unfinalized state causes two major failures in `syncer.SyncObject`:

1. **Incorrect `DirtyThreshold` Sanity Check Violation:**
   GCS metadata stat queries sometimes report the stale size of unfinalized zonal objects as `X` (`srcSize = X`), even though the object contains actual data that GCSFuse has successfully downloaded and staged locally upto Y (Y > X). 
   When local edits are made, the local staged file has `sr.DirtyThreshold > X`. The sanity check:
   ```go
   if sr.DirtyThreshold > srcSize { ... error("stat returned weird DirtyThreshold") }
   ```
   was triggered because `DirtyThreshold > X` was compared against `srcSize = X`, causing application I/O errors and failing the sync.

2. **Silent Drop of Truncate to X:**
   If a user truncated an unfinalized file to `X` bytes:
   * Local file size `sr.Size` was `X` and `sr.DirtyThreshold` was `X`.
   * GCS metadata size `srcSize` was reported as `X`.
   * The early clean-return check:
     ```go
     if sr.Size == srcSize && sr.DirtyThreshold == srcSize { return }
     ```
     incorrectly matched (`X == X && X == X`), causing GCSFuse to return early and silently drop the truncation. Stale data remained on GCS, violating consistency.

---

### Changes
For Zonal Bucket we bypass the dirty sanity check and only rely on mtime for the early return optimisations.

---

b/519352477

### Testing details
1. Manual - Yes
2. Unit tests - Yes
3. Integration tests - part of presubmits

NONE

### DIFF
--- a/internal/gcsx/syncer.go
+++ b/internal/gcsx/syncer.go
@@ -193,9 +193,19 @@ func (os *syncer) SyncObject(
 	}
 
 	// If the content hasn't been dirtied (i.e. it is the same size as the source
-	// object, and no bytes within the source object have been dirtied), we're
-	// done.
-	if sr.Size == srcSize && sr.DirtyThreshold == srcSize {
+	// object, and no bytes within the source object have been dirtied).
+	tempFileIsUnmodifiedAndLocalMatchesGCS := sr.Size == srcSize && sr.DirtyThreshold == srcSize
+
+	// We return early if the file is unmodified.
+	//
+	// We handle this for two different object states:
+	// 1. Finalized objects (Standard buckets): We return early if the file is unmodified.
+	// 2. Unfinalized objects (Zonal/Rapid buckets): Unfinalized objects can report a
+	//    stale metadata size of 0. If the local temp file size is 0 and GCS returns 0,
+	//    we cannot distinguish between a genuinely empty file and a file containing
+	//    data that we want to truncate to 0. Thus, we bypass this optimization when
+	//    the local temp file size is 0 to ensure truncate-to-0 is always uploaded.
+	if (!srcObject.Finalized.IsZero() || sr.Size != 0) && tempFileIsUnmodifiedAndLocalMatchesGCS {
 		return
 	}
 
--- a/internal/gcsx/syncer.go
+++ b/internal/gcsx/syncer.go
@@ -180,7 +180,7 @@ func (os *syncer) SyncObject(
 		return os.fullCreator.Create(ctx, objectName, srcObject, sr.Mtime, os.chunkRetryDeadlineSecs, os.chunkTransferTimeoutSecs, content)
 	}
 
-	// TODO(b/520290147): Remomve unfinalized checks once stat results become consistent on zonal buckets.
+	// TODO(b/520290147): Remove unfinalized checks once stat results become consistent on zonal buckets.
 	// Make sure the dirty threshold makes sense.
 	// Note: This check is not needed for unfinalized objects as they are always uploaded in their entirety.
 	srcSize := int64(srcObject.Size)
@@ -207,7 +207,7 @@ func (os *syncer) SyncObject(
 	// object, and no bytes within the source object have been dirtied).
 	tempFileIsUnmodifiedAndLocalMatchesGCS := sr.Size == srcSize && sr.DirtyThreshold == srcSize
 
-	// TODO(b/520290147): Remomve unfinalized checks once stat results become consistent on zonal buckets.
+	// TODO(b/520290147): Remove unfinalized checks once stat results become consistent on zonal buckets.
 	// This check is specifically for finalized objects (Standard buckets) where GCS metadata
 	// size is fully trustworthy. sr.Mtime == nil alone is not sufficient because a file could
 	// be modified (making sr.Mtime non-nil), but then truncated back to its original size (or
--- a/internal/gcsx/syncer.go
+++ b/internal/gcsx/syncer.go
@@ -193,35 +193,35 @@ func (os *syncer) SyncObject(
 		return
 	}
 
+	// We return early if the file is unmodified or content in unchanged.
+	//
+	// Scenario 1: If the local staged file was never written to or truncated, sr.Mtime
+	// remains nil. This is a 100% reliable indicator that no local changes were ever
+	// made, allowing us to always return early safely for both finalized and unfinalized objects.
+	if sr.Mtime == nil {
+		return
+	}
+
+	// Scenario 2: Size and dirty threshold comparison.
 	// If the content hasn't been dirtied (i.e. it is the same size as the source
 	// object, and no bytes within the source object have been dirtied).
 	tempFileIsUnmodifiedAndLocalMatchesGCS := sr.Size == srcSize && sr.DirtyThreshold == srcSize
 
-	// We return early if the file is unmodified.
+	// TODO(b/520290147): Remomve unfinalized checks once stat results become consistent on zonal buckets.
+	// This check is specifically for finalized objects (Standard buckets) where GCS metadata
+	// size is fully trustworthy. sr.Mtime == nil alone is not sufficient because a file could
+	// be modified (making sr.Mtime non-nil), but then truncated back to its original size (or
+	// have its changes reverted) such that it matches the source GCS object size/content exactly.
+	// For finalized objects, this check allows us to still optimize and return early in
+	// such cases, avoiding a redundant upload.
 	//
-	// We handle this for two different object states:
-	// 1. Finalized objects (Standard buckets): We return early if the file is unmodified.
-	// 2. Unfinalized objects (Zonal/Rapid buckets): Unfinalized objects can report a
-	//    stale metadata size Y that is less than their actual size X. If we locally
-	//    truncate the file to Y, both the local size and dirty threshold become Y,
-	//    which matches the stale metadata size Y. We cannot distinguish this from an
-	//    unmodified file of size Y. Thus, we bypass this optimization for
-	//    unfinalized objects to prevent truncations from being silently skipped.
-	//
-	// We use two checks to determine if we can return early:
-	//
-	// - sr.Mtime == nil: If the file was never written to or truncated, sr.Mtime remains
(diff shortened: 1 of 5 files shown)


internal/gcsx/syncer.go
```diff
@@ -192,9 +193,19 @@ func (os *syncer) SyncObject(
...
+	tempFileIsUnmodifiedAndLocalMatchesGCS := sr.Size == srcSize && sr.DirtyThreshold == srcSize
+
+	// We return early if the file is unmodified.
+	//
+	// We handle this for two different object states:
+	// 1. Finalized objects (Standard buckets): We return early if the file is unmodified.
+	// 2. Unfinalized objects (Zonal/Rapid buckets): Unfinalized objects can report a
+	//    stale metadata size of 0. If the local temp file size is 0 and GCS returns 0,
+	//    we cannot distinguish between a genuinely empty file and a file containing
+	//    data that we want to truncate to 0. Thus, we bypass this optimization when
```
raj-prince: Does it only happen in truncate operation or with modification at 0 also?

internal/gcsx/syncer_test.go
```diff
@@ -359,6 +360,80 @@ func (t *SyncerTest) SyncObjectShouldInvokeFullObjectCreatorWhenSrcObjectIsNil()
...
+
+	o, err := t.call()
+
+	// It should bypass the weird dirty threshold error and succeed.
+	AssertEq(nil, err)
+	ExpectEq(t.fullCreator.o, o)
+	ExpectTrue(t.fullCreator.called)
+}
+
+func (t *SyncerTest) UnfinalizedObjectDoesNotReturnEarlyOnTruncateToZero() {
```
raj-prince, line 390: Can we follow AAA structure here?

meet2mky: Done

internal/gcsx/syncer.go
meet2mky: Only in the truncate to 0 we can't provide the early return optimisation.

In rest scenarios if the edit actually happens even at 0 or mid of the file including truncate then following happens as we must update the object on GCS.

In other case when the file is not updated at all then calling sync would return early as there is nothing to update on the GCS.

In Zonal the full creator is always invoked irrespective of edit type (truncate downwards, edit within the file, appends to the file)

---
### Scenario 1: Overwriting a single byte on an unfinalized zonal object (actual size 10, GCS reports 0) (Single byte can be anywhere on the file).

1. **Initial Setup:**
   * **GCS State:** The object is unfinalized. It contains 10 bytes of actual data, but GCS metadata reports `srcSize = 0`.
   * **Staging:** GCSFuse reads the unfinalized object, downloads the 10 bytes, and populates the local temp file (`sr.Size = 10`, `sr.DirtyThreshold = 10`).

2. **The Modification:**
   * The application writes a single byte at offset 0 (overwriting a bit).
   * GCSFuse updates the dirty threshold: `dirtyThreshold = min(10, 0) = 0`.
   * Staged temp file state: `sr.Size = 10`, `sr.DirtyThreshold = 0`.

3. **Step-by-Step Check Execution in `SyncObject`:**
   * **DirtyThreshold Sanity Check:** Bypassed since the object is unfinalized (`!srcObject.Finalized.IsZero()` is `false`).
   * **Early Clean-Return Check:** `sr.Size == srcSize` (10 == 0) is `false`, so `tempFileIsUnmodifiedAndLocalMatchesGCS` is `false`. The early return check is bypassed.
   * **Determining Creator:** `sr.DirtyThreshold == srcSize` (0 == 0) is `false`. The compose optimization is bypassed.

4. **Upload Execution:**
   * GCSFuse enters the `else` block and invokes the **Full Creator** (`os.fullCreator.Create(...)`), uploading the entire local 10-byte temp file from offset 0, successfully overwriting and updating GCS with the modified contents.


---
### Scenario 2: Truncating an unfinalized zonal object to 0 bytes (actual size 10, GCS reports 0).

1. **Initial Setup:**
   * **GCS State:** The object is unfinalized. It contains 10 bytes of actual data, but GCS metadata reports `srcSize = 0`.
   * **Staging:** GCSFuse reads the unfinalized object, downloads the 10 bytes, and populates the local temp file (`sr.Size = 10`, `sr.DirtyThreshold = 10`).

2. **The Modification:**
   * The application calls `Truncate(0)`.
   * GCSFuse updates the stats: `sr.Size = 0`, `sr.DirtyThreshold = min(10, 0) = 0`.
   * Staged temp file state: `sr.Size = 0`, `sr.DirtyThreshold = 0`.

3. **Step-by-Step Check Execution in `SyncObject`:**
   * **DirtyThreshold Sanity Check:** Bypassed since the object is unfinalized (`!srcObject.Finalized.IsZero()` is `false`).
   * **Early Clean-Return Check:**
     * `tempFileIsUnmodifiedAndLocalMatchesGCS` evaluates to `0 == 0 && 0 == 0` $\implies$ `true`.
     * The condition `(!srcObject.Finalized.IsZero() || sr.Size != 0)` evaluates to `(false || false)` $\implies$ `false`.
     * The unified check `(false) && true` evaluates to `false`.
     * **Result:** GCSFuse bypasses the early return optimization, preventing the truncate-to-zero operation from being silently skipped.

4. **Upload Execution:**
   * GCSFuse enters the `else` block and invokes the **Full Creator** (`os.fullCreator.Create(...)`), uploading the 0-byte file, successfully truncating the object on GCS.

raj-prince (dismissed):

internal/gcsx/syncer.go
```diff
@@ -181,8 +181,9 @@ func (os *syncer) SyncObject(
 	}
 
 	// Make sure the dirty threshold makes sense.
+	// Note: This check is not needed for unfinalized objects as they are always uploaded in their entirety.
 	srcSize := int64(srcObject.Size)
-	if sr.DirtyThreshold > srcSize {
+	if !srcObject.Finalized.IsZero() && sr.DirtyThreshold > srcSize {
```
abhishek10004: nit: You can use IsUnfinalized() for better readability. Same at other places.

internal/gcsx/syncer.go
```diff
@@ -192,9 +193,19 @@ func (os *syncer) SyncObject(
...
+	// We return early if the file is unmodified.
+	//
+	// We handle this for two different object states:
+	// 1. Finalized objects (Standard buckets): We return early if the file is unmodified.
+	// 2. Unfinalized objects (Zonal/Rapid buckets): Unfinalized objects can report a
+	//    stale metadata size of 0. If the local temp file size is 0 and GCS returns 0,
+	//    we cannot distinguish between a genuinely empty file and a file containing
+	//    data that we want to truncate to 0. Thus, we bypass this optimization when
+	//    the local temp file size is 0 to ensure truncate-to-0 is always uploaded.
+	if (!srcObject.Finalized.IsZero() || sr.Size != 0) && tempFileIsUnmodifiedAndLocalMatchesGCS {
```
abhishek10004: As discussed offline, we cannot always return early when sr.size !=0.
There is a possibility that when local file is truncated to stale size returned by GCS, we might return early which is not correct.

Also, put a to do to revert this change once we have consistent stat call for Rapid.

internal/gcsx/syncer.go
meet2mky: Done

internal/gcsx/syncer.go
meet2mky: Added a TODO. Based on offline discussion on edge cases current implementation covers everything. Thanks!

internal/gcsx/syncer.go
```diff
@@ -192,16 +194,24 @@ func (os *syncer) SyncObject(
...
+	//    stale metadata size Y that is less than their actual size X. If we locally
+	//    truncate the file to Y, both the local size and dirty threshold become Y,
+	//    which matches the stale metadata size Y. We cannot distinguish this from an
+	//    unmodified file of size Y. Thus, we bypass this optimization for
+	//    unfinalized objects to prevent truncations from being silently skipped.
+	//
+	// In both cases, if the file was never written to or truncated, sr.Mtime remains
+	// nil. This is a 100% reliable indicator that no local changes were ever made,
+	// allowing us to always return early safely.
+	if sr.Mtime == nil || (!srcObject.IsUnfinalized() && tempFileIsUnmodifiedAndLocalMatchesGCS) {
```
raj-prince: nit: can we add comment for other check?

meet2mky: Done

internal/gcsx/syncer.go
```diff
@@ -192,16 +194,34 @@ func (os *syncer) SyncObject(
...
-	// done.
-	if sr.Size == srcSize && sr.DirtyThreshold == srcSize {
-		return
-	}
+	// object, and no bytes within the source object have been dirtied).
+	tempFileIsUnmodifiedAndLocalMatchesGCS := sr.Size == srcSize && sr.DirtyThreshold == srcSize
 
-	// Sanity check: the branch above should ensure that by the time we get here,
-	// the stat result's mtime is non-nil.
-	if sr.Mtime == nil {
```
vadlakondaswetha, line 203: can't we just rely on the mtime and remove the other check?

if you intend to keep both the checks, split the condition into multiple if clauses

if sr.mtime == nil {
return
}

if finalized && (sr.Size == srcSize && sr.DirtyThreshold == srcSize) {
return
}

meet2mky: There is a use case to keep the  other check as well along with the sr.mtime check. I will split the check in two parts.
CC @raj-prince

meet2mky: Done, Have split the check.

raj-prince (dismissed):

raj-prince (approved):

vadlakondaswetha (approved):

PR #4763, opened 2026-06-06
test(integration): migrate TPC operations tests to config file

PranjalC100:
### Description
Migrates the `operations` integration tests on the TPC endpoint to execute from the common configuration file (`test_config.yaml`) rather than relying on hardcoded test runner logic.

**Key Changes:**
* Adds a new `TPC` boolean property to `ConfigItem` in `config.go` to identify TPC-specific configurations.
* Declares standard `flat`/`hns`/`zonal` compatibility for TPC configurations in `test_config.yaml`, combined with `tpc: true`.
* Modifies `BuildFlagSets()` in `setup.go` to correctly filter configurations by matching the active endpoint with the config (`tpcMatches := (TestOnTPCEndPoint() == testCase.TPC)`).
* Configures `setup.go` to use specific TPC credentials (`/tmp/sa.key.json`), endpoint (`storage.apis-tpczero.goog:443`), and universe domain (`apis-tpczero.goog`) when `TestOnTPCEndPoint()` is true.
* Conditionally initializes the storage client using standard `storage.NewClient` for TPC endpoints, while retaining `storage.NewGRPCClient` for others.
* Deletes the custom/hardcoded `RunTestOnTPCEndPoint` flow in `operations_test.go` and transitions it to the standard config setup.
* Cleans up redundant local helpers in `operations_test.go` in favor of using package-level ones instead.

Fixes b/469970353

### Testing details
1. Manual - Checked local compilation using `go test -c ./tools/integration_tests/operations` and checked general repo lint/compilation using `make buildTest`.
2. Unit tests - N/A
3. Integration tests - N/A (requires TPC endpoint and credentials)

N/A

### DIFF
--- a/tools/integration_tests/util/setup/setup.go
+++ b/tools/integration_tests/util/setup/setup.go
@@ -607,14 +607,26 @@ func BucketType(ctx context.Context, testBucket string) (bucketType string, err 
 	defer cancel()
 	var opts []option.ClientOption
 	opts = append(opts, experimental.WithGRPCBidiReads())
-	if keyFile != "" {
+	if TestOnTPCEndPoint() {
+		cred, err := auth2.GetCredentials("/tmp/sa.key.json")
+		if err != nil {
+			return "", fmt.Errorf("failed to get credentials for TPC: %w", err)
+		}
+		opts = append(opts, option.WithEndpoint("storage.apis-tpczero.goog:443"), option.WithAuthCredentials(cred))
+	} else if keyFile != "" {
 		cred, err := auth2.GetCredentials(keyFile)
 		if err != nil {
 			return "", fmt.Errorf("failed to get credentials: %w", err)
 		}
 		opts = append(opts, option.WithAuthCredentials(cred))
 	}
-	storageClient, err := storage.NewGRPCClient(ctx, opts...)
+	var storageClient *storage.Client
+	if TestOnTPCEndPoint() {
+		storageClient, err = storage.NewClient(ctx, opts...)
+	} else {
+		storageClient, err = storage.NewGRPCClient(ctx, opts...)
+	}
+
 	if err != nil {
 		return "", fmt.Errorf("failed to create storage client: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


meet2mky (changes requested):
TPC Build is still failing. Please see the kokoro failure.

anushka567:
@PranjalC100 Can you please address the requested changes by Mohit? Thanks!

tools/integration_tests/util/setup/setup.go
meet2mky, line 603: Can this method use the exported functions from this utility?

tools/integration_tests/util/client

It provides both storage client creation and storage control client creation.

tools/integration_tests/util/setup/setup.go
```diff
@@ -676,8 +688,8 @@ func BuildFlagSets(cfg test_suite.TestConfig, bucketType string, run string) [][
 				isCompatible = false
 			}
 		}
-
-		if isCompatible && (run == "" || run == testCase.Run) {
+		tpcMatches := (TestOnTPCEndPoint() == testCase.TPC)
```
meet2mky: Nit: tpcRun instead of tpcMatches

PranjalC100: done.

tools/integration_tests/util/setup/setup.go
PranjalC100: We actually cannot use the common client utility (`tools/integration_tests/util/client`) here inside `setup.go` because `util/client` already imports `util/setup` (for helpers like `setup.TestOnTPCEndPoint()`, `setup.KeyFile()`, `setup.BillingProject()`, etc.). 

Attempting to import `util/client` into `setup.go` causes a fatal Go compiler error due to an import cycle (`import cycle not allowed: util/setup -> util/client -> util/setup`). Therefore, `setup.go` needs to instantiate its own storage client directly.

Refactoring this would involve changes across all the test packages.

meet2mky: Oh, That's a bummer :(

meet2mky (approved):
LGTM

PR #4778, opened 2026-06-15
fix(patch): benchmarking and unsupported_path test fixes to v3.8.2

PranjalC100:
### Description
This PR backports two integration test fixes to the `v3.8.2_release` branch for GKE.

1. **Benchmarking Fix (PR #4577):** Prevents `setup.MntDir` from being overwritten by an empty directory during GKE execution.
2. **Unsupported Path Fix (PR #4623):** Resolves test failures in `unsupported_path` caused by path joining logic.

### Linked Issues
- **https://buganizer.corp.google.com/issues/496956438#comment5**

### DIFF
--- a/tools/integration_tests/test_config.yaml
+++ b/tools/integration_tests/test_config.yaml
@@ -69,6 +69,7 @@ list_large_dir:
 operations:
   - mounted_directory: "${MOUNTED_DIR}"
     test_bucket: "${BUCKET_NAME}"
+    only_dir: "${ONLY_DIR}"
     configs:
       - flags:
         - "--metadata-cache-ttl-secs=0,--enable-streaming-writes=false"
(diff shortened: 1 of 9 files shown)


tools/integration_tests/test_config.yaml
```diff
@@ -786,25 +787,23 @@ flag_optimizations:
         run_on_gke: false
 
 unsupported_path:
-- mounted_directory: "${MOUNTED_DIR}"
-  test_bucket: "${BUCKET_NAME}"
-  only_dir: "${ONLY_DIR}"
```
anushka567, line 791: Why did we remove the only_dir config for unsupported_path test package?

anushka567 (approved):

tools/integration_tests/test_config.yaml
PranjalC100: Because we were not running **only_dir** tests for `unsupported_path` test package. Incase we add **only_dir** in` test_config.yaml`, even if we ignore it, GKE will start running it with **only_dir** as well.

PR #4781, opened 2026-06-16
fix: resolve CPU busy-loop in GenBlockPool.Get on exhaustion

kislaykishore:
### Problem
When write operations take longer to complete, GCSFuse cannot release memory blocks back to the pool quickly and the block pool remains exhausted for longer periods.

During these periods, write threads calling `GenBlockPool.Get()` spend their time spinning in a CPU busy-loop (the `default` branch in the `select` loop), leading to near-permanent CPU starvation of the GCSFuse 

### Solution 
We removed the busy-loop. When the pool is exhausted and no new blocks can be allocated, the write thread blocks cleanly (using Go's sleeping channel/mutex mechanisms, consuming **0% CPU**). 

#### Before fix
In an Orbax checkpoint writes workload, the GetBlockPool was taking ~50% of CPU for the `WriteFile operation`
image

#### After fix
The entire call disappears from the profile. The common `appendBuffer` call that was taking 53% of the CPU earlier now takes only 5%
image

b/524416466

### Testing details
Manual - Tested manually by adding logs.
Unit tests - Added
Integration tests - NA

No

### DIFF
--- a/internal/block/block_pool.go
+++ b/internal/block/block_pool.go
@@ -95,6 +95,10 @@ func NewGenBlockPool[T GenBlock](blockSize int64, maxBlocks int64, reservedBlock
 	}, nil
 }
 
+// Get returns a block. It returns an existing block if it's ready for reuse or
+// creates a new one if required.
+// Not thread-safe, calling from multiple goroutines may lead to memory leaks because
+// of race conditions.
 func (bp *GenBlockPool[T]) Get() (T, error) {
 	// Try to get a block immediately (non-blocking).
 	b, err := bp.TryGet()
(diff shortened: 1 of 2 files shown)


internal/block/block_pool.go
```diff
@@ -95,27 +97,97 @@ func NewGenBlockPool[T GenBlock](blockSize int64, maxBlocks int64, reservedBlock
...
+	return bp.allocateNewBlock(true)
+}
+
+const (
+	stateUndecided int32 = iota
+	stateLocalWon
+	stateGlobalWon
+)
+
+func (bp *GenBlockPool[T]) waitAndGetConcurrent() (T, error) {
```
raj-prince, line 139: Can't we simplify the implementation, do we need to keep the atomic state? Something like below:

```go
func (bp *GenBlockPool[T]) waitAndGetConcurrent() (T, error) {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	acquiredCh := make(chan struct{}) // Unbuffered channel

	go func() {
		err := bp.globalMaxBlocksSem.Acquire(ctx, 1)
		if err == nil {
			select {
			case acquiredCh <- struct{}{}:
				// Main thread successfully read the permit signal!
			case <-ctx.Done():
				// Main thread took the local block and returned.
				// We must release the acquired global permit!
				bp.globalMaxBlocksSem.Release(1)
			}
		}
	}()

	select {
	case b := <-bp.freeBlocksCh:
		b.Reuse()
		return b, nil

	case <-acquiredCh:
		// Non-blocking check: can we reuse a local block instead of allocating?
		select {
		case b := <-bp.freeBlocksCh:
			bp.globalMaxBlocksSem.Release(1)
			b.Reuse()
			return b, nil
		default:
			return bp.allocateNewBlock(true)
		}
	}
}
```

raj-prince (commented):
Looks good, added a comment to simplify the waitAndGetConcurrent() implementation if we can.

internal/block/block_pool.go
kislaykishore: Per Jetski:

Consider the case where a local block becomes free at the exact same moment that the helper goroutine successfully acquires the global semaphore permit:

* Helper Goroutine succeeds: Acquire(ctx, 1) completes successfully. A global permit is now acquired.

* Helper Goroutine enters its select:
  - Since the main thread is still waiting in its select block, a reader is available on acquiredCh.
  - ctx.Done() is not closed yet.
  - The helper goroutine's select chooses the first case because it's ready: case acquiredCh <- struct{}{}:.

Crucial Go behavior: Because acquiredCh is an unbuffered channel, sending to it blocks the helper goroutine until the main thread actually receives the value. The helper goroutine is now blocked mid-send.

* Main Goroutine wakes up:
  - In the main thread's select, both cases are now ready: case b := <-bp.freeBlocksCh and case <-acquiredCh.
  - Go's select statement selects a case pseudo-randomly when multiple cases are ready.
  - Go chooses case b := <-bp.freeBlocksCh.

* Main Goroutine exits:
  - The main thread reuses the local block, returns it, and exits waitAndGetConcurrent().
  - As it exits, defer cancel() runs, which closes ctx.Done().

The Leak:
* The helper goroutine was already committed to the case acquiredCh <- struct{}{}: branch and was blocked waiting for a receiver.
* Since the main thread has exited, there will never be another receiver on acquiredCh.
* The helper goroutine remains blocked forever trying to send to acquiredCh. It can never reach bp.globalMaxBlocksSem.Release(1).
This results in a permanent goroutine leak and a permanent semaphore permit leak.

kislaykishore: I'm not fully convinced though. I've incorporated your feedback. Thanks!

raj-prince (approved):

PR #4784, opened 2026-06-16
feat(metrics): Adding the metadata_cache read count metric

thrivikram-karur-g:
### Description
Adding a metric to capture the metadata cache reads.
Hits & misses, more detail on the cache lookup and if cached entry is positive or negative.

b/512002923

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

N/A

### DIFF
--- a/internal/cache/metadata/stat_cache.go
+++ b/internal/cache/metadata/stat_cache.go
@@ -15,7 +15,10 @@
 package metadata
 
 import (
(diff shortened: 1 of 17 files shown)


thrivikram-karur-g:
Closing this PR as we are going with a different implementation approach with accurate metric counts

PR #4788, opened 2026-06-17
test(fix flaky test): fixing flaky TestZonalFileCacheReaderTestSuite

raj-prince:
### Description
- **Problem:** The TestZonalFileCacheReaderTestSuite/Test_ReadAt_SequentialRangeRead test was flaky because the asynchronous background downloader goroutines (downloader.Job) were being leaked after each test completed. When TearDownTest() executed os.RemoveAll(t.cacheDir), it raced against those active background jobs trying to write to that same directory, causing device or resource busy or mock assertion failures.
- **Solution:** Updated TearDownTest in file_cache_reader_test.go
 to call t.jobManager.Destroy() before removing the directory. This guarantees that all background downloader jobs are instantly cancelled and stopped, releasing their file descriptors and terminating their goroutines, making the directory cleanup completely safe and race-free.

b/524993281

### Testing details
1. Manual - yes, by running 30 times.
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/gcsx/file_cache_reader_test.go
+++ b/internal/gcsx/file_cache_reader_test.go
@@ -110,9 +110,15 @@ func (t *fileCacheReaderTest) SetupTest() {
 }
 
 func (t *fileCacheReaderTest) TearDownTest() {
-	t.reader.Destroy()
-	t.reader_unfinalized_object.Destroy()
-	t.jobManager.Destroy()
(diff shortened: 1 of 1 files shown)


meet2mky (approved):
LGTM

kislaykishore (approved):

PR #4801, opened 2026-06-24
test(e2e): making client-protocol explicit to make agnostic of default value

raj-prince:
### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/tools/integration_tests/test_config.yaml
+++ b/tools/integration_tests/test_config.yaml
@@ -17,7 +17,7 @@ explicit_dir:
     test_bucket: "${BUCKET_NAME}" # To be passed by both gcsfuse and gke tests
     configs:
       - flags:
-          - "--implicit-dirs=false"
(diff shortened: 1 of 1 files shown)


tools/integration_tests/test_config.yaml
meet2mky, line 1: Hi @raj-prince , Can you check with @PranjalC100 if we have to use comma separate mount options or space separated mount option?

I think both should be working as of now but let's keep the file consistent on whether we use the comma or space. 

Can be done in separate PR as well. What do you think @PranjalC100?

PranjalC100: I've added another commit in the same PR, to keep the flags comma separated.

meet2mky (approved):
LGTM

PranjalC100 (approved):

PR #4802, opened 2026-06-25
feat(storage): introduce IsDirectory flag to bypass appendable API

vipnydav:
### Description
This PR introduces a new `IsDirectory` flag in `gcs.CreateObjectRequest` to explicitly distinguish 0-byte directory placeholders from regular files at the storage layer. 

Previously, `CreateObject` and `CreateObjectChunkWriter` would default to using the appendable upload API for *all* objects when rapid writes were enabled. While Hierarchical Namespace (HNS) and Zonal buckets bypass this path entirely by using the `CreateFolder` API, **Flat Pirlo buckets** still rely on `CreateObject` to create 0-byte directory placeholders. 

Because we do not want to use the appendable API to write these simple 0-byte directory markers to the rapid storage class, setting `IsDirectory: true` explicitly bypasses the appendable API (`wc.Append = false`). 

b/527767043

### Testing details
1. **Manual** - NA
2. **Unit tests** - Done
3. **Integration tests** - Automated

N/A

### DIFF
--- a/internal/canned/canned.go
+++ b/internal/canned/canned.go
@@ -71,8 +71,9 @@ func MakeFakeBucket(ctx context.Context) (b gcs.Bucket) {
 		_, err := b.CreateObject(
 			ctx,
 			&gcs.CreateObjectRequest{
-				Name:     k,
-				Contents: strings.NewReader(v),
+				Name:        k,
+				Contents:    strings.NewReader(v),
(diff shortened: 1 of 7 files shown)


vipnydav:
Closing as anyways the zero byte objects would be uptiered to rapid class, so no benefit in creating in standard class.

PR #4803, opened 2026-06-26
test: add memory and performance benchmarks for MapLRU and StatCache

vedantdas-source:
### Description
Adding memory profiling tests and CPU benchmarks for the GCSfuse caching layer. 

A baseline metric of the current architecture's footprint is required to evaluate performance for any further cache implementations. These tests measure the performance of the standard `MapLRU` and the `StatCache` wrapper.

**Changes Included:**
* **Memory Profiling (`TestMapCacheWorkloads` & `TestStatCacheMemoryWorkloads`)**: Simulates a 1,000,000-file workload across three different string geometries (`flat`, `nested`, and `deeply_nested`) to accurately measure the heap cost of the structural overhead and string pointers.
* **CPU Benchmarks (`BenchmarkLRU_...`)**: Adds targeted performance benchmarks to measure nanoseconds-per-operation for `Insert`, `LookUp`, and `ErasePrefix` across the different workloads.

[b/512335162](https://b.corp.google.com/issues/512335162)

### Testing details
1. Manual - NA
2. Unit tests - Added standalone benchmark files `lru_memory_test.go`, `lru_workload_benchmark_test.go`, and `statcache_memory_test.go` to be executed explicitly via `go test` and `go test -bench`.
3. Integration tests - NA

NA

### DIFF
--- a/internal/cache/lru/lru_memory_test.go
+++ b/internal/cache/lru/lru_memory_test.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -81,9 +81,9 @@ func TestMapCacheWorkloads(t *testing.T) {
 		t.Logf("=== WORKLOAD: %s (%d files) ===", w, count)
 
 		// 1. Pure MapLRU
+		paths := generatePaths(w, count)
 		baseMem := getMemStats()
 
-		paths := generatePaths(w, count)
 		pureCache := lru.NewCache(capacity)
 		for _, p := range paths {
 			pureCache.Insert(p, dummyValue{})
--- a/internal/cache/lru/lru_memory_test.go
+++ b/internal/cache/lru/lru_memory_test.go
@@ -42,20 +42,20 @@ func generatePaths(workload string, count int) []string {
 			paths = append(paths, fmt.Sprintf("file_%d.txt", i))
 		}
 	case "nested":
-		// e.g. 1000 dirs, 1000 files each
-		for d := 0; d < 1000; d++ {
-			for f := 0; f < 1000; f++ {
-				paths = append(paths, fmt.Sprintf("dir_%04d/file_%04d.txt", d, f))
-			}
+		filesPerDir := 1000
+		for i := 0; i < count; i++ {
+			d := i / filesPerDir
+			f := i % filesPerDir
+			paths = append(paths, fmt.Sprintf("dir_%04d/file_%04d.txt", d, f))
 		}
 	case "deeply_nested":
-		// e.g. 50 classes, 100 batches, 200 images = 1,000,000 keys
-		for c := 0; c < 50; c++ {
-			for b := 0; b < 100; b++ {
-				for i := 0; i < 200; i++ {
-					paths = append(paths, fmt.Sprintf("projects/ai-vision/training/class_%04d/batch_%04d/img_%06d.jpg", c, b, i))
-				}
-			}
+		batches := 100
+		images := 200
+		for i := 0; i < count; i++ {
+			c := i / (batches * images)
+			b := (i / images) % batches
+			img := i % images
+			paths = append(paths, fmt.Sprintf("projects/ai-vision/training/class_%04d/batch_%04d/img_%06d.jpg", c, b, img))
 		}
 	}
 
@@ -86,7 +86,7 @@ func TestMapCacheWorkloads(t *testing.T) {
 
 		pureCache := lru.NewCache(capacity)
 		for _, p := range paths {
-			pureCache.Insert(p, dummyValue{})
+			_, _ = pureCache.Insert(p, dummyValue{})
 		}
 		alloc := getMemStats()
 		pureMem := alloc - baseMem
--- a/internal/cache/lru/lru_memory_test.go
+++ b/internal/cache/lru/lru_memory_test.go
@@ -38,12 +38,12 @@ func generatePaths(workload string, count int) []string {
 
 	switch workload {
 	case "flat":
-		for i := 0; i < count; i++ {
+		for i := range count {
 			paths = append(paths, fmt.Sprintf("file_%d.txt", i))
 		}
 	case "nested":
 		filesPerDir := 1000
-		for i := 0; i < count; i++ {
+		for i := range count {
 			d := i / filesPerDir
 			f := i % filesPerDir
 			paths = append(paths, fmt.Sprintf("dir_%04d/file_%04d.txt", d, f))
@@ -51,7 +51,7 @@ func generatePaths(workload string, count int) []string {
 	case "deeply_nested":
 		batches := 100
 		images := 200
-		for i := 0; i < count; i++ {
+		for i := range count {
 			c := i / (batches * images)
 			b := (i / images) % batches
 			img := i % images
@@ -59,6 +59,8 @@ func generatePaths(workload string, count int) []string {
 		}
 	}
 
+	//shuffle the paths to simulate a realistic, unpredictable workload
+	//inserting keys in perfectly sorted lexicographical order can benefit or penalize certain tree/map data structures
 	rand.Shuffle(len(paths), func(i, j int) {
 		paths[i], paths[j] = paths[j], paths[i]
 	})
--- a/internal/cache/lru/lru_memory_test.go
+++ b/internal/cache/lru/lru_memory_test.go
@@ -93,9 +93,8 @@ func TestMapCacheWorkloads(t *testing.T) {
 		alloc := getMemStats()
 		pureMem := alloc - baseMem
 		runtime.KeepAlive(pureCache)
-		pureCache = nil
+		runtime.KeepAlive(paths)
 
-		t.Logf("%-20s Heap Used: %10.2f MB", "Pure MapLRU", float64(pureMem)/(1024*1024))
-		t.Logf("")
+		t.Logf("%-20s Heap Used: %10.2f MB\n", "MapLRU", float64(pureMem)/(1024*1024))
 	}
 }
--- a/internal/cache/lru/lru_memory_test.go
+++ b/internal/cache/lru/lru_memory_test.go
@@ -1,0 +1,99 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lru_test
+
+import (
+	"fmt"
+	"math/rand"
+	"runtime"
+	"runtime/debug"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/cache/lru"
+)
+
+func getMemStats() uint64 {
+	runtime.GC()
+	runtime.GC()
+	debug.FreeOSMemory()
+	var m runtime.MemStats
+	runtime.ReadMemStats(&m)
+	return m.Alloc
+}
+
+func generatePaths(workload string, count int) []string {
+	var paths []string
+
+	switch workload {
+	case "flat":
+		for i := 0; i < count; i++ {
+			paths = append(paths, fmt.Sprintf("file_%d.txt", i))
+		}
+	case "nested":
+		// e.g. 1000 dirs, 1000 files each
+		for d := 0; d < 1000; d++ {
+			for f := 0; f < 1000; f++ {
+				paths = append(paths, fmt.Sprintf("dir_%04d/file_%04d.txt", d, f))
+			}
+		}
+	case "deeply_nested":
+		// e.g. 50 classes, 100 batches, 200 images = 1,000,000 keys
+		for c := 0; c < 50; c++ {
+			for b := 0; b < 100; b++ {
+				for i := 0; i < 200; i++ {
+					paths = append(paths, fmt.Sprintf("projects/ai-vision/training/class_%04d/batch_%04d/img_%06d.jpg", c, b, i))
+				}
+			}
+		}
+	}
+
+	rand.Shuffle(len(paths), func(i, j int) {
+		paths[i], paths[j] = paths[j], paths[i]
+	})
+
+	return paths
+}
+
+type dummyValue struct{}
+
+func (d dummyValue) Size() uint64 {
+	return 10
+}
+
+func TestMapCacheWorkloads(t *testing.T) {
+	count := 1000000 // 1 Million files
+	capacity := uint64(count * 1000)
+	workloads := []string{"flat", "nested", "deeply_nested"}
+
+	for _, w := range workloads {
+		t.Logf("=== WORKLOAD: %s (%d files) ===", w, count)
+
+		// 1. Pure MapLRU
+		baseMem := getMemStats()
+
+		paths := generatePaths(w, count)
+		pureCache := lru.NewCache(capacity)
+		for _, p := range paths {
+			pureCache.Insert(p, dummyValue{})
+		}
+		alloc := getMemStats()
+		pureMem := alloc - baseMem
+		runtime.KeepAlive(pureCache)
+		pureCache = nil
+
+		t.Logf("%-20s Heap Used: %10.2f MB", "Pure MapLRU", float64(pureMem)/(1024*1024))
+		t.Logf("")
+	}
+}
--- a/internal/cache/lru/lru_workload_benchmark_test.go
+++ b/internal/cache/lru/lru_workload_benchmark_test.go
@@ -96,7 +96,7 @@ func runBenchmarks(b *testing.B, name string, depth int) {
 	itemsPerPrefix := 1000
 	keys := generateKeys(prefixCount, itemsPerPrefix, depth)
 
-	// pre-calculate which keys belong to which prefix so we can
+	// Pre-calculate which keys belong to which prefix so we can
 	// quickly repair the cache in benchmarkErasePrefix
 	var prefixes []string
 	prefixMap := make(map[string][]string)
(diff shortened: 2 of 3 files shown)


internal/cache/lru/lru_memory_test.go
```diff
@@ -0,0 +1,99 @@
...
+	runtime.ReadMemStats(&m)
+	return m.Alloc
+}
+
+func generatePaths(workload string, count int) []string {
+	var paths []string
+
+	switch workload {
+	case "flat":
+		for i := 0; i < count; i++ {
```
kislaykishore: Modernize - Please make this change here and elsewhere
```suggestion
		for i := range count {
```

internal/cache/lru/lru_memory_test.go
```diff
@@ -0,0 +1,99 @@
...
+		for c := 0; c < 50; c++ {
+			for b := 0; b < 100; b++ {
+				for i := 0; i < 200; i++ {
+					paths = append(paths, fmt.Sprintf("projects/ai-vision/training/class_%04d/batch_%04d/img_%06d.jpg", c, b, i))
+				}
+			}
+		}
+	}
+
+	rand.Shuffle(len(paths), func(i, j int) {
```
kislaykishore, line 64: Please add a comment to explain why we shuffling.

internal/cache/lru/lru_memory_test.go
```diff
@@ -0,0 +1,99 @@
...
+	case "deeply_nested":
+		// e.g. 50 classes, 100 batches, 200 images = 1,000,000 keys
+		for c := 0; c < 50; c++ {
+			for b := 0; b < 100; b++ {
+				for i := 0; i < 200; i++ {
+					paths = append(paths, fmt.Sprintf("projects/ai-vision/training/class_%04d/batch_%04d/img_%06d.jpg", c, b, i))
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
vedantdas-source: Changed the scripts to dynamically generate paths based on count

internal/cache/metadata/statcache_memory_test.go
```diff
@@ -0,0 +1,87 @@
...
+	case "deeply_nested":
+		// e.g. 50 classes, 100 batches, 200 images = 1,000,000 keys
+		for c := 0; c < 50; c++ {
+			for b := 0; b < 100; b++ {
+				for i := 0; i < 200; i++ {
+					paths = append(paths, fmt.Sprintf("projects/ai-vision/training/class_%04d/batch_%04d/img_%06d.jpg", c, b, i))
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
vedantdas-source: changed script to dynamically generate paths based on count

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"fmt"
+	"math/rand"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/cache/lru"
+)
+
+func generateKeys(prefixCount, itemsPerPrefix, depth int) []string {
+	var keys []string
+	for p := 0; p < prefixCount; p++ {
```
kislaykishore: Modernize here and elsewhere
```suggestion
	for p := range prefixCount {
```

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,147 @@
...
+	rand.Shuffle(len(keys), func(i, j int) {
+		keys[i], keys[j] = keys[j], keys[i]
+	})
+	return keys
+}
+
+func benchmarkInsert(b *testing.B, cache *lru.Cache, keys []string) {
+	data := testData{Value: 1, DataSize: 10}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
kislaykishore:
```suggestion
	for b.Loop() {
```

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+func benchmarkLookup(b *testing.B, cache *lru.Cache, keys []string) {
+	data := testData{Value: 1, DataSize: 10}
+	for _, key := range keys {
+		_, _ = cache.Insert(key, data)
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
kislaykishore:
```suggestion
	for b.Loop() {
```

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,147 @@
...
+	data := testData{Value: 1, DataSize: 10}
+
+	for _, keysInPrefix := range prefixMap {
+		for _, key := range keysInPrefix {
+			_, _ = cache.Insert(key, data)
+		}
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
kislaykishore:
```suggestion
	for b.Loop() {
```

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,147 @@
...
+	prefixCount := 1000
+	itemsPerPrefix := 1000
+	keys := generateKeys(prefixCount, itemsPerPrefix, depth)
+
+	// Pre-calculate which keys belong to which prefix so we can
+	// quickly repair the cache in benchmarkErasePrefix
+	var prefixes []string
+	prefixMap := make(map[string][]string)
+
+	for p := 0; p < prefixCount; p++ {
```
kislaykishore:
```suggestion
	for p := range prefixCount {
```

internal/cache/metadata/statcache_memory_test.go
```diff
@@ -0,0 +1,99 @@
...
+		sharedCache := lru.NewCache(capacity)
+		statCache := metadata.NewStatCacheBucketView(sharedCache, "")
+
+		for _, p := range paths {
+			statCache.Insert(&gcs.MinObject{Name: p}, expiration)
+		}
+
+		alloc := getMemStats()
+		pureMem := alloc - baseMem
+		runtime.KeepAlive(statCache)
```
kislaykishore, line 95: Should we keep the "paths" alive as well?

internal/cache/lru/lru_memory_test.go
```diff
@@ -0,0 +1,99 @@
...
+		paths := generatePaths(w, count)
+		baseMem := getMemStats()
+
+		pureCache := lru.NewCache(capacity)
+		for _, p := range paths {
+			_, _ = pureCache.Insert(p, dummyValue{})
+		}
+		alloc := getMemStats()
+		pureMem := alloc - baseMem
+		runtime.KeepAlive(pureCache)
```
kislaykishore, line 95: Same comment here: should we keep paths alive as well?

internal/cache/metadata/statcache_memory_test.go
vedantdas-source: As the string headers are stored inside the cache itself, the GC wont reclaim them. So keeping the cache alive until memory measurements are done would be fine.

internal/cache/lru/lru_memory_test.go
vedantdas-source: As the string headers are stored inside the cache itself, the GC wont reclaim them. So keeping the cache alive until memory measurements are done would be fine.

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Done. As we cycle through the keys we still have to keep a separate index variable though

internal/cache/lru/lru_memory_test.go
vedantdas-source: Done

internal/cache/lru/lru_memory_test.go
vedantdas-source: Done

internal/cache/lru/lru_memory_test.go
kislaykishore: That's dependent on the implementation. When we move to the Radix tree implementation, it'd matter and we'll use the same test there as well.

internal/cache/lru/lru_memory_test.go
```diff
@@ -0,0 +1,102 @@
...
+		runtime.KeepAlive(paths)
+		pureCache = nil
```
kislaykishore: What's the point of setting pureCache = nil here? Won't it automatically be GC'ed in the next iteration?

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,159 @@
...
+	//inserting keys in perfectly sorted lexicographical order can benefit or penalize certain tree/map data structures
+	rand.Shuffle(len(keys), func(i, j int) {
+		keys[i], keys[j] = keys[j], keys[i]
+	})
+	return keys
+}
+
+func benchmarkInsert(b *testing.B, cache *lru.Cache, keys []string) {
+	data := testData{Value: 1, DataSize: 10}
+	b.ResetTimer()
```
kislaykishore: This is not needed.
```suggestion
```

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,159 @@
...
+		i++
+	}
+}
+
+func benchmarkLookup(b *testing.B, cache *lru.Cache, keys []string) {
+	data := testData{Value: 1, DataSize: 10}
+	for _, key := range keys {
+		_, _ = cache.Insert(key, data)
+	}
+	b.ResetTimer()
```
kislaykishore: Not needed - b.Loop() already resets the timer.
```suggestion
```

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,159 @@
...
+func benchmarkErasePrefix(b *testing.B, cache *lru.Cache, prefixMap map[string][]string, prefixes []string) {
+	data := testData{Value: 1, DataSize: 10}
+
+	for _, keysInPrefix := range prefixMap {
+		for _, key := range keysInPrefix {
+			_, _ = cache.Insert(key, data)
+		}
+	}
+
+	b.ResetTimer()
```
kislaykishore: This is redundant.
```suggestion
```

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,159 @@
...
+		b.StartTimer()
+		cache.EraseEntriesWithGivenPrefix(prefix)
+		//reset the timer so that we only measure erasure time
+		b.StopTimer()
+
+		//restore the map
+		keysToRestore := prefixMap[prefix]
+		for _, key := range keysToRestore {
+			_, _ = cache.Insert(key, data)
+		}
```
kislaykishore, line 103: How about we recreate the cache from scratch for every loop?

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,159 @@
...
+	return keys
+}
+
+func benchmarkInsert(b *testing.B, cache *lru.Cache, keys []string) {
+	data := testData{Value: 1, DataSize: 10}
+	b.ResetTimer()
+
+	i := 0
+	for b.Loop() {
+		key := keys[i%len(keys)]
```
kislaykishore, line 61: With this in b.Loop, we will overwrite the existing keys. So, it's not pure insert but insert + overwrite. Is that intentional?

internal/cache/lru/lru_workload_benchmark_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	for p := range prefixCount {
+		prefix := ""
+		if depth == 0 {
+			prefix = fmt.Sprintf("prefix%d-", p)
+		} else {
+			for d := range depth {
+				prefix += fmt.Sprintf("dir%d/", p*depth+d)
+			}
+		}
```
kislaykishore: This looks same as the `generateKeys` method, doesn't it?

internal/cache/lru/lru_memory_test.go
vedantdas-source: Removed. While explicitly zeroing out the memory was meant as a defensive safeguard, the GC would automatically handle it in the subsequent loop iteration.

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Removed as no operations prior to the actual insert operations are performed.

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Yes it is. Because the b.Loop() runs as many times as required to get to a particular time, which could eventually wrap around our array of 1 million pre generated keys.  Generating string keys on the fly would skew the timer.

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Here the b.ResetTimer() is required as we build the cache first.When Go starts this function, the stopwatch starts at line 1. Without it the benchmark would also measure the time for inserting those entries.

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Here the b.ResetTimer() is required as we build the cache first.When Go starts this function, the stopwatch starts at line 1. Without it the benchmark would also measure the time for inserting those entries.

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Currently we only insert 1000 entries into the cache to reconstruct it. If we completely build the cache from scratch with 1M entries on every iteration, the test could take a lot of time to finish. I agree with your concern as it increases the complexity of the code. Would you like me to reconstruct the cache it would be more readable ?

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Yes, it is redundant. Moved the logic inside the `generateKeys` method itself

internal/cache/lru/lru_workload_benchmark_test.go
kislaykishore: Please refer to this: https://go.dev/blog/testing-b-loop#old-benchmark-loop-problems. b.Loop() resets the timer.

internal/cache/lru/lru_workload_benchmark_test.go
kislaykishore: Please check my other comment regarding this: https://github.com/GoogleCloudPlatform/gcsfuse/pull/4803/changes#r3502879419

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: My mistake! I didn't realize b.Loop() automatically resets the timer now. Removed b.ResetTimer()

internal/cache/lru/lru_workload_benchmark_test.go
vedantdas-source: Removed b.ResetTimer()

internal/cache/lru/lru_memory_test.go
```diff
@@ -0,0 +1,100 @@
...
+		pureCache := lru.NewCache(capacity)
+		for _, p := range paths {
+			_, _ = pureCache.Insert(p, dummyValue{})
+		}
+		alloc := getMemStats()
+		pureMem := alloc - baseMem
+		runtime.KeepAlive(pureCache)
+
+		t.Logf("%-20s Heap Used: %10.2f MB", "Pure MapLRU", float64(pureMem)/(1024*1024))
+		t.Logf("")
```
kislaykishore: Why do we need this?

internal/cache/metadata/statcache_memory_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		t.Logf("%-20s Heap Used: %10.2f MB", "StatCache", float64(pureMem)/(1024*1024))
+		t.Logf("")
```
kislaykishore: Why is this needed?

internal/cache/lru/lru_memory_test.go
vedantdas-source: We need runtime.KeepAlive() here to prevent the Go from destroying the cache before we can measure it. When we force runtime.GC() to run right before taking our memory snapshot, the Garbage Collector sees that the cache is dead and completely sweeps it from RAM which causes memAfter to actually be smaller than baseMem and results in a integer underflow.

internal/cache/metadata/statcache_memory_test.go
vedantdas-source: We need runtime.KeepAlive() here to prevent the Go from destroying the cache before we can measure it. When we force runtime.GC() to run right before taking our memory snapshot, the Garbage Collector sees that the cache is dead and completely sweeps it from RAM which causes memAfter to actually be smaller than baseMem and results in a integer underflow.

internal/cache/lru/lru_memory_test.go
vedantdas-source: if we remove the runtime.KeepAlive(pure cache) line from the script it reports around 1.5 million terabytes of memory consumption which is a integer underflow.

internal/cache/metadata/statcache_memory_test.go
kislaykishore: Why are we doing t.Logf("") at the end?

internal/cache/metadata/statcache_memory_test.go
kislaykishore: We would like to use the same benchmark in Radix cache as well. Would this not help there? Plus even though the strings are not reclaimed, the slice could get reclaimed, right?

internal/cache/metadata/statcache_memory_test.go
vedantdas-source: That is just to add an empty newline between each workload iteration

kislaykishore: you could simply add \n in the previous statement, right?

vedantdas-source: Done.

internal/cache/metadata/statcache_memory_test.go
vedantdas-source: Yes, the slice backing array itself would get reclaimed without it. we'd need paths kept alive to ensure a perfect 1-to-1 baseline comparison

kislaykishore (approved):

PR #4804, opened 2026-06-26
feat(storage): implement atomic compose via DeleteSourceObjects

PranjalC100:
### Description
This pull request implements an atomic compose operation by introducing a `DeleteSourceObjects` boolean inside the `ComposeObjectsRequest`. It leverages the new GCS Compose API functionality that allows atomically deleting source objects on the server-side the exact millisecond a compose operation completes. 

**Impact:**
This change fundamentally resolves race conditions, early deletion billing spikes, and "soft delete" charges commonly encountered during append-heavy AI/ML workloads.

**Key Changes:**
* **Core Implementation:** Passed the `DeleteSourceObjects: true` flag during GCS object composition via `ComposeObjectCreator` and removed the manual `DeleteObject` deferred calls.
* **Garbage Collector (Fallback):** The background garbage collector was intentionally retained as a resilient fallback to clean up orphaned chunks if a process crashes before compose.
- **Mock & Test Updates**: Updated `fake/bucket.go` to simulate source deletion and added unit tests in `bucket_tests.go`, `bucket_handle_test.go`, and `compose_object_creator_test.go`.

### Link to the issue in case of a bug fix:
[b/430606937](https://buganizer.corp.google.com/issues/430606937)

### Testing details
1. **Manual:** Executed standard test sweeps and compilation checks locally via `make build`.
2. **Unit tests:** Verified `go test -race ./internal/storage/... ./internal/gcsx/...` passes cleanly.
3. **Integration tests:** Triggered via PR labels (`execute-integration-tests`, `execute-integration-tests-on-zb`).

N/A


internal/gcsx/compose_object_creator.go
meet2mky, line 1: I think the same PR should modify the garbage collector code? wdyt

PranjalC100: I will modify the garbage collector code in the next PR. Discussed and agreed upon offline.

anushka567 (dismissed):
LGTM.

internal/gcsx/compose_object_creator.go
```diff
@@ -107,17 +106,16 @@ func (oc *composeObjectCreator) Create(
...
-			ctx,
-			&gcs.DeleteObjectRequest{
-				Name:       tmp.Name,
-				Generation: 0, // Delete the latest generation of temporary object.
-			})
-
-		if err == nil && deleteErr != nil {
-			err = fmt.Errorf("DeleteObject: %w", deleteErr)
+		if err != nil {
+			_ = oc.bucket.DeleteObject(
```
abhishek10004: In case this delete fails, we should log this or return this along with the original error.

abhishek10004 (commented):
Left one minor comment. Rest looks good. Haven't checked the UTs, will rely on previous reviewer's review for that.

internal/gcsx/compose_object_creator.go
PranjalC100: ACK. Added the error logging incase deleteObject fails.

abhishek10004 (approved):
LGTM

PR #4808, opened 2026-06-27
feat(Flaky_Test): use t.TempDir() instead of hardcoded path and migrate ContentCache tests to testify

charith87:
Helps in standard IDE integration, moving away from unmaintained library, and improving test isolation by replacing the hardcoded /tmp path with t.TempDir().


geertj:
LGTM.

Thanks @charith87 this kind of improvements are sorely needed. I was running into both issues with ogletest, and hard coded test directories (including in $HOME) myself as well.

geertj (approved):

PR #4815, opened 2026-06-27
test(build): support build cache inside tests via env var

geertj:
Checking GCSFUSE_USE_BUILD_CACHE environment variable in buildBinaries. If it is set to "true", we skip creating temporary, isolated GOPATH and GOCACHE folders and compile in standard module mode.

This keeps GCSFuse building cleanly/isolated by default, but allows unit tests to set GCSFUSE_USE_BUILD_CACHE to speed up TestVersion execution time from ~24s to ~3.6s.

TAG=agy
CONV=8c7b5ab6-91ae-4a84-911e-1788917d5ab8

### DIFF
--- a/tools/build_gcsfuse/main.go
+++ b/tools/build_gcsfuse/main.go
@@ -41,6 +41,7 @@ import (
 	"os/exec"
 	"path"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


tools/build_gcsfuse/main.go
```diff
@@ -76,36 +76,42 @@ func buildBinaries(dstDir, srcDir, version, arch string, buildArgs []string) (er
...
 	}
 
-	// Create a directory to become GOPATH for our build below.
-	gopath, err := os.MkdirTemp("", "build_gcsfuse_gopath")
-	if err != nil {
-		err = fmt.Errorf("TempDir: %w", err)
-		return
-	}
-	defer os.RemoveAll(gopath)
+	clean := os.Getenv("GCSFUSE_USE_BUILD_CACHE") != "1"
```
charith87: We can use strconv.ParseBool() which will take care of all cases like true, 1 etc.

geertj: Updated

charith87 (approved):

PR #4818, opened 2026-06-27
test(fs/handle): fix BufferedReader resource leak timeout in file_test

geertj:
In Test_ReadWithReadManager_ConcurrentReadsWithBufferedReader: Added invocation of resp.Callback() after validating each read response to correctly release outstanding data slice references held by the BufferedReader.

Without this callback, BufferedReader.Destroy would wait until a 10-second timeout expired before releasing resources, causing the test case to hang for 10.15 seconds and print a warning. With the callback, the test executes cleanly in 0.15s (a ~67x speedup).

TAG=agy
CONV=8c7b5ab6-91ae-4a84-911e-1788917d5ab8

### DIFF
--- a/internal/fs/handle/file_test.go
+++ b/internal/fs/handle/file_test.go
@@ -1441,6 +1441,10 @@ func (t *fileTest) Test_ReadWithReadManager_ConcurrentReadsWithBufferedReader() 
 				Offset: offset,
 			}, int32(readSize))
 
+			if resp.Callback != nil {
+				defer resp.Callback()
+			}
+
 			assert.NoError(t.T(), err)
 			assert.Equal(t.T(), readSize, resp.Size)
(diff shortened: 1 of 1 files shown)


internal/fs/handle/file_test.go
```diff
@@ -1444,6 +1444,9 @@ func (t *fileTest) Test_ReadWithReadManager_ConcurrentReadsWithBufferedReader()
 			assert.NoError(t.T(), err)
 			assert.Equal(t.T(), readSize, resp.Size)
 			results[index] = util.ConvertReadResponseToBytes(resp.Data, resp.Size)
+			if resp.Callback != nil {
+				resp.Callback()
```
charith87: We can do defer resp.Callback() for assertion failure cases

geertj: Good point, updated.

charith87 (approved):

PR #4824, opened 2026-06-30
test(pirlo): Add tests to validate the finalization logic and metadata eventual consistency for Rapid bucket

vipnydav:
### Description

This PR renames and generalizes the `rapid_appends` integration test package to `rapid` and introduces new test coverage for zonal bucket environments. 

**Key Changes:**
1. **Rename test package:** Renamed directories, packages, config structures, and test scripts from `rapid_appends` to `rapid` to represent a broader set of rapid validation tests.
2. **Add Finalize test suite:** Introduced `TestFinalizeRapidWritesTestSuite` in `tools/integration_tests/rapid/finalize_test.go` to verify file closure behaviors in both finalized and unfinalized states depending on the `--finalize-file-for-rapid` configuration.
3. **Add Stat & List test suite:** Added `TestStatAndListTestSuite` in `tools/integration_tests/rapid/stat_and_list_test.go` to validate that eventually we get correct size for objects created in the GCSFuse test directory through directory listing and stat operations.

b/526451412, b/526451679,b/526451122

### Testing details
1. Manual - Verified by running on zonal bucket in same and different zones and flat and hns buckets.
2. Unit tests - NA

### DIFF
--- a/tools/integration_tests/rapid_operations/helpers_test.go
+++ b/tools/integration_tests/rapid_operations/helpers_test.go
@@ -1,0 +1,85 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rapid_operations
+
+import (
+	"math/rand/v2"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/client"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup"
+	"github.com/stretchr/testify/require"
+)
+
+func (t *BaseSuite) createGCSFile(useAppendableAPI bool, fileSize int64) (filePath string, content []byte) {
+	t.T().Helper()
+
+	t.fileName = fileNamePrefix + setup.GenerateRandomString(5)
+	filePath = path.Join(t.primaryMount.testDirPath, t.fileName)
+	content, err := operations.GenerateRandomData(fileSize)
+	require.NoError(t.T(), err)
+
+	// Create file directly in GCS using Go SDK with the chosen API.
+	err = client.CreateObjectWithAPI(testEnv.ctx, testEnv.storageClient, path.Join(testDirName, t.fileName), content, useAppendableAPI)
+	require.NoError(t.T(), err)
+	return filePath, content
+}
+
+// declare a function type for read and verify
+type readAndVerifyFunc func(t *testing.T, filePath string, expectedContent []byte)
+
+func readSequentiallyAndVerify(t *testing.T, filePath string, expectedContent []byte) {
+	file, err := os.OpenFile(filePath, os.O_RDONLY, setup.FilePermission_0600)
+	require.NoError(t, err)
+	readContent, err := operations.ReadFileSequentially(file, 1024*1024)
+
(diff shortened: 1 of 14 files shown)


tools/integration_tests/rapid_operations/helpers_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"path"
+	"testing"
+	"time"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/operations"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup"
+	"github.com/stretchr/testify/require"
+)
+
+func (t *BaseSuite) createGCSFile(useAppendableAPI bool, fileSize int64) (filePath string, content []byte) {
```
abhishek10004, line 29: We should already have something for this. Can you please check once?

tools/integration_tests/rapid/reads_after_appends_test.go
```diff
@@ -85,7 +33,7 @@ func (t *SingleMountReadsTestSuite) runAppendAndReadTest(verifyFunc readAndVerif
 	t.createUnfinalizedObject()
 	defer t.deleteUnfinalizedObject()
 
-	appendFileHandle := operations.OpenFileInMode(t.T(), path.Join(t.primaryMount.testDirPath, t.fileName), fileOpenModeAppend|syscall.O_DIRECT)
```
abhishek10004, line 88: Why were we running it with O_DIRECT earlier?

vipnydav: As per discussion with @anushka567, we used `O_DIRECT` mostly to keep it consistent with other methods where we created file.

But as we discussed offline we should test with `O_DIRECT` as we want to test the GCS behavior and page cache should not be involved.

For my kernel version this was failing with `O_DIRECT` but passed in newer kernel versions. 

I'll remove this change for now and try to debug this issue separately.

tools/integration_tests/rapid_operations/helpers_test.go
vipnydav: This method allows you to use different API while creating file which is specific to `Pirlo` bucket. Added this to integration test utility, so that it can be used by other Tests in future as well.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -277,6 +277,25 @@ func CreateFinalizedObjectOnGCS(ctx context.Context, client *storage.Client, obj
 	return nil
 }
 
+// CreateObjectWithAPI creates an object on GCS with the given content, allowing the choice of whether to use the appendable API.
+func CreateObjectWithAPI(ctx context.Context, client *storage.Client, object string, content []byte, useAppendableAPI bool) error {
+	bucket, object := setup.GetBucketAndObjectBasedOnTypeOfMount(object)
+	o := getBucketHandle(client, bucket).Object(object)
+
+	wc := o.NewWriter(ctx)
+	wc.Append = useAppendableAPI
```
abhishek10004, line 286: How would writer switch to resumable when append is set to false?

vipnydav: Here is the logic for choosing Buffer Sender in Go SDK: https://github.com/googleapis/google-cloud-go/blob/6f4e42b8f8f28c602c81f424c824c8e328485fed/storage/grpc_writer.go#L298
- If Append is set to true then Appendable API is used
- If Chunk Size is set to zero then Single-Shot uploads are used
- Else Resumable Uploads are used.

abhishek10004: Got it. Thanks.

tools/integration_tests/util/client/storage_client.go
```diff
@@ -277,6 +277,25 @@ func CreateFinalizedObjectOnGCS(ctx context.Context, client *storage.Client, obj
...
+	wc.FinalizeOnClose = true
+
+	if _, err := wc.Write(content); err != nil {
+		return fmt.Errorf("wc.Write failed for object %q: %w", object, err)
+	}
+	if err := wc.Close(); err != nil {
+		return fmt.Errorf("wc.Close failed for object %q: %w", object, err)
+	}
+	operations.WaitForSizeUpdate(setup.IsZonalBucketRun(), operations.WaitDurationAfterCloseZB)
```
abhishek10004: We do not need to wait if we're finalizing the object on close.

vipnydav: Updated

abhishek10004 (approved):

PR #4825, opened 2026-06-30
perf(stat-cache memory): storing expiry nanoseconds (int64) instead of time.Time

raj-prince:
### Description
- Stat cache entries stores expiration time as time.Time struct (24 bytes), which is used to check expiry. This seems redundant and only storing the expirationNs-int64 (8 bytes) will be enough to check expiration. Saves 16bytes per entry.
- Without change:
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/cache/metadata
cpu: AMD EPYC 7B13
BenchmarkStatCache_Insert-64                     1724638               749.4 ns/op           299 B/op          6 allocs/op
BenchmarkStatCache_AddNegativeEntry-64           1679168               677.1 ns/op           224 B/op          5 allocs/op
BenchmarkStatCache_LookUp-64                     5474361               214.5 ns/op            71 B/op          2 allocs/op
```
- With change:
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/cache/metadata
cpu: AMD EPYC 7B13
BenchmarkStatCache_Insert-64                     1714723               745.1 ns/op           283 B/op          6 allocs/op
BenchmarkStatCache_AddNegativeEntry-64           2020422               647.2 ns/op           204 B/op          5 allocs/op
BenchmarkStatCache_LookUp-64                     6030410               195.4 ns/op            55 B/op          2 allocs/op
```

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/cache/metadata/stat_cache.go
+++ b/internal/cache/metadata/stat_cache.go
@@ -111,9 +111,9 @@ type statCacheBucketView struct {
 // An entry in the cache, pairing an object with the expiration time for the
 // entry. Nil object means negative entry.
 type entry struct {
-	m          *gcs.MinObject
-	f          *gcs.Folder
-	expiration time.Time
+	m            *gcs.MinObject
+	f            *gcs.Folder
+	expirationNS int64
 	// Set to true only for implicit directory entries. This flag will always remain false for negative entries and explicit objects.
 	implicitDir bool
 }
@@ -200,8 +200,8 @@ func (sc *statCacheBucketView) Insert(m *gcs.MinObject, expiration time.Time) {
 
 	// Insert an entry.
 	e := entry{
-		m:          m,
-		expiration: expiration,
+		m:            m,
+		expirationNS: expiration.UnixNano(),
 	}
 
 	if _, err := sc.sharedCache.Insert(name, e); err != nil {
@@ -236,8 +236,8 @@ func (sc *statCacheBucketView) InsertImplicitDir(objectName string, expiration t
 
 	// Insert an entry.
 	e := entry{
-		implicitDir: true,
-		expiration:  expiration,
+		implicitDir:  true,
+		expirationNS: expiration.UnixNano(),
 	}
 
 	if _, err := sc.sharedCache.Insert(name, e); err != nil {
@@ -250,8 +250,8 @@ func (sc *statCacheBucketView) AddNegativeEntry(objectName string, expiration ti
 
 	// Insert a negative entry.
 	e := entry{
-		m:          nil,
-		expiration: expiration,
+		m:            nil,
+		expirationNS: expiration.UnixNano(),
 	}
 
 	if _, err := sc.sharedCache.Insert(name, e); err != nil {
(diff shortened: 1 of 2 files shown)


internal/cache/metadata/stat_cache_benchmark_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+func BenchmarkStatCache_Insert(b *testing.B) {
+	lc := lru.NewCache(util.MiBsToBytes(100))
+	sc := metadata.NewStatCacheBucketView(lc, "")
+	expiration := time.Now().Add(time.Hour)
+
+	b.ResetTimer()
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
```
kislaykishore, line 36: Use b.Loop()

internal/cache/metadata/stat_cache_benchmark_test.go
```diff
@@ -0,0 +1,84 @@
...
+		name := fmt.Sprintf("file-%d", i)
```
kislaykishore, line 37: Using sprintf affects the benchmarking results. Prefer using strings.Builder with strconv for formatting integers.

internal/cache/metadata/stat_cache_benchmark_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
+)
+
+func BenchmarkStatCache_Insert(b *testing.B) {
+	lc := lru.NewCache(util.MiBsToBytes(100))
+	sc := metadata.NewStatCacheBucketView(lc, "")
+	expiration := time.Now().Add(time.Hour)
+
+	b.ResetTimer()
+	b.ReportAllocs()
```
kislaykishore, line 34: Why is this needed? You could achieve the same using benchmem

internal/cache/metadata/stat_cache_benchmark_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs"
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
+)
+
+func BenchmarkStatCache_Insert(b *testing.B) {
+	lc := lru.NewCache(util.MiBsToBytes(100))
+	sc := metadata.NewStatCacheBucketView(lc, "")
+	expiration := time.Now().Add(time.Hour)
+
+	b.ResetTimer()
```
kislaykishore, line 33: If you use b.loop, this will not be needed.

kislaykishore (commented):
Gave some comments for the first method. Please incorporate them into other methods as well.

internal/cache/metadata/stat_cache.go
```diff
@@ -111,9 +111,9 @@ type statCacheBucketView struct {
 // An entry in the cache, pairing an object with the expiration time for the
 // entry. Nil object means negative entry.
 type entry struct {
-	m          *gcs.MinObject
-	f          *gcs.Folder
-	expiration time.Time
+	m            *gcs.MinObject
+	f            *gcs.Folder
+	expirationNS int64
```
meet2mky, line 116: Why do we need nanoseconds precision here?

meet2mky: Our ttls are in seconds. Wdyt?

raj-prince:
Closing in favor of - https://github.com/GoogleCloudPlatform/gcsfuse/pull/4857

PR #4837, opened 2026-07-02
feat(kernel range reader): Add thread-safe LIFO BufferPool with mmap allocation and sync.Pool fallback

abhishek10004:
### Description
# Description

This PR introduces a thread-safe **`BufferPool`** in GCSFuse to manage a pool of buffers. These buffers are used to optimize large read paths (such as the kernel reader path). 

The pool dynamically allocates static buffers via `mmap` on-demand up to a configured limit (`maxStatic`). When the static pool is exhausted, or if an allocation fails, it transparently falls back to a `sync.Pool` of Go-allocated buffers. This design prevents goroutines from blocking and ensures high throughput under heavy concurrent read workloads.

---

# Why Stack-Based (LIFO) Approach?

We chose a **LIFO (Last-In, First-Out) stack** (`staticBuffers [][]byte`) for managing the static buffers for the following reasons:

1. **Cache Locality:** The buffer most recently returned to the pool is the most likely to still reside in the CPU's cache hierarchy (L1/L2/L3). Reusing it immediately (LIFO) maximizes cache hits and reduces memory access latency compared to a FIFO queue.
2. **Low Overhead:** Slice append and pop operations on a LIFO stack are extremely fast ($O(1)$) and avoid complex index bookkeeping.
3. **Reduced Lock Contention:**
   * **Lock-free Slot Reservation:** The pool uses an atomic Compare-And-Swap (CAS) counter (`allocated`) to reserve slots up to `maxStatic` without holding any locks.
   * **Short Critical Section:** The mutex (`bp.mu`) is only held for the duration of pushing to or popping from the slice, which takes only a few nanoseconds.
   * This ensures the pool scales exceptionally well under high contention

---

# Benchmarking Results

We ran parallel benchmarks on an `Intel(R) Xeon(R) CPU @ 2.20GHz` (Linux/amd64) simulating concurrent workers (1 to 50 goroutines) requesting and releasing `1MiB` buffers.

### 1. Steady State (Warmed Cache)
*Static buffers are pre-allocated/warmed-up before the benchmark begins.*

| Concurrency (CPU Threads) | Throughput (ns/op) | Memory Allocation (B/op) | Allocs/op |
| :--- | :--- | :--- | :--- |
| **1** (No Contention) | **64.19 ns/op** | 0 B/op | 0 allocs/op |
| **2** | **83.64 ns/op** | 0 B/op | 0 allocs/op |
| **4** | **207.40 ns/op** | 0 B/op | 0 allocs/op |
| **8** | **295.70 ns/op** | 0 B/op | 0 allocs/op |
| **16** | **308.40 ns/op** | 0 B/op | 0 allocs/op |
| **32** | **242.20 ns/op** | 0 B/op | 0 allocs/op |
| **50** (High Contention) | **201.10 ns/op** | 0 B/op | 0 allocs/op |

### 2. On-Demand Allocation (Cold Cache)
*Static buffers are allocated on-demand via `mmap` during the benchmark.*

| Concurrency (CPU Threads) | Throughput (ns/op) | Memory Allocation (B/op) | Allocs/op |
| :--- | :--- | :--- | :--- |
| **1** (No Contention) | **61.62 ns/op** | 0 B/op | 0 allocs/op |
| **2** | **86.26 ns/op** | 0 B/op | 0 allocs/op |
| **4** | **202.40 ns/op** | 0 B/op | 0 allocs/op |
| **8** | **296.50 ns/op** | 0 B/op | 0 allocs/op |
| **16** | **306.70 ns/op** | 0 B/op | 0 allocs/op |
| **32** | **229.00 ns/op** | 0 B/op | 0 allocs/op |
| **50** (High Contention) | **211.70 ns/op** | 0 B/op | 0 allocs/op |

### Key Takeaways
* **Negligible Cold Start Overhead:** Over a typical benchmark run, the cost of the initial `mmap` allocations on a cold cache is negligible, performing virtually identically to the warmed cache.
* **Excellent Scalability:** Even under high contention (50 concurrent goroutines), the time per operation remains low (~200ns).

b/530541554

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/fs/buffer_pool.go
+++ b/internal/fs/buffer_pool.go
@@ -1,0 +1,176 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fs
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"syscall"
+	"unsafe"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
+)
+
+// BufferPool manages a pool of bufferSize buffers.
+// It allocates static buffers via mmap on-demand up to maxStatic.
+// If mmap allocation fails or more buffers are needed concurrently, it falls back
+// to a sync.Pool of Go-allocated buffers.
+type BufferPool struct {
+	mu            sync.Mutex
+	staticBuffers [][]byte // LIFO stack of static buffers
+	staticMap     sync.Map // map[uintptr][]byte for routing
+	fallbackPool  sync.Pool
+	bufferSize    int
+	maxStatic     int
+	allocated     int32 // atomic count of allocated static buffers
+}
+
+// NewBufferPool creates and initializes a BufferPool.
+func NewBufferPool(maxStatic int, bufferSize int) *BufferPool {
+	return &BufferPool{
+		staticBuffers: make([][]byte, 0, maxStatic),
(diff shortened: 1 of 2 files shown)


abhishek10004:
Closing this PR. Will be going with sync pool for now and use mmap for allocation only with dynamic scaling based implementation.

PR #4842, opened 2026-07-03
feat(kernel range reader): Support vectored reads with kernel readers

abhishek10004:
### Description
This PR introduces support for vectored reads (`[][]byte`) in `gcsfuse kernel readers`. This allows the kernel to request reads into multiple non-contiguous memory buffers, which are then populated sequentially.

This PR is based on the `abhishek/buffer_pool_1mib` branch.

## Key Changes
1. **API Extensions**: Added `Buffers [][]byte` and `Size int64` fields to `gcsx.ReadRequest` to support multiple destination buffers.
2. **`VectoredWriter` Utility**: Introduced `VectoredWriter` which implements `io.Writer` and `io.ReaderFrom` to sequentially write into a slice of byte slices.
3. **`MrdInstance` Refactoring**: Refactored `MrdInstance.Read` to accept an `io.Writer` instead of a flat `[]byte` buffer, allowing it to support both single-buffer and multi-buffer writers.
4. **`KernelRangeReader` Integration**: Integrated vectored reads using `VectoredWriter` and `io.CopyN`.
5. **`KernelMRDReader` Integration & Optimization**:
   * **Vectored Path**: Uses `VectoredWriter` to write into the requested `Buffers`.
   * **Standard Path (Optimized)**: For standard reads (where `Buffers` is empty), it continues to use `bytes.NewBuffer(req.Buffer[:0])`. This avoids the overhead of `VectoredWriter` and preserves the exact performance characteristics of the original implementation.
   * **Short-read Retries**: Supports retries for both paths, slicing either the standard buffer or the vectored buffers depending on the request type.

---

### 1. Implementing `io.ReaderFrom` on `VectoredWriter`
* **Rationale**: By implementing `io.ReaderFrom`, `VectoredWriter` can read directly from an `io.Reader` (like the GCS HTTP response stream) into the target user-space buffers. This bypasses intermediate allocations and avoids the CPU overhead of a double-copy (i.e., reading into a temporary buffer and then copying to the destination slices).

### 2. Using `io.CopyN` instead of `io.ReadFull` in a Loop
In `KernelRangeReader`, we use `io.CopyN(vectoredWriter, reader, bytesToRead)` rather than looping over individual buffers and calling `io.ReadFull`.
* **Rationale**: 
  * **Simplicity**: Looping over buffers manually requires complex state tracking (managing slice bounds, remaining bytes, and handling partial reads/EOFs across buffer boundaries).
  * **Delegation**: `io.CopyN` combined with `VectoredWriter.ReadFrom` delegates the loop complexity entirely to the writer's internal mechanics, keeping the reader code clean and declarative.
  * **Efficiency**: Since `VectoredWriter` implements `io.ReaderFrom`, `io.CopyN` will internally invoke `ReadFrom`, maintaining the zero-allocation path.

### 3. Refactoring `MrdInstance.Read` to accept `io.Writer`
* **Rationale**: Decoupling `MrdInstance` from a flat `[]byte` allows it to support both standard reads (by wrapping the `[]byte` in a buffer writer) and vectored reads (by passing `VectoredWriter`) through a single unified interface.

### 4. Dual-Path Writer Selection in `KernelMRDReader`
* **Rationale**: To ensure no performance regression for standard (non-vectored) reads, we avoid using `VectoredWriter` when `req.Buffers` is empty. Instead, we wrap the standard `req.Buffer` in a `bytes.Buffer` (using `req.Buffer[:0]` to start with zero length but retain capacity). This matches the exact behavior and performance of the previous implementation.

b/530765199

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


abhishek10004:
Have opened https://github.com/GoogleCloudPlatform/gcsfuse/pull/4845 directly on master.

PR #4844, opened 2026-07-06
feat(cache): introduce the core radix trie structure 

vedantdas-source:
### Description
This PR introduces only the core, internal tree structure (radixTree and radixNode) along with the core operations: `insertNode`, `getNode`, `deleteNode`, and `compressPathUpwards`. It does not yet include LRU eviction logic or the lru.Cache wrapper.

Design Overview  : A Radix Tree (or compressed trie) is a space-optimized tree where nodes with only a single child are merged with their parents. By storing directory paths as string prefixes across nodes, we can instantly locate or sever entire directory trees just by walking the prefix.

To map this multi-way tree efficiently in Go memory, we use the Left-Child Right-Sibling (LCRS) design pattern. Instead of storing a slice of children ` []*radixNode`, each node only points to its first child, and that child points to its sibling.

### Testing details
1. Manual - NA
2. Unit tests - Added `radix_test.go` to thoroughly test string prefix splitting, path compression, and exact-match leaf lookup.
3. Integration tests - NA

### DIFF
--- a/internal/cache/lru/radix.go
+++ b/internal/cache/lru/radix.go
@@ -63,42 +63,21 @@ func (n *radixNode) getChild(b byte) *radixNode {
 // addChild adds a child node, maintaining sorted order by the first byte of the prefix.
 func (n *radixNode) addChild(newChild *radixNode) {
 	newChild.parent = n
-	newChild.sibling = nil
-
-	if n.child == nil {
-		n.child = newChild
-		return
+	pcurr := &n.child
+	for *pcurr != nil && (*pcurr).prefix[0] < newChild.prefix[0] {
+		pcurr = &(*pcurr).sibling
 	}
-
-	if newChild.prefix[0] < n.child.prefix[0] {
-		newChild.sibling = n.child
-		n.child = newChild
-		return
-	}
-
-	curr := n.child
-	for curr.sibling != nil && curr.sibling.prefix[0] < newChild.prefix[0] {
-		curr = curr.sibling
-	}
-
-	newChild.sibling = curr.sibling
-	curr.sibling = newChild
+	newChild.sibling = *pcurr
+	*pcurr = newChild
 }
 
 // removeChild directly removes a child node by reference from the sibling list.
 func (n *radixNode) removeChild(childToRemove *radixNode) {
-	if n.child == childToRemove {
-		n.child = childToRemove.sibling
-		return
-	}
-
-	curr := n.child
-	for curr != nil && curr.sibling != childToRemove {
-		curr = curr.sibling
-	}
-
-	if curr != nil {
-		curr.sibling = childToRemove.sibling
+	for pcurr := &n.child; *pcurr != nil; pcurr = &(*pcurr).sibling {
+		if *pcurr == childToRemove {
+			*pcurr = childToRemove.sibling
+			break
+		}
 	}
 }
 
@@ -141,17 +120,11 @@ func (t *radixTree) insertNode(key string, value ValueType) (*radixNode, bool) {
 			parent: node,
 		}
 
-		if node.child == child {
-			splitNode.sibling = child.sibling
-			node.child = splitNode
-		} else {
-			curr := node.child
-			for curr != nil && curr.sibling != child {
-				curr = curr.sibling
-			}
-			if curr != nil {
+		for pcurr := &node.child; *pcurr != nil; pcurr = &(*pcurr).sibling {
+			if *pcurr == child {
 				splitNode.sibling = child.sibling
-				curr.sibling = splitNode
+				*pcurr = splitNode
+				break
 			}
 		}
 
@@ -233,17 +206,11 @@ func (t *radixTree) compressPathUpwards(curr *radixNode) {
 			onlyChild.prefix = curr.prefix + onlyChild.prefix
 			onlyChild.parent = curr.parent
 
-			if curr.parent.child == curr {
-				onlyChild.sibling = curr.sibling
-				curr.parent.child = onlyChild
-			} else {
-				prev := curr.parent.child
-				for prev != nil && prev.sibling != curr {
-					prev = prev.sibling
-				}
-				if prev != nil {
+			for pcurr := &curr.parent.child; *pcurr != nil; pcurr = &(*pcurr).sibling {
+				if *pcurr == curr {
 					onlyChild.sibling = curr.sibling
-					prev.sibling = onlyChild
+					*pcurr = onlyChild
+					break
 				}
 			}
 			curr = curr.parent
--- a/internal/cache/lru/radix.go
+++ b/internal/cache/lru/radix.go
@@ -56,6 +56,10 @@ func (n *radixNode) getChild(b byte) *radixNode {
 		if curr.prefix[0] == b {
 			return curr
 		}
+		//as the siblings are arranged in lexicographical order we can exit early
+		if curr.prefix[0] > b {
+			break
+		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


internal/cache/lru/radix.go
```diff
@@ -0,0 +1,221 @@
...
+	newChild.sibling = *pcurr
+	*pcurr = newChild
+}
+
+// removeChild directly removes a child node by reference from the sibling list.
+func (n *radixNode) removeChild(childToRemove *radixNode) {
+	for pcurr := &n.child; *pcurr != nil; pcurr = &(*pcurr).sibling {
+		if *pcurr == childToRemove {
+			*pcurr = childToRemove.sibling
+			break
```
kislaykishore:
```suggestion
			return
```

internal/cache/lru/radix.go
```diff
@@ -0,0 +1,221 @@
...
+				t.size++
+			}
+			node.value = value
+			return node, isNew
+		}
+
+		child := node.getChild(search[0])
+		if child == nil {
+			newLeaf := &radixNode{
+				prefix: strings.Clone(search),
```
kislaykishore, line 111: I believe you are using `strings.Clone` to allow the main string to get GC'ed when GCSFuse holds only the substring. Add a comment to say that.

internal/cache/lru/radix.go
```diff
@@ -0,0 +1,221 @@
...
+	i := 0
+	for i < len(a) && i < len(b) && a[i] == b[i] {
+		i++
+	}
+	return i
+}
+
+// getChild finds a child node whose prefix starts with the given byte.
+func (n *radixNode) getChild(b byte) *radixNode {
+	for curr := n.child; curr != nil; curr = curr.sibling {
```
kislaykishore, line 55: Given that we are keeping the siblings sorted, should we break early if we find that `cur.prefix[0] > b`

internal/cache/lru/radix.go
```diff
@@ -0,0 +1,221 @@
...
+	for pcurr := &n.child; *pcurr != nil; pcurr = &(*pcurr).sibling {
+		if *pcurr == childToRemove {
+			*pcurr = childToRemove.sibling
+			break
+		}
+	}
+}
+
+// insertNode inserts a new key into the radix tree and returns the leaf node.
+func (t *radixTree) insertNode(key string, value ValueType) (*radixNode, bool) {
```
kislaykishore, line 89: Passing `nil` value will cause tree corruption. Either don't insert or throw an error if the value is nil.

internal/cache/lru/radix.go
```diff
@@ -0,0 +1,221 @@
...
+			if isNew {
+				t.size++
+			}
+			node.value = value
+			return node, isNew
+		}
+
+		child := node.getChild(search[0])
+		if child == nil {
+			newLeaf := &radixNode{
```
kislaykishore, line 110: Consider using a `sync.Pool` as a future enhancement to reduce GC load.

internal/cache/lru/radix_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Test Overwrite (should not increase size)
+	node2, isNew2 := tree.insertNode("foo/bar", val2)
+	assert.False(t, isNew2)
+	assert.Equal(t, val2, node2.value)
+	assert.Equal(t, 1, tree.size)
+
+	// Test non-existent get
+	_, ok = tree.getNode("foo/baz")
+	assert.False(t, ok)
```
kislaykishore: Ideally, a unit-test should test just one unit of work. Create separate tests for each of these cases. Also, please follow the Arrange-Act-Assert segregation.
Ditto for other tests.

internal/cache/lru/radix_test.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func TestRadixTree_PrefixSplitting(t *testing.T) {
+	tree := newRadixTree()
+
+	tree.insertNode("foo/bar", mockValue{1})
+	tree.insertNode("foo/baz", mockValue{2}) // Splits "foo/ba" -> "r", "z"
+
+	assert.Equal(t, 2, tree.size)
```
kislaykishore: Please keep assertions at the end.

internal/cache/lru/radix_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Both should share the same parent routing node "foo/ba"
+	assert.Equal(t, n1.parent, n2.parent)
+	assert.Equal(t, "foo/ba", n1.parent.prefix)
+	assert.Nil(t, n1.parent.value) // Parent is just a routing node
+}
+
+func TestRadixTree_DeleteAndCompress(t *testing.T) {
+	tree := newRadixTree()
+
```
kislaykishore: Remove blank line.

internal/cache/lru/radix_test.go
```diff
@@ -0,0 +1,90 @@
...
+	tree.insertNode("foo/bar", mockValue{1})
+	tree.insertNode("foo/baz", mockValue{2})
+
```
kislaykishore: Remove blank line

internal/cache/lru/radix_test.go
```diff
@@ -0,0 +1,90 @@
...
+	n1, _ := tree.getNode("foo/bar")
+	n2, _ := tree.getNode("foo/baz")
+
+	// Both should share the same parent routing node "foo/ba"
+	assert.Equal(t, n1.parent, n2.parent)
+	assert.Equal(t, "foo/ba", n1.parent.prefix)
+	assert.Nil(t, n1.parent.value) // Parent is just a routing node
+}
+
+func TestRadixTree_DeleteAndCompress(t *testing.T) {
```
kislaykishore, line 113: See if you can split this into two test

internal/cache/lru/radix.go
vedantdas-source: Yes, that was the exact purpose of using  `strings.Clone` instead of simple string slicing which was done in the PoC which helped us achieve much more memory savings.

internal/cache/lru/radix.go
vedantdas-source: Yes, I completely missed that. Thanks for pointing it out.

internal/cache/lru/radix.go
vedantdas-source: The radix Cache's `Insert` function (which would be implemented in further commits) will have checks for that so I dont think that would be really necessary. Added checks here as well as a safeguard.

internal/cache/lru/radix.go
vedantdas-source: Understood, will keep it in mind.

internal/cache/lru/radix_test.go
vedantdas-source: Noted. Will use the Arrange-Act-Assert pattern for all further tests.

internal/cache/lru/radix.go
```diff
@@ -0,0 +1,230 @@
...
+
+// getChild finds a child node whose prefix starts with the given byte.
+func (n *radixNode) getChild(b byte) *radixNode {
+	for curr := n.child; curr != nil; curr = curr.sibling {
+		if curr.prefix[0] == b {
+			return curr
+		}
+		//as the siblings are arranged in lexicographical order we can exit early
+		if curr.prefix[0] > b {
+			break
```
kislaykishore:
```suggestion
			return nil
```

kislaykishore (approved):

PR #4845, opened 2026-07-06
feat(kernel range reader): Support vectored reads with kernel readers

abhishek10004:
### Description
This PR introduces support for vectored reads (`[][]byte`) in `gcsfuse kernel readers`. 

## Key Changes
1. **API Extensions**: Added `Buffers [][]byte` and `Size int64` fields to `gcsx.ReadRequest` to support multiple destination buffers.
2. **`VectoredWriter` Utility**: Introduced `VectoredWriter` which implements `io.Writer` and `io.ReaderFrom` to sequentially write into a slice of byte slices.
3. **`MrdInstance` Refactoring**: Refactored `MrdInstance.Read` to accept an `io.Writer` instead of a flat `[]byte` buffer, allowing it to support both single-buffer and multi-buffer writers.
4. **`KernelRangeReader` Integration**: Integrated vectored reads using `VectoredWriter` and `io.CopyN`.
5. **`KernelMRDReader` Integration & Optimization**:
   * **Vectored Path**: Uses `VectoredWriter` to write into the requested `Buffers`.
   * **Standard Path (Optimized)**: For standard reads (where `Buffers` is empty), it continues to use `bytes.NewBuffer(req.Buffer[:0])`. This avoids the overhead of `VectoredWriter` and preserves the exact performance characteristics of the original implementation.
   * **Short-read Retries**: Supports retries for both paths, slicing either the standard buffer or the vectored buffers depending on the request type.

---

### 1. Implementing `io.ReaderFrom` on `VectoredWriter`
* **Rationale**: By implementing `io.ReaderFrom`, `VectoredWriter` can read directly from an `io.Reader` (like the GCS HTTP response stream) into the target user-space buffers. This bypasses intermediate allocations and avoids the CPU overhead of a double-copy (i.e., reading into a temporary buffer and then copying to the destination slices).

### 2. Using `io.CopyN` instead of `io.ReadFull` in a Loop
In `KernelRangeReader`, we use `io.CopyN(vectoredWriter, reader, bytesToRead)` rather than looping over individual buffers and calling `io.ReadFull`.
* **Rationale**: 
  * **Simplicity**: Looping over buffers manually requires complex state tracking (managing slice bounds, remaining bytes, and handling partial reads/EOFs across buffer boundaries).
  * **Delegation**: `io.CopyN` combined with `VectoredWriter.ReadFrom` delegates the loop complexity entirely to the writer's internal mechanics, keeping the reader code clean and declarative.
  * **Efficiency**: Since `VectoredWriter` implements `io.ReaderFrom`, `io.CopyN` will internally invoke `ReadFrom`, maintaining the zero-allocation path.

### 3. Refactoring `MrdInstance.Read` to accept `io.Writer`
* **Rationale**: Decoupling `MrdInstance` from a flat `[]byte` allows it to support both standard reads (by wrapping the `[]byte` in a buffer writer) and vectored reads (by passing `VectoredWriter`) through a single unified interface.

### 4. Dual-Path Writer Selection in `KernelMRDReader`
* **Rationale**: To ensure no performance regression for standard (non-vectored) reads, we avoid using `VectoredWriter` when `req.Buffers` is empty. Instead, we wrap the standard `req.Buffer` in a `bytes.Buffer` (using `req.Buffer[:0]` to start with zero length but retain capacity). This matches the exact behavior and performance of the previous implementation.

b/530765199

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/gcsx/kernel_readers/kernel_range_reader.go
```diff
@@ -63,12 +63,13 @@ func (krr *KernelRangeReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest)
...
 
-	endOffset := req.Offset + int64(len(req.Buffer))
-	if endOffset > int64(obj.Size) {
-		endOffset = int64(obj.Size)
-	}
+	limit := int64(obj.Size) - req.Offset
+	bytesToRead := req.GetReadSize(limit)
+	endOffset := req.Offset + bytesToRead
 
 	reader, err := krr.bucket.NewReaderWithReadHandle(
```
raj-prince, line 81: Don't see clobbered error checked we have in another reader.

internal/gcsx/vectored_writer.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+			return n, err
+		}
+	}
+	return n, nil
+}
+
+// GetVectoredBuffers returns the buffers to be used for the read operation and the actual size to read.
+// The returned size will not exceed the total capacity of the buffers, or the provided limit (if limit > 0).
+func GetVectoredBuffers(req *ReadRequest, limit int64) ([][]byte, int64) {
```
raj-prince: Why are we keeping this function in this file? Why not in reader.go or the logic around ReadRequest?

internal/gcsx/mrd_instance.go
```diff
@@ -151,10 +150,9 @@ func (mi *MrdInstance) getMRDEntry() (*MRDEntry, error) {
 	return entry, nil
 }
 
-// Read downloads data from the GCS object into the provided buffer starting at the offset.
-// It handles the details of selecting a valid MRD entry, locking it, and waiting for the async download to complete.
-func (mi *MrdInstance) Read(ctx context.Context, p []byte, offset int64, metrics metrics.MetricHandle) (int, error) {
-	if len(p) == 0 {
+// Read downloads data from the GCS object into the provided io.Writer starting at the offset.
+func (mi *MrdInstance) Read(ctx context.Context, w io.Writer, offset int64, length int64, metrics metrics.MetricHandle) (int, error) {
```
raj-prince: What if writer doesn't have capacity to store the content of given length? Just want to make sure, it returns error and we have proper handling for that, given now buffer and downloaded-length is decoupled.

We can add a unit test for this scenario.

internal/gcsx/kernel_readers/kernel_mrd_reader.go
```diff
@@ -102,21 +104,70 @@ func (kmr *KernelMRDReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (
...
+
 		var bytesReadOnRetry int
-		bytesReadOnRetry, err = kmr.mrdInstance.Read(ctx, retryBuffer, retryOffset, kmr.metrics)
+		bytesReadOnRetry, err = kmr.mrdInstance.Read(ctx, retryWriter, retryOffset, sizeToRead-int64(bytesRead), kmr.metrics)
 		bytesRead += bytesReadOnRetry
 	}
 	return gcsx.ReadResponse{Size: bytesRead}, err
 }
 
+func sliceBuffers(buffers [][]byte, offset int) [][]byte {
```
raj-prince: Can we simplify the implementation by re-using offset variable only? Something like below?

```go
func sliceBuffers(buffers [][]byte, offset int) [][]byte {
	if offset <= 0 {
		return buffers
	}
	for i, b := range buffers {
		if offset < len(b) {
			if offset == 0 {
				return buffers[i:]
			}
			result := make([][]byte, 0, len(buffers)-i)
			result = append(result, b[offset:])
			result = append(result, buffers[i+1:]...)
			return result
		}
		offset -= len(b)
	}
	return nil
}

```

raj-prince (commented):
Looks good overall, few minor comments.

internal/gcsx/vectored_writer.go
```diff
@@ -0,0 +1,92 @@
...
+// limitations under the License.
+
+package gcsx
+
+import "io"
+
+// VectoredWriter implements io.Writer and writes data sequentially into a slice of byte slices.
+type VectoredWriter struct {
+	buffers [][]byte
+	offset  int
```
vadlakondaswetha: index here is tracking the index from the list of buffers, but offset is specific to a buffer. can you make another struct which stores the buffers and offset and use it in vectored write.

internal/gcsx/kernel_readers/kernel_range_reader.go
```diff
@@ -90,7 +91,15 @@ func (krr *KernelRangeReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest)
 		}
 	}()
 
-	n, err := io.ReadFull(reader, req.Buffer[:endOffset-req.Offset])
+	var n int
+	if len(req.Buffers) == 0 {
```
vadlakondaswetha, line 114: why if-else, are we going to support nonvectored reads for kernel range reader flow?

internal/gcsx/kernel_readers/kernel_mrd_reader.go
```diff
@@ -102,21 +104,70 @@ func (kmr *KernelMRDReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (
...
 
 	var err error
-	bytesRead, err = kmr.mrdInstance.Read(ctx, req.Buffer, req.Offset, kmr.metrics)
-	if isShortRead(bytesRead, len(req.Buffer), err) {
-		logger.Tracef("Short read detected: read %d bytes out of %d requested. Retrying...", bytesRead, len(req.Buffer))
+	var writer io.Writer
+	var sizeToRead int64
+	var buffers [][]byte
+
+	if len(req.Buffers) == 0 {
```
vadlakondaswetha: why are we doing changes for mrd flow. this requires functional and perf testing to ensure nothing is broken. can we keep out of this release

internal/gcsx/kernel_readers/kernel_range_reader.go
abhishek10004: Yes, we would be supporting non vectored reads as well. For smaller read requests, i.e. requests within 1M, jacobsa would be sending the original buffer which it had used to read from /dev/fuse and we would continue using that.

There was one way to simplify this code by having NewVectoredWriter extract this logic and return a wrapping [][]byte in any case (even when the underlying buffer is from jacobsa), but that would have resulted in an extra allocation on heap for smaller reads. Hence, I went with this approach (for smaller reads, there is no extra allocation like it was the case earlier).

internal/gcsx/kernel_readers/kernel_mrd_reader.go
```diff
@@ -81,7 +83,7 @@ func (kmr *KernelMRDReader) CheckInvariants() {
 // requested byte range.
 func (kmr *KernelMRDReader) ReadAt(ctx context.Context, req *gcsx.ReadRequest) (gcsx.ReadResponse, error) {
 	// If the destination buffer is empty, there's nothing to read.
-	if len(req.Buffer) == 0 {
+	if len(req.Buffer) == 0 && len(req.Buffers) == 0 {
 		return gcsx.ReadResponse{}, nil
 	}
```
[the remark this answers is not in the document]
abhishek10004: Not needed. MRD implementation in Go SDK handles negative offsets.

internal/gcsx/vectored_writer.go
abhishek10004: I had implemented it this way to minimize any extra allocation. The idea was that all buffers before the index would be fully written to. Hence, offset would always be used in context to the buffer pointed to by index.

I have now changed the implementation to have another struct which keeps offset along with buffer.

internal/gcsx/kernel_readers/kernel_mrd_reader.go
abhishek10004: The changes for MRD were being done because they were safe changes and would help us in future when we decide to support vectored reads for rapid as well. They were safe as we'd not be invoking that path for now (we'd still be going with single buffer and hence using the old path which is in production).

Discussed this offline with Swetha and I'd now add those changes after the next release.

raj-prince (approved):

vadlakondaswetha (approved):

PR #4846, opened 2026-07-06
feat(native-retries): wire Control Plane mount retry predicate (ShouldRetryOnMount)

dangarhs:
### Description
- Added `EnableMountRetries bool` field to `storageutil.StorageClientConfig` to transport the `--enable-mount-retries` flag from the CLI configuration layer into the storage layer.
- Mapped `newConfig.GcsRetries.EnableMountRetries` to `storageClientConfig.EnableMountRetries` in `createStorageHandle()` inside `cmd/legacy_main.go`.
- Added a `withRetryOnMount` client wrapper function to configure the Control Plane storage client to handle retries during mount initialization.
- Wrapped the control client with `withRetryOnMount` instead of `withRetryOnStorageLayout` inside `NewStorageHandle()` of `internal/storage/storage_handle.go`.
- Updated `storageControlClientWithRetry.GetStorageLayout()` in `internal/storage/control_client_wrapper.go` to use `storageutil.ShouldRetryOnMount` when `enableRetriesOnMount` is active, enabling autonomous self-healing for `404 NotFound` (missing bucket) and `403 PermissionDenied` errors during mount initialization.
- Restricted mount retries strictly to the Control Plane (`GetStorageLayout`) during mount setup, leaving Data Plane runtime I/O completely untouched and safe from infinite retry loops.

N/A

### Testing details
1. Manual - Verified that Data Plane runtime operations remain unchanged and fail fast on 404/non-retryable errors.
2. Unit tests - Added unit tests in `internal/storage/control_client_wrapper_test.go`:
   - `TestGetStorageLayout_MountRetriesEnabled_Retry404ThenSuccess`: Verifies that `GetStorageLayout` successfully retries transient 404 errors during initialization and succeeds.
   - `TestGetStorageLayout_MountRetriesDisabled_404FailsImmediately`: Verifies that `GetStorageLayout` fails immediately on 404 errors when the feature is disabled.
   - `TestGetStorageLayout_MountRetriesEnabled_AllAttemptsTimeOut`: Verifies correct propagation of deadline exceeded errors when all retry attempts time out.
3. Integration tests - N/A

No

### DIFF
--- a/internal/storage/control_client_wrapper.go
+++ b/internal/storage/control_client_wrapper.go
@@ -100,6 +100,8 @@ type storageControlClientWithRetry struct {
 	enableRetriesOnStorageLayoutAPI bool
 	// Whether or not to enable retries for folder APIs.
 	enableRetriesOnFolderAPIs bool
+	// Whether or not to enable mount retries for GetStorageLayout call.
+	enableRetriesOnMount bool
 }
 
 func (sccwros *storageControlClientWithRetry) GetStorageLayout(ctx context.Context,
@@ -113,6 +115,9 @@ func (sccwros *storageControlClientWithRetry) GetStorageLayout(ctx context.Conte
 		return sccwros.raw.GetStorageLayout(attemptCtx, req, opts...)
 	}
 
+	if sccwros.enableRetriesOnMount {
+		return storageutil.ExecuteWithCustomShouldRetryAtLogLevel(ctx, sccwros.retryConfig, "GetStorageLayout", req.Name, req.RequestId, apiCall, storageutil.ShouldRetryOnMount, logger.LevelInfo)
+	}
 	return storageutil.ExecuteWithRetryAtLogLevel(ctx, sccwros.retryConfig, "GetStorageLayout", req.Name, req.RequestId, apiCall, logger.LevelInfo)
 }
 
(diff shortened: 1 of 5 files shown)


internal/storage/control_client_wrapper.go
```diff
@@ -113,7 +115,12 @@ func (sccwros *storageControlClientWithRetry) GetStorageLayout(ctx context.Conte
 		return sccwros.raw.GetStorageLayout(attemptCtx, req, opts...)
 	}
 
-	return storageutil.ExecuteWithRetryAtLogLevel(ctx, sccwros.retryConfig, "GetStorageLayout", req.Name, req.RequestId, apiCall, logger.LevelInfo)
+	shouldRetry := storageutil.ShouldRetryWithoutLogging
+	if sccwros.enableMountRetries {
+		shouldRetry = storageutil.ShouldRetryOnMount
+	}
+
+	return storageutil.ExecuteWithCustomShouldRetryAtLogLevel(ctx, sccwros.retryConfig, "GetStorageLayout", req.Name, req.RequestId, apiCall, shouldRetry, logger.LevelInfo)
```
meet2mky: Nit: Can we simplify this a bit
```suggestion
	if sccwros.enableRetriesOnMountOperations {
		return storageutil.ExecuteWithCustomShouldRetryAtLogLevel(ctx, sccwros.retryConfig, "GetStorageLayout", req.Name, req.RequestId, apiCall, storageutil.ShouldRetryOnMount, logger.LevelInfo)
	}
	return storageutil.ExecuteWithRetryAtLogLevel(ctx, sccwros.retryConfig, "GetStorageLayout", req.Name, req.RequestId, apiCall, logger.LevelInfo)
```

internal/storage/control_client_wrapper.go
```diff
@@ -100,6 +100,8 @@ type storageControlClientWithRetry struct {
 	enableRetriesOnStorageLayoutAPI bool
 	// Whether or not to enable retries for folder APIs.
 	enableRetriesOnFolderAPIs bool
+	// Whether or not to enable mount retries for GetStorageLayout call.
+	enableMountRetries bool
```
meet2mky: Nit: keeping it consistent with above variables.
```suggestion
	enableRetriesOnMount bool
```

internal/storage/control_client_wrapper.go
meet2mky, line 227: Can you add a method here named **_withRetryOnMount_** here and use that?

dangarhs: Is it for separating the startup mount retries with GetStorageLayout calls?

internal/storage/control_client_wrapper.go
dangarhs: Ack.

internal/storage/control_client_wrapper.go
meet2mky: I Think, this is how existing retries are working.

PR #4849, opened 2026-07-07
feat(cache): add LRU mechanics to Radix Trie

vedantdas-source:
### Description
This PR embeds a doubly-linked list directly into the `radixNode` structure and introduces the core LRU management functions: `pushFront`, `moveToFront`, `remove`, and `evictOne`.

Instead of maintaining a separate list data structure (like the standard `container/list`) , the `prev`and `next` pointers are incorporated directly into the `radixNode` struct. This "intrusive" design completely eliminates the need for separate list-element allocations when tracking cache eviction order, saving memory and GC pressure.

`evictOne()` performs `O(1)` removal of the tail node from both the linked list and the core tree structure in constant time.

### Testing details
1. Manual - NA
2. Unit tests - Appended unit tests to `radix_test.go` to test `pushFront`, `moveToFront`, `remove`, and `evictOne` pointer mutations.
3. Integration tests - NA

### DIFF
--- a/internal/cache/lru/radix.go
+++ b/internal/cache/lru/radix.go
@@ -26,6 +26,9 @@ type radixNode struct {
 	parent  *radixNode
 	child   *radixNode
 	sibling *radixNode
+	// LRU Linked List pointers
+	prev *radixNode
+	next *radixNode
 }
 
 // this radixTree struct will be replaced by a RadixCache struct afterwards with additional necessary fields
@@ -33,6 +36,10 @@ type radixNode struct {
 type radixTree struct {
 	root *radixNode
 	size int
+	// Head and tail of the LRU Doubly Linked List
+	head *radixNode
+	tail *radixNode
+	len  int
 }
 
 func newRadixTree() *radixTree {
@@ -190,6 +197,7 @@ func (t *radixTree) deleteNode(node *radixNode) {
 	if node == nil || node.value == nil {
 		return
 	}
+	t.remove(node) // Unlink from LRU list
 
 	node.value = nil
 	t.size--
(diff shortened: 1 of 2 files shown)


internal/cache/lru/radix.go
```diff
@@ -228,3 +235,69 @@ func (t *radixTree) compressPathUpwards(curr *radixNode) {
...
+	node := t.tail
+	if node == nil {
+		return nil
+	}
+	evictedEntry := node.value
+	// Remove from LRU list and then fully delete from the tree
+	t.remove(node)
+	t.deleteNode(node)
+	return evictedEntry
+}
```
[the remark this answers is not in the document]
vedantdas-source: This was done to separate the core tree logic from the LRU logic. The Radix Cache wrapper struct that would be incorporated further on would have checked this in its `Erase` function. 
Still, to provide an additional safeguard and avoid dangling pointers `t.remove(node)` is now invoked in the `deleteNode` function

internal/cache/lru/radix.go
```diff
@@ -26,13 +26,20 @@ type radixNode struct {
...
 
 // this radixTree struct will be replaced by a RadixCache struct afterwards with additional necessary fields
 // radixTree encapsulates the core tree structure (LRU logic to be added in next PR).
 type radixTree struct {
 	root *radixNode
 	size int
+	// Head and tail of the LRU Doubly Linked List
+	head *radixNode
+	tail *radixNode
+	len  int
```
kislaykishore: It's unclear what's the difference between size and len. Please add documentation.

internal/cache/lru/radix.go
```diff
@@ -228,3 +236,72 @@ func (t *radixTree) compressPathUpwards(curr *radixNode) {
...
+}
+
+func (t *radixTree) evictOne() ValueType {
+	node := t.tail
+	if node == nil {
+		return nil
+	}
+	evictedEntry := node.value
+	// Remove from LRU list and then fully delete from the tree
+	t.remove(node)
```
kislaykishore, line 306: Why is this call needed when `t.deleteNode(node)` internally calls `remove(node)`?

internal/cache/lru/radix.go
```diff
@@ -228,3 +236,72 @@ func (t *radixTree) compressPathUpwards(curr *radixNode) {
...
+		t.tail = node.prev
+	}
+	node.prev = nil
+	node.next = t.head
+	if t.head != nil {
+		t.head.prev = node
+	}
+	t.head = node
+	if t.tail == nil {
+		t.tail = node
```
kislaykishore: Is this statement dead-code? If not, could you give me an example where this code-path gets hit?

internal/cache/lru/radix_test.go
```diff
@@ -128,3 +128,156 @@ func TestRadixTree_DeleteAndCompress(t *testing.T) {
 	assert.Equal(t, "foo/baz", n2.prefix)
 	assert.Equal(t, tree.root, n2.parent) // Path compressed all the way up to the root!
 }
+
+func TestRadixTree_LRU_PushFront(t *testing.T) {
+	// Arrange
```
kislaykishore: Please remove the "// Arrange", "// Act" and "// Assert" tags.

internal/cache/lru/radix.go
vedantdas-source: I will be removing the `remove(node)` call from the `deleteNode` function afterwards because the Radix Cache wrapper function `Erase` would take care of that, hence I kept it there and I wanted to separate core Trie operations from LRU operations but I changed it to resolve [this comment](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4849#discussion_r3534501558) . will remove the `remove(node)` call from `deleteNode`

internal/cache/lru/radix.go
vedantdas-source: This `if t.tail == nil` statement will be necessary when the entire LRU is empty. So when there are no LRU pointers (`t.head` and `t.tail` point to nothing) that will be the case when this would be required as we correctly update `t.head = node` but if there are no entries `t.tail` must also point to `node` .

kislaykishore: But if the LRU Cache is empty, there won't be a node that could be moved to the front, right?

vedantdas-source: Ok, that was my mistake. I copied the `tail` check from `pushFront` so I got a little confused. Removed the dead code.

internal/cache/lru/radix.go
vedantdas-source: Done, Added comments regarding the same. 
- `size`: Tracks the number of value-bearing nodes currently in the Radix Tree structure.
- `len`: Tracks the number of nodes currently chained together in LRU list.

In a perfect cache, these two numbers will always be exactly equal. Hence I was planning to use this as a invariant check.

This is different than the `currentSize` and `maxSize` fields that would be added further on.

kislaykishore (approved):

PR #4850, opened 2026-07-07
feat(kernel range reader): Add buffer pooling support for large reads on regional buckets

abhishek10004:
### Description
# Description

This PR introduces buffer pooling for large and vectored read operations to optimize memory usage and reduce garbage collection overhead. By reusing allocated buffers via a sync pool, we minimize memory churn during sequential and large file reads.

> [!NOTE]
> Buffer pooling support for larger reads is only supported for **regional buckets** when **kernel reader** is enabled.

## Key Changes

* **Buffer Pooling & Lifecycle Management (`internal/fs/fs.go`)**
  * Integrated `readBufferPool` into read operations to dynamically allocate and recycle byte buffers for read requests (supported only for regional buckets with kernel reader enabled).
  * Implemented immediate buffer reclamation: unused buffers or zero-copy hits return their buffers to the pool right away.
  * Added callback handling to ensure borrowed buffers are properly returned to the pool after the caller finishes consuming the read data.

* **Buffer Limiting Utility (`internal/util/util.go`)**
  * Added a zero-allocation helper function `LimitBuffers(buffers [][]byte, limit int) [][]byte`.
  * Reslices multi-buffer slices in-place to cap total data size without allocating new slice headers or arrays.

* **Testing (`internal/fs/large_read_regional_test.go`, `internal/util/util_test.go`)**
  * Added unit tests for `LimitBuffers` covering boundary conditions, multiple buffer spans, and edge cases.
  * Introduced regional large read test suite (`large_read_regional_test.go`) verifying successful vectored reads on regional buckets and confirming proper fallback/error behavior (`ENOTSUP`) for zonal buckets or when kernel reader support is disabled.

b/530765312

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -27,6 +27,7 @@ import (
 	"reflect"
 	"slices"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -60,6 +61,9 @@ import (
 	"github.com/jacobsa/timeutil"
 	"github.com/spf13/viper"
 )
+
+// readPoolBufferSize is the size of each buffer in readBufferPool (1 MiB).
+const readPoolBufferSize = util.MiB
 
 type ServerConfig struct {
 	// A clock used for cache expiration. It is *not* used for inode times, for
@@ -225,6 +229,11 @@ func NewFileSystem(ctx context.Context, serverCfg *ServerConfig) (fuseutil.FileS
 		globalMaxWriteBlocksSem:    semaphore.NewWeighted(serverCfg.NewConfig.Write.GlobalMaxBlocks),
 		globalMaxReadBlocksSem:     semaphore.NewWeighted(serverCfg.NewConfig.Read.GlobalMaxBlocks),
 		globalMetadataPrefetchSem:  semaphore.NewWeighted(serverCfg.NewConfig.MetadataCache.MetadataPrefetchMaxWorkers),
+		readBufferPool: sync.Pool{
+			New: func() any {
+				return new([readPoolBufferSize]byte)
+			},
+		},
 	}
 
 	// Initialize MRD cache if enabled
@@ -676,6 +685,9 @@ type fileSystem struct {
 
 	// mrdCache manages the cache of inactive MultiRangeDownloaders.
 	mrdCache *lru.Cache
+
+	// readBufferPool is a sync.Pool of readPoolBufferSize (1 MiB) buffers.
+	readBufferPool sync.Pool
 }
 
 ////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


internal/fs/fs.go
```diff
@@ -3118,23 +3130,74 @@ func (fs *fileSystem) ReadFile(
...
-			Buffer: op.Dst,
-			Offset: op.Offset,
-		}
-		resp, err = fh.ReadWithKernelReader(ctx, req)
+		resp, err = fh.ReadWithKernelReader(ctx, &req)
 		op.BytesRead = resp.Size
-		op.Data = resp.Data
-		op.Callback = resp.Callback
+		if useReadBufferPool {
+			if len(resp.Data) > 0 {
```
vadlakondaswetha, line 3168: why all this logic. why can't the read create buffers based on the need.

abhishek10004: I'm not very clear on the ask here.

IIUC, you're asking why do we have this buffer pool here instead of creating the buffers in kernel range reader. There are a couple of reasons for that.
1. If we have to read from inode (in case there are local changes), then we need to allocate buffers for that which would be outside the kernel range reader implementation.
2. In future, if we extend this read capability to other read paths as well, having the buffer pool at this layer would readily allow us to do that.

If you're asking about the additional handling on the returned response: Yes, I'm doing additional handling here (since currently kernel range reader does not allocate buffers on its own or sets any callback) but that is being done for future proofing the code. The ReadResponse struct says readers can return data directly from their internal buffers via Data field or can set a callback. We're not doing that currently but I've added the logic so that in case something changes in the readers (let's say we can get buffers from upper layers etc.), it would be handled automatically here without any additional changes.

Let me know if you meant something else.

internal/fs/fs.go
```diff
@@ -3118,23 +3130,74 @@ func (fs *fileSystem) ReadFile(
...
+
+	useReadBufferPool := op.Size > 0 && op.Dst == nil
+	if useReadBufferPool {
+		if !fs.newConfig.FileSystem.EnableKernelReader || fh.Inode().Bucket().BucketType().IsRapid() {
+			logger.Errorf("ReadFile: buffer pool allocation is only supported for regional buckets with"+
+				" kernel reader enabled (EnableKernelReader: %v, IsRapid: %v)",
+				fs.newConfig.FileSystem.EnableKernelReader,
+				fh.Inode().Bucket().BucketType().IsRapid())
+			fh.Inode().Unlock()
+			return syscall.ENOTSUP
```
raj-prince, line 3146: Just to confirm, customer will never face this scenario. Given we are enabling the MaxRead only in case of regional bucket & kernel reader.

abhishek10004:
Closing this PR as we're moving towards lazy allocation based approach for the buffers - https://github.com/GoogleCloudPlatform/gcsfuse/pull/4852

PR #4854, opened 2026-07-08
fix(streaming writes): change log severity for streaming writes fallback from INFO to TRACE

anushka567:
### Description
This PR addresses an issue where GCS Fuse was emitting excessive logs when falling back to legacy staged writes from streaming writes. The `INFO` and `WARN` level logs were spamming user logs (e.g., when writing to existing files, performing out-of-order writes, or reaching concurrent write limits). 

To prevent this log spam and improve the product experience, the severity of these fallback logs has been reduced to `TRACE` level. This ensures the information remains available for debugging purposes when explicitly requested, without polluting the default log output. Additionally, proper metrics are already in place that record the counts and the specific reasons for the fallback, ensuring that no observability information is lost despite the reduction in log verbosity. Documentation has also been updated accordingly. 

[b/528701083](https://buganizer.corp.google.com/issues/528701083)
### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA

None. Only logging severity is changed.

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -755,7 +755,7 @@ func (f *FileInode) writeUsingBufferedWrites(ctx context.Context, data []byte, o
 	}
 	// Fall back to temp file for Out-Of-Order Writes.
 	if errors.Is(err, bufferedwrites.ErrOutOfOrderWrite) {
-		logger.Infof("Out of order write detected. File %s will now use legacy staged writes. "+StreamingWritesSemantics, f.name.String())
+		logger.Tracef("Out of order write detected. File %s will now use legacy staged writes. "+StreamingWritesSemantics, f.name.String())
 		// Finalize the object.
 		err = f.flushUsingBufferedWriteHandler(ctx)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


internal/fs/inode/file.go
```diff
@@ -1237,7 +1237,7 @@ func (f *FileInode) InitBufferedWriteHandlerIfEligible(ctx context.Context, open
 			TraceHandle:              f.traceHandle,
 		})
 		if errors.Is(err, block.CantAllocateAnyBlockError) {
-			logger.Warnf("File %s will use legacy staged writes because concurrent streaming write "+
```
meet2mky, line 1240: Do you want to move this to trace as well? This would prevent us from knowing if the streaming write blocks limit is being hit for customer workloads.

CC: @vadlakondaswetha

anushka567: FYI, the fallback metric count also records the reason for fallback.

meet2mky: Ack!

meet2mky (approved):
LGTM

thrivikram-karur-g (approved):
lgtm

thommahoney-google (approved):

anushka567:
Merging since playbook for staging writes investigation based on server side logs is present.

kislaykishore (approved):

PR #4855, opened 2026-07-08
test(gke_machine_type): pass test_config.yaml to flag_optimizations in GKE machine_type_test

anushka567:
### Description
Resolves fatal errors caused by missing `FlagOptimizations` configurations after the removal of hardcoded test fallback configs in a recent commit. Previously, `setup_test.go` would fall back to manual configs, but it now strictly requires `test_config.yaml` to be passed.
[b/532193472](b/532193472)
### Testing details
1. Manual - Yes.
2. Unit tests - NA
3. Integration tests - NA


meet2mky (approved):
LGTM

PR #4856, opened 2026-07-08
feat(cache): implement Radix cache functionality

vedantdas-source:
### Description
This PR introduces the final unified `radixCache` implementation for the internal cache module. It incorporates the previously built LCRS (Left-Child Right-Sibling) tree and LRU list into a single struct that implements the `lru.Cache` interface supporting the following functions :
```
	Insert(key string, value ValueType) ([]ValueType, error)
	Erase(key string) (value ValueType)
	LookUp(key string) (value ValueType)
	LookUpWithoutChangingOrder(key string) (value ValueType)
	UpdateWithoutChangingOrder(key string, value ValueType) error
	UpdateSize(key string, sizeDelta uint64) error
	EraseEntriesWithGivenPrefix(prefix string)
```

By avoiding wrapper structs, we minimize memory allocations .

### Testing details
1. Manual - NA
2. Unit tests -  Added tests in `radix_test.go` to cover various scenarios:
     - Inserting new entries and handling nil values.
    - Retrieving, overwriting, and checking for non-existent nodes
    - Mutating LRU list order and severing specific nodes from the cache
    - Enforcing global cache capacity bounds and strict LRU eviction policies
    - Performing passive size updates and lookups without mutating the LRU order
    - Ensuring thread safety across highly concurrent reads, writes, and evictions to prevent race conditions
4. Integration tests - NA


internal/cache/lru/radix.go
```diff
@@ -46,11 +62,81 @@ type radixTree struct {
...
+		maxSize: maxSize,
+		root:    &radixNode{},
+	}
+	c.mu = locker.NewRW("RadixCache", c.checkInvariants)
+	return c
+}
+
+func (c *radixCache) checkInvariants() {
+	// INVARIANT: maxSize > 0
+	if c.maxSize <= 0 {
```
kislaykishore: maxSize is a uint64 - it will never be negative, right?

internal/cache/lru/radix.go
```diff
@@ -31,9 +34,22 @@ type radixNode struct {
...
-// radixTree encapsulates the core tree structure (LRU logic to be added in next PR).
-type radixTree struct {
+// radixCache encapsulates the core tree structure and implements the lru.Cache interface.
+type radixCache struct {
+	/////////////////////////
+	// Constant data
+	/////////////////////////
+
+	// INVARIANT: maxSize > 0
+	maxSize uint64
```
kislaykishore, line 44: Let's use int64. uint64 should be used sparingly - [go/totw/159](https://goto.google.com/totw/159)

internal/cache/lru/radix.go
kislaykishore: Why are we not failing the construction if `c.maxSize == 0`?

internal/cache/lru/radix.go
```diff
@@ -253,57 +339,279 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	_, ok := c.getNode(key)
+	if !ok {
+		return ErrEntryNotExist
+	}
+
+	// Update currentSize accounting
+	// Note: This may temporarily violate currentSize <= maxSize invariant
```
kislaykishore: Wouldn't this mean that the `checkInvariants` call would panic?

internal/cache/lru/radix.go
```diff
@@ -46,11 +62,81 @@ type radixTree struct {
 	// len is the number of nodes currently linked in the LRU list.
 	// In a properly functioning cache, size and len must always be equal.
 	len int
```
kislaykishore, line 60: I'd recommend against keeping len as well as size since they represent the same conceptual quantity.

internal/cache/lru/radix.go
```diff
@@ -253,57 +339,279 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+
+	node.value = value
+	return nil
+}
+
+// UpdateSize updates the currentSize accounting when an entry's size has changed.
+// This is needed for entries whose size grows incrementally (e.g., sparse files).
+// Eviction is deferred until the next Insert() call.
+// The entry's order in the LRU is not changed.
+func (c *radixCache) UpdateSize(key string, sizeDelta uint64) error {
```
kislaykishore, line 507: Do we have tests for this method?

internal/cache/lru/radix.go
```diff
@@ -253,57 +339,279 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+	}
+
+	// If search == 0 at the start, the prefix is exact to 'node'
+	if node != c.root {
+		parent := node.parent
+		parent.removeChild(node)
+		c.sweepAndUnlink(node)
+		c.compressPathUpwards(parent)
+	}
+}
```
kislaykishore: Could you check if the `if node != c.root` code block can be removed entirely with the following change?
```suggestion
		if lcp == len(search) {
			// We found the exact node where the prefix ends.
			// Sever it entirely from the tree structure
			node.removeChild(child)

			// Now sweep the detached subtree to fix LRU and currentSize
			c.sweepAndUnlink(child)
			c.compressPathUpwards(node)
			return
		}
		if lcp == len(child.prefix) {
			search = search[len(child.prefix):]
			node = child
			continue
		}
		return
	}
}
```

internal/cache/lru/radix.go
vedantdas-source: The `gcs.MinObject.Size`, `gcs.Object.Size`, and `FileInfo` are all strictly typed as `uint64` . If we change the LRU cache to use `int64`, we have to change the `ValueType.Size()` interface signature to `int64` as well. This would require us to scatter the `int64() ` type casts across a lot of unrelated files just to avoid compilation problems.

internal/cache/lru/radix.go
vedantdas-source: Yes , removed this invariant check and instead introduced a check for that while instantiating the cache.

vedantdas-source: Added a `maxsize == 0` check in the `NewRadixCache` method

internal/cache/lru/radix.go
vedantdas-source: Done, I've removed the size field from the Cache and updated all the functions that used it.

internal/cache/lru/radix.go
vedantdas-source: Added tests for `UpdateSize` method.

internal/cache/lru/radix.go
vedantdas-source: this is the same bug from the original map cache . I've added the eviction logic to `UpdateSize` to resolve it, and I'll make the corresponding fix in the map cache as well.

internal/cache/lru/radix.go
vedantdas-source: Absolutely possible. Evaluating `lcp == len(search)` first catches it immediately and deletes the child without bubbling out of the loop. This completely eliminates the need for the trailing block. Applied the changes.

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,274 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+
+func (c *radixCache) EraseEntriesWithGivenPrefix(prefix string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	//this check was done to comply with the existing behaviour of the EraseEntriesWithGivenPrefix function
+	//the mapCache uses strings.HasPrefix which returns true for any string compared with ""
+	if prefix == "" {
+		c.sweepAndUnlink(c.root)
```
kislaykishore: Is sweepAndUnlink really necessary? Since we just need to clear the cache, how about we just reset the cache state (an O(1) operation) and let Golang GC handle the collection?

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,274 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+// The entry's order in the LRU is not changed.
+func (c *radixCache) UpdateSize(key string, sizeDelta uint64) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	_, ok := c.getNode(key)
+	if !ok {
+		return ErrEntryNotExist
+	}
+
```
kislaykishore, line 515:
```suggestion
if node.value.Size()+sizeDelta > c.maxSize {
    return ErrInvalidEntrySize
}
```

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,274 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
 }
+
+////////////////////////////////////////////////////////////////////////
+// Cache interface
+////////////////////////////////////////////////////////////////////////
+
+// Insert the supplied value into the cache, overwriting any previous entry for
+// the given key. The value must be non-nil.
+// Also returns a slice of ValueType evicted by the new inserted entry.
+func (c *radixCache) Insert(key string, value ValueType) ([]ValueType, error) {
```
kislaykishore, line 390: Add test coverage for Insert("", val)

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,388 @@
...
-		tree.moveToFront(node2)
-
-		assert.Equal(t, node2, tree.head)
-		assert.Equal(t, node1, tree.tail)
-		assert.Nil(t, node2.prev)
-		assert.Equal(t, node3, node2.next)
-		assert.Equal(t, node2, node3.prev)
-		assert.Equal(t, node1, node3.next)
-		assert.Equal(t, node3, node1.prev)
+	t.Run("DeferredEviction", func(t *testing.T) {
```
kislaykishore: This isn't deferred eviction, right? It evicts immediately.

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,274 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+		oldNode.value = value
+		c.moveToFront(oldNode)
+	} else {
+		node, _ := c.insertNode(key, value)
+		c.pushFront(node)
+		c.currentSize += valueSize
+	}
+
+	var evictedValues []ValueType
+	//Evict until we're at or below maxSize
```
kislaykishore:
```suggestion
	// Evict until we're at or below maxSize
```

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,274 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+
+	return nil
+}
+
+func (c *radixCache) EraseEntriesWithGivenPrefix(prefix string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	//this check was done to comply with the existing behaviour of the EraseEntriesWithGivenPrefix function
+	//the mapCache uses strings.HasPrefix which returns true for any string compared with ""
```
kislaykishore: Add leading space
```suggestion
	// this check was done to comply with the existing behaviour of the EraseEntriesWithGivenPrefix function
	// the mapCache uses strings.HasPrefix which returns true for any string compared with ""
```

internal/cache/lru/radix_test.go
vedantdas-source: Correct. I changed the `UpdateSize` method but forgot to change the test name and logic to reflect this fix. Done.

internal/cache/lru/radix.go
vedantdas-source: Done. Made the check before updating `currentSize`

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,281 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+		c.currentSize -= n.value.Size()
+		c.remove(n)
+		n.value = nil
+	}
+
+	// Recurse through children
+	var nextChild *radixNode
+	for curr := n.child; curr != nil; curr = nextChild {
+		nextChild = curr.sibling
+		c.sweepAndUnlink(curr)
```
kislaykishore: Avoid tail-recursion.

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,281 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+
+	// Update accounting
+	c.currentSize -= evictedEntry.Size()
+
 	// Remove from LRU list and then fully delete from the tree
-	t.remove(node)
-	t.deleteNode(node)
+	c.remove(node)
+	c.deleteNode(node)
 	return evictedEntry
```
kislaykishore: Go [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)
```suggestion
	node := c.tail
if node == nil {
return nil
}
return c.eraseInternal(node)
```

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,395 @@
...
+	// inserting again which should evict key1 because key1 is looked up without
+	// changing order
+	key3 := "burrito3"
+	data3 := testData{Value: 3, DataSize: 5}
+	insertAndAssert(t, cache, key3, data3, []int64{23}, nil)
+}
+
+// This will detect race if we run the test with `-race` flag.
+// We get the race condition failure if we remove lock from Insert or Erase method.
+func TestRadixCache_RaceCondition(t *testing.T) {
```
kislaykishore, line 392: Please include `UpdateSize` and `EraseEntriesWithGivenPrefix` methods in this test.

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,281 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+	}
+
+	if node.value.Size()+sizeDelta > c.maxSize {
+		return ErrInvalidEntrySize
+	}
+
+	// Update currentSize accounting
+	c.currentSize += sizeDelta
+
+	//Evict until we're at or below maxSize to maintain invariants
```
kislaykishore: Please add a space after "//" here and elsewhere.
```suggestion
	// Evict until we're at or below maxSize to maintain invariants
```

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,281 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+		c.moveToFront(oldNode)
+	} else {
+		node, _ := c.insertNode(key, value)
+		c.pushFront(node)
+		c.currentSize += valueSize
+	}
+
+	var evictedValues []ValueType
+	// Evict until we're at or below maxSize
+	for c.currentSize > c.maxSize {
```
kislaykishore: Coding defensively in case `c.currentSize` computation goes haywire.
```suggestion
	for c.currentSize > c.maxSize && c.tail != nil {
```

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,281 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+
+	if node.value.Size()+sizeDelta > c.maxSize {
+		return ErrInvalidEntrySize
+	}
+
+	// Update currentSize accounting
+	c.currentSize += sizeDelta
+
+	//Evict until we're at or below maxSize to maintain invariants
+	for c.currentSize > c.maxSize {
```
kislaykishore: Code 
```suggestion
	for c.currentSize > c.maxSize && c.tail != nil {
```

internal/cache/lru/radix_test.go
vedantdas-source: Done.

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,281 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+
+// LookUp a previously-inserted value for the given key. Return nil if no
+// value is present.
+func (c *radixCache) LookUp(key string) (value ValueType) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	node, ok := c.getNode(key)
+	if !ok {
+		return
```
kislaykishore: Be consistent
```suggestion
		return nil
```

internal/cache/lru/radix.go
vedantdas-source: Ok we will use a stack for the DFS instead of recursion. Done

internal/cache/lru/radix.go
vedantdas-source: Done

internal/cache/lru/radix.go
kislaykishore: How about this:
```
func (c *radixCache) sweepAndUnlink(node *radixNode) {
	curr := node
	for curr != nil {
		// Clean up LRU list and size accounting for the current node
		if curr.value != nil {
			c.currentSize -= curr.value.Size()
			c.remove(curr)
			curr.value = nil
		}
		// Advance to next node in pre-order traversal
		if curr.child != nil {
			curr = curr.child
		} else {
			// Backtrack up parent chain until we find an ancestor with an unvisited sibling,
			// or we reach the root of the detached subtree.
			for curr != node && curr.sibling == nil {
				curr = curr.parent
			}
			if curr == node {
				break
			}
			curr = curr.sibling
		}
	}
}
```

kislaykishore: A slightly less complex implementation:
```
func (c *radixCache) sweepAndUnlink(node *radixNode) {
	curr := node
	for curr != nil {
		// Clean up LRU list and size accounting for the current node
		if curr.value != nil {
			c.currentSize -= curr.value.Size()
			c.remove(curr)
			curr.value = nil
		}
		// Advance to next node in pre-order traversal
		if curr.child != nil {
			curr = curr.child
                        continue
		}
		// Backtrack up parent chain until we find an ancestor with an unvisited sibling,
		// or we reach the root of the detached subtree.
		for curr != node && curr.sibling == nil {
			curr = curr.parent
		}
		if curr == node {
			break
		}
               curr = curr.sibling
	}
}
```

vedantdas-source: Ok, this would be a better iterative DFS approach as it would eliminate the memory that I was using for the stack. Thanks !

internal/cache/lru/radix.go
```diff
@@ -253,57 +322,280 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+	}
+
+	oldNode, exists := c.getNode(key)
+	if exists {
+		c.currentSize -= oldNode.value.Size()
+		c.currentSize += valueSize
+		oldNode.value = value
+		c.moveToFront(oldNode)
+	} else {
+		node, _ := c.insertNode(key, value)
```
kislaykishore: Why are we ignoring the error?

kislaykishore: We already did some pre-work when we invoked the `c.getNode(key)` to traverse the Radix tree. Calling c.insertNode would do a second traversal. Given that we are concerned about performance, would it make sense to reuse the traversal. I guess, you could refactor `insertNode` to perform lookup and insertion simultaneously.

internal/cache/lru/radix.go
```diff
@@ -99,20 +175,17 @@ func (n *radixNode) removeChild(childToRemove *radixNode) {
 }
 
 // insertNode inserts a new key into the radix tree and returns the leaf node.
-func (t *radixTree) insertNode(key string, value ValueType) (*radixNode, bool) {
+func (c *radixCache) insertNode(key string, value ValueType) (*radixNode, bool) {
```
kislaykishore: There is code-duplication in the sibling-replacement logic between `insertNode` and `compressPathUpwards`. Perhaps, it can be refactored out as a separate method.

internal/cache/lru/radix.go
vedantdas-source: We return a boolean indicating if a new node has been created or has been overwritten. As we insert this node inside the `else` block after checking if the node exists or not, we are sure it would return true.
I think it will be better to remove this return value from the function signature because we are not really using it anywhere else and we already make checks before using `insertNode`

vedantdas-source: We can change the `insertNode` to return the `ValueType` interface so we can check for the existence of a node there and if it doesn't we can just insert and return a nil value to indicate there was no such node/entry present.

internal/cache/lru/radix.go
vedantdas-source: Added a helper function `replaceChild` to reuse that logic in both functions

internal/cache/lru/radix.go
```diff
@@ -98,23 +174,32 @@ func (n *radixNode) removeChild(childToRemove *radixNode) {
...
 }
 
+// replaceChild finds oldChild in the sibling linked-list and replaces it with newChild,
+// seamlessly preserving the rest of the sibling chain.
+func (n *radixNode) replaceChild(oldChild, newChild *radixNode) {
+	for pcurr := &n.child; *pcurr != nil; pcurr = &(*pcurr).sibling {
+		if *pcurr == oldChild {
+			newChild.sibling = oldChild.sibling
+			*pcurr = newChild
+			break
```
kislaykishore:
```suggestion
			return
```

internal/cache/lru/radix.go
```diff
@@ -253,57 +328,278 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+		return nil, ErrInvalidEntry
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	valueSize := value.Size()
+	if valueSize > c.maxSize {
+		return nil, ErrInvalidEntrySize
+	}
```
kislaykishore:
```suggestion
	valueSize := value.Size()
	if valueSize > c.maxSize {
		return nil, ErrInvalidEntrySize
	}
	c.mu.Lock()
	defer c.mu.Unlock()
```

internal/cache/lru/radix.go
```diff
@@ -253,57 +328,278 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+}
+
+// Erase any entry for the supplied key, also returns the value of erased key.
+func (c *radixCache) Erase(key string) (value ValueType) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	node, ok := c.getNode(key)
+	if !ok {
+		return
```
kislaykishore: Avoid naked returns here and elsewhere.

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,409 @@
...
+	assert.Equal(t, uint64(2), cache.LookUp("b").Size())
+}
 
-	t.Run("move head to front", func(t *testing.T) {
-		tree := newRadixTree()
-		node1, _ := tree.insertNode("foo/1", mockValue{10})
-		node2, _ := tree.insertNode("foo/2", mockValue{20})
-		tree.pushFront(node1)
-		tree.pushFront(node2)
+func TestRadixCache_EraseCacheWithGivenPrefixWithSomeEntriesEvictedDueToCacheSize(t *testing.T) {
```
kislaykishore, line 198: Please add a test for empty prefix deletion (effectively dropping the entire cache)

internal/cache/lru/update_size_bug_test.go
```diff
@@ -0,0 +1,70 @@
...
+	// 3. Notify mapCache of the growth.
+	// mapCache simply adds sizeDelta to c.currentSize without double counting.
+	err = mapCache.UpdateSize("file.txt", sizeDelta)
+	assert.NoError(t, err, "mapCache should succeed when updated size (100) <= maxSize (100)")
+
+	// 4. Notify radixCache of the growth.
+	// radixCache evaluates node.value.Size() (100) + sizeDelta (50) = 150 > maxSize (100).
+	// This double-counts sizeDelta and incorrectly rejects a valid update.
+	err = radixCache.UpdateSize("file.txt", sizeDelta)
+	assert.ErrorIs(t, err, lru.ErrInvalidEntrySize, "radixCache incorrectly fails due to double counting (100 + 50 = 150 > 100)")
```
kislaykishore: This test shows the disparity between the mapCache and radixCache due to a double counting bug. Please fix the bug and incorporate such a test in the test-suite.

vedantdas-source: I've already fixed this issue in a previous commit. I also updated the your relevant tests to match, and everything is passing successfully.

internal/cache/lru/radix.go
```diff
@@ -253,57 +328,278 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+	defer c.mu.Unlock()
+
+	node, ok := c.getNode(key)
+	if !ok {
+		return ErrEntryNotExist
+	}
+
+	if node.value.Size() > c.maxSize {
+		return ErrInvalidEntrySize
+	}
```
kislaykishore: `mapCache` doesn't do this check - are we violating the contract?

kislaykishore: Please add a test as well for this.

vedantdas-source: I added that check to return an error early on if there is an attempt to increase the size beyond the capacity. This could cause some unexpected behaviour for like a sparse downloader where they constantly keep updating the size. I've removed this check so we gracefully evict during the cleanup cycle

internal/cache/lru/radix.go
```diff
@@ -253,57 +328,274 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+
+	node, oldValue := c.insertNode(key, value)
+	if oldValue != nil {
+		c.currentSize -= oldValue.Size()
+		c.currentSize += valueSize
+		c.moveToFront(node)
+	} else {
+		c.pushFront(node)
+		c.currentSize += valueSize
+	}
```
kislaykishore:
```suggestion
	if node, oldValue := c.insertNode(key, value); oldValue != nil {
		c.currentSize += valueSize - oldValue.Size()
		c.moveToFront(node)
	} else {
		c.pushFront(node)
		c.currentSize += valueSize
	}
```

internal/cache/lru/radix.go
```diff
@@ -253,57 +328,274 @@ func (t *radixTree) moveToFront(node *radixNode) {
...
+			curr = curr.child
+			continue
+		}
+		// Backtrack up parent chain until we find an ancestor with an unvisited sibling,
+		// or we reach the root of the detached subtree.
+		for curr != node && curr.sibling == nil {
+			curr = curr.parent
+		}
+		if curr == node {
+			break
```
kislaykishore:
```suggestion
			return
```

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,472 @@
...
+	insertAndAssert(t, cache, "burrito", testData{Value: 23, DataSize: 4}, []int64{}, nil)
 
-		assert.Equal(t, 0, tree.len)
-		assert.Nil(t, tree.head)
-		assert.Nil(t, tree.tail)
-	})
+	deletedEntry := cache.Erase("taco")
+	assert.Nil(t, deletedEntry)
+
+	assert.Equal(t, int64(23), cache.LookUp("burrito").(testData).Value)
```
kislaykishore, line 223:
```suggestion
	deletedEntry := cache.Erase("taco")
	
	assert.Nil(t, deletedEntry)
	assert.Equal(t, int64(23), cache.LookUp("burrito").(testData).Value)
```

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,472 @@
...
-		tree := newRadixTree()
-		node1, _ := tree.insertNode("foo/1", mockValue{10})
-		node2, _ := tree.insertNode("foo/2", mockValue{20})
-		tree.pushFront(node1)
+	t.Run("Immediate Eviction", func(t *testing.T) {
+		cache := lru.NewRadixCache(100)
+		data1 := testData{Value: 1, DataSize: 10}
+		data2 := testData{Value: 2, DataSize: 70}
+
+		_, _ = cache.Insert("key1", data1)
```
kislaykishore, line 239:
```suggestion
		data2 := testData{Value: 2, DataSize: 70}
		_, _ = cache.Insert("key1", data1)
```

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,472 @@
...
-	node2, _ := tree.insertNode("foo/2", mockValue{20})
-	tree.pushFront(node1)
-	tree.pushFront(node2)
+func TestRadixCache_UpdateSize_ExceedsMaxSize(t *testing.T) {
+	cache := lru.NewRadixCache(100)
+
+	data := &testData{Value: 1, DataSize: 50}
+
+	_, err := cache.Insert("file.txt", data)
+	assert.NoError(t, err)
```
kislaykishore:
```suggestion
	require.NoError(t, err)
```

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,472 @@
...
-	node1, _ := tree.insertNode("foo/1", mockValue{10})
-	node2, _ := tree.insertNode("foo/2", mockValue{20})
-	tree.pushFront(node1)
-	tree.pushFront(node2)
+func TestRadixCache_UpdateSize_ExceedsMaxSize(t *testing.T) {
+	cache := lru.NewRadixCache(100)
+
+	data := &testData{Value: 1, DataSize: 50}
+
+	_, err := cache.Insert("file.txt", data)
```
kislaykishore, line 253:
```suggestion
	cache := lru.NewRadixCache(100)
	data := &testData{Value: 1, DataSize: 50}
	_, err := cache.Insert("file.txt", data)
```

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,472 @@
...
+	assert.NoError(t, err)
+
+	data.DataSize = 150
```
kislaykishore, line 256:
```suggestion
	data.DataSize = 150
```

internal/cache/lru/radix_test.go
```diff
@@ -12,227 +12,472 @@
...
+
+	err = cache.UpdateSize("file.txt", 100)
+	assert.NoError(t, err)
+
+	assert.Nil(t, cache.LookUp("file.txt"))
```
kislaykishore, line 260:
```suggestion
	err = cache.UpdateSize("file.txt", 100)
	
	assert.NoError(t, err)
	assert.Nil(t, cache.LookUp("file.txt"))
```

kislaykishore (commented):
Some nits

internal/cache/lru/radix.go
vedantdas-source: Done.

kislaykishore (approved):

PR #4857, opened 2026-07-09
perf: Optimize statCache expiration timestamp by using int64 instead of time.Time

vedantdas-source:
### Description
In the `StatCache` the `entry` struct tracks TTL expiration using a standard Go `time.Time` object. This `time.Time` struct consumes 24 bytes of memory which can be reduced to 8 bytes using `int64` for the expiration timestamp. This saves 16 bytes of memory per object. 

#### Changes Made
- Changed `expiration` field in the internal cache `entry` from `time.Time` to `int64`
- Cache insertion methods (`Insert`, `InsertImplicitDir`, `AddNegativeEntry`, etc.) now convert the incoming `expiration time.Time` parameter to seconds using `.Unix()`
- Lookup methods now evaluate expiration against `now.Unix()`

#### Benchmarking results 
Using `time.Time` : 
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/cache/metadata
BenchmarkStatCache_Insert-48                     3033651               409.2 ns/op           272 B/op          4 allocs/op
BenchmarkStatCache_AddNegativeEntry-48           7874757               149.2 ns/op            80 B/op          2 allocs/op
BenchmarkStatCache_LookUp-48                    19632806                61.15 ns/op           17 B/op          0 allocs/op
```
Using `int64` : 
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/cache/metadata
BenchmarkStatCache_Insert-48                     3174098               382.8 ns/op           256 B/op          4 allocs/op
BenchmarkStatCache_AddNegativeEntry-48           8443263               141.3 ns/op            64 B/op          2 allocs/op
BenchmarkStatCache_LookUp-48                    22825635                54.30 ns/op           11 B/op          0 allocs/op
```

### Testing details
1. Manual - NA
2. Unit tests - Added `stat_cache_benchmark_test.go` to measure memory and CPU performance of `StatCache` operations.
3. Integration tests - NA


internal/cache/metadata/stat_cache.go
```diff
@@ -113,7 +113,7 @@ type statCacheBucketView struct {
 type entry struct {
 	m          *gcs.MinObject
 	f          *gcs.Folder
-	expiration time.Time
+	expiration int64 // expiration timestamp
```
raj-prince, line 116: Could you please check, what happens when `--stat-cache-ttl` is set to -1? I feel, expiration will get overflow and leads to quick eviction. 

In that case, I would suggest to create an utility method to handle corner scenario.

```go
func toUnixNano(t time.Time) int64 {
	if t.IsZero() {
		return math.MinInt64
	}
	// Dates beyond April 2262 (math.MaxInt64 / 1e9 seconds) overflow int64 nanoseconds.
	if t can overflow during conversion {
        return math.MaxInt64
	}
	return t.UnixNano()
}
```

raj-prince (changes requested):

internal/cache/metadata/stat_cache.go
vedantdas-source: Checked for `--stat-cache-ttl = -1`. It sets the `TtlSecs` to `maxSupportedTTLInSeconds = math.MaxInt64 / int64(time.Second)` which wraps around to become negative hence indicates that it has been expired. Added the utility function and used it for conversion

PR #4859, opened 2026-07-09
perf: Optimize folder UpdateTime timestamp by using int64 instead of time.Time 

vedantdas-source:
### Description
By default,`Folder.UpdateTime` was stored as a `time.Time` value, which occupies 24 bytes. By converting this field to a Unix nanosecond `int64` representation which uses 8 bytes, we can safely reduce the structural footprint of every `Folder` object .

#### Changes Made
- Changed `Folder.UpdateTime` from time.Time to `int64`.
- Added a zero-value check `!t.IsZero()` in `GCSFolder` to ensure that uninitialized folder timestamps default to 0 instead of a negative value.
- Updated the mock bucket generation and FUSE assertions to compare the equivalent `UnixNano()` values.

#### Benchmarking results
Using `time.Time`
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs
BenchmarkGCSFolder-48           12818931                94.74 ns/op      96 B/op          2 allocs/op
```

Using `int64`
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs
BenchmarkGCSFolder-48           14288154                87.04 ns/op      72 B/op          2 allocs/op
```

### Testing details
1. Manual - NA
2. Unit tests - Added a `BenchmarkGCSFolder` benchmark in `folder_test.go` to compare memory and CPU performance.
3. Integration tests - NA


internal/storage/gcs/folder.go
```diff
@@ -16,21 +16,25 @@ package gcs
...
 
 type Folder struct {
 	Name       string
-	UpdateTime time.Time
+	UpdateTime int64
 }
 
 func GCSFolder(bucketName string, attrs *controlpb.Folder) *Folder {
 	// Setting the parameters in Folder and doing conversions as necessary.
+	var updateTime int64
```
raj-prince, line 30: Lets also add a unit test, which ensures 0 UpdatedTime in case of un-initialized time.

raj-prince (commented):
LGTM.

internal/storage/fake/bucket.go
```diff
@@ -1205,6 +1205,9 @@ func (b *bucket) CreateFolder(ctx context.Context, folderName string) (*gcs.Fold
 }
 
 func (b *bucket) RenameFolder(ctx context.Context, folderName string, destinationFolderId string) (*gcs.Folder, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
```
meet2mky, line 1209: Why a

meet2mky, line 1209: Why are we taking b.mu.Lock() here?. It doesn't seem to the part of perf optimization.

internal/storage/gcs/folder.go
vedantdas-source: Done.

internal/storage/fake/bucket.go
vedantdas-source: Yes, It is not related to the perf optimization, but `RenameFolder` iterates over and mutates b.folders and b.objects without holding `b.mu.Lock()` (unlike `CreateFolder` and `DeleteFolder`, which do hold the lock) and this was flagged by the review bot, hence I made that change. Would you prefer this to be removed from this PR ?

meet2mky: It's fine. It's just looking at the PR I couldn't relate it immediately why was the change added.

vedantdas-source:
#4924

PR #4860, opened 2026-07-09
perf(mem): Optimize folder UpdateTime timestamp by using int64 instead of time.Time

vedantdas-source:
### Description
By default,`Folder.UpdateTime` was stored as a `time.Time` value, which occupies 24 bytes. By converting this field to a Unix nanosecond `int64` representation which uses 8 bytes, we can safely reduce the structural footprint of every `Folder` object .

#### Changes Made
- Changed `Folder.UpdateTime` from time.Time to `int64`.
- Added a zero-value check `!t.IsZero()` in `GCSFolder` to ensure that uninitialized folder timestamps default to 0 instead of a negative value.
- Updated the mock bucket generation and FUSE assertions to compare the equivalent `UnixNano()` values.

#### Benchmarking results
Using `time.Time`
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs
BenchmarkGCSFolder-48           12818931                94.74 ns/op      96 B/op          2 allocs/op
```

Using `int64`
```
goos: linux
goarch: amd64
pkg: github.com/googlecloudplatform/gcsfuse/v3/internal/storage/gcs
BenchmarkGCSFolder-48           14288154                87.04 ns/op      72 B/op          2 allocs/op
```

### Testing details
1. Manual - NA
2. Unit tests - Added a `BenchmarkGCSFolder` benchmark in `folder_test.go` to compare memory and CPU performance.
3. Integration tests - NA

### DIFF
--- a/internal/storage/gcs/folder.go
+++ b/internal/storage/gcs/folder.go
@@ -16,21 +16,25 @@ package gcs
 
 import (
 	"strings"
-	"time"
 
 	"cloud.google.com/go/storage/control/apiv2/controlpb"
 )
 
 type Folder struct {
 	Name       string
-	UpdateTime time.Time
+	UpdateTime int64
 }
 
 func GCSFolder(bucketName string, attrs *controlpb.Folder) *Folder {
 	// Setting the parameters in Folder and doing conversions as necessary.
+	var updateTime int64
+	if ts := attrs.GetUpdateTime(); ts != nil {
+		updateTime = ts.GetSeconds()*1e9 + int64(ts.GetNanos())
+	}
+
 	return &Folder{
 		Name:       getFolderName(bucketName, attrs.Name),
-		UpdateTime: attrs.GetUpdateTime().AsTime(),
+		UpdateTime: updateTime,
 	}
 }
 
--- a/internal/storage/gcs/folder_test.go
+++ b/internal/storage/gcs/folder_test.go
@@ -16,7 +16,6 @@ package gcs
 
 import (
 	"testing"
-	"time"
 
 	"cloud.google.com/go/storage/control/apiv2/controlpb"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


internal/storage/gcs/folder.go
```diff
@@ -16,21 +16,26 @@ package gcs
...
 }
 
 func GCSFolder(bucketName string, attrs *controlpb.Folder) *Folder {
 	// Setting the parameters in Folder and doing conversions as necessary.
+	var updateTime int64
+
+	t := attrs.GetUpdateTime().AsTime()
+	if !t.IsZero() {
+		updateTime = t.UnixNano()
+	}
```
kislaykishore, line 33: Instead of extracting the protobuf time: `GetUpdateTime()` and then converting it to Golang time, followed by converting it to nanoseconds, you could consider doing the following. Notice how the `AsTime()` method is implemented [here](https://github.com/protocolbuffers/protobuf-go/blob/96a179180f0ad6bba9b1e7b6e38d0affb0168e9a/types/known/timestamppb/timestamp.pb.go#L201)
```suggestion
	var updateTime int64
	if ts := attrs.GetUpdateTime(); ts != nil {
	  updateTime = ts.GetSeconds()*1e9 + int64(ts.GetNanos())
	}
```

internal/storage/gcs/folder_test.go
```diff
@@ -48,5 +48,33 @@ func TestGCSFolder(t *testing.T) {
...
+
+func TestGCSFolder_UninitializedTime(t *testing.T) {
+	attrs := controlpb.Folder{
+		Name:           TestFolderName,
+		Metageneration: 10,
+	}
+
+	gcsFolder := GCSFolder(TestBucketName, &attrs)
+
+	assert.Equal(t, int64(0), gcsFolder.UpdateTime)
```
kislaykishore: Nit - feel free to ignore
```suggestion
	assert.EqualValues(t, 0, gcsFolder.UpdateTime)
```

internal/storage/gcs/folder_test.go
```diff
@@ -48,5 +48,33 @@ func TestGCSFolder(t *testing.T) {
 	gcsFolder := GCSFolder(TestBucketName, &attrs)
 
 	assert.Equal(t, attrs.Name, gcsFolder.Name)
-	assert.Equal(t, attrs.UpdateTime.AsTime(), gcsFolder.UpdateTime)
+	assert.Equal(t, attrs.UpdateTime.AsTime().UnixNano(), gcsFolder.UpdateTime)
```
kislaykishore: Use literals in assertions: go/tott/741 e.g.
```
timestamp := &timestamppb.Timestamp {
    Seconds: 123456789,
     Nanos: 987654321,
 }
...
...
assert.Equal(t, 123234345,  gcsFolder.UpdateTime)
```

vedantdas-source: Done

raj-prince (approved):

PR #4862, opened 2026-07-10
perf: Optimize dirInode prevDirListingTimeStamp by using atomic.Int64 instead of time.Time 

vedantdas-source:
### Description
This PR optimizes the `dirInode` struct by converting the `prevDirListingTimeStamp` field from a `time.Time` struct into an `atomic.Int64` representing Unix nanoseconds. This reduces memory overhead per cached directory and eliminates a latent thread-safety data race during concurrent kernel cache TTL checks.

- **Latent Data Race Fix (Concurrency Safety)** :  Previously, `ShouldInvalidateKernelListCache` was performing concurrent, lock-free reads on the 24-byte `time.Time` struct. Because a 24-byte struct cannot be read in a single CPU instruction, this exposed the cache invalidation path to memory tearing (reading a partially-updated timestamp) during highly parallel FUSE directory lookups. By migrating to `atomic.Int64.Load()` and `.Store()`, we guarantee 100% thread-safe, lock-free reads at the hardware level without introducing any Mutex overhead.

- **Memory Optimization** : A time.Time struct consumes 24 bytes in Go. An `atomic.Int64` consumes 8 bytes. Because GCSFuse heavily caches directories, shrinking this auxiliary field removes exactly 16 bytes from every `dirInode` in memory (reducing its size from 344B to 328B). This results in a ~1.6 MB heap RAM savings per 100,000 cached directories.

All existing directory and inode tests pass successfully.
### Testing details
1. Manual - NA
2. Unit tests - Added a `Test_ShouldInvalidateKernelListCache_RaceCondition` to check for Race condition that is fixed by using `atomic.Int64`. To use the same test for the version of dirInode with `time.Time` change the `d.prevDirListingTimeStamp.Store(d.cacheClock.Now().UnixNano())`  to `d.prevDirListingTimeStamp = d.cacheClock.Now()`
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/dir.go
+++ b/internal/fs/inode/dir.go
@@ -265,7 +265,7 @@ type dirInode struct {
 	// (via kernel) the directory listing from the filesystem.
 	// Specially used when kernelListCacheTTL > 0 that means kernel list-cache is
 	// enabled.
-	prevDirListingTimeStamp time.Time
+	prevDirListingTimeStamp atomic.Int64
 
 	metadataCacheTtlSecs int64
 
@@ -634,7 +634,6 @@ func (d *dirInode) CancelCurrDirPrefetcher() {
 	if d.prefetcher != nil {
 		d.prefetcher.Cancel()
 	}
-	return
 }
 
 // UpdateSize is a no-op for implicit directories. These directories are not
@@ -1105,7 +1104,7 @@ func (d *dirInode) ReadEntryCores(ctx context.Context, tok string) (cores map[Na
 		return
 	}
 
-	d.prevDirListingTimeStamp = d.cacheClock.Now()
+	d.prevDirListingTimeStamp.Store(d.cacheClock.Now().UnixNano())
 	return
 }
 
@@ -1508,11 +1507,12 @@ func (d *dirInode) LocalFileEntries(localFileInodes map[Name]Inode) (localEntrie
 func (d *dirInode) ShouldInvalidateKernelListCache(ttl time.Duration) bool {
 	// prevDirListingTimeStamp.IsZero() true means listing has not happened yet, and we should
 	// invalidate for clean start.
-	if d.prevDirListingTimeStamp.IsZero() {
+	prevNS := d.prevDirListingTimeStamp.Load()
+	if prevNS == 0 {
 		return true
 	}
 
-	cachedDuration := d.cacheClock.Now().Sub(d.prevDirListingTimeStamp)
+	cachedDuration := time.Duration(d.cacheClock.Now().UnixNano() - prevNS)
 	return cachedDuration >= ttl
 }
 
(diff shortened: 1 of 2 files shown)


internal/fs/inode/dir.go
```diff
@@ -265,7 +265,7 @@ type dirInode struct {
 	// (via kernel) the directory listing from the filesystem.
 	// Specially used when kernelListCacheTTL > 0 that means kernel list-cache is
 	// enabled.
-	prevDirListingTimeStamp time.Time
+	prevDirListingTimeStamp atomic.Int64
```
kislaykishore, line 268: Are we fixing a concurrency bug that we are fixing here by using atomic integers? If so, could you also add a test that demonstrates the problem?

internal/fs/inode/dir_test.go
```diff
@@ -1253,7 +1253,7 @@ func (t *DirTest) TestReadEntryCores_NonEmpty_ImplicitDirsEnabled() {
 	t.validateCore(cores, "symlink", false, metadata.SymlinkType, symlinkName)
 	assert.ElementsMatch(t.T(), []string{dirInodeName + "../", dirInodeName + "/"}, unsupportedPaths)
 	// Make sure prevDirListingTimeStamp is initialized.
-	require.False(t.T(), d.prevDirListingTimeStamp.IsZero())
+	require.False(t.T(), d.prevDirListingTimeStamp.Load() == 0)
```
kislaykishore: Similar changes elsewhere.
```suggestion
	assert.NotZero(t.T(), d.prevDirListingTimeStamp.Load())
```

internal/fs/inode/dir_test.go
```diff
@@ -1238,7 +1238,7 @@ func (t *DirTest) TestReadEntryCores_NonEmpty_ImplicitDirsEnabled() {
 	// Nil prevDirListingTimeStamp
 	d := t.in.(*dirInode)
 	require.NotNil(t.T(), d)
-	require.True(t.T(), d.prevDirListingTimeStamp.IsZero())
+	require.True(t.T(), d.prevDirListingTimeStamp.Load() == 0)
```
kislaykishore:
```suggestion
	require.Zero(t.T(), d.prevDirListingTimeStamp.Load())
```

internal/fs/inode/dir.go
vedantdas-source: Yes. I've added a test to check for that. We can run a similiar test to check for the race condition in the original code. (by changing to `d.prevDirListingTimeStamp = d.cacheClock.Now()` in the `Test_ShouldInvalidateKernelListCache_RaceCondition` test i've added) and running the `go test -race ./internal/fs/inode/...`

internal/fs/inode/dir_test.go
```diff
@@ -2402,3 +2403,36 @@ func (t *DirTest) TestMetadataPrefetcher_InitializationGuards() {
 		})
 	}
 }
+
+func (t *DirTest) Test_ShouldInvalidateKernelListCache_RaceCondition() {
```
kislaykishore, line 2407: Not blocking this PR - Please log a bug for it to demonstrate that this bug exists in master along with the impact that this issue can have.

kislaykishore (approved):

internal/fs/inode/dir_test.go
vedantdas-source: Ok. I'll discuss it with prince a bit more and will take the necessary steps.

PR #4868, opened 2026-07-10
fix(bucket type optimization): Add support for multiple bucket types in bucket type flag optimization

abhishek10004:
### Summary
Enables specifying multiple bucket types using a comma-separated string (e.g. `bucket-type: "zonal, pirlo"`) in `params.yaml` for flag optimization rules, eliminating duplicate entries for identical configurations across bucket types.

### Key Changes
- **Configuration Types (`cfg/shared/types.go`)**:
  - Added `BucketTypeList` type with custom `UnmarshalYAML` support to parse CSV-style strings (`"zonal, pirlo"` or `"zonal"`).
  - Updated `BucketTypeOptimization` struct to use `BucketTypes BucketTypeList` (formerly single string `BucketType`).
- **Optimization Evaluation (`cfg/optimize.go`)**:
  - Updated `getOptimizedValue` to iterate over all bucket types listed in `BucketTypes` when matching against the active mount's bucket type.
- **Config Generator & Validation (`tools/config-gen/`)**:
  - Updated parser to validate `BucketTypeList` rules, ensuring no duplicate bucket types or empty lists are specified per flag.
  - Updated `config.tpl` and `config_test.tpl` templates to generate and test multi-bucket-type optimization rules.
- **Parameter Simplification (`cfg/params.yaml` & `cfg/config.go`)**:
  - Consolidated duplicate `zonal` and `pirlo` optimization entries in `params.yaml` into single `- bucket-type: "zonal, pirlo"` rules and regenerated `cfg/config.go`.

### Testing
- Added unit tests in `cfg/shared/types_test.go` and `tools/config-gen/parser_test.go` verifying valid CSV parsing and error handling for invalid/duplicate entries.
- Verified all generated code (`go generate`) and passed all existing/new unit tests (`go test ./cfg/... ./tools/config-gen/...`).

b/533339400

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/cfg/shared/types.go
+++ b/cfg/shared/types.go
@@ -27,13 +27,14 @@ type BucketTypeList []string
 
 func (b *BucketTypeList) UnmarshalYAML(value *yaml.Node) error {
 	var rawStrings []string
-	if value.Kind == yaml.ScalarNode {
+	switch value.Kind {
+	case yaml.ScalarNode:
 		rawStrings = []string{value.Value}
-	} else if value.Kind == yaml.SequenceNode {
+	case yaml.SequenceNode:
 		if err := value.Decode(&rawStrings); err != nil {
 			return err
 		}
-	} else {
+	default:
 		return fmt.Errorf("bucket-type must be a string or list of strings, got %v", value.Kind)
 	}
 
--- a/cfg/shared/types.go
+++ b/cfg/shared/types.go
@@ -16,28 +16,38 @@ package shared
 
 import (
 	"fmt"
+	"strings"
 
 	"gopkg.in/yaml.v3"
 )
 
-// BucketTypeList represents a list of bucket types, supporting both scalar string (e.g. "zonal")
-// and YAML sequence (e.g. ["zonal", "pirlo"]) in params.yaml.
+// BucketTypeList represents a list of bucket types, supporting scalar strings (e.g. "zonal" or "zonal, pirlo")
+// and YAML sequences (e.g. ["zonal", "pirlo"]) in params.yaml.
 type BucketTypeList []string
 
 func (b *BucketTypeList) UnmarshalYAML(value *yaml.Node) error {
+	var rawStrings []string
 	if value.Kind == yaml.ScalarNode {
-		*b = []string{value.Value}
-		return nil
-	}
-	if value.Kind == yaml.SequenceNode {
-		var slice []string
-		if err := value.Decode(&slice); err != nil {
+		rawStrings = []string{value.Value}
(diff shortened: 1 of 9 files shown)


cfg/shared/types.go
```diff
@@ -14,6 +14,32 @@
 
 package shared
 
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+// BucketTypeList represents a list of bucket types, supporting both scalar string (e.g. "zonal")
```
meet2mky: Is there any specific reason to support both? if no, we can support the [] only keeping the code minimum. Wdyt?

cfg/shared/types.go
```diff
@@ -28,8 +54,8 @@ type MachineBasedOptimization struct {
 
 // BucketTypeOptimization defines a bucket-type-based optimization.
 type BucketTypeOptimization struct {
-	BucketType string `yaml:"bucket-type"`
-	Value      any    `yaml:"value"`
+	BucketType BucketTypeList `yaml:"bucket-type"`
```
meet2mky: Nit: Should it be bucketType or BucketTypes now?
```suggestion
	BucketTypes BucketTypeList `yaml:"bucket-types"`
```

meet2mky (commented):
Looks good. Minor comments!

cfg/shared/types.go
kislaykishore: +1 - why not support csv values alone similar to flags e.g. "zonal,pirlo"?

kislaykishore (commented):
Changes look good. Just a couple of nits.

cfg/shared/types.go
abhishek10004: Added support for both - single entry ("zonal") and list (["zonal","flat"]) because it needed just this Unmarshal function for the extra support. Otherwise it would only support the list (which would need users to use [] for even a single entry, like ["zonal"]). Supporting for both seemed better to me.

Regarding adding support for csv values, "zonal,pirlo". I have done that as well.

meet2mky: There are no user here IMO. It's just the bucket type mapping to optimisation using params.yaml. Supporting multiple ways can cause developers to have inconsistent config if both style succeed.

meet2mky (approved):
LGTM

cfg/shared/types.go
abhishek10004: Yeah, I've now limited it to string only ("zonal" or "zonal, pirlo") after discussion with Kislay.

kislaykishore (approved):

PR #4869, opened 2026-07-10
perf(inode): optimized lookupCount struct to remove unnecessary fields

vedantdas-source:
### Description
Every inode (`FileInode`, `dirInode`, `SymlinkInode`, etc.) embeds a `lookupCount` struct to track active FUSE kernel lookups. Previously, this struct was defined as:

```go
type lookupCount struct {
    id        fuseops.InodeID // 8 bytes
    count     uint64          // 8 bytes
    destroyed bool            // 1 byte
}
```
Due to 64-bit compiler alignment padding, the 1-byte boolean padded the struct out to 24 bytes. Furthermore, the `destroyed` boolean was dead code (it was checked, but never set), and the `id` field was redundant because the parent inodes already store their own ID.

#### Changes Made
- **Type Aliasing**: Converted the struct into a pure type alias (`type lookupCount int64`), dropping the footprint to exactly 8 bytes.
- **State Merging**: The active count and "destroyed" state are now merged into a single `int64`. A value `>= 0` represents the active count, and `-1` represents the destroyed state.
- **Redundancy Removal**: Removed the redundant `.Init(id)` calls. The `.Inc(id)` and `.Dec(id)` methods now accept the `fuseops.InodeID` dynamically from the parent inode to preserve existing safety panic messages.
- **Overflow Protection**: Casted the internal count check to `uint64` in `.Dec()` to prevent negative wrap-around overflow bypasses in the event of an abnormally large decrement request.
 
Existing test passes seamlessly.
### Testing details
1. Manual - NA
2. Unit tests - Added `internal/fs/inode/lookup_count_test.go` . Validated exact bounds checking, underflow panics, wrap-around protection (`math.MaxUint64`), and idempotent destruction.
3. Integration tests - NA

### DIFF
--- a/internal/fs/inode/lookup_count.go
+++ b/internal/fs/inode/lookup_count.go
@@ -24,40 +24,36 @@ import (
 // paranoid panics. External synchronization is required.
 //
 // May be embedded within a larger struct. Use Init to initialize.
-type lookupCount struct {
-	id        fuseops.InodeID
-	count     uint64
-	destroyed bool
-}
+type lookupCount int64
 
-func (lc *lookupCount) Init(id fuseops.InodeID) {
-	lc.id = id
-}
-
-func (lc *lookupCount) Inc() {
-	if lc.destroyed {
-		panic(fmt.Sprintf("Inode %v has already been destroyed", lc.id))
+func (lc *lookupCount) Inc(id fuseops.InodeID) {
+	if *lc == -1 {
+		panic(fmt.Sprintf("Inode %v has already been destroyed", id))
 	}
 
-	lc.count++
+	(*lc)++
 }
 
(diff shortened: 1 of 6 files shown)


internal/fs/inode/lookup_count.go
```diff
@@ -24,40 +24,36 @@ import (
...
-type lookupCount struct {
-	id        fuseops.InodeID
-	count     uint64
-	destroyed bool
-}
-
-func (lc *lookupCount) Init(id fuseops.InodeID) {
-	lc.id = id
-}
+type lookupCount int64
```
raj-prince, line 27: Lets add a TODO to get rid of this struct at all, we can use struct embedding (with Inc/Dec lookUpCount) method to re-use the code.

raj-prince (approved):

PR #4871, opened 2026-07-12
feat(auth): add --experimental-auth-token-file for short-lived OAuth2 access tokens

### LINKED ISSUE #4686: Feature request: --auth-token-file for short-lived OAuth2 access tokens
## Use case
FUSE is being used in short-lived microvm sandbox environments like [e2b](https://e2b.dev), Modal, Daytona, and similar Firecracker-microVM-based sandboxes. Many of these providers offer fast spin up and exist in an api platform (not always self hosted where workload identity is an option). 

in these environments:
  - The orchestrator already has authenticated GCP credentials and can mint a short-lived oauth access token like with sts Credential Access Boundary downscoping, scoped to a specific bucket prefix on the orchestrator side.
  - The orchestrator wants to hand that bearer token to `gcsfuse` inside the sandbox, mount, and move on.
  - The sandbox lifetime is much shorter than the token TTL (sandbox idle timeout often ~120s, token TTL typically 1h), so refresh is structurally unnecessary.

## Current options and why they don't fit

  - `--key-file <SA-JSON>`: shipping a long-lived service account private key into a sandbox where AI-generated code is going to execute is a non-starter from a security perspective. We want short-lived, prefix-scoped credentials. A typical pattern is a single gcs bucket with per user prefixes, and 1 sandbox per user, so a token scoped to a single users prefix and mounted with --only-dir is perfect. 

  - `--token-url <http-url>`: works, but forces us to run an HTTP server inside the sandbox just to serve a static token. Concretely, our current setup needs:

  1. A small HTTP daemon baked into the sandbox template, listening on `127.0.0.1:8001`, that reads a token file and returns it on `GET /token`:
     ```python
     from http.server import BaseHTTPRequestHandler, HTTPServer

     class Handler(BaseHTTPRequestHandler):
         def do_GET(self):
             with open("/tmp/gcs-token", "rb") as f:
                 body = f.read()
             self.send_response(200)
             self.send_header("Content-Type", "application/json")
             self.end_headers()
             self.wfile.write(body)

     HTTPServer(("127.0.0.1", 8001), Handler).serve_forever()

  2. Orchestrator-side glue to write the token file and start that daemon before gcsfuse runs:
  echo '{"access_token":"ya29...","expires_in":3600,"token_type":"Bearer"}' > /tmp/gcs-token
  nohup /opt/token-server.py > /dev/null 2>&1 &
  sudo gcsfuse --token-url=http://127.0.0.1:8001/token --only-dir=user-prefix bucket /home/user

  This works but it's ~50 lines of glue (HTTP daemon + lifecycle + readiness check) per integration just to satisfy the "fetch-via-URL" constraint. It also adds a localhost listening port for the lifetime of the
  sandbox, which is one more thing to harden against accidentally exposing.

  Proposal: --auth-token-file <path>

  Add a flag that points at a file containing an OAuth2 token response, matching the existing --key-file API shape:

  echo '{"access_token":"ya29...","expires_in":3600,"token_type":"Bearer"}' > /tmp/gcs-token
  sudo gcsfuse --auth-token-file=/tmp/gcs-token --only-dir=user-prefix bucket /home/user

  The file would contain the same JSON shape that --token-url endpoints already return today, so the credential-loader could reuse most of the existing code path:

  {
    "access_token": "ya29.<bearer>",
    "expires_in": 3600,
    "token_type": "Bearer"
  }

  Behaviorally:
  - On mount, read the file once and construct an oauth2.Token directly. No JWT signing, no metadata server call, no HTTP.
  - When the cached token nears expiry, re-read the file (so callers who care can refresh by rewriting the file from outside) — or fail if expires_in has elapsed and the file hasn't been touched. Either is fine;
  the latter is simpler.

  For our use case the sandbox always dies before the token expires, so even a "read once, fail on expiry" implementation would be sufficient.

  Alternatives considered

  - --auth-token <bearer> on argv: would leak via ps, /proc/*/cmdline, shell history, audit logs. File-based is strictly better for the same convenience.
  - GOOGLE_OAUTH_ACCESS_TOKEN env var: same ps//proc exposure as argv. File-based avoids it.
  - Workload Identity Federation via external_account JSON: works in some environments but requires a token-source URL or executable, which has the same "you need a callback" problem we're trying to avoid.

## Willing to contribute

  Happy to send a PR if there's interest. Would appreciate guidance on:
  - Preferred flag name (--auth-token-file? --access-token-file? something else)
  - Behavior when expires_in elapses with no file rewrite (fail vs. continue and let GCS reject)
  - Whether the file should be the OAuth2 JSON shape above, or a simpler bearer-only file (one line: the access token string)

asdiamond:
Fixes #4686

## What

Adds `--experimental-auth-token-file <path>` (config: `gcs-auth.experimental-auth-token-file`), a flag that points at a file containing an OAuth2 token response in JSON format, used directly as the GCS credential:

```bash
echo '{"access_token":"ya29...","expires_in":3600,"token_type":"Bearer"}' > /tmp/gcs-token
gcsfuse --experimental-auth-token-file=/tmp/gcs-token --only-dir=user-prefix bucket /mnt/gcs
```

This targets short-lived sandbox environments (Firecracker microVM providers like e2b, Modal, Daytona) where an orchestrator mints a downscoped, prefix-scoped access token (e.g. via STS Credential Access Boundaries) and hands it to gcsfuse inside the sandbox — without shipping a long-lived SA key into an environment executing untrusted code, and without running a localhost HTTP server just to satisfy `--token-url`.

## Design decisions (per maintainer guidance on #4686)

- **Flag name**: `--experimental-auth-token-file` (@trinadhkotturu preferred `auth-token-file`, @anushka567 asked for the `experimental-` prefix for external contributions).
- **Fail at mount, not at GCS**: the token file is read eagerly when the token source is constructed, so mounting fails immediately if the token is missing, malformed, empty, or expired (@trinadhkotturu's recommendation).
- **File shape**: the OAuth2 token response JSON, consistent with what `--token-url` endpoints return (@anushka567's recommendation). The relative `expires_in` field is anchored to the file read time; an absolute `expiry` field (RFC 3339), when present, takes precedence.
- **Refresh**: the source is wrapped in `oauth2.ReuseTokenSource`, so the file is re-read when the cached token expires. An external process may therefore refresh the credential by rewriting the file; if the file isn't refreshed, requests fail with an "invalid or expired token" error rather than hitting GCS with a stale credential.
- **Mutual exclusion**: combining the new flag with `--key-file` or `--token-url` is rejected at config validation time, to fail early instead of silently picking one. Happy to relax this to a documented precedence order if you'd prefer.

## Changes

- `cfg/params.yaml` + regenerated `cfg/config.go` (via `go generate`): new param, type `resolvedPath`.
- `internal/auth/file_token_source.go`: new `fileTokenSource` + `NewTokenSourceFromTokenFile`, mirroring the existing `proxyTokenSource` structure.
- `internal/storage/storageutil/auth_client_option.go`: token-file branch in `GetClientAuthOptionsAndToken` (google-lib-auth flow).
- `internal/auth/auth.go` / `internal/storage/storageutil/client.go`: token-file support in the legacy `GetTokenSource` flow; also replaced the deprecated `golang.org/x/net/context` import with `context` in `client.go` to satisfy the `govet` inline check on the touched line.
- `cfg/validate.go`: `isValidGcsAuthConfig` mutual-exclusion check.
- Tests: `internal/auth/file_token_source_test.go` (valid/expired/no-expiry/malformed/missing/empty-token/re-read cases), `storageutil/auth_client_option_test.go`, `cfg/validate_test.go`, and a happy-path + conflict cases in `cmd/root_test.go`, per the dev guide.

## Testing

- `go test ./cmd/... ./cfg/... ./internal/auth/... ./internal/storage/storageutil/...` passes on linux/amd64 (Docker, golang:1.26).
- `go test -run=PATTERN_THAT_DOES_NOT_MATCH_ANYTHING ./...` (buildTest) compiles cleanly on linux/amd64.
- `golangci-lint run -E=unused --new-from-rev=master` reports 0 issues on the touched packages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -594,6 +594,8 @@ type FileSystemConfig struct {
 type GcsAuthConfig struct {
 	AnonymousAccess bool `yaml:"anonymous-access"`
 
+	ExperimentalAuthTokenFile ResolvedPath `yaml:"experimental-auth-token-file"`
+
 	KeyFile ResolvedPath `yaml:"key-file"`
 
 	ReuseTokenFromUrl bool `yaml:"reuse-token-from-url"`
@@ -1026,6 +1028,8 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 	if err := flagSet.MarkHidden("enable-unsupported-path-support"); err != nil {
 		return err
 	}
+
+	flagSet.StringP("experimental-auth-token-file", "", "", "Experimental: Absolute path to a file containing an OAuth2 token response in JSON format ({\"access_token\": \"...\", \"expires_in\": 3600, \"token_type\": \"Bearer\"}), to be used directly as the GCS credential. Mounting fails if the token is expired. The file is re-read when the token expires, so an external process may refresh the credential by rewriting the file.")
 
 	flagSet.BoolP("experimental-enable-dentry-cache", "", false, "When enabled, it sets the Dentry cache entry timeout same as metadata-cache-ttl. This enables kernel to use cached entry to map the file paths to inodes, instead of making LookUpInode calls to GCSFuse.")
 
@@ -1637,6 +1641,10 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 	}
 
 	if err := v.BindPFlag("enable-unsupported-path-support", flagSet.Lookup("enable-unsupported-path-support")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("gcs-auth.experimental-auth-token-file", flagSet.Lookup("experimental-auth-token-file")); err != nil {
 		return err
 	}
 
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -644,6 +644,8 @@ type FileSystemConfig struct {
 type GcsAuthConfig struct {
 	AnonymousAccess bool `yaml:"anonymous-access"`
 
+	ExperimentalAuthTokenFile ResolvedPath `yaml:"experimental-auth-token-file"`
+
 	KeyFile ResolvedPath `yaml:"key-file"`
 
 	ReuseTokenFromUrl bool `yaml:"reuse-token-from-url"`
@@ -1090,6 +1092,8 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 	if err := flagSet.MarkHidden("enable-unsupported-path-support"); err != nil {
 		return err
 	}
+
+	flagSet.StringP("experimental-auth-token-file", "", "", "Experimental: Absolute path to a file containing an OAuth2 token response in JSON format ({\"access_token\": \"...\", \"expires_in\": 3600, \"token_type\": \"Bearer\"}), to be used directly as the GCS credential. Mounting fails if the token is expired. The file is re-read when the token expires, so an external process may refresh the credential by rewriting the file.")
 
 	flagSet.BoolP("experimental-enable-dentry-cache", "", false, "When enabled, it sets the Dentry cache entry timeout same as metadata-cache-ttl. This enables kernel to use cached entry to map the file paths to inodes, instead of making LookUpInode calls to GCSFuse.")
 
@@ -1749,6 +1753,10 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 	}
 
 	if err := v.BindPFlag("enable-unsupported-path-support", flagSet.Lookup("enable-unsupported-path-support")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("gcs-auth.experimental-auth-token-file", flagSet.Lookup("experimental-auth-token-file")); err != nil {
 		return err
 	}
 
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -1093,7 +1093,7 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	flagSet.StringP("experimental-auth-token-file", "", "", "Experimental: Absolute path to a file containing an OAuth2 token response in JSON format ({\"access_token\": \"...\", \"expires_in\": 3600, \"token_type\": \"Bearer\"}), to be used directly as the GCS credential. Mounting fails if the token is expired. The file is re-read when the token expires, so an external process may refresh the credential by rewriting the file.")
+	flagSet.StringP("experimental-auth-token-file", "", "", "Experimental: Absolute path to a file containing an OAuth2 token response in JSON format ({\"access_token\": \"...\", \"expires_in\": 3600, \"token_type\": \"Bearer\"}), to be used directly as the GCS credential. A relative \"expires_in\" is measured from the file's modification time; an absolute RFC 3339 \"expiry\" field takes precedence if present. Mounting fails if the token is expired. The file is re-read when the token expires, so an external process may refresh the credential by replacing the file; do so atomically (write a temporary file, then rename it over the token file). The file is a credential: restrict its permissions to the owner, as with key-file.")
 
 	flagSet.BoolP("experimental-enable-dentry-cache", "", false, "When enabled, it sets the Dentry cache entry timeout same as metadata-cache-ttl. This enables kernel to use cached entry to map the file paths to inodes, instead of making LookUpInode calls to GCSFuse.")
 
--- a/internal/auth/file_token_source.go
+++ b/internal/auth/file_token_source.go
@@ -1,0 +1,69 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"time"
+
+	"golang.org/x/oauth2"
+)
+
+// fileTokenSource is an oauth2.TokenSource that reads an OAuth2 token
+// response from a JSON file on disk.
+type fileTokenSource struct {
+	path string
+}
+
+func (ts fileTokenSource) Token() (*oauth2.Token, error) {
+	contents, err := os.ReadFile(ts.path)
+	if err != nil {
+		return nil, fmt.Errorf("fileTokenSource cannot read token file: %w", err)
+	}
+
+	token := &oauth2.Token{}
+	if err := json.Unmarshal(contents, token); err != nil {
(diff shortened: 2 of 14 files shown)


internal/auth/file_token_source.go
```diff
@@ -0,0 +1,69 @@
...
+	if err := json.Unmarshal(contents, token); err != nil {
+		return nil, fmt.Errorf("fileTokenSource cannot decode token file %q: %w", ts.path, err)
+	}
+
+	// The expires_in field is relative, so anchor it to the time the file was
+	// read to get an absolute expiry. An absolute expiry field, if present in
+	// the file, takes precedence.
+	if token.Expiry.IsZero() && token.ExpiresIn > 0 {
+		token.Expiry = time.Now().Add(time.Duration(token.ExpiresIn) * time.Second)
+	}
```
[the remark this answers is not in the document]
asdiamond: This finding is a false positive on both points:

1. **`oauth2.Token` does have an `ExpiresIn` field.** It was added in `golang.org/x/oauth2` v0.17.0 ([CL 545056](https://go-review.googlesource.com/c/oauth2/+/545056)), and this repo pins **v0.36.0** (see `go.mod`). From `token.go` in v0.36.0:

   ```go
   // ExpiresIn is the OAuth2 wire format "expires_in" field,
   // which specifies how many seconds later the token expires,
   // relative to an unknown time base approximately around "now".
   // It is the application's responsibility to populate
   // `Expiry` from `ExpiresIn` when required.
   ExpiresIn int64 `json:"expires_in,omitempty"`
   ```

   So there is no compilation error — the branch builds and all unit tests pass — and `json.Unmarshal` does populate `expires_in`, since the struct tag exists.

2. **`token.Expiry` is not left zero.** Immediately after unmarshalling, this function does exactly what the library doc above prescribes:

   ```go
   if token.Expiry.IsZero() && token.ExpiresIn > 0 {
       token.Expiry = time.Now().Add(time.Duration(token.ExpiresIn) * time.Second)
   }
   ```

   so `oauth2.ReuseTokenSource` sees a real expiry and re-reads the file when the token expires. `Test_NewTokenSourceFromTokenFile_Success` asserts the parsed token has a non-zero, future `Expiry`, and `Test_NewTokenSourceFromTokenFile_ExpiredToken` covers the expired case.

The suggested auxiliary `tokenJSON` struct was necessary with x/oauth2 versions before v0.17.0, but is redundant on the version this project uses.

asdiamond:
@googlebot I signed it!

asdiamond:
@anushka567 can you please approve the workflow to run the tests? :)

cfg/config.go
```diff
@@ -1027,6 +1029,8 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	flagSet.StringP("experimental-auth-token-file", "", "", "Experimental: Absolute path to a file containing an OAuth2 token response in JSON format ({\"access_token\": \"...\", \"expires_in\": 3600, \"token_type\": \"Bearer\"}), to be used directly as the GCS credential. Mounting fails if the token is expired. The file is re-read when the token expires, so an external process may refresh the credential by rewriting the file.")
```
anushka567, line 1032: > so an external process may refresh the credential by rewriting the file.

How are we safeguarding against corruption to this file via malicious processes?

anushka567, line 1032: Is there a potential race condition involved where token file is read directly while the external process rewrites the token file in non-atomic fashion.

internal/auth/file_token_source.go
```diff
@@ -0,0 +1,69 @@
...
+	token := &oauth2.Token{}
+	if err := json.Unmarshal(contents, token); err != nil {
+		return nil, fmt.Errorf("fileTokenSource cannot decode token file %q: %w", ts.path, err)
+	}
+
+	// The expires_in field is relative, so anchor it to the time the file was
+	// read to get an absolute expiry. An absolute expiry field, if present in
+	// the file, takes precedence.
+	if token.Expiry.IsZero() && token.ExpiresIn > 0 {
+		token.Expiry = time.Now().Add(time.Duration(token.ExpiresIn) * time.Second)
```
anushka567, line 47: What is the rationale behind using the file access time as the anchor point for the `expires_in` field instead of file modification time based on when the "external" process modified the file with the refreshed credential?

rakesh-163:
Hello! Would love to know the status of this PR.

asdiamond:
@rakesh-163 the pr is in review

cfg/config.go
asdiamond: Good question. The short answer is that the token file is a credential and gets the same protection model as `--key-file`: filesystem permissions set by whoever provisions it. gcsfuse can't defend against a process that already has write access to the credential file, and the same is true of a service account key today. What this flag changes is the blast radius: a process that tampers with or steals the file gets a short-lived, downscoped token instead of a long-lived key.

That said, I've added two things in c79bd3ce4:

- **Permissions warning at mount time.** If the token file is writable by group or others, gcsfuse logs a warning suggesting `chmod 600` (mirrors what ssh does for key files). It's a warning rather than a hard failure so as not to break setups where the file lives on a filesystem with unusual permission semantics.
- **Corruption is handled, not trusted.** Any content that fails to parse, or parses to a token without an `access_token`, is rejected before it's ever sent to GCS. A corrupted file therefore surfaces as a mount error (or a request error on refresh), never as a silently wrong credential.

The flag help text now states explicitly that the file is a credential and should be permission-restricted like `key-file`.

asdiamond: Yes, that race is real. If a writer truncates and rewrites the file in place, a read landing mid-write can observe an empty or partially written file, which showed up as a spurious decode error on the GCS request that triggered the refresh.

Addressed in c79bd3ce4 from both sides:

- **Documented contract.** The help text now tells writers to replace the file atomically (write a temp file in the same directory, then `rename` it over the token file). This is the same convention kubelet uses for projected service account tokens, so orchestrators typically already do it.
- **Reader tolerance.** `fileTokenSource.Token()` now retries a read that observes a missing, empty, or malformed file (the three states consistent with a non-atomic rewrite in progress), up to 5 attempts with 10ms exponential backoff, ~150ms total. Definitive failures such as an expired token or a missing `access_token` are not retried. Tests cover the truncate-then-write and remove-then-recreate cases.

Also worth noting: the read stats the open file descriptor rather than the path, so the modification time used for `expires_in` (see the other thread) always belongs to the same contents that were read, even if the file is replaced concurrently.

internal/auth/file_token_source.go
asdiamond: You're right, and I've switched it. Anchoring `expires_in` to the read time was optimistic: if the writer minted a token with 3600s of life and gcsfuse read the file 40 minutes later, we'd compute an expiry 40 minutes past the real one and send a dead token to GCS.

In c79bd3ce4, `expires_in` is now measured from the file's modification time, which approximates when the external process wrote the refreshed credential. An absolute RFC 3339 `expiry` field, if present, still takes precedence, and the help text now recommends it for writers that want exact control (mtime can be skewed by copies, image bakes, or host/guest clock drift). Tests added for: expires_in anchored to a backdated mtime, a token whose expires_in has fully elapsed since mtime being rejected at mount, and `expiry` winning over `expires_in`.

asdiamond:
@anushka567 I have responded to the review comments, can you please re-review?

PR #4873, opened 2026-07-12

vedantdas-source:
### Description

#### Changes Made

### Testing details
1. Manual - NA
- `TestMinObject_UpdatedTime`

3. Integration tests - NA


internal/storage/gcs/object.go
```diff
@@ -86,13 +86,30 @@ type MinObject struct {
...
-	Updated         time.Time
-	Finalized       time.Time
+	Updated         int64
+	Finalized       int64
 	Metadata        map[string]string
 	ContentEncoding string
 	CRC32C          *uint32 // Missing for CMEK buckets
 }
 
+func TimeToNS(t time.Time) int64 {
```
kislaykishore, line 100: For the time: 1970-01-01 00:00:00 UTC i.e. time.Unix(0,0)
* TimeToNS(time.Unix(0,0).UTC()) -> will return 0.
* However, NSToTime(0) != time.Unix(0,0).UTC()


Example here: https://go.dev/play/p/uPlzB7nhicY

vedantdas-source: We can map the `time.Unix(0, 0)` to `math.MinInt64` so that
- `time.Time{}` (IsZero) strictly maps to 0 nanoseconds, preserving natural zero-initialization for new `MinObjects`.
- `time.Unix(0,0)` is explicitly mapped to `math.MinInt64`.

And as `math.MinInt64` represents the year 1677 it's impossible for a valid GCS object to possess it.And because every single read passes through `NSToTime()`, the `math.MinInt64` is safely decoded back into the standard `time.Unix(0,0)` struct.

internal/storage/fake/testing/bucket_tests.go
```diff
@@ -1376,6 +1376,8 @@ func (t *copyTest) DestinationDoesntExist() {
 	AssertNe(nil, statMinObj)
 	AssertNe(nil, statExtObjAttr)
 	statObj := storageutil.ConvertMinObjectAndExtendedObjectAttributesToObject(statMinObj, statExtObjAttr)
+	dst.Updated = gcs.NSToTime(gcs.TimeToNS(dst.Updated))
```
kislaykishore: What's this test testing? Seems like a test for `gcs.NSToTime` and `gcs.TimeToNS` since it's true for any value of dst.Updated. Same case for other tests in this file.

internal/fs/inode/file_mock_bucket_test.go
```diff
@@ -345,9 +345,9 @@ func (t *FileMockBucketTest) TestAttributes_NoChangeInAttributes() {
 	t.bucket.AssertExpectations(t.T())
 	// Attributes should NOT be updated because the size hasn't increased.
 	assert.Equal(t.T(), initialSize, attrs.Size)
-	assert.Equal(t.T(), initialTime, attrs.Mtime)
+	assert.Equal(t.T(), gcs.NSToTime(gcs.TimeToNS(initialTime)), attrs.Mtime)
```
kislaykishore: This can hide conversion errors. Ditto everywhere else.  It also violates: [go/tott/741](https://goto.google.com/tott/741) but given that this is pre-existing code, it's okay to ignore it.
```suggestion
	assert.Equal(t.T(), initialTime.UTC().Round(0), attrs.Mtime)
```

kislaykishore (changes requested):

internal/storage/fake/testing/bucket_tests.go
vedantdas-source: The `DeepEquals` checks compares unexported fields as well. This was done to strip Go's hidden monotonic clock reading from `dst` so that correctly checks against `statObj` (which lost its monotonic clock when it was compressed into an `int64` inside `MinObject`)

internal/storage/gcs/object.go
```diff
@@ -86,13 +87,37 @@ type MinObject struct {
...
+		return time.Time{}
+	}
+	if ns == math.MinInt64 {
+		return time.Unix(0, 0).UTC()
+	}
+	return time.Unix(0, ns).UTC()
+}
+
+func (m *MinObject) UpdatedTime() time.Time   { return NSToTime(m.Updated) }
+func (m *MinObject) FinalizedTime() time.Time { return NSToTime(m.Finalized) }
```
kislaykishore: Move these method right below the `MinObject` struct.

vedantdas-source: Done

kislaykishore (approved):

PR #4875, opened 2026-07-13
feat(native-retries): wire Control Plane mount retry predicate (ShouldRetryOnMount)

dangarhs:
### Description
- Wires the new control plane mount retry predicate (`ShouldRetryOnMount`) for `GetStorageLayout` calls when the `--enable-mount-retries` flag is enabled.
- Prevents dynamic mounts (`bucketName == ""` or `bucketName == "_"`) from inheriting mount-time retries to avoid runtime lookup hangs.
- Resolves EnableMountRetries on StorageClientConfig during storage handle creation for dynamic mounts.

b/530765797

### Testing details
1. Manual - NA
2. Unit tests - Added unit tests in `legacy_main_test.go` and `storage_handle_test.go`. Verified all tests in `cmd` and `internal/storage` pass cleanly:
   `go test -v ./cmd/... ./internal/storage/...`
3. Integration tests - NA

No.

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -142,7 +142,7 @@ func getConfigForUserAgent(mountConfig *cfg.Config) string {
 	}
 	return strings.Join(parts, ":")
 }
-func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle metrics.MetricHandle, isGKE bool) (storageHandle storage.StorageHandle, err error) {
+func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle metrics.MetricHandle, isGKE bool, isDynamicMount bool) (storageHandle storage.StorageHandle, err error) {
 	storageClientConfig := storageutil.StorageClientConfig{
 		ClientProtocol:                          newConfig.GcsConnection.ClientProtocol,
 		MaxConnsPerHost:                         int(newConfig.GcsConnection.MaxConnsPerHost),
@@ -151,6 +151,7 @@ func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle m
 		MaxRetrySleep:                           newConfig.GcsRetries.MaxRetrySleep,
 		MaxRetryAttempts:                        int(newConfig.GcsRetries.MaxRetryAttempts),
 		RetryMultiplier:                         newConfig.GcsRetries.Multiplier,
+		EnableMountRetries:                      newConfig.GcsRetries.EnableMountRetries && !isDynamicMount,
 		UserAgent:                               userAgent,
 		CustomEndpoint:                          newConfig.GcsConnection.CustomEndpoint,
 		KeyFile:                                 string(newConfig.GcsAuth.KeyFile),
@@ -202,7 +203,7 @@ func mountWithArgs(bucketName string, mountPoint string, newConfig *cfg.Config, 
 	if bucketName != canned.FakeBucketName {
 		userAgent := getUserAgent(newConfig.AppName, getConfigForUserAgent(newConfig), logger.MountInstanceID(fsName(bucketName)))
 		logger.Info("Creating Storage handle...")
-		storageHandle, err = createStorageHandle(newConfig, userAgent, metricHandle, isGKE)
+		storageHandle, err = createStorageHandle(newConfig, userAgent, metricHandle, isGKE, isDynamicMount(bucketName))
 		if err != nil {
 			err = fmt.Errorf("failed to create storage handle using createStorageHandle: %w", err)
 			return
--- a/internal/storage/control_client_wrapper.go
+++ b/internal/storage/control_client_wrapper.go
@@ -91,6 +91,8 @@ type storageControlClientWithRetry struct {
 
 	// Whether or not to enable retries for folder APIs.
 	enableRetriesOnFolderAPIs bool
+	// Whether or not to enable mount retries for GetStorageLayout call.
+	enableRetriesOnMount bool
 }
 
 func (sccwros *storageControlClientWithRetry) GetStorageLayout(ctx context.Context,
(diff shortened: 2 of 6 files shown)


internal/storage/control_client_wrapper.go
```diff
@@ -178,6 +184,13 @@ func WithRetriesOnFolderAPI() ControlClientOption {
 	}
 }
 
+// WithMountRetries returns an option that enables mount retries for GetStorageLayout call.
+func WithMountRetries() ControlClientOption {
```
meet2mky: Since we have both (mountRetries flag and also isDynamic) in the storageClientConfig so I think this can enabled by default without having this WithMountRetries helper and we can remove the mountRetries also from the clientConfigState.

similarly, in the NewStorageControlClient 

```
func NewStorageControlClient(raw StorageControlClient, clientConfig *storageutil.StorageClientConfig, opts ...ControlClientOption) StorageControlClient {
	var state clientConfigState
	for _, opt := range opts {
		opt(&state)
	}

	// Unwrap raw client to avoid nested wrappers.
	rawClient := raw
	var existingConfig *storageutil.RetryConfig
	// Check if the client is wrapped with billing project header injection first, and peel it off.
	if sccwbp, ok := rawClient.(*storageControlClientWithBillingProject); ok {
		rawClient = sccwbp.raw
	}
	// Check if the client is wrapped with retry logic next, and peel it off while capturing
	// its existing retry configuration.
	if sccwros, ok := rawClient.(*storageControlClientWithRetry); ok {
		rawClient = sccwros.raw
		existingConfig = sccwros.retryConfig
	}

	// 1. Wrap with retry capability
	retryConfig := existingConfig
	if retryConfig == nil && clientConfig != nil {
		retryConfig = storageutil.NewRetryConfig(clientConfig)
	}
	wrapped := &storageControlClientWithRetry{
		raw:         rawClient,
		retryConfig: retryConfig,
	}
	// 2. Enable retries on mount if configured and supported
	if clientConfig != nil {
		wrapped.enableRetriesOnMount = clientConfig.EnableMountRetries && !clientConfig.IsDynamicMount
	}
	// 3. Wrap with retries on folder APIs if configured
	if state.folderRetries {
		wrapped.enableRetriesOnFolderAPIs = true
	}
	// 4. Wrap with billing project if configured
	if state.billingProject != "" {
		return &storageControlClientWithBillingProject{raw: wrapped, billingProject: state.billingProject}
	}
	return wrapped
}
```

Wdyt?

dangarhs: Good suggestion, we dont need additional setting in clientConfigState now. Refactored it. Please check.

cmd/legacy_main.go
```diff
@@ -202,7 +204,8 @@ func mountWithArgs(bucketName string, mountPoint string, newConfig *cfg.Config,
 	if bucketName != canned.FakeBucketName {
 		userAgent := getUserAgent(newConfig.AppName, getConfigForUserAgent(newConfig), logger.MountInstanceID(fsName(bucketName)))
 		logger.Info("Creating Storage handle...")
-		storageHandle, err = createStorageHandle(newConfig, userAgent, metricHandle, isGKE)
+		isDynamicMount := isDynamicMount(bucketName)
+		storageHandle, err = createStorageHandle(newConfig, userAgent, metricHandle, isGKE, isDynamicMount)
```
meet2mky: Nit:
```suggestion
		storageHandle, err = createStorageHandle(newConfig, userAgent, metricHandle, isGKE, isDynamicMount(bucketName))
```

internal/storage/storage_handle.go
meet2mky, line 1: Change in this file is also not required so we can revert it.

internal/storage/control_client_wrapper_test.go
```diff
@@ -219,6 +219,83 @@ func (t *StorageLayoutRetryWrapperTest) TestGetStorageLayout_AllAttemptsTimeOut(
 	t.mockRawClient.AssertExpectations(t.T())
 }
 
+func (t *StorageLayoutRetryWrapperTest) TestGetStorageLayout_MountRetriesEnabled_Retry404ThenSuccess() {
+	// Arrange
```
meet2mky: Nit: // Arrange and // Act comments are not needed so we can remove it in this and other new tests.

meet2mky (dismissed):
Looks good some minor comments.

internal/storage/control_client_wrapper.go
meet2mky: Thanks

meet2mky:
Add bug on the PR description

internal/storage/control_client_wrapper_test.go
dangarhs: Just kept it consistent with the file earlier. Removed now.

internal/storage/storage_handle.go
dangarhs: Ack.

cmd/legacy_main.go
dangarhs: Done

meet2mky (dismissed):

cmd/legacy_main.go
```diff
@@ -171,6 +172,7 @@ func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle m
 		LocalSocketAddress:                      newConfig.GcsConnection.ExperimentalLocalSocketAddress,
 		EnableGrpcMetrics:                       newConfig.Metrics.ExperimentalEnableGrpcMetrics,
 		IsGKE:                                   isGKE,
+		IsDynamicMount:                          isDynamicMount,
```
raj-prince: Do we really need to pass this information to storage-config? IIUC, we just need to disable the mount-retries, why can't we pass the resolved value EnableMountRetries?

```go
// Compute EnableMountRetries directly when building storageClientConfig:
storageClientConfig := storageutil.StorageClientConfig{
    // ...
    EnableMountRetries: newConfig.GcsRetries.EnableMountRetries && !isDynamicMount,
}
```

raj-prince (changes requested):

cmd/legacy_main.go
dangarhs: Preferred not to mutate StorageClientConfig here to keep user flag same and let the storage client evaluate the dynamic mount retry behavior internally.

raj-prince: Unless not important we should not increase the storage layer API surface (by adding mount related config). Storage layer should not worry about dynamic-mount or not, should only worry about If mount-retry was enabled or not.

So, better to resolve while passing to storage package.

dangarhs: Updated to pass the pre-resolved EnableMountRetries value directly to StorageClientConfig, please check.

raj-prince (approved):

PR #4876, opened 2026-07-13
feat(kernel range reader): Optimize Kernel Reader default values based on bucket type

abhishek10004:
### Description

This PR introduces bucket-type specific default values and optimizations for Kernel Range Reader (KRR) and replaces the raw kernel pages configuration flag with a request-size based configuration.

### Key Changes

- **Bucket-Type Optimized KRR Defaults (`cfg/config_util.go`, `cfg/params.yaml`)**
  - Added distinct default limits for **Rapid** (`zonal`, `pirlo`) vs. **Non-Rapid** (`flat`, `hierarchical`) bucket types:
    - **Rapid Buckets**: Max request size = `1 MiB`, Max background = `192`, Max read-ahead = `16 MiB`.
    - **Non-Rapid Buckets**: Max request size = `16 MiB`, Max background = `96`, Max read-ahead = `128 MiB`.
  - Updated bucket-type optimization rules in `cfg/config.go` and `cfg/params.yaml` for `max-background`, `congestion-threshold`, `max-read-ahead-kb`, and `fuse-max-request-size-kb`.

- **Flag Replacement: `--fuse-max-request-size-kb` (`cfg/`, `cmd/`)**
  - Replaced the hidden `--fuse-max-pages-limit` flag with `--fuse-max-request-size-kb`.
  - Dynamically calculates the FUSE `MaxPages` setting (`MaxPagesForRequestSizeKb`) based on the host kernel page size (`os.Getpagesize()`).

- **Kernel Parameter Tuning (`cmd/mount.go`, `internal/fs/fs.go`)**
  - **Non-GKE Environments**: Applies pre-mount kernel `max_pages` settings when KRR is enabled for non-dynamic mounts if higher than the current system limit.
  - **GKE Environments**: Generalized post-mount kernel settings (`read_ahead_kb`, `congestion_threshold`) application via `kernel-params-file` to all bucket types when KRR is enabled (previously restricted to rapid buckets).

- **Dependency & Testing Updates**
  - Updated `github.com/jacobsa/fuse` dependency to support setting `MaxPages` in `fuse.MountConfig`.
  - Updated unit tests, config validation tests, and integration tests (`zonal_bucket_optimization_test.go`) to reflect the new flags and bucket-type optimization behavior.

b/534234379

### Testing details
1. Manual - In progress
2. Unit tests - Done
3. Integration tests - NA


cfg/config_util.go
```diff
@@ -23,27 +23,64 @@ import (
...
+	if requestSizeKb <= 0 {
+		return 0
+	}
+	return int((requestSizeKb * 1024) / int64(kernelPageSize))
 }
 
-func DefaultCongestionThreshold() int {
-	// 75 % of DefaultMaxBackground
-	return (3 * DefaultMaxBackground()) / 4
+func (bt BucketType) DefaultMaxBackground() int {
```
abhishek10004: Validating this for smaller machines as well.

cfg/config.go
```diff
@@ -32,7 +32,14 @@ var AllFlagOptimizationRules = map[string]shared.OptimizationRules{"file-system.
...
 			},
-			Value: int64(DefaultCongestionThreshold()),
+			Value: int64(BucketTypeZonal.DefaultCongestionThreshold()),
+		},
+		{
+			BucketTypes: shared.BucketTypeList{
+				"flat",
+				"hierarchical",
+			},
+			Value: int64(BucketTypeHierarchical.DefaultCongestionThreshold()),
```
meet2mky: Can this be misleading, It has flat in the list and value says hierarchical?

cfg/config.go
```diff
@@ -1157,9 +1200,9 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
...
+	flagSet.IntP("fuse-max-request-size-kb", "", BucketTypeZonal.DefaultFuseMaxRequestSizeKb(), "Sets the target maximum request size in KiB that FUSE can process in a single request (currently used to control read requests only). This is translated to the kernel max_pages limit based on host page size. As max_pages_limit is a global, machine-level configuration across all mounts, the host's limit is only updated if the calculated pages value is greater than the current system limit. Note that the FUSE kernel max_pages limit can be set to at most 65535 (fuse_max_max_pages), so values exceeding 65535 pages are capped at 65535.  Additionally, on GKE, the system-wide setting is capped to 16 MiB (16384 KiB) by default by the CSI driver. If needed to be set beyond that on GKE, the user has to manually increase the value on the node before GCSFuse mounting begins.")
```
meet2mky: Can we add validation that this must be greater than 0. It would help in handling in the lower layer. Also we can add validation that fuse-max-request-size-kb <=(fuseMaxPagesLimit = 65535) * pageSize. If user provides outside of range then just return invalid argument error.

cfg/config_util.go
```diff
@@ -16,34 +16,79 @@ package cfg
...
 	// on a single VM. Revise the numbers if you plan to support higher bandwidth VMs.
-	maxBackgroundLimit = 192
+	maxBackgroundRapidLimit = 192
+
+	// maxBackgroundNonRapidLimit configures the default max-background kernel fuse
+	// setting for non-rapid (regional and multi-regional) buckets.
+	maxBackgroundNonRapidLimit = 96
+
+	// maxReadAheadRapidKb configures the default max-read-ahead-kb setting for rapid buckets (16 MiB).
+	maxReadAheadRapidKb = 16 * 1024
```
meet2mky: How these constants are being decided?

cfg/config_util.go
```diff
@@ -16,34 +16,79 @@ package cfg
...
+}
+
+// MaxPagesForRequestSizeKb converts a request size in KiB to the number of kernel pages.
+// The FUSE kernel setting fuse_max_max_pages is capped at 65535 (math.MaxUint16).
+func MaxPagesForRequestSizeKb(requestSizeKb int64) int {
+	if requestSizeKb <= 0 {
+		return 0
+	}
+	bytes := requestSizeKb * 1024
+	pageSize := int64(kernelPageSize)
```
meet2mky: Can the page size be more than int limits?

cfg/config_util.go
```diff
@@ -16,34 +16,79 @@ package cfg
...
+	pages := (bytes + pageSize - 1) / pageSize
+	if pages > math.MaxUint16 {
+		return math.MaxUint16
```
meet2mky: We can add front loaded validations then this would be simplified instead of automatic capping we can just return to user that this is invalid configuration and will not take effect.

cfg/config.go
abhishek10004: Because the values are going to be same for flat & hierarchical for now, I've clubbed them together in the bucket type optimization in params.yaml and just used  BucketTypeHierarchical.DefaultCongestionThreshold() as the default (to save double specification).
If you feel strongly about this, I can specify defaults separately for each bucket type in param specification.

cfg/config_util.go
abhishek10004: Based on the testing done during POC time.

cfg/validate.go
meet2mky, line 38:
```suggestion
	ProfileAIMLCheckpointing                  = "aiml-checkpointing"
	FuseMaxPagesLimit                         = 65535 // (2^16 - 1) : Maximum page limit allowed by linux kernel
```

cfg/validate.go
```diff
@@ -114,6 +114,20 @@ func isValidSequentialReadSizeMB(size int64) error {
...
+	}
+	if requestSizeKb > int64(math.MaxInt-kernelPageSize+1)/1024 {
+		return fmt.Errorf("invalid value for fuse-max-request-size-kb: %d; value is too large and causes overflow", requestSizeKb)
+	}
+	pages := MaxPagesForRequestSizeKb(int(requestSizeKb))
+	if pages > fuseMaxPagesLimit {
+		return fmt.Errorf("invalid value for fuse-max-request-size-kb: %d; resulting page count %d exceeds maximum allowed value of %d", requestSizeKb, pages, fuseMaxPagesLimit)
+	}
+	return nil
+}
```
meet2mky, line 127:
```suggestion
func isValidFuseMaxRequestSizeKb(requestSizeKb int64) error {
	if requestSizeKb <= 0 {
		return fmt.Errorf("invalid value for fuse-max-request-size-kb: %d; should be > 0", requestSizeKb)
	}
	pageSizeKb := int64(os.Getpagesize()) / 1024
	if requestSizeKb > FuseMaxPagesLimit*pageSizeKb {
		return fmt.Errorf("invalid value for fuse-max-request-size-kb: %d; exceeds maximum allowed limit of %d", requestSizeKb, FuseMaxPagesLimit*pageSizeKb)
	}
	return nil
}
```

cfg/validate.go
```diff
@@ -346,6 +360,12 @@ func ValidateConfig(v *viper.Viper, config *Config) error {
 		return fmt.Errorf("error parsing gcs-connection config: %w", err)
 	}
 
+	if v.IsSet("file-system.fuse-max-request-size-kb") || config.FileSystem.FuseMaxRequestSizeKb != 0 {
```
meet2mky: Nit: This check can be removed right?

config.FileSystem.FuseMaxRequestSizeKb != 0

cmd/mount.go
```diff
@@ -190,6 +201,10 @@ func getFuseMountConfig(fsName string, newConfig *cfg.Config) *fuse.MountConfig
 		EnableAsyncReads: newConfig.FileSystem.EnableKernelReader,
 	}
 
+	if newConfig.FileSystem.FuseMaxRequestSizeKb != 0 {
```
meet2mky: Nit: prefer > 0 as it's easier to read it's a positive value.

cmd/mount.go
```diff
@@ -147,6 +148,16 @@ be interacting with the file system.`)
 
 	fsName := fsName(bucketName)
 
+	// Apply pre mount kernel settings in non-GKE environments for non dynamic mounts when kernel reader is enabled.
+	if !isDynamicMount(bucketName) && !cfg.IsGKEEnvironment(mountPoint) && newConfig.FileSystem.EnableKernelReader {
+		maxPages := cfg.MaxPagesForRequestSizeKb(int(newConfig.FileSystem.FuseMaxRequestSizeKb))
```
meet2mky, line 153: Do we need to ensure if this is > 0 and the existing limit is < request page size kb and also should we restore after mount succeed or it should be fine in GKE because it only succeeds when we run as root and on multiple mounts of GCSFuse on same VM there is no syncronization mechanism?

meet2mky (commented):
Changes look good overall some minor comments.

cmd/mount.go
abhishek10004: We're checking and updating the system limit only if that is less than what we want to set (update only happens when ShouldUpdateMaxPagesLimit() returns true)

Yeah, there is no easy way to synchronize on a VM (there is no single coordinator etc. that you can leverage for setting this; each GCSFuse process would be running independently).

cfg/config.go
raj-prince: +1, maybe we can rename it to standard (vs rapid/zonal) - differentiation based on storage class - https://docs.cloud.google.com/storage/docs/storage-classes.

cfg/config_util.go
```diff
@@ -16,34 +16,72 @@ package cfg
...
 	"os"
 	"runtime"
 	"strings"
 	"time"
 )
 
 const (
-	// maxBackgroundLimit configures the upper limit for max-background kernel fuse
-	// setting. This is more than sufficient to saturate the 200 Gbps network bandwidth
+	// maxBackgroundRapidLimit configures the upper limit for max-background kernel fuse
```
raj-prince: nit: Should we group the value for rapid vs standard (or non-rapid) bucket for better readability?

```go
// Rapid bucket defaults (zonal, pirlo)
const (
	rapidMaxBackground    = 192
	rapidMaxReadAheadKb   = 16 * 1024 // 16 MiB
	rapidMaxRequestSizeKb = 1024      // 1 MiB
)

// Non-rapid bucket defaults (flat, hierarchical)
const (
	nonRapidMaxBackground    = 96
	nonRapidMaxReadAheadKb   = 128 * 1024 // 128 MiB
	nonRapidMaxRequestSizeKb = 16 * 1024  // 16 MiB
)
```

raj-prince (commented):
Looks good, couple of nits.

cfg/config_util.go
abhishek10004: > +1, maybe we can rename it to standard (vs rapid/zonal) - differentiation based on storage class - https://docs.cloud.google.com/storage/docs/storage-classes.

This comes back to the thing that we had discussed earlier. We do not want to add bucket groups/classes in the bucket type (right now bucket type maps 1:1 with GCS bucket types). We can add something separate called StorageClass and use that for default* methods. This will be fine for now but in case, we'd need to change values for individual bucket types in future, we'd have to move the default* methods to BucketType again.

cfg/config.go
abhishek10004: Created StorageClass for better readability

meet2mky (approved):
LGTM

raj-prince (approved):

PR #4877, opened 2026-07-14
test(machine_type_test): fix flag parsing and mounted directory in gke/machine_type_test

PranjalC100:
### Description
It modifies the go test command to include the `-args` flag before passing the test-specific arguments. This ensures that custom flags like `--integrationTest` and `--mountedDirectory` are correctly passed to the compiled test binary instead of being incorrectly interpreted by the go test command itself.

b/534386424

### Testing details
1. Manual - via KOKORO
2. Unit tests - NA
3. Integration tests - NA


meet2mky (approved):
LGTM

anushka567 (approved):

PR #4878, opened 2026-07-14
feat(cache): add the arena-backed radix tree cache implementation 

vedantdas-source:
### Description
This PR introduces initial structure of the `arenaRadix` cache.

To reduce memory footprint and garbage collection overhead, this implementation moves away from traditional pointer-based nodes. Instead, it utilizes a Left-Child Right-Sibling (LCRS) representation backed by a slice-based arena allocator (`[]arenaRadixNode`) and 32-bit indices. 

**Changes made**
* added `arenaRadixNode` and `arenaRadix` core structs.
* Internal arena memory management via a custom freelist (`allocateNode`, `freeNode`), entirely eliminating heap allocations during tree node creation.
* Helper utilities for deterministic key hashing and full path reconstruction.

NA

### Testing details
1. Manual - NA
2. Unit tests - Added a comprehensive suite in `arena_radix_test.go` to explicitly verify the internal tree mechanics and lru operations
3. Integration tests - NA

None. This PR introduces new internal logic alongside the existing implementations and does not yet mutate any public package APIs or interfaces.


kislaykishore:
Please add the LRU implementation too in the same PR and ensure that the tests that pass with other LRU implementations pass with this one as well.

internal/cache/lru/arena_radix.go
```diff
@@ -0,0 +1,156 @@
...
+type arenaRadix struct {
+	nodes    []arenaRadixNode
+	freeHead uint32
+	root     uint32
+}
+
+// FNV-1a 64-bit hash
+func hashString(s string) uint64 {
+	var h uint64 = 14695981039346656037
+	for i := 0; i < len(s); i++ {
```
kislaykishore: Modernize
```suggestion
	for i := range len(s) {
```

internal/cache/lru/arena_radix.go
```diff
@@ -0,0 +1,156 @@
...
+		h ^= uint64(s[i])
+		h *= 1099511628211
```
kislaykishore: Add constants for these magic numbers and explain their meanings.

internal/cache/lru/arena_radix.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return h
+}
+
+func (c *arenaRadix) getFullKey(nodeID uint32) string {
+	var parts []string
```
kislaykishore: How about we provision the some capacity to the slices in advance.
```suggestion
	parts := make([]string, 0, 10)
```

kislaykishore: Consider using a sync.Pool

internal/cache/lru/arena_radix.go
```diff
@@ -0,0 +1,156 @@
...
+func (c *arenaRadix) removeChild(nID uint32, childToRemoveID uint32) {
+	for pcurr := &c.nodes[nID].child; *pcurr != nilNode; pcurr = &c.nodes[*pcurr].sibling {
+		if *pcurr == childToRemoveID {
+			*pcurr = c.nodes[childToRemoveID].sibling
+
+			c.nodes[childToRemoveID].sibling = nilNode
+			c.nodes[childToRemoveID].parent = nilNode
+
+			return
+		}
```
kislaykishore: Reduce nesting
```suggestion
		if *pcurr != childToRemoveID {
		  continue
		}
		*pcurr = c.nodes[childToRemoveID].sibling

		c.nodes[childToRemoveID].sibling = nilNode
		c.nodes[childToRemoveID].parent = nilNode
```

vedantdas-source: Ack

internal/cache/lru/arena_radix.go
vedantdas-source: Yes, we can definitely provide a very small capacity initially.

vedantdas-source: Done. Refactored the `getFullKey` function to use `sync.Pool`

vedantdas-source:
> Please add the LRU implementation too in the same PR and ensure that the tests that pass with other LRU implementations pass with this one as well.

Done. Added the functionality of the LRU operations as well as tests that the existing lru cache are tested against.

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,397 @@
...
+
+// getNodeKey finds a leaf node ID by key.
+func (c *arenaRadix) getNodeKey(key string) (uint32, bool) {
+	nodeID, ok := c.nodeMap[hashString(key)]
+	if ok && c.nodes[nodeID].value != nil {
+		if c.verifyKey(nodeID, key) {
+			return nodeID, true
+		}
+	}
+	return nilNode, false
```
kislaykishore, line 292: How do you handle hash-collisions?
Say there are two keys, `key1` and `key2` and `hashString(key1) == hashString(key2)`. Assume that `c.nodeMap[hashString(key1)] == nodeId of key1`.

Now, let's see what happens when `Erase("key2")` is invoked. The `Erase(key)` has the following logic:
```
func (c *arenaRadix) Erase(key string) (value ValueType) {
	c.mu.Lock()
	defer c.mu.Unlock()

	nodeID, ok := c.getNodeKey(key)
	if !ok {
		return nil
	}

	return c.eraseInternal(nodeID)
}
```

* After acquiring the exclusive lock, `c.getNodeKey(key2)` will get called.
* `c.getNodeKey(key2)` will return nilNode, false because:
  - c.nodeMap[hashString(key2)] will return the nodeID corresponding to key1.
  - the verification that the nodeID is actually for key2, `verifyKey` method will get called and will return false since, the nodeID is for `key1`
  - `nilNode, false` will be returned and nothing will get erased.

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,397 @@
...
+package lru
+
+import (
+	"math"
+	"strings"
+
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/locker"
+)
+
+const nilNode uint32 = math.MaxUint32
```
kislaykishore, line 23: What does nilNode represent?

internal/cache/lru/arena_radix.go
```diff
@@ -0,0 +1,397 @@
...
+			return false
+		}
+		if key[end-prefixLen:end] != prefix {
+			return false
+		}
+		end -= prefixLen
+		curr = c.nodes[curr].parent
+	}
+
+	return end == 0
```
kislaykishore: If a node is disconnected or orphaned (e.g., during sub-tree deletions or parent re-assignments where curr.parent == nilNode), curr becomes nilNode after processing the node's prefix. If end reached 0 at that moment, verifyKey returns true, even though the node is not connected to c.root.

Impact
When getNodeKey checks nodeMap lookups, an unlinked or freed node whose parent pointer is nilNode will be validated as a legitimate key in the trie, returning stale or corrupted data.

Example test that fails:

```
func TestArenaRadixLRU_OrphanedNodeMapAndVerifyKeyBug(t *testing.T) {
	cache := NewArenaRadixCache(100)

	// 1. Create multi-level radix tree hierarchy
	_, err := cache.Insert("sub1/a.txt", testValue{100})
	assert.NoError(t, err)
	_, err = cache.Insert("sub1/sub2/b.txt", testValue{200})
	assert.NoError(t, err)
	_, err = cache.Insert("sub1/sub2/c.txt", testValue{300})
	assert.NoError(t, err)

	// 2. Erase sub-prefix "sub1/sub2/"
	cache.EraseEntriesWithGivenPrefix("sub1/sub2/")

	// 3. Check nodeMap for stale key hashes
	c := cache.(*arenaRadix)
	hashB := hashString("sub1/sub2/b.txt")
	hashC := hashString("sub1/sub2/c.txt")

	_, existsB := c.nodeMap[hashB]
	_, existsC := c.nodeMap[hashC]

	assert.False(t, existsB, "nodeMap should not contain stale hash for 'sub1/sub2/b.txt'")
	assert.False(t, existsC, "nodeMap should not contain stale hash for 'sub1/sub2/c.txt'")
}
```

internal/cache/lru/radix_arena_lru.go
```diff
@@ -0,0 +1,313 @@
...
+
+		lcp := longestCommonPrefix(search, c.nodes[childID].prefix)
+
+		if lcp == len(search) {
+			// We found the exact node where the prefix ends.
+			// Sever it entirely from the tree structure
+			c.removeChild(nodeID, childID)
+
+			// Now sweep the detached subtree to fix LRU and currentSize
+			c.freeSubtree(childID)
```
kislaykishore, line 276: `removeChild` sets `childID.parent = nilNode` before `freeSubtree` runs. Key re-computation fails, leaving stale map references pointing to freed nodes.

kislaykishore:
Please fix the failing tests too.

internal/cache/lru/radix_arena.go
vedantdas-source: `nilNode` acts as a custom `nil` pointer. As we do not use pointers anymore, we need to have a separate `uint32` value to represent a null pointer.

internal/cache/lru/radix_arena.go
vedantdas-source: As we had discussed earlier, we intend to revert to a classic tree traversal whenever we have a hash collision, but i forgot to add the logic for the same inside of  `getNodeKey`. Added it now.

internal/cache/lru/arena_radix.go
vedantdas-source: Fixed.

internal/cache/lru/radix_arena_lru.go
vedantdas-source: Done. I've temporarily set the parent pointer before `freeSubtree`.

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	return end == 0 && curr == c.root
+}
+
+// getNodeKey finds a leaf node ID by key.
+func (c *arenaRadix) getNodeKey(key string) (uint32, bool) {
+	nodeID, ok := c.nodeMap[hashString(key)]
+	if ok && c.nodes[nodeID].value != nil {
+		if c.verifyKey(nodeID, key) {
```
kislaykishore, line 262: Would you want to update the `nodeMap` if the `verifyKey` operation succeeds?

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,423 @@
...
+////////////////////////////////////////////////////////////////////////
+// Cache interface
+////////////////////////////////////////////////////////////////////////
+
+func (c *arenaRadix) eraseInternal(nodeID uint32) (value ValueType) {
+	deletedEntry := c.nodes[nodeID].value
+	c.currentSize -= deletedEntry.Size()
+
+	// Prevent hash collision cross-deletions
+	hash := hashString(c.getFullKey(nodeID))
```
kislaykishore, line 414: `c.getFullKey` constructs the entire string which requires string concatenation and adds to GC load. What if we perform the hash computation like this:
```
func (c *arenaRadix) hashNodeKey(nodeID uint32) uint64 {
	var stackBuf [64]uint32
	stack := stackBuf[:0]

	curr := nodeID
	for curr != c.root && curr != nilNode {
		stack = append(stack, curr)
		curr = c.nodes[curr].parent
	}

	var h uint64 = offset64
	for i := len(stack) - 1; i >= 0; i-- {
		prefix := c.nodes[stack[i]].prefix
		for j := 0; j < len(prefix); j++ {
			h ^= uint64(prefix[j])
			h *= prime64
		}
	}
	return h
}
```

internal/cache/metadata/stat_cache_test.go
```diff
@@ -543,7 +547,7 @@ func (t *StatCacheTest) Test_ShouldReturnHitTrueWhenOnlyObjectAlreadyHasEntry()
 }
 
 func (t *StatCacheTest) Test_ShouldEvictEntryOnFullCapacityIncludingFolderSize() {
-	localCache := lru.NewCache(uint64(2700))
+	localCache := lru.NewCache(uint64(2600))
```
kislaykishore: How can we ensure that all the StatCache tests run against ArenaRadix as well?

kislaykishore (commented):
LGTM - minor comments.

internal/cache/lru/radix_arena.go
vedantdas-source: If the `verifyKey` operation succeds we are good. 
I think we should update the `nodeMap` if we find a hash collision i.e when we have to do a tree traversal. Say for example we have `fileA` and `fileB` which have the same hash. 
- We insert `fileA`. All good
- We then insert `fileB` and the `nodeMap` is updated

If we then have lots of lookups for `fileA` we never overwrite the `nodeMap`  when the hash collision happens which it does in this case. This could lead us to perform a tree traversal each time for a `fileA` lookup. So it would be better to update the map.
I've added the logic to overwrite the `nodeMap`  when we have hash collision

internal/cache/lru/radix_arena.go
vedantdas-source: Yes this would be much better instead of doing `hashString(c.getFullKey(nodeID))` we can simplify it by just having a function which hashes a key based on the `nodeID`. I've added the function, ran some internal tests, and changed `radix_arena_lru.go` and `radix_arena.go` to use this instead.

internal/cache/metadata/stat_cache_test.go
vedantdas-source: I've refactored the hard coded `lru.NewCache` to a dynamic `cacheFactory` where we go through all of the cache implementations which executes the test for all implementations.

internal/cache/lru/radix_arena_lru.go
```diff
@@ -0,0 +1,316 @@
...
+// without changing the order of entries in the cache. Return nil if no value
+// is present.
+//
+// Note: Because this look up doesn't change the order, it only acquires and
+// releases read lock.
+func (c *arenaRadix) LookUpWithoutChangingOrder(key string) (value ValueType) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	nodeID, ok := c.getNodeKey(key)
```
kislaykishore, line 183: We take a read-lock but `c.getNodeKey` mutates the `c.nodeMap`. This can lead to application crashing. Either ensure that `c.nodeMap` in this case doesn't have any side-effects, or take a write lock.
If you agree, please add a test to repro the crash and fix it.

internal/cache/lru/radix_arena_lru.go
```diff
@@ -0,0 +1,316 @@
...
+func (c *arenaRadix) freeSubtree(nodeID uint32) {
+	if nodeID == nilNode {
+		return
+	}
+	currID := nodeID
+	for currID != nilNode {
+		if c.nodes[currID].value != nil {
+			c.currentSize -= c.nodes[currID].value.Size()
+			c.remove(currID)
+			delete(c.nodeMap, c.hashNodeKey(currID))
```
kislaykishore: In case of hash-collision, this can lead to cross-deletion.
```suggestion
hash := c.hashNodeKey(currID)
if c.nodeMap[hash] == currID {
   delete(c.nodeMap, hash) 
}
```

internal/cache/lru/radix_arena_lru.go
vedantdas-source: Yes , when I added the logic to overwrite an entry on hash collision, it made the `getNodekey` a write operation. So we'd have to acquire a write lock.
I have tested it, but adding the same in the `radix_arena_test.go` wont work because we have to manually set `c.nodeMap[hashB] = c.nodeMap[hashA]` which is not possible as `nodeMap` is unexported.

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,437 @@
...
+		h ^= uint64(s[i])
+		h *= prime64
+	}
+	return h
+}
+
+// hashNodeKey computes the FNV-1a hash of the full key for a given node
+// without allocating a concatenated string, reducing GC pressure.
+func (c *arenaRadix) hashNodeKey(nodeID uint32) uint64 {
+	var stackBuf [64]uint32
```
kislaykishore: To not require slice expansion for longer file names.
```suggestion
	var stackBuf [128]uint32
```

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,437 @@
...
+}
+
+// getChild finds a child node whose prefix starts with the given byte.
+func (c *arenaRadix) getChild(nID uint32, b byte) uint32 {
+	n := &c.nodes[nID]
+	for curr := n.child; curr != nilNode; curr = c.nodes[curr].sibling {
+		if c.nodes[curr].prefix[0] == b {
+			return curr
+		}
+		//as the siblings are arranged in lexicographical order we can exit early
```
kislaykishore, line 131:
```suggestion
		// as the siblings are arranged in lexicographical order we can exit early
```

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,437 @@
...
+	sibling uint32
+	prev    uint32
+	next    uint32
+}
+
+// arenaRadix encapsulates the core tree structure and implements the lru.Cache interface.
+type arenaRadix struct {
+	maxSize     uint64
+	currentSize uint64
+	mu          locker.RWLocker
```
kislaykishore: You don't need an RWLocker now - an exclusive mutex would do just as well.

internal/cache/lru/radix_arena.go
```diff
@@ -0,0 +1,437 @@
...
+		prefix := c.nodes[stack[i]].prefix
+		for j := 0; j < len(prefix); j++ {
+			h ^= uint64(prefix[j])
+			h *= prime64
+		}
+	}
+	return h
+}
+
+func (c *arenaRadix) allocateNode() uint32 {
```
kislaykishore, line 84: With uint32 as pointers, we have enforced a cache limit. How can we ensure that the cache doesn't panic when we breach that limit? Maybe we should trigger evictions.

internal/cache/lru/radix_arena.go
vedantdas-source: Done

internal/cache/lru/radix_arena.go
vedantdas-source: We can trigger evictions when we insert, to make space for the new nodes if we hit the limit of 2<sup>32<sup> elements

kislaykishore (approved):

PR #4879, opened 2026-07-14
fix(metrics): Fixing double counting issue in read bytes count metrics in Kernel Range Reader & Buffered Reader

abhishek10004:
### Description
Currently read bytes count is getting double counted in Kernel Range Reader and Buffered reader since monitoringReadCloser reader (which wraps the GCS reader) increments the read bytes count internally. Hence, the readers don't have to increment it.

b/534618517

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/fs/metrics_test.go
+++ b/internal/fs/metrics_test.go
@@ -641,7 +641,6 @@ func TestReadFile_KernelRangeReaderMetrics(t *testing.T) {
 	require.NoError(t, err, "ReadFile")
 	waitForMetricsProcessing()
 
-	require.NoError(t, err, "ReadFile")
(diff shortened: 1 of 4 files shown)


internal/fs/metrics_test.go
```diff
@@ -609,6 +609,44 @@ func TestReadFile_KernelMRDReaderMetrics(t *testing.T) {
...
+		Offset: 0,
+		Size:   int64(len(content)),
+		Dst:    make([]byte, len(content)),
+	}
+
+	err = server.ReadFile(ctx, readOp)
+	require.NoError(t, err, "ReadFile")
+	waitForMetricsProcessing()
+
+	require.NoError(t, err, "ReadFile")
```
```suggestion
```

kislaykishore (dismissed):

kislaykishore (approved):

PR #4903, opened 2026-07-20
test: Remove flat bucket test configurations for Pirlo

vipnydav:
### Description
This PR removes the flat bucket test configurations for Pirlo in tools/integration_tests/test_config.yaml. Pirlo currently does not support flat buckets, so these test scenarios are no longer applicable and are being removed.

b/536789480

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


meet2mky:
@vipnydav Can you take a look at gemini comments.

meet2mky (changes requested):
PTAL

meet2mky (approved):

PR #4906, opened 2026-07-21
fix: Fix the gke machine type test failure

vadlakondaswetha:
1. Symptom
The GCSFuse CSI Driver multi-arch build was failing during the Cloud Build execution at the docker manifest create step. It threw a no such manifest error for the newly built _linux_amd64 image, despite the logs showing that the image was successfully compiled and pushed:

2. Root Cause
The issue was caused by a combination of deprecation and tool incompatibility:

GCR Deprecation: The build was originally configured to push to Container Registry (gcr.io), which is being sunset. We needed to migrate to Artifact Registry (us-docker.pkg.dev).
Docker CLI Bug in Cloud Build: The official Cloud Build step container (gcr.io/cloud-builders/docker) uses an outdated Docker version (20.10.24).
OCI Manifest Incompatibility: While buildx successfully pushed the OCI-compliant images to Artifact Registry, the older Docker CLI version in the container had known bugs that prevented it from correctly inspecting or recognizing OCI manifests in Artifact Registry, resulting in the false no such manifest error.

3. Resolution:
We modified csi_driver_build.yml to:

Migrate Registry: Point REGISTRY to Artifact Registry (us-docker.pkg.dev).
Upgrade Docker CLI: Added a step to install a modern Docker CLI version (docker-ce-cli v28.1.1) inside the builder container before running the makefile.
Update Auth: Authenticated the new Docker CLI against us-docker.pkg.dev using gcloud.

Manual run on kokoro is passing:

image

### DIFF
--- a/csi_driver_build.yml
+++ b/csi_driver_build.yml
@@ -95,14 +95,21 @@ steps:
   - '-c'
   - |
     set -e
-    # --- Install gcloud ---
+    # --- Install gcloud and newer Docker CLI ---
     apt-get update && apt-get install -y apt-transport-https ca-certificates gnupg curl
+    
+    # Docker repo
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /usr/share/keyrings/docker.gpg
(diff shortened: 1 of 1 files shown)


meet2mky (approved):
LGTM, verified on kokoro builds.

abhishek10004 (approved):

PR #4912, opened 2026-07-22
fix(GKE test): Ensuring new tests for read/write request size verification does not run on GKE

abhishek10004:
### Description
Ensuring new tests for read/write request size verification does not run on GKE as these are already being checked in
https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1514. Can be enabled later after checking the log-file parsing later on.

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - Done (Pending on GKE)


meet2mky (approved):
LGTM

PR #4916, opened 2026-07-23
feat: Reducing the memory used by fileInode - part1

vadlakondaswetha:
1. attrs Field Removal (Files, Directories, Symlinks):

> - Removed fuseops.InodeAttributes field (128 bytes) and replaced it with time.Time (24 bytes) for mtime. Attributes are now constructed on-the-fly when requested.
> - Savings: 104 bytes per cached Inode.

2. WriteContext Decoupling (Files only):

> - Moved active write-only fields (config, semaphore, tracing, metrics handles) out of FileInode struct and passed them dynamically during active write operations.
> - Savings: 48 bytes per FileInode.

**Size Improvements per inode entry (verified via unsafe.SizeOf)**

- FileInode - 31% (original size: 448, optimized size: 336)
- DirInode - 30% (original size: 344, optimized size: 240)
- Symlink Inode - 40% (original size: 256, optimized size: 152)

**Performance & Memory Savings at Scale**
For large directory caches (common in GCSFuse mounts), these optimizations translate to significant resident memory (RSS) savings:

For 1 Million Cached Files:

> Heap memory saved: 145 MB
> Estimated RSS Savings (with Go GC 2x multiplier): ~290 MB

For 10 Million Cached Files:

> Heap memory saved: 1.41 GB
> Estimated RSS Savings: ~2.83 GB

### Testing details
1. Manual - Done
2. Unit tests - Existing tests will cover
3. Integration tests -Existing tests


internal/fs/inode/file.go
```diff
@@ -1089,6 +1085,7 @@ func (f *FileInode) updateInodeStateAfterFlush(minObj *gcs.MinObject) {
 func (f *FileInode) updateInodeStateAfterSync(minObj *gcs.MinObject) {
 	if minObj != nil && !f.localFileCache {
 		f.src = *minObj
+		f.mtime = parseMtime(minObj)
```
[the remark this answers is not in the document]
vadlakondaswetha: updateInodeStateAfterFlush calls updateInodeStateAFterSync(). so it is handled.

internal/fs/inode/file.go
```diff
@@ -54,6 +53,39 @@ const (
...
+			mtime = timestamp
+		} else {
+			logger.Warnf("parseMtime: Error parsing gcsfuse_mtime %q: %v", formatted, err)
+		}
+	}
+	return mtime
+}
+
 type FileInode struct {
 	/////////////////////////
```
abhishek10004, line 90: What is the expected saving per inode?

internal/fs/inode/file.go
```diff
@@ -1224,7 +1222,8 @@ func (f *FileInode) InitBufferedWriteHandlerIfEligible(ctx context.Context, open
 			// instead of reaching out to the server for latest metadata. This is done to avoid
 			// forceful overwrites of local and latest object metadata with possibly stale server
 			// response. Since appends happen at the same generation, StatObject() call is redundant.
-			latestGcsObj = storageutil.ConvertMinObjectToObject(&f.src)
+			minObj := f.src
+			latestGcsObj = storageutil.ConvertMinObjectToObject(&minObj)
```
abhishek10004: Why are we not passing f.src directly?

internal/fs/inode/file.go
```diff
@@ -431,7 +452,6 @@ func (f *FileInode) IsUsingBWH() bool {
 //
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) Source() *gcs.MinObject {
-	// Make a copy, since we modify f.src.
```
abhishek10004, line 434: Should we keep this comment?

internal/fs/inode/file_test.go
```diff
@@ -191,24 +190,47 @@ func (t *FileTest) validateMrdInstanceMinObject() {
 	t.T().Helper()
 	// Validate only for rapid buckets
 	if t.in.bucket.BucketType().IsRapid() {
-		// Validate MinObject in inode and MRDInstance points to different copy of MinObject.
```
abhishek10004, line 194: Why are we changing this?

internal/fs/inode/file_test.go
```diff
@@ -191,24 +190,47 @@ func (t *FileTest) validateMrdInstanceMinObject() {
...
+		assert.Equal(t.T(), minObj.Size, actual.Size)
+		assert.Equal(t.T(), minObj.ContentEncoding, actual.ContentEncoding)
 	}
 }
 
 func (t *FileTest) validateMrdWrapperMinObject() {
 	t.T().Helper()
 	// Validate only for rapid buckets
 	if t.in.bucket.BucketType().IsRapid() {
-		// Validate MinObject in inode and MRDWrapper points to different copy of MinObject.
```
abhishek10004, line 205: Same as above. Why are we changing this?

internal/fs/inode/file.go
abhishek10004: mentioned in the description

internal/fs/inode/file.go
vadlakondaswetha: I made some optimizations to how we store minObj. Will raise a seperate PR for those changes. Some changes got into this PR.

abhishek10004 (approved):
LGTM

PR #4920, opened 2026-07-23
feat: Added strong consistency mode for openFile call

vadlakondaswetha:
OpenFile today doesn't make any GCS call and relies on inode created prior to file open. Because of this, if any changes are made on GCS after metadata is cached, we dont get that latest information. 

Now we added a mode when enabled, we check with GCS if the file is clobbered, if yes return ESTALE to the kernel. 
upon getting ESTALE, kernel will repeat the same cycle i.e, lookupinode, openFile.
When the openFile is called 2nd time, we will again check with GCS and now that we have the latest data we return succesful handleID to the kernel.

Note: If we return ESTALE in openFile, kernel makes a fresh lookupinode call to get latest inodeId and then calls openFile. If we return ESTALE in 2nd openFile it will return the error to application. It only retries once on ESTALE error.

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - Added

### DIFF
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -314,6 +314,7 @@ func (f *FileInode) CheckClobbered(ctx context.Context) (err error) {
 	}
 	if clobbered {
 		return &gcsfuse_errors.FileClobberedError{
+			Err:        errors.New("file was clobbered due to generation/metageneration mismatch"),
 			ObjectName: f.name.GcsObjectName(),
 		}
 	}
--- a/internal/fs/inode/file.go
+++ b/internal/fs/inode/file.go
@@ -304,6 +304,22 @@ func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool, inclu
 	}
 }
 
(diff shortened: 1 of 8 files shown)


internal/fs/inode/file.go
```diff
@@ -304,6 +304,23 @@ func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool, inclu
 	}
 }
 
+// CheckClobbered checks if the file has been clobbered by making a stat call to GCS.
+//
+// LOCKS_REQUIRED(f.mu)
+func (f *FileInode) CheckClobbered(ctx context.Context) (err error) {
```
raj-prince, line 310: Should we add unit test to validate this returns FileClobberedError if clobbered?

raj-prince (dismissed):
LGTM.

Maybe we can update the PR description or doc to clarify - The first ESTALE is retried transparently by the Linux kernel VFS. If ESTALE persists on the retried attempt, it is returned to the application.

raj-prince (approved):

PR #4922, opened 2026-07-23
feat: Making write-block-size-mb to accept fractional values

vadlakondaswetha:
changing write-block-size-mb to accept fractional values so users can specify the size in kb. This helps in reducing the memory size for small files. 

Today the same size buffer is created on go-sdk also. So effectively we are creating 2MB buffers even when 1K file is written. Changing this flag to parse fractional values will help users configure values in KB also.

### Testing details
1. Manual - done
2. Unit tests - added
3. Integration tests - NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -813,7 +813,7 @@ type WorkloadInsightConfig struct {
 }
 
 type WriteConfig struct {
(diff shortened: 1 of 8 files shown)


abhishek10004 (approved):
LGTM

PR #4925, opened 2026-07-23

vedantdas-source:
### Description

#### Changes Made

### Testing details
1. Manual - NA
- `TestMinObject_UpdatedTime`

3. Integration tests - NA

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -142,7 +142,7 @@ func getConfigForUserAgent(mountConfig *cfg.Config) string {
 	}
 	return strings.Join(parts, ":")
 }
-func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle metrics.MetricHandle, isGKE bool) (storageHandle storage.StorageHandle, err error) {
+func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle metrics.MetricHandle, isGKE bool, isDynamicMount bool) (storageHandle storage.StorageHandle, err error) {
 	storageClientConfig := storageutil.StorageClientConfig{
 		ClientProtocol:                          newConfig.GcsConnection.ClientProtocol,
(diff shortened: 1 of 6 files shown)


internal/gcsx/garbage_collect.go
```diff
@@ -53,7 +53,7 @@ func garbageCollectOnce(
 	group.Go(func() (err error) {
 		defer close(staleNames)
 		for o := range minObjects {
-			if now.Sub(o.Updated) < stalenessThreshold {
+			if now.Sub(o.UpdatedTime()) < stalenessThreshold {
```
kislaykishore: Could UpdatedTime() return 0 here leading to underflow? This can easily happen when `updated` is uninitialized in case of tests. Secondly, it could happen during struct creation in case this field is left initialized. Please check that this works in all those scenarios.

vedantdas-source: Done. I've added a check to explicitly check if the time is uninitialized

internal/storage/gcs/object.go
```diff
@@ -86,13 +87,37 @@ type MinObject struct {
...
-	Updated         time.Time
-	Finalized       time.Time
+	Updated         int64
+	Finalized       int64
 	Metadata        map[string]string
 	ContentEncoding string
 	CRC32C          *uint32 // Missing for CMEK buckets
 }
 
+func (m MinObject) UpdatedTime() time.Time   { return NSToTime(m.Updated) }
```
kislaykishore: Pointer receivers might be more efficient here.

vedantdas-source: I originally had pointers receivers but for consistency with the other methods on `MinObject` (like `IsUnfinalized` and `HasContentEncodingGzip`), and to allow calling these getters on non-addressable `MinObject` values, we defined `UpdatedTime` and `FinalizedTime` with value receivers instead of pointer receivers.

kislaykishore (dismissed):

PR #4927, opened 2026-07-23
perf(inode): optimized lookupCount struct to remove unnecessary fields

vedantdas-source:
### Description
Every inode (`FileInode`, `dirInode`, `SymlinkInode`, etc.) embeds a `lookupCount` struct to track active FUSE kernel lookups. Previously, this struct was defined as:

```go
type lookupCount struct {
    id        fuseops.InodeID // 8 bytes
    count     uint64          // 8 bytes
    destroyed bool            // 1 byte
}
```
Due to 64-bit compiler alignment padding, the 1-byte boolean padded the struct out to 24 bytes. Furthermore, the `destroyed` boolean was dead code (it was checked, but never set), and the `id` field was redundant because the parent inodes already store their own ID.

#### Changes Made
- **Type Aliasing**: Converted the struct into a pure type alias (`type lookupCount int64`), dropping the footprint to exactly 8 bytes.
- **State Merging**: The active count and "destroyed" state are now merged into a single `int64`. A value `>= 0` represents the active count, and `-1` represents the destroyed state.
- **Redundancy Removal**: Removed the redundant `.Init(id)` calls. The `.Inc(id)` and `.Dec(id)` methods now accept the `fuseops.InodeID` dynamically from the parent inode to preserve existing safety panic messages.
- **Overflow Protection**: Casted the internal count check to `uint64` in `.Dec()` to prevent negative wrap-around overflow bypasses in the event of an abnormally large decrement request.
 
Existing test passes seamlessly.
[b/538386610](b/538386610), [b/539089367](b/539089367)

### Testing details
1. Manual - NA
2. Unit tests - Added `internal/fs/inode/lookup_count_test.go` . Validated exact bounds checking, underflow panics, wrap-around protection (`math.MaxUint64`), and idempotent destruction.
3. Integration tests - NA


internal/fs/inode/lookup_count.go
```diff
@@ -16,48 +16,40 @@ package inode
...
-	if lc.destroyed {
-		panic(fmt.Sprintf("Inode %v has already been destroyed", lc.id))
+func (lc *lookupCount) IncrementLookupCount() {
+	val := atomic.LoadInt64((*int64)(lc))
+	if val == -1 {
+		panic("Inode has already been destroyed")
 	}
-
-	lc.count++
+	atomic.AddInt64((*int64)(lc), 1)
```
kislaykishore: IncrementLookupCount loads count non-atomically before atomic.AddInt64. A concurrent Destroy() (atomic.StoreInt64(-1)) between the check and add results in -1 + 1 = 0, overwriting the destroyed sentinel -1 and resurrecting the inode.
Ditto for "DecrementLookupCount"

Implement Compare-And-Swap (CAS) loop:
```
for {
  old := lc.val.Load();
    if old == -1 {
      panic(...)
  };
  if lc.val.CompareAndSwap(old, old+1) {
    return
  } 
}.

vedantdas-source: Done.

PR #4936, opened 2026-07-25
refactor: change size computation for radix cache

vedantdas-source:
### Description
This PR addresses an issue where the `stat_cache` was computing memory sizes for Radix cache entries using the structural overhead constant derived for the Go map cache (515 bytes). Because the underlying tree implementation in the Radix cache is much more memory efficient, this inaccurately inflated the cache's reported memory footprint and could cause premature LRU evictions. 

Changes include:
- `stat_cache.go`: Updated `NewStatCacheBucketView` to use type checking to determine if the active cache is a Radix cache.
- `stat_cache.go`: Updated the `entry.Size()` function to dynamically add `475` bytes of overhead for Radix cache entries (based on `TestStatCacheMemoryWorkloads` benchmark calculations) instead of the Map cache's `515` bytes. 
- `stat_cache_test.go`: Adjusted cache sizes in unit tests to account for the new accurate memory footprints and ensure capacity-based evictions trigger exactly when expected.

### Testing details
1. Manual - NA
2. Unit tests - Updated existing `Test_ExpiresLeastRecentlyUsed` multi-bucket and single-bucket tests in `stat_cache_test.go` to balance the capacity sizes. 
3. Integration tests - NA

### DIFF
--- a/internal/cache/metadata/stat_cache.go
+++ b/internal/cache/metadata/stat_cache.go
@@ -15,6 +15,7 @@
 package metadata
 
 import (
+	"fmt"
 	"math"
 	"time"
 
@@ -90,9 +91,14 @@ type StatCache interface {
 // For dynamic-mount (mount for multiple buckets), pass bn as bucket-name.
 // For static-mout (mount for single bucket), pass bn as "".
 func NewStatCacheBucketView(sc lru.Cache, bn string) StatCache {
+	isRadix := false
+	if fmt.Sprintf("%T", sc) == "*lru.radixCache" || fmt.Sprintf("%T", sc) == "*lru.arenaRadix" {
+		isRadix = true
+	}
 	return &statCacheBucketView{
 		sharedCache: sc,
 		bucketName:  bn,
+		isRadix:     isRadix,
 	}
 }
 
@@ -109,6 +115,7 @@ type statCacheBucketView struct {
 	// using the same shared lru.Cache object.
 	// It can be empty ("").
 	bucketName string
+	isRadix    bool
 }
 
 // An entry in the cache, pairing an object with the expiration time for the
@@ -119,6 +126,7 @@ type entry struct {
 	expiration int64
 	// Set to true only for implicit directory entries. This flag will always remain false for negative entries and explicit objects.
 	implicitDir bool
+	isRadix     bool
 }
 
 // Size returns the approximate memory-size (resident set size) of the receiver entry.
@@ -135,7 +143,11 @@ type entry struct {
 func (e entry) Size() uint64 {
 	size := uint64(util.UnsafeSizeOf(&e) + util.NestedSizeOfGcsMinObject(e.m))
 	if e.m != nil {
-		size += 515
+		if e.isRadix {
+			size += 475 // Deduced radix cache overhead (Map's 515 minus ~40 bytes saved per entry per benchmark)
+		} else {
+			size += 515
+		}
 	}
 
 	if e.f != nil {
@@ -231,6 +243,7 @@ func (sc *statCacheBucketView) Insert(m *gcs.MinObject, expiration time.Time) {
 	e := entry{
 		m:          mCopy,
 		expiration: timeToUnixNano(expiration),
+		isRadix:    sc.isRadix,
 	}
 
 	if _, err := sc.sharedCache.Insert(name, e); err != nil {
@@ -267,6 +280,7 @@ func (sc *statCacheBucketView) InsertImplicitDir(objectName string, expiration t
 	e := entry{
 		implicitDir: true,
 		expiration:  timeToUnixNano(expiration),
+		isRadix:     sc.isRadix,
 	}
 
 	if _, err := sc.sharedCache.Insert(name, e); err != nil {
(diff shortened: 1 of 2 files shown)


internal/cache/metadata/stat_cache.go
```diff
@@ -90,9 +91,12 @@ type StatCache interface {
 // For dynamic-mount (mount for multiple buckets), pass bn as bucket-name.
 // For static-mout (mount for single bucket), pass bn as "".
 func NewStatCacheBucketView(sc lru.Cache, bn string) StatCache {
+	t := reflect.TypeOf(sc)
+	isRadix := t != nil && (t.String() == "*lru.radixCache" || t.String() == "*lru.arenaRadix")
```
kislaykishore, line 95: Have you considered extending the statCacheBucketView and create a separate struct for stat_cache e.g. statCacheRadixBucketView. You could override the Size() method in that.

internal/cache/metadata/stat_cache.go
```diff
@@ -135,7 +141,16 @@ type entry struct {
...
 	size := uint64(util.UnsafeSizeOf(&e) + util.NestedSizeOfGcsMinObject(e.m))
 	if e.m != nil {
-		size += 515
+		if e.isRadix {
+			// Deduced radix cache overhead.
+			// At a representative scale of 1,000,000 entries, memory benchmarks
+			// (see internal/cache/lru/lru_memory_test.go) show MapLRU adds ~136 bytes/entry
+			// and RadixLRU adds ~94 bytes/entry (saving ~42 bytes).
+			// Subtracting this savings from the map overhead of 515 gives us roughly 475 bytes.
+			size += 475
```
kislaykishore: We multiple these differences with HeapSizeToRssConversionFactor of 2 later on. So, should we subtract 42 or 21?

internal/cache/metadata/stat_cache.go
```diff
@@ -119,6 +124,7 @@ type entry struct {
 	expiration int64
 	// Set to true only for implicit directory entries. This flag will always remain false for negative entries and explicit objects.
 	implicitDir bool
+	isRadix     bool
```
kislaykishore: Every entry in the cache will have this field. Could we get rid of it and have this data at the stat-cache level?

internal/cache/metadata/stat_cache.go
```diff
@@ -135,7 +141,16 @@ type entry struct {
 func (e entry) Size() uint64 {
 	size := uint64(util.UnsafeSizeOf(&e) + util.NestedSizeOfGcsMinObject(e.m))
 	if e.m != nil {
-		size += 515
+		if e.isRadix {
```
kislaykishore: Could you add documentation how do you compute the size of the cache in total including how it takes the tree-structure, folder vs files, positive vs negative entries etc. into account

internal/cache/metadata/stat_cache.go
kislaykishore: Alternatively, externalize the logic of size computation into a function and pass it to the cache constructor depending on whether it's Radix or not. The string comparison looks brittle.

vedantdas-source: i considered the same approach , i had originally planned on cleanly computing size with say a standalone `CalculateEntrySize` function which would calculate size based on the `isRadix` flag and pass it into the cache constructor during mount. But this turned out to be a highly invasive change because the `lru.Cache` implementation strictly relies on the `ValueType` interface (which enforces that cached objects must implement their own `Size() uint64` method), we can't easily bypass the `entry.Size()` method. To avoid a huge blast radius I preferred this approach. This could cause some problems if in the future if we have 2-3 more cache implementations but for the current scenario I think it should be fine. Wdyt ?

internal/cache/metadata/stat_cache.go
vedantdas-source: From a memory perspective this doesn't cost us any additional memory as the `implicitDir` field causes an 8 bit padding so our `isRadix` field comes under that.

internal/cache/metadata/stat_cache.go
vedantdas-source: The benchmarks that I ran uses `runtime.ReadMemStats(&m)` and looks at `m.Alloc` which strictly measures Heap allocations, not RSS so if we subtracted there we would essentially be saying we save 21 bytes on average instead of 42 on the heap. I think we could have a separate conversion factor for Radix cache say `HeapSizetoRssConversionFactorRadix` ? Wdyt?

internal/cache/metadata/stat_cache.go
kislaykishore: But padding does cost memory. It's a trade-off for faster access, right?

vedantdas-source: what i meant by saying it doesn't cost us additional memory is that without the `isRadix` field we have: `m` (8) + `f` (8) + `expiration` (8) + `implicitDir` (1) = 25 bytes which gets padded to 32 bytes (multiple of 8)

With `isRadix`: `m` (8) + `f` (8) + `expiration` (8) + `implicitDir` (1) + `isRadix` (1) = 26 bytes this still gets padded out to 32 bytes. Because we are just utilizing 1 byte of the dead padding space that the compiler was already allocating and wasting, it doesn't increase our memory footprint.

PR #4938, opened 2026-07-25

vedantdas-source:
### Description

#### Changes Made

### Testing details
1. Manual - NA
- `TestMinObject_UpdatedTime`

3. Integration tests - NA

### DIFF
--- a/internal/storage/gcs/object.go
+++ b/internal/storage/gcs/object.go
@@ -16,6 +16,7 @@ package gcs
 
 import (
 	"crypto/md5"
+	"fmt"
 	"time"
 
 	storagev1 "google.golang.org/api/storage/v1"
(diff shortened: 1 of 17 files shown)


kislaykishore (commented):
Please implement String() in MinObject so that the timestamps get correctly printed in logs.

internal/storage/gcs/object.go
```diff
@@ -119,5 +120,50 @@ func (mo MinObject) HasContentEncodingGzip() bool {
...
+	return time.Unix(0, ns).UTC()
+}
+
+func (mo MinObject) UpdatedTime() time.Time {
+	return NSToTime(mo.Updated)
+}
+
+func (mo MinObject) FinalizedTime() time.Time {
+	return NSToTime(mo.Finalized)
+}
```
kislaykishore: The methods for MinObject should be right next to the struct itself. Please move them up.

vedantdas-source: Done

internal/storage/gcs/object.go
```diff
@@ -119,5 +128,42 @@ func (mo MinObject) HasContentEncodingGzip() bool {
...
+	if ns == 0 {
+		return time.Time{}
+	}
+	return time.Unix(0, ns).UTC()
+}
+
+// String implements the fmt.Stringer interface to provide a human-readable
+// representation of the MinObject in logs, converting the int64
+// timestamps back into readable time formats.
+func (mo MinObject) String() string {
```
kislaykishore: You missed to move the String() method.

vedantdas-source: My mistake, missed it.

kislaykishore (approved):

PR #4943, opened 2026-07-26
fix(inode): report symlink size as the length of its target

vyncint:
### Description

`SymlinkInode.Attributes` never assigned its `size` return value, so it returned the zero value and `stat` reported a size of 0 for every symlink.

```go
func (s *SymlinkInode) Attributes(
	ctx context.Context, clobberedCheck bool) (size uint64, mtime time.Time, nlink uint32, err error) {
	mtime = s.mtime
	nlink = 1
	return
}
```

`stat(2)` requires `st_size` of a symlink to be the length of the pathname it contains, without a terminating null byte. The intended semantics were already documented in this same file, on `UpdateSize` a few lines above:

```go
// The size of a symlink is its target's length, not the backing object's size.
```

So the contract was written down; `Attributes` just did not implement it. For a legacy symlink the backing object is empty, which is why the reported size came out as 0 rather than as some other wrong number.

The fix is one line: `size = uint64(len(s.target))`.

Scope check: `fs.getAttributes` is the only caller of `Attributes`, and it assigns the value straight to `attr.Size`. The rename path reads `Source()` (the backing object's size) rather than `Attributes`, so it is unaffected. The change is limited to the size reported to the kernel.

https://github.com/GoogleCloudPlatform/gcsfuse/issues/2273

I posted the root-cause analysis there first, per the dev guide, including the open question below. Happy to close this PR if you would rather handle it yourselves or scope it differently.

### Testing details

1. Manual - NA. Verified through the composite tests below, which exercise a real mount rather than only the inode layer.
2. Unit tests - updated `TestAttributes`, which previously asserted `uint64(0)`. Added `TestAttributesSizeIsTargetLength`, covering the legacy metadata representation with a zero-size backing object, a non-zero-size backing object (so the assertion cannot pass by coincidence), an empty target, and a multi-byte target. Added `TestAttributesSizeForStandardSymlink` for the content-based representation.
3. Integration tests - not run; I do not have a test bucket. The composite suite was run instead: `./internal/fs/...` passes, including a new `SymlinkTest.LinkSizeIsTargetLength` that creates symlinks through the mount and checks `Lstat` reports the target length.

As a negative control I reverted the one-line fix and confirmed all three new tests fail with `actual: 0x0`, so they do test the change rather than pass incidentally.

Six packages fail in my container (`benchmarks/internal/percentile`, `internal/block`, `internal/bufferedread`, `internal/cache/file`, `internal/cache/util`, `internal/gcsx`). I checked these against a clean checkout and the failing set is identical, so they are environmental - the container runs as root, which defeats the permission-based cases - and unrelated to this change.

Yes, in observable behaviour, and it is worth an explicit decision rather than being waved through.

`stat` on a symlink now reports the target length instead of 0. Anything that reads `st_size` for a symlink sees a different number - though 0 was never a value a POSIX-conforming caller could use, and tools like `tar`, `rsync` and `ls -l` expect the target length.

Concretely, four existing assertions pinned the old value and are updated here:

- `internal/fs/inode/symlink_test.go` - `TestAttributes`
- `internal/fs/local_modifications_test.go` - `SymlinkTest.CreateLink`, twice, via `Lstat` and via the readdir entry
- `internal/fs/caching_test.go` - `CachingWithImplicitDirsTest.SymlinksWork`
- `internal/fs/foreign_modifications_test.go` - `ForeignModsTest.Symlink`

My reading is that these were written to match the implementation rather than to encode a deliberate decision, but that judgement is yours. The GCS object size assertion in `CreateLink` (`ExpectEq(0, m.Size)`) is left alone - that one is correctly 0, since a legacy symlink's target lives in metadata and the object really is empty.

### DIFF
--- a/internal/fs/caching_test.go
+++ b/internal/fs/caching_test.go
@@ -395,7 +395,7 @@ func (t *CachingWithImplicitDirsTest) SymlinksWork() {
 	AssertEq(nil, err)
 
 	ExpectEq("bar", fi.Name())
-	ExpectEq(0, fi.Size())
+	ExpectEq(len("foo"), fi.Size())
 	ExpectEq(filePerms|os.ModeSymlink, fi.Mode())
 
 	// Stat the target via the link.
--- a/internal/fs/foreign_modifications_test.go
+++ b/internal/fs/foreign_modifications_test.go
@@ -895,7 +895,7 @@ func (t *ForeignModsTest) Symlink() {
 	AssertEq(nil, err)
 
 	ExpectEq("foo", fi.Name())
-	ExpectEq(0, fi.Size())
+	ExpectEq(len("bar/baz"), fi.Size())
 	ExpectEq(filePerms|os.ModeSymlink, fi.Mode())
 
 	// Read the link.
--- a/internal/fs/inode/symlink.go
+++ b/internal/fs/inode/symlink.go
@@ -232,6 +232,10 @@ func (s *SymlinkInode) Destroy() (err error) {
 
 func (s *SymlinkInode) Attributes(
 	ctx context.Context, clobberedCheck bool) (size uint64, mtime time.Time, nlink uint32, err error) {
+	// POSIX requires st_size of a symlink to be the length of the target path,
+	// excluding the terminating null byte. It is not the size of the backing GCS
+	// object, which is zero for legacy symlinks that keep the target in metadata.
+	size = uint64(len(s.target))
 	mtime = s.mtime
 	nlink = 1
 	return
--- a/internal/fs/inode/symlink_test.go
+++ b/internal/fs/inode/symlink_test.go
@@ -118,9 +118,86 @@ func TestAttributes(t *testing.T) {
 			// Check expected values
 			require.NoError(t, err)
 			assert.Equal(t, uint32(1), nlink)
-			assert.Equal(t, uint64(0), size)
+			assert.Equal(t, uint64(len("target")), size)
 		})
 	}
+}
+
+func TestAttributesSizeIsTargetLength(t *testing.T) {
+	tests := []struct {
+		name     string
+		target   string
+		metadata map[string]string
+		// objectSize is the size of the backing GCS object, which must not be
+		// what Attributes reports.
+		objectSize uint64
+	}{
+		{
+			name:     "LegacySymlink",
+			target:   "/some/fairly/long/target/path",
+			metadata: map[string]string{inode.SymlinkMetadataKey: "/some/fairly/long/target/path"},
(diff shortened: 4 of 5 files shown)


anushka567:
@vyncint  Can you please push an empty commit to retrigger CI (kokoro specifically) ? Thanks!

vyncint:
Done — pushed `5a7a68e`, an empty commit (identical tree to `84c3180`, 0 files changed), so the diff is untouched at 113/-6 across the same 5 files.

The GitHub Actions checks have re-triggered on the new head. Kokoro has not reported yet; if it stays quiet, note the PR carries `execute-integration-tests` but not `kokoro:run` — I left that one alone since applying it is yours to decide. Happy to push another if you need it.

vyncint:
Hi @anushka567, Kokoro is red, but the log is on fusion2. Could you paste the failing test name?

Checked from outside first: no e2e test asserts a symlink's st_size through the mount, and the one that looks like a hit ("Legacy symlink size should be 0" in symlink_suites_test.go) reads the GCS object's own size via objHandle.Attrs, not what gcsfuse reports. linux-tests is green.

4 of 7 recent PRs that actually ran this suite also failed, so it may be a flake, but I'd rather see the name than assume. Happy to push another empty commit if you'd prefer a re-run.

anushka567:
@vyncint This was a recent regression for GCSFuse CI and the fix was merged a while back (commit- `f3eecbc22588ef1da2b5113ef66111efc072fe17`). Can you please rebase the branch on master to pick the changes, thanks!

anushka567:
fyi, the failing test package is stale_handle and the failure is unrelated to the changes made in this PR.

vyncint:
Hi @anushka567, I rebased onto master — bb68de1 now includes f3eecbc. Single commit, diff unchanged at +113/-6 across the same 5 files.

Heads up: the force-push briefly landed on master's tip, which auto-closed the PR. Reopened right away — sorry for the notification noise.

The ci workflow is at action_required, so it needs your approval to run.

vyncint:
Hi @anushka567, could you help me apply kokoro:run again? The label gets consumed on each run, and I can't apply it myself.

vyncint:
Hi @anushka567,

Thanks for applying the label. Everything is green now, including Kokoro.

The rebase picked up f3eecbc, and the stale_handle failure is gone. The diff is
unchanged at +113/-6 across the same 5 files.

Ready for review whenever you have a moment.

anushka567 (approved):
LGTM. Will need codeowner approval for merge.

charith87 (approved):
Relying on Anushka's Review

PR #4945, opened 2026-07-27
refactor(integration_tests): introduce WriterOptions and streamline GCS writer test utilities

vipnydav:
### Description

This PR refactors the GCS writer initialization in integration tests by introducing a `WriterOptions` struct and streamlined helper methods (`NewWriterWithOptions` / `CreateObjectWithOptions`). This reduces boilerplate, removes redundant network calls, and improves write error handling.

Crucially, this PR also ensures that the `RAPID` storage class is explicitly set whenever the appendable API is used. While Zonal buckets are exclusively `RAPID`, Pirlo buckets support both `RAPID` and `STANDARD` classes, but the appendable API is only supported on `RAPID`. This explicit assignment prevents test failures when a bucket's default storage class differs.

### Testing details
1. Manual - Done
2. Unit tests - NA

No

### DIFF
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -203,28 +203,55 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
 	return string(content), nil
 }
 
-// WriterOptions contains optional parameters for creating objects on GCS.
-type WriterOptions struct {
-	StorageClass     string
-	UseAppendableAPI bool
-	FinalizeOnClose  *bool
+// WriterOption is a function that configures writer properties.
+type WriterOption func(*writerConfig)
+
+type writerConfig struct {
+	storageClass     string
+	useAppendableAPI bool
+	finalizeOnClose  bool
 }
 
-// NewWriterWithOptions creates a storage.Writer configured with the given WriterOptions.
-func NewWriterWithOptions(ctx context.Context, o *storage.ObjectHandle, opts WriterOptions) *storage.Writer {
-	wc := o.NewWriter(ctx)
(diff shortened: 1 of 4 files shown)


tools/integration_tests/util/client/storage_client.go
```diff
@@ -203,54 +203,42 @@ func ReadChunkFromGCS(ctx context.Context, client *storage.Client, object string
...
-	// Changes specific to zonal bucket
-	var attrs *storage.BucketAttrs
-	attrs, err = getBucketHandle(client, o.BucketName()).Attrs(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get attributes for bucket %q: %w", o.BucketName(), err)
+// WriterOptions contains optional parameters for creating objects on GCS.
+type WriterOptions struct {
+	StorageClass     string
+	UseAppendableAPI bool
+	FinalizeOnClose  *bool
```
abhishek10004: you can use an enum, which has a default value and then 2 other values for true & false. This can help you detect if anyone has explicitly set this or not.

abhishek10004: something like this:

type FinalizeMode int

const (
    FinalizeDefault FinalizeMode = iota // 0 (Zero-value default)
    FinalizeTrue
    FinalizeFalse
)

abhishek10004: gemini suggested another option as well - https://paste.googleplex.com/5576790805446656. You can check that out as well.

vipnydav: Used the Functional Options Pattern as suggested by Gemini.

abhishek10004 (approved):
LGTM

PR #4947, opened 2026-07-27
feat(native-retries): retry OAuth2/STS token acquisition and Metadata Server readiness during mount

dangarhs:
### Description

This PR adds retry logic for OAuth2/STS token acquisition and Metadata Server (MDS) readiness during mount and runtime. 

* **Retries transient errors:** Extends `determineRetryAction` to classify temporary OAuth2, STS, and MDS errors (e.g., HTTP 400, 408, 429, 500-599) as transient retryable errors.
* **Fails fast on permanent errors:** Stops retrying immediately on permanent authentication errors like HTTP 401 and 403.

b/538079553

### Testing details
1. Manual - Verified error classification and retry loops behavior.
2. Unit tests - Added unit test coverage for classification logic.

No.


internal/storage/storageutil/custom_retry.go
```diff
@@ -76,6 +86,25 @@ func determineRetryAction(err error) retryAction {
...
+		code := retrieveErr.Response.StatusCode
+		if isTransientHTTPStatusCode(code) {
+			return retryTransientMDSError
+		}
+		if code == http.StatusUnauthorized {
+			return retry401
+		}
+		if code == http.StatusForbidden {
+			return retry403
+		}
```
meet2mky: Are these two error mentioned in the doc?

internal/storage/storageutil/custom_retry.go
```diff
@@ -76,6 +86,25 @@ func determineRetryAction(err error) retryAction {
...
+	}
+
+	var metaErr *metadata.Error
+	if errors.As(err, &metaErr) && isTransientHTTPStatusCode(metaErr.Code) {
```
meet2mky: Here 401 and 403 doesn't happen in case of metadata.Error?

internal/storage/storageutil/custom_retry.go
```diff
@@ -33,8 +36,8 @@ type retryAction int
 const (
 	// noRetry indicates the error is not retryable.
 	noRetry retryAction = iota
-	// retryTransient indicates the error is transient and retryable as per Go-SDK retry policy.
-	retryTransient
+	// retryTransientMDSError indicates the error is transient and retryable as per Go-SDK or MDS/OAuth retry policy.
```
meet2mky: Add separate enum for the MDS Transient error.

internal/storage/storageutil/custom_retry.go
dangarhs: Yeah, there is no authentication layer for MDS and also 403 is permanent when the request header doesnt contain a specific header.

meet2mky: If there is no authentication layer then which layer sends the 429 and 5xx here?

internal/storage/storageutil/custom_retry.go
```diff
@@ -51,6 +56,13 @@ const (
 
 const ErrStrBucketNotExist = "bucket does not exist"
 
+func isTransientHTTPStatusCode(code int) bool {
+	return code == http.StatusBadRequest ||
+		code == http.StatusRequestTimeout ||
+		code == http.StatusTooManyRequests ||
+		(code >= 500 && code <= 599)
+}
+
```
meet2mky, line 77: Let's rename this method to isTransientMDSError(err error) bool  {

}
and put logic for both the retrieveError and metadata.Error in this.

internal/storage/storageutil/custom_retry.go
```diff
@@ -97,7 +128,7 @@ func determineRetryAction(err error) retryAction {
 // This method is same as ShouldRetry except it doesn't add warning logs.
 func ShouldRetryWithoutLogging(err error) bool {
 	switch determineRetryAction(err) {
-	case retryTransient, retry401, retryUnauthenticated:
+	case retryTransient, retryTransientMDSError, retry401, retryUnauthenticated:
```
meet2mky: Why do we need this enum here because it's by default covered in ShouldRetryOnMount which we would use wherever the enable mount retries scenario is required?

internal/storage/storageutil/custom_retry.go
dangarhs: Done.

internal/storage/storageutil/custom_retry.go
dangarhs: By auth layer, I mean MDS doesnt authorize using oauth right? so, 401 can never be returned.

internal/storage/storageutil/custom_retry.go
dangarhs: Ack. Missed while reverting the universeDomain retries.

internal/storage/storageutil/custom_retry.go
dangarhs: as discussed offline, I have removed these 2 errors as they are not present in the doc but while testing, they seem to be genuine error scenarios. Lets discuss separately.

internal/storage/storageutil/custom_retry.go
dangarhs: Done.

meet2mky (approved):
Looks good.

internal/storage/storageutil/custom_retry.go
```diff
@@ -51,6 +56,25 @@ const (
...
+
+	if errors.As(err, &retrieveErr) && retrieveErr.Response != nil {
+		code = retrieveErr.Response.StatusCode
+	} else if errors.As(err, &metaErr) {
+		code = metaErr.Code
+	} else {
+		return false
+	}
+
+	return code == http.StatusBadRequest ||
```
abhishek10004, line 72: Why is bad request a retryable error?

internal/storage/storageutil/custom_retry.go
```diff
@@ -76,6 +100,10 @@ func determineRetryAction(err error) retryAction {
 		}
 	}
 
+	if isTransientMDSError(err) {
+		return retryTransientMDSError
```
abhishek10004, line 104: We're returning a new action but I don't see any handling based on this. Are we just doing this so that some of the errors for which we were returning NoRetry action now returns a different action?

dangarhs: Yes, exactly. Scope is to only retry for mount-time(ShouldRetryOnMount), while runtime(ShouldRetryWithoutLogging) still evaluates to false for transient errors.

internal/storage/storageutil/custom_retry.go
dangarhs: 400 can occur during OAuth/STS token exchange due to transient startup delays (like Identity provider sync delays). The error is observed as `invalid_grant`

abhishek10004: Got it. Also, one other question: why are we checking for 404, 429 & >= 500 && <= 599 in this as well. These are all retryable in general and shouldRetry handles it correctly and returns retryTransient.
Either remove them from here, or maybe put a comment above isTransientMDSError specifying that this should be called after shouldRetry

abhishek10004 (approved):
LGTM. Left a comment. Please check.

internal/storage/storageutil/custom_retry.go
dangarhs: `ShouldRetry()` only handles `*googleapi.Error` and gRPC status errors—it does not handle `*oauth2.RetrieveError` or `*metadata.Error`, so we need to check 429 & 5xx separately here. (Also, 404 is not included in isTransientMDSError and returns noRetry).

PR #4948, opened 2026-07-28
feat: Added httpmtls option for client-protocol flag

vadlakondaswetha:
Today GCSFuse creates a custom http client which doesn't support mtls. The go storage sdk's client creation support mtls. Hence for users to do mtls via GCSFuse, added a new option under client protocol called httpwithmtls. When users sets this:

- It will bypass GCSFuse custom client creation and uses go storage sdks http client.
- Explicit enablement of DNS caching via GCSFuse will not work. OS level enablement is not effected.
- Http transport level tracing will not work since we can't override the transport settings as http client is created inside go-storage sdk.
- By default go storage sdk tries to create http2 client and falls back to http1 if the server doesn't support it.

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - Added


cfg/types.go
```diff
@@ -39,19 +39,20 @@ func (o Octal) MarshalText() ([]byte, error) {
...
 type Protocol string
 
 const (
-	HTTP1 = "http1"
-	HTTP2 = "http2"
-	GRPC  = "grpc"
+	HTTP1        = "http1"
+	HTTP2        = "http2"
+	GRPC         = "grpc"
+	HTTPWithMtls = "httpwithmtls"
```
abhishek10004: nit:http-with-mlts ot http-mlts?

tools/integration_tests/test_config.yaml
```diff
@@ -94,6 +94,13 @@ operations:
     test_bucket: "${BUCKET_NAME}"
     only_dir: "${ONLY_DIR}"
     configs:
+      - flags:
+          - "--client-protocol=httpwithmtls"
+        compatible:
+          flat: true
+          hns: true
+          zonal: false
+        run_on_gke: true
```
abhishek10004, line 103: I'm not sure if this will work directly on GKE. Might be better to test it first and then only enable it on GKE.

abhishek10004 (dismissed):
LGTM. Left a minor comment for the param value.

1
abhishek10004, line 1: Seems like an extra file getting checked in. Could you please remove this?

abhishek10004 (approved):
LGTM

PR #4952, opened 2026-07-28
feat(otel logger): Add OpenTelemetry (OTEL) logging support

abhishek10004:
## Description

**Objective:** 
Introduce OpenTelemetry (OTEL) logging support to GCSFuse, providing a more robust, standardized, and observable logging infrastructure.

**Key Changes:**
* **OTEL Logger Integration:** Introduced the necessary configuration (`cfg/params.yaml`, `cfg/config.go`) and dependencies to enable OpenTelemetry logging. 
* **OTEL Exporters Setup:** Initialized and configured the OTEL log exporters (`internal/monitor/otelexporters.go`) to handle log telemetry data, including adding local testing support.
* **Logger Optimization:** Updated the internal logger (`internal/logger/logger.go`) to seamlessly route logs through OTEL while actively minimizing heap allocations for improved performance.
* **Command Line Hookup:** Integrated the new OTEL logging configuration and lifecycle into the main startup flow (`cmd/legacy_main.go`).

**Testing:**
* Added comprehensive unit tests for the updated logger and OTEL exporters (`internal/logger/logger_test.go`, `internal/monitor/otelexporters_test.go`).
* Implemented extensive integration tests specifically tailored to validate OTEL logging end-to-end (`tools/integration_tests/otel_logs/`).

b/535703440

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Done


cfg/config.go
```diff
@@ -1939,6 +1935,14 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
+	if err := v.BindPFlag("logging.otel-logging-enabled", flagSet.Lookup("otel-logging-enabled")); err != nil {
```
vadlakondaswetha: our name convention is generally enable-otel-logging

cfg/params.yaml
```diff
@@ -864,6 +854,13 @@ params:
     default: false
     hide-flag: true
 
+  - config-path: "logging.enable-otel-logging"
+    flag-name: "enable-otel-logging"
```
kislaykishore: Let's have the flags as experimental to begin with.

internal/logger/logger.go
```diff
@@ -294,12 +298,60 @@ func (f *loggerFactory) createJsonOrTextHandler(writer io.Writer, levelVar *slog
...
+	} else if f.sysWriter != nil {
+		localHandler = f.createJsonOrTextHandler(f.sysWriter, levelVar, prefix)
+	} else {
+		localHandler = f.createJsonOrTextHandler(os.Stdout, levelVar, prefix)
 	}
 
-	if f.sysWriter != nil {
-		return f.createJsonOrTextHandler(f.sysWriter, levelVar, prefix)
+	if f.enableOtelLogging {
+		otelHandler := otelslog.NewHandler("gcsfuse")
```
kislaykishore, line 312: OTel logs remain statically at INFO (dropping 100% of DEBUG/TRACE logs and leaking INFO logs when --severity=ERROR
```suggestion
		otelHandler := otelslog.NewHandler("gcsfuse", otelslog.WithLevel(levelVar))
```

internal/monitor/otelexporters.go
```diff
@@ -205,3 +217,96 @@ func getResource(ctx context.Context, mountID string) (*resource.Resource, error
...
+	}
+
+	exporter, err := otlploghttp.New(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	processor := log.NewBatchProcessor(
+		exporter,
+		log.WithExportMaxBatchSize(2000),
```
kislaykishore, line 355: Batch processor queue starvation: MaxQueueSize defaults to 2048. With WithExportMaxBatchSize(2000), the queue headroom is only 48 records, causing silent log drops during FUSE I/O bursts.

Configure `log.WithMaxQueueSize(16384)` and `log.WithExportTimeout(10*time.Second)`.

internal/monitor/otelexporters.go
```diff
@@ -205,3 +217,96 @@ func getResource(ctx context.Context, mountID string) (*resource.Resource, error
...
+	}
+
+	// Use authenticated client for GCP endpoints.
+	if strings.Contains(endpoint, "googleapis.com") {
+		ts, err := auth.GetTokenSourceWithScope(ctx, string(authConfig.KeyFile), authConfig.TokenUrl, authConfig.ReuseTokenFromUrl, "https://www.googleapis.com/auth/logging.write")
+		if err != nil {
+			logger.Errorf("Error getting GCP authenticated token source for logs: %v", err)
+			return nil, err
+		}
+		client := oauth2.NewClient(ctx, ts)
```
kislaykishore, line 338: oauth2.NewClient creates an HTTP client with no timeout potential indefinite hangs. Add timeout.

internal/auth/auth.go
```diff
@@ -91,8 +91,17 @@ func GetTokenSource(
 	tokenUrl string,
 	reuseTokenFromUrl bool,
 ) (tokenSrc oauth2.TokenSource, err error) {
-	// Create the oauth2 token source.
-	const scope = storagev1.DevstorageFullControlScope
+	return GetTokenSourceWithScope(ctx, keyFile, tokenUrl, reuseTokenFromUrl, storagev1.DevstorageFullControlScope)
+}
+
+// GetTokenSourceWithScope generates the token-source for a specific scope.
+func GetTokenSourceWithScope(
```
kislaykishore, line 98: Could you verify if this works on GKE with host-network?

internal/monitor/otelexporters.go
```diff
@@ -205,3 +217,96 @@ func getResource(ctx context.Context, mountID string) (*resource.Resource, error
...
+func SetupOTelLogExporter(ctx context.Context, endpoint string, mountID string, authConfig cfg.GcsAuthConfig, configuredProjectID string) (common.ShutdownFn, error) {
+	projectID := getProjectID(ctx, authConfig, configuredProjectID)
+	res, err := getResource(ctx, mountID)
+	if err != nil {
+		logger.Errorf("Error while fetching resource for logs: %v", err)
+		return nil, err
+	}
+
+	if projectID != "" {
+		projRes, _ := resource.New(ctx, resource.WithAttributes(attribute.String("gcp.project_id", projectID)))
```
kislaykishore: Include resource.WithSchemaURL(semconv.SchemaURL) and semconv.CloudAccountIDKey.String(projectID).
Don't use raw strings such as `gcp.project_id`

tools/integration_tests/otel_logs/setup_test.go
```diff
@@ -0,0 +1,191 @@
...
+func TestMain(m *testing.M) {
+	setup.ParseSetUpFlags()
+
+	if err := startMockServer(); err != nil {
+		log.Fatalf("Failed to start mock server: %v", err)
+	}
+	defer stopMockServer()
+
+	// Create a temp config file for gcsfuse
+	tempFile, err := os.CreateTemp("", "otel_config_*.yaml")
```
kislaykishore, line 144: While the file itself is deleted, the file-descriptor is not closed.

internal/logger/logger.go
abhishek10004: Added level in dual handler and using that to filter logs

internal/monitor/otelexporters.go
abhishek10004: Used a queue size of 8K and timeout of 10s

internal/auth/auth.go
abhishek10004: I've just refactored GetTokenSource to create GetTokenSourceWithScope and taking scope as a parameter(instead of hardcoded one). Since, GetTokenSource was already there, this should also work.

internal/monitor/otelexporters.go
```diff
@@ -205,3 +217,104 @@ func getResource(ctx context.Context, mountID string) (*resource.Resource, error
...
+		logger.Errorf("Error while fetching resource for logs: %v", err)
+		return nil, err
+	}
+
+	if projectID != "" {
+		projRes, _ := resource.New(ctx,
+			resource.WithSchemaURL(semconv.SchemaURL),
+			resource.WithAttributes(semconv.CloudAccountIDKey.String(projectID)),
+		)
+		res, err = resource.Merge(res, projRes)
```
kislaykishore, line 316: Could you double check if this "Merge" statement is not throwing an error? You are merging resources from two different schemas.

internal/monitor/otelexporters.go
```diff
@@ -205,3 +217,104 @@ func getResource(ctx context.Context, mountID string) (*resource.Resource, error
...
+		if err != nil {
+			logger.Errorf("Error getting GCP authenticated token source for logs: %v", err)
+			return nil, err
+		}
+		client := oauth2.NewClient(ctx, ts)
+		client.Timeout = 30 * time.Second
+		opts = append(opts, otlploghttp.WithHTTPClient(client))
+	}
+
+	exporter, err := otlploghttp.New(ctx, opts...)
```
kislaykishore, line 343: What if the permission is missing - will the code keep retrying indefinitely or will it stop after some tries. If not, this could spam logs a lot. For metrics we have implementation `permissionAwareExporter`. Perhaps we should do something similar here.

internal/logger/logger.go
```diff
@@ -294,12 +298,70 @@ func (f *loggerFactory) createJsonOrTextHandler(writer io.Writer, levelVar *slog
...
+}
+
+// dualHandler dispatches logs to a local handler and an OpenTelemetry handler.
+type dualHandler struct {
+	local    slog.Handler
+	otel     slog.Handler
+	levelVar *slog.LevelVar
+}
+
+func (d *dualHandler) Enabled(ctx context.Context, level slog.Level) bool {
```
kislaykishore, line 325: Is this method redundant. Won't the check in line#195 be sufficient?

internal/logger/logger.go
```diff
@@ -294,12 +298,70 @@ func (f *loggerFactory) createJsonOrTextHandler(writer io.Writer, levelVar *slog
...
+		err2 = d.otel.Handle(ctx, r)
+	}
+
+	if err1 != nil && err2 != nil {
+		return fmt.Errorf("local err: %v, otel err: %v", err1, err2)
 	}
+	if err1 != nil {
+		return err1
+	}
+	return err2
```
kislaykishore: Use errors.Join to join errors?

internal/logger/logger.go
```diff
@@ -294,12 +298,70 @@ func (f *loggerFactory) createJsonOrTextHandler(writer io.Writer, levelVar *slog
...
+
+func (d *dualHandler) Enabled(ctx context.Context, level slog.Level) bool {
+	if level < d.levelVar.Level() {
+		return false
+	}
+	return d.local.Enabled(ctx, level) || d.otel.Enabled(ctx, level)
+}
+
+func (d *dualHandler) Handle(ctx context.Context, r slog.Record) error {
+	if r.Level < d.levelVar.Level() {
```
kislaykishore, line 333: Is this check redundant?

internal/logger/logger.go
kislaykishore:
```suggestion
	return errors.Join(
        d.local.Handle(ctx, r),
        d.otel.Handle(ctx, r),
    )
```

internal/logger/logger.go
abhishek10004: DualHandler implements Slog.Handler interface and hence needs to implement Enabled function

internal/logger/logger.go
abhishek10004: yes, it's a defensive check in case someone directly calls Handle without Enabled (bypassing the sLogger front end interface)

internal/logger/logger.go
abhishek10004: I've used errors.Join but have still kept separate logging for the 2 handlers for ease of debugging

internal/monitor/otelexporters.go
```diff
@@ -205,3 +258,111 @@ func getResource(ctx context.Context, mountID string) (*resource.Resource, error
...
+			}
+		}
+	} else if authConfig.TokenUrl == "" {
+		creds, err := google.FindDefaultCredentials(ctx)
+		if err == nil && creds.ProjectID != "" {
+			return creds.ProjectID
+		}
+	}
+
+	if envProj := os.Getenv("GOOGLE_CLOUD_PROJECT"); envProj != "" {
```
kislaykishore, line 288: Are we propagating this environment variable - will this work in background mode?

internal/monitor/otelexporters.go
```diff
@@ -205,3 +258,111 @@ func getResource(ctx context.Context, mountID string) (*resource.Resource, error
...
+				attribute.String("gcp.project_id", projectID),
+			),
+		)
+		res, err = resource.Merge(res, projRes)
+		if err != nil {
+			logger.Errorf("Error merging project ID into resource: %v", err)
+		}
+	}
+
+	opts := []otlploghttp.Option{otlploghttp.WithEndpoint(endpoint)}
```
kislaykishore: Consider adding `otlploghttp.WithCompression(otlploghttp.GzipCompression)` to reduce n/w payload.

kislaykishore (approved):

PR #4958, opened 2026-07-28
feat(cfg): setup config.proto code generation and privacy scrubber

PranjalC100:
### Description
This pull request introduces automated Protobuf schema generation (`cfg/config.proto`) for gcsfuse configuration and telemetry, deriving the schema directly from `cfg/params.yaml`.
#### Key Highlights:
* **Flat Protobuf Schema Generation**: Uses `config_proto.tpl` and updates `tools/config-gen/main.go` to generate a flat `message Config` in `cfg/config.proto` under package `gcsfuse.cfg.v1` (`go_package: github.com/googlecloudplatform/gcsfuse/v3/cfg/pb`).
* **Explicit Tag Validation & Reserved Tags**: Proto tags are statically defined in `params.yaml`. Fields are sorted strictly by `proto-tag`, and retired tags are exported as `reserved` fields to guarantee schema stability and backwards compatibility across versions.
* **Telemetry Privacy Scrubber**: In `tools/config-gen/parser.go`, high-risk text/path types (`string`, `resolvedPath`, `[]string`) are automatically converted into safe boolean presence flags (e.g., `is_<field>_set`), preventing PII leakage while retaining telemetry utility.
* **Low-Risk Enum Preservation**: Custom string enums (`protocol`, `directPathStrategy`, `logSeverity`) are retained as strings for maintainability and readability without requiring integer mapping registries.
* Bounded metadata (`profile`, `machine-type`) are whitelisted as `string` fields rather than scrubbed to booleans, enabling workload profile and GCE node classification in Bigstore telemetry without exposing freeform user data.

### Testing details
1. **Manual**: Verified generated `cfg/config.proto` matches formatting, proto3 syntax, and field tag ordering.
2. **Unit tests**: Verified unit tests in `tools/config-gen`:
   - `parser_test.go`: `TestPopulateProtoMetadataExhaustive`
   - `type_template_data_gen_test.go`: `TestComputeProtoFields`, `TestFormatReservedTags`
   All existing and new tests pass with `go test -race ./tools/config-gen/...`.
3. **Integration tests**: N/A (code generation tool and schema definitions only; all presubmit checks pass).

### Bug: 
[549008004](https://b.corp.google.com/issues/549008004)
### Backward Compatibility
Enforced by design. Older flag definitions (even deprecated ones) retain their explicit tags to prevent tag collisions and schema breaks in legacy GCSFuse versions.


cfg/config.proto
```diff
@@ -0,0 +1,252 @@
...
+message GcsAuthConfig {
+  bool anonymous_access = 1;
+  bool is_key_file_set = 2;
+  bool reuse_token_from_url = 3;
+  bool is_token_url_set = 4;
+}
+
+message GcsConnectionConfig {
+  bool is_billing_project_set = 1;
+  int32 client_protocol = 2;
```
PranjalC100: Hi @kislaykishore 
Currently, this is generated as an **int32**. I'd like to propose serializing this (and other 2 custom string enums like **log-severity** and **grpc-path-strategy**) directly as strings rather than mapping them to integer variants (e.g., grpc=1, http1=0).
While integer mapping theoretically saves ~10-15 bytes in the final payload, it introduces the overhead of maintaining a custom mapping registry across the codebase. Serializing them natively as strings is significantly more readable, easier to maintain, and still keeps us well within our payload constraints. What do you think?

kislaykishore: Sounds good. We can use strings as well. Does our PDD allow for this?

tools/config-gen/type_template_data_gen_test.go
```diff
@@ -0,0 +1,165 @@
...
+`
+	err := os.WriteFile(protoFilePath, []byte(protoContent), 0644)
+	require.NoError(t, err)
+
+	existingTags, err := parseExistingProtoTags(protoFilePath)
+	require.NoError(t, err)
+
+	params := []Param{
+		// Existing root flag
+		{
```
kislaykishore, line 35: Why not exclude app-name? One that it's a string field and I think we discussed to not send string fields. Secondly, I think it's already included in the user-agent. It helps to not require us to deserialize it and immediately infer the app name for analysis.

tools/config-gen/templates/config_proto.tpl
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// GENERATED CODE - DO NOT EDIT MANUALLY.
+
+syntax = "proto3";
+
+package cfg;
```
kislaykishore: "cfg" is quite a conflicting choice. Instead keep the package a bit more specific: gcsfuse.cfg.v1. You can use a go-pkg  option.

cfg/config.proto
PranjalC100: Yes the PDD allows this. Quoting from the PDD:
* **Low-Risk Primitive & Custom Types (Directly Retained):** These types represent bounded numeric thresholds, binary toggles, or predefined domain enums. They don’t contain customer-identifiable metadata and are safe to transmit in their native binary formats.
* **Example:** bool, octal ,duration ,float64. Custom Enums (client-protocol, directPathStrategy, logSeverity)

tools/config-gen/type_template_data_gen_test.go
PranjalC100: I was thinking of still keeping it in the UserAgent field. But it doesn't make that much sense to have it in the config, since it only indicates a boolean presence. 
I'll exclude `app-name` from this config.

tools/config-gen/templates/config_proto.tpl
PranjalC100: Makes sense. I will update the protobuf package to `gcsfuse.cfg.v1`. I am already utilizing the go_package option`github.com/googlecloudplatform/gcsfuse/v3/cfg/pb` Will update them. (line 21)

tools/config-gen/parser.go
```diff
@@ -67,9 +69,57 @@ func parseParamsYAMLStr(paramsYAMLStr string) (paramsYAML ParamsYAML, err error)
...
+	"[]string":     true,
+}
+
+// populateProtoMetadata maps YAML types to Protobuf types, converting strings to booleans for telemetry privacy.
+func populateProtoMetadata(params []Param) error {
+	// Assign proto type and field name
+	for i := range params {
+		param := &params[i]
+		if param.FlagName == "app-name" {
+			continue
```
kislaykishore: Let's not hardcode it. It's okay to send an extra param to keep the code and analysis less complex.

cfg/config.proto
```diff
@@ -0,0 +1,256 @@
...
+// GENERATED CODE - DO NOT EDIT MANUALLY.
+
+syntax = "proto3";
+
+package gcsfuse.cfg.v1;
+
+option go_package = "github.com/googlecloudplatform/gcsfuse/v3/cfg/pb";
+
+
+message CloudProfilerConfig {
```
kislaykishore: The tag numbers require old params.yaml to be present and cannot be generated from scratch later on. This makes the design brittle. I'd rather we hardcode the tag numbers in the params.

tools/config-gen/parser.go
PranjalC100: Okay, will remove this then.

cfg/config.proto
PranjalC100: Implemented this in as per the offline discussion: https://github.com/GoogleCloudPlatform/gcsfuse/pull/5036. Rebasing this PR.

cfg/config.proto
```diff
@@ -0,0 +1,183 @@
...
+  bool is_metrics_experimental_otel_metrics_endpoint_set = 126;
+  bool is_metrics_experimental_otel_metrics_project_id_set = 127;
+  int32 metrics_prometheus_port = 128;
+  int64 metrics_stackdriver_export_interval = 129;
+  bool metrics_use_new_names = 130;
+  int32 metrics_workers = 131;
+  bool is_mount_id_set = 132;
+  int32 mrd_pool_size = 133;
+  bool is_only_dir_set = 134;
+  bool is_profile_set = 135;
```
kislaykishore, line 159: What's our take on string fields like logging.format and profile? Do we want to hide them as well?

cfg/config.proto
```diff
@@ -0,0 +1,183 @@
...
+  int64 metadata_cache_deprecated_stat_cache_ttl = 110;
+  int64 metadata_cache_deprecated_type_cache_ttl = 111;
+  bool metadata_cache_enable_metadata_prefetch = 112;
+  bool metadata_cache_enable_nonexistent_type_cache = 113;
+  bool metadata_cache_experimental_enable_optimized_metadata_cache = 114;
+  bool is_metadata_cache_experimental_metadata_prefetch_on_mount_set = 115;
+  int32 metadata_cache_metadata_prefetch_entries_limit = 116;
+  int32 metadata_cache_metadata_prefetch_max_workers = 117;
+  int32 metadata_cache_negative_ttl_secs = 118;
+  int32 metadata_cache_stat_cache_max_size_mb = 119;
```
kislaykishore, line 143: Consider using sint64 for integers. It could be more memory-efficient for negative numbers.

PranjalC100: Done. Migrated all config integer fields to sint64.

cfg/config.proto
PranjalC100: I have whitelisted `profile` and `machine-type` in `config-gen` to emit as string since they are bounded enums/hardware metadata, while keeping freeform text scrubbed as booleans.

tools/config-gen/parser.go
```diff
@@ -84,6 +89,59 @@ func isValidDataType(dt string) bool {
 	return slices.Contains(supportedDataTypes, dt)
 }
 
+var highRiskTextTypes = map[string]bool{
+	"string":       true,
+	"resolvedPath": true,
+	"[]string":     true,
+}
+
+var whitelistedTextConfigPaths = map[string]bool{
```
kislaykishore: AllowList is more inclusive.
```suggestion
var allowlistedTextConfigPaths = map[string]bool{
```

kislaykishore (approved):

PR #4960, opened 2026-07-29
fix(integration_tests): fallback to MOUNTED_DIR environment variable correctly

PranjalC100:
### Description
This pull request fixes the environment variable parsing for the mounted directory in the integration test suite. While `test_config.yaml` successfully evaluated the `MOUNTED_DIR` environment variable, the fallback logic inside `setup.go` did not cleanly map it to the underlying configuration structs when invoked locally or when the flag/env var was passed directly. 

Changes included:
* Enabled direct fallback to `os.Getenv("MOUNTED_DIR")` within the `setup.MountedDirectory()` global flag parser, mirroring the existing fallback behavior for `TestBucket()`.
* Updated `setup.SetGlobalVars` to explicitly fallback `cfg.GKEMountedDirectory` to the resolved command-line flag/environment string if it was absent in the YAML config.
* Added clarifying documentation comments to the flag registrations to explicitly denote environment variable support and order of precedence.

N/A

### Testing details
1. Manual - Verified that the integration tests can dynamically parse the `MOUNTED_DIR` environment variable through `go test` and fallback correctly in local executions.
2. Unit tests - N/A
3. Integration tests - N/A

N/A


meet2mky (approved):
LGTM

PR #4964, opened 2026-07-30
fix: Differentiating between not found and estale during fileopen

vadlakondaswetha:
Handling Not found seperately to avoid getting another lookup call during fileopen.

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - Added


internal/fs/inode/file.go
```diff
@@ -273,7 +273,7 @@ func (f *FileInode) checkInvariants() {
 // | Other GCS Stat Error                         | N/A          | Other     | (nil, false, <GCS Error>)                   |
 //
 // LOCKS_REQUIRED(f.mu)
-func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool, includeExtendedObjectAttributes bool) (o *gcs.Object, b bool, err error) {
+func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool, includeExtendedObjectAttributes bool) (o *gcs.Object, b bool, isNotFound bool, err error) {
```
abhishek10004: Let's update the comment above as well.

Also, instead of b, let's use isClobbered in the return param list as welll

internal/fs/inode/file.go
```diff
@@ -956,7 +963,7 @@ func (f *FileInode) fetchLatestGcsObject(ctx context.Context) (*gcs.Object, erro
 	// properties and using that when object is synced below. StatObject by
 	// default sets the projection to full, which fetches all the object
 	// properties.
-	latestGcsObj, isClobbered, err := f.clobbered(ctx, true, true)
+	latestGcsObj, isClobbered, _, err := f.clobbered(ctx, true, true)
```
abhishek10004, line 966: Since we now have isNotFound as well, it will be better to use that and return proper error from here.

abhishek10004 (dismissed):
LGTM. Left a few minor comments.

internal/fs/inode/file.go
vadlakondaswetha: ESTALE for network based filesystems covers ENOENT also . So i dont see a need to handle ESTALE & ENOENT at all places unless both the errors require different handling (eg. openFile) https://screenshot-v2.corp.google.com/3kvMH4ic7UnAwi5

abhishek10004 (approved):
LGTM

PR #4966, opened 2026-07-30
feat: Adding new directpath fallback stragegy

vadlakondaswetha:
Today we only support using gRPC with directPath and fallback to http when directPath is not available.

Added a new option to use gRPC Cloud path when direct path is not available.

### Testing details
1. Manual - Done
2. Unit tests - Done
3. Integration tests - NA

### DIFF
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -690,7 +690,7 @@ params:
     type: "directPathStrategy"
     usage: >-
       Strategy for DirectPath connectivity when client-protocol=grpc.
-      Options: 'direct-path-only' (fail if unavailable), 'direct-path-with-fallback' (always fallback to HTTP/1 when direct path is not available).
+      Options: 'direct-path-only' (fail if unavailable), 'direct-path-with-fallback' (always fallback to HTTP/1 when direct path is not available), 'direct-path-with-grpc-fallback' (fallback to standard gRPC when direct path is not available).
     default: "direct-path-with-fallback"
     hide-flag: true
 
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -690,7 +690,7 @@ params:
     type: "directPathStrategy"
     usage: >-
       Strategy for DirectPath connectivity when client-protocol=grpc.
-      Options: 'direct-path-only' (fail if unavailable), 'direct-path-with-fallback' (always fallback to HTTP/1 when direct path is not available), 'direct-path-with-grpc-fallback' (fallback to standard gRPC when direct path is not available).
+      Options: 'direct-path-only' (fail if unavailable), 'direct-path-with-fallback' (always fallback to HTTP/1 when direct path is not available), 'direct-path-with-grpc-fallback' (fallback to standard gRPC cloud path when direct path is not available).
     default: "direct-path-with-fallback"
     hide-flag: true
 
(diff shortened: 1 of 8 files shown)


cfg/params.yaml
```diff
@@ -690,7 +690,7 @@ params:
     type: "directPathStrategy"
     usage: >-
       Strategy for DirectPath connectivity when client-protocol=grpc.
-      Options: 'direct-path-only' (fail if unavailable), 'direct-path-with-fallback' (always fallback to HTTP/1 when direct path is not available).
+      Options: 'direct-path-only' (fail if unavailable), 'direct-path-with-fallback' (always fallback to HTTP/1 when direct path is not available), 'direct-path-with-grpc-fallback' (fallback to standard gRPC cloud path when direct path is not available).
```
abhishek10004, line 693: Should we update this to direct-path-with-http-fallback now since we already have a separate grpc fallback strategy? Do we know if anyone outside has started using that older option value?

vadlakondaswetha: I thought about it. It becomes tricky when we start recommending to customers. Depending on the version they are using they have to pass direct-path-with-fallback or direct-path-with-http-fallback. On the other hand, its a hidden flag. so should we take the risk?

Other option would be to backport all version?

cfg/params.yaml
```diff
@@ -690,7 +690,7 @@ params:
     type: "directPathStrategy"
     usage: >-
```
abhishek10004, line 691: Should we also mention that this is only used for non rapid?

internal/storage/storage_handle.go
```diff
@@ -496,14 +508,28 @@ func (sh *storageClient) createNonBidiGRPCClientWithHttpFallback(ctx context.Con
...
+	// Check if the error is due to DP failure. If not, return it.
+	if !isDirectPathFailure(err) {
+		logger.Errorf("Grpc client creation failed with non-DirectPath error: %v", err)
+		return nil, err
+	}
+
 	// We will reach here when we failed to create a grpc client with direct path.
-	// Decide whether to create a http client based on grpPathStrategy param.
+	// Decide what to do based on grpPathStrategy param.
+	// Ideally this condition will never be true, since we are checking for non-DP error above.
```
abhishek10004, line 526: +1. This condition would never be true. Let's remove this check.

internal/storage/storage_handle.go
```diff
@@ -560,3 +586,10 @@ func (sh *storageClient) BucketHandle(ctx context.Context, bucketName string, bi
 
 	return
 }
+
+func isDirectPathFailure(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), directPathVerificationErrorPrefix)
```
abhishek10004, line 594: This is fine for now but we should define a new error and then use Errors.Is or Errors.As

cfg/params.yaml
abhishek10004: Yeah, I would not like to backport to all versions.

Let's go with this for now. Meanwhile, we can check the logs and if we don't see any external use of this flag, then let's make the change.

abhishek10004 (approved):
LGTM. Left a few minor comments.

PR #4969, opened 2026-07-30
fix(integration_tests on gke): disable mounting test suite on GKE for httpmtls protocol

meet2mky:
### Description
Mounting integration test package is not suppose to run on GKE as it literally tries to mount GCSFuse which is not possible on GKE environment. This was accidentally enabled in #4948 for httpmtls protocol.

image

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


abhishek10004 (approved):
LGTM

Issue #4976, opened 2026-08-03
Feature Request: Directory write guard (ENOSPC above max-dir-entries), streaming readdir, and documented limits — remaining scope from #4696


chaitanyapantheor:
## Summary

Follow-up to #4696, which was closed as completed but is only partially addressed. #4736 (still open) implements just the warning-threshold log (feature #2), as a hidden experimental flag. This issue tracks the remaining, substantive scope from #4696.

### 1. Write guard — block writes above a configurable entry limit
Reject `O_CREAT` on a directory at/above a configurable `max-dir-entries` with `ENOSPC`, plus a structured `WARNING` log. This is the corruption-prevention mechanism the original request centered on: read-path mitigations don't stop a directory from growing further into the degradation/corruption zone (~50k read degradation, 100k+ corruption risk). Returning `ENOSPC` ("no space left on device") gives applications a standard, actionable signal.

```yaml
file-system:
  max-dir-entries: 50000   # block new creates + warn when the parent dir hits this
```

### 3. Streaming / incremental readdir
Stream `ListObjects` entries back to the kernel incrementally instead of buffering the full result, to cap peak memory and latency at 100k+ entries. (Larger change; likely its own PR.)

### 4. Documented limits
Publish a supported maximum directory size (entries per prefix) with mitigation guidance (subdirectory sharding, HNS buckets, etc.).

## Status
- Feature #2 (warning-threshold log): #4736 (open)
- Feature #1 (write guard): PR incoming (see below once opened)
- Features #3, #4: not started

Refs #4696, #4736.

chaitanyapantheor:
Feature #1 (write guard) PR is up: #4977 — adds `file-system.max-dir-entries` (reject `O_CREAT` with `ENOSPC` when the directory is already at/above the limit, default `0` = disabled).

Still open under this tracking issue: streaming/incremental readdir (#3) and documented directory-size limits (#4). Warning-threshold log (#2) is #4736.

PR #4977, opened 2026-08-03
feat(fs): add file-system.max-dir-entries write guard (ENOSPC above limit)

chaitanyapantheor:
### Description

Adds a `file-system.max-dir-entries` flag / config field. When it is greater than zero, creating a new file in a directory that already contains at least this many entries is rejected with `ENOSPC` and a `WARNING` is logged. `0` (the default) keeps the guard disabled, so behavior is unchanged unless a user opts in.

This implements the **write-guard** requested in #4696 — the corruption-prevention piece. Read-path mitigations (kernel list cache TTL, larger stat cache) reduce listing cost but do not stop a directory from growing further into the size range where listing degrades and metadata state can corrupt. Returning `ENOSPC` ("no space left on device") gives applications and scripts a standard, actionable signal at the source.

Implementation notes:
- The entry count is **bounded**: it uses `ReadDescendants(ctx, limit)`, which stops once it has seen the limit — mirroring the existing `file-system.rename-dir-limit` idiom. A directory comfortably below the limit only pays to list its own (few) entries; the cost is only incurred for directories at/near the limit, which are exactly the ones to protect.
- The guard is enforced at the `CreateFile` op (a single choke point covering both the empty-object and streaming/local create paths).
- Directories that don't support descendant listing (e.g. the multi-bucket mount root, which returns `ENOSYS`) are left to the normal create path.

Scope: this PR covers requested feature #1 (write guard) only. The warning-threshold log (#2) is #4736; streaming readdir (#3) and documented limits (#4) are tracked in #4976.

Implements part of #4696. Remaining-scope tracking issue: #4976.

### Testing details
1. Manual - Not run locally (FUSE mount is Linux-only; dev machine is macOS).
2. Unit tests - `go test ./cfg` passes with the new flag. Added `internal/fs/max_dir_entries_test.go` (fsTest-harness behavioral test: creates into a directory until a create is rejected, asserts the rejection is `ENOSPC` and occurs after no more than `limit` successes; plus a below-limit success case). Verified it compiles under `GOOS=linux go test -c ./internal/fs/`.
3. Integration tests - The `internal/fs` FUSE-backed suite runs in the project's Linux CI.

No. The guard is off by default (`max-dir-entries: 0`); existing behavior is unchanged unless a user sets a positive limit.

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -559,6 +559,8 @@ type FileSystemConfig struct {
 
 	MaxBackground int64 `yaml:"max-background"`
 
+	MaxDirEntries int64 `yaml:"max-dir-entries"`
+
 	MaxReadAheadKb int64 `yaml:"max-read-ahead-kb"`
 
 	RenameDirLimit int64 `yaml:"rename-dir-limit"`
@@ -1184,6 +1186,8 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.IntP("max-conns-per-host", "", 0, "The max number of TCP connections allowed per server. This is effective when client-protocol is set to 'http1'. A value of 0 indicates no limit on TCP connections (limited by the machine specifications).")
 
+	flagSet.IntP("max-dir-entries", "", 0, "If greater than zero, block creating new files in a directory that already contains at least this many entries, returning ENOSPC and logging a warning. 0 (the default) disables the guard.")
+
 	flagSet.IntP("max-idle-conns-per-host", "", 100, "The number of maximum idle connections allowed per server.")
 
 	flagSet.IntP("max-read-ahead-kb", "", 0, "Sets max kernel-read-ahead for the mount in KiB. 0 means system default. Requires sudo permission to set this value, otherwise the value will be ignored and system default will be used.")
@@ -1786,6 +1790,10 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 	}
 
 	if err := v.BindPFlag("gcs-connection.max-conns-per-host", flagSet.Lookup("max-conns-per-host")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("file-system.max-dir-entries", flagSet.Lookup("max-dir-entries")); err != nil {
 		return err
 	}
 
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -524,6 +524,15 @@ params:
         - bucket-type: "pirlo"
           value: "DefaultMaxBackground()"
 
+  - config-path: "file-system.max-dir-entries"
+    flag-name: "max-dir-entries"
+    type: "int"
+    usage: >-
+      If greater than zero, block creating new files in a directory that
+      already contains at least this many entries, returning ENOSPC and logging
+      a warning. 0 (the default) disables the guard.
+    default: "0"
+
   - config-path: "file-system.max-read-ahead-kb"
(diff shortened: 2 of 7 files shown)


chaitanyapantheor:
Thanks for the review — addressed all three findings in the latest commit:

**1. Recursive vs. flat counting + lock contention.** Replaced `ReadDescendants` with a new `DirInode.CountDirEntriesUpTo(ctx, limit)` that lists **direct entries only** (`Delimiter: "/"`, via `listObjectsAndBuildCores`) and stops once the count reaches the limit. It performs the GCS listing **without holding the parent inode's lock** (mirroring `readObjectsUnlocked`) and does not refresh caches, so a parent with few direct entries is no longer blocked by a full subdirectory, and file creation is no longer serialized behind the network call.

**2. Completeness.** The guard is now enforced in `MkDir`, `MkNode`, `CreateSymlink`, and `Rename` (destination parent, only for cross-directory moves — in-place renames don't change the destination's count), in addition to `CreateFile`.

**3. Test coverage.** Added:
- `TestDirectEntryCountIsNotRecursive` — a parent with several subdirectories each filled just under the limit (recursive count far above the limit) still accepts a direct create; this fails under recursive counting.
- `TestMkDirAndSymlinkBlockedWhenDirIsFull` — once a directory is full, `MkDir` and `CreateSymlink` are rejected with `ENOSPC`; the attempted entries are intentionally not cleaned up.

`go test ./cfg` passes; `internal/fs` builds and the test binary compiles under `GOOS=linux` (the FUSE suite runs in CI).

chaitanyapantheor:
Thanks for the review — all three points are addressed in the current implementation (`68167b2`), and I've just pushed `8c2a34f` to fix the failing Lint.

**1. Recursive counting + lock contention (both fixed).** The count no longer uses `ReadDescendants` (recursive, `Delimiter: ""`). It now uses a new `DirInode.CountDirEntriesUpTo`, which lists **flat** through `listObjectsAndBuildCores` (hardcoded `Delimiter: "/"`), so only **direct** entries are counted — a parent with a few subdirs that themselves hold many files is no longer blocked. It's **bounded** (stops as soon as the count reaches the limit) and performs the GCS list **without holding the inode lock** (mirroring `readObjectsUnlocked`), so concurrent operations on the directory are not serialized behind the network call.

**2. Completeness (fixed).** The guard is enforced at every entry-creating op, not just `CreateFile`: `MkDir`, `MkNode`, `CreateSymlink`, and cross-directory `Rename` (destination parent; in-place renames are skipped since they don't change the destination's count).

**3. Test coverage (added).** `TestDirectEntryCountIsNotRecursive` builds subdirectories whose recursive descendant count far exceeds the limit and asserts a direct create in the parent still succeeds (this would fail under recursive counting). `TestMkDirAndSymlinkBlockedWhenDirIsFull` asserts `MkDir` and `CreateSymlink` into a full directory are rejected with `ENOSPC`, and deliberately does **not** clean up those artifacts so a leftover entry is evidence of failure.

**CI.** Fixed the `Lint` failure (staticcheck `QF1008` — redundant embedded `fsTest` selectors in the test). Verified locally: `gofmt` clean, `go test ./cfg` passes, `GOOS=linux go vet ./internal/fs/ ./internal/fs/inode/` and `GOOS=linux go test -c ./internal/fs` both compile, and `golangci-lint` reports no issues on the changed file. (`zizmor-scan` is the Actions-workflow security scanner and is unrelated — this PR touches no workflow files.)

PR #4980, opened 2026-08-03
fix: Fixing only-dir mounting for managed folders

vadlakondaswetha:
We are doing getSTorageLayout for the bucket even incase of only-dir mounting. Instead also passing the directory name as part of getSTorageLayout call for only-dir mounting. This is important to avoid permission denied errors when user doesn't have access on bucket but has access to a specific folder inside the bucket (managed folder usecase)

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -397,7 +397,8 @@ func (sh *storageClient) lookupBucketType(bucketName string) (*gcs.BucketType, e
 	}
 
 	startTime := time.Now()
-	logger.Infof("GetStorageLayout <- (%s)", bucketName)
+	prefix := sh.clientConfig.OnlyDir
+	logger.Infof("GetStorageLayout <- (%s, prefix: %q)", bucketName, prefix)
 	storageLayout, err := sh.getStorageLayout(bucketName)
 	duration := time.Since(startTime)
 
@@ -405,7 +406,7 @@ func (sh *storageClient) lookupBucketType(bucketName string) (*gcs.BucketType, e
 		return nil, err
 	}
 
-	logger.Infof("GetStorageLayout -> (%s) %v msec", bucketName, duration.Milliseconds())
+	logger.Infof("GetStorageLayout -> (%s, prefix: %q) %v msec", bucketName, prefix, duration.Milliseconds())
 
 	// TODO (b/483608308): Once GetStorageLayout starts returning Pirlo bucket type,
 	// update this logic to use the response instead of inferring from clientConfig.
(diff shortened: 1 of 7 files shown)


internal/storage/storage_handle.go
```diff
@@ -388,15 +397,16 @@ func (sh *storageClient) lookupBucketType(bucketName string) (*gcs.BucketType, e
 	}
 
 	startTime := time.Now()
-	logger.Infof("GetStorageLayout <- (%s)", bucketName)
+	prefix := sh.clientConfig.OnlyDir
+	logger.Infof("GetStorageLayout <- (%s, prefix: %q)", bucketName, prefix)
```
abhishek10004, line 401: This logging and time measurement for GetStorageLayout should be inside the getStorageLayout function defined below and not here. Since, we're making changes in this function, let's clean this up as well.

internal/storage/storage_handle.go
```diff
@@ -388,15 +397,16 @@ func (sh *storageClient) lookupBucketType(bucketName string) (*gcs.BucketType, e
 	}
 
 	startTime := time.Now()
```
abhishek10004, line 399: Also, let's start time after the log statement is printed.

internal/storage/storage_handle.go
```diff
@@ -418,18 +428,28 @@ func (sh *storageClient) lookupBucketType(bucketName string) (*gcs.BucketType, e
...
 
-	return stoargeLayout, err
+	return storageLayout, err
 }
 
 // NewStorageHandle creates control client and stores client config to allow dynamic
 // creation of http or grpc client.
 func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClientConfig, billingProject string) (sh StorageHandle, err error) {
+	// Sanitize the onlyDir path for only-dir mounting
+	if clientConfig.OnlyDir != "" {
```
abhishek10004, line 445: What about the case when OnlyDir has .. at the start? Is that getting handled?

internal/storage/storage_handle.go
vadlakondaswetha: I dont see any diff printing it here or there. Not sure why we even created a seperate method just for the call.

internal/storage/storage_handle.go
vadlakondaswetha: Whats the reason?

internal/storage/storage_handle.go
abhishek10004: Agreed that a separate method was not needed but since it's here, it would be better if logging etc. is also handled in that function call only.

internal/storage/storage_handle.go
abhishek10004: I don't like the logging time getting included in the actual measurement. Hence, I prefer to log outside of the measurement blocks.

abhishek10004 (approved):
LGTM

internal/storage/storage_handle.go
vadlakondaswetha: lets keep that method as a simple wrapper. lets not add some logic here and other logic there.

internal/storage/storage_handle.go
vadlakondaswetha: I prefer not changing the existing code in this PR as there is no strong reason why start time should be after logging.

PR #4981, opened 2026-08-03
perf(storage): Optimize RenameFolder LRO polling strategy to improve client-side latency

meet2mky:
## Summary
This PR improves client-side observed latency for long-running operations (such as folder rename `RenameFolder` on HNS buckets) by replacing `resp.Wait(ctx)` with an optimized time-based linear polling strategy (`PollLRO`).

## Motivation & Problem Statement
Currently, operations like `bucketHandle.RenameFolder` call `resp.Wait(ctx)` from the Go Cloud Storage Control API client library to await Long Running Operation (LRO) completion. 
Under the hood, `resp.Wait(ctx)` relies on `cloud.google.com/go/longrunning`, which enforces a hardcoded `gax.Backoff` starting at **1.0 second** with **full random jitter** (`rand[1ns, 1s]`) and doubling up to **60 seconds**.

### Pitfalls of Current `resp.Wait(ctx)`:
1. **High Initial Delay**: Small/empty folder renames complete on GCS in 10ms–50ms. If Poll #0 misses completion by 1ms, the client immediately sleeps up to **1,000 ms** (avg 500 ms) before retrying.
2. **Aggressive Escalation & Large Gaps**: Sleep intervals double rapidly (`1s -> 2s -> 4s -> 8s -> 16s -> 32s -> 60s`). A 20-second server rename currently incurs **11.5 seconds of idle client sleep** before the next status poll occurs at 31.5s.

---

## Solution
We implement `PollLRO` in `internal/storage/storageutil/lro_poller.go` using a **Time-Based Linear Delay Schedule**:

The polling sleep interval dynamically grows linearly relative to the elapsed wall-clock time of the operation:
`delay(t) = min(Max, (Max / CapTime) * t)`

1. **Low Initial Overhead**: The delay starts smoothly from a minimum floor of **50ms**, ensuring that fast operations (completing under 1s) are detected almost instantly.
2. **Proportional Detection Lag**: The worst-case detection lag (client-side latency overhead) is mathematically bounded to exactly `Max / CapTime` of the elapsed time. With `Max = 30s` and `CapTime = 10m`, this means the overhead is strictly **5% of the operation time**.
3. **Graceful Ceiling**: The delay caps at a maximum of **30 seconds** (at `t = 10 minutes`), protecting GCS Control API quota and reducing background gRPC traffic for multi-minute operations.
4. **Jitter**: A multiplicative jitter of **±10%** (`0.9x to 1.1x`) is applied to prevent synchronized retries across concurrent operations from causing API thundering herds.

---

## Latency Overhead Gain Comparison
By bounding the polling gap to a 5% slope rather than a rapidly doubling exponential curve, we save significant overhead on medium-to-large operations.

| Server Operation Time ($T_{server}$) | Current Code Overhead (`resp.Wait`) | Proposed Code Overhead (Linear 5% + 50ms Floor) | Latency Overhead Saved |
|---|---|---|---|
| **20 ms** (Ultra-Small) | ~500 ms | **~50 ms** | **~450 ms saved** |
| **500 ms** (Medium-Small) | ~750 ms | **~50 ms** | **~700 ms saved** |
| **20.0 s** (Medium-Large) | **11.50 s** (11,500 ms) | **~1.00 s** (1,000 ms) | **10.50 seconds saved (91% less overhead)** |
| **30.0 s** (Large) | **1.50 s** (1,500 ms) | **~1.50 s** (1,500 ms) | Performs equally well without aggressive 30s+ gaps |
| **2.0 mins** (Very Large) | **28.00 s** (28,000 ms) | **~6.00 s** (6,000 ms) | **22.0 seconds saved (78% less overhead)** |

---

## Key Changes
- **`internal/storage/storageutil/lro_poller.go`**:
  - Implemented `PollLRO(ctx, op, cfg)` with the linear delay schedule.
  - Added default configuration constants (`DefaultLROPollMin = 50ms`, `DefaultLROPollMax = 30s`, `DefaultLROPollCapTime = 10m`).
- **`internal/storage/bucket_handle.go`**:
  - Replaced `resp.Wait(ctx)` with `storageutil.PollLRO(ctx, resp, storageutil.DefaultLROPollConfig())`.
  - Added `TODO(b/540568787)` tracking Go-SDK native polling configuration support.
- **`internal/storage/storageutil/lro_poller_test.go`**:
  - Updated AAA unit tests covering immediate success/error, multi-retry polling, transient RPC errors (429s, Auth), context cancellation, and bounds validation (`Min > Max`, `CapTime <= 0`).

---

## Verification & Testing
Ran unit tests across `internal/storage/`:
```bash
go test -v ./internal/storage/storageutil/...
go test -v ./internal/storage/...

### CODE REVIEW

File: internal/storage/storageutil/rename_folder_poller.go [hunk 1]
```diff
@@ -0,0 +1,128 @@
...
+	"context"
+	"time"
+
+	"cloud.google.com/go/storage/control/apiv2/controlpb"
+	"github.com/googleapis/gax-go/v2"
+)
+
+const (
+	// DefaultRenamePollInitial is the fixed sleep duration between retry polls during the initial fast phase.
+	DefaultRenamePollInitial = 500 * time.Millisecond
```
vadlakondaswetha: do we have numbers on how long does Rename takes on a empty directory, directory with < 10 files, 100 files etc. i am trying to understand if 500ms delay is acceptable or should we start with even smaller value and increase upto 500ms even in the Fast phase.

internal/storage/storageutil/rename_folder_poller.go
```diff
@@ -0,0 +1,128 @@
...
+	// Poll #0: Immediate status check right after operation creation to handle instantaneous completions.
+	folder, err := op.Poll(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if op.Done() {
+		return folder, nil
+	}
+
+	initial := cfg.Initial
```
vadlakondaswetha: I dont understand this code. you are passing the default values to this method and verifying again if the values are valid or not.

can you just validate the config and throw an error if not valid.

internal/storage/storageutil/rename_folder_poller.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// DefaultRenamePollMultiplier is the backoff factor applied to sleep intervals after the fast phase window expires.
+	DefaultRenamePollMultiplier = 1.1
+
+	// DefaultRenamePollMax is the maximum backoff sleep duration cap for long-running folder rename operations.
+	DefaultRenamePollMax = 30 * time.Second
+)
+
+// RenamePollConfig holds configuration parameters for polling long-running operations.
+type RenamePollConfig struct {
```
vadlakondaswetha: can we make this generic and call as LRO polling config.

meet2mky: done!

internal/storage/storageutil/rename_folder_poller.go
meet2mky: Done, removed the fallback throwing error on invalid configuration.

internal/storage/storageutil/rename_folder_poller.go
meet2mky: for empty directories LRO completes in around ~50ms and for folders which have 100 subfolders in them it completes in around ~300ms. We also do a first poll immediately after the LRO creation so if the operation complete fast enough they should get the result immediately. Changing frequency to higher than this can result in more QPS.

image

meet2mky: The initial 500ms is also having full jitter so the actual sleep is somewhere between 1ns to 500ms (Avg is 250ms).

meet2mky:
gemini review

internal/storage/storageutil/lro_poller.go
```diff
@@ -0,0 +1,128 @@
...
+
+		// Apply exponential backoff after fast phase window.
+		if time.Since(startTime) >= cfg.FastPhaseWindow {
+			backoff = min(cfg.Max, time.Duration(float64(backoff)*cfg.Multiplier))
+		}
+
+		// Add full jitter to avoid the thundering herd problem.
+		// The actual wait time is a random value between 1ns and the current backoff.
+		pause := time.Duration(rand.Int63n(int64(backoff))) + 1
+		timer.Reset(pause)
```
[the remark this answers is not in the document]
meet2mky: Fixed both issues, Jitter is improved to be 80 to 100% of 500ms and explicitly added retry pattern similar to other API(s) in GRPC Client using ShouldRetryWithoutLogging to handle Auth Error/ 429 and other socket level errors etc.

internal/storage/storageutil/lro_poller.go
```diff
@@ -0,0 +1,132 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return zero, ctx.Err()
+		case <-timer.C:
+		}
+
+		result, err = op.Poll(ctx)
+		if err != nil {
+			if !ShouldRetryWithoutLogging(err) {
```
abhishek10004: There can be an issue in this case. If a completed LRO fails with a transient operation, then we can run into an infinite loop.
Once op.Done is true, we don't make a network call again and keep on returning the cached value.

So, we should update check for op.Done as well

!ShouldRetryWithoutLogging(err) --> !ShouldRetryWithoutLogging(err) || op.Done()

abhishek10004: Same handling should be done above as well.

internal/storage/storageutil/lro_poller.go
```diff
@@ -0,0 +1,132 @@
...
+type LROPoller[T any] interface {
+	Poll(ctx context.Context, opts ...gax.CallOption) (T, error)
+	Done() bool
+}
+
+// PollLRO polls the given LRO using a two-stage adaptive polling strategy.
+func PollLRO[T any](ctx context.Context, op LROPoller[T], cfg LROPollConfig) (T, error) {
+	var zero T
+
+	if cfg.Initial <= 0 {
```
abhishek10004: We should add another validation for initial to be less than max as well

internal/storage/storageutil/lro_poller.go
meet2mky: Thanks 👍🏻 , I have added at both places.

internal/storage/storageutil/lro_poller.go
meet2mky: Done 💯

internal/storage/storageutil/rename_folder_poller.go
abhishek10004: I was reading about this and the time depends on how many batches the sub folders would fit into. It typically forms a batch of ~50 subfolders and then rename it in one transaction (So, a rename where the number of subfolders fit into one batch would be complete in one transaction and rename would be done when the API returns)


So, we can take the initial time to be slightly more than what a single shot rename would take and then maybe slowly ramp it upto 500 ms, if needed, or just use that smaller value till the time we start using the multiplier. WDYT @meet2mky?

cc @vadlakondaswetha

meet2mky: LRO latency is not just about the single transaction cost, it has some cost of scheduling delay at server side. From the client side we poll immediately after the LRO is successfully created. So if it got completed within the time when the LRO was created using first request and by the time we fetched it's status in the subsequent request then it should be fine. If it's not we poll in next 500ms later. Your approach is also okay but it would complicate the logic more IMO. There doesn't seem to be many workload (apart from checkpointing) which depends on LRO latency since it's already a long running operation.

meet2mky: I have changed the strategy to be more dynamic. With this approach we have capped the detection delay on the client side to be maximum of 5% only in worst case scenarios. I.e, assuming a server LRO complete in X seconds then client will in worst case scenario know by 1.05*X seconds.

internal/storage/storageutil/lro_poller.go
```diff
@@ -0,0 +1,129 @@
...
+
+const (
+	// DefaultLROPollMin is the minimum backoff sleep duration for long-running operations.
+	DefaultLROPollMin = 50 * time.Millisecond
+
+	// DefaultLROPollMax is the maximum backoff sleep duration cap for long-running operations.
+	DefaultLROPollMax = 30 * time.Second
+
+	// DefaultLROPollCapTime is the time at which the delay reaches DefaultLROPollMax.
+	DefaultLROPollCapTime = 10 * time.Minute
```
abhishek10004, line 34: maybe set it to 5 mins?

abhishek10004 (dismissed):
LGTM

internal/storage/storageutil/lro_poller.go
meet2mky: I think it's fine as 30s of 10 min is 5% but If I set it to 5 min then it should 10% detection lag in client side.

internal/storage/storageutil/lro_poller.go
```diff
@@ -0,0 +1,129 @@
...
+		if err := ctx.Err(); err != nil {
+			return zero, err
+		}
+
+		elapsed := time.Since(startTime)
+		ms := min(float64(cfg.Max.Milliseconds()), max(float64(cfg.Min.Milliseconds()), slope*float64(elapsed.Milliseconds())))
+		// Apply ±10% multiplicative jitter (0.9x to 1.1x of the computed delay)
+		// to prevent synchronized retries from multiple operations hitting the API at the same time.
+		jitterFactor := 0.9 + 0.2*rand.Float64()
+		pause := time.Duration(ms*jitterFactor) * time.Millisecond
```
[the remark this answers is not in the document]
meet2mky: Good catch. Done!

abhishek10004 (approved):
LGTM

PR #4984, opened 2026-08-03
test(implicit_dir): bifurcate implicit_dir tests for rapid writes flagsets

vipnydav:
### Description
Bifurcates implicit_dir tests into TestImplicitDirBase and TestImplicitDirLocalFileRapidWritesEnabled while preserving both static and persistent mounting.

### Testing details
1. Unit tests - NA
2. Integration tests - Done

No

### DIFF
--- a/tools/integration_tests/implicit_dir/implicit_dir_test.go
+++ b/tools/integration_tests/implicit_dir/implicit_dir_test.go
@@ -20,13 +20,16 @@ import (
 	"log"
 	"os"
 	"path"
+	"strings"
 	"testing"
 
 	"cloud.google.com/go/storage"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/client"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting/persistent_mounting"
+	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/mounting/static_mounting"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup"
-	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/setup/implicit_and_explicit_dir_setup"
 	"github.com/googlecloudplatform/gcsfuse/v3/tools/integration_tests/util/test_suite"
+	"github.com/stretchr/testify/suite"
 )
 
 const ExplicitDirInImplicitDir = "explicitDirInImplicitDir"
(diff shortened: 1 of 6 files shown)


tools/integration_tests/implicit_dir/local_file_test.go
```diff
@@ -22,66 +22,86 @@ import (
...
+type implicitDirLocalFileTest struct {
+	isRapidWritesEnabled bool
+	suite.Suite
+}
+
+func (i *implicitDirLocalFileTest) TearDownTest() {
+	setup.SaveGCSFuseLogFileInCaseOfFailure(i.T())
+}
+
+func TestImplicitDirLocalFileRapidWritesEnabled(t *testing.T) {
```
anushka567: There is a name mismatch between test_config.yaml (run: `TestImplicitDirRapidWritesEnabled`) and local_file_test.go (`TestImplicitDirLocalFileRapidWritesEnabled`). Because the names don't match, setup.BuildFlagSets returns an empty array []. If you run this test on a Pirlo bucket, it returns PASS (0.00s) without generating any subtests or running any assertions.

tools/integration_tests/implicit_dir/implicit_dir_test.go
```diff
@@ -80,11 +125,12 @@ func TestMain(m *testing.M) {
 		}
 	}()
 
-	// 3. Build the flag sets dynamically from the config.
-	flags := setup.BuildFlagSets(cfg.ImplicitDir[0], bucketType, "")
+	// 3. Set up test directory for test bucket.
+	setup.SetUpTestDirForTestBucket(testEnv.cfg)
```
anushka567, line 134: In TestMain, `setup.SetUpTestDirForTestBucket(testEnv.cfg)` is called unconditionally. It needs to be skipped if GKEMountedDirectory is present.

tools/integration_tests/test_config.yaml
```diff
@@ -36,20 +36,31 @@ implicit_dir:
...
+          - "--experimental-enable-pirlo,--enable-rapid-writes=true"
         run_on_pirlo:
           hns:
             same_zone: true
             different_zone: false
-        run_on_gke: true
+        run_on_gke: false
+        run: TestImplicitDirRapidWritesEnabled
+      - flags:
+          - "--experimental-enable-pirlo,--enable-rapid-writes=false"
```
anushka567: Is dropping the `--implicit-dirs` flag for pirlo tests intentional ? considering the test package is dedicated to this functionality?

tools/integration_tests/implicit_dir/local_file_test.go
vipnydav: Thanks for catching that.

tools/integration_tests/implicit_dir/implicit_dir_test.go
vipnydav: Done

tools/integration_tests/test_config.yaml
vipnydav: It was intentional as for HNS Buckets we check for implicit directories even if the `--implicit-dirs` flag is not passed but I'd add it for consistency.

anushka567 (approved):
LGTM.

abhishek10004 (approved):
Approving based on Anushka's review.

PR #4998, opened 2026-08-09
perf(cache): optimize stat_cache memory-usage

kislaykishore:
### Description
* Reduce the memory footprint of entries in the metadata-cache
* Add stat-cache benchmarks

b/538391384

### Memory Workload Benchmark (1 Million Entries)

| Workload (1M Entries) | Baseline (`master- 0673d1998`) | Optimized Branch | Absolute Memory Saved | Heap Memory Reduction |
|:---|:---:|:---:|:---:|:---:|
| **Flat** | `282.34 MB` | `267.06 MB` | **`-15.28 MB`** | **`-5.41%`** |
| **Nested** | `282.06 MB` | `266.89 MB` | **`-15.17 MB`** | **`-5.38%`** |
| **Deeply Nested** | `282.15 MB` | `266.80 MB` | **`-15.35 MB`** | **`-5.44%`** |

---

### CPU & Allocation Microbenchmarks (`go test -bench -benchmem`)

| Benchmark Operation | Baseline Latency | Optimized Latency | $\Delta$ Latency | Baseline Allocations | Optimized Allocations | $\Delta$ Memory Allocated |
|:---|:---:|:---:|:---:|:---:|:---:|:---:|
| **`LookUp`** | `65.08 ns/op` | `61.50 ns/op` | **`-5.50%`** *(faster)* | `18 B/op (0 allocs)` | `12 B/op (0 allocs)` | **`-33.3%`** *(-6 B/op)* |
| **`Insert`** | `334.00 ns/op` | `334.60 ns/op` | `+0.18%` *(parity)* | `160 B/op (3 allocs)` | `144 B/op (3 allocs)` | **`-10.0%`** *(-16 B/op)* |
| **`AddNegativeEntry`** | `142.10 ns/op` | `141.30 ns/op` | `-0.56%` *(parity)* | `96 B/op (2 allocs)` | `80 B/op (2 allocs)` | **`-16.7%`** *(-16 B/op)* |

---

### Struct Layout & Per-Entry Impact

| Attribute | Baseline (`master`) | Optimized Branch | Difference |
|:---|:---:|:---:|:---:|
| **`entry` Struct Size (`unsafe.Sizeof`)** | `48 bytes` | `32 bytes` | **`-16 bytes` (`-33.3%`)** |
| **`expiration` Field Type** | `time.Time` (24 bytes) | `int64` (8 bytes) | **`-16 bytes`** |
| **Calculated RSS per Entry (`entry.Size()`)** | Base $\approx 1304\text{ B}$ | Base $\approx 1272\text{ B}$ | **`-32 bytes` (with factor $2.0$)** |
| **Direct Heap Memory per 1M Entries** | $\approx 282\text{ MB}$ | $\approx 267\text{ MB}$ | **$\approx -15.3\text{ MB}$** |

### Testing details
1. Manual - Ran `make build` with zero compilation errors and zero lint warnings.
2. Unit tests - Ran `go test -v ./internal/cache/metadata/...` (all unit and memory workload tests passing).
3. Integration tests - N/A

N/A

### DIFF
--- a/internal/cache/metadata/stat_cache.go
+++ b/internal/cache/metadata/stat_cache.go
@@ -24,6 +24,9 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
 )
 
+// 9223372036 is math.MaxInt64 / 1,000,000,000 (max seconds representable in nanoseconds)
+const maxSeconds int64 = math.MaxInt64 / 1000000000
+
 // A cache mapping from name to most recent known record for the object of that
 // name. External synchronization must be provided.
 type StatCache interface {
@@ -113,7 +116,7 @@ type statCacheBucketView struct {
 type entry struct {
 	m          *gcs.MinObject
 	f          *gcs.Folder
-	expiration time.Time
+	expiration int64
 	// Set to true only for implicit directory entries. This flag will always remain false for negative entries and explicit objects.
 	implicitDir bool
 }
@@ -143,6 +146,25 @@ func (e entry) Size() uint64 {
 	size = uint64(math.Ceil(util.HeapSizeToRssConversionFactor * float64(size)))
 
 	return size
+}
+
+// timeToUnixNano safely converts a time.Time to Unix nanoseconds.
+// It prevents int64 overflow for dates beyond the year 2262, returning
+// math.MaxInt64 instead, which safely mimics "never expire".
+func timeToUnixNano(t time.Time) int64 {
+	if t.IsZero() {
+		return math.MinInt64
+	}
+
+	unixSecs := t.Unix()
+	if unixSecs >= maxSeconds {
+		return math.MaxInt64
+	}
+	if unixSecs <= -maxSeconds {
+		return math.MinInt64
+	}
+
+	return t.UnixNano()
 }
 
 // Should the supplied object for a new positive entry replace the given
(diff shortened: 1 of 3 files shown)


internal/cache/metadata/stat_cache.go
```diff
@@ -24,6 +24,9 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/util"
 )
 
+// 9223372036 is math.MaxInt64 / 1,000,000,000 (max seconds representable in nanoseconds)
+const maxSeconds int64 = math.MaxInt64 / 1000000000
```
abhishek10004: nit: can use int64(time.Second) or 1e9 for better readability

internal/cache/metadata/stat_cache_test.go
abhishek10004: These kind of comments are wrong now since we've updated the struct. We should update these as well otherwise these causes confusion.

abhishek10004 (commented):
Changes look good. Just one suggestion to update the comments describing the size as well since that has now changed.
In future, we can look to further improve the tests to reduce effort whenever the struct gets modified.

internal/cache/metadata/stat_cache_test.go
```diff
@@ -168,18 +168,18 @@ func (t *StatCacheTest) Test_KeysPresentButEverythingIsExpired() {
 }
 
 func (t *StatCacheTest) Test_FillUpToCapacity() {
-	assert.Equal(t.T(), 3, capacity) // maxSize = 3 * 1640 = 4920 bytes
+	assert.Equal(t.T(), 3, capacity) // maxSize = 3 * 1720 = 5160 bytes
```
abhishek10004, line 171: just curious. Since we've reduced the size, how did this increase?

abhishek10004 (approved):

internal/cache/metadata/stat_cache_test.go
kislaykishore: I think the comment was out-of-date.

PR #5003, opened 2026-08-10
fix(cfg): enable finalize-on-close for aiml-checkpointing profile

vipnydav:
### Description
This PR enables `finalize-on-close` by default for the `aiml-checkpointing` profile to prevent checkpoint loading failures caused by stale metadata.

**Background:**
As discovered in b/540250863 during checkpoint benchmarking, when a directory rename occurs, GCSFuse invalidates cached entries under the old folder prefix. A subsequent `ListObjects` call on the new directory re-populates the cache. 

However, because zonal buckets use the appendable API where objects are not finalized on write by default, `ListObjects` can return a stale, intermediate object size for these unfinalized objects. Since the `aiml-checkpointing` profile sets an infinite metadata cache TTL (`metadata-cache.ttl-secs = -1`), this stale size is cached permanently in RAM, which ultimately causes `OUT_OF_RANGE` errors during checkpoint loading. 

Enabling finalize on close explicitly finalizes the object, avoiding the asynchronous metadata propagation delay for unfinalized appendable objects.

b/544523257

### Testing details
1. Manual - NA
2. Unit tests - Added test cases for `aiml-checkpointing` profile in `cfg/config_test.go`
3. Integration tests - NA

NA

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -78,6 +78,12 @@ var AllFlagOptimizationRules = map[string]shared.OptimizationRules{"file-system.
 			Value: bool(true),
 		},
 	},
+	Profiles: []shared.ProfileOptimization{
+		{
+			Name:  "aiml-checkpointing",
+			Value: bool(true),
+		},
+	},
 }, "file-system.fuse-max-request-size-kb": {
 	BucketTypeOptimization: []shared.BucketTypeOptimization{
 		{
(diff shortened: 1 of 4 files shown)


charith87:
Adding @abhishek10004 and @vadlakondaswetha since its ZB related.

PR #5004, opened 2026-08-10
fix(storage): honor --enable-rapid-writes flag during object overwrites for pirlo buckets

vipnydav:
### Description
When overwriting an existing object, GCSFuse currently copies the original object's storage class into the new request. If an object with the `RAPID` storage class in a pirlo bucket is overwritten while the GCSFuse mount flag `--enable-rapid-writes` is set to `false`, the request fails with an `InvalidArgument` error. This occurs because the combination of using the Resumable upload API (which is used when rapid writes are disabled) and explicitly passing the `RAPID` storage class is rejected by GCS.

This PR fixes the issue by explicitly clearing the storage class (`req.StorageClass = ""`) for pirlo buckets when rapid writes are disabled. This ensures GCSFuse strictly honors the mount flag rather than inheriting the original object's `RAPID` storage class, allowing the overwritten pirlo bucket object to correctly fall back to the bucket's default storage class instead.

b/543390252

### Testing details
1. Manual - NA
2. Unit tests - NA

No.

### DIFF
--- a/internal/storage/bucket_handle.go
+++ b/internal/storage/bucket_handle.go
@@ -200,9 +200,14 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 
 	// When rapid writes are enabled on an RCU bucket, objects should be explicitly
 	// created with the RAPID storage class in the zonal cache rather than
-	// defaulting to the bucket's default storage class
+	// defaulting to the bucket's default storage class.
+	// If rapid writes are disabled, we explicitly clear the storage class to prevent
+	// inheriting the RAPID storage class from an overwritten object and write it to
+	// the bucket's default storage class.
 	if bh.BucketType().Pirlo == gcs.PirloStateRapidWritesEnabled {
 		req.StorageClass = storageClassRapid
+	} else if bh.BucketType().Pirlo == gcs.PirloStateRapidWritesDisabled {
+		req.StorageClass = ""
 	}
 
 	obj := bh.getObjectHandleWithPreconditionsSet(req)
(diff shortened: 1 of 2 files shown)


internal/storage/bucket_handle.go
```diff
@@ -200,9 +200,15 @@ func (bh *bucketHandle) CreateObject(ctx context.Context, req *gcs.CreateObjectR
 
 	// When rapid writes are enabled on an RCU bucket, objects should be explicitly
 	// created with the RAPID storage class in the zonal cache rather than
-	// defaulting to the bucket's default storage class
-	if bh.BucketType().Pirlo == gcs.PirloStateRapidWritesEnabled {
+	// defaulting to the bucket's default storage class.
+	// If rapid writes are disabled, we explicitly clear the storage class to prevent
+	// inheriting the RAPID storage class from an overwritten object and write it to
+	// the bucket's default storage class.
+	switch bh.BucketType().Pirlo {
```
vadlakondaswetha, line 203: you can extract this logic into a common method and use at both the places

vipnydav: Done

vadlakondaswetha: we should avoid overriding variables in a different function as it is difficult to main in long run, which method is modifiying what and the order.

i would suggest adding a method getSTorageClassForCreateObject and use this in the CreateObject method
req.storageClass = getStorageClassForCreateObject(req)

vipnydav: Agreed. Updated the helper method to return the storage class and assign it to `req.StorageClass` directly in the caller.

vadlakondaswetha (approved):

PR #5005, opened 2026-08-10
feat(native-retries): retry token acquisition and refresh via retrying token provider

dangarhs:
### Description
- Added `retryingTokenProvider` (`auth.TokenProvider`) and `retryingTokenSource` (`oauth2.TokenSource`) in `internal/storage/storageutil/token_retry.go` to retry transient Metadata Server (MDS) and OAuth2 errors (`isTransientMDSError`) using `ExecuteWithCustomShouldRetryAtLogLevel`.
- Updated `GetClientAuthOptionsAndToken` in `internal/storage/storageutil/auth_client_option.go` to wrap credentials and token sources with the retry decorators when `EnableMountRetries` is enabled.
- Retries both initial token source creation (`auth2.NewTokenSourceFromURL` and `auth2.GetCredentials`) during mount as well as subsequent `Token()` / `ProvideToken()` calls during runtime token refresh.
- Added comprehensive unit tests in `internal/storage/storageutil/token_retry_test.go`.

b/538080458

### Testing details
1. Manual - NA
2. Unit tests - Added unit tests in `token_retry_test.go` verifying first-attempt success, retry on transient errors, fast failure on non-transient errors, and exhausted retry attempts.
3. Integration tests - NA

No

### DIFF
--- a/internal/storage/storageutil/auth_client_option.go
+++ b/internal/storage/storageutil/auth_client_option.go
@@ -30,11 +30,40 @@ import (
 
 // GetClientAuthOptionsAndToken returns client options and a token source using either a token URL or fallback to key file/ADC.
 func GetClientAuthOptionsAndToken(ctx context.Context, config *StorageClientConfig) ([]option.ClientOption, oauth2.TokenSource, error) {
+	retryConfig := NewRetryConfig(config)
+
 	// If Token URL is provided, attempt to fetch token source directly.
 	if config.TokenUrl != "" {
-		tokenSrc, err := auth2.NewTokenSourceFromURL(ctx, config.TokenUrl, config.ReuseTokenFromUrl)
+		var tokenSrc oauth2.TokenSource
+		var err error
+
+		if config.EnableMountRetries {
+			apiCall := func(attemptCtx context.Context) (oauth2.TokenSource, error) {
+				return auth2.NewTokenSourceFromURL(attemptCtx, config.TokenUrl, config.ReuseTokenFromUrl)
+			}
+			tokenSrc, err = ExecuteWithCustomShouldRetryAtLogLevel(
+				ctx,
+				retryConfig,
+				"NewTokenSourceFromURL",
+				"token-source",
+				uuid.NewString(),
+				apiCall,
+				ShouldRetryOnMount,
+				logger.LevelInfo,
+			)
+		} else {
+			tokenSrc, err = auth2.NewTokenSourceFromURL(ctx, config.TokenUrl, config.ReuseTokenFromUrl)
+		}
+
 		if err != nil {
 			return nil, nil, fmt.Errorf("while fetching token source: %w", err)
+		}
+
+		if config.EnableMountRetries {
+			tokenSrc = &retryingTokenSource{
+				base:        tokenSrc,
+				retryConfig: retryConfig,
+			}
 		}
 
 		clientOpts := []option.ClientOption{option.WithTokenSource(tokenSrc)}
@@ -42,12 +71,40 @@ func GetClientAuthOptionsAndToken(ctx context.Context, config *StorageClientConf
 	}
 
 	// Fallback: Use key file credentials.
-	cred, err := auth2.GetCredentials(config.KeyFile)
+	var cred *auth.Credentials
+	var err error
+
+	if config.EnableMountRetries {
+		apiCall := func(_ context.Context) (*auth.Credentials, error) {
+			return auth2.GetCredentials(config.KeyFile)
+		}
+		cred, err = ExecuteWithCustomShouldRetryAtLogLevel(
+			ctx,
+			retryConfig,
+			"GetCredentials",
+			"credentials",
+			uuid.NewString(),
+			apiCall,
+			ShouldRetryOnMount,
+			logger.LevelInfo,
+		)
+	} else {
+		cred, err = auth2.GetCredentials(config.KeyFile)
+	}
+
 	if err != nil {
 		return nil, nil, fmt.Errorf("while fetching credentials: %w", err)
 	}
 
-	tokenSrc := oauth2adapt.TokenSourceFromTokenProvider(cred.TokenProvider)
+	tokenProvider := cred.TokenProvider
+	if config.EnableMountRetries {
+		tokenProvider = &retryingTokenProvider{
+			base:        cred.TokenProvider,
+			retryConfig: retryConfig,
+		}
+	}
+
+	tokenSrc := oauth2adapt.TokenSourceFromTokenProvider(tokenProvider)
 
 	var domain string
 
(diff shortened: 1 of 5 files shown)


internal/storage/storageutil/auth_client_option.go
```diff
@@ -30,24 +30,82 @@ import (
 
 // GetClientAuthOptionsAndToken returns client options and a token source using either a token URL or fallback to key file/ADC.
 func GetClientAuthOptionsAndToken(ctx context.Context, config *StorageClientConfig) ([]option.ClientOption, oauth2.TokenSource, error) {
```
meet2mky:
```suggestion
func GetClientAuthOptionsAndTokenSource(ctx context.Context, config *StorageClientConfig) ([]option.ClientOption, oauth2.TokenSource, error) {
```
Nit: Should we rename this method for more clarity on it's use?

internal/storage/storageutil/auth_client_option.go
```diff
@@ -30,24 +30,82 @@ import (
...
 	if config.TokenUrl != "" {
-		tokenSrc, err := auth2.NewTokenSourceFromURL(ctx, config.TokenUrl, config.ReuseTokenFromUrl)
+		var tokenSrc oauth2.TokenSource
+		var err error
+
+		if config.EnableMountRetries {
+			apiCall := func(attemptCtx context.Context) (oauth2.TokenSource, error) {
+				return auth2.NewTokenSourceFromURL(attemptCtx, config.TokenUrl, config.ReuseTokenFromUrl)
+			}
+			tokenSrc, err = ExecuteWithCustomShouldRetryAtLogLevel(
```
meet2mky:
```suggestion
			tokenSrc, err = ExecuteWithCustomShouldRertry(
```
Use this method otherwise every call will become verbose.

internal/storage/storageutil/auth_client_option.go
```diff
@@ -30,24 +30,82 @@ import (
...
+			"GetCredentials",
+			"credentials",
+			uuid.NewString(),
+			apiCall,
+			ShouldRetryOnMount,
+			logger.LevelInfo,
+		)
+	} else {
+		cred, err = auth2.GetCredentials(config.KeyFile)
+	}
```
meet2mky: why does this require retrying? if it's key file then it shouldn't make any MDS interaction or external API(s).

internal/storage/storageutil/auth_client_option.go
```diff
@@ -58,14 +116,16 @@ func GetClientAuthOptionsAndToken(ctx context.Context, config *StorageClientConf
...
-		retryConfig := NewRetryConfig(config)
-
 		apiCall := func(attemptCtx context.Context) (string, error) {
 			d, err := cred.UniverseDomain(attemptCtx)
 			return d, err
 		}
 
-		domain, err = ExecuteWithRetryAtLogLevel(ctx, retryConfig, "cred.UniverseDomain", "credentials", uuid.NewString(), apiCall, logger.LevelInfo)
+		if config.EnableMountRetries {
+			domain, err = ExecuteWithCustomShouldRetryAtLogLevel(ctx, retryConfig, "cred.UniverseDomain", "credentials", uuid.NewString(), apiCall, ShouldRetryOnMount, logger.LevelInfo)
```
meet2mky, line 83: Same here let's use the one which is less verbose.

meet2mky (changes requested):
Added some comments, PTAL.

internal/storage/storageutil/auth_client_option.go
dangarhs: It also acts as fallback for ADC which checks the MDS, so I added it to handle transient MDS errors. But, I just checked that the SDK memoizes that MDS failure via `sync.Once`. So retry is not needed.

internal/storage/storageutil/auth_client_option.go
dangarhs: Done.

internal/storage/storageutil/auth_client_option.go
dangarhs: Ack.

internal/storage/storageutil/auth_client_option.go
dangarhs: Ack.

internal/storage/storageutil/token_retry.go
```diff
@@ -0,0 +1,73 @@
...
+		ShouldRetryOnMount,
+	)
+}
+
+// retryingTokenSource wraps an oauth2.TokenSource to retry transient MDS/OAuth errors.
+type retryingTokenSource struct {
+	ctx         context.Context
+	base        oauth2.TokenSource
+	retryConfig *RetryConfig
+}
```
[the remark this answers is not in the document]
meet2mky: +1, Ref: go/how-to-use-a-context#rule-4

meet2mky:
PTAL at gemini comments.

dangarhs:
> PTAL at gemini comments.

Done.

internal/storage/storageutil/token_retry.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (r *retryingTokenSource) Token() (*oauth2.Token, error) {
+	apiCall := func(_ context.Context) (*oauth2.Token, error) {
+		return r.base.Token()
+	}
+
+	return ExecuteWithCustomShouldRetry(
+		context.Background(),
+		r.retryConfig,
+		"Token",
```
meet2mky: Nit: Let's change "Token" -> "TokenSource.Token" to make it consistent with "TokenProvider.Token"

meet2mky (dismissed):
Looks good to me.

internal/storage/storageutil/auth_client_option.go
```diff
@@ -58,14 +74,16 @@ func GetClientAuthOptionsAndToken(ctx context.Context, config *StorageClientConf
...
-		retryConfig := NewRetryConfig(config)
-
 		apiCall := func(attemptCtx context.Context) (string, error) {
 			d, err := cred.UniverseDomain(attemptCtx)
 			return d, err
 		}
 
-		domain, err = ExecuteWithRetryAtLogLevel(ctx, retryConfig, "cred.UniverseDomain", "credentials", uuid.NewString(), apiCall, logger.LevelInfo)
+		if config.EnableMountRetries {
+			domain, err = ExecuteWithCustomShouldRetry(ctx, retryConfig, "cred.UniverseDomain", "credentials", uuid.NewString(), apiCall, ShouldRetryOnMount)
```
meet2mky: This was already to LogLevel INFO should we keep it at that log level only? I think this would not make logs noisy since it's going to be only at start and once.

dangarhs: There is already one INFO log level at L92 when its successful, so having TRACE here should be ok according to me.

internal/storage/storageutil/token_retry.go
dangarhs: Ack.

internal/storage/storageutil/auth_client_option.go
meet2mky: It logs the response but not when it started calling so I think we can add it and it will be just a single Log.

dangarhs: Done.

meet2mky (approved):
LGTM

PR #5013, opened 2026-08-11
test(write_test): log empty MediaLink attribute as a known gRPC limitation

vipnydav:
### Description
The `MediaLink` field is intentionally excluded from the GCS gRPC API specification because it is an HTTP-specific functionality. Previously, the integration test incorrectly attributed the missing `MediaLink` attribute to RAPID/zonal buckets. This PR updates the logic in `write_test.go` to correctly identify the empty `MediaLink` as a known gRPC limitation, simplifying the condition for Pirlo buckets and updating the log message accordingly.

### Testing details
1. Manual - Done
2. Unit tests - NA
3. Integration tests - NA.

None.

### DIFF
--- a/tools/integration_tests/operations/write_test.go
+++ b/tools/integration_tests/operations/write_test.go
@@ -120,8 +120,9 @@ func (w *writeOperationsTest) validateObjectAttributes(attr1, attr2 *storage.Obj
 		w.T().Error("Expected CRC32 attributes to be non 0")
 	}
 	if attr1.MediaLink == "" || attr2.MediaLink == "" {
-		if setup.IsZonalBucketRun() || (setup.IsPirloBucketRun() && w.isRapidWritesEnabled) {
-			w.T().Logf("media link is empty, but it is a known limitation in RAPID/zonal buckets.")
+		// TODO: Align test storage client protocol with the mount protocol. Note that when using gRPC to verify attributes for regional buckets, MediaLink will be empty.
+		if setup.IsZonalBucketRun() || setup.IsPirloBucketRun() {
+			w.T().Logf("media link is empty, but it is a known limitation in gRPC.")
 		} else {
 			w.T().Errorf("Expected media link to be non empty")
 		}


tools/integration_tests/operations/write_test.go
```diff
@@ -120,8 +120,8 @@ func (w *writeOperationsTest) validateObjectAttributes(attr1, attr2 *storage.Obj
 		w.T().Error("Expected CRC32 attributes to be non 0")
 	}
 	if attr1.MediaLink == "" || attr2.MediaLink == "" {
```
meet2mky, line 122: Nit: Shoudn't this also handle standard bucket with gRPC protocol?

vipnydav: I just checked, and the `operations` test package currently isn't run with `--client-protocol=grpc`. Also, for GKE, we don't have visibility into the flags used to mount the bucket. To support this, we would need to adopt a pattern of explicitly naming our tests (e.g., gRPC enabled/disabled) and passing the flag accordingly.

I can add a `TODO` to implement gRPC handling in the future. However, since we don't currently run these tests for gRPC, I think we shouldn't block this PR on it.

meet2mky: Please see: https://github.com/GoogleCloudPlatform/gcsfuse/blob/7c26b311cb059ef152cea97cf1953f334d540205/tools/integration_tests/test_config.yaml#L108

vipnydav: Added a `TODO`

vipnydav: Checking why this didn't fail for gRPC. Although it is documented that mediaLinks are not supported for gRPC: go/gcs-grpc-decisions#excluded-fields-from-object-spec-05-october-2022

vipnydav: I checked how this works, and the `--client-protocol=grpc` flag actually doesn't cause the `MediaLink` to be empty here. 

Here is why:
The `--client-protocol=grpc` flag only controls how the GCSFuse daemon connects to GCS. But when the test checks the object attributes, it uses its own [Go SDK client](https://github.com/GoogleCloudPlatform/gcsfuse/blob/7c26b311cb059ef152cea97cf1953f334d540205/tools/integration_tests/util/client/storage_client.go#L116).

For standard buckets, the test uses an HTTP client. Since it uses HTTP, the `MediaLink` is populated normally. 

For Zonal and Pirlo buckets, the test creates a gRPC client instead. The Go SDK over gRPC always leaves the `MediaLink` empty, regardless of how the object was originally created.

I tested this manually to be sure: if you create an object over gRPC but read it back over HTTP, the `MediaLink` is still there. It is only empty when you read the metadata using a gRPC client. Because our tests only use the gRPC client for Zonal and Pirlo runs, checking if it is a Zonal or Pirlo run is the right condition to know when the `MediaLink` will be empty.

meet2mky: Let's add a TODO to align the testing correctly so ensure the test client uses same protocol what GCSFuse used for mounting because it can mask or hide some bugs in future.

vipnydav: Agreed. Added a TODO.

meet2mky (approved):

PR #5017, opened 2026-08-11
feat(metrics): Add OpenTelemetry (OTLP) metrics exporting support

abhishek10004:
### Description
This PR introduces the ability to export GCSFuse metrics using OpenTelemetry (OTLP over HTTP), providing an alternative to the existing Cloud Monitoring metrics exporter. 

### Key Changes
* **New Configuration Flags**: Added experimental flags to control OTel metric exports:
  * `--experimental-enable-otel-metrics`: Enables the OTel metrics exporter (requires `cloud-metrics-export-interval-secs > 0`).
  * `--experimental-otel-metrics-endpoint`: Specifies the OTLP HTTP endpoint (defaults to `telemetry.googleapis.com`).
  * `--experimental-otel-metrics-project-id`: Overrides the GCP project ID for OTel metrics.
* **OTLP Metrics Exporter Integration**: Implemented `setupOtelMetricsEndpoint` to initialize the `otlpmetrichttp` exporter with GCP authentication support (when using `googleapis.com`).
* **Permission-Aware Degradation**: Wrapped the metrics exporter to gracefully handle `PermissionDenied` or `403 Forbidden` errors by disabling future export attempts instead of continuously failing.
* **Refactoring**: Consolidated shared logic for OTel resource generation and GCP auth client creation to be used consistently across both logs and metrics exporters.
* **Testing**: Added comprehensive unit tests (`otelexporters_test.go`) and end-to-end integration tests (`otel_metrics_test.go`) to validate the exporter's setup, permission handling, and metric delivery.

## Testing
- [x] Added unit tests for metric exporter setup and permission-aware error handling.
- [x] Added E2E integration tests to verify successful metric exporting.

b/535703440

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


internal/monitor/otelexporters.go
```diff
@@ -193,9 +206,74 @@ func (e *permissionAwareLogExporter) ForceFlush(ctx context.Context) error {
...
+
+	// 2. Check typed Google API error
+	var gErr *googleapi.Error
+	if errors.As(err, &gErr) && gErr.Code == http.StatusForbidden {
+		return true
+	}
+
+	// 3. Check specific HTTP 403 string patterns (e.g., OTLP HTTP exporter errors)
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "403 forbidden") ||
```
vadlakondaswetha, line 229: why are we not checking for just 403 string as earlier ?

internal/monitor/otelexporters.go
```diff
@@ -369,3 +431,44 @@ func SetupOTelLogExporter(ctx context.Context, endpoint string, mountID string,
 		return provider.Shutdown(ctx)
 	}, nil
 }
+
+// getOtelResource creates a base OTel resource and merges the GCP project ID if provided.
+func getOtelResource(ctx context.Context, mountID string, projectID string) (*resource.Resource, error) {
+	res, err := getResource(ctx, mountID)
+	if err != nil {
+		logger.Errorf("Error while fetching resource: %v", err)
+		if res == nil {
```
vadlakondaswetha, line 440: what is rationale behind returning only when res == nil and not for every error?

internal/monitor/otelexporters.go
```diff
@@ -193,9 +206,74 @@ func (e *permissionAwareLogExporter) ForceFlush(ctx context.Context) error {
...
+		strings.Contains(msg, "googleapi: error 403") ||
+		strings.Contains(msg, "\"code\": 403") ||
+		strings.Contains(msg, "\"code\":403")
+}
+
 func metricFormatter(m metricdata.Metrics) string {
 	return cloudMonitoringMetricPrefix + strings.ReplaceAll(m.Name, ".", "/")
 }
+
+func setupOtelMetricsEndpoint(ctx context.Context, endpoint string, authConfig cfg.GcsAuthConfig, intervalSecs int64) ([]metric.Option, error) {
```
vadlakondaswetha, line 241: do we need to repeat all this logic for otel logging exporter.
if not, why?

tools/integration_tests/otel_metrics/setup_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+var (
+	mockServerURL        string
+	mockServer           *http.Server
+	metricRecords        [][]byte
+	metricMu             sync.Mutex
+	configFileForGCSFuse string
+)
+
+func startMockServer() error {
```
vadlakondaswetha, line 47: i see this is all same as otel_logs. why are we not reusing stuff and put all the otel related tests in single package. https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/integration_tests/otel_logs/otel_logs_test.go

internal/monitor/otelexporters.go
abhishek10004: Because it was prone to false positives.

For example, if the OTel collector is running on a port containing "403" (such as 1403 or 4030), a connection refused error like dial tcp 127.0.0.1:1403: connect: connection refused will match this check and permanently disable the exporter.

internal/monitor/otelexporters.go
abhishek10004: In OpenTelemetry, resource.New returns a valid, partially-constructed *resource.Resource alongside an error (such as ErrPartialResource) when dynamic detectors like gcp.NewDetector() fail or run outside GCP (e.g., local testing).

If we returned on every err != nil, exporter initialization would fail entirely in non-GCP environments despite having valid static attributes (service.name, service.version, service.instance.id). Checking if res == nil allows us to log the detection error while gracefully proceeding with the partial resource.

internal/monitor/otelexporters.go
abhishek10004: Yes, the setup flow is needed for both, but the common logic has already been factored out into shared helpers:

1. SDK Type Separation: OTel Go maintains distinct packages and non-interchangeable option/exporter types for metrics (otlpmetrichttp) vs logs (otlploghttp).
2. Different OAuth Scopes: Metrics targets Cloud Monitoring (monitoring.write), whereas logs targets Cloud Logging (logging.write).
3. Pipeline Mechanisms: Metrics requires a metric.PeriodicReader tied to export intervals, while logs uses an asynchronous log.BatchProcessor with batch queueing.

All the underlying shared logic (OAuth client generation with scopes via getGcpOtelHttpClient, endpoint security check via isOtelEndpointInsecure, resource creation via getOtelResource, and 403 error handling via isPermissionDenied) is already deduplicated and shared across both setups.

vadlakondaswetha (approved):

PR #5020, opened 2026-08-12
refactor(locker & lru): Refactor `locker` and `lru` to isolate invariant checking and eliminate global state

meet2mky:
### Description
This PR addresses a severe O(N²) performance bottleneck in our tests caused by global lock state, and paves the way for explicitly configuring `locker` instances across the codebase.
Previously, `gEnableInvariantsCheck` in `locker.go` was a global variable. When a unit test enabled it, the O(N) `checkInvariants` loop was forcefully enabled for *all* concurrent tests sharing the process, causing large stress tests (like `lru_memory_test.go`) to hit massive slowdowns and flake/timeout in CI. 
This change introduces an explicit `locker.Options` configuration struct to replace the global state. Components can now explicitly configure and scope their lock behaviors (invariants checking and debug messaging) per instance.

### Key Changes
* **`internal/locker` Refactoring**: 
  * Introduced `locker.Options` and `NewWithOptions` / `NewRWWithOptions` to allow for explicit, localized lock configuration.
  * Marked existing global variables (`gEnableInvariantsCheck`, `gEnableDebugMessages`) and legacy constructors (`New`, `NewRW`) with `TODO`s for future migration and deprecation.
* **`internal/cache/lru` Migration**:
  * Added `NewCacheWithOptions` to allow constructing an LRU cache with specific lock configurations.
  * Migrated `lru_test.go` to explicitly pass `locker.Options{EnableInvariantsCheck: true}` instead of mutating global state. This successfully isolates the expensive invariant checks to just these tests!
* **`internal/fs` Lint Fix**: 
  * Grouped typed constants in `fs_test.go` to resolve a minor lint error.

### Why this matters
By passing down `locker.Options` instead of mutating package-level variables, tests will no longer accidentally sabotage the performance of other concurrent tests. It also sets up the architecture to be more idiomatic, testable, and dependency-injected moving forward.

[b/545172419](https://b.corp.google.com/issues/545172419)

### Testing details
This fixes the flake detector that was timing out from many days -

image

1. Manual - NA
2. Unit tests - Passed `go test ./internal/cache/lru/... ./internal/locker/...`
3. Integration tests - NA

NA (Legacy constructors and global functions are preserved for backward compatibility during migration)

### DIFF
--- a/internal/locker/locker.go
+++ b/internal/locker/locker.go
@@ -23,35 +23,45 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
 )
 
+// TODO: Migrate all usages of these global variables to explicitly instantiate and pass locker.Options.
 var gEnableInvariantsCheck bool
 var gEnableDebugMessages bool
 
-// Enable the check for invariants in the locks. Must be set before creating
+// EnableInvariantsCheck enables the check for invariants in the locks. Must be set before creating
 // any lockers.
+// TODO: Deprecate and delete this function once all components migrate to using locker.Options.
 func EnableInvariantsCheck() {
 	gEnableInvariantsCheck = true
 }
 
-// Enable the debug messages to diagnose dead locks. Must be set before creating
+// EnableDebugMessages enables the debug messages to diagnose dead locks. Must be set before creating
 // any lockers.
+// TODO: Deprecate and delete this function once all components migrate to using locker.Options.
 func EnableDebugMessages() {
 	gEnableDebugMessages = true
 }
 
 type Locker sync.Locker
 
-// Returns a locker with potential capability for debugging.
-func New(name string, check func()) Locker {
+// Options contains configuration for creating lockers.
+// By using options, callers can avoid relying on the package-level global variables.
+type Options struct {
+	EnableInvariantsCheck bool
+	EnableDebugMessages   bool
+}
+
+// NewWithOptions returns a locker with potential capability for debugging based on options.
+func NewWithOptions(name string, check func(), opts Options) Locker {
 	var l Locker = &sync.Mutex{}
 
-	if gEnableInvariantsCheck {
+	if opts.EnableInvariantsCheck {
 		l = &checker{
 			locker: l,
 			check:  check,
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/locker/locker.go
```diff
@@ -61,6 +71,24 @@ func New(name string, check func()) Locker {
...
+
+// NewWithChecker returns a locker that always checks invariants on Lock/Unlock.
+// TODO: Deprecate and delete this function once all components migrate to using locker.Options.
+func NewWithChecker(name string, check func()) Locker {
+	return NewWithOptions(name, check, Options{
+		EnableInvariantsCheck: true,
+		EnableDebugMessages:   gEnableDebugMessages,
+	})
+}
+
```
kislaykishore: while unrelated to this PR, there appears to be a concurrency issue in line 122 - we are logging d.holder. However, d.holder is a struct variable which is mutable via the `Unlock` method call. Instead, we should define a stack variable `holder`  and log that.
Consider removing `holder` from the struct as well.

internal/locker/locker.go
```diff
@@ -61,6 +71,24 @@ func New(name string, check func()) Locker {
...
+	return NewWithOptions(name, check, Options{
+		EnableInvariantsCheck: gEnableInvariantsCheck,
+		EnableDebugMessages:   gEnableDebugMessages,
+	})
+}
+
+// NewWithChecker returns a locker that always checks invariants on Lock/Unlock.
+// TODO: Deprecate and delete this function once all components migrate to using locker.Options.
+func NewWithChecker(name string, check func()) Locker {
+	return NewWithOptions(name, check, Options{
```
kislaykishore: Why is this method needed?

meet2mky: Thanks for catching. It's not needed. I added it initially and then I went ahead with NewWithOptions. I have removed it now.

internal/locker/locker.go
meet2mky: That is correct. I have also found this during this fix. Let's fix it in a follow up PR as this change is purely refactor while fixing race issue is code fix.

kislaykishore (approved):

PR #5021, opened 2026-08-13
feat(native-retries): report mount status and error classification to CSI sidecar

dangarhs:
### Description
- Implement mount status reporting subsystem in `internal/mountstatus` to communicate mount status to CSI sidecar over a named pipe specified via `GCSFUSE_STATUS_FD`.
- Report mount success (`codes.OK`) and mount failure classifications (`codes.PermissionDenied`, `codes.NotFound`).

b/545400301

### Testing details
1. Manual - Verified build and diff against master.
2. Unit tests - Ran storageutil unit tests.
3. Integration tests - NA

No

### DIFF
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -44,6 +44,7 @@ import (
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/logger"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/monitor"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/mount"
+	"github.com/googlecloudplatform/gcsfuse/v3/internal/mountstatus"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/profiler"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/storage/storageutil"
@@ -52,6 +53,7 @@ import (
 	"github.com/jacobsa/fuse"
 	"github.com/kardianos/osext"
 	"github.com/spf13/viper"
+	"google.golang.org/grpc/codes"
 )
 
 const (
@@ -386,6 +388,8 @@ func logGCSFuseMountInformation(mountInfo *mountInfo) {
 func Mount(mountInfo *mountInfo, bucketName, mountPoint string) (err error) {
 	newConfig := mountInfo.config
 
+	mountstatus.InitStatusReporter(newConfig.Foreground)
+
 	var logExporterShutdownFn common.ShutdownFn
 	if newConfig.Foreground {
 		err = logger.InitLogFile(newConfig.Logging, fsName(bucketName), newConfig.MountId)
@@ -516,6 +520,7 @@ func Mount(mountInfo *mountInfo, bucketName, mountPoint string) (err error) {
 			}
 			// Print the success message in the log-file/stdout depending on what the logger is set to.
 			logger.Info(SuccessfulMountMessage)
+			mountstatus.ReportStatus(codes.OK, SuccessfulMountMessage)
 			callDaemonizeSignalOutcome(nil)
 		}
 
--- a/internal/mountstatus/status.go
+++ b/internal/mountstatus/status.go
@@ -1,0 +1,111 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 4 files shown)


cmd/legacy_main.go
```diff
@@ -524,6 +528,9 @@ func Mount(mountInfo *mountInfo, bucketName, mountPoint string) (err error) {
 			// mounting gcsfuse in foreground mode. But this is important to avoid
 			// losing error logs when run in the background mode.
 			logger.Errorf("%s: %v\n", UnsuccessfulMountMessagePrefix, err)
+			if code, ok := storageutil.DetermineMountStatusCode(err); ok {
```
meet2mky: Let's not add this right now to keep the scope of changes minimum for native-retries. We can do this once we fully migrate to status file FD later.

internal/mountstatus/status.go
```diff
@@ -0,0 +1,167 @@
...
+
+var (
+	statusMu   sync.Mutex
+	statusFile *os.File
+)
+
+// MountState defines the lifecycle stage of GCSFuse mounting.
+type MountState string
+
+const (
```
meet2mky: Let's not add the MountState as it's not going to be be useful.

internal/mountstatus/status.go
```diff
@@ -0,0 +1,167 @@
...
+		return
+	}
+
+	f, err := openStatusPipe(statusFDStr)
+	if err != nil {
+		logger.Warnf("Failed to initialize status reporter from GCSFUSE_STATUS_FD: %v", err)
+		return
+	}
+
+	statusMu.Lock()
```
meet2mky: We don't really want to have mutex because it's not expected to call InitStatusReporter more than once or concurrently. 

I think as a defence in the init we can check is statusFile is not nil then return early without doing re-initialization.

internal/mountstatus/status.go
```diff
@@ -0,0 +1,167 @@
...
+
+// ReportInProgress reports an intermediate state during mount retries (e.g. transient 403/404).
+func ReportInProgress(code codes.Code, errMsg string) {
+	ReportStatus(StateInProgress, code, errMsg)
+}
+
+// ReportSuccess reports a successful mount completion (StateSuccess with codes.OK) and closes the pipe.
+func ReportSuccess(errMsg string) {
+	ReportStatus(StateSuccess, codes.OK, errMsg)
+	closeStatusPipe()
```
meet2mky: Why do we need to close the pipe?

internal/mountstatus/status.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil, fmt.Errorf("fcntl F_GETFL failed: %w", err)
+	}
+
+	if accMode := flags & unix.O_ACCMODE; accMode != unix.O_WRONLY && accMode != unix.O_RDWR {
+		return nil, errors.New("descriptor is not open for writing")
+	}
+
+	// Set close-on-exec so child processes do not inherit the status pipe.
+	unix.CloseOnExec(fd)
+	return f, nil
```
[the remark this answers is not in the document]
dangarhs: InitStatusReporter executes only once at process startup on an externally provided FD, so there is no risk of an accumulating leak and Go's runtime finalizer cleans up unreferenced *os.File handles on failure.

internal/mountstatus/status.go
dangarhs: Not required. Removed.

internal/mountstatus/status.go
dangarhs: Removed MountState and pipe closing, statusFile never changes, so no mutex is needed.

internal/mountstatus/status_test.go
```diff
@@ -0,0 +1,261 @@
...
+func setStatusFileForTesting(f *os.File) func() {
+	old := statusFile
+	statusFile = f
+
+	return func() {
+		statusFile = old
+	}
+}
+
+func TestOpenStatusPipe(t *testing.T) {
```
meet2mky: Nit: Let's break down the test in table driven manner or simple unit tests, same for other.

meet2mky (dismissed):
Looks good to me. Just a comment about splitting test into simple unit tests.

internal/mountstatus/status_test.go
dangarhs: This is taken care of.

internal/mountstatus/status_test.go
meet2mky, line 1: Nit: Let's remove the empty lines within unit tests unless they are separating Arrange Act Assert block.

meet2mky (dismissed):
Looks good. Could you remove the empty lines in unit test file. Thanks!

internal/mountstatus/status_test.go
dangarhs: Done

meet2mky (approved):
LGTM

PR #5022, opened 2026-08-13
fix: Setting expires field explicitly based on expiresIn

vadlakondaswetha:
go auth2 library populates expiry field based on expiresIn. Since GCSFuse is doing custom unmarshalling we need to explicitly populate it to handle scenarios where the token only has expiresIn field set.
https://github.com/golang/oauth2/blob/master/internal/token.go#L79
bug fix: https://buganizer.corp.google.com/u/1/issues/543949965

### Testing details
1. Manual - Done
2. Unit tests - Added
3. Integration tests - NA

### DIFF
--- a/internal/auth/token_source.go
+++ b/internal/auth/token_source.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"time"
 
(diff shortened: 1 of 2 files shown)


abhishek10004 (approved):
LGTM

PR #5024, opened 2026-08-13
fix(locker): resolve data race by removing holder struct field

meet2mky:
### Description
This PR fixes a data race condition in the deadlock detection logic of the `locker` and `rw_locker` packages.

**Root Cause:**
When a lock was held for exactly 5 seconds, the `time.AfterFunc` background timer would fire to log a deadlock warning. This background goroutine read the `holder` string from the `debugger` struct at the exact same time the main goroutine cleared it (`d.holder = ""`) inside the `Unlock()` method. This concurrent read and write to the same memory address without synchronization resulted in a data race.

**Fix:**
* Removed the `holder` struct field from the `debugger` and `rwDebugger` types since it wasn't needed.
* In `Lock()`, captured the stack trace into a local, immutable `holder` variable and passed it into the closure for the background timer, ensuring thread-safe access.
* Added a `nil` check for the timer cleanup in `Unlock()` to prevent nil pointer panics and preserve the standard library's `"sync: unlock of unlocked mutex"` panic message when unlocking an uninitialized locker.

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

### DIFF
--- a/internal/locker/locker.go
+++ b/internal/locker/locker.go
@@ -98,7 +98,6 @@ func (c *checker) Unlock() {
 type debugger struct {
 	locker Locker
 	name   string
-	holder string
 	timer  *time.Timer
 }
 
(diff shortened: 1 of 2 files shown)


internal/locker/locker.go
```diff
@@ -107,17 +106,18 @@ func (d *debugger) Lock() {
 
 	buf := make([]byte, 2048)
 	runtime.Stack(buf, false /* all */)
-	d.holder = string(buf)
+	holder := string(buf)
```
kislaykishore: To avoid uninitialized values in string. Ditto for rw_locker.go
```
n := runtime.Stack(buf, false /* all */)
holder := string(buf[:n])
```

meet2mky: Thanks for catching this. I have updated the code at both places.

kislaykishore (approved):

PR #5025, opened 2026-08-13
fix(caching): only cache negative entries from authoritative listings

### LINKED ISSUE #5018: Windowed listings can poison the stat cache and cause false `ENOENT` for existing implicit directories
## Describe the issue

When gcsfuse mounts a flat-namespace bucket with implicit directories and the negative metadata cache enabled, an empty **windowed** `ListObjects` result can overwrite an existing positive stat-cache entry for the directory with a negative entry.

For the duration of `--metadata-cache-negative-ttl-secs` (5 seconds by default), gcsfuse can then return `ENOENT` for a directory that still exists, as well as for paths beneath it. The result is served from the local cache without a Cloud Storage request, so there may be no corresponding failed request in Cloud Storage logs or metrics. The directory becomes visible again after the negative entry expires.

The [metadata prefetcher](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/fs/inode/dir_prefetcher.go#L134-L173) provides a concrete path to this state. Once a directory is classified as large, later prefetches use the looked-up child name as [`ListObjectsRequest.StartOffset`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/storage/gcs/request.go#L367-L369). If that name sorts after all existing children, Cloud Storage correctly returns an empty listing for that window. [`fastStatBucket.insertListing`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/storage/caching/fast_stat_bucket.go#L118-L160) currently treats the empty result as proof that the entire prefix does not exist and calls `AddNegativeEntry` for the directory.

This appears to affect the negative stat-cache implementation introduced by [PR #4729](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4729) in [commit `d830c5c77`](https://github.com/GoogleCloudPlatform/gcsfuse/commit/d830c5c77de7995d5d5d143a389832c8b732c859), released in [v3.11.0](https://github.com/GoogleCloudPlatform/gcsfuse/releases/tag/v3.11.0). I reproduced it on unpatched [commit `446c9ea04`](https://github.com/GoogleCloudPlatform/gcsfuse/commit/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5) and observed the production symptom with [gcsfuse 3.11.2](https://github.com/GoogleCloudPlatform/gcsfuse/releases/tag/v3.11.2).

### User-visible impact

- Existing directories transiently return `ENOENT`.
- All lookups beneath the affected directory can fail during the negative-cache window.
- Multiple processes using the same mount can fail at approximately the same time.
- The condition normally self-heals after the negative TTL, making it appear intermittent.
- Because the rejection can occur entirely in the cache, the failure may have no matching Cloud Storage request or server-side error.

### Relevant configuration

The production observation used a mount equivalent to:

```bash
gcsfuse --implicit-dirs BUCKET_NAME /path/to/mountpoint
```

All metadata-cache settings were left at their defaults, which means:

- Negative metadata caching enabled ([`--metadata-cache-negative-ttl-secs=5`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/cfg/params.yaml#L1053-L1060))
- Metadata prefetch enabled ([`--enable-metadata-prefetch=true`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/cfg/params.yaml#L990-L998))
- [`--metadata-prefetch-entries-limit=5000`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/cfg/params.yaml#L1031-L1041)

The bucket uses a flat namespace, and the affected directory contained approximately 6,400 children — above the default `--metadata-prefetch-entries-limit=5000`, which is what causes the directory to be classified as large and subsequent prefetches to be windowed.

Setting `--metadata-cache-negative-ttl-secs=0` avoided the observed failures under otherwise equivalent load.

## System and version

- OS: Linux
- Platform: GCE VM
- Version: gcsfuse 3.11.2
- Bucket type: flat namespace (non-HNS)
- Also reproduced deterministically at the storage/cache layer on [commit `446c9ea04`](https://github.com/GoogleCloudPlatform/gcsfuse/commit/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5)

## Steps to reproduce

TRACE logs were not captured during the production event; the deterministic reproducer below substitutes for them. It reproduces the cache corruption without requiring a FUSE mount or access to a live bucket, using the real [`fastStatBucket`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/storage/caching/fast_stat_bucket.go), the real [LRU/stat-cache implementation](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/cache/metadata/stat_cache.go), and the repository's [fake Cloud Storage backend](https://github.com/GoogleCloudPlatform/gcsfuse/tree/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/storage/fake).

Add this test to `internal/storage/caching/integration_test.go` on an unpatched checkout:

```go
func TestIntegration_WindowedListingDoesNotPoisonExistingDirectory(t *testing.T) {
	deps := setupIntegrationTest(t)
	const dirName = "dir/"

	_, err := storageutil.CreateObject(
		deps.ctx,
		deps.wrapped,
		dirName+"a",
		[]byte("x"),
	)
	require.NoError(t, err)

	// Populate a positive cache entry for the existing implicit directory.
	listing, err := deps.bucket.ListObjects(deps.ctx, &gcs.ListObjectsRequest{
		Prefix:    dirName,
		Delimiter: "/",
	})
	require.NoError(t, err)
	require.NotEmpty(t, listing.MinObjects)

	statReq := &gcs.StatObjectRequest{
		Name:               dirName,
		FetchOnlyFromCache: true,
	}

	dir, _, err := deps.bucket.StatObject(deps.ctx, statReq)
	require.NoError(t, err)
	require.NotNil(t, dir)

	// Model a metadata-prefetch listing whose window begins after the final
	// child. The empty result describes only this window, not the whole prefix.
	listing, err = deps.bucket.ListObjects(deps.ctx, &gcs.ListObjectsRequest{
		Prefix:      dirName,
		Delimiter:   "/",
		StartOffset: dirName + "zzz",
	})
	require.NoError(t, err)
	require.Empty(t, listing.MinObjects)
	require.Empty(t, listing.CollapsedRuns)
	require.Empty(t, listing.ContinuationToken)

	// The implicit directory still exists and should remain positively cached.
	dir, _, err = deps.bucket.StatObject(deps.ctx, statReq)
	assert.NoError(t, err)
	assert.NotNil(t, dir)
}
```

Run only the reproducer using the repository's Go version in Docker:

```bash
docker run --rm \
  --mount type=bind,src="$PWD",dst=/workspace \
  -w /workspace \
  golang:1.26.5 \
  go test ./internal/storage/caching \
    -run TestIntegration_WindowedListingDoesNotPoisonExistingDirectory \
    -count=1 -v
```

### Actual behavior

On the unpatched implementation, the final lookup fails because the empty windowed listing replaced the positive `dir/` entry with a negative entry:

```text
gcs.NotFoundError: negative cache entry for dir/
```

The test therefore fails at the final `assert.NoError` and `assert.NotNil` checks even though `dir/a` still exists in the wrapped bucket.

### Expected behavior

An empty windowed or continuation-page listing should not create a negative cache entry for the whole prefix. The existing positive entry for `dir/` should remain available, and the test should pass.

## Root cause analysis

[`fastStatBucket.insertListing`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/storage/caching/fast_stat_bucket.go#L118-L160) currently uses the absence of objects and collapsed runs as sufficient evidence that a non-root directory does not exist:

```go
isNegativeCacheEnabled := b.negativeCacheTTL > 0 && !b.enableEmptyManagedFolders
isEmptyNonRootDir := !dirHasContents && dirName != ""
if isNegativeCacheEnabled && isEmptyNonRootDir {
	b.cache.AddNegativeEntry(dirName, b.clock.Now().Add(b.negativeCacheTTL))
}
```

That inference is sound only when the listing is authoritative for the complete prefix. It is not sound when:

- `ListObjectsRequest.StartOffset` is set, because the response covers only the namespace at or after that offset;
- the request has a continuation token, because it is not the first page; or
- the response has a continuation token, because the listing is incomplete.

[`AddNegativeEntry`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/cache/metadata/stat_cache.go#L43-L46) intentionally overwrites an existing cache entry ([implementation](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/cache/metadata/stat_cache.go#L248-L260)). Consequently, an empty result for only part of the namespace can replace a valid positive implicit-directory entry.

The metadata-prefetch path makes this reachable in normal operation:

1. A prefetch exceeding `metadata-prefetch-entries-limit` [marks the directory as large](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/fs/inode/dir_prefetcher.go#L142-L179).
2. A later [uncached child lookup triggers the prefetcher](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/fs/inode/dir.go#L770-L776), which [builds a windowed listing request](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/fs/inode/dir.go#L909-L933) using that child name as `StartOffset`.
3. If the name sorts after every existing child, the windowed listing is empty.
4. `insertListing` writes a negative entry for the entire directory prefix.
5. If the alternative file-form key is also negative, [`dirInode.LookUpChild` returns not found from cache](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/fs/inode/dir.go#L668-L714) without contacting Cloud Storage.
6. The false result persists until the negative-cache TTL expires.

The `StartOffset` behavior is documented by the [Cloud Storage Objects `list` API](https://cloud.google.com/storage/docs/json_api/v1/objects/list).

## Production observation

This was observed with a directory containing approximately 6,400 children while 13 processes created small files at an aggregate rate of approximately 55 files per second. All 13 processes received `ENOENT` from `open(..., O_CREAT)` at approximately the same time. The mount recovered without intervention within several seconds.

Two additional same-day observations on the same prefix included a recursive tree walk failing and a directory listing returning a child whose subsequent `open` returned `ENOENT`. Cloud Storage metrics showed no corresponding service-side error responses for the relevant window. Disabling negative metadata caching eliminated the failures under equivalent load.

## Suggested resolution

Only derive a negative directory entry from an empty listing when the result is authoritative for the complete prefix. For example, require all of the following before calling `AddNegativeEntry`:

```go
req.StartOffset == "" &&
req.ContinuationToken == "" &&
listing.ContinuationToken == ""
```

Note that [`insertListing` currently receives only the listing and directory name](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/storage/caching/fast_stat_bucket.go#L118-L122), so this guard requires plumbing the `*gcs.ListObjectsRequest` through from [`ListObjects`](https://github.com/GoogleCloudPlatform/gcsfuse/blob/446c9ea04ad54f6b8621b8022b32dcd85ccf45a5/internal/storage/caching/fast_stat_bucket.go#L461-L483) into `insertListing`.

An unwindowed request returning no contents and no continuation token remains valid evidence that the prefix is empty. `MaxResults` does not require a separate guard in that case: if zero results are returned without a continuation token, the cap did not truncate any results.

I can submit a focused PR with the change and regression coverage if maintainers agree with this direction.

## Additional context

A closely matching failure mode was [flagged by the automated review](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4729#discussion_r3297136734) (gemini-code-assist) on the original negative stat-cache PR: an empty paginated listing could overwrite a valid positive entry and cause `ENOENT`.

The concern was [acknowledged by the PR author](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4729#discussion_r3301272488), and the listing-derived negative insertion was removed during that review.

The merged implementation again creates negative entries from empty listings, but does not distinguish complete-prefix listings from windowed or paginated listings.

No duplicate public issue was found when searching for combinations of `ENOENT`, implicit directories, negative stat cache, metadata prefetch, `StartOffset`, and empty listings.

nk412:
### Description

Prevent empty windowed or paginated object listings from creating a negative stat-cache entry for an entire implicit-directory prefix.

`fastStatBucket.insertListing` previously treated every empty listing as evidence that the requested directory did not exist. That is not authoritative when the request uses `StartOffset` or a continuation token, or when the response contains a continuation token. In those cases, the listing represents only part of the namespace, and inserting a negative entry can overwrite a valid positive directory entry and cause transient `ENOENT` responses.

This change:

- passes the originating `ListObjectsRequest` to `insertListing`;
- limits listing-derived negative entries to unwindowed, complete listings;
- preserves negative caching for authoritative empty listings; and
- adds request-level coverage for start offsets and continuation tokens, plus a regression test using the real LRU/stat-cache implementation.

Fixes #5018

### Testing details

1. Manual - Verified the deterministic regression fails on the unpatched implementation with `gcs.NotFoundError: negative cache entry for dir/` and passes with this change. `make build` also passes.
2. Unit tests - Passed in `golang:1.26.5` Docker:
   - `go test ./internal/storage/caching -count=1`
   - `CGO_ENABLED=0 go test ./internal/storage/... -count=1`
   - `go test -race ./internal/storage/caching -count=1`
   - `go vet ./internal/storage/caching`
   - `CGO_ENABLED=0 go build ./...`
3. Integration tests - Added `TestIntegration_WindowedListingDoesNotPoisonExistingDirectory`, which exercises `fastStatBucket` with the real LRU/stat cache and fake GCS backend. A live FUSE/GCS end-to-end test was not run.

N/A. The changed method is internal, and no exported API, configuration, or existing authoritative-listing behavior changes.

### DIFF
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -115,6 +115,15 @@ func (b *fastStatBucket) insertMultiple(objs []*gcs.Object) {
 	}
 }
 
+// isAuthoritativeListing returns true if the listing represents a complete,
+// un-windowed, and non-paginated view of the entire requested prefix.
+func isAuthoritativeListing(req *gcs.ListObjectsRequest, listing *gcs.Listing) bool {
+	if req == nil || listing == nil {
+		return false
+	}
+	return req.StartOffset == "" && req.ContinuationToken == "" && listing.ContinuationToken == ""
+}
+
 // LOCKS_EXCLUDED(b.mu)
 // insertListing caches all objects and sub-directories discovered during a GCS listing.
 // It explicitly handles the "implicit directory" edge case where a directory exists
@@ -155,17 +164,13 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 	// If enableEmptyManagedFolders is true, do not negatively cache empty directories,
 	// otherwise existing positive entries for valid empty managed folders get overwritten.
 	//
-	// An empty listing proves the directory doesn't exist only when it is
-	// authoritative for the whole prefix: a windowed listing (StartOffset or
-	// ContinuationToken set on the request) covers a slice of the namespace, and
-	// an empty page that carries a continuation token is an incomplete listing.
-	// Caching a negative entry from such a listing would overwrite a valid
-	// positive entry and serve ENOENT for the entire subtree until the negative
-	// TTL expires.
+	// Only an authoritative listing proves the directory doesn't exist: caching
+	// a negative entry from a windowed or paginated listing would overwrite a
+	// valid positive entry and serve ENOENT for the entire subtree until the
+	// negative TTL expires.
 	isNegativeCacheEnabled := b.negativeCacheTTL > 0 && !b.enableEmptyManagedFolders
 	isEmptyNonRootDir := !dirHasContents && dirName != ""
-	isAuthoritativeListing := req.StartOffset == "" && req.ContinuationToken == "" && listing.ContinuationToken == ""
-	if isNegativeCacheEnabled && isEmptyNonRootDir && isAuthoritativeListing {
+	if isNegativeCacheEnabled && isEmptyNonRootDir && isAuthoritativeListing(req, listing) {
 		b.cache.AddNegativeEntry(dirName, b.clock.Now().Add(b.negativeCacheTTL))
 	}
 
--- a/internal/storage/caching/fast_stat_bucket.go
+++ b/internal/storage/caching/fast_stat_bucket.go
@@ -119,7 +119,7 @@ func (b *fastStatBucket) insertMultiple(objs []*gcs.Object) {
 // insertListing caches all objects and sub-directories discovered during a GCS listing.
 // It explicitly handles the "implicit directory" edge case where a directory exists
 // only as a prefix to other objects.
-func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing, dirName string) {
+func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing, req *gcs.ListObjectsRequest) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
@@ -129,6 +129,7 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 		return
 	}
 
+	dirName := req.Prefix
 	expiration := b.clock.Now().Add(b.primaryCacheTTL)
 
 	// 1. Parent Directory Inference (Implicit Check)
@@ -153,9 +154,18 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
 	// Negative Cache (Only if it's a directory and there are NO contents)
 	// If enableEmptyManagedFolders is true, do not negatively cache empty directories,
 	// otherwise existing positive entries for valid empty managed folders get overwritten.
+	//
+	// An empty listing proves the directory doesn't exist only when it is
+	// authoritative for the whole prefix: a windowed listing (StartOffset or
+	// ContinuationToken set on the request) covers a slice of the namespace, and
+	// an empty page that carries a continuation token is an incomplete listing.
+	// Caching a negative entry from such a listing would overwrite a valid
+	// positive entry and serve ENOENT for the entire subtree until the negative
+	// TTL expires.
 	isNegativeCacheEnabled := b.negativeCacheTTL > 0 && !b.enableEmptyManagedFolders
 	isEmptyNonRootDir := !dirHasContents && dirName != ""
-	if isNegativeCacheEnabled && isEmptyNonRootDir {
+	isAuthoritativeListing := req.StartOffset == "" && req.ContinuationToken == "" && listing.ContinuationToken == ""
+	if isNegativeCacheEnabled && isEmptyNonRootDir && isAuthoritativeListing {
 		b.cache.AddNegativeEntry(dirName, b.clock.Now().Add(b.negativeCacheTTL))
 	}
 
@@ -474,7 +484,7 @@ func (b *fastStatBucket) ListObjects(
 	}
 
 	if b.isTypeCacheDeprecated {
-		b.insertListing(ctx, listing, req.Prefix)
+		b.insertListing(ctx, listing, req)
 	} else {
 		// note anything we found.
 		b.insertMultipleMinObjects(ctx, listing.MinObjects)
--- a/internal/storage/caching/fast_stat_bucket_test.go
+++ b/internal/storage/caching/fast_stat_bucket_test.go
@@ -1062,6 +1062,33 @@ func (t *ListObjectsTest_InsertListing) EmptyPageWithResponseContinuationTokenDo
 	t.callAndVerifyWithRequest(context.TODO(), false, listing, req, nil, nil, false)
 }
 
+// A non-empty windowed listing (StartOffset set) must insert the returned
+// objects and still infer the parent implicit directory.
+func (t *ListObjectsTest_InsertListing) NonEmptyListingWithStartOffsetInsertsPositiveEntries() {
+	listing := &gcs.Listing{
+		MinObjects: []*gcs.MinObject{
+			{Name: "dir/c", Size: 10},
+		},
+	}
+	req := &gcs.ListObjectsRequest{
+		Prefix:      "dir/",
+		StartOffset: "dir/b",
+	}
+	expectedInserts := []*gcs.MinObject{
+		{Name: "dir/c", Size: 10},
+	}
+	expectedImplicitDirs := []string{"dir/"}
+	t.callAndVerifyWithRequest(
+		context.TODO(),
+		false, // isHNS
+		listing,
+		req,
+		expectedInserts,
+		expectedImplicitDirs,
+		false, // expectNegativeEntry
+	)
+}
+
 func (t *ListObjectsTest_InsertListing) EmptyListing() {
 	listing := &gcs.Listing{}
 	expectedInserts := []*gcs.MinObject{}
--- a/internal/storage/caching/fast_stat_bucket_test.go
+++ b/internal/storage/caching/fast_stat_bucket_test.go
@@ -1078,6 +1078,7 @@ func (t *ListObjectsTest_InsertListing) NonEmptyListingWithStartOffsetInsertsPos
 		{Name: "dir/c", Size: 10},
 	}
 	expectedImplicitDirs := []string{"dir/"}
+
 	t.callAndVerifyWithRequest(
 		context.TODO(),
 		false, // isHNS
--- a/internal/storage/caching/fast_stat_bucket_test.go
+++ b/internal/storage/caching/fast_stat_bucket_test.go
@@ -1007,6 +1007,11 @@ func (t *ListObjectsTest_InsertListing) SetUp(ti *TestInfo) {
 }
 
 func (t *ListObjectsTest_InsertListing) callAndVerify(ctx context.Context, isHNS bool, listing *gcs.Listing, prefix string, expectedInserts []*gcs.MinObject, expectedImplicitDirs []string) {
+	expectNegativeEntry := len(listing.MinObjects) == 0 && len(listing.CollapsedRuns) == 0 && prefix != "" && listing.ContinuationToken == ""
+	t.callAndVerifyWithRequest(ctx, isHNS, listing, &gcs.ListObjectsRequest{Prefix: prefix}, expectedInserts, expectedImplicitDirs, expectNegativeEntry)
+}
+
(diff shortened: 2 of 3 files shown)


internal/storage/caching/fast_stat_bucket.go
```diff
@@ -153,9 +154,18 @@ func (b *fastStatBucket) insertListing(ctx context.Context, listing *gcs.Listing
...
+	// authoritative for the whole prefix: a windowed listing (StartOffset or
+	// ContinuationToken set on the request) covers a slice of the namespace, and
+	// an empty page that carries a continuation token is an incomplete listing.
+	// Caching a negative entry from such a listing would overwrite a valid
+	// positive entry and serve ENOENT for the entire subtree until the negative
+	// TTL expires.
 	isNegativeCacheEnabled := b.negativeCacheTTL > 0 && !b.enableEmptyManagedFolders
 	isEmptyNonRootDir := !dirHasContents && dirName != ""
-	if isNegativeCacheEnabled && isEmptyNonRootDir {
+	isAuthoritativeListing := req.StartOffset == "" && req.ContinuationToken == "" && listing.ContinuationToken == ""
```
meet2mky: Nit: Can we refactor this into a helper method along with it's definition. 
```suggestion
// isAuthoritativeListing returns true if the listing represents a complete,
// un-windowed, and non-paginated view of the entire requested prefix.
func isAuthoritativeListing(req *gcs.ListObjectsRequest, listing *gcs.Listing) bool {
	if req == nil || listing == nil {
		return false
	}
	return req.StartOffset == "" && req.ContinuationToken == "" && listing.ContinuationToken == ""
}

```

internal/storage/caching/fast_stat_bucket_test.go
```diff
@@ -1019,17 +1024,44 @@ func (t *ListObjectsTest_InsertListing) callAndVerify(ctx context.Context, isHNS
...
+
+// An empty page that carries a continuation token is an incomplete listing and
+// must not produce a negative entry for the directory.
+func (t *ListObjectsTest_InsertListing) EmptyPageWithResponseContinuationTokenDoesNotCacheNegativeEntry() {
+	listing := &gcs.Listing{ContinuationToken: "next-page"}
+	req := &gcs.ListObjectsRequest{Prefix: "dir/"}
+
+	t.callAndVerifyWithRequest(context.TODO(), false, listing, req, nil, nil, false)
+}
+
```
meet2mky, line 1064: Can you please add a positive test as well?

```
// A non-empty windowed listing (StartOffset set) must insert the returned
// objects and still infer the parent implicit directory.
func (t *ListObjectsTest_InsertListing) NonEmptyListingWithStartOffsetInsertsPositiveEntries() {
	listing := &gcs.Listing{
		MinObjects: []*gcs.MinObject{
			{Name: "dir/c", Size: 10},
		},
	}
		Prefix:      "dir/",
		StartOffset: "dir/b",
	}
	expectedInserts := []*gcs.MinObject{
		{Name: "dir/c", Size: 10},
	}
	expectedImplicitDirs := []string{"dir/"}
	t.callAndVerifyWithRequest(
		context.TODO(),
		false, // isHNS
		listing,
		req,
		expectedInserts,
		expectedImplicitDirs,
		false, // expectNegativeEntry
	)
}

```

meet2mky (commented):
Changes look good 👍🏻. Can you address the small comment. Thanks!

nk412:
Done. Thanks for the eyes, @meet2mky

meet2mky:
> Done. Thanks for the eyes, @meet2mky 

Can you also format unit tests in AAA format separating them with an empty lines. Sorry for back and forth.

nk412:
Done! The new positive test now has the arrange block separated from the call. The other added tests already had that separation. Happy to add explicit Arrange/Act/Assert labels if that's what you meant.

alleaditya (approved):

meet2mky (approved):
LGTM

PR #5026, opened 2026-08-14
fix(bufferedwrites): prevent memory leak & race in UploadHandler.Destroy

meet2mky:
## Description
When a file handle is destroyed (e.g., after a failed upload or unlink), `UploadHandler.Destroy` did not cancel the ongoing GCS upload context or wait for the background `uploader` goroutine to finish processing its current chunk.

This caused the `uploader` goroutine to return its in-flight block back to the block pool *after* `ClearFreeBlockChannel` had already executed and finished teardown. In production, this resulted in severe consequences:
1. **Memory Leaks:** The late-arriving block was abandoned in the pool, meaning `Deallocate()` was never called and the underlying `mmap` memory was never freed.
2. **Global Write Degredation:** The global semaphore permit (`globalMaxBlocksSem`) tied to the abandoned block was never released. Over time, these leaked permits would exhaust the semaphore capacity, causing all new writes across the GCSFuse mount to use staged writes.

This PR fixes the teardown sequence and synchronization:
- Cancels the GCS upload context inside `Destroy()` to quickly abort the in-flight upload.
- Adds `uh.wg.Wait()` after draining `uploadCh` to ensure the `uploader` goroutine fully completes and releases its block before teardown continues.
- Introduces `TestDestroyWithUploaderGoroutineInProgress` to assert strict synchronization between `Destroy` and the `uploader`.

## How Has This Been Tested?
- Ran `go test -v -count=100 -run TestBufferedWriteTestSuite/TestDestroyShouldClearFreeBlockChannel ./internal/bufferedwrites/...` and verified it passes 100% of the time.
- Implemented and verified the new `TestDestroyWithUploaderGoroutineInProgress` unit test to directly trigger and assert this race condition.
- Confirmed that the new test fails without the fix and succeeds with it.

b/546299575

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

NA

### DIFF
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -98,6 +98,7 @@ func (uh *UploadHandler) Upload(ctx context.Context, block block.Block) error {
 
 	err := uh.ensureWriter(ctx)
 	if err != nil {
+		uh.wg.Done()
 		return fmt.Errorf("uh.ensureWriter() failed: %v", err)
 	}
 	// Start the uploader goroutine but only once.
--- a/internal/bufferedwrites/upload_handler.go
+++ b/internal/bufferedwrites/upload_handler.go
@@ -269,20 +269,28 @@ func (uh *UploadHandler) AwaitBlocksUpload() {
 }
 
 func (uh *UploadHandler) Destroy() {
-	// Move all pending blocks to freeBlockCh and close the channel if not done.
+	// Cancel the context to abort any ongoing GCS upload.
+	if uh.cancelFunc != nil {
+		uh.cancelFunc()
+	}
+
+	// Drain all pending blocks from uploadCh and return them to the block pool,
+	// then close uploadCh and wait for the uploader goroutine to finish.
 	for {
 		select {
 		case currBlock, ok := <-uh.uploadCh:
-			// Not ok means channel closed. Return.
+			// If ok is false, the channel was already closed (e.g. by Finalize).
 			if !ok {
 				return
 			}
 			uh.blockPool.Release(currBlock)
-			// Marking as wg.Done to ensure any waiters are unblocked.
+			// Mark as wg.Done for each drained block to ensure wg.Wait() can unblock.
 			uh.wg.Done()
 		default:
-			// This will get executed when there are no blocks pending in uploadCh and its not closed.
+			// Executed when there are no more blocks pending in uploadCh and it is not closed.
 			close(uh.uploadCh)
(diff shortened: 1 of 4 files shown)


internal/bufferedwrites/upload_handler_test.go
```diff
@@ -610,3 +610,36 @@ func (t *UploadHandlerTest) createBlocks(count int) []block.Block {
...
+	}).Return(4, nil)
+	t.mockBucket.On("BucketType").Return(gcs.BucketType{})
+	t.mockBucket.On("CreateObjectChunkWriter", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(writer, nil)
+	blocks := t.createBlocks(1)
+	b := blocks[0]
+	_, err := b.Write([]byte("taco"))
+	require.NoError(t.T(), err)
+	err = t.uh.Upload(context.Background(), b)
+	require.NoError(t.T(), err)
+	time.Sleep(50 * time.Millisecond) // Give the uploader goroutine time to pull the block and call Write()
```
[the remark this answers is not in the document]
meet2mky: Fixed

internal/bufferedwrites/upload_handler.go
abhishek10004, line 102: The wait in Destroy exposes a waitgroup leak here. We should call uh.wg.Done() when ensureWriter fails.

abhishek10004: Also, let's add a UT for this as well.

meet2mky: Done. Thanks for catching this. I found one more instance where it was leaked when current block is partially filled and destroy is called with Sync() or Close() causing it to leak. Have added UT and change for that as well.

internal/bufferedwrites/upload_handler_test.go
```diff
@@ -253,6 +253,18 @@ func (t *UploadHandlerTest) TestUploadWhenCreateObjectWriterFails() {
 	require.Error(t.T(), err)
 	assert.ErrorContains(t.T(), err, "createObjectWriter")
 	assert.ErrorContains(t.T(), err, "taco")
+	// Destroy should complete without deadlocking on uh.wg.Wait().
```
abhishek10004: Let's move this up next to the upload block to follow the AAA pattern?

abhishek10004 (dismissed):
LGTM. Left a minor comment

abhishek10004 (approved):

PR #4993, opened 2026-08-04
fix(storage): enforce single attempt retryer on DirectPath verification handle

vipnydav:
### Description
Disable Go SDK internal retries for the `BucketHandle` used during `verifyDirectPathConnectivity` by setting `Retryer(storage.WithMaxAttempts(1))` directly on `bucketHandle`.

Previously, `sc.SetRetry(storage.WithMaxAttempts(1))` was called after `sc.Bucket(bucketName)` had already cloned the client retry config, causing `bucketHandle` to ignore `SetRetry` and execute 15-second internal SDK retries per attempt when the backend returned `UNAVAILABLE` for cross-location DirectPath requests.

Additionally, temporarily increased expected cross-region mount timeouts to 10 seconds in `mount_timeout_test.go` with a TODO to revert after cl/958817829 rolls out to production.

### Link to the issue in case of a bug fix.
https://b.corp.google.com/issues/542608390

### Testing details
1. Manual - Verified fast fallback (~7s total) using Go reproducer script.
2. Unit tests - NA
3. Integration tests - Automated

### Any backward incompatible change? If so, please explain.
No.

PR #5027, opened 2026-08-14
feat(grpc): bypass DirectPath connectivity check for rapid buckets

anushka567:
### Description
This PR modifies the `createGRPCClientHandle` logic to bypass the `verifyDirectPathConnectivity` dummy-stat check for Rapid (and Pirlo) buckets. 
Since Rapid buckets primarily rely on gRPC for high throughput and bidirectional streaming, skipping this verification eliminates unnecessary initialization latency and potential connection failures on startup. DirectPath verification failures remain strictly fatal for standard regional buckets.

**Test Determinism**: Explicitly set `AnonymousAccess = true` in the standard test to strip environment credentials (e.g., GCE Application Default Credentials). This ensures the dummy `Attrs` call strictly fails with a `PermissionDenied` rather than authenticating and returning a `NotFound` (which the check ignores), guaranteeing deterministic failures across all CI environments.

### Testing details
1. Manual - yes.
2. Unit tests - Added.
3. Integration tests - NA

No.

### DIFF
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -209,7 +209,7 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 
 	// Set the production level retry config.
 	defer func() {
-		logger.Infof("Applying production retry config after DirectPath verification.")
+		logger.Infof("Applying production retry config.")
 		setRetryConfig(ctx, sc, clientConfig)
 	}()
 
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -187,7 +187,7 @@ func setRetryConfig(ctx context.Context, sc *storage.Client, clientConfig *stora
 }
 
 // Followed https://pkg.go.dev/cloud.google.com/go/storage#hdr-Experimental_gRPC_API to create the gRPC client.
-func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig, isbucketRapid bool, enableBidiConfig bool, bucketName string, billingProject string) (*storage.Client, error) {
+func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.StorageClientConfig, isBucketRapid bool, enableBidiConfig bool, bucketName string, billingProject string) (*storage.Client, error) {
 	if err := os.Setenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS", "true"); err != nil {
 		return nil, fmt.Errorf("error setting direct path env var: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


internal/storage/storage_handle.go
```diff
@@ -209,17 +209,16 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 
 	// Set the production level retry config.
 	defer func() {
```
meet2mky: This defer shouldn't be required right? because we never mutate the client retry policy.

anushka567: Yes, as already modified in PR #4993 , this defer is no longer needed as we never modify the global client while making the verifyDP check. Updated.

internal/storage/storage_handle.go
```diff
@@ -207,19 +207,12 @@ func createGRPCClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 		return nil, fmt.Errorf("NewGRPCClient: %w", err)
 	}
 
-	// Set the production level retry config.
-	defer func() {
```
meet2mky, line 211: Hi @anushka567 , I think my initial understanding was that we were restoring to production retry config because we mutated it but it seems that was the only place it was every being set for GRPC clients.

Let's refactor the code. 

In the getClient() method let's add this line in the end before returning client ~

```
	// Set Retry Config on all type (GRPC/HTTP) Storage Clients.
	setRetryConfig(ctx, client, &sh.clientConfig)
```

We can remove the duplicate code of setting this when creating the GRPC/HTTP client.

internal/storage/storage_handle.go
meet2mky: This should be changed like this ~

	// Add DirectPath enforcement - operations performed via the client will fail if DirectPath is not available.
	// Rapid buckets do not support DirectPath enforcement headers.
	if !isBucketRapid {
		clientOpts = append(clientOpts, experimental.WithDirectConnectivityEnforced())
	}

meet2mky: Nit: Rename variable to `isRapidBucket`.

meet2mky, line 263: Let's close the client here as it would need to be discarded to save resource.

meet2mky (changes requested):
Some comments. PTAL.

internal/storage/storage_handle.go
anushka567: The PR was for only removing the verify DP check for rapid. Removing the forcedDP will be taken up in different PR with test.

anushka567: Updated. 

Rapid buckets do not support DirectPath enforcement headers. > Here we are making conscious choice not to enforce due to server side behavior, hence rephrased a bit.

internal/storage/storage_handle.go
anushka567: In case of dynamic mount, getClient() can be invoked multiple times where we return the cached client created at time of first reference. Adding the SetRetryConfig() to getClient() means every reference will invoke this again and again which is unnecessary as it is one-time action we need to perform. Hence, moved it to client creation method.

meet2mky (approved):
LGTM

PR #5032, opened 2026-08-17
feat(otel exporter): Disabling otel exporter test for GKE & Zonal

abhishek10004:
### Description
Disabling OTEL exporter e2e tests for GKE. Also, we don't need to check for zonal buckets as well as it has nothing to do with bucket type.

b/546350504

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA


meet2mky (approved):
LGTM

PR #5036, opened 2026-08-19
feat(cfg): add explicit proto-tag and retired-params validation to params.yaml

PranjalC100:
### Description
This change introduces explicit, immutable Protobuf tags and retired parameter tracking to `params.yaml` along with comprehensive parser validation:
1. **Explicit Proto Tags**: Assigned sequential `proto-tag: <int>` ($1 \dots 158$) to all active parameters in `cfg/params.yaml`.
2. **Retired Parameters Tracking (`retired-params`)**: Added `retired-params: []` header and usage documentation in `cfg/params.yaml` to track deleted parameters for historical Protobuf telemetry decoding.
3. **Parser Validation (`tools/config-gen/parser.go`)**:
   - Added `RetiredParam` struct and `ParamsYAML.RetiredParams`.
   - Unified `supportedDataTypes` and `isValidDataType` helper across active and retired parameter validation.
   - Implemented `validateProtoTags` enforcing positive tags ($>0$), tag uniqueness, type validity, and sequence contiguity ($1 \dots \text{maxTag}$) with zero unrecorded gaps.
   - Skipped tag validation for legacy CLI-only flags with empty `config-path`.
4. **Unit Tests (`tools/config-gen/parser_test.go`)**:
   - Added comprehensive table-driven tests (`TestValidateProtoTags`) covering sequential tags, retired param gap-filling, last tag retired, duplicate tags, collisions, empty config paths, invalid data types, and gaps.
   - Updated existing mock YAML fixtures with valid `proto-tag`s.

b/548947258

### Testing details
1. **Manual**: N/A
2. **Unit tests**: `go test -race -count=1 ./tools/config-gen/... ./cfg/...`.
3. **Integration tests**: N/A

N/A

### DIFF
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -33,13 +33,11 @@
 #
 ###################################### DOCUMENTATION ENDS ######################
 
-# Add retired parameters here when deleted from active params list.
-# Preserves Protobuf schema definition for uniform historical decoding.
+# Add retired proto-tag numbers here when deleted from active params list.
+# Preserves tag reservation for Protobuf schema generation.
 retired-params: []
-# Example of a retired parameter entry:
-#   - config-path: "file-cache.enable-crc"
-#     proto-tag: 28
-#     type: "bool"
+# Example of retired parameters entry:
+#   retired-params: [28, 42]
 
 # Keep this section at the top of the file.
 machine-type-groups:
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -22,6 +22,7 @@
 # default: The default value of the param.
 # deprecated: Specifies whether the param is deprecated. This will cause warnings when the user specifies the flag.
 # deprecationWarning: The warning that should be issued when the user specifies a deprecated flag.
+# hide-flag: Whether to hide flag from the helpdoc.
 # proto-tag: Unique, immutable integer field number assigned to the parameter for Protobuf wire serialization.
 #
 # The params must be kept sorted in ascending order.
--- a/cfg/params.yaml
+++ b/cfg/params.yaml
@@ -1,4 +1,4 @@
-# Copyright 2024 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -22,7 +22,7 @@
 # default: The default value of the param.
 # deprecated: Specifies whether the param is deprecated. This will cause warnings when the user specifies the flag.
 # deprecationWarning: The warning that should be issued when the user specifies a deprecated flag.
-# hide-flag: Whether to hide flag from the helpdoc.
+# proto-tag: Unique, immutable integer field number assigned to the parameter for Protobuf wire serialization.
 #
 # The params must be kept sorted in ascending order.
 # The linear order between any two params, p1 and p2 is defined as following:
@@ -31,6 +31,14 @@
 # 3. If p1.config-path != "" && p2.config-path == "" then p1 < p2 and vice versa.
 #
 ###################################### DOCUMENTATION ENDS ######################
+
+# Add retired parameters here when deleted from active params list.
+# Preserves Protobuf schema definition for uniform historical decoding.
+retired-params: []
+# Example of a retired parameter entry:
+#   - config-path: "file-cache.enable-crc"
+#     proto-tag: 28
+#     type: "bool"
 
 # Keep this section at the top of the file.
 machine-type-groups:
@@ -64,17 +72,20 @@ machine-type-groups:
 
 params:
   - config-path: "app-name"
+    proto-tag: 1
     flag-name: "app-name"
     type: "string"
     usage: "The application name of this mount."
     default: ""
 
   - config-path: "cache-dir"
+    proto-tag: 2
     flag-name: "cache-dir"
     type: "resolvedPath"
     usage: "Enables file-caching. Specifies the directory to use for file-cache."
 
   - config-path: "cloud-profiler.allocated-heap"
+    proto-tag: 3
     flag-name: "cloud-profiler-allocated-heap"
     type: "bool"
     usage: "Enables allocated heap (HeapProfileAllocs) profiling. This only works when --enable-cloud-profiler is set to true."
@@ -82,6 +93,7 @@ params:
     hide-flag: true
 
   - config-path: "cloud-profiler.cpu"
+    proto-tag: 4
     flag-name: "cloud-profiler-cpu"
     type: "bool"
     usage: "Enables cpu profiling. This only works when --enable-cloud-profiler is set to true."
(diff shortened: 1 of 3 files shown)


tools/config-gen/parser.go
```diff
@@ -67,9 +75,89 @@ func parseParamsYAMLStr(paramsYAMLStr string) (paramsYAML ParamsYAML, err error)
...
+
+	// 3. Contiguity / gap check: ensure all numbers from 1 to maxTag are accounted for.
+	for tag := 1; tag <= maxTag; tag++ {
+		if _, isAssigned := assignedTagsMap[tag]; !isAssigned {
+			return fmt.Errorf("missing proto-tag %d in sequence 1..%d", tag, maxTag)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
PranjalC100: ACK.

tools/config-gen/parser.go
```diff
@@ -67,9 +75,91 @@ func parseParamsYAMLStr(paramsYAMLStr string) (paramsYAML ParamsYAML, err error)
...
+	}
+
+	// 3. Contiguity check: ensure all tags from 1 to maxTag are accounted for.
+	if len(assignedTagsMap) != maxTag {
+		for tag := 1; tag <= len(assignedTagsMap)+1; tag++ {
+			if _, isAssigned := assignedTagsMap[tag]; !isAssigned {
+				return fmt.Errorf("missing proto-tag %d in sequence 1..%d", tag, maxTag)
+			}
+		}
+	}
```
kislaykishore: Reduce nesting
```suggestion
	if len(assignedTagsMap) == maxTag {
	  return nil
	}
	for tag := 1; tag <= len(assignedTagsMap)+1; tag++ {
		if _, isAssigned := assignedTagsMap[tag]; !isAssigned {
			return fmt.Errorf("missing proto-tag %d in sequence 1..%d", tag, maxTag)
		}
	}
	return nil
```

cfg/params.yaml
```diff
@@ -22,7 +22,7 @@
 # default: The default value of the param.
 # deprecated: Specifies whether the param is deprecated. This will cause warnings when the user specifies the flag.
 # deprecationWarning: The warning that should be issued when the user specifies a deprecated flag.
-# hide-flag: Whether to hide flag from the helpdoc.
```
kislaykishore, line 25: Why did we remove this?

cfg/params.yaml
```diff
@@ -32,6 +32,14 @@
 #
 ###################################### DOCUMENTATION ENDS ######################
 
+# Add retired parameters here when deleted from active params list.
+# Preserves Protobuf schema definition for uniform historical decoding.
+retired-params: []
+# Example of a retired parameter entry:
+#   - config-path: "file-cache.enable-crc"
+#     proto-tag: 28
```
kislaykishore: why not just specify the retired-params as something like:
`retired-params: [28, 42]?`

tools/config-gen/parser_test.go
```diff
@@ -325,6 +330,194 @@ params:
...
+				{FlagName: "p1", ConfigPath: "p1", ProtoTag: 1},
+				{FlagName: "p3", ConfigPath: "p3", ProtoTag: 3},
+			},
+			retiredParams: []RetiredParam{
+				{ConfigPath: "p2", ProtoTag: 2, Type: "bool"},
+			},
+			expectErr: false,
+		},
+		{
+			name: "Valid_with_last_tag_retired",
```
kislaykishore: Please start the names with lowercase. This is to keep invoking tests in the CLI simpler.

cfg/params.yaml
PranjalC100: We need to keep the `config-path` and `type` so the **deserializer** can remain **version-independent** and properly parse older configs without classifying deleted flags as unknown fields.
As discussed, if we only store the tag numbers, the deserializer won't know the original flag names or data types for the retired tags. 
Otherwise, we'd have to rely on the existing `config.proto` to keep this information, which means `params.yaml` would no longer act as the single source of truth.

cfg/params.yaml
PranjalC100: restored.

cfg/params.yaml
kislaykishore: But protobuf automatically handles reserved tags and ignores them safely. So, why can't the deserializer be version independent.

tools/config-gen/parser_test.go
PranjalC100: Done! I've updated the subtest names to start with a lowercase letter for easier CLI invocation.

cfg/params.yaml
PranjalC100: Protobuf ignores `reserved` tags during binary deserialization, but GCSFuse config deserialization operates on YAML files and CLI flags via `config-path`.

Why we can't just use `[28, 42]` and `reserved` tags:
* **No `reserved` in `config.proto`**: To keep the deserializer truly version-independent, we will retain all field definitions in `config.proto` rather than using `reserved`. Otherwise, deserializing or debugging configs/telemetry from an older release (e.g., `3.13.0` after a flag is retired in `3.14.0`) would fail to resolve tag names and types.
* **YAML Backwards Compatibility**: If `params.yaml` only stored tag numbers, the parser wouldn't know the `config-path` or datatype for retired tags. Running a newer GCSFuse release with an older `config.yaml` would fail with an unknown field error instead of safely parsing and ignoring it.
* **Single Source of Truth (Option A)**: As agreed in our chat (Option A vs. Option B), retaining `config-path` and `type` in `retired-params` keeps `params.yaml` self-contained. It ensures deterministic regeneration of `config.proto` without relying on historical file state.

tools/config-gen/parser.go
PranjalC100: ACK.

cfg/params.yaml
kislaykishore: What's the issue with failing to resolve config from an old release? It will just ignore those reserved fields.

PranjalC100: I think we may loose visibility when debugging customer issues/adoption across versions.

For example, suppose a customer runs `v3.13.0` with `file-cache.enable-crc` (tag `28`). If we retire that flag in `v3.14.0` and only emit `reserved 28;` in `config.proto`, any tooling or on-call engineer inspecting that customer's `v3.13.0` telemetry against our master decoder will silently drop tag `28`. I completely lose visibility into what configuration options the customer actually had enabled.

Keeping `config-path` and `type` in `params.yaml` keeps it self-sufficient and allows our uniform decoder on master to inspect telemetry from older releases without blind spots.

Retaining retired fields in `config.proto` adds zero runtime or network overhead—unset fields consume 0 bytes on the wire. Most importantly, it ensures we can deterministically regenerate `config.proto` from `params.yaml` without depending on schemas from previous releases or git history.

kislaykishore: I don't see it as a big plus TBH. If we remove a flag, we know it's not very useful and there is no usage of that flag.

PranjalC100: That's true of course. So should I then just store the retired tag numbers in the `params.yaml`?

kislaykishore: Yes

PranjalC100: ACK. Pushed the changes in the next commit.

kislaykishore (approved):

PR #5040, opened 2026-08-21
test: move simulated clock advancement into eventually blocks within

saisagar1998:
### Description
In `InactiveTimeoutReaderTestSuite`, `Test_Read_ReconnectFails` and `Test_Read_TimeoutAndSuccessfulReconnect` rely on simulated clock advancements to trigger inactivity timeouts. Moving the `AdvanceTime` call inside the `require.Eventually` polling block ensures that if the background goroutine hasn't scheduled/processed the timer yet, advancing time during polling retries guarantees the timeout condition is triggered and checked reliably without timing flakiness.

b/546433355

### Testing details
1. Manual - NA
2. Unit tests - Ran TestInactiveTimeoutReaderTestSuite successfully.
3. Integration tests - NA
NA

### DIFF
--- a/internal/gcsx/inactive_timeout_reader.go
+++ b/internal/gcsx/inactive_timeout_reader.go
@@ -51,6 +51,9 @@ type InactiveTimeoutReader struct {
 	object *gcs.MinObject
 	bucket gcs.Bucket
 
+	timeout time.Duration
+	clock   clock.Clock
+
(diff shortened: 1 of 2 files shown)


charith87:
@vipnydav could you take a first look since its regarding inactive streams.

abhishek10004 (approved):
LGTM

charith87 (approved):

PR #5042, opened 2026-08-24
docs: add known issue for DirectPath verification hang

vipnydav:
### Description
Adds a known issue for GCSFuse versions 3.10.x and 3.11.x documenting a 1.5-minute hang during DirectPath connectivity verification due to an internal SDK retry bug when a retryable error is encountered.

b/542608390

### Testing details
1. Manual - NA
2. Unit tests - NA
3. Integration tests - NA

None.


docs/known-issues.md
```diff
@@ -22,3 +22,4 @@ This document lists known issues and bugs in GCSFuse, their impact, and the rele
...
+| DirectPath connectivity verification can hang for up to ~1.5 minutes because of an internal SDK retry bug when a retryable error is encountered.                                                                                                                                                                                                                                                                                                                                                                                                        | v3.10.0, v3.11.0, v3.11.1, v3.11.2 | v3.12.0                                | [#4993](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4993)                                                                     |
```
abhishek10004: should we add that this can only happen during mount time (& is a one time thing) so that users don't confuse this being the reason for hangs etc. which might be due to other reasons

abhishek10004 (approved):

PR #5043, opened 2026-08-24
fix(gcsx): prioritize context cancellation over concurrent operation results in multi-range downloader

saisagar1998:
### Description
When a read context is cancelled, Go's `select` statement between `<-ctx.Done()` and `<-done` chooses pseudo-randomly if both channels are ready simultaneously. In cases where the background download operation completes before or concurrently with cancellation evaluation, receiving from `done` would previously return the read result or EOF instead of respecting context cancellation when interrupts are enabled.

This change checks `ctx.Err()` when handling results from the `done` channel in `MrdInstance.Read` and `MultiRangeDownloaderWrapper.Read` to ensure context cancellation is prioritized when interrupts are enabled. It also updates the unit tests in `multi_range_downloader_wrapper_test.go` to test both zero and non-zero sleep cases.

b/546318093

### Testing details
1. Unit tests - `go test -v -count=1 ./internal/gcsx -run TestMRDWrapperTestSuite`

NA

### DIFF
--- a/internal/gcsx/multi_range_downloader_wrapper_test.go
+++ b/internal/gcsx/multi_range_downloader_wrapper_test.go
@@ -231,7 +231,7 @@ func (t *mrdWrapperTest) TestReadContextCancelledWithInterruptsDisabled() {
 			bytesRead, err := t.mrdWrapper.Read(ctx, make([]byte, t.object.Size), 0, int64(t.object.Size), metrics.NewNoopMetrics(), tracing.NewNoopTracer(), false)
 
 			require.NoError(t.T(), err)
-			assert.Equal(t.T(), 100, bytesRead)
+			assert.Equal(t.T(), int(t.object.Size), bytesRead)
 		})
 	}
 }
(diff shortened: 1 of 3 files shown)


charith87:
Adding @abhishek10004 since its regarding MRD

internal/gcsx/multi_range_downloader_wrapper_test.go
```diff
@@ -174,29 +174,66 @@ func (t *mrdWrapperTest) Test_Read_ShortRead() {
...
+			ctx, cancel := context.WithCancel(context.Background())
+			cancel()
 
-	bytesRead, err := t.mrdWrapper.Read(ctx, make([]byte, t.object.Size), 0, int64(t.object.Size), metrics.NewNoopMetrics(), tracing.NewNoopTracer(), false)
+			bytesRead, err := t.mrdWrapper.Read(ctx, make([]byte, t.object.Size), 0, int64(t.object.Size), metrics.NewNoopMetrics(), tracing.NewNoopTracer(), false)
 
-	require.NoError(t.T(), err)
-	assert.Equal(t.T(), 100, bytesRead)
+			require.NoError(t.T(), err)
+			assert.Equal(t.T(), 100, bytesRead)
```
abhishek10004: let's use t.object.Size instead of hardcoded value

abhishek10004 (dismissed):
LGTM. Left a minor comment.

internal/gcsx/multi_range_downloader_wrapper_test.go
saisagar1998: done

abhishek10004 (approved):
LGTM

PR #5048, opened 2026-08-25
test(cache): remove kernel-dependent unaligned O_DIRECT test case

kislaykishore:
### Description
- Removed the kernel-dependent `writer offset not multiple of 4096` subtest from `Test_CopyUsingMemoryAlignedBuffer` in `internal/cache/util/util_test.go`. On modern Linux kernels (Linux 6.1+, including Ubuntu 24.04), `O_DIRECT` writes at unaligned offsets are supported (or handled via buffered fallback) by the kernel and do not fail with `EINVAL`, causing the test to fail.
- Updated GitHub workflows (`ci.yml` and `flake-detector.yml`) to run on `ubuntu-latest` instead of pinning to `ubuntu-22.04`.

Fixes b/496763741

### Testing details
1. Manual - Ran `make build` and verified clean compilation, formatting, and linting.
2. Unit tests - Ran `go test -v -race -count=1 ./internal/cache/...` with all tests passing.
3. Integration tests - NA

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
 
   linux-tests:
     needs: filter
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 25
 
     steps:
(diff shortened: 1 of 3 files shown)


abhishek10004 (approved):
LGTM

PR #5052, opened 2026-08-26
fix: backport cherry-picks and remediate dependency CVEs on v3.11.3

### LINKED ISSUE #5018: Windowed listings can poison the stat cache and cause false `ENOENT` for existing implicit directories

kislaykishore:
### Description

This PR backports critical fixes and upgrades vulnerable Go dependencies for the `v3.11.3_release` branch:

1. **Backport PR #5025** (cherry-pick commit `26e67a0ed7b31e5a1b5ec0671507a8be62a9fddb` / upstream `0ce7ebc30e6729912422b08315b0f86dd52ef102`):
   - `fix(caching)`: only cache negative entries from authoritative listings.
   - Prevents empty windowed or paginated object listings from creating a negative stat-cache entry for an entire implicit-directory prefix, preventing transient `ENOENT` errors.

2. **Backport PR #4993** (cherry-pick commit `ad2543347280d0496024e59453f70c337665e82c` / upstream `2b2f0573ecaefc67bd188799766f3cadf3bc76c6`):
   - `fix(storage)`: enforce single attempt retryer on DirectPath verification handle.
   - Sets `Retryer(storage.WithMaxAttempts(1))` directly on `bucketHandle` to avoid 15-second internal SDK retries per attempt when the backend returns `UNAVAILABLE`.

3. **Backport PR #5027** (cherry-pick commit `450646ca2399f6875b0cf1bf5768424b93dde72e` / upstream `7b39292788a42c304802366ec3b822f6cf634ce1`):
   - `feat(grpc)`: bypass DirectPath connectivity check for rapid buckets.
   - Skips DirectPath check when mounting rapid/pirlo buckets while preserving checks for standard buckets.

4. **Dependency CVE Vulnerability Remediation** (commit `9e34ef5e25a7280eb3d147b7119697946517c84b`):
   - Upgraded direct and transitive Go dependencies to remediate vulnerabilities reported by `govulncheck`:
     - `google.golang.org/grpc`: updated to safe version v1.82.1.
     - `golang.org/x/text`: updated to safe version v0.39.0.
     - `golang.org/x/net`: updated to safe version v0.56.0.
     - `golang.org/x/crypto`: updated to safe transitive version v0.53.0 via `go mod tidy`.
     - Updated Go toolchain to Go 1.26.6 in `go.mod` and `.go-version`.

- Fixes #5018 (PR #5025)
- https://b.corp.google.com/issues/542608390 (PR #4993)
- Rapid bucket DirectPath check bypass (PR #5027)

### Testing details

1. **Manual**:
   - Verified clean compilation with `go build ./...`.
   - Audited vulnerability status with `govulncheck ./...` showing 0 reachable vulnerabilities.
2. **Unit tests**:
   - Executed repository standard test suite per `Makefile:67-68`:
     `CGO_ENABLED=0 go test -timeout 5m -count 1 $(go list ./... | grep -v "tools/integration_tests\|internal/cache/...") && CGO_ENABLED=0 go test -timeout 5m -p 1 -count 1 ./internal/cache/...`
   - Verified `internal/cache/...` passes cleanly with `-p 1 -count 1` (`CGO_ENABLED=0 go test -timeout 5m -p 1 -count 1 ./internal/cache/...`), resolving parallel test interference in `internal/cache/file/downloader` caused by shared cache directory access.
   - `go test -count 1 ./internal/storage/...` passed.
   - `go test -count 1 ./internal/storage/caching/...` passed.
3. **Integration tests**:
   - Verified that regression test `TestIntegration_WindowedListingDoesNotPoisonExistingDirectory` in `internal/storage/caching` passes.
   - Verified that `tools/integration_tests/mount_timeout` compiles cleanly via `go test -c ./tools/integration_tests/mount_timeout`.

None. All changes preserve existing external APIs, CLI flags, and mount configurations.


kislaykishore:
Replacing with PR from designated branch v3.11.3-cherrypicks-and-cve-fixes

kislaykishore:
Superceded by PR from designated branch v3.11.3-cherrypicks-and-cve-fixes

PR #5061, opened 2026-08-28
test(cache): fix race condition in parallel download mock expectation

kislaykishore:
### Description
The mock expectation for chunk 4 explicitly mandated `req.ReadHandle != nil`. Because worker goroutines race concurrently to acquire chunks from the job queue, any worker (including one that has not yet opened a read handle) may pick up chunk 4. Under heavy scheduler contention or race detector instrumentation, this non-deterministic worker execution led to intermittent mock expectation mismatches.

**Solution / Fix Mechanism:**
- Relaxed the chunk 4 mock expectation to accept either nil or non-nil ReadHandle.

b/553889964

### Testing details
1. Manual - Verified compilation and formatting via `make build`.
2. Unit tests - Executed with race detector: `go test -v -race ./internal/cache/file/downloader/...` (PASS — 0 race warnings, 0 failures).
3. Integration tests - N/A (Unit test mock fix).

N/A

### DIFF
--- a/internal/cache/file/downloader/parallel_downloads_job_testify_test.go
+++ b/internal/cache/file/downloader/parallel_downloads_job_testify_test.go
@@ -15,13 +15,13 @@
 package downloader
 
 import (
+	"bytes"
+	"context"
 	"io"
 	"os"
 	"strings"
 	"sync"
 	"testing"
-
-	"context"
 
 	"github.com/googlecloudplatform/gcsfuse/v3/cfg"
 	"github.com/googlecloudplatform/gcsfuse/v3/internal/cache/data"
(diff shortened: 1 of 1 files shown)


internal/cache/file/downloader/parallel_downloads_job_testify_test.go
```diff
@@ -128,9 +128,9 @@ func (t *ParallelDownloaderJobTestifyTest) Test_ParallelDownloadObjectToFile_New
 		return req.Range.Start == rangeR3.Start && req.Range.Limit == rangeR3.Limit
 	})).Run(incrementCounters).Return(readerR3, nil).Once()
 
-	// Chunk 4 (R4): ReadHandle should not be nil
+	// Chunk 4 (R4): ReadHandle can be nil or propagated depending on which worker picks it up. Match primarily on range.
 	t.mockBucket.On("NewReaderWithReadHandle", mock.Anything, mock.MatchedBy(func(req *gcs.ReadObjectRequest) bool {
-		return req.Range.Start == rangeR4.Start && req.Range.Limit == rangeR4.Limit && req.ReadHandle != nil
```
charith87, line 133: Will this not check the case when a Readhandle to be re-used. We are relaxing this now and might miss this case.

kislaykishore: Added another test to validate the ReadHandle reuse.

charith87 (approved):

PR #5066, opened 2026-08-31
refactor(integration_tests): generalize WaitForSizeUpdate and rename wait constants to rapid

vipnydav:
### **Description**
1. Generalized [`WaitForSizeUpdate`](file:///usr/local/google/home/vipinydv/gcsfuse/tools/integration_tests/util/operations/operations.go#L83) parameter name from `isZonal bool` to `isUnfinalized bool` to represent unfinalized objects for both Zonal and Pirlo runs.
2. Renamed wait duration constants `WaitDurationAfterFlushZB` $\rightarrow$ `WaitDurationAfterFlushRapid` and `WaitDurationAfterCloseZB` $\rightarrow$ `WaitDurationAfterCloseRapid` in [`file_operations.go`](file:///usr/local/google/home/vipinydv/gcsfuse/tools/integration_tests/util/operations/file_operations.go#L56).
3. In `file_operations.go`, updated `CloseFiles`, `CloseFile`, and `CloseFileShouldNotThrowError` to pass `setup.IsZonalBucketRun()` since only Zonal leaves files unfinalized on close (Pirlo finalizes on close by default and does not need to sleep 1s).
4. In `storage_client.go` and `symlink_suites_test.go`, passed `wc.Append && !wc.FinalizeOnClose` to dynamically wait only when an unfinalized appendable object was actually created.
5. Updated `SyncFile` in `file_operations.go` and `reads_after_appends_test.go` to pass `setup.IsZonalBucketRun() || setup.IsPirloBucketRun()` for unfinalized flushes/appends.

### **Testing details**
1. Manual - Done
2. Unit tests - NA

### **Any backward incompatible change? If so, please explain.**
NA

### DIFF
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -292,7 +292,7 @@ func CreateObjectWithOptions(ctx context.Context, client *storage.Client, object
 	if err := wc.Close(); err != nil {
 		return fmt.Errorf("wc.Close failed for object %q: %w", object, err)
 	}
-	operations.WaitForSizeUpdate(setup.IsZonalBucketRun(), operations.WaitDurationAfterCloseZB)
+	operations.WaitForSizeUpdate(wc.Append && !wc.FinalizeOnClose, operations.WaitDurationAfterCloseRapid)
 	return nil
 }
 
@@ -406,7 +406,7 @@ func UploadGcsObjectWithPreconditions(ctx context.Context, client *storage.Clien
 		if err := w.Close(); err != nil {
 			log.Printf("Failed to close GCS object gs://%s/%s: %v", bucketName, objectName, err)
 		}
-		operations.WaitForSizeUpdate(setup.IsZonalBucketRun(), operations.WaitDurationAfterCloseZB)
+		operations.WaitForSizeUpdate(w.Append && !w.FinalizeOnClose, operations.WaitDurationAfterCloseRapid)
 	}()
 
 	filePathToUpload := localPath
--- a/tools/integration_tests/util/client/storage_client.go
+++ b/tools/integration_tests/util/client/storage_client.go
@@ -292,7 +292,8 @@ func CreateObjectWithOptions(ctx context.Context, client *storage.Client, object
 	if err := wc.Close(); err != nil {
 		return fmt.Errorf("wc.Close failed for object %q: %w", object, err)
 	}
-	operations.WaitForSizeUpdate(wc.Append && !wc.FinalizeOnClose, operations.WaitDurationAfterCloseRapid)
+	isUnfinalizedObject := wc.Append && !wc.FinalizeOnClose
+	operations.WaitForSizeUpdate(isUnfinalizedObject, operations.WaitDurationAfterCloseRapid)
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


tools/integration_tests/util/operations/operations.go
```diff
@@ -80,9 +81,15 @@ func ExecuteGcloudCommand(command string) ([]byte, error) {
 }
 
 // WaitForSizeUpdate waits for a specified time duration to ensure that stat()
-// call returns correct size for unfinalized object.
-func WaitForSizeUpdate(isZonal bool, duration time.Duration) {
-	if isZonal {
+// call returns correct size for unfinalized object in zonal or pirlo bucket runs.
```
abhishek10004: Would there be tests in pirlo where we'd like to test stat instantly after write (when finalize-file-for-rapid is true)? Have you made sure that they're not calling this or how are we handling them?

vipnydav: Thanks for catching this!

I checked across `test_config.yaml` and noticed that for Pirlo, we only run with unfinalized objects on close (`--finalize-file-for-rapid=false`) in `rapid_operations`. In all other suites running on Pirlo, files are finalized on close by default.

I have updated the PR accordingly:
1. In `file_operations.go`, `CloseFile*` methods now only pass `setup.IsZonalBucketRun()`. This ensures that for all normal Pirlo runs, we don't introduce an unnecessary 1-second sleep on close, and tests asserting `stat` after close run with zero delay.
2. In `storage_client.go` and `symlink_suites_test.go`, we now check `wc.Append && !wc.FinalizeOnClose` to dynamically wait only when an unfinalized object was actually created.
3. `SyncFile` and `reads_after_appends_test.go` continue to wait for both Zonal and Pirlo (`setup.IsZonalBucketRun() || setup.IsPirloBucketRun()`) since object is still unfinalized.

tools/integration_tests/util/operations/file_operations.go
```diff
@@ -174,15 +174,15 @@ func CloseFiles(t *testing.T, files []*os.File) {
 		err := file.Close()
 		assert.NoError(t, err)
 	}
-	WaitForSizeUpdate(setup.IsZonalBucketRun(), WaitDurationAfterCloseZB)
+	WaitForSizeUpdate(setup.IsZonalBucketRun(), WaitDurationAfterCloseRapid)
```
abhishek10004, line 179: why are we not using "setup.IsZonalBucketRun() || setup.IsPirloBucketRun()" here?

tools/integration_tests/util/client/storage_client.go
```diff
@@ -292,7 +292,7 @@ func CreateObjectWithOptions(ctx context.Context, client *storage.Client, object
 	if err := wc.Close(); err != nil {
 		return fmt.Errorf("wc.Close failed for object %q: %w", object, err)
 	}
-	operations.WaitForSizeUpdate(setup.IsZonalBucketRun(), operations.WaitDurationAfterCloseZB)
+	operations.WaitForSizeUpdate(wc.Append && !wc.FinalizeOnClose, operations.WaitDurationAfterCloseRapid)
```
abhishek10004: it's fine to be more verbose in tests for better readability. Think if creating a separate properly named variable would be better here

tools/integration_tests/util/operations/file_operations.go
vipnydav: Stale metadata is only returned for unfinalized objects. Since we are closing the writer here rather than just syncing it, the object will be finalized (for pirlo, 'finalize on close' is the default behavior). Because the object is finalized, it will immediately return the correct size, so we don't need to wait. We only need to handle unfinalized pirlo objects in the `rapid_operations` package, and this method isn't used there.

tools/integration_tests/util/client/storage_client.go
vipnydav: Agreed.

tools/integration_tests/util/operations/file_operations.go
abhishek10004: maybe add a comment to change the condition (to add wait for pirlo) if this is used in future for pirlo. Same for CloseFileShouldNotThrowError

abhishek10004 (dismissed):
LGTM. Left a minor comment.

tools/integration_tests/util/operations/file_operations.go
vipnydav: Added.

abhishek10004 (approved):
LGTM

PR #5069, opened 2026-08-31
feat: add direct S2A authentication support for HTTP transport

sruthi-talluri:
### Description
This PR adds Direct S2A (Secure Session Agent) authentication support to GCSFuse over HTTP (client-protocol: `http1` and `http2`) and gRPC.

Direct S2A authentication is enabled whenever the `--s2a-address` CLI flag (or `s2a-address` configuration field) is specified:
- Over HTTP/1.1 and HTTP/2: Uses `s2a.NewS2ADialTLSContextFunc` with optional local SPIFFE ID identity (`--s2a-spiffe-id`) to establish mTLS transport channels directly with Google Cloud Storage endpoints.
- Over gRPC: Configures gRPC client dial options with S2A channel credentials (`s2a.NewClientCreds`) for both storage and control client channels.
- When `--s2a-address` is provided, standard OAuth2 token fetching is bypassed (`option.WithoutAuthentication()`), relying entirely on S2A mTLS handshake credentials.
- When `--s2a-address` is omitted or empty, standard credential resolution paths (OAuth2, ADC, etc.) are preserved unchanged.

### Verification & Testing

1. Unit Tests
All unit tests in `cfg`, `cmd`, and `internal/storage` pass cleanly (100% OK):
- Verified flag parsing and config mappings in `cfg` and `cmd`.
- Added unit tests for enabled, disabled, and fallback S2A configurations in `internal/storage/storage_handle_test.go` and `internal/storage/storageutil/client_test.go`.

2. Live Staging Host uVM End-to-End CUJ Test
Verified in the Agent Compute Engine (ACE) staging environment against host pool `gcsfuse-s2a-integration` (`gcsfuse-s2a-integration-0` in `us-central1-staginga`) running the E2E shared storage test suite (`UvmSharedStorageE2eIntegrationTest`):
- Test Invocations: [Sponge Link 1](http://sponge2/2ebf36dd-9f60-457a-8c35-558baabbc255) | [Sponge Link 2](http://sponge2/8dbc39c7-2a5c-4af7-8a3a-56f6717d4f7f) | [Latest Invocations](http://sponge2/fcd8f300-432b-4add-a0b8-6f600694e502)
- Status: PASSED (OK)
- Flow Verified:
  - Primary and isolated uVMSpaces created.
  - Writer uVM launched with delegated SPIFFE identity, mounted GCS via Direct S2A GCSFuse binary, and wrote test file `/shared/test_shared_<nonce>.txt`.
  - Reader uVM in the same space mounted storage via GCSFuse + S2A, read `/shared/test_shared_<nonce>.txt`, and verified cross-uVM consistency.
  - Isolated uVM in a distinct space verified file isolation (`FILE_NOT_FOUND`).

3. Direct S2A Mount Protocol Verification
Verified direct S2A mounting on host instance `gcsfuse-s2a-integration-0` with local S2A daemon (169.254.169.254:8083):
- HTTP/1.1 (`--client-protocol=http1`): Initiated S2A mTLS handshake with Google Cloud Storage (`GetStorageLayout` returned in 79ms), mounted bucket, and unmounted cleanly.
- HTTP/2 (`--client-protocol=http2`): Initiated S2A mTLS handshake with Google Cloud Storage (`GetStorageLayout` returned in 54ms), mounted bucket, and unmounted cleanly.

### Link to Issue / Bug
Bug: b/550402283

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -648,6 +648,10 @@ type GcsAuthConfig struct {
 
 	ReuseTokenFromUrl bool `yaml:"reuse-token-from-url"`
 
+	S2aAddress string `yaml:"s2a-address"`
+
+	S2aSpiffeId string `yaml:"s2a-spiffe-id"`
+
 	TokenUrl string `yaml:"token-url"`
 }
 
@@ -1461,6 +1465,10 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 
 	flagSet.BoolP("reuse-token-from-url", "", true, "If false, the token acquired from token-url is not reused.")
 
+	flagSet.StringP("s2a-address", "", "", "S2A daemon address for direct S2A authentication.")
+
+	flagSet.StringP("s2a-spiffe-id", "", "", "Local SPIFFE ID to assume for direct S2A authentication.")
+
 	flagSet.IntP("sequential-read-size-mb", "", 200, "File chunk size to read from GCS in one call. Need to specify the value in MB. ChunkSize less than 1MB is not supported")
 
 	flagSet.DurationP("stackdriver-export-interval", "", 0*time.Nanosecond, "Export metrics to stackdriver with this interval. A value of 0 indicates no exporting.")
@@ -2101,6 +2109,14 @@ func BindFlags(v *viper.Viper, flagSet *pflag.FlagSet) error {
 	}
 
 	if err := v.BindPFlag("gcs-auth.reuse-token-from-url", flagSet.Lookup("reuse-token-from-url")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("gcs-auth.s2a-address", flagSet.Lookup("s2a-address")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("gcs-auth.s2a-spiffe-id", flagSet.Lookup("s2a-spiffe-id")); err != nil {
 		return err
 	}
 
--- a/cmd/legacy_main.go
+++ b/cmd/legacy_main.go
@@ -174,6 +174,8 @@ func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle m
 		TracingEnabled:             cfg.IsTracingEnabled(newConfig),
 		EnableHTTPDNSCache:         newConfig.GcsConnection.EnableHttpDnsCache,
 		LocalSocketAddress:         newConfig.GcsConnection.ExperimentalLocalSocketAddress,
+		S2AAddress:                 newConfig.GcsAuth.S2aAddress,
+		S2ASpiffeID:                newConfig.GcsAuth.S2aSpiffeId,
 		EnableGrpcMetrics:          newConfig.Metrics.ExperimentalEnableGrpcMetrics,
 		IsGKE:                      isGKE,
 		WriteConfig:                &newConfig.Write,
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -110,6 +110,10 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
 		creds, err := s2a.NewClientCreds(&s2a.ClientOptions{
 			S2AAddress:    clientConfig.S2AAddress,
 			LocalIdentity: localIdentity,
+			// GCS is a Google endpoint serving a WebPKI certificate, so S2A must
+			// validate the peer chain against Google roots rather than treating it
+			// as a SPIFFE SVID. See the equivalent comment in storageutil.
+			VerificationMode: s2a.ConnectToGoogle,
 		})
 		if err != nil {
 			return nil, fmt.Errorf("failed to create S2A gRPC credentials: %w", err)
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -29,6 +29,7 @@ import (
 	control "cloud.google.com/go/storage/control/apiv2"
 	"cloud.google.com/go/storage/control/apiv2/controlpb"
 	"cloud.google.com/go/storage/experimental"
+	"github.com/google/s2a-go"
 	"github.com/google/uuid"
 	"github.com/googleapis/gax-go/v2"
 	"github.com/googlecloudplatform/gcsfuse/v3/cfg"
@@ -101,7 +102,21 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
 	}
 
 	// Configure authentication.
-	if clientConfig.AnonymousAccess {
+	if clientConfig.S2AAddress != "" {
+		var localIdentity s2a.Identity
+		if clientConfig.S2ASpiffeID != "" {
+			localIdentity = s2a.NewSpiffeID(clientConfig.S2ASpiffeID)
+		}
+		creds, err := s2a.NewClientCreds(&s2a.ClientOptions{
+			S2AAddress:    clientConfig.S2AAddress,
+			LocalIdentity: localIdentity,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create S2A gRPC credentials: %w", err)
+		}
+		clientOpts = append(clientOpts, option.WithGRPCDialOption(grpc.WithTransportCredentials(creds)))
+		clientOpts = append(clientOpts, option.WithoutAuthentication())
+	} else if clientConfig.AnonymousAccess {
 		clientOpts = append(clientOpts, option.WithoutAuthentication())
 	} else if clientConfig.EnableGoogleLibAuth {
 		var authOpts []option.ClientOption
@@ -262,7 +277,7 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 	var clientOpts []option.ClientOption
 	var tokenSrc oauth2.TokenSource = nil
 
-	if clientConfig.AnonymousAccess {
+	if clientConfig.S2AAddress != "" || clientConfig.AnonymousAccess {
 		clientOpts = append(clientOpts, option.WithoutAuthentication())
 	} else if clientConfig.EnableGoogleLibAuth {
 		var authOpts []option.ClientOption
--- a/internal/storage/storage_handle.go
+++ b/internal/storage/storage_handle.go
@@ -115,6 +115,9 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
 			return nil, fmt.Errorf("failed to create S2A gRPC credentials: %w", err)
 		}
 		clientOpts = append(clientOpts, option.WithGRPCDialOption(grpc.WithTransportCredentials(creds)))
+		// When S2A is enabled, authentication is handled via mTLS at the transport level.
+		// Passing WithoutAuthentication() bypasses the default OAuth2 credential flow via Metadata Server.
+		// If not, every request will have an OAuth token attached and it will get used instead of S2A.
 		clientOpts = append(clientOpts, option.WithoutAuthentication())
 	} else if clientConfig.AnonymousAccess {
 		clientOpts = append(clientOpts, option.WithoutAuthentication())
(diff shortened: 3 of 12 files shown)


cmd/legacy_main.go
```diff
@@ -174,11 +174,16 @@ func createStorageHandle(newConfig *cfg.Config, userAgent string, metricHandle m
...
 		EnableHTTPDNSCache:         newConfig.GcsConnection.EnableHttpDnsCache,
 		LocalSocketAddress:         newConfig.GcsConnection.ExperimentalLocalSocketAddress,
+		S2AAddress:                 newConfig.GcsAuth.S2aAddress,
+		S2ASpiffeID:                newConfig.GcsAuth.S2aSpiffeId,
 		EnableGrpcMetrics:          newConfig.Metrics.ExperimentalEnableGrpcMetrics,
 		IsGKE:                      isGKE,
 		WriteConfig:                &newConfig.Write,
 	}
 	logger.Infof("UserAgent = %s\n", storageClientConfig.UserAgent)
+	if storageClientConfig.S2AAddress != "" {
```
vadlakondaswetha: Entire config is already logged by GCSFuse, we need not log individual parameters again.

cfg/config.go
```diff
@@ -643,10 +643,16 @@ type FileSystemConfig struct {
 type GcsAuthConfig struct {
 	AnonymousAccess bool `yaml:"anonymous-access"`
 
+	EnableS2a bool `yaml:"enable-s2a"`
```
vadlakondaswetha: when the user sets s2aAdress we trigger S2a flow. why do you think we need another flag to control the s2a authentication?

internal/storage/storageutil/client.go
```diff
@@ -148,6 +165,23 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
...
 			Timeout: storageClientConfig.HttpClientTimeout,
 		}
+	} else if storageClientConfig.EnableS2A && storageClientConfig.S2AAddress != "" {
+		// When S2A is enabled, authentication is handled via mTLS at the transport level.
+		httpClient = &http.Client{
+			Transport: transport,
+			Timeout:   storageClientConfig.HttpClientTimeout,
+		}
+		// Setting UserAgent through RoundTripper middleware
+		httpClient.Transport = &userAgentRoundTripper{
```
vadlakondaswetha: to avoid duplicating this logic can we do something like this:

if storageClientCofnig.AnonymounsAccess {
   create client
   return
} else if S2A flow {
   create http client
} else {
  create http client
}

add useragent
add tracing support
return

internal/storage/storage_handle.go
```diff
@@ -275,7 +290,9 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 	var clientOpts []option.ClientOption
 	var tokenSrc oauth2.TokenSource = nil
 
-	if clientConfig.AnonymousAccess {
+	if clientConfig.EnableS2A && clientConfig.S2AAddress != "" {
+		clientOpts = append(clientOpts, option.WithoutAuthentication())
+	} else if clientConfig.AnonymousAccess {
```
vadlakondaswetha: can we combine the s2a check also into this if clause

cfg/config.go
sruthi-talluri: The ACE team (Zixuan Wang / Theodore Preduta) requested an explicit boolean flag (--enable-s2a) to mirror how Envoy guards S2A integration (google3/net/envoy/source/extensions/transport_sockets/s2a). Because the S2A daemon infrastructure is currently deployed in TC staging and not yet in production.

internal/storage/storageutil/client.go
sruthi-talluri: Updated CreateHttpClient as suggested: AnonymousAccess returns early, the base transport is selected based on S2A vs OAuth2, and UserAgent / Tracing wrappers are applied once at the end.

internal/storage/storage_handle.go
sruthi-talluri: Combined the S2A and AnonymousAccess checks into a single if clause.

internal/storage/storageutil/client.go
```diff
@@ -145,41 +161,48 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
...
-		httpClient = &http.Client{
+		return &http.Client{
 			Timeout: storageClientConfig.HttpClientTimeout,
-		}
+		}, nil
+	}
+
+	var clientTransport http.RoundTripper = transport
+	if storageClientConfig.S2AAddress != "" {
+		// When S2A is enabled, authentication is handled via mTLS at the transport level.
```
vadlakondaswetha: why did you create a if block which has no logic. can you do this check?
if storageClientConfig.S2AAddress == "" {
}

internal/storage/storageutil/client.go
```diff
@@ -145,41 +161,48 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
 		// While the "WithUserAgent" option could set a custom User-Agent, it's incompatible
 		// with the "WithHTTPClient" option, preventing the direct injection of a user agent
 		// when authentication is skipped.
-		httpClient = &http.Client{
+		return &http.Client{
```
vadlakondaswetha: why are we returning earlier for anonymous access?

internal/storage/storage_handle.go
```diff
@@ -101,7 +102,21 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
...
+		}
+		creds, err := s2a.NewClientCreds(&s2a.ClientOptions{
+			S2AAddress:    clientConfig.S2AAddress,
+			LocalIdentity: localIdentity,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create S2A gRPC credentials: %w", err)
+		}
+		clientOpts = append(clientOpts, option.WithGRPCDialOption(grpc.WithTransportCredentials(creds)))
+		clientOpts = append(clientOpts, option.WithoutAuthentication())
```
vadlakondaswetha, line 125: please add a comment why we are doing withoutAuthentication for S2A

internal/storage/storageutil/client.go
sruthi-talluri: Done, I have refactored this block entirely. The empty check is removed, and the logic is now combined with the AnonymousAccess flow so that we do not skip the common setup steps

internal/storage/storageutil/client.go
sruthi-talluri: Done, returning early was causing the flow to skip the User-Agent and tracing wrappers. I have refactored the logic so that both AnonymousAccess and S2A now flow through the bottom block, ensuring all the middleware is applied uniformly.

internal/storage/storage_handle.go
sruthi-talluri: Done. I added a comment explaining that mTLS authentication is already handled at the transport level by S2A, so we use WithoutAuthentication to bypass the default OAuth2 credential flow.

internal/storage/storage_handle.go
```diff
@@ -101,7 +102,23 @@ func createClientOptionForGRPCClient(ctx context.Context, clientConfig *storageu
...
+		creds, err := s2a.NewClientCreds(&s2a.ClientOptions{
+			S2AAddress:    clientConfig.S2AAddress,
+			LocalIdentity: localIdentity,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create S2A gRPC credentials: %w", err)
+		}
+		clientOpts = append(clientOpts, option.WithGRPCDialOption(grpc.WithTransportCredentials(creds)))
+		// When S2A is enabled, authentication is handled via mTLS at the transport level.
+		// Passing WithoutAuthentication() bypasses the default OAuth2 credential flow via Metadata Server.
```
vadlakondaswetha, line 123: Please add this line also: If not, every request will have a oauth token attached and it will get used instead of S2A.

internal/storage/storage_handle.go
```diff
@@ -275,7 +292,7 @@ func createHTTPClientHandle(ctx context.Context, clientConfig *storageutil.Stora
 	var clientOpts []option.ClientOption
 	var tokenSrc oauth2.TokenSource = nil
 
-	if clientConfig.AnonymousAccess {
+	if clientConfig.S2AAddress != "" || clientConfig.AnonymousAccess {
```
vadlakondaswetha: please add the comment here also why we are doing wihtoutauthentication when s2a is enabled

internal/storage/storageutil/client.go
```diff
@@ -130,56 +145,53 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
...
+			DialTLSContext:    dialTLSContext,
 			Proxy:             http.ProxyFromEnvironment,
 			DisableKeepAlives: true,
 			MaxConnsPerHost:   storageClientConfig.MaxConnsPerHost,
 			ForceAttemptHTTP2: true,
 		}
 	}
 
+	var clientTransport http.RoundTripper = transport
 	if storageClientConfig.AnonymousAccess {
```
vadlakondaswetha: Please do not add code like  - empty if-else blocks

internal/storage/storage_handle.go
sruthi-talluri: Done. Added the comment explaining that omitting WithoutAuthentication would cause an OAuth token to be attached to each request and used instead of S2A.

internal/storage/storage_handle.go
sruthi-talluri: Done. Separated the S2A and anonymous access checks and added the comment explaining why WithoutAuthentication is passed when S2A is enabled.

internal/storage/storageutil/client.go
sruthi-talluri: Done. Refactored the condition to !storageClientConfig.AnonymousAccess && storageClientConfig.S2AAddress == "" to wrap with oauth2.Transport directly, eliminating the empty branch blocks.

internal/storage/storageutil/client.go
```diff
@@ -113,11 +115,24 @@ func CreateHttpClient(storageClientConfig *StorageClientConfig, tokenSrc oauth2.
...
+			LocalIdentity: localIdentity,
+		})
+	}
+
 	var transport *http.Transport
 	// Using http1 makes the client more performant.
 	if storageClientConfig.ClientProtocol == cfg.HTTP1 {
 		transport = &http.Transport{
 			DialContext:         dialer.DialContext,
+			DialTLSContext:      dialTLSContext,
```
vadlakondaswetha, line 185: did you test if http1 & http2 are working fine. I am wondering if setting empty dialTLSContext has any implications

vadlakondaswetha (dismissed):

internal/storage/storageutil/client.go
sruthi-talluri: Both were tested, but I did have to make few changes. 

The empty dialTLSContext itself is harmless: it's a func-typed field, so nil is identical to leaving it unset — hasCustomTLSDialer() returns false and dialConn takes the standard path, unchanged from before this PR.

The http1 path was broken though. s2a-go hardcodes ALPN to h2 and its dialer returns a *tls.Conn, so http.Transport sees NegotiatedProtocol h2. Our http1 branch sets TLSNextProto to an empty map to disable HTTP/2, so it would send HTTP/1.1 over an h2-negotiated connection and fail with malformed HTTP response "\x00\x00$\x04..." (the server's SETTINGS frame). 
My earlier http1 check only hit GetStorageLayout, which goes over the gRPC control client, so it missed this.

Fixed: the http1 branch now builds the config via s2a.NewTLSClientConfigFactory and pins NextProtos to http/1.1. http2 is unchanged. Verified against s2a-go's fake S2A server — http1 negotiates HTTP/1.1, http2 negotiates h2, and the old path reproducibly fails. 

Happy to add regression tests if you'd like them here.

PR #5075, opened 2026-09-03
feat(cfg): compile protobuf and implement config serialization

PranjalC100:
### Description
`cfg/config.proto` defines the telemetry schema and privacy scrubbing rules for GCSFuse configuration flags. 
At runtime, GCSFuse will log its active mount configuration to Bigstore telemetry via User-Agent headers without leaking sensitive paths or PII.

This PR implements the runtime serialization layer that bridges the in-memory configuration (`cfg.Config`) with the protobuf schema (`pb.Config`):

* **Compiled Proto Bindings (`cfg/pb/config.pb.go`)**: Compiles `cfg/config.proto` into Go protobuf structs using `sint64` (ZigZag encoding) to minimize wire size for default/negative sentinels.
* **Envelope Message (`UserAgentConfig`)**: Wraps `pb.Config` inside an outer `UserAgentConfig` message envelope.
* **Runtime Serialization (`cfg/config_proto.go`)**:
  * `(config *Config) ToProto() *pb.Config`: Transforms the active runtime config into the sanitized `pb.Config`.
  * `(config *Config) SerializeConfigToProtoBase64() (string, error)`: Wraps `pb.Config` into `pb.UserAgentConfig`, marshals it, and encodes it into a URL-safe unpadded Base64 string (`base64.RawURLEncoding`).
* **Zero-Setup Codegen (`buf.gen.yaml` & `main.go`)**: Uses `buf generate` running directly via `go run github.com/bufbuild/buf/cmd/buf@v1.50.0 generate` and `protoc-gen-go@v1.36.12` without requiring any local C++ `protoc` installation or setup scripts.

---

#### Manual Changes
| File | Description |
| :--- | :--- |
| `buf.gen.yaml` | Buf configuration for compiling `cfg/config.proto` using pure-Go `protoc-gen-go`. |
| `main.go` | Adds `//go:generate` directive for `buf generate`. |
| `tools/config-gen/main.go` | Registers `config_proto_mapper` template in the config-gen pipeline. |
| `tools/config-gen/type_template_data_gen.go` | Implements `computeProtoMappings` for Go accessors, `sint64` numeric mappings, and whitelisted strings (`machine-type`, `profile`). |
| `tools/config-gen/type_template_data_gen_test.go` | Unit tests for proto mapping logic and whitelisted fields. |
| `tools/config-gen/templates/config_proto.tpl` | Protobuf schema template defining `UserAgentConfig` envelope and `Config`. |
| `tools/config-gen/templates/config_proto_mapper.tpl` | Template for `cfg/config_proto.go` (`ToProto` and `SerializeConfigToProtoBase64`). |
| `cfg/config_proto_test.go` | Table-driven unit tests verifying nil-safety, all data types & PII scrubbing rules (`ToProto`), and valid URL-safe Base64 regex format (`SerializeConfigToProtoBase64`). |

#### Autogenerated Files
| File | Description |
| :--- | :--- |
| `cfg/config.proto` | Generated Protobuf schema. |
| `cfg/pb/config.pb.go` | Protobuf Go structs compiled from `cfg/config.proto` via `buf generate`. |
| `cfg/config_proto.go` | Runtime serialization engine implementing `ToProto` and `SerializeConfigToProtoBase64`. |

---

### Link to Issue
[558660838](https://buganizer.corp.google.com/issues/558660838)

---

### Testing Details
* **Unit Tests (`./cfg/...`)**: Passed
* **Generator Unit Tests (`./tools/config-gen/...`)**: Passed


main.go
```diff
@@ -40,6 +40,8 @@ func logPanic() {
 // Refer https://go.dev/blog/generate for details.
 //
 //go:generate go run -C tools/config-gen . --paramsFile=../../cfg/params.yaml --outDir=../../cfg --templateDir=templates
+//go:generate protoc --go_out=. --go_opt=module=github.com/googlecloudplatform/gcsfuse/v3 cfg/config.proto
+//go:generate go fmt ./cfg/...
```
kislaykishore, line 44: We currently format as part of the `make generate` build rule. You can get rid of this statement.

tools/config-gen/templates/config_proto_mapper.tpl
```diff
@@ -0,0 +1,56 @@
...
+	return &pb.Config{
+	{{- range .ProtoMappings}}
+		{{.ProtoFieldName}}: {{.GoExpression}},
+	{{- end}}
+	}
+}
+
+// SerializeConfigToProtoBase64 maps the given Config to its flat Protobuf representation,
+// marshals it using proto3 binary encoding, and encodes it into URL-safe Base64 without padding.
+func SerializeConfigToProtoBase64(config *Config) (string, error) {
```
kislaykishore, line 44: How about we make these methods part of the config struct? It'll be more ergonomic to execute config.ToProto() or config.SerializeConfigToProtoBase64.
```suggestion
func (config *Config) SerializeConfigToProtoBase64() (string, error) {
```

cfg/config_proto_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	protoConfig := ToProto(&config)
+	require.NotNil(t, protoConfig)
+
+	m := protoConfig.ProtoReflect()
+	fields := m.Descriptor().Fields()
+	require.Greater(t, fields.Len(), 0, "expected proto fields in descriptor")
+
+	var unsetFields []string
+	for i := 0; i < fields.Len(); i++ {
```
kislaykishore, line 201: Modernize.
```suggestion
	for i := range fields.Len() {
```

cfg/config_proto_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+func TestToProto_Nil(t *testing.T) {
+	assert.Nil(t, ToProto(nil))
+}
+
+func TestSerializeConfigToProtoBase64_Nil(t *testing.T) {
+	str, err := SerializeConfigToProtoBase64(nil)
+	assert.NoError(t, err)
```
kislaykishore, line 35: Add blank lines between Arrange, Act and assert blocks.
```suggestion
	assert.NoError(t, err)
	
```

cfg/config_proto_test.go
```diff
@@ -0,0 +1,209 @@
...
+			ClientProtocol:    "http1",
+			HttpClientTimeout: 30 * time.Second,
+		},
+	}
+
+	encoded, err := SerializeConfigToProtoBase64(config)
+	require.NoError(t, err)
+	assert.NotEmpty(t, encoded)
+
+	decodedBytes, err := base64.RawURLEncoding.DecodeString(encoded)
```
kislaykishore, line 139: This is a cyclic test - the method `DecodeString` is testing itself. Tomorrow, if DecodeString's implementation becomes buggy, this method would still succeed. Instead, let's use actual bytes for your asserts.

cfg/config_proto_test.go
```diff
@@ -0,0 +1,209 @@
...
+// converts it to pb.Config, and uses Protobuf reflection to ensure every generated field in pb.Config
+// has received a non-zero value.
+func TestToProto_AutomatedReflection(t *testing.T) {
+	var config Config
+	populateNonZero(reflect.ValueOf(&config).Elem())
+
+	protoConfig := ToProto(&config)
+	require.NotNil(t, protoConfig)
+
+	m := protoConfig.ProtoReflect()
```
kislaykishore, line 196: This test is complicated with the use of reflection. A test should be readable and should have as little logic as possible. I would recommend we just test for all data-types once instead of having such a test that populates every field in the configuration. Because if you do the latter, who is going to test the test's logic!

.github/workflows/ci.yml
```diff
@@ -48,6 +48,10 @@ jobs:
         go-version-file: '.go-version'
     - name: Install goimports
       run: go install golang.org/x/tools/cmd/goimports@latest
+    - name: Install Protoc & protoc-gen-go
+      run: |
+        sudo apt-get update && sudo apt-get install -y protobuf-compiler
+        go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.12
```
kislaykishore, line 54: Do we need to make corresponding changes in CD pipeline as well?

.github/workflows/ci.yml
```diff
@@ -48,6 +48,10 @@ jobs:
         go-version-file: '.go-version'
     - name: Install goimports
       run: go install golang.org/x/tools/cmd/goimports@latest
+    - name: Install Protoc & protoc-gen-go
+      run: |
+        sudo apt-get update && sudo apt-get install -y protobuf-compiler
```
kislaykishore, line 53: We need to ensure that the protobuf-compiler is also pinned to a specific version. Otherwise, as the protobuf-compiler version changes in future Ubuntu versions, the formatting and imports check will begin to fail.
Perhaps using a docker/podman image both on our dev machines as well as in the CI/CD pipelines might be the way to get the version pinned.

main.go
PranjalC100: ACK.

.github/workflows/ci.yml
PranjalC100: Verified that GCSFuse CD pipeline only runs `build_gcsfuse` (`go build`) against the checked-in source code and does not run go generate or formatting checks.

.github/workflows/ci.yml
PranjalC100: Adding a shared script (`tools/scripts/install_proto_tools.sh`) used by both local development (`make generate`) and CI (`ci.yml`) that automatically downloads and pins both `protoc (v36.1)` and `protoc-gen-go (v1.36.12)`, so developers and CI always use the exact same versions without worrying about local setup or Ubuntu upgrades.

cfg/config_proto_test.go
PranjalC100: ACK.

cfg/config_proto_test.go
PranjalC100: ACK.

cfg/config_proto_test.go
PranjalC100: Removed `populateNonZero` and `TestToProto_AutomatedReflection` and replaced them with a table-driven `TestToProto_AllDataTypes`. It explicitly tests each data type (bool, int64, Octal, float64, Duration, enums, and PII-scrubbed strings/paths/slices) using proto.Equal.

cfg/config_proto_test.go
PranjalC100: Replaced the cyclic `config.ToProto()` comparison with a table-driven test asserting against exact expected `protobuf wire bytes ([]byte)` and expected `Base64 strings`.

tools/config-gen/templates/config_proto_mapper.tpl
PranjalC100: Updated `config_proto_mapper.tpl` and regenerated `cfg/config_proto.go` so both are **receiver methods** on `Config (config.ToProto()` and `config.SerializeConfigToProtoBase64()`

cfg/config_proto_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			encoded, err := tc.config.SerializeConfigToProtoBase64()
+
+			require.NoError(t, err)
+			assert.Equal(t, tc.expectedBase64, encoded)
+			assert.NotContains(t, encoded, "=")
+			assert.NotContains(t, encoded, "+")
+			assert.NotContains(t, encoded, "/")
```
kislaykishore, line 219: Why not just ensure that `expectedBase64` doesn't contain "=" and other symbols?
```suggestion
```

cfg/config_proto_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+			decodedBytes, err := base64.RawURLEncoding.DecodeString(encoded)
```
kislaykishore, line 221: Test should ensure a single thing - let's split this into three different subtests

tools/scripts/install_proto_tools.sh
kislaykishore, line 1: Could you check out `buf` for proto-generation? That will remove the requirement of installing `protoc`. Refer: https://github.com/bufbuild/buf

cfg/config_proto_test.go
PranjalC100: Okay will check that `tc.expectedBase64` consists only of the characters permitted in `base64.RawURLEncoding`.

cfg/config_proto_test.go
PranjalC100: Done as per discussion.

cfg/config_proto_test.go
```diff
@@ -0,0 +1,247 @@
...
+			})
+
+			t.Run("DecodedBytes", func(t *testing.T) {
+				encoded, err := tc.config.SerializeConfigToProtoBase64()
+				require.NoError(t, err)
+
+				decodedBytes, err := base64.RawURLEncoding.DecodeString(encoded)
+
+				require.NoError(t, err)
+				assert.Equal(t, tc.expectedBytes, decodedBytes)
```
kislaykishore, line 228: This is flaky - proto serialization is not always deterministic. So, SerializeConfigToProtoBase64() might produce different results and that'll lead to different bytes generated by `DecodeString`. I guess we should get rid of this test.

.github/workflows/ci.yml
PranjalC100: Switched to `pure-Go buf (v1.50.0)` and `protoc-gen-go@v1.36.12` via go run in //go:generate, completely removing OS protoc dependencies while strictly pinning compiler versions across all environments.

cfg/config_proto_test.go
PranjalC100: ACK. Removed this test.

cfg/config.proto
```diff
@@ -21,6 +21,10 @@ package gcsfuse.cfg.v1;
 
 option go_package = "github.com/googlecloudplatform/gcsfuse/v3/cfg/pb";
 
+message UserAgentConfig {
```
kislaykishore:
```suggestion
message ConfigPayload {
```

tools/config-gen/type_template_data_gen_test.go
```diff
@@ -126,3 +126,104 @@ func TestConstructTypeTemplateData(t *testing.T) {
 	require.Len(t, loggingMsg.Fields, 1)
 	assert.Equal(t, "Severity", loggingMsg.Fields[0].FieldName)
 }
+
+func TestComputeProtoMappings(t *testing.T) {
+	params := []Param{
```
kislaykishore: Please include the expected "Go expression" and field names in the test table as well so that output and inputs are together. This will make the test more readable.

tools/config-gen/type_template_data_gen.go
```diff
@@ -70,6 +70,78 @@ func computeProtoFields(params []Param) []protoFieldTemplateData {
...
+	ProtoFieldName string
+	GoExpression   string
+	ProtoTag       int
+}
+
+func computeProtoMappings(params []Param) ([]protoMappingTemplateData, error) {
+	mappings := make([]protoMappingTemplateData, 0, len(params))
+	for _, p := range params {
+		if p.ConfigPath == "" || p.ProtoFieldName == "" || p.ProtoType == "" {
+			continue
```
kislaykishore, line 83: Should we throw validation errors if a param is not deprecated and yet protofieldname or prototype are blanks?

cfg/config.proto
PranjalC100: ACK.

tools/config-gen/type_template_data_gen.go
PranjalC100: We actually already guarantee this upstream in [parser.go](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/config-gen/parser.go#L229). All non-deprecated params are validated to have a ConfigPath, which in turn ensures [populateProtoMetadata()](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/tools/config-gen/parser.go#L129) will assign their **ProtoFieldName** and **ProtoType**. 

The continue here safely skips legacy deprecated flags (which intentionally have no ConfigPath), but an active param will never have blank proto metadata at this stage.

tools/config-gen/type_template_data_gen_test.go
PranjalC100: Converted the tests in this file to table-driven format, containing both input & output/expected values.

tools/config-gen/type_template_data_gen_test.go
```diff
@@ -111,18 +170,153 @@ func TestConstructTypeTemplateData(t *testing.T) {
...
+
+	mappings, err := computeProtoMappings(params)
+
+	require.NoError(t, err)
+	require.Len(t, mappings, 7)
+	mappingIdx := 0
+	for _, tc := range testCases {
+		if tc.isSkipped {
+			continue
+		}
```
kislaykishore: What are we testing by skipping the test case? If it's to verify that it is successfully ignored then let's have a separate test for that e.g.

```
func TestDeprecatedParam(t *testing.T) {
  mappings, err := computeProtoMappings([]Param{
				FlagName:       "deprecated-flag",
				ConfigPath:     "",
				Type:           "bool",
				ProtoType:      "",
				ProtoFieldName: "",
				ProtoTag:       0,
			})
  require.NoError(t, err)
  
  assert.Empty(t, mappings) 
}
```

tools/config-gen/type_template_data_gen_test.go
```diff
@@ -111,18 +170,153 @@ func TestConstructTypeTemplateData(t *testing.T) {
...
+		assert.Equal(t, tc.expectedProtoFieldName, mappings[mappingIdx].ProtoFieldName)
+		assert.Equal(t, tc.expectedGoExpression, mappings[mappingIdx].GoExpression)
+		assert.Equal(t, tc.expectedProtoTag, mappings[mappingIdx].ProtoTag)
+		mappingIdx++
+	}
```
kislaykishore: Tests should be minimal and focussed [go/tott/319](https://goto.google.com/tott/319). Get rid of the expectedProtoTag as well since it's always the same as ProtoTag. Same comment for `TestComputeProtoFields` as well.

```suggestion
	for _, tc := range testCases {
	    mappings, err := computeProtoMappings([]{tc.param})
	    require.Len(t, mappings, 1)
	    
	    mapping = mappings[0]
		assert.Equal(t, tc.expectedProtoFieldName, mapping.ProtoFieldName)
		assert.Equal(t, tc.expectedGoExpression, mapping.GoExpression)
		assert.Equal(t, tc.expectedProtoTag, mapping.ProtoTag)
	}
```

tools/config-gen/type_template_data_gen_test.go
PranjalC100: Done! Refactored the table tests to focus strictly on single-param mappings and moved the deprecated flag checks into dedicated unit tests.

tools/config-gen/type_template_data_gen_test.go
PranjalC100: Done!

kislaykishore (approved):

PR #5080, opened 2026-09-07
fix(cfg): auto-adjust max-read-ahead-kb to match fuse-max-request-size-kb when kernel reader is enabled

kislaykishore:
### Description
When kernel reader is enabled and a user sets a large `fuse-max-request-size-kb` (e.g. 255 MiB), GCSFuse kernel read calls were previously limited to 128 MiB for regional buckets due to constraints from the `max-read-ahead-kb` default.

This change:
- Automatically adjusts `max-read-ahead-kb` to `max(max-read-ahead-kb, fuse-max-request-size-kb)` when `enable-kernel-reader` is enabled and the user did not explicitly set `max-read-ahead-kb`.
- Preserves explicit user configuration of `max-read-ahead-kb` (via CLI or config file) even if lower than `fuse-max-request-size-kb`.
- Retains existing default read-ahead values when `fuse-max-request-size-kb` is lower than the default or unset.
- Leaves `max-read-ahead-kb` unchanged when the kernel reader is disabled.

b/558219582

### Testing details
1. Manual - Verified compilation, format, and static analysis via `make build` and `golangci-lint`.
3. Integration tests - Tested the behavior using Tessellations checkpoint benchmark.

N/A

### DIFF
--- a/cfg/rationalize.go
+++ b/cfg/rationalize.go
@@ -176,6 +176,18 @@ func resolveOnlyDir(c *Config) {
 	}
 }
 
+// resolveKernelReadAhead auto-adjusts max-read-ahead-kb to match fuse-max-request-size-kb
+// when kernel reader is enabled, unless the user explicitly configured max-read-ahead-kb.
+func resolveKernelReadAhead(v *viper.Viper, c *Config) {
+	if !c.FileSystem.EnableKernelReader {
+		return
+	}
+	if v != nil && v.IsSet("file-system.max-read-ahead-kb") {
+		return
+	}
+	c.FileSystem.MaxReadAheadKb = max(c.FileSystem.MaxReadAheadKb, c.FileSystem.FuseMaxRequestSizeKb)
+}
+
 // Rationalize updates the config fields based on the values of other fields.
 func Rationalize(v *viper.Viper, c *Config, optimizedFlags []string) error {
 	var err error
@@ -198,6 +210,7 @@ func Rationalize(v *viper.Viper, c *Config, optimizedFlags []string) error {
 	resolveFileCacheAndBufferedReadConflict(v, c)
 	resolveGCSRetriesConfig(&c.GcsRetries)
 	resolveOnlyDir(c)
+	resolveKernelReadAhead(v, c)
 
 	return nil
 }
--- a/cfg/rationalize_test.go
+++ b/cfg/rationalize_test.go
@@ -930,7 +930,7 @@ func TestResolveOnlyDir(t *testing.T) {
 	}
 }
 
-func Test_resolve_kernel_read_ahead(t *testing.T) {
+func TestResolveKernelReadAhead(t *testing.T) {
 	testCases := []struct {
 		name                string
 		userSetFlags        map[string]any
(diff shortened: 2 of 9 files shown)


cmd/config_rationalization_test.go
```diff
@@ -111,3 +113,364 @@ func TestRationalizeCloudMetricsExportIntervalSecs(t *testing.T) {
...
+	cmdArgs := append([]string{"gcsfuse"}, args...)
+	cmdArgs = append(cmdArgs, "a")
+	cmd.SetArgs(convertToPosixArgs(cmdArgs, cmd))
+	if err = cmd.Execute(); err != nil {
+		return nil, err
+	}
+	return &info, nil
+}
+
+func Test_rationalize_kernel_read_ahead_cli_flags(t *testing.T) {
```
abhishek10004: nit: use Pascal case for test function name

cfg/rationalize_test.go
```diff
@@ -929,3 +929,346 @@ func TestResolveOnlyDir(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var v *viper.Viper
+			if !tc.nilViper {
+				v = viper.New()
+				for key, val := range tc.userSetFlags {
+					v.Set(key, val)
+				}
+			}
+
```
abhishek10004: nit: empty line not needed here. Similar for other tests as well.

cfg/rationalize.go
```diff
@@ -176,6 +176,18 @@ func resolveOnlyDir(c *Config) {
...
+// resolveKernelReadAhead auto-adjusts max-read-ahead-kb to match fuse-max-request-size-kb
+// when kernel reader is enabled, unless the user explicitly configured max-read-ahead-kb.
+func resolveKernelReadAhead(v *viper.Viper, c *Config) {
+	if !c.FileSystem.EnableKernelReader {
+		return
+	}
+	if v != nil && v.IsSet("file-system.max-read-ahead-kb") {
+		return
+	}
+	c.FileSystem.MaxReadAheadKb = max(c.FileSystem.MaxReadAheadKb, c.FileSystem.FuseMaxRequestSizeKb)
```
abhishek10004, line 188: Let's update the documentation for this flag in params.yaml file to reflect this as well

cmd/config_rationalization_test.go
```diff
@@ -111,3 +113,364 @@ func TestRationalizeCloudMetricsExportIntervalSecs(t *testing.T) {
...
+
+func Test_rationalize_kernel_read_ahead_viper_is_set(t *testing.T) {
+	testCases := []struct {
+		name          string
+		yamlContent   string
+		cliArgs       []string
+		expectedIsSet bool
+	}{
+		{
+			name:          "cli_flag_zero_is_set",
```
abhishek10004: nit: the name sounds slightly confusing, can we use a simpler name like cli_flag_is_set_with_value_zero? similar for other configs in this test

cmd/config_rationalization_test.go
abhishek10004: Similar for other tests throughout this PR

abhishek10004 (approved):
LGTM

PR #5085, opened 2026-09-08
feat(flags): support enable-grpc-by-default flag for dynamic gRPC selection

dangarhs:
### Description
- Added hidden CLI flag `--enable-grpc-by-default` (and config parameter `gcs-connection.enable-grpc-by-default`, default `false`) to consume the signal from the GKE GCSFuse CSI Driver (cl/973240436).
- When `--enable-grpc-by-default` is set, GCSFuse switches `ClientProtocol` to `grpc`.
- Preserves explicit customer overrides (e.g. `--client-protocol=http1`, `http2`, `grpc`, or `httpmtls`).
- Safely retains `http1` default for existing clusters / non-GKE setups where the flag is omitted/false.

b/556501118

### Testing details
1. Manual - Verified CLI flag parsing and resolution across permutations with compiled `gcsfuse` binary mounts.
2. Unit tests - Added unit test coverage:
   - `cfg/rationalize_test.go`: Added `TestResolveClientProtocol` covering all 4 non-redundant scenarios (`enable-grpc-by-default` true/false with explicit user override or omitted).
   - Ran `go test ./cfg/...` and `go test ./cmd/...` (all passed).
3. Integration tests - Ran full test suite across packages.

No

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -658,6 +658,8 @@ type GcsConnectionConfig struct {
 
 	CustomEndpoint string `yaml:"custom-endpoint"`
 
+	EnableGrpcByDefault bool `yaml:"enable-grpc-by-default"`
+
 	EnableHttpDnsCache bool `yaml:"enable-http-dns-cache"`
 
 	ExperimentalEnableJsonRead bool `yaml:"experimental-enable-json-read"`
@@ -1022,6 +1024,12 @@ func BuildFlagSet(flagSet *pflag.FlagSet) error {
 	flagSet.BoolP("enable-google-lib-auth", "", true, "Enable google library authentication method to fetch the credentials")
 
 	if err := flagSet.MarkHidden("enable-google-lib-auth"); err != nil {
+		return err
+	}
+
+	flagSet.BoolP("enable-grpc-by-default", "", false, "If true, enables gRPC protocol by default (e.g. for new GKE deployments).")
+
+	if err := flagSet.MarkHidden("enable-grpc-by-default"); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


cfg/config.go
```diff
@@ -658,6 +658,8 @@ type GcsConnectionConfig struct {
 
 	CustomEndpoint string `yaml:"custom-endpoint"`
 
+	EnableGrpcByDefault bool `yaml:"enable-grpc-by-default"`
```
vadlakondaswetha, line 661: should we just call enable-grpc instead of enable-grpc-by-default

cfg/rationalize_test.go
```diff
@@ -1267,3 +1267,73 @@ func TestRationalizeWithBucketOptimization(t *testing.T) {
...
+			userSetFlags: map[string]any{
+				ClientProtocolConfigKey: "http1",
+			},
+			expectedProtocol: HTTP1,
+		},
+		{
+			name: "enable-grpc-by-default is false and client passes grpc",
+			config: &Config{
+				GcsConnection: GcsConnectionConfig{
+					ClientProtocol:      GRPC,
```
vadlakondaswetha, line 1315: what happens if we set this to http1?

dangarhs: If we set this to http1, the test would fail because Rationalize() will return without making any change when it sees the user flag is set. We are setting GRPC here to mock viper unmarshaled (--client-protocol=grpc) flag. I didnt add test for that because its an impossible state.

cfg/config.go
dangarhs: Technically we can, but I prefer enable-grpc-by-default.. if enable-grpc is passed with explicit client-protocol, fuse will honor client-protocol. So, this is just a fallback default rather than a strict override. 

How about default-to-grpc ?

vadlakondaswetha (approved):

PR #5091, opened 2026-09-14
test(list_large_dir): check if bucket dir is empty before creating files and clean up in teardown

anushka567:
### Description
Currently in the `list_large_dir` integration test suite:
1. `SetupSuite` unconditionally ran `DeleteAllObjectsWithPrefix` to delete pre-existing objects before mounting.
2. `prepareTestDirectory` created and uploaded 12,000 files and directories on GCS regardless of whether the bucket directory was already populated.
3. Test teardown did not clean up the generated files and directories on GCS, leaving thousands of objects behind in the test bucket after test runs.

### Changes
- Added `IsDirEmptyOnGCS(ctx, client, bucketName, prefix)` helper in `tools/integration_tests/util/client/storage_client.go` to check whether objects exist under a GCS bucket prefix.
- Updated `prepareTestDirectory` in `list_dir_with_twelve_thousand_files_test.go` to check if the bucket directory is empty, and only create & upload files / directories if it is empty.
- Removed unconditional object deletion from `SetupSuite`.
- Added `DeleteAllObjectsWithPrefix` cleanup in `TearDownTest` to properly clean up test objects after each test execution.

### Testing details
- Manual: Yes, bucket left in clean state after test run complete.
- Unit tests: Verified package compilation
- Integration tests: Automated


tools/integration_tests/list_large_dir/list_dir_with_twelve_thousand_files_test.go
```diff
@@ -235,14 +236,22 @@ func prepareTestDirectory(t *testing.T, withExplicitDirs bool, withImplicitDirs
 		t.Fatalf("Failed to create directory: %v", err)
 	}
 
-	createFilesAndUpload(t, testDirPathOnBucket)
-
-	if withExplicitDirs {
-		testdataCreateExplicitDir(t, testEnv.ctx, testEnv.storageClient, testDirPathOnBucket)
+	bucketName, dirPathInBucket := operations.SplitBucketNameAndDirPath(t, testDirPathOnBucket)
+	isEmpty, err := IsDirEmptyOnGCS(testEnv.ctx, testEnv.storageClient, bucketName, dirPathInBucket)
```
abhishek10004, line 240: We're not doing anything when the directory is not empty. We should either fail fast here or call DeleteAllObjectsWithPrefix here only to recover

anushka567: Added logic to delete the existing files before proceeding, if deletion fails, fail the test.

abhishek10004 (approved):

PR #5092, opened 2026-09-14
feat(grpc-default-on): dynamic client protocol selection and directpath fallback

dangarhs:
### Description
This PR implements dynamic gRPC client protocol selection and DirectPath fallback for GCSFuse:

- **Protocol Resolution (`resolveClientProtocol` in `cfg/rationalize.go`)**:
  - When `client-protocol` is explicitly specified by the user (`v.IsSet(ClientProtocolConfigKey)`), honors the user-provided protocol and sets `c.GcsConnection.EnableGrpcByDefault = false`.
  - When `client-protocol` is not explicitly specified, selects `GRPC` and sets `c.GcsConnection.EnableGrpcByDefault = true` if either:
    - `--enable-grpc-by-default=true` is set, **or**
    - Running in a **GKE environment** (`IsGKEEnvironment(os.Args[len(os.Args)-1])`) on a **`high-performance` machine** (`machineTypeToGroupMap[c.MachineType] == "high-performance"`).
  - Otherwise, retains `HTTP1` with `EnableGrpcByDefault = false`.
- **DirectPath Enforcement & HTTP/1.1 Fallback (`internal/storage/storage_handle.go` & `cmd/legacy_main.go`)**:
  - Wires `EnableGrpcByDefault: newConfig.GcsConnection.EnableGrpcByDefault` into `StorageClientConfig` (`internal/storage/storageutil/client.go`).
  - In `createGRPCClientHandle`, enforces DirectPath (`experimental.WithDirectConnectivityEnforced()` and `verifyDirectPathConnectivity`) when `enforceDirectPath := !isBucketRapid && clientConfig.EnableGrpcByDefault` is `true`.
  - When `EnableGrpcByDefault` is `false` (explicit `--client-protocol=grpc`), skips DirectPath verification and never falls back to HTTP, allowing the Go Storage SDK to handle CloudPath fallback.
  - When `EnableGrpcByDefault` is `true` and DirectPath verification fails on a non-rapid bucket, `createNonBidiGRPCClientWithHttpFallback` updates `sh.clientConfig.ClientProtocol = cfg.HTTP1` and falls back to creating and caching an HTTP/1.1 client (`sh.httpClient`).
  - Rapid/Zonal buckets continue to unconditionally use `grpcClientWithBidiConfig` (`!isBucketRapid` is `false`).

- Ticket: b/556501233
- Design Doc: go/grpc-client-strategy-for-gcsfuse

### Testing details
1. Manual - Verified client selection and fallback paths.
2. Unit tests:
   - `cfg/rationalize_test.go` (`TestResolveClientProtocol`): Covers GKE vs. GCE across `high-performance` and standard machine types, `--enable-grpc-by-default`, and explicit `client-protocol` overrides (`http1`, `grpc`).
   - `internal/storage/storage_handle_test.go` (`TestStorageHandleTestSuite`):
     - `TestGetClient_Regional_ExplicitHTTP`, `TestGetClient_Rapid_ExplicitHTTP_SilentlyOverridesToGRPC`, `TestGetClient_ExplicitGRPC`, `TestGetClient_Rapid_DefaultGRPC`
     - `TestGetClient_Regional_EnableGrpcByDefault_DirectPathFails_FallbackToHTTP`
     - `TestGetClient_Regional_EnableGrpcByDefault_CachedDirectPathClient_Reused`
     - `TestGetClient_Regional_EnableGrpcByDefault_ContextCanceled_DoesNotFallbackToHTTP`
3. Integration tests - NA

No


internal/storage/storage_handle.go
```diff
@@ -460,62 +471,161 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
...
+	if err == nil {
+		if verifyErr := verifyDirectPathConnectivity(ctx, &sh.clientConfig, bucketName, grpcClient, billingProject); verifyErr != nil {
+			logger.Warnf("DirectPath verification failed for bucket %q: %v. Falling back to HTTP client.", bucketName, verifyErr)
+			sh.mu.Lock()
+			if sh.directPathFailedBuckets == nil {
+				sh.directPathFailedBuckets = make(map[string]bool)
+			}
+			sh.directPathFailedBuckets[bucketName] = true
+			sh.mu.Unlock()
+			// verifyDirectPathConnectivity already closes the unusable gRPC client.
```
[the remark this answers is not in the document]
dangarhs: Fixed in ec22b4147. Note that we specifically guard with `ctx.Err() != nil` rather than `!errors.Is(verifyErr, context.DeadlineExceeded)`. Because `verifyDirectPathConnectivity` enforces its own internal 15-second probe timeout via `NewCustomRetryConfig`, a genuine network blackhole or unreachable DirectPath endpoint surfaces as `context.DeadlineExceeded`. Checking `ctx.Err() != nil` cleanly separates user-initiated caller aborts (which should not poison the cache) from legitimate 15-second DirectPath probe timeouts (which must be memoized so subsequent lookups do not repeatedly stall for 15s).

internal/storage/storage_handle.go
```diff
@@ -460,62 +471,161 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
...
+		return client, err
+	}
+
+	// Release sh.mu before performing potentially slow DirectPath network verification
+	// to prevent head-of-line blocking for unrelated buckets and requests.
+	sh.mu.Unlock()
+
+	// Acquire dpMu to serialize DirectPath verification attempts.
+	sh.dpMu.Lock()
+	defer sh.dpMu.Unlock()
```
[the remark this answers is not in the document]
dangarhs: Fixed in ec22b4147 by replacing `dpMu` with `singleflight.Group` keyed by `bucketName` in `getOrCreateHighEndVMClient`. While `baseDirInode.LockForChildLookup()` currently serializes dynamic bucket lookups at the FUSE layer via an exclusive lock (`d.mu.Lock()`), implementing `singleflight` and `sh.mu` here ensures `storageClient` remains strictly self-contained, encapsulates its own thread-safety invariants, and is future-proofed against head-of-line blocking once upstream FUSE directory locks are decoupled.

internal/storage/storageutil/custom_retry.go
```diff
@@ -140,6 +183,10 @@ func ShouldRetryWithRetryContext(err error, retryCtx *storage.RetryContext) bool
 	if !ShouldRetryWithoutLogging(err) {
 		return false
 	}
+	if isDirectPathUnavailable(err) && retryCtx != nil && retryCtx.Attempt >= maxDirectPathUnavailableAttempts {
+		logger.Warnf("DirectPath unavailable retry attempts exhausted (%d). Failing without infinite retries.", retryCtx.Attempt)
+		return false
+	}
```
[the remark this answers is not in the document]
dangarhs: Fixed in ec22b4147 by updating the condition to `retryCtx.Attempt >= maxDirectPathUnavailableAttempts - 1` (and logging `retryCtx.Attempt + 1`), strictly enforcing the 5-attempt ceiling.

internal/storage/storage_handle.go
```diff
@@ -497,20 +511,19 @@ func (sh *storageClient) createNonBidiGRPCClientWithHttpFallback(ctx context.Con
...
+		return nil, ctx.Err()
 	}
 
-	// When grpcPathStrategy=DirectPathWithFallback, create a http client.
+	// When EnableGrpcByDefault is true and DirectPath verification fails, fall back to HTTP.
 	logger.Infof("Grpc dp is not available and falling back to Http.")
+	sh.clientConfig.ClientProtocol = cfg.HTTP1
 	if sh.httpClient == nil {
 		sh.httpClient, err = createHTTPClientHandle(ctx, &sh.clientConfig)
 	}
```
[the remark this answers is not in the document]
dangarhs: BucketHandle / getClient is only invoked once at startup for static mounts, and in dynamic mounts all SetUpBucket calls are already serialized under the root baseDirInode exclusive lock (LockForChildLookup). Therefore, concurrent calls to getClient cannot occur and an extra mutex in storageClient is unnecessary

cfg/rationalize.go
```diff
@@ -188,17 +189,21 @@ func resolveKernelReadAhead(v *viper.Viper, c *Config) {
...
+// or GKE high-performance machine).
+func resolveClientProtocol(v *viper.Viper, c *Config) {
 	if v != nil && v.IsSet(ClientProtocolConfigKey) {
+		c.GcsConnection.EnableGrpcByDefault = false
 		return
 	}
 
-	if c.EnableGrpcByDefault {
-		c.ClientProtocol = GRPC
+	isGKE := len(os.Args) > 0 && IsGKEEnvironment(os.Args[len(os.Args)-1])
```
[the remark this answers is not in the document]
dangarhs: gcsfuse already requires mountPoint to be the final element of os.Args during daemonization (cmd/legacy_main.go:452: args[len(args)-1] = mountPoint), and /dev/fd/N mountpoints are exclusively passed by the GKE CSI driver as the final positional argument.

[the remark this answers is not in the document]
dangarhs: Tracking this improvement under: b/556501227

cfg/rationalize_test.go
```diff
@@ -1270,34 +1270,70 @@ func TestRationalizeWithBucketOptimization(t *testing.T) {
...
+				GcsConnection: GcsConnectionConfig{
+					ClientProtocol:      HTTP1,
+					EnableGrpcByDefault: false,
+				},
+			},
+			expectedProtocol:            GRPC,
+			expectedEnableGrpcByDefault: true,
+		},
+		{
+			name:       "enable-grpc-by-default is true and client does not pass protocol",
```
vadlakondaswetha: why is enable-grpc-by-default is true not set as part of userSetFlags map?

cfg/rationalize_test.go
```diff
@@ -1270,34 +1270,70 @@ func TestRationalizeWithBucketOptimization(t *testing.T) {
...
+		expectedProtocol            Protocol
+		expectedEnableGrpcByDefault bool
 	}{
 		{
-			name: "enable-grpc-by-default is false and client does not pass protocol",
+			name:       "standard machine on GKE without flags keeps HTTP1",
+			mountPoint: "/dev/fd/3",
 			config: &Config{
+				MachineType: "e2-standard-4",
 				GcsConnection: GcsConnectionConfig{
```
vadlakondaswetha: what is config vs usersetflags in these tests?

cfg/rationalize.go
```diff
@@ -188,17 +189,21 @@ func resolveKernelReadAhead(v *viper.Viper, c *Config) {
...
+	isHighPerfMachine := machineTypeToGroupMap[c.MachineType] == "high-performance"
+
+	if c.GcsConnection.EnableGrpcByDefault || (isGKE && isHighPerfMachine) {
```
vadlakondaswetha: why are we not doing this based on the config we have today: https://github.com/GoogleCloudPlatform/gcsfuse/blob/399e0116b9490b89e3a1d4d184f0716131033821/cfg/params.yaml#L623. Something like saying apply this optimization to GKE in the config itself?

dangarhs: Main reason for not doing this is to keep the entire protocol resolution logic at one place. 

During implementation, I had done analysis of using optimization approach. Since the env detection flag is currently not supported in params.yaml (existing machine-based-optimization applies to both GCE and GKE).

I think the best way to support that is by adding is-gke directly in machine-based-optimization, and then adding this optimization on gcs-connection.enable-grpc-by-default in cfg/params.yaml:

```
    optimizations:
      machine-based-optimization:
        - group: "high-performance"
          value: true
          is-gke: true
```
          
We would also need to wire this up across cfg and the tools/config-gen templates to regenerate cfg/config.go and cfg/config_test.go. Even with that in place, since ApplyOptimizations only sets `EnableGrpcByDefault = true` in isolation, we will still need resolveClientProtocol in cfg/rationalize.go to set `ClientProtocol = GRPC` and handle explicit --client-protocol overrides:

```
func resolveClientProtocol(v *viper.Viper, c *Config) {
	if v != nil && v.IsSet(ClientProtocolConfigKey) {
		c.GcsConnection.EnableGrpcByDefault = false
		return
	}
	if c.GcsConnection.EnableGrpcByDefault {
		c.GcsConnection.ClientProtocol = GRPC
	}
}
```

I think we can take this refactor separately as follow-up, because this will require additional testing. Or do let me know if you think otherwise.

cfg/rationalize_test.go
dangarhs: config is the struct post-unmarshaling (with defaults + user values), and userSetFlags just mocks v.IsSet() in viper so Rationalize knows which flags were explicitly passed by the user vs defaulted.

cfg/rationalize_test.go
dangarhs: Because userSetFlags is only used for v.IsSet() checks, and in resolveClientProtocol we only check v.IsSet() for client-protocol and read c.GcsConnection.EnableGrpcByDefault directly from config.

cfg/rationalize.go
dangarhs: @vadlakondaswetha I have refactored it to use config optimization. Removed Rationalize and Kept everything in the StorageClient as discussed.

internal/storage/storage_handle.go
```diff
@@ -471,13 +482,24 @@ func (sh *storageClient) getClient(ctx context.Context, isBucketRapid bool, buck
...
 		if sh.grpcClientWithBidiConfig == nil {
-			sh.grpcClientWithBidiConfig, err = createGRPCClientHandle(ctx, &sh.clientConfig, isBucketRapid, true, bucketName, billingProject)
+			sh.grpcClientWithBidiConfig, err = createGRPCClientHandle(ctx, &sh.clientConfig, false, true, bucketName, billingProject)
 		}
 		return sh.grpcClientWithBidiConfig, err
 	}
 
+	if sh.clientConfig.ClientProtocol == "" {
+		if sh.clientConfig.EnableGrpcByDefault {
+			sh.clientConfig.ClientProtocol = cfg.GRPC
```
vadlakondaswetha: I would prefer not updating the mount config in the code. even though its not a problem here for dynamic mounts, in other places it might become an issue.

internal/storage/storage_handle_test.go
```diff
@@ -1281,3 +1283,146 @@ func (testSuite *StorageHandleTest) TestBucketHandle_NonHNS_AccessCheck_WithPref
...
+			if tc.expectBidiClient {
+				assert.Equal(testSuite.T(), sh.grpcClientWithBidiConfig, client)
+				assert.Nil(testSuite.T(), sh.httpClient)
+				_ = sh.grpcClientWithBidiConfig.Close()
+			}
+		})
+	}
+}
+
+func (testSuite *StorageHandleTest) TestGetClient_Regional_EnableGrpcByDefault_DirectPathSucceeds_UsesGRPCAndReusesCachedClient() {
```
vadlakondaswetha: we shouldn't test two usecases in a single test

dangarhs: Ack. Done.

internal/storage/storage_handle.go
dangarhs: Couldnt find any impact due to this but I agree that updating the mount configs in code can lead to code maintainability issues. Fixed it.

internal/storage/storage_handle.go
```diff
@@ -469,18 +480,31 @@ func NewStorageHandle(ctx context.Context, clientConfig storageutil.StorageClien
...
+		if sh.grpcClient == nil {
+			sh.grpcClient, err = createGRPCClientHandle(ctx, &sh.clientConfig, false, false, bucketName, billingProject)
+		}
+		return sh.grpcClient, err
+	}
+
+	// Default to HTTP1 when client-protocol is unset, or use the explicitly requested HTTP protocol.
+	if sh.clientConfig.ClientProtocol == "" || sh.clientConfig.ClientProtocol == cfg.HTTP1 || sh.clientConfig.ClientProtocol == cfg.HTTP2 || sh.clientConfig.ClientProtocol == cfg.HTTPMtls {
 		if sh.httpClient == nil {
 			sh.httpClient, err = createHTTPClientHandle(ctx, &sh.clientConfig)
```
vadlakondaswetha, line 509: if the clientProtocol is empty we will end up creating a http2 client here: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/storageutil/client.go#L116, you need the extract the protocol into a variable and pass it to the httpClientHandle method

clientProtocol = sh.clientConfig.ClientProtocol
if clientProtocol = "" {
   clientProtocol = cfg.HTTP1
}

createHttpClientHandle(ctx, clientProtocol, &sh.clientConfig)

Can we write a unit test case to validate http1 vs http2 client creation.

Also did run the manual tests to see what client is being generated? We need to think about how to write integration tests to check which client is being created.

vadlakondaswetha: infact dont use the reference clientConfig.ClientProtocol anywhere, so we understand what all places need to be updated.

dangarhs: I tried with a separate `getClientProtocol` as discussed but for fallback to http1 we still needed to pass the actual protocol used(i.e. http1 and not grpc). Hence, I had to use the local variable only. PTAL.

Reason it didnt get caught earlier,

1. In storage_handle_test we can only check sh.httpClient != nil (since storage.Client does not export its internal http.Client).
2. At https://github.com/GoogleCloudPlatform/gcsfuse/blob/8448a242a/internal/storage/storageutil/client_test.go#L58-L76- only asserting httpClient != nil without inspecting transport.ForceAttemptHTTP2 and https://github.com/GoogleCloudPlatform/gcsfuse/blob/8448a242a/internal/storage/storageutil/test_util.go#L29 defaults ClientProtocol to cfg.HTTP1

PR #5093, opened 2026-09-15
feat(metrics): Implement and capture the metadata cache read count metric.

thrivikram-karur-g:
### Description
Implementing the **metadata_cache/read_count** metric to capture the hits and misses with detailed reason for miss or hit as a counter metric to understand the **metadata_cache** performance.
`metadata_cache/read_count` was already defined in `metrics/metrics.yaml` but nothing ever
recorded it. This PR wires up emission and export.

**Summary of the code changes:**

- Emitted from exactly one place — `dirInode.LookUpChild` — with `cache_hit`,
  `entry_status` (positive/negative) and `lookup_detail` (found/ttl_expired/not_found).
- `stat_cache.go` derives the classification, `fast_stat_bucket.go` carries the miss reason
  up via `CacheMissError`, `dir.go` emits, `otelexporters.go` allowlists the prefix.
  
  
Perf test results on the PR:

pr_perf_results_5093

we do not see any regression and all bandwidth fluctuations seems normal and within healthy range.

b/512002923

### Testing details
1. Manual 

Executed the below manual test plan for checking the metrics count updates for the **metadata_cache/read_count** metric.

Ran GCSFuse with the below command.

Start GCSFuse with Prometheus metric reporting enabled on port `8080`:
```bash
go run . --foreground \
  --enable-buffered-read \
  --implicit-dirs \
  --client-protocol http1 \
  --prometheus-port=8080 \
  --enable-metadata-prefetch=false \
  --metadata-cache-negative-ttl-secs=120 \
  --metadata-cache-ttl-secs=120 \
  --log-severity=trace \
  <BUCKET_NAME> <MOUNT_DIR>
 ```
 
 Query the exposed Prometheus metrics endpoint to inspect the metadata_cache/read_count counter and its attributes:

bash
curl -s http://localhost:8080/metrics | grep "metadata_cache/read_count"

Verified manually the below test cases for corresponding change in metric counts queried from the prometheus end point.

| Scenario | Test Case | Executed Command(s) | Expected Metric Output Delta |
| :--- | :--- | :--- | :--- |
| **1** | **Cold File Miss** | `stat <MOUNT_DIR>/nonexistent_file.txt` | `metadata_cache_read_count_total{cache_hit="false",lookup_detail="not_found"} +1` |
| **2** | **Cold Directory Miss** | `stat <MOUNT_DIR>/nonexistent_dir/` | `metadata_cache_read_count_total{cache_hit="false",lookup_detail="not_found"} +1` |
| **3** | **Warm File Hit (Positive)** | 1. `stat <MOUNT_DIR>/existing_file.txt`<br>2. `stat <MOUNT_DIR>/existing_file.txt` | **1st stat:** `{cache_hit="false",lookup_detail="not_found"} +1`<br>**2nd stat:** `{cache_hit="true",entry_status="positive",lookup_detail="found"} +1` |
| **4** | **Warm Directory Hit (Positive)** | 1. `stat <MOUNT_DIR>/existing_dir/`<br>2. `stat <MOUNT_DIR>/existing_dir/` | **1st stat:** `{cache_hit="false",lookup_detail="not_found"} +1`<br>**2nd stat:** `{cache_hit="true",entry_status="positive",lookup_detail="found"} +1` |
| **5** | **Warm File Hit (Negative)** | 1. `stat <MOUNT_DIR>/nonexistent_file.txt`<br>2. `stat <MOUNT_DIR>/nonexistent_file.txt` | **1st stat:** `{cache_hit="false",lookup_detail="not_found"} +1`<br>**2nd stat:** `{cache_hit="true",entry_status="negative",lookup_detail="found"} +1` |
| **6** | **Warm Directory Hit (Negative)** | 1. `stat <MOUNT_DIR>/nonexistent_dir/`<br>2. `stat <MOUNT_DIR>/nonexistent_dir/` | **1st stat:** `{cache_hit="false",lookup_detail="not_found"} +1`<br>**2nd stat:** `{cache_hit="true",entry_status="negative",lookup_detail="found"} +1` |
| **7** | **Positive File TTL Expired** | 1. `stat <MOUNT_DIR>/existing_file.txt`<br>2. `sleep 125 && stat <MOUNT_DIR>/existing_file.txt` | **1st stat:** `{cache_hit="false",lookup_detail="not_found"} +1`<br>**2nd stat:** `{cache_hit="false",entry_status="positive",lookup_detail="ttl_expired"} +1` |
| **8** | **Positive Directory TTL Expired** | 1. `stat <MOUNT_DIR>/existing_dir/`<br>2. `sleep 125 && stat <MOUNT_DIR>/existing_dir/` | **1st stat:** `{cache_hit="false",lookup_detail="not_found"} +1`<br>**2nd stat:** `{cache_hit="false",entry_status="positive",lookup_detail="ttl_expired"} +1` |
| **9** | **Negative Entry TTL Expired** | 1. `stat <MOUNT_DIR>/nonexistent_file.txt`<br>2. `sleep 125 && stat <MOUNT_DIR>/nonexistent_file.txt` | **1st stat:** `{cache_hit="false",lookup_detail="not_found"} +1`<br>**2nd stat:** `{cache_hit="false",entry_status="negative",lookup_detail="ttl_expired"} +1` |
| **10** | **Directory Listing (No Stat Inflation)** | `ls --color=never <MOUNT_DIR>/` | **No metric delta** for `metadata_cache_read_count` *(ReadDir does not issue per-entry stat calls)* |

2. Unit tests - Added
3. Integration tests - NA

N/A

### CODE REVIEW

File: internal/storage/caching/fast_stat_bucket.go [hunk 1]
```diff
@@ -460,12 +479,11 @@ func (b *fastStatBucket) StatObject(
 		// Otherwise, return MinObject and nil ExtendedObjectAttributes.
 		m = entry
 		return
-	}
-
-	// Cache Miss Handling
-	if req.FetchOnlyFromCache {
+	} else if req.FetchOnlyFromCache {
```
thrivikram-karur-g, line 482: I do not see why we need to change this . we can keep is simple if as there is return in the previous if anyways.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -599,11 +621,11 @@ func (b *fastStatBucket) GetFolder(ctx context.Context, req *gcs.GetFolderReques
 		}
 
 		return entry, nil
-	}
-
-	if req.FetchOnlyFromCache {
+	} else if req.FetchOnlyFromCache {
```
thrivikram-karur-g, line 624: Similar comment on the unnecessary else if as commented previously for a similar code block.

metrics/helper.go
```diff
@@ -20,3 +20,15 @@ func CaptureGCSReadMetrics(mh MetricHandle, readType ReadType, downloadBytes int
 	mh.GcsReadCount(1, readType)
 	mh.GcsDownloadBytesCount(downloadBytes, readType)
 }
+
+// IsMonitoringEnabled returns true if the metric handle is non-nil and not a no-op implementation.
+func IsMonitoringEnabled(mh MetricHandle) bool {
```
thrivikram-karur-g, line 25: Do we really need this check, given we have no-op implementation already ? Need to verify it's usage and audit it's necessity.

internal/fs/inode/file.go
```diff
@@ -274,6 +277,9 @@ func (f *FileInode) checkInvariants() {
 //
 // LOCKS_REQUIRED(f.mu)
 func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool, includeExtendedObjectAttributes bool) (o *gcs.Object, isClobbered bool, isNotFound bool, err error) {
+	if f.IsLocal() {
```
thrivikram-karur-g: why are we adding this check ?
This seems like a change of behaviour. Need to audit this before proceeding ahead. If it is only for fixing some edge case unit tests, would be better to get rid of both the tests and this check.
Let us not make additional optimisations that are not needed for correcting the metric as part of this PR and blow up the change scope.

internal/fs/inode/file.go
```diff
@@ -282,6 +288,9 @@ func (f *FileInode) clobbered(ctx context.Context, forceFetchFromGcs bool, inclu
 		ReturnExtendedObjectAttributes: includeExtendedObjectAttributes,
 	}
 	m, e, err := f.bucket.StatObject(ctx, req)
+	if !forceFetchFromGcs && metrics.IsMonitoringEnabled(f.metricHandle) {
```
thrivikram-karur-g, line 288: It is not clear why clobbered check is always recording metrics in a particular way. Is it validated that this is the only possible test case and why is the metric not recorded at some other layer in the code stack trace ?

internal/fs/inode/dir.go
```diff
@@ -694,24 +710,37 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
 
-		// 3. Both lookups resulted in cache hits (no cacheMiss errors) with no results found,
-		// indicating a negative cache entry. Return nil to indicate the entry doesn't exist from cache.
-		if dirErr == nil && fileErr == nil {
+		// 3. Negative entry check:
+		// If either candidate is a negative hit, return nil to indicate the entry does not exist from cache.
+		if dirErr == nil || fileErr == nil {
+			d.recordCacheMetric(true, metrics.EntryStatusNegativeAttr, metrics.LookupDetailFoundAttr)
 			return nil, nil
 		}
```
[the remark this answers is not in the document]
thrivikram-karur-g: Agree, this seems like a critical change and needs to be fixed, looks unintended.

internal/fs/inode/dir.go
```diff
@@ -391,19 +396,19 @@ func (d *dirInode) checkInvariants() {
 }
 
 func (d *dirInode) lookUpChildFile(ctx context.Context, name string) (*Core, error) {
-	return findExplicitInode(ctx, d.Bucket(), NewFileName(d.Name(), name), false)
+	return findExplicitInode(ctx, d.Bucket(), NewFileName(d.Name(), name), false, d.IsTypeCacheDeprecated())
```
thrivikram-karur-g, line 399: Explain the type cache deprecation check in the summary of code changes in PR description, do this for all the changes for ease of review for the reviewer.

internal/fs/inode/dir.go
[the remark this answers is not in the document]
thrivikram-karur-g: We reverted this back to && and seems to work fine.

internal/fs/inode/dir.go
thrivikram-karur-g: This is no longer needed and the check has been removed.

internal/fs/inode/file.go
thrivikram-karur-g: No longer valid and it is now not being considered being an edge case for stat cache access to record metric due to further complications in implementation and scope creep.

internal/fs/inode/file.go
thrivikram-karur-g: no longer applicable and has been removed.

internal/storage/caching/fast_stat_bucket.go
thrivikram-karur-g: Needed, as we need entryStatus, detail variables in scope.

internal/storage/caching/fast_stat_bucket.go
thrivikram-karur-g: Needed, same as above.

metrics/helper.go
thrivikram-karur-g: Removing as metricHandle can never be nil.

thrivikram-karur-g:
Passed Integration tests.

thrivikram-karur-g:
Perf tests also have been run

internal/fs/inode/dir.go
```diff
@@ -704,14 +712,28 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
 		}
 
-		// 3. Both lookups resulted in cache hits (no cacheMiss errors) with no results found,
-		// indicating a negative cache entry. Return nil to indicate the entry doesn't exist from cache.
+		// 3. Negative entry check:
+		// Both lookups must be cache hits (no cacheMiss errors) with no results found for us to
+		// conclude the entry does not exist. If only one candidate is a negative hit, the other
+		// candidate may still exist in GCS, so we must fall through and query GCS.
 		if dirErr == nil && fileErr == nil {
+			d.metricHandle.MetadataCacheReadCount(1, true, metrics.EntryStatusNegativeAttr, metrics.LookupDetailFoundAttr)
```
kislaykishore: Why are we logging this here as opposed to logging it at the fs.go level?

kislaykishore: In case of retries, or operations such as renames, will we end up recording multiple metric events? If so, is that intentional?

internal/fs/inode/cache_outcome.go
```diff
@@ -0,0 +1,74 @@
...
+// itself has already warmed.
+//
+// This is a no-op when the context carries no outcome, which is the case for
+// unit tests that drive LookUpChild directly.
+//
+// Not safe for concurrent use. All call sites are on LookUpChild's own
+// goroutine; in particular none are inside the errgroup workers that
+// lookUpUnknownType spawns.
+func recordCacheOutcome(ctx context.Context, cacheHit bool, entryStatus metrics.EntryStatus, lookupDetail metrics.LookupDetail) {
+	outcome, ok := ctx.Value(cacheOutcomeKey{}).(*CacheOutcome)
```
thrivikram-karur-g: we can skip recording the cache outcome if metrics are disabled ...

thrivikram-karur-g: updated

internal/fs/inode/dir.go
thrivikram-karur-g: This was the core place where we were recording the logic hence captured the metric here but since retry case would not be solved here, moved the metric to context and captured it in LookUpChild method in fs.go, going further up in call stack would blow up the scope of changes a lot.

thrivikram-karur-g: wasn't intentional, rename works fine, retry needs moving capture further up in call stack, achieved using context changes as we discussed sometime ago. update the PR as well accordingly.

thrivikram-karur-g:
Updated the perf results run on the PR with the latest changes.

metrics/metrics_test_utils.go
```diff
@@ -101,7 +101,7 @@ func VerifyCounterMetric(t *testing.T, ctx context.Context, reader *metric.Manua
 	}
 
 	require.True(t, foundMetric, "metric %s not found", metricName)
-	require.Fail(t, "Data point for attributes %v not found in %s metric", attrs, metricName)
+	require.Failf(t, "data point not found", "no data point matching attributes %v in metric %s", attrs, metricName)
```
kislaykishore, line 104: what's the reason behind making these changes?

thrivikram-karur-g:
Re-Ran the integration tests with the updated PR changes and they passed.

internal/cache/metadata/stat_cache.go
```diff
@@ -56,6 +57,9 @@ type StatCache interface {
 	// entry, or the entry has expired according to the supplied current time.
 	LookUp(name string, now time.Time) (hit bool, m *gcs.MinObject)
 
+	// LookUpDetail returns the current object entry along with entry status and lookup detail.
+	LookUpDetail(name string, now time.Time) (hit bool, m *gcs.MinObject, entryStatus metrics.EntryStatus, detail metrics.LookupDetail)
```
kislaykishore, line 61: I find `EntryKind` or `EntryType` better than EntryStatus. Status should represent things like whether the entry is expired or not.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -599,11 +618,11 @@ func (b *fastStatBucket) GetFolder(ctx context.Context, req *gcs.GetFolderReques
...
```
kislaykishore, line 621: Why change if to `else if`? It adds more cognitive load. Same comment elsewhere.

internal/storage/caching/fast_stat_bucket.go
```diff
@@ -302,6 +305,14 @@ func (b *fastStatBucket) lookUp(name string) (hit bool, m *gcs.MinObject) {
 	return
 }
 
+func (b *fastStatBucket) lookUpDetail(name string) (hit bool, m *gcs.MinObject, entryStatus metrics.EntryStatus, detail metrics.LookupDetail) {
```
kislaykishore, line 308: Let's rename it `LookupFileDetail`

internal/fs/inode/dir.go
```diff
@@ -704,14 +708,35 @@ func (d *dirInode) LookUpChild(ctx context.Context, name string) (*Core, error)
...
+			recordMetadataCacheOutcome(ctx, true, metrics.EntryStatusPositiveAttr, metrics.LookupDetailFoundAttr)
 			return fileResult, nil
 		}
 
-		// 3. Both lookups resulted in cache hits (no cacheMiss errors) with no results found,
-		// indicating a negative cache entry. Return nil to indicate the entry doesn't exist from cache.
+		// 3. Negative entry check:
+		// Both lookups must be cache hits (no cacheMiss errors) with no results found for us to
+		// conclude the entry does not exist. If only one candidate is a negative hit, the other
+		// candidate may still exist in GCS, so we must fall through and query GCS.
```
kislaykishore, line 718: If the GCS call is still happening - should it be recorded as a cache-hit or a miss?

internal/fs/inode/metadata_cache_outcome.go
```diff
@@ -0,0 +1,84 @@
...
+//   - A cache miss means a GCS network call was made. A miss always trumps a hit
+//     so that any operation requiring a GCS call is classified as a cache miss
+//     overall.
+//   - Subsequent hits that merely observe an entry warmed by an earlier miss will
+//     not overwrite the miss.
+//
+// This is a no-op when the context carries no outcome, which is the case when
+// metrics are disabled or for unit tests that drive LookUpChild directly.
+//
+// Not safe for concurrent use. All call sites are on LookUpChild's own
```
kislaykishore, line 62: This looks fragile. While currently it might work, it's a maintenance hazard. I understand that performance could be a concern in which case, we should see how to fix that.  But without that, it could lead to hard-to-find bugs.

internal/storage/caching/fast_stat_bucket.go
thrivikram-karur-g: will keep variables out of if scope

